PR #1, opened 2017-07-14
Supporting Shadow Roots

e111077:
Adds support for `renderTo(this.shadowRoot)`


e111077:
Oops ignore this haha

e111077:
I can't believe I made the first PR to Lit and it was a fuckup. This is so embarrassingly Elliott

PR #7, opened 2017-07-26
`renderTo` overwrites if the container's current `TemplateInstance` has a different `Template`.

### LINKED ISSUE #6: Static text content of node does not change when re-rendered
Consider the following contrived usage of lit:

```javascript

html`<div>foo</div>`.renderTo(container);
html`<div>bar</div>`.renderTo(container);
```

## What I expected

```html
<div>
  <div>bar</div>
</div>
```

## What I got

```html
<div>
  <div>foo</div> <!-- The text content didn't change! -->
</div>
```

bicknellr:
Fixes #6.

### DIFF
--- a/src/labs/lit-extended.ts
+++ b/src/labs/lit-extended.ts
@@ -44,8 +44,10 @@ import { TemplateResult, AttributePart, TemplateInstance, TemplatePart, Part, Te
  *
  */
 export function renderExtendedTo(result: TemplateResult, container: Element|DocumentFragment) {
-  let instance = container.__templateInstance as ExtendedTemplateInstance|TemplateInstance|undefined;
-  if (instance && instance.template === result.template) {
+  let instance = container.__templateInstance as any;
+  if (instance &&
+      instance instanceof ExtendedTemplateInstance &&
+      instance.template === result.template) {
     instance.update(result.values);
     return;
   }
--- a/src/labs/lit-extended.ts
+++ b/src/labs/lit-extended.ts
@@ -44,16 +44,22 @@ import { TemplateResult, AttributePart, TemplateInstance, TemplatePart, Part, Te
  *
  */
 export function renderExtendedTo(result: TemplateResult, container: Element|DocumentFragment) {
-  let instance = container.__templateInstance as TemplateInstance;
-  if (instance === undefined) {
-    instance = new ExtendedTemplateInstance(result.template);
-    container.__templateInstance = instance;
-    const fragment = instance._clone();
+  let instance = container.__templateInstance as ExtendedTemplateInstance|TemplateInstance|undefined;
+  if (instance && instance.template === result.template) {
     instance.update(result.values);
-    container.appendChild(fragment);
-  } else {
-    instance.update(result.values);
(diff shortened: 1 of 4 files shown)


src/labs/lit-extended.ts
```diff
@@ -44,16 +44,22 @@ import { TemplateResult, AttributePart, TemplateInstance, TemplatePart, Part, Te
...
 export function renderExtendedTo(result: TemplateResult, container: Element|DocumentFragment) {
-  let instance = container.__templateInstance as TemplateInstance;
-  if (instance === undefined) {
-    instance = new ExtendedTemplateInstance(result.template);
-    container.__templateInstance = instance;
-    const fragment = instance._clone();
-    instance.update(result.values);
-    container.appendChild(fragment);
-  } else {
+  let instance = container.__templateInstance as ExtendedTemplateInstance|TemplateInstance|undefined;
```
bicknellr: This doesn't need `ExtendedTemplateInstance|`.

bicknellr:
This doesn't work for `ExtendedTemplateInstance`: if there was an old `TemplateInstance` attached to the container and you attempt to `renderExtendedTo` it, then it won't update the instance.

justinfagnani:
I'm confused by your last comment, are you saying this PR doesn't work in that case, or the code didn't before?

src/labs/lit-extended.ts
```diff
@@ -44,16 +44,22 @@ import { TemplateResult, AttributePart, TemplateInstance, TemplatePart, Part, Te
...
+  if (instance && instance.template === result.template) {
```
justinfagnani: I think you want to add an instanceof check for `ExtendedTemplateInstance`

src/lit-html.ts
```diff
@@ -48,16 +48,22 @@ export class TemplateResult {
...
-    let instance = container.__templateInstance as TemplateInstance;
-    if (instance === undefined) {
-      instance = new TemplateInstance(this.template);
-      container.__templateInstance = instance;
-      const fragment = instance._clone();
-      instance.update(this.values);
-      container.appendChild(fragment);
-    } else {
+    let instance = container.__templateInstance as TemplateInstance|undefined;
+    if (instance && instance.template === this.template) {
```
justinfagnani: also instanceof check here

bicknellr:
This PR doesn't work in that case. I'm about to push a fix though.

src/labs/lit-extended.ts
justinfagnani: also, explicitly check again `undefined`, it's faster because it skips a ToBoolean on `instance`

bicknellr: updated

bicknellr: updated

src/lit-html.ts
bicknellr: updated

bicknellr:
`renderExtendedTo` should now cause a new `ExtendedTemplateInstance` to be created, even if a plain `TemplateInstance` is attached to the container and has the same template.

PR #9, opened 2017-07-26
Optimizations

cdata:
Currently the only optimization is to prefer a simple list walking loop over `Range`. @sorvell to provide the performance improvement numbers.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -316,7 +316,7 @@ export class NodePart extends Part {
             // Since this is the last part we'll use, set it's endNode to the
             // container's endNode. Setting the value of this part will clean
             // up any residual nodes from a previously longer iterable.
-            this.clear(itemPart.endNode);
+            this.clear(itemPart.endNode.previousSibling!);
             itemPart.endNode = this.endNode;
           }
           itemEnd = itemPart.endNode;
@@ -348,12 +348,12 @@ export class NodePart extends Part {
     }
   }
 
-  clear(startNode :Node = this.startNode) {
+  clear(startNode: Node = this.startNode) {
     this._previousValue = undefined;
 
     let node = startNode.nextSibling!;
 
-    while (node && node !== this.endNode) {
+    while (node !== null && node !== this.endNode) {
       let next = node.nextSibling!;
       node.parentNode!.removeChild(node);
       node = next;
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -46,7 +46,7 @@ export class TemplateResult {
 
 /**
  * Renders a template to a container.
- * 
+ *
  * To update a container with new values, reevaluate the template literal and
  * call `render` with the new result.
  */
@@ -316,6 +316,9 @@ export class NodePart extends Part {
             // Since this is the last part we'll use, set it's endNode to the
             // container's endNode. Setting the value of this part will clean
             // up any residual nodes from a previously longer iterable.
+
+            // Remove previousSibling, since we want itemPart.endNode to be
+            // removed as part of the clear operation.
             this.clear(itemPart.endNode.previousSibling!);
             itemPart.endNode = this.endNode;
           }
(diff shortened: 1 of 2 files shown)


src/lit-html.ts
```diff
@@ -307,11 +307,7 @@ export class NodePart extends Part {
             // Since this is the last part we'll use, set it's endNode to the
             // container's endNode. Setting the value of this part will clean
             // up any residual nodes from a previously longer iterable.
-            const range = document.createRange();
-            range.setStartBefore(itemPart.endNode);
-            range.setEndBefore(this.endNode);
-            range.deleteContents();
-            range.detach();
+            this.clear(itemPart.endNode);
```
justinfagnani: This looks like a change in behavior. Previously the last itemPart's endNode was removed, and it's endNode property set to the container's endNode. Now it looks like, because of the `let node = startNode.nextSibling!;` line below, that the part's endNode will be left around

src/lit-html.ts
```diff
@@ -307,11 +307,7 @@ export class NodePart extends Part {
             // Since this is the last part we'll use, set it's endNode to the
             // container's endNode. Setting the value of this part will clean
             // up any residual nodes from a previously longer iterable.
```
justinfagnani: this comment is out of date, I missed updating it yesterday

src/lit-html.ts
```diff
@@ -343,15 +339,17 @@ export class NodePart extends Part {
     }
   }
 
-  clear() {
+  clear(startNode :Node = this.startNode) {
```
justinfagnani: formatting: no space after parameter name, and 1 space befre type

src/lit-html.ts
```diff
@@ -343,15 +339,17 @@ export class NodePart extends Part {
...
     this._previousValue = undefined;
-    const range = document.createRange();
-    range.setStartAfter(this.startNode);
-    range.setEndBefore(this.endNode);
-    range.deleteContents();
-    range.detach();
-  }
 
+    let node = startNode.nextSibling!;
```
justinfagnani: I think rather than a default parameter, you can do this to get the old behavior:

```ts
clear(startNode?: Node) {
  let node = startNode === undefined ? this.startNode.nextSibling! : startNode;
  ...
}
```

src/lit-html.ts
justinfagnani: It'd be nice to have a test that failed for this. I have a few where I check the childNodes of the container, not just innerHTML, so we're sure there's no orphan marker nodes hanging around

src/lit-html.ts
```diff
@@ -343,15 +339,17 @@ export class NodePart extends Part {
...
+
+    while (node && node !== this.endNode) {
```
justinfagnani: I'm trying to do explicit checks for `undefined` and `null` in here, especially in loops, on advice from the V8 team which says that it's faster by avoiding a ToBoolean conversion. I'm also using `===`, and relying somewhat on TypeScript (which has very good typings for the DOM) to yell when I check for the wrong nully value. So:
```ts
while (node !== null && node !== this.endNode) {
```

sorvell:
This has a pretty significant impact on a test that updates 5000 text bindings:
* before: ~250ms
* after: 15ms

src/lit-html.ts
cdata: Oops, I missed the subtle `setStartBefore` vs `setStartAfter`. Good catch.

src/lit-html.ts
cdata: Will fix.

src/lit-html.ts
cdata: Default value seems way cleaner to read than a branching assignment. Just my opinion.

src/lit-html.ts
cdata: Will fix.

src/lit-html.ts
justinfagnani: Seems fine, but maybe comment like 310 then

cdata: Done.

PR #10, opened 2017-07-26
Handle special case for setting text content

cdata:
@sorvell to confirm if bytesize trade-off is worth it.


sorvell:
Seems good. 

This has a minor impact on a test that updates 5000 text bindings:

with https://github.com/PolymerLabs/lit-html/pull/9: ~15ms
with https://github.com/PolymerLabs/lit-html/pull/9 + this PR: ~9ms

src/lit-html.ts
```diff
@@ -330,6 +330,9 @@ export class NodePart extends Part {
         itemStart = itemEnd;
       }
       this._previousValue = itemParts;
+    } else if (this.startNode.nextSibling! === this.endNode.previousSibling! &&
+        this.startNode.nextSibling!.nodeType === Node.TEXT_NODE) {
+      this.startNode.nextSibling!.textContent = value;
```
justinfagnani: Can we make sure we have a test that updates an expression from a simple value to a complex one. I think it'll pass, but I don't see it covered yet. Something like:

```typescript
let value = 'foo';
const t = () => html`<div>${value}</div>`;
render(t(), container);
assert.equals(container.innerHTML, '<div>foo</div>');

value = html`<span>bar</span>`;
render(t(), container);
assert.equals(container.innerHTML, '<div><span>bar</span></div>');
```

justinfagnani:
I think 40% is very worth it :)

src/lit-html.ts
cdata: Will fix.

cdata:
Gonna go ahead and merge this into the other PR.

matthewp:
As far as size, I always just inline `3` instead of `Node.TEXT_NODE`, etc. those are easy enough to remember. You only have a couple of instances of constant usage in the code though.

PR #14, opened 2017-07-29
Add support for Promises

justinfagnani:
Simple change, but seems to handle the basics correctly.

Still need to decide what rejections should do. Currently they do nothing, should they could clear the part? Right now a template author realistically would have to explicitly handle rejections, which might be good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,8 +162,6 @@ const render = () => html`
 `;
 ```
 
-These features compose so you can include iterables of thunks that return arrays of nested templates, etc...
-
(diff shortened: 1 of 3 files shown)


TimvdLippe:
I like the explicit handling of rejections. If not caught, they should produce errors in the console

justinfagnani:
@TimvdLippe VMs should be able to give the unhandled rejection warning on the console. An async stack trace should point devs directly to the expression that yielded the promise.

PR #18, opened 2017-07-29
Initial support for async iterables

### LINKED ISSUE #15: Async Iterable Support
Now that there's a PR for Promises (#14), we should think about async iterables.

Async iterables would allow for very easy to express incrementally rendered lists. To control how fast items are rendered into the list, just control how values are resolved into the iterable. Streaming server responses could also easily be rendered incrementally.

The part that's trickier over Promise support is that we can't just call `setValue()` in the callback anymore, we have to append new item Parts, and might have to give up on part reuse if it's too complex.

justinfagnani:
Fixes #15 

Add support for async iterables, which can be used to stream data into a part over time.

Combined with #14 this brings the gzipped+babili size to 2k, so I do wonder if they're worth it. It is a little bit tricky to implement this inside a template function because handing races requires statefulness.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -350,6 +350,24 @@ export class NodePart extends Part {
         itemStart = itemEnd;
       }
       this._previousValue = itemParts;
+    } else if (value && value[Symbol.asyncIterator]) {
+      this.clear();
+      (async () => {
+        for await (const v of value) {
+          let itemStart = this.startNode;
+          if (this.startNode.nextSibling !== this.endNode) {
+            itemStart = new Text();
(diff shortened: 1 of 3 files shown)


justinfagnani:
I'm going to close this in favor of adding support externally. Two reasons:

1. As small as this feature is, it's big enough that accumulating enough of these small features could double the size of the library.
2. It is a bit more opinionated than just Iterable support, due to the fact that async streams are often used to represent a single value changing over time, as opposed to an Array or Iterable which nearly always represents multiple values. So appending new content is not the only valid approach, and giving users control via a directive is probably the better option.

PR #19, opened 2017-07-31
Add until() to labs/

justinfagnani:
Lets templates display default content until a Promise resolves:

```js
html`<div>${until(promise, html`<span>loading...</span>`)}</div>`
```

### DIFF
--- a/src/test/labs/until_test.ts
+++ b/src/test/labs/until_test.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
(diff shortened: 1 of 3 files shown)


src/test/labs/until_test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const assert = chai.assert;
+
+suite('until', () => {
+
+  test('displays defaultContent immediately', async () => {
+    const container = document.createElement('div');
+    let resolve: (v: any) => void;
+    const promise = new Promise((res, _) => {resolve = res;});
+    render(html`<div>${until(promise, html`<span>loading...</span>`)}</div>`, container);
```
TimvdLippe: defaultContent can be a regular string as well right?

justinfagnani: It can be any supported value - TemplateResult, scalars, Iterables, Functions, etc...

PR #22, opened 2017-08-02
refactor NodePart::setValue

zhaoz:
* break into individual functions depending on type
 * improves readability
 * moves function args to be monomorphic
   * improved some private benchmarks by about 40%

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -265,102 +265,110 @@ export class NodePart extends Part {
     this.endNode = endNode;
   }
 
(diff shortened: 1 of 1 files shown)


zhaoz:
Some test are broken, but they seem to be broken on master as well.

justinfagnani:
Thanks @zhaoz!

All tests are passing on master for me. The Promise-based tests are failing on this PR, I suspect you just missed those changes.

justinfagnani:
I merged a version that passes tests in #23

PR #23, opened 2017-08-02
refactor NodePart::setValue

justinfagnani:
* break into individual functions depending on type
 * improves readability
 * moves function args to be monomorphic
   * improved some private benchmarks by about 40%


justinfagnani:
@zhaoz I fixed the tests and made some methods private. GOing to merge this one and close #22

Issue #24, opened 2017-08-02
Optimizations


justinfagnani:
Tracking issue for optimization ideas we want to try when we have time.

- [ ] - Change template instantiation for first render: Instead of TemplateInstance.new(), _clone(), update(), where clone imports the template, then does a tree walk of the clone to index the parts, then update() sets values, causing a bunch of insertNodes, combine into one method and do a tree walk of the template creating static nodes and parts values in one pass. This assumes appendChild is faster than insertNode.
- [x] - Dirty-checking for simple values. This should make updates way faster. PR #35
- [x] - Change the Part.setValue() API to take a list of values and an index to start from. This eliminates a slice() call. PR #42 
- [x] - Eliminate the near-full tree walk on instance first-render (alternate to first optimization). Maybe use childNodes like Polymer? PR #50 
- [ ] - Optimize for NodeParts that are the only child. They don't need start and end marker nodes.
- [x] - Micro-optimizations around ordering of if/else cascades, instanceof, etc.

justinfagnani:
FYI @sorvell @cdata @kevinpschaaf @zhaoz

developit:
You might find manually walking trees to be faster than TreeWalker for the specific case being used here.
I did some performance analysis here:
https://github.com/developit/strip-dom-whitespace/issues/1#issuecomment-267873172

justinfagnani:
Closing this, but opened #147 to track optimizing NodeParts that are the only child

PR #35, opened 2017-08-07
Add dirty-checking for primitive values

justinfagnani:
Seems to give a ~5% improvement on first render, and ~25% improvement on updates for a deep recursion benchmark.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -268,10 +268,32 @@ export class NodePart extends Part {
   setValue(value: any): void {
     value = this._getValue(value);
 
-    if (value instanceof Node) {
-      this._previousValue = this._setNode(value);
+    if (value === null ||
+        !(typeof value === 'object' || typeof value === 'function')) {
+      // Handle primitive values
+
+      // If the value didn't change, do nothing
+      if (value === this._previousValue) {
+        return;
+      }
+
+      if (this.startNode.nextSibling! === this.endNode.previousSibling! &&
+          this.startNode.nextSibling!.nodeType === Node.TEXT_NODE) {
+        // If we only have a single text node between the markers, we can just
+        // set its value, rather than replacing it.
+        // TODO(justinfagnani): Can we just check if _previousValue is
+        // primitive?
+        this.startNode.nextSibling!.textContent = value;
+        this._previousValue = value;
+      } else {
+        this._previousValue = this._setText(value);
+      }
     } else if (value instanceof TemplateResult) {
       this._previousValue = this._setTemplateResult(value);
+    } else if (value[Symbol.iterator]) {
+      this._previousValue = this._setIterable(value);
+    } else if (value instanceof Node) {
+      this._previousValue = this._setNode(value);
     } else if (value && value.then !== undefined) {
       value.then((v: any) => {
         if (this._previousValue === value) {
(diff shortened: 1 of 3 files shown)


justinfagnani:
Some of the first render improvement could have come from simply reordering the TemplateResult check to come earlier in the if/else cascade.

justinfagnani:
Addresses part of #24

src/lit-html.ts
```diff
@@ -268,23 +268,34 @@ export class NodePart extends Part {
   setValue(value: any): void {
     value = this._getValue(value);
 
-    if (value instanceof Node) {
-      this._previousValue = this._setNode(value);
+    if (value === undefined ||
```
rictic: Recommend documenting this branch. The rest are pretty clear but this one is doing a lot of subtle stuff.

TimvdLippe: For reference, [this](https://github.com/Polymer/polymer/blob/master/lib/mixins/property-accessors.html#L607) is the dirty check as implemented by Polymer

justinfagnani: Yes, this is trying to compare primitives only. I'll try that above.

@TimvdLippe that function is odd, it's using `===` compares even on objects

src/lit-html.ts
```diff
@@ -268,23 +268,34 @@ export class NodePart extends Part {
...
+        value === null ||
+        typeof value === 'string' ||
+        typeof value === 'number' ||
+        typeof value === 'boolean') {
+      if (value === this._previousValue) {
+        return;
+      }
+      if (this.startNode.nextSibling! === this.endNode.previousSibling! &&
+          this.startNode.nextSibling!.nodeType === Node.TEXT_NODE) {
```
justinfagnani: yes, as long as no one messes w/ the dom outside of lit-html's control

src/lit-html.ts
justinfagnani: done

justinfagnani:
@rictic PTAL

src/lit-html.ts
```diff
@@ -268,24 +268,41 @@ export class NodePart extends Part {
...
+        // If we only have a single text node between the markers, we can just
+        // set its value, rather than replacing it.
+        // TODO(justinfagnani): Can we just check if _previousValue is
+        // primitive?
+        this.startNode.nextSibling!.textContent = value;
+        this._previousValue = value;
+      } else {
+        this._previousValue = this._setText(value);
```
sorvell: should we make `_setText` do the optimization above?

src/lit-html.ts
```diff
@@ -268,24 +268,41 @@ export class NodePart extends Part {
   setValue(value: any): void {
     value = this._getValue(value);
 
-    if (value instanceof Node) {
-      this._previousValue = this._setNode(value);
+    if (value === null ||
+        !(typeof value === 'object' || typeof value === 'function')) {
```
sorvell: Why avoid for functions? Only for directives?

src/lit-html.ts
```diff
@@ -268,24 +268,41 @@ export class NodePart extends Part {
...
+        this._previousValue = value;
+      } else {
+        this._previousValue = this._setText(value);
+      }
     } else if (value instanceof TemplateResult) {
       this._previousValue = this._setTemplateResult(value);
+    } else if (value[Symbol.iterator]) {
+      this._previousValue = this._setIterable(value);
+    } else if (value instanceof Node) {
+      this._previousValue = this._setNode(value);
```
sorvell: Why no setter for the promise case? (below this, not changed here)

src/lit-html.ts
justinfagnani: Only because functions can have properties. I know boxed primitives can, but it seems very rare.

src/lit-html.ts
justinfagnani: yeah

src/lit-html.ts
justinfagnani: Just because it was so small. I can do that too

Issue #36, opened 2017-08-07
Single expression producing two sets of marker nodes


justinfagnani:
```javascript
html`<div>${foo}</div>`
```

is producing 4 marker nodes. Ultimately it should produce none in this case, but at least with the current approach I'd expect it to produce only 2.

justinfagnani:
Fixed by #39

PR #39, opened 2017-08-09
Use string literal parts as marker nodes

justinfagnani:
It turns out we don't need to explicitly create any marker nodes at all - we get them naturally from the string literal parts. When splitting a text node content with the marker sentinel, we're guaranteed to get strings at the beginning, end, and in-between every expression - exactly where we need markers.

This simplifies template prep, makes it easier to understand, and should speed up template prep and cloning, thought I didn't see much of a change on a deeply recursive first render test. It reduces the minified+gzipped size by 30 bytes.

A follow up optimization could be to eliminate the first and last nodes when they're empty, ie #36. Given this didn't change times too much, maybe it's not worth it.


justinfagnani:
@sorvell PTAL

PR #40, opened 2017-08-09
Fix until test

justinfagnani:
For some reason the until test was failing on Chrome, but only when devtools were closed :/ It was passing fine on Safari and Firefox.

Waiting one event loop turn fixed it.

### DIFF
--- a/src/test/lib/until_test.ts
+++ b/src/test/lib/until_test.ts
@@ -30,6 +30,7 @@ suite('until', () => {
     assert.equal(container.innerHTML, '<div><span>loading...</span></div>');
     resolve!('foo');
     await promise;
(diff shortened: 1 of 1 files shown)


src/test/lib/until_test.ts
```diff
@@ -30,6 +30,7 @@ suite('until', () => {
     assert.equal(container.innerHTML, '<div><span>loading...</span></div>');
     resolve!('foo');
     await promise;
+    await new Promise((r) => setTimeout(() => r()));
```
valdrinkoshi: nitty nit: are all these bindings needed? should be enough to just do this
```js
await new Promise(r => setTimeout(r));

PR #43, opened 2017-08-09
Add the directive() function, don’t treat function values as directives by default

### LINKED ISSUE #29: Better ergonomics around Function values
Function values are a sharp edge of the API. It's very easy to pass a function to a template expecting it to be used as an event handler, when instead it's invoked immediately as a template extension.

One possible way around this is to not invoke functions during rendering unless they have a special property, which could be set by a utility function:

```typescript
import { directive } from '../lit-html.js';

export repeat = directive((part: NodePart) => { ... }
```

Since function literals written by template authors are probably going to be written many, many times more than directives, it makes sense to optimize for their experience.

The one downside it this makes guarded (exception-safe) expressions longer. We'd have to vend a directive that ran a function immediately in a try/catch block:

```typescript
html`<input value=${safe(_=> this.foo.bar)}></input>`;
```

justinfagnani:
Fixes #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ HTML templates, via JavaScript template literals
 
 ### Examples
 
-`lit-html` can be used standalone and directly to help manage some DOM:
+`lit-html` can be used standalone and directly to help manage DOM:
 ```javascript
 const helloTemplate = (name) => html`<div>Hello ${name}!</div>`;
 
@@ -25,7 +25,7 @@ render(helloTemplate('Steve'), document.body);
 render(helloTemplate('Kevin'), document.body);
 ```
 
-But may also be common to use `lit-html` with a component system that calls `render()` for you, similar to React components:
+But it may also be common to use `lit-html` with a component system that calls `render()` for you, similar to React components:
 
 _(this example uses JS Class Fields, an upcoming specification)_
 ```javascript
@@ -52,7 +52,7 @@ class MyElement extends CoolLitMixin(HTMLElement) {
 `lit-html` has four main goals:
 
 1. Efficient updates of previously rendered DOM.
-2. Expressiveness and easy access the JavaScript state that needs to be injected into DOM.
+2. Expressiveness and easy access to the JavaScript state that needs to be injected into DOM.
 3. Standard JavaScript without required build steps, understandable by standards-compliant tools.
 4. Very small size.
 
@@ -68,7 +68,7 @@ A JavaScript template literal is a string literal that can have other JavaScript
 `My name is ${name}.`
 ``` 
 
-A _tagged_ template literal is preceded by a special template tag function:
+A _tagged_ template literal is prefixed with a special template tag function:
 
 ```javascript
 let name = 'Monica';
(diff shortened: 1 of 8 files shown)


src/test/lit-html_test.ts
```diff
@@ -121,17 +121,18 @@ suite('lit-html', () => {
...
-      test('renders thunks that throw as empty text', () => {
-        const container = document.createElement('div');
-        render(html`<div>${(_:any)=>{throw new Error('e')}}</div>`, container);
-        assert.equal(container.innerHTML, '<div></div>');
-      });
+      // test('renders thunks that throw as empty text', () => {
+      //   const container = document.createElement('div');
+      //   render(html`<div>${(_:any)=>{throw new Error('e')}}</div>`, container);
+      //   assert.equal(container.innerHTML, '<div></div>');
+      // });
```
cdata: Remove?

src/test/lit-html_test.ts
```diff
@@ -567,15 +569,15 @@ suite('lit-html', () => {
...
       });
 
-      test('accepts thunks that throw as empty text', () => {
-        part.setValue((_:any)=>{throw new Error('e')});
-        assert.equal(container.innerHTML, '');
-      });
+      // test('accepts thunks that throw as empty text', () => {
+      //   part.setValue((_:any)=>{throw new Error('e')});
+      //   assert.equal(container.innerHTML, '');
+      // });
```
cdata: Remove?

src/test/lit-html_test.ts
justinfagnani: done

src/test/lit-html_test.ts
justinfagnani: done

README.md
```diff
@@ -162,24 +162,45 @@ const render = () => html`
 `;
 ```
 
-### Function Values / Directives
+### Directives
 
-A function valued expression can be used for error handling and stateful rendering.
+Directives Functions are functions that can extend lit-html by directly interacting with the Part API.
```
sorvell: "Directives are functions"... seems like "Directive Function" isn't helping but if you insist on this name then reword to avoid "Functions are functions"

README.md
```diff
@@ -162,24 +162,45 @@ const render = () => html`
...
 
-If an expression returns a function, the function is called with the `Part` its populating, inside a try/catch block.
+Directive Functions are created from regular functions by passing them to the `directive()` function:
```
sorvell: "Directives are..."

README.md
```diff
@@ -162,24 +162,45 @@ const render = () => html`
...
 
-This makes it safe from exceptions:
+```javascript
+html`<div>${directive((part) => { part.setValue('Hello')})}</div>`
+```
+
+Directives Functions will often be returned from another function which takes arguments, which will generally be called the directive. That is, a directive is usually called inline in a template, and then returns a Directive Function to use a Part:
```
sorvell: "Directive Functions" thing again... and at the end do you want "to use _as_ a Part" ?

README.md
justinfagnani: fixed up

README.md
justinfagnani: changed

README.md
justinfagnani: removed

justinfagnani:
@sorvell PTAL

src/lib/lit-extended.ts
```diff
@@ -34,13 +34,11 @@ import { TemplateResult, AttributePart, TemplateInstance, TemplatePart, Part, Te
...
  *
- * To set an event handler, prefix the attribute name with `on-` and use a
- * function to return the handler, so that the event handler itself is not
- * called as a template directive.
+ * To set an event handler, prefix the attribute name with `on-`:
  *
  * Example:
  *
- *     html`<button on-click=${_=> this.onClickHandler}>Buy Now</button>`
+ *     html`<button on-click=${(e)=> this.onClickHandler(e)}>Buy Now</button>`
```
kenchris: When there is only one argument you don't need () ... e => this.onClickHandler(e) works fine

README.md
```diff
@@ -162,24 +162,43 @@ const render = () => html`
...
+```
 
-And is a useful extension point:
+This example increments a counter on every render:
 
-const render = () => html`<div>${(part) => part.setValue((part.previousValue + 1) || 0)}</div>`;
+```javascript
+const render = () => html`
+  <div>
+    ${directive((part) => part.setValue((part.previousValue + 1) || 0))}
```
kenchris: Couldn't you expand this so that it is clear where part comes from and where previousValue is defined?

src/lib/lit-extended.ts
justinfagnani: Style-wise I always include the parens. THey are optional, but I find it more consistent.

PR #46, opened 2017-08-11
Simplify the handling of Arrays and Iterables

justinfagnani:
This isn't a big performance improvement, but it really simplifies handling of iterables, using a for/of loop, rather than manually using the JS iteration protocol.

It should also make Array handling more easily backward compatible with IE11. We just need an Array.isArray() polyfill, and either just the existence of `Symbol` or an extra guard in case `Symbol` doesn't exist.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -231,7 +231,6 @@ export class AttributePart extends Part implements MultiPart {
 
   constructor(instance: TemplateInstance, element: Element, name: string, strings: string[]) {
     super(instance);
-    console.assert(element.nodeType === 1 /* ELEMENT_NODE */);
     this.element = element;
     this.name = name;
     this.strings = strings;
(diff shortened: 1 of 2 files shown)


justinfagnani:
Also fixes #51. Added regression test.

src/lit-html.ts
```diff
@@ -332,63 +331,51 @@ export class NodePart extends Part implements SinglePart {
...
-
-    if (current.done) {
-      // Empty iterable, just clear
+    // an Iterable, but it lets us recurse easily and efficiently update Arrays
+    // of TemplateResults that will be commonly returned from expressions like:
+    // array.map((i) => html`${i}`), by reusing existing TemplateInstances.
+
+    if (!Array.isArray(this._previousValue)) {
       this.clear();
+      this._previousValue = [];
```
sorvell: Is this the only spot the previousValue is set? Why not just always set it?

justinfagnani: Because I don't want to clear it if the previous values already is an array, since it will contain the NodeParts from the previous render. Adding a comment.

justinfagnani:
@sorvell PTAL

PR #49, opened 2017-08-12
Add unsafeHTML directive

justinfagnani:


src/lib/unsafe-html.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Renders the result as HTML, rather than text.
+ * 
+ * Note, this is unsafe to use with any user-provided input that hasn't been
+ * sanitized or escaped, as it may lead to cross-site-scripting
+ * vulnerabilities.
+ */
+export const unsafeHTML = (value: any) =>
+  directive((_part: NodePart) => {
+    tmp.innerHTML = value;
```
sorvell: You might want to perf check the difference between creating a new template here and setting innerHTML when there's significant existing content.

justinfagnani: Appears to make no difference. I'll go with the simpler approach of creating a new element each time.

justinfagnani:
@sorvell PTAL

Issue #45, opened 2017-08-11
bindings in <template> not working


valdrinkoshi:
```js
const who = 'world';
render(html`<template id="res">hello ${who}</template>`, document.body);

// `hello {{}}` instead of `hello world` :(
console.log(document.getElementById('res').innerHTML);
```

justinfagnani:
Oh, interesting...

I suppose we have to walk into nested template contents when preparing a template.

justinfagnani:
@valdrinkoshi #50 starts us on the path to being able to fix this. Because it does a recursive walk on the template, rather than using a TreeWalker, we can choose to special-case <template>s and recurse into them.

How critical is this?

valdrinkoshi:
The fact that we're using a template literal with a template element within it it's a bit strange, I know, but nevertheless I think it's important we support it for inter-op.

I've been updating [iron-overlay](https://github.com/PolymerLabs/iron-overlay) to use lit-html.
`<iron-overlay>` gets a template via distribution, stamps it and hosts it in a stacking-context-safe node (e.g. `<body>`). It relies on `<template>` so that the overlay content gets created and appended only when needed.
Another similar example is `<iron-list>`, which gets the item template via distribution.
If I want to use `<iron-list>`, I have to pass a template:
```js
const templateResult = html`
<iron-list items=${props.items}>
  <template>
    <div> lit-html binding: ${props.userName}, polymer binding: [[item]] </div>
  </template>
</iron-list>`;

**Comment by justinfagnani** [comment]
Well, with #50 I think we can do this, but it might be a bit confusing to deal with two separate data-binding systems. The lit-html bindings are going to change the `<template>`'s content. You have that right above, but I wonder how many would get that distinction immediately.

It's be nice to have an iron-list that took something even a bit more abstract than template elements, maybe a function that produces DOM (something like `(item, index) => Node`).

**Comment by mkorenko** [comment]
What about event binding?
```html
<iron-list items="${items}">
  <template>
    <div on-click="${(e) => this.onItemClick(e)}">
      [[item]]
    </div>
  </template>
</iron-list>
```
Does not work in #50

PR #50, opened 2017-08-13
Refactor node finding to use DFS instead of TreeWalker.

justinfagnani:
This is a somewhat speculative PR. It's performance neutral in a couple of microbenchmarks I'm trying out, but those microbenchmarks are pretty bad. It _should_ be perf positive with larger templates.

This is one optimization I mentioned in #9. An alternative is to not use cloning for template instantiation, but walk the template, appending into the container, finding parts, and inserting values in one pass. Ideally we'd try both, with some more realistic benchmarks, before committing, but considering that this one is at worst perf-neutral, and opens up the door to more easily fixing #45 it might be worthwhile to commit this anyway.

This does add ~100 bytes to the minified+gziped output size, at 1880b, putting us pretty close to 2kb 🤔

Edit: A couple of situations where this should clearly visit less nodes than TreeWalker:
- Static subtrees will be skipped, since no path terminates in them
- An element with a node binding as the first child can skip traversing the rest of its children

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -96,7 +96,7 @@ const exprMarker = '{{}}';
 export class TemplatePart {
   constructor(
     public type: string,
-    public index: number,
+    public path: number[],
     public name?: string,
     public rawName?: string,
     public strings?: string[]) {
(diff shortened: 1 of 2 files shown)


justinfagnani:
Closing this for now, but the code will be here for reference if we decide to investigate this route again.

PR #52, opened 2017-08-13
Change how rendering is extended to use a part callback instead of subclassing.

justinfagnani:


justinfagnani:
I'm down for a lot more experimentation here, let's brainstorm offline.

PR #53, opened 2017-08-18
Add SVG support & demo

justinfagnani:


justinfagnani:
Uh, I didn't actually merged this, just moved some branch pointers around that made it look like it

PR #57, opened 2017-08-18
Add SVG support & demo

justinfagnani:
Adds a new `svg` template tag, so you can declare that templates should produce nodes in the SVG namespace:

```javascript
html`
  <svg>
    <circle/>
    ${data.map(d => svg`<line ... />`)}
  </svg>
`
```

### DIFF
--- a/demo/clock.html
+++ b/demo/clock.html
@@ -1,0 +1,19 @@
+
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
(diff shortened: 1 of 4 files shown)


demo/clock.html
```diff
@@ -0,0 +1,8 @@
+<html>
```
azakus: License header please!

justinfagnani: done

kristoferjoseph:
Would this approach be needed to use `xlink` attributes?
Having trouble rendering svg symbols with:
```
  <svg>
    <use xlink:href="#${href}"/>
  </svg>
```

PR #75, opened 2017-08-27
Use workaround for TypeScript template literals bug

gdi2290:
https://github.com/Microsoft/TypeScript/issues/17956


justinfagnani:
Thanks for the PR @gdi2290!

I'd like to avoid a special function to use in TypeScript, and make a fix that works on other broken implementations of template literals. I was thinking of feature detecting proper template behavior, like so:

```ts
const t = () => ((s) => s)``;
const cachesTemplates = t() === t();
```

Then join the strings array if `cachesTemplates` is false.

src/lit-html.ts
```diff
@@ -12,21 +12,41 @@
...
 
 /**
  * Interprets a template literal as an HTML template that can efficiently
  * render to and update a container.
  */
 export function html(strings: TemplateStringsArray, ...values: any[]): TemplateResult {
-  let template = templates.get(strings);
+  let key: string = undefined;
+  let template: TemplateStringsArray = undefined;
+  if (shouldCacheTemplates) {
```
gdi2290: how about this @justinfagnani?

src/lit-html.ts
```diff
@@ -12,21 +12,39 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+/**
+ * TypeScript has a problem with precompiling templates literals
+ * https://github.com/Microsoft/TypeScript/issues/17956
+ */
+const shouldCacheTemplates = ((_t: any) => _t() === _t())(() => ((s: any) => s)``);
```
justinfagnani: I would name this `cachesTemplates`, as in "the browser caches template literals properly"

justinfagnani: `s: any` -> `s: TemplateStringsArray`

justinfagnani: no underscore for `_t`. We use leading underscores to mean a parameter is not used, but this is.

src/lit-html.ts
```diff
@@ -12,21 +12,39 @@
...
+
 // The first argument to JS template tags retain identity across multiple
 // calls to a tag for the same literal, so we can cache work done per literal
 // in a Map.
 const templates = new Map<TemplateStringsArray, Template>();
+const cacheTemplates = new Map<string, TemplateStringsArray>();
```
justinfagnani: We don't need a separate template cache, just change the key type on the first to `TemplateStringsArray|string`

src/lit-html.ts
```diff
@@ -12,21 +12,39 @@
...
 
 /**
  * Interprets a template literal as an HTML template that can efficiently
  * render to and update a container.
  */
 export function html(strings: TemplateStringsArray, ...values: any[]): TemplateResult {
+  if (shouldCacheTemplates) {
+    const _key = strings.join('{{typescriptProblems}}');
```
justinfagnani: since the Template constructor already joins the literals parts, we could just reuse that as the key:

```ts
let key = cachesTemplates ? strings : strings.join(exprMarker);
let template = templates.get(key);
if (template === undefined) {
  template = new Template(strings, cachesTemplates ? undefined : key)
  templates.set(key, template);
}
```

```ts
class Template {
  constructor(strings: TemplateStringsArray, html?: string) { ... }
}
```

gdi2290: `exprMarker` uses `Math.random()` which means the key will always be different. I also isolated the code with the assumption that it will be deleted when TypeScript follows the spec. We can also RegExp the `exprMarker` in the `Template` with one using `Math.random()`

gdi2290:
@justinfagnani I ended up keeping the two Maps due to typescript problems. you can edit it after the merge if you prefer one Map

justinfagnani:
What TypeScript problems did you hit?

gdi2290:
https://travis-ci.org/PolymerLabs/lit-html/builds/269422414?utm_source=github_status&utm_medium=notification
Since you're saying `template` can hold both TemplateStringsArray and string as the key you have to manually declare the types. If you manually declare the type you also need to say the value returned can also be undefined

justinfagnani:
@gdi2290 this works for me with no type warnings:

```ts
const cacheTemplates = ((t: any) => t() === t())(() => ((s: TemplateStringsArray) => s)``);

// The first argument to JS template tags retain identity across multiple
// calls to a tag for the same literal, so we can cache work done per literal
// in a Map.
const templates = new Map<TemplateStringsArray|string, Template>();

/**
 * Interprets a template literal as an HTML template that can efficiently
 * render to and update a container.
 */
export function html(strings: TemplateStringsArray, ...values: any[]): TemplateResult {
  const key = cacheTemplates ? strings.join('{{typescriptProblems}}') : strings;
  let template = templates.get(key);

  if (template === undefined) {
    template = new Template(strings);
    templates.set(key, template);
  }

  return new TemplateResult(template, values);
}```

**Comment by gdi2290** [comment]
@justinfagnani thanks for working with me on the PR. I made the correct changes

**Comment by justinfagnani** [comment]
@gdi2290 no problem! I'm glad to have contributors :)

This is almost there, but I think the conditional on whether to generate a key is backwards. We wouldn't know without tests that are compiled to ES5 though. I'll open an issue for that.

File: src/lit-html.ts [hunk 5]
```diff
@@ -12,22 +12,30 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+/**
+ * TypeScript has a problem with precompiling templates literals
+ * https://github.com/Microsoft/TypeScript/issues/17956
+ */
+const cacheTemplates = ((t: any) => t() === t())(() => ((s: TemplateStringsArray) => s)``);
```
justinfagnani: I think given that this is used backwards in the ternary below (and I did it too!), that it should have a clear comment and clearer name. I originally suggested `cachesTemplates` as in "the browser caches templates like it should", but that one extra "s" may be too subtle. Maybe `templateStringsAreCached`, so that the use reads like:

```ts
const key = templateStringsAreCached ? strings : strings.join('{{typescriptProblems}}');
```

src/lit-html.ts
```diff
@@ -12,22 +12,30 @@
...
 
 /**
  * Interprets a template literal as an HTML template that can efficiently
  * render to and update a container.
  */
-export function html(
-    strings: TemplateStringsArray, ...values: any[]): TemplateResult {
-  let template = templates.get(strings);
+export function html(strings: TemplateStringsArray, ...values: any[]): TemplateResult {
+  const key = cacheTemplates ? strings.join('{{typescriptProblems}}') : strings;
```
justinfagnani: This ternary is backwards. `cacheTemplates` is true when the environment has the correct behavior.

justinfagnani:
Actually, we don't need to compile all the tests to ES5 just to test this. We can include a snippet of compiled code in our tests:

TypeScript compiles this:

```ts
const t = () => html`abc${1}def`;
```

to this:

```js
var t = function () {
    return (_a = ["abc", "def"], _a.raw = ["abc", "def"], html(_a, 1));
    var _a;
};
```

So we can make a test like:

```ts
const t = function () {
    let _a;
    return (_a = ["abc", "def"], _a.raw = ["abc", "def"], html(_a, 1));
};
const result1 = t();
const result2 = t();
assert.strictEqual(result1.template, result2.template);
```

I changed `var`s to `const` or `let`, and put `let _a` before the return to respect the TDZ, but otherwise it's the same and will demonstrates the bug.

justinfagnani:
@gdi2290 if you open your PR to commits from maintainers I can add the fix and test

gdi2290:
@justinfagnani I added you to my fork so you can push changes

justinfagnani:
Merged! Thanks a lot @gdi2290!

justinfagnani:
To clarify for anyone following (especially @zhaoz you'll need this fix), this fixes usage in TypeScript only when lit-html and the client code are both compiled to ES5 with TypeScript, since the feature detection only fails if lit-html is itself compiled. This is the case for scenarios where the app is in charge of compiling all ES2015+ code to ES5 when it needs to and does so with the same compiler across the whole app.

PR #77, opened 2017-08-27
Setup clang-format, tslint and editorconfig

justinfagnani:


src/lib/repeat.ts
```diff
@@ -40,14 +43,16 @@ export function repeat<T>(items: Iterable<T>, keyFnOrTemplate: KeyFn<T>|ItemTemp
     let state = stateCache.get(part);
     if (state === undefined) {
       state = {
-        keyMap: keyFn && new Map(),
-        parts: [],
+        keyMap : keyFn && new Map(),
```
stramel: This is interesting, don't think I have seen anyone use this style before..

gdi2290: same, would prefer no space between the colon

src/lit-html.ts
```diff
@@ -394,17 +395,20 @@ export class NodePart implements SinglePart {
...
-  } else if (templatePart.type === 'node') {
-    return new NodePart(instance, node, node.nextSibling!);
-  }
-  throw new Error(`Unknown part type ${templatePart.type}`);
-}
+export type PartCallback = (instance: TemplateInstance,
+                            templatePart: TemplatePart, node: Node) => Part;
+
+export const defaultPartCallback =
+    (instance: TemplateInstance, templatePart: TemplatePart, node: Node):
```
gdi2290: not sure how this is better lol

stramel: I'm guessing it was to fix line length? /shrug

package.json
```diff
@@ -13,16 +13,19 @@
     "build": "tsc",
     "gen-docs": "typedoc --readme none --mode modules --excludeNotExported --excludePrivate --exclude **/*_test.ts --out ./docs/api .",
     "pretest": "npm run posttest; ln -s node_modules bower_components",
-    "test": "wct -l chrome",
+    "test": "npm run lint && wct -l chrome",
     "posttest": "rm -f bower_components",
-    "checksize": "uglifyjs lit-html.js -mc --toplevel | gzip -9 | wc -c"
+    "checksize": "uglifyjs lit-html.js -mc --toplevel | gzip -9 | wc -c",
+    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i",
```
rictic: I think this PR is missing the .clang-format file

package.json
```diff
@@ -13,7 +13,7 @@
     "build": "tsc",
     "gen-docs": "typedoc --readme none --mode modules --excludeNotExported --excludePrivate --exclude **/*_test.ts --out ./docs/api .",
     "pretest": "npm run posttest; ln -s node_modules bower_components",
-    "test": "wct -l chrome",
+    "test": "npm run lint && wct -l chrome",
```
rictic: I'd recommend running lint after tests. Most of the time your editor will tell you about lint errors as you work.

I would recommend building before testing though. Something like:

```
"test": "npm run build && wct -l chrom && npm run lint",
```

src/lib/repeat.ts
justinfagnani: Same, but every formatter I've tried - prettier, clang and tsfmt - have their warts. We can file bugs on clang.

src/lit-html.ts
justinfagnani: yep. You take the bad with the consistency

package.json
justinfagnani: done. fyi a lot of our projects have the previous test command

package.json
justinfagnani: Ahhh, much better!

src/lib/repeat.ts
justinfagnani: fixed with a .clang-format

justinfagnani:
@rictic PTAL

Issue #79, opened 2017-08-31
Tests don‘t test anything


surma:
The tests currently don’t run at all.

Adding `-p` to the `wct` invocation shows that some libraries don’t load and as a result no tests are being run.

![screen shot 2017-08-31 at 12 06 45 pm](https://user-images.githubusercontent.com/234957/29920335-f217a256-8e44-11e7-8cb7-bf2a5e3c9c6e.png)

justinfagnani:
Those errors are because we're still working on WCT for modules/npm and some of the WCT's automatically added dependencies aren't installed. They don't cause failures though, and the tests are running for me locally and on CI: https://travis-ci.org/PolymerLabs/lit-html 89 passing on master and they definitely fail for PRs that get something wrong.

What version of WCT are you running?

surma:
Huh, I just ran `npm i` (so `wct` is at 6.0.1) and it didn’t work. I just re-did that on a completely fresh clone and same thing is happening. I’ll investigate – if it runs on CI  the problem is probably on my end.

surma:
As far as I can tell, mocha/wct finishes before the tests have been loaded asynchronously. Seems like we are in some kind of race-condition here. I’ll see if I can reliably delay wct until the tests are loaded.

surma:
Never mind, not a race condition, but wct prefers Chrome Stable over Canary, which doesn’t have modules. Manually opening the tests in Canary makes them work. I’ll look into ensuring it’s Canary, although this might not be worth it considering M61 is around the corner. Closing this for now.

surma:
For completeness sake: I used `-l canary` for the time being.

PR #80, opened 2017-08-31
Switch to Karma and fix tests

surma:
This is a **proposed** fix for #79 

I didn’t know `wct`  so wasn’t quite comfortable fixing it within that. I made a switch to [Karma](https://karma-runner.github.io/). All files that end with `_test.js` are loaded and run automatically. Using Karma will also allow us to test in other browsers in the future with literally a one line edit in `karma.conf.js`.

Your call if you want this.


justinfagnani:
I'd like to stay with WCT because 1) It's working, at least for me and CI, 2) We're updating WCT to better work with modules and npm and this is the one project we're using it on with CI right now.

PR #84, opened 2017-09-01
Update LICENSE

stramel:
These changes allow GitHub to determine the License type and display it on the repo.


justinfagnani:
@azakus ? What's this file supposed to have? I think I accidentally chose some option in a tool to generate it. Should I replace it with, say: https://github.com/Polymer/polymer-analyzer/blob/master/LICENSE

stramel:
This was based off what GitHub generates for the LICENSE file from their templates. I'm wondering if it is triggering off that first line.

PR #87, opened 2017-09-02
feat(tsconfig): inlineSources

gdi2290:
> Emit the source alongside the sourcemaps within a single file; requires `--inlineSourceMap` or `--sourceMap` to be set.

https://www.typescriptlang.org/docs/handbook/compiler-options.html


justinfagnani:
Thanks!

PR #102, opened 2017-09-12
Fix double accounting of old parts.

### LINKED ISSUE #92: repeat() fails on multiple rerender
Here’s a test case:

```js
    test('can rerender multiple times', () => {
      let items = [1, 2, 3, 4, 5];
      const t = () => html`${repeat(items, (i: number) => html`
            <li>item: ${i}</li>`)}`;
      render(t(), container);
      render(t(), container);
      render(t(), container);
      render(t(), container);
      assert.equal(container.innerHTML, `<li>item: 1</li><li>item: 2</li><li>item: 3</li><li>item: 4</li><li>item: 5</li>`);
    })
```

Error:
screen shot 2017-09-04 at 2 31 30 pm

cdata:
Fixes #92

### DIFF
--- a/src/lib/repeat.ts
+++ b/src/lib/repeat.ts
@@ -57,7 +57,6 @@ export function repeat<T>(
 
     const itemParts = [];
     let index = 0;
-    let oldPartsIndex = 0;
     let currentMarker: Node|undefined;
 
     for (const item of items) {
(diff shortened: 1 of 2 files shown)


src/lib/repeat.ts
```diff
@@ -74,7 +73,7 @@ export function repeat<T>(
       // Try to reuse a part, either keyed or from the list of previous parts
       // if there's no keyMap
       let itemPart =
-          keyMap === undefined ? oldParts[oldPartsIndex++] : keyMap.get(key);
+          keyMap === undefined ? oldParts[0] : keyMap.get(key);
```
justinfagnani: I see how this fixes the bug, but I don't like the accounting code here, especially the splice + indexOf below. Seems like we shouldn't need both an array and map of old parts. If we have a map, I'd hope we don't need to search through list to find the part there. Hoping we can clean this up a lot :)

cdata: 👍 I'm assuming that most of this will be refactored out in a future change.

PR #103, opened 2017-09-13
Add a failing swap test

sunesimonsen:
This failing test was discovered by
https://github.com/sunesimonsen/lit-html-tests/blob/fix-repeat-92/test/repeat.spec.js#L18


sunesimonsen:
Superseeded by https://github.com/PolymerLabs/lit-html/pull/108 which is move appropriate now that fix-repeat-92 is merged. But it still seems to be a problem.

PR #108, opened 2017-09-22
Add a failing swap test

sunesimonsen:
This failing test was discovered by
https://github.com/sunesimonsen/lit-html-tests/blob/master/test/repeat.spec.js#L12

I first saw the problem when playing with a web component example where I was swapping colors randomly:

swap

### DIFF
--- a/src/test/lib/repeat_test.ts
+++ b/src/test/lib/repeat_test.ts
@@ -62,6 +62,24 @@ suite('repeat', () => {
       assert.strictEqual(children1[2], children2[0]);
     });
 
+    test('swaps are stable', () => {
+      const t = (items: number[]) =>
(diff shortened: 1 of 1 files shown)


justinfagnani:
Thanks for the tests! These were fixed in #161 and I just committed a version of them I made locally that failed pre-161 and pass now.

sunesimonsen:
Cool, I'll do a rerun with the property based test to see if it finds something new.

sunesimonsen:
The initial property based test is passing now for 1000 iterations swapping a random number of items. But When I tried to extend the test a bit I think I found an infinite loop https://github.com/PolymerLabs/lit-html/pull/173

PR #110, opened 2017-09-25
[TASK] Add a very simple demo for basic testing

daKmoR:
It's a very basic thing to test basic functionality in browsers. e.g. for working on IE/Edge.

### DIFF
--- a/demo/simple.html
+++ b/demo/simple.html
@@ -1,0 +1,42 @@
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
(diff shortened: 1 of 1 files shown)


demo/simple.html
```diff
@@ -0,0 +1,42 @@
...
+-->
+<!doctype html>
+<html>
+  <body>
+    <div id="demo"></div>
+    
+    <script type="module">
+      import {html, render} from '../lit-html.js';
+
+      var header = 'Initial Value (will change in 2 seconds)';
```
AliMd: It's better to use `let`

demo/simple.html
```diff
@@ -0,0 +1,42 @@
...
+      var demo = document.getElementById('demo');
```
AliMd: And `const`

daKmoR:
I now moved this here https://github.com/daKmoR/lit-html-demos
if some demos are still needed just let me know and I'm happy to provide a PR.

PR #112, opened 2017-09-25
Support for Edge/IE11

daKmoR:
just a small thing to support EDGE (tested on "Edge 40; EdgeHTML 15").

IE11 apparently renders the html but won't insert the values - needs some more investigation
=> maybe another PR if I can figure it out

@justinfagnani sooo this actually also work in IE11 :)
2 Limitations so far:
1) Last node can't be a node with a variable
```
// => works in IE11
return html`
	<h2>${header}</h2>
	<p>ending</p>
`;

// does not work in IE11
return html`
	<h2>${header}</h2>
`;
````
2) You need a polyfill for template tag and ES6 Symbols
=> so this should be somewhere noted

I will copy this to an Issue

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -144,7 +144,7 @@ export class Template {
     this.element = document.createElement('template');
     this.element.innerHTML = this._getHtml(strings, svg);
     const walker = document.createTreeWalker(
-        this.element.content, 5 /* elements & text */);
+        this.element.content, 5 /* elements & text */, undefined, false);
     let index = -1;
     let partIndex = 0;
     const nodesToRemove = [];
@@ -192,7 +192,7 @@ export class Template {
           // Generate a new text node for each literal section
           // These nodes are also used as the markers for node parts
           for (let i = 0; i < lastIndex; i++) {
-            parent.insertBefore(new Text(strings[i]), node);
+            parent.insertBefore(document.createTextNode(strings[i]), node);
             this.parts.push(new TemplatePart('node', index++));
           }
         } else if (!node.nodeValue!.trim()) {
@@ -346,7 +346,7 @@ export class NodePart implements SinglePart {
       // primitive?
       node.textContent = value;
     } else {
-      this._setNode(new Text(value));
+      this._setNode(document.createTextNode(value));
     }
     this._previousValue = value;
   }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by daKmoR** [comment]
oh no using createTextNode fails 2 tests... will take a look tomorrow - it's way past bed time already :p

File: src/lit-html.ts [hunk 1]
```diff
@@ -143,8 +143,9 @@ export class Template {
     this.svg = svg;
     this.element = document.createElement('template');
     this.element.innerHTML = this._getHtml(strings, svg);
+    // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
     const walker = document.createTreeWalker(
-        this.element.content, 5 /* elements & text */);
+        this.element.content, 5 /* elements & text */, (<any>null), false);
```
justinfagnani: style nit: prefer `as` casts: `null as any`

src/lit-html.ts
```diff
@@ -487,8 +488,9 @@ export class TemplateInstance {
     const fragment = document.importNode(this.template.element.content, true);
 
     if (this.template.parts.length > 0) {
+      // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
       const walker =
-          document.createTreeWalker(fragment, 5 /* elements & text */);
+          document.createTreeWalker(fragment, 5 /* elements & text */, (<any>null), false);
```
justinfagnani: `null as any`

src/lit-html.ts
```diff
@@ -346,7 +347,7 @@ export class NodePart implements SinglePart {
       // primitive?
       node.textContent = value;
     } else {
-      this._setNode(new Text(value));
+      this._setNode(document.createTextNode(value));
```
justinfagnani: `this._setNode(document.createTextNode(value === undefined ? '' : value));` will fix the two failing tests.

justinfagnani:
Thanks @daKmoR!

Note that Edge and IE have another problem where they don't iterate attributes in document order. I've been working on that a bit as well. We also need to get Edge testing up and running.

daKmoR:
very nice :) I love passing tests :)

justinfagnani:
Thanks a lot!

PR #115, opened 2017-09-28
Fix iterables not clearing previous values when updating with smaller lists

### LINKED ISSUE #106: Iterable cache isn't always set correctly
It seems there's a bug in the way iterables are rendered.

Currently this happens:
* Render 5 TemplateResult instances
* All 5 render correctly
* Render 1 TemplateResult instance (same template, new values)
* 1 TemplateResult is updated, 4 DOM nodes are [removed](https://github.com/PolymerLabs/lit-html/blob/9d6f581f77533cb47783d0271d6d968b1f7690de/src/lit-html.ts#L418)
* Render 5 TemplateResult instances
* `_previousValue` _still_ contains 5 instances, all 5 are [updated](https://github.com/PolymerLabs/lit-html/blob/9d6f581f77533cb47783d0271d6d968b1f7690de/src/lit-html.ts#L391)
* *4 new nodes are not appended*, so 1 is rendered and the remaining 4 which were previously removed never return

This seems to be because `_previousValue` is not re-computed when the clear() call runs (when extra DOM nodes are removed). So in future renders, lit-html believes 5 nodes still exist (they don't, only 1 is in the dom tree at this point).

justinfagnani:
Fixes #106

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -488,9 +488,10 @@ export class TemplateInstance {
     const fragment = document.importNode(this.template.element.content, true);
 
     if (this.template.parts.length > 0) {
-      // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
-      const walker =
-          document.createTreeWalker(fragment, 5 /* elements & text */, null as any, false);
+      // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be
+      // null
+      const walker = document.createTreeWalker(
+          fragment, 5 /* elements & text */, null as any, false);
 
       const parts = this.template.parts;
       let index = 0;
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -382,9 +382,9 @@ export class NodePart implements SinglePart {
       this._previousValue = [];
     }
 
-    // Lets of keep track of how many items we stamped so we can clear leftover
+    // Lets us keep track of how many items we stamped so we can clear leftover
     // items from a previous render
-    const itemParts = this._previousValue;
+    const itemParts = this._previousValue as any[];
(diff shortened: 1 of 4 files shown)


src/lit-html.ts
```diff
@@ -384,7 +384,7 @@ export class NodePart implements SinglePart {
 
     // Lets of keep track of how many items we stamped so we can clear leftover
```
cdata: `lets of => lets us`

src/lit-html.ts
```diff
@@ -416,6 +416,7 @@ export class NodePart implements SinglePart {
       this._previousValue = undefined;
     } else if (partIndex < itemParts.length) {
       const lastPart = itemParts[partIndex - 1];
+      itemParts.length = partIndex;
```
cdata: 🎉 the first time I have ever seen someone set length on an `Array` is today 🎂

src/test/lit-html_test.ts
```diff
@@ -451,8 +435,24 @@ suite('lit-html', () => {
         assert.equal(container.innerHTML, '<div>321</div>');
       });
 
+      test('updates arrays 2', () => {
```
cdata: Personal opinion: test titles like this ought to be more descriptive of what the test is actually doing.

src/lit-html.ts
justinfagnani: fixed

src/lit-html.ts
justinfagnani: first time for everything!

I added a comment

src/test/lit-html_test.ts
justinfagnani: fixed

Issue #116, opened 2017-09-30
Exception with whitespace next to a part and at the start of end of an element


justinfagnani:
Report from @zhaoz:

```js
window.dostuff = function(container) {
  const b = 'hello';
  render(html`<span>${b} </span>`, container);
}
```

 ```
lit-html.ts:405 Uncaught TypeError: Cannot read property 'previousSibling' of null
    at NodePart._setText (lit-html.ts:405)
    at NodePart.setValue (lit-html.ts:378)
    at TemplateInstance.update (lit-html.ts:543)
    at render (lit-html.ts:99)
    at render (lit-extended.ts:48)
    at window.dostuff (lit-thumbnail.js:86)
    at <anonymous>:1:1```

justinfagnani:
This was fixed in #117

PR #117, opened 2017-09-30
Only remove whitespace between, or at the start or end of, elements.

### LINKED ISSUE #114: Whitespace between parts is removed
This makes it hard/impossible to render two values with a space in between them.

Failing test.
```ts
      test('renders parts with whitespace between them', () => {
        render(html`<div>${'foo'} ${'bar'}</div>`, container);
        console.log(container.innerHTML);
        assert.equal(container.innerHTML, '<div>foo bar</div>');
      });
```

justinfagnani:
Fixes #114 and #116

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -201,7 +201,8 @@ export class Template {
           }
         }
       } else if (node.nodeType === 3 /* Node.TEXT_NODE */) {
-        const strings = node.nodeValue!.split(attributeMarker);
+        const nodeValue = node.nodeValue!;
+        const strings = nodeValue.split(attributeMarker);
         if (strings.length > 1) {
           const parent = node.parentNode!;
           const lastIndex = strings.length - 1;
@@ -220,10 +221,18 @@ export class Template {
             parent.insertBefore(document.createTextNode(strings[i]), node);
             this.parts.push(new TemplatePart('node', index++));
           }
-        } else if (!node.nodeValue!.trim()) {
-          nodesToRemove.push(node);
-          currentNode = previousNode;
-          index--;
+        } else {
+          // Strip whitespace-only nodes, only between elements, or at the
+          // beginning or end of elements.
+          const previousSibling = node.previousSibling;
+          const nextSibling = node.nextSibling;
+          if ((previousSibling === null || previousSibling.nodeType === 1)
+              && (nextSibling === null || nextSibling.nodeType === 1)
+              && !nodeValue.trim()) {
+            nodesToRemove.push(node);
+            currentNode = previousNode;
+            index--;
+          }
         }
       } else if (
           node.nodeType === 8 /* Node.COMMENT_NODE */ &&
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -160,17 +160,19 @@ export class Template {
     // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
     const walker = document.createTreeWalker(
         this.element.content,
-        133 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT */,
-        null as any, false);
+        133 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT */
+        ,
+        null as any,
+        false);
     let index = -1;
     let partIndex = 0;
     const nodesToRemove: Node[] = [];
 
-    // The actual previous node, accounting for removals: if a node is removed it will
-    // never be the previousNode.
-    let previousNode: Node | undefined;
+    // The actual previous node, accounting for removals: if a node is removed
+    // it will never be the previousNode.
+    let previousNode: Node|undefined;
     // Used to set previousNode at the top of the loop.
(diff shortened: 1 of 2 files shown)


src/lit-html.ts
```diff
@@ -220,10 +221,18 @@ export class Template {
...
-        } else if (!node.nodeValue!.trim()) {
-          nodesToRemove.push(node);
-          currentNode = previousNode;
-          index--;
+        } else {
+          // Strip whitespace-only nodes, only between elements, or at the
+          // beginning or end of elements.
+          const previousSibling = node.previousSibling;
+          const nextSibling = node.nextSibling;
+          if ((previousSibling === null || previousSibling.nodeType === 1)
```
cdata: Please indicate what node type this is as you have done elsewhere in the source.

src/lit-html.ts
```diff
@@ -220,10 +221,18 @@ export class Template {
             parent.insertBefore(document.createTextNode(strings[i]), node);
             this.parts.push(new TemplatePart('node', index++));
           }
-        } else if (!node.nodeValue!.trim()) {
-          nodesToRemove.push(node);
-          currentNode = previousNode;
-          index--;
+        } else {
+          // Strip whitespace-only nodes, only between elements, or at the
+          // beginning or end of elements.
```
cdata: I wonder if these rules are even satisfactory to match what the browser might do with whitespace.

src/lit-html.ts
```diff
@@ -220,10 +221,18 @@ export class Template {
...
+          const previousSibling = node.previousSibling;
+          const nextSibling = node.nextSibling;
+          if ((previousSibling === null || previousSibling.nodeType === 1)
+              && (nextSibling === null || nextSibling.nodeType === 1)
+              && !nodeValue.trim()) {
```
cdata: I find it stylistically unexpected that there are such explicit checks for null and yet we rely on coercion in this case. No action required.

src/test/lit-html_test.ts
```diff
@@ -366,7 +354,9 @@ suite('lit-html', () => {
               ${'baz'}
               <p>${'qux'}</p>
             </div>`, container);
-        assert.equal(container.innerHTML, `<div foo="bar">baz<p>qux</p></div>`);
+        assert.equal(container.innerHTML, `<div foo="bar">
+              baz
+              <p>qux</p></div>`);
```
cdata: The name of this test could stand to be improved. It is a fair - albeit vague - description of what is being tested, but there is no reason why "stuff" _must_ include significant whitespace. A better test name will help to preserve such details as the test changes over time.

cdata: It may even make sense to break this into a couple of tests that each focus on a particular detail of rendering complex "stuff." For example, one test that focuses on nesting and one that focuses on multiline significant whitespace.

src/lit-html.ts
```diff
@@ -220,10 +221,18 @@ export class Template {
             parent.insertBefore(document.createTextNode(strings[i]), node);
             this.parts.push(new TemplatePart('node', index++));
           }
-        } else if (!node.nodeValue!.trim()) {
-          nodesToRemove.push(node);
-          currentNode = previousNode;
-          index--;
+        } else {
```
cdata: These conditions are starting to become dense, rendering the `while` loop difficult to consider in its entirety. Consider factoring these conditional paths into methods for readability.

zhaoz: +1

src/lit-html.ts
zhaoz: or use an enum.

src/lit-html.ts
justinfagnani: The difficulty in splitting out methods is the amount of mutable state they share. It would require either a lot of parameters and return values, plus assignments to the shared state, or splitting into closures within the constructor, which might create more closure allocation per template.

I'd like to take a look at refactoring separately because of the amount of changes it'd require.

src/lit-html.ts
justinfagnani: Any whitespace stripping like this is going to slightly change layout. @kevinpschaaf added the original stripping as a perf optimization, and this is backing some of it out for the bug fixes. Soon I think we want to look at how we might offer some control over this, but I think this is a better middle ground.

src/lit-html.ts
justinfagnani: added the comment

src/lit-html.ts
justinfagnani: This is a good point. Done

src/test/lit-html_test.ts
justinfagnani: I completely agree, and this test _might_ be obsolete by now, but at the time I wrote it I had very little coverage of specific scenarios that this was catching. This was serving as a canary for situations I didn't know to test yet.

Since adding it I've tried to add a new test for every time I break this one, which is basically your second comment. I think for a better version of this kind of coverage we should add some larger real-world-style templates and goldens.

justinfagnani:
@cdata PTAL

justinfagnani:
I need to get a release out with these fixes. Going to merge on @zhaoz's approval, but @cdata if you want more changes I can make another PR.

cdata:
@justinfagnani Please file issues that document the outstanding comments.

PR #118, opened 2017-10-01
Add asyncAppend and asyncReplace directives

### LINKED ISSUE #15: Async Iterable Support

justinfagnani:
This also updates the directive API, so that directives no longer return values, but have to call the part.setValue() API.

Fixes #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,48 @@ const render = () => html`
 `;
 ```
 
+#### `asyncAppend(asyncIterable)` and `asyncReplace(asyncIterable)`
+
+`asyncAppend` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
+appending each new value after the previous.
+
+`asyncReplace` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
+replacing the previous value with the new value.
+
+Example:
+
+```javascript
+/**
+ * Returns an async iterable that yields the number of seconds since `date`,
+ * approximately every second.
+ */
+async function* secondsAgo(date) {
+  const seconds = (Date.now() - date.getTime()) / 1000;
+  const wholeSeconds = Math.floor(seconds);
+  yield wholeSeconds;
+  yield* await new Promise((res) => {
+    setTimeout(() => {
+      res(secondsAgo(date));
+    }, 1000 - (seconds - wholeSeconds) * 1000);
+  });
+}
+
+render(html`
+  This example started ${asyncReplace(secondsAgo(new Date()))} seconds ago.
+`, document.body);
+```
+
+In the near future, `ReadableStream`s will be async iterables, enabling streaming `fetch()`
+directly into a template:
+
+```javascript
+const streamingResponse = (async () => {
+  const response = await fetch('https://cors-anywhere.herokuapp.com/http://stuff.mit.edu/afs/sipb/contrib/pi/pi-billion.txt');
+  return response.body.getReader();
+})();
+render(html`π is: ${asyncAppend(streamingResponse)}`, document.body);
+```
+
 ### Promises
 
 Promises are rendered when they resolve, leaving the previous value in place until they do. Races are handled, so that if an unresolved Promise is overwritten, it won't update the template when it finally resolves.
--- a/README.md
+++ b/README.md
@@ -167,6 +167,16 @@ const render = () => html`
 `;
 ```
 
+### Promises
+
+Promises are rendered when they resolve, leaving the previous value in place until they do. Races are handled, so that if an unresolved Promise is overwritten, it won't update the template when it finally resolves.
+
+```javascript
+const render = () => html`
+  The response is ${fetch('sample.txt').then((r) => r.text())}.
+`;
+```
+
 ### Directives
 
 Directives are functions that can extend lit-html by directly interacting with the Part API.
@@ -242,54 +252,49 @@ const render = () => html`
 
 #### `asyncAppend(asyncIterable)` and `asyncReplace(asyncIterable)`
 
-`asyncAppend` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
-appending each new value after the previous.
+JavaScript asynchronous iterators provide a generic interface for asynchronous sequential access to data. Much like an iterator, a consumer requests the next data item with a a call to `next()`, but with asynchronous iterators `next()` returns a `Promise`, allowing the iterator to provide the item when it's ready.
 
-`asyncReplace` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
+lit-html offers two directives to consume asynchronous iterators:
+
+ * `asyncAppend` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
+appending each new value after the previous.
+ * `asyncReplace` renders the values of an [async iterable](https://github.com/tc39/proposal-async-iteration),
 replacing the previous value with the new value.
 
 Example:
 
 ```javascript
+
+const wait = (t) => new Promise((resolve) => setTimeout(resolve, t));
+
 /**
- * Returns an async iterable that yields the number of seconds since `date`,
- * approximately every second.
+ * Returns an async iterable that yields increasing integers.
  */
-async function* secondsAgo(date) {
-  const seconds = (Date.now() - date.getTime()) / 1000;
-  const wholeSeconds = Math.floor(seconds);
-  yield wholeSeconds;
-  yield* await new Promise((res) => {
-    setTimeout(() => {
-      res(secondsAgo(date));
-    }, 1000 - (seconds - wholeSeconds) * 1000);
-  });
+async function* countUp() {
+  let i = 0;
+  while (true) {
+    yield i++;
+    await wait(1000);
+  }
 }
 
 render(html`
-  This example started ${asyncReplace(secondsAgo(new Date()))} seconds ago.
+  Count: <span>${asyncReplace(countUp())}</span>.
 `, document.body);
 ```
 
-In the near future, `ReadableStream`s will be async iterables, enabling streaming `fetch()`
-directly into a template:
+In the near future, `ReadableStream`s will be async iterables, enabling streaming `fetch()` directly into a template:
 
 ```javascript
+// Endpoint that returns a billion digits of PI, streamed.
+const url =
+    'https://cors-anywhere.herokuapp.com/http://stuff.mit.edu/afs/sipb/contrib/pi/pi-billion.txt';
+
 const streamingResponse = (async () => {
-  const response = await fetch('https://cors-anywhere.herokuapp.com/http://stuff.mit.edu/afs/sipb/contrib/pi/pi-billion.txt');
+  const response = await fetch(url);
   return response.body.getReader();
 })();
 render(html`π is: ${asyncAppend(streamingResponse)}`, document.body);
-```
-
-### Promises
-
-Promises are rendered when they resolve, leaving the previous value in place until they do. Races are handled, so that if an unresolved Promise is overwritten, it won't update the template when it finally resolves.
-
-```javascript
-const render = () => html`
-  The response is ${fetch('sample.txt').then((r) => r.text())}.
-`;
 ```
 
 ### Composability
--- a/src/lib/async-append.ts
+++ b/src/lib/async-append.ts
@@ -1,0 +1,88 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {directive, NodePart} from '../lit-html.js';
+
+/**
+ * A directive that renders the items of an async iterable[1], appending new
+ * values after previous values, similar to the built-in support for iterables.
+ *
+ * Async iterables are objects with a [Symbol.asyncIterator] method, which
+ * returns an iterator who's `next()` method returns a Promise. When a new
+ * value is available, the Promise resolves and the value is appended to the
+ * Part controlled by the directive. If another value other than this
+ * directive has been set on the Part, the iterable will no longer be listened
+ * to and new values won't be written to the Part.
+ *
+ * [1]: https://github.com/tc39/proposal-async-iteration
+ *
+ * @param value An async iterable
+ * @param mapper An optional function that maps from (value, index) to another
+ *     value. Useful for generating templates for each item in the iterable.
+ */
+export const asyncAppend = <T>(
+    value: AsyncIterable<T>, mapper?: (v: T, index?: number) => any) =>
+    directive(async (part: NodePart) => {
+      // If we've already set up this particular iterable, we don't need
+      // to do anything.
+      if (value === part._previousValue) {
+        return;
+      }
+      part.clear();
+      part._previousValue = value;
+
+      // We keep track of item Parts across iterations, so that we can
+      // share marker nodes between consecutive Parts.
+      let itemPart;
+      let i = 0;
+
+      for await (let v of value) {
+        // Check to make sure that value is the still the current value of
+        // the part, and if not bail because a new value owns this part
+        if (part._previousValue !== value) {
+          break;
+        }
+
+        // As a convenience, because functional-programming-style
+        // transforms of iterables and async iterables requires a library,
+        // we accept a mapper function. This is especially convenient for
+        // rendering a template for each item.
+        if (mapper !== undefined) {
+          v = mapper(v, i);
+        }
+
+        // Like with sync iterables, each item induces a Part, so we need
+        // to keep track of start and end nodes for the Part.
+        // Note: Because these Parts are not updatable like with a sync
+        // iterable (if we render a new value, we always clear), it may
+        // be possible to optimize away the Parts and just re-use the
+        // Part.setValue() logic.
+
+        let itemStartNode = part.startNode;
+
+        // Check to see if we have a previous item and Part
+        if (itemPart !== undefined) {
+          // Create a new node to separate the previous and next Parts
+          itemStartNode = document.createTextNode('');
+          // itemPart is currently the Part for the previous item. Set
+          // it's endNode to the node we'll use for the next Part's
+          // startNode.
+          itemPart.endNode = itemStartNode;
+          part.endNode.parentNode!.insertBefore(itemStartNode, part.endNode);
+        }
+        itemPart = new NodePart(part.instance, itemStartNode, part.endNode);
+        itemPart.setValue(v);
+        i++;
+      }
+    });
--- a/src/lib/async-append.ts
+++ b/src/lib/async-append.ts
@@ -39,7 +39,6 @@ export const asyncAppend = <T>(
       if (value === part._previousValue) {
         return;
       }
-      part.clear();
       part._previousValue = value;
 
       // We keep track of item Parts across iterations, so that we can
@@ -48,6 +47,12 @@ export const asyncAppend = <T>(
       let i = 0;
 
       for await (let v of value) {
+        // When we get the first value, clear the part. This let's the previous
+        // value display until we can replace it.
+        if (i === 0) {
+          part.clear();
+        }
+
         // Check to make sure that value is the still the current value of
         // the part, and if not bail because a new value owns this part
         if (part._previousValue !== value) {
--- a/src/lib/async-replace.ts
+++ b/src/lib/async-replace.ts
@@ -1,0 +1,73 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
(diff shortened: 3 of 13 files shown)


justinfagnani:
cc @jakearchibald how's this API look for your streaming use cases?

justinfagnani:
@sorvell adding you as a reviewer since you've used this branch

src/test/lib/test-async-iterable.ts
```diff
@@ -0,0 +1,50 @@
...
+  Object.defineProperty(Symbol, 'asyncIterator', {value: Symbol()});
+}
+
+export class TestAsyncIterable<T> implements AsyncIterable<T> {
+  /**
+   * A Promise that resolves with the next value to be returned by the
+   * async iterable returned from iterable()
+   */
+  private _nextValue: Promise<T> =
+      new Promise((res, _) => this._resolveNextValue = res);
```
cdata: Inconsistent naming here. `res` is used vs `resolve` bellow. In general, I prefer full words over abbreviations.

src/test/lib/test-async-iterable.ts
```diff
@@ -0,0 +1,50 @@
...
+        new Promise((ressolve, _) => this._resolveNextValue = ressolve);
+    // Resolves the previous value of _nextValue (now currentValue in this
+    // scope), making `yield await this._nextValue` go.
+    currentResolveValue(value);
+    // Waits for the value to be emitted
+    await currentValue;
+    // Need to wait for one more microtask for value to be rendered, but only
+    // when devtools is closed. Waiting for rAF might be more reliable, but
+    // this waits the minimum that seems reliable now.
+    await Promise.resolve();
```
cdata: This is subtle. It might be better for this to be awaited explicitly in the test, so that it is clear to a reader how timing works in practice.

src/test/lib/async-replace_test.ts
```diff
@@ -0,0 +1,111 @@
...
+    assert.equal(container.innerHTML, '<div>hello</div>');
+
+    await iterable.push('bar');
+    assert.equal(container.innerHTML, '<div>hello</div>');
+  });
+
+  test('renders new value over a pending iterable', async () => {
+    const t = (v: any) => html`<div>${v}</div>`;
+    // This is a little bit of an odd usage of directives as values, but it
+    // it possible, and we check here that asyncAppend plays nice in this case
```
cdata: `it it => it is`

cdata: `asyncAppend => asyncReplace`

src/test/lib/until_test.ts
```diff
@@ -22,20 +22,56 @@ const assert = chai.assert;
...
+    assert.equal(container.innerHTML, '<div>bar</div>');
+
+    deferred.resolve('foo');
+    await deferred.promise;
+    assert.equal(container.innerHTML, '<div>bar</div>');
+  });
+
 });
+
+class Deferred<T> {
```
cdata: Consider splitting this out into a test helper library.

src/test/lib/async-replace_test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  let container: HTMLDivElement;
+  let iterable: TestAsyncIterable<string>;
+
+  setup(() => {
+    container = document.createElement('div');
+    iterable = new TestAsyncIterable<string>();
+  });
+
+  test('renders an async iterable', async () => {
```
cdata: This actually tests a very particular rendering strategy. It's more like, "replaces content as the async iterable yields new values".

src/test/lib/async-replace_test.ts
```diff
@@ -0,0 +1,111 @@
...
+    assert.equal(container.innerHTML, '<div></div>');
+
+    await iterable.push('foo');
+    assert.equal(container.innerHTML, '<div>foo</div>');
+
+    await iterable.push('bar');
+    assert.equal(container.innerHTML, '<div>bar</div>');
+  });
+
+  test('renders an async iterable containing undefined', async () => {
```
cdata: It looks like many of these test names were copied from `asyncRender`. However, they deserve descriptions specific to the directive under test. Try to think of these test names as a form of documentation.

src/test/lib/async-append_test.ts
```diff
@@ -0,0 +1,111 @@
...
+import {TestAsyncIterable} from './test-async-iterable.js';
+
+const assert = chai.assert;
+
+// Set Symbol.asyncIterator on browsers without it
+if (typeof Symbol !== undefined && Symbol.asyncIterator === undefined) {
+  Object.defineProperty(Symbol, 'asyncIterator', {value: Symbol()});
+}
+
+suite('asyncAppend', () => {
```
cdata: I would like to see a test here for rendering lists of element nodes. It seems like a common case, and also one that has tripped us up in the past (particularly when it comes to getting the intersections between two lists right).

cdata:
Given that we have gone through a lot of exploration around state management, and have much yet to do, I would like to have a better understanding of which existing state management "paradigms du jour" advocate and/or benefit from async action when rendering.

Could you include some examples in the PR description before we merge?

justinfagnani:
Rebased to pick up the new Travis config and testing  on FF, Safari and Edge.

justinfagnani:
> Given that we have gone through a lot of exploration around state management, and have much yet to do, I would like to have a better understanding of which existing state management "paradigms du jour" advocate and/or benefit from async action when rendering.

I'm not sure if they do or not... I'm thinking about this less in terms of existing state management and more in terms of supporting the relatively primitive platform types, which async iterables are now, much like Promises.

There are useful cases this enables that don't use any state management solution or framework. For example streaming fetch (when ReadableStreams become async iterables):

```js
const streamingResponse = (async () => {
  const response = await fetch('https://cors-anywhere.herokuapp.com/http://stuff.mit.edu/afs/sipb/contrib/pi/pi-billion.txt');
  return response.body.getReader();
})();
render(html`π is: ${asyncAppend(streamingResponse)}`, document.body);
```

I could add some examples like that.

src/lit-html.ts
```diff
@@ -312,13 +313,14 @@ export const directive = <F extends DirectiveFn>(f: F): F => {
   return f;
 };
 
+const isDirective = (o: any) =>
+    typeof o === 'function' && o.__litDirective === true;
+
+const directiveValue = {};
```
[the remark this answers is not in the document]
justinfagnani: I'm being very careful with Symbol usage to more easily support IE11 without requiring too many polyfills.

src/lit-html.ts
```diff
@@ -299,8 +299,9 @@ export class Template {
 export const getValue = (part: Part, value: any) => {
```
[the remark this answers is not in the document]
justinfagnani: I can add it in this PR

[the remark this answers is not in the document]

src/lib/async-append.ts
```diff
@@ -0,0 +1,67 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {directive, NodePart} from '../lit-html.js';
+
+/**
+ * Render items of an async iterable, appending new values to previous values.
```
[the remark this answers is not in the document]
justinfagnani: done

src/lib/async-append.ts
```diff
@@ -0,0 +1,67 @@
...
+              // Like iterables, each item induces a part, so we need to keep
+              // track of start and end nodes. Because these parts are not
+              // updatable like an iterable, it may be possible to optimize away
+              // the parts and just re-use the setValue() logic.
+              let itemStart = part.startNode;
+              if (itemPart !== undefined) {
+                itemStart = new Text();
+                // Set up the previous item's endNode to the new part's
+                // startNode
+                itemPart.endNode = itemStart;
```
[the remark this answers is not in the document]
justinfagnani: I tried to clarify in comments. Inside this conditional, before we've created a Part for the new item, `itemPart` is referring to the previous item's part. Then after the conditional, we create a new Part.

src/lib/async-replace.ts
```diff
@@ -0,0 +1,60 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {directive, NodePart} from '../lit-html.js';
+
+// const previousIterables = new WeakMap<NodePart, AsyncIterable<any>>();
+
+/**
+ * Render items of an async iterable, replacing previous values.
```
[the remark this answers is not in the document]
justinfagnani: done

src/lit-html.ts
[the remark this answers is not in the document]
justinfagnani: added

src/test/lib/async-replace_test.ts
```diff
@@ -0,0 +1,111 @@
...
+import {asyncReplace} from '../../lib/async-replace.js';
+import {html, render} from '../../lit-html.js';
+
+import {TestAsyncIterable} from './test-async-iterable.js';
+
+const assert = chai.assert;
+
+// Set Symbol.asyncIterator on browsers without it
+if (typeof Symbol !== undefined && Symbol.asyncIterator === undefined) {
+  Object.defineProperty(Symbol, 'asyncIterator', {value: Symbol()});
```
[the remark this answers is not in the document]
justinfagnani: It might, but shouldn't matter for the tests

src/test/lib/async-replace_test.ts
justinfagnani: fixed

src/test/lib/async-replace_test.ts
justinfagnani: both don

src/test/lib/async-replace_test.ts
justinfagnani: fixed

src/test/lib/async-replace_test.ts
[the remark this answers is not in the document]
cdata: Nope, only really matters for correctness (and seems trivial to do).

src/test/lib/test-async-iterable.ts
```diff
@@ -0,0 +1,50 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+// Set Symbol.asyncIterator on browsers without it
+if (typeof Symbol !== undefined && Symbol.asyncIterator === undefined) {
+  Object.defineProperty(Symbol, 'asyncIterator', {value: Symbol()});
+}
+
+export class TestAsyncIterable<T> implements AsyncIterable<T> {
```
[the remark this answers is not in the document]
justinfagnani: done

src/test/lib/test-async-iterable.ts
justinfagnani: done

src/test/lib/test-async-iterable.ts
```diff
@@ -0,0 +1,50 @@
...
+      new Promise((res, _) => this._resolveNextValue = res);
+  private _resolveNextValue: (value: T) => void;
+
+  async * [Symbol.asyncIterator]() {
+    while (true) {
+      yield await this._nextValue;
+    }
+  }
+
+  async push(value: any): Promise<void> {
```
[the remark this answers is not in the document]
justinfagnani: done

src/test/lib/test-async-iterable.ts
```diff
@@ -0,0 +1,50 @@
...
+    const currentValue = this._nextValue;
+    const currentResolveValue = this._resolveNextValue;
+    this._nextValue =
+        new Promise((ressolve, _) => this._resolveNextValue = ressolve);
```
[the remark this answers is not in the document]
justinfagnani: done

src/test/lib/test-async-iterable.ts
justinfagnani: Because I only stumbled on the need for this when devtools was closed, I'd like to keep it in one documented place. I'm not yet sure why this wasn't needed with devtools open.

src/test/lib/until_test.ts
justinfagnani: done

cdata:
> I'm not sure if they do or not... I'm thinking about this less in terms of existing state management and more in terms of supporting the relatively primitive platform types, which async iterables are now, much like Promises.

It matters that these patterns are considered useful for rendering UI, though. If these primitives are not used for rendering UI -
 and certainly some systems would advocate explicitly against using them in that capacity - then: YAGNI.

> There are useful cases this enables that don't use any state management solution or framework.

The user who streams fetch and progressively handles results is probably doing so in the context of an integrated application (rather than a stand-alone component). It seems likely to me that they will be doing this within the prescribed structure of a framework or state management system.

Who is the user that is doing this without a framework and/or state management system? What is that user building?

These patterns are hard to coordinate across an integrated application. If I have five component instances and all of them want to access data on the network, it is inefficient if they all fetch and operate on their own instances of the same JSON.

So, what I want to see is how these patterns can be made manageable at the scale of an integrated application. Someone out there may have demonstrated it, but we can too if necessary (this would take more time). However, my bias is that it is more difficult and unwieldy to coordinate components that don't have some centralized management scheme.

justinfagnani:
> Who is the user that is doing this without a framework and/or state management system? What is that user building?

I'm not sure, and not sure it matters in this context. This is a fairly low-level library that is composing and exposing relatively primitive data types and operations. By supporting these primitives user can combine them to do interesting things on top of them. I don't think this library should concern itself too much with state management.

cdata:
> I don't think this library should concern itself too much with state management.

I'm confused then. These directives are prescribing a strategy for state management.

justinfagnani:
> I'm confused then. These directives are prescribing a strategy for state management.

I don't think so. They're adding a way to deal with a platform type. It's the async analogy to iterables. Are iterables state management?

justinfagnani:
I'm just filling out the lower-right quadrant here:

| | Sync | Async |
|-|-|-|
|**Single**|`I`|`Promise<I>`|
|**Many**|`I[]`|*`AsyncIterable<I>`*|

cdata:
If these directives are offering features that are orthogonal to rendering, then they should probably be in their own project.

If they are directly related to rendering (and I think the change makes it clear they are) then I would like to see my above concerns addressed.

Iterables are used as a corner stone by some trending state management systems (see [redux-saga](https://github.com/redux-saga/redux-saga).

I'm asking you to do this research and share it with the rest of us to make the use case clear. We have been doing a lot of research on the topic of state management, and what you prescribe here can influence our messaging to users.

justinfagnani:
I just don't understand why we're talking about state management here.

cdata:
Well, then I would ask you to re-read my comments.

justinfagnani:
I have. Statement management systems may tie into some of the low-level capabilities in lit, but the capabilities themselves are in now way specific to state management. I don't see why this PR needs to be tied to a discussion of state manegement at all.

cdata:
Please review my earlier comments and address them in good faith. This claim:

> the capabilities themselves are in no way specific to state management

is simply not true. Your own examples demonstrate the relationship between application state and these directives. The example you used above is interesting, but some state management systems would tell you that fetching should not happen there.

justinfagnani:
> Please review my earlier comments and address them in good faith.

*I have*

> This claim:

>> the capabilities themselves are in no way specific to state management

> is simply not true. Your own examples demonstrate the relationship between application state and these directives. The example you used above is interesting, but some state management systems would tell you that fetching should not happen there.

Should I care what some state management systems have to say about that? Where is this requirement coming from?

This is starting to be a major distraction from the core features being added in this PR. I would politely ask that we move on.

sorvell:
I think I see both your points here @cdata and @justinfagnani. My concern is not explicitly about the value of the pattern but more that the number of directives here is creeping up. I don't feel super strongly about it yet, but it may make sense to separate some or all of the directives from the main project in the interest of reducing maintenance burden, cognitive load, and (specifically relevant here) bar for inclusion.

justinfagnani:
@sorvell So far developing these directives in-repo has really helped test out the directive API. It was this PR that showed that directives need to be handled slightly differently in core to not use their return value. Keeping all tests and infrastructure in one place and everything tested on atomic commits is very useful.

There are still other directives I consider fairly core like `guard` and `when` that I'd like to get to. Maybe soon we should bring back labs/ and figure out what can go in there.

justinfagnani:
Added two examples to the README. I'm working on more extensive non-README docs that can host more examples.

README.md
```diff
@@ -240,6 +240,48 @@ const render = () => html`
...
+
+In the near future, `ReadableStream`s will be async iterables, enabling streaming `fetch()`
+directly into a template:
+
+```javascript
+const streamingResponse = (async () => {
+  const response = await fetch('https://cors-anywhere.herokuapp.com/http://stuff.mit.edu/afs/sipb/contrib/pi/pi-billion.txt');
+  return response.body.getReader();
+})();
+render(html`π is: ${asyncAppend(streamingResponse)}`, document.body);
```
danizord: `streamingResponse` is a promise here, no?

justinfagnani:
Rebased on master and formatted

README.md
```diff
@@ -240,6 +240,48 @@ const render = () => html`
 `;
 ```
 
+#### `asyncAppend(asyncIterable)` and `asyncReplace(asyncIterable)`
+
```
richardanaya: I think maybe a bit of text at the beginning of this readme section could help the readers mind flow into what is about to be discussed:

"Asynchronous iterators offer the ability to continuously offer new data to be rendered into an element. lit-html offers two ways to consume that data into your element:"

README.md
```diff
@@ -240,6 +240,48 @@ const render = () => html`
...
+Example:
+
+```javascript
+/**
+ * Returns an async iterable that yields the number of seconds since `date`,
+ * approximately every second.
+ */
+async function* secondsAgo(date) {
+  const seconds = (Date.now() - date.getTime()) / 1000;
```
richardanaya: Perhaps there is a simpler form example? I find datetime in this example  somewhat distracting. Perhaps setTimeout(, 1000) is sufficient?

richardanaya:
@cdata @justinfagnani I just wanted to add that I am EXCITED to finally see new primitives like asynchronous part replacement (by promises, async iterables, or observers) for DOM rendering. I I think the simple state binding, immutable unidirectional dataflows are useful for heirarchal component construction ( traditional web apps).  But as a component builder ( who often works in trying to build my web components in black box environment without assumption of the paradigms above me ), I'm happy to finally see some tools to help me be more concise.

@cdata to attempt to address your concern. One observation I have about this PR is how little the core lit-html library is affected! =) you don't have to use asyncReplace or asyncAppend (which to you might be wasted bytes for your paradigm of choice).  I think the work here being done lives up to a core ideal i've seen in this team for extensibility that I think is just wonderful, and fostering of a future of experimentation and personalization in DOM rendering that I think will be useful.

README.md
jridgewell: Yah, that's what I'm reading it as.

src/lib/async-append.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export const asyncAppend = <T>(
+    value: AsyncIterable<T>, mapper?: (v: T, index?: number) => any) =>
+    directive(async (part: NodePart) => {
+      // If we've already set up this particular iterable, we don't need
+      // to do anything.
+      if (value === part._previousValue) {
+        return;
+      }
+      part.clear();
```
jridgewell: Can we wait to clear? That way we could use something like `until` to get a default value while the first promise is resolving.

src/lib/async-replace.ts
```diff
@@ -0,0 +1,73 @@
...
+          if (value === part._previousValue) {
+            return;
+          }
+
+          // We nest a new part to keep track of previous item values separately
+          // of the iterable as a value itself.
+          const itemPart =
+              new NodePart(part.instance, part.startNode, part.endNode);
+
+          part.clear();
```
jridgewell: Ditto.

src/lit-html.ts
```diff
@@ -438,6 +449,7 @@ export class NodePart implements SinglePart {
 
   private _setText(value: string): void {
     const node = this.startNode.nextSibling!;
+    value = value === undefined ? '' : value;
```
jridgewell: Make a test please.

richardanaya:
I know we want to get this one done right. There's also some pretty important bugfixes in current master. I offer one suggestion of a point release for 0.7.x

README.md
danizord: So I think the code should be:
```js
(async () => render(html`π is: ${asyncAppend(await streamingResponse)}`, document.body))();
```

README.md
justinfagnani: added something similar

README.md
justinfagnani: Made a much simpler `countUp()` example :)

src/lib/async-append.ts
justinfagnani: Good point, and I'll do that, but...

I think a lot of the directive, especially when dealing with async values, don't compose well right now. `until(asyncIterable, fallback)` won't quite work. I could spot-fix a few things, but I want to think on this a little and see if there's a more general solution.

src/lib/async-replace.ts
justinfagnani: done

PR #123, opened 2017-10-05
Fix bug with templates with comments

### LINKED ISSUE #122: Templates with comments fail
We didn't have any tests with bindings after comments. Since comment nodes are iterated over in template setup, they need to be iterated over in part instantiation. Currently they end up misaligned.

justinfagnani:
Fixes #122


justinfagnani:
Merging TBR

PR #129, opened 2017-10-07
fix(NodePart): reuse same Node elements

### LINKED ISSUE #128: Rerendering a node part causes the node to be removed, then re-added
If a node or array of nodes is passed to a template, the nodes are rendered as expected — but subsequent renders cause the nodes to be removed and then re-added each time, which is inefficient.

See http://jsbin.com/fomepuj/edit?html,output. Clicking the Update button shows the DOM updating (the nodes briefly flash color in Chrome DevTools), even though the same nodes are being passed in each time. Since no data is actually changing, I'd expect subsequent renders to not update the DOM.

In `NodePart.setValue`, it looks like the `value instanceof Node` case could be optimized by skipping the update if `value === this._previousValue`.

Is the current behavior intentional? If not, I'm happy to submit a fix.

gdi2290:
closes #128

current workaround
http://jsbin.com/nahoyatoxe/edit?html,output

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -410,6 +410,9 @@ export class NodePart implements SinglePart {
   }
 
   private _setNode(value: Node): void {
+    if (this._previousValue === value) {
+      return;
+    }
     this.clear();
     this._insert(value);
     this._previousValue = value;


justinfagnani:
@gdi2290 thanks! Can you add a test? You should be able to use a MutationObserver to detect if the node is removed and added.

PR #130, opened 2017-10-09
fix(ie11): empty tags get no text node so join strings with ' '

daKmoR:
Example Template: `<h2></h2><p>end</p>`
On IE11 empty tags like `<h2></h2>` will not have an empty text node
inside. So the Treewalker will go h2 -> p -> exit. All other
browsers assume that inside there is an empty text. So they go
h2 -> emptyTextNode (that will be filled with data) -> p ...
To force empty text node for IE11 we join with ' '

Replace for of loops of arrays with forEach. They apparently only
get transpiled to be usable for objects.

Replace new Text() with document.createTextNode('')

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -274,9 +274,9 @@ export class Template {
     }
 
     // Remove text binding nodes after the walk to not disturb the TreeWalker
-    for (const n of nodesToRemove) {
+    nodesToRemove.forEach((n) => {
       n.parentNode!.removeChild(n);
-    }
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


daKmoR:
This will break a lot of test as the static data would need to be changed...
It could also be done only for IE11? (howto properly detect?)

Tested in Chrome 61, Firefox 56, Edge 15 and so far I didn't saw any side effects. But this should probably  be more thoroughly tested...

If I should adjust the tests or you have other ideas to make this just let me know.

So far I'm pretty happy that this seems to work quite nicely :) even ShadyCSS plays along :)

daKmoR:
ping - should I do the needed changes for that or do you propose a different approach?

justinfagnani:
@daKmoR I'll take a look at the asap. I merged the other IE PR, so can you rebase? Thanks!

justinfagnani:
Also, this needs a test that fails in IE before the change. I'll IE testing setup soon.

daKmoR:
@justinfagnani
I will rebase it tomorrow but the only change would be from `const html = a.join('');` to `const html = a.join(' ');`
I'm still not sure if it's the best approach - it will ensure an empty textnode everywhere but there will be a space everywhere then in all the templates even for e.g. chrome where it would not be needed...
don't know if it may have any performance impacts...
but I guess it's the only way to get it working on IE11...

daKmoR:
something with the rebase went wrong but so I did a fore push with a cherry pick on the latest master... 

so that should be fine now.

I currently can't run the tests as running them in the browser doesn't seem to work? and I in my current dev setup I can only run it on IE11 via browser...

daKmoR:
ditch in favor of https://github.com/PolymerLabs/lit-html/pull/143

Issue #111, opened 2017-09-25
Tests don't run in Firefox, IE, Edge


daKmoR:
I get for example this result in latest Firefox
```
0 passing tests 
test suite passed
Evaluated 0 tests in 16ms.
```

PS: using polyserve and going to http://localhost:8081/test/

JanMiksovsky:
@daKmoR I see all unit tests pass in Firefox. What problem do you see?

daKmoR:
they do run now :)

PR #132, opened 2017-10-12
Edge unit tests pass

JanMiksovsky:
This is a partial fix for #111 for Microsoft Edge. It fixes the 48 unit tests which were broken on Edge so that all 96 tests now pass.

_This still relies on native ES modules_, which must be enabled on Edge through about:flags, "Enable experimental JavaScript features". This PR just fixes the unit tests. To get everything to pass in production Edge (and, eventually, IE), separate work will be required to create a bundled build.

One thing to review: the loop that processed attributes needed to be rewritten to deal with the fact that Edge and IE don't necessarily return attributes in document order. The new loop extracts the expected string name for the attribute from the corresponding member of `strings`, then uses that to look up the actual `Attr` instance on the element. Because of that change in sequence, the relevant regex `match` call had to be updated: rather than matching against the entire string, the `$` (match end of string) directive was dropped. This still appears to find the attribute name correctly, at least as far as evidenced in the unit tests.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -248,7 +248,7 @@ export class Template {
         // placholder, or empty text, add a marker node.
         if (node.previousSibling === null ||
             node.previousSibling !== previousNode) {
-          parent.insertBefore(new Text(), node);
+          parent.insertBefore(document.createTextNode(''), node);
         } else {
           index--;
         }
@@ -258,7 +258,7 @@ export class Template {
         // We don't have to check if the next node is going to be removed,
         // because that node will induce a marker if so.
         if (node.nextSibling === null) {
-          parent.insertBefore(new Text(), node);
+          parent.insertBefore(document.createTextNode(''), node);
         } else {
           index--;
         }
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -188,7 +188,7 @@ export class Template {
         // attributes are not guaranteed to be returned in document order. In
         // particular, Edge/IE can return them out of order, so we cannot assume
         // a correspondance between part index and attribute index.
-        
+
         // Do a first pass to count attributes that correspond to parts.
         const attributesWithParts: string[][] = [].filter.call(attributes, (attribute: Attr) =>
           attribute.value.split(attrOrTextRegex).length > 1
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -184,24 +184,30 @@ export class Template {
           continue;
         }
         const attributes = node.attributes;
-        for (let i = 0; i < attributes.length; i++) {
-          const attribute = attributes.item(i);
-          const attributeStrings = attribute.value.split(attrOrTextRegex);
-          if (attributeStrings.length > 1) {
-            // Get the template literal section leading up to the first
-            // expression in this attribute attribute
-            const attributeString = strings[partIndex];
-            // Trim the trailing literal value if this is an interpolation
-            const rawNameString = attributeString.substring(
-                0, attributeString.length - attributeStrings[0].length);
-            // Find the attribute name
-            const rawName = rawNameString.match(/((?:\w|[.\-_$])+)=["']?$/)![1];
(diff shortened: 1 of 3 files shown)


daKmoR:
@JanMiksovsky did you also so the change needed for IE11?
https://github.com/PolymerLabs/lit-html/pull/130

IE11 works a little different when it comes to "empty" text nodes apparently

is this solved by this as well or could you incorporate it?

justinfagnani:
This is great @JanMiksovsky!

I'm a little worried about the change in the attribute name regex, but I'll try adding some more test cases to see if it's a problem.

src/lit-html.ts
```diff
@@ -184,24 +184,30 @@ export class Template {
...
-            partIndex += attributeStrings.length - 1;
-            i--;
-          }
+        // Per https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap,
+        // attributes are not guaranteed to be returned in document order. In
+        // particular, Edge/IE can return them out of order, so we cannot assume
+        // a correspondance between part index and attribute index.
+
+        // Do a first pass to count attributes that correspond to parts.
+        const attributesWithParts: string[][] = [].filter.call(attributes, (attribute: Attr) =>
```
azakus: This allocates an array every time. Maybe use `Array.prototype.filter.call`?

src/lit-html.ts
```diff
@@ -184,24 +184,30 @@ export class Template {
...
+          attribute.value.split(attrOrTextRegex).length > 1
+        );
+        // Loop that many times, but don't use loop index for anything.
+        for (let i = 0; i < attributesWithParts.length; i++) {
+          // Get the template literal section leading up to the first
+          // expression in this attribute attribute
+          const stringForPart = strings[partIndex];
+          // Find the attribute name
+          // Trim the trailing literal value if this is an interpolation
+          const attributeNameInPart = stringForPart.match(/((?:\w|[.\-_$])+)=["']?/)![1];
```
azakus: Pull regex out into a constant.

JanMiksovsky:
Great, glad to help. We're rewriting the Elix components in an FRP style, and are interested in using lit-html as the rendering portion. Both Edge/IE are mission critical for our clients.

If it'd be of use, we could also take a shot at modifying the build and test page to support testing on platforms that don't support modules. It looks like that'd entail changing the TypeScript build target to ES2015. That'd obviously effect the dev/test cycle, though, as debugging would take place in more heavily transpiled code. Thoughts on that?

JanMiksovsky:
@daKmoR Thanks for the heads up on empty text nodes in IE. Once there's an IE-compatible build in place, I can look at checking the issue you mention.

justinfagnani:
@JanMiksovsky WCT is capable of compiling on the fly no, so we shouldn't need to alter the tsconfig

justinfagnani:
@JanMiksovsky this is failing for attribute bindings that come after attribute literals:

```ts
      test('renders to an attribute expression after an attribute literal', () => {
        render(html`<div a="b" foo="${'bar'}"></div>`, container);
        assert.equal(container.innerHTML, '<div a="b" foo="bar"></div>');
      });
```

daKmoR:
It should also be possible to just use latest polyserv as it will transpile to requirejs and AMD modules that can be used on IE11

justinfagnani:
This regex works, but disallows quotes in attribute values, even when they would be allowed by HTML: `((?:\w|[.\-_$])+)=["']?[^"']*$`

justinfagnani:
This regex might be better, it allows quotes that aren't the attribute value delimiter: `((?:\w|[.\-_$])+)=(?:[^"']*|(?:["][^"]*)|(?:['][^']*))$`

JanMiksovsky:
Thanks for merging.

@azakus @justinfagnani I didn't have a chance to adopt your suggestions before this got merged. Did you want to make those changes yourselves, or should I make a follow-up PR?

PR #134, opened 2017-10-13
Fix binding to attributes after non-bound attributes

justinfagnani:


justinfagnani:
@sorvell the issue this fixes is breaking some users, could you take a look?

justinfagnani:
The failures on Edge are due to it reordering attributes and the expectation being too strict. Fixing...

richardanaya:
Thank you!

justinfagnani:
Ok, all tests green on all browsers now.

PR #137, opened 2017-10-17
fix: add polyfills so test can run on IE11

daKmoR:
With this tests at least run on IE11 :p
```
Test Results
54 failing tests
20 passing tests
Evaluated 73 tests in 520ms.
```

once this is done I guess there is a need for some follow up PRs for IE11


justinfagnani:
With the latest changes to get tests passing in Edge, I see 39 pass / 57 fail 👍

PR #139, opened 2017-10-19
Add a failing nested templates test

HitkoDev:
Test rendering nested templates when parent template contains preceding nodes, such as:
```js
html`<a>${foo}</a>${partial}`
```

Currently, lit-html throws `Cannot read property 'parentNode' of null` when trying to render such templates.


justinfagnani:
Thanks for the test @HitkoDev! I was able to reduce the test to any expression following an element.

I filed #148 as an issue for this, and have a fix in #149

PR #143, opened 2017-10-20
fix(IE11): manually create empty text nodes => 60 passing tests

daKmoR:
Finally found a better approach to bring IE11 on board. Still not perfect but it's 60 passing tests and the basic functionality is working :)

```
// from
Test Results
57 failing tests
39 passing tests
// to
Test Results
36 failing tests
60 passing tests
```

especially with Sub Templates there is still something going on...

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "wct-browser-legacy": "^0.0.1-pre.10",
     "web-component-tester": "^6.3.0",
     "@webcomponents/template": "^1.1.0",
-    "get-own-property-symbols": "^0.9.2"
+    "babel-polyfill": "^6.26.0"
   },
   "typings": "lit-html.d.ts",
   "dependencies": {}
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -281,9 +281,9 @@ export class Template {
     }
 
     // Remove text binding nodes after the walk to not disturb the TreeWalker
-    for (const n of nodesToRemove) {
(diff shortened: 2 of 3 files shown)


daKmoR:
hmmm a Firefox Promise Test fails because `babel-polyfill` apparently messes with Firefox'es promise...

should we write that Promise Test differently? or search for another Promise Polyfill?

justinfagnani:
I'd like to see if we can't just make the template preparation code smarter. Querying and adding text nodes in cloning is going to slow things down a lot I suspect.

justinfagnani:
It'd be nice if there were a way to simulate the IE behavior of parsing in Chrome. We can walk a template a remove empty text nodes, it's just having to right hook in place to do it.

daKmoR:
yeah doing this in cloning may be not the best place... (maybe it can be done while creating the template - didn't test it as the content is just `<h2></h2>` and IE sees no empty text node and Chrome sees one I'm not sure - maybe it somehow remembers it inside the template :p

I also thought about getting rid of all those empty text nodes... while looking around I stumbled over https://www.sitepoint.com/removing-useless-nodes-from-the-dom/
not sure if it can just be applied while creating the template - even if same concern as above...

probably need to test that to be sure

daKmoR:
Close in favor of https://github.com/PolymerLabs/lit-html/pull/202

justinfagnani:
@daKmoR thanks for all your work here!

PR #153, opened 2017-10-25
Optimizes Template#_getHtml

jridgewell:
Two main optimizations:

1. The regexs are _really_ bad at predicting where the end of a string
is. This leads to bad performance when using a `*` quantifier to match
everything to the end. It's much faster to use `#indexOf` and
`#lastIndexOf` to find the `<` and `>`, and do some comparisons.
   - https://jsperf.com/html-template-tag-context

2. String concatenation is much faster than building an array to join
it.
   - https://jsperf.com/string-concatenation/47
     - (This even compares an already made array, it's super slow without
   even building the array in a loop!)

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -296,7 +296,7 @@ export class Template {
   private _getHtml(strings: TemplateStringsArray, svg?: boolean): string {
     const l = strings.length;
     let html = '';
-    let isTextBinding = false;
+    let isTextBinding = true;
     for (let i = 0; i < l - 1; i++) {
       const s = strings[i];
       html += s;
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -111,20 +111,24 @@ export function render(
  */
 const attributeMarker = `{{lit-${Math.random()}}}`;
 
-/**
- * Regex to scan the string preceding an expression to see if we're in a text
- * context, and not an attribute context.
- *
- * This works by seeing if we have a `>` not followed by a `<`. If there is a
- * `<` closer to the end of the strings, then we're inside a tag.
- */
-const textRegex = />[^<]*$/;
-const hasTagsRegex = /[^<]*/;
 const textMarkerContent = '_-lit-html-_';
 const textMarker = `<!--${textMarkerContent}-->`;
 const attrOrTextRegex = new RegExp(`${attributeMarker}|${textMarker}`);
 const lastAttributeNameRegex =
     /((?:\w|[.\-_$])+)=(?:[^"']*|(?:["][^"]*)|(?:['][^']*))$/;
+
+/**
+ * Finds the closing index of the last closed HTML tag.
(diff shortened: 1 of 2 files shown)


src/lit-html.ts
```diff
@@ -111,20 +111,24 @@ export function render(
...
 
-/**
- * Regex to scan the string preceding an expression to see if we're in a text
- * context, and not an attribute context.
- *
- * This works by seeing if we have a `>` not followed by a `<`. If there is a
- * `<` closer to the end of the strings, then we're inside a tag.
- */
-const textRegex = />[^<]*$/;
-const hasTagsRegex = /[^<]*/;
```
jridgewell: BTW, this was broken beforehand. Any string passes, including `"<test"`, `"test<"`, and `""`. I'll add a test later tonight.

justinfagnani:
Wow, these look like great improvements @jridgewell, thanks for the PR!

I have some outstanding PRs that touch this code, so I'll merge their tests with this change and see where we're at. Looking forward for the tests on the tag detection too!

daKmoR:
hmmm wanted to test this out...
but it breaks my template here
`<button on-click=${(e) => { this.run(); }}>`
current output `<button>`
in this branch `(e) => { this.run(); }`

so code works fine in the current master but if I take this branch it does not... strange that no test catches this?

justinfagnani:
I have a test that's broken at master and in the `nested` branch, but fixed by this change, exactly as pointed out by @jridgewell in the comment about the regexes:

```ts
      test('renders attributes bindings after text bindings', () => {
        render(html`
          <div>${''}</div>
          <div foo=${'bar'}></div>
        `, container);
        assert.equal(container.innerHTML, '<div></div><div foo="bar"></div>');
      });
```

justinfagnani:
@daKmoR this test, in `lit-extended_test.ts` is passing for me with this change:

```ts
    test('renders a function to a property', () => {
      render(html`<button on-click=${(e: Event) => {console.log(e)}}></button>`, container);
      assert.equal(container.innerHTML, '<button></button>');
    });
```

Can you make a test that fails we can look at?

daKmoR:
hey, so apparently it only was a problem in combination
```
test.only('renders combination of static attributes and click handlers', () => {
	render(html`<div class="test"><button on-click="${(e) => {console.log(e)}}"></button></div>`, container);
	assert.equal(container.innerHTML, '<div class="test"><button></button></div>');
});
```

that test fails in the `optimize-getHtml` branch.

Once you rebase to latest master it's already fixed :)

jridgewell:
Tests added.

justinfagnani:
Just did some testing. This is pretty much perf-neutral on the Marko Search Results benchmark, but seem like a win on another boot-speed benchmark I have, which makes sense as this savings is isolated to template creation time. It's also < 20 bytes larger, and fixes a critical but.

Thanks a lot @jridgewell!

PR #160, opened 2017-11-02
Normalize template string part iteration

jridgewell:
The template string iteration that's used in `Template#_getHtml` is pretty nifty. This reuses it in `AttributePart#setValue`, since it's very much the same.


justinfagnani:
@jridgewell looks great, but can you rebase?

jridgewell:
Rebased.

PR #161, opened 2017-11-02
Optimize repeat

### LINKED ISSUE #95: repeat() slows down adding rows extrimely
I have created `lit-html` version for the [js-framework-benchmark](https://github.com/krausest/js-framework-benchmark) (you can find it [here](https://github.com/Lodin/js-framework-benchmark/tree/lit-html/lit-html-v0.6.0-keyed)). 

There is a code I wrote: 
```javascript
render(html`${this.store.data.map(i => html`
  <tr class="${i.id === this.store.selected ? 'danger' : ''}">
    <td class="col-md-1">${i.id}</td>
    <td class="col-md-4">
      <a on-click="${() => this.select(i.id)}">${i.label}</a>
    </td>
    <td class="col-md-1">
      <a on-click="${() => this.delete(i.id)}">
        <span class="glyphicon glyphicon-remove" aria-hidden="true"></span>
      </a>
    </td>
    <td class="col-md-6"></td>
  </tr>`
)}`, this.tbody);
```
It appends a single row to the table. Then, there is results of benchmarking:
* Creating 10,000 rows took `2485.26`
* Adding 1,000 rows to existing table took `604.47`

But if I replace `this.store.data.map` with `repeat`: 
```javascript
render(html`${repeat(this.store.data, i => i.id, i => html`
  ... same code
)}`, this.tbody);
```
The benchmark results become following: 
* Creating 10,000 rows took `2448.795`
* Adding 1,000 rows to existing table took `7411.785`

These results keeps repeating in the same numbers (adding 1,000 rows time can vary from `5000` to `15000`), so I assume they are true. You can also clone my repo and check the results.

### LINKED ISSUE #157: repeat: keyFn makes perf worse?
(Im not sure if this is a dublicate of https://github.com/PolymerLabs/lit-html/issues/95)

I repeat over a list of about 500 products and change the ordering afterwards (to test perf). The product-list have a unique field called id that is a Number. 

Example using repeat 
https://github.com/filipbech/lit-element-example/blob/using-repeat/demo.js#L46-L48 

when just using ```products.map(product=>html`...`)``` it ends up being alot faster when re-ordering... if I take out the keyFn it runs about as fast as the .map-version (which is what i expected).

It also seems to take all my elements out and put them in again, so all connectedCallbacks are fired...

jridgewell:
This pretty seriously optimizes the repeat directive.

Now, it creates an extra text node for every item (it creates an explicit start node for every item, instead of reusing the previous' end node). That's not _perfect_, but it simplifies a lot of code and makes it run _much_ faster. I think it's a pretty good tradeoff.

Fixes #95.
Fixes #157.

### DIFF
--- a/src/lib/repeat.ts
+++ b/src/lib/repeat.ts
@@ -17,12 +17,7 @@ import {directive, DirectiveFn, NodePart} from '../lit-html.js';
 export type KeyFn<T> = (item: T) => any;
 export type ItemTemplate<T> = (item: T, index: number) => any;
 
-interface State {
-  keyMap: Map<any, NodePart>;
-  parts: NodePart[];
-}
-
-const stateCache = new WeakMap<NodePart, State>();
+const keyMapCache = new WeakMap<NodePart, Map<any, NodePart>>();
 
 function cleanMap(part: NodePart, key: any, map: Map<any, NodePart>) {
   if (!part.startNode.parentNode) {
@@ -46,19 +41,13 @@ export function repeat<T>(
 
   return directive((part: NodePart): any => {
 
-    let state = stateCache.get(part);
-    if (state === undefined) {
-      state = {
-        keyMap: new Map(),
-        parts: [],
-      };
-      stateCache.set(part, state);
+    let keyMap = keyMapCache.get(part);
+    if (keyMap === undefined) {
+      keyMap = new Map(),
+      keyMapCache.set(part, keyMap);
     }
     const container = part.startNode.parentNode as HTMLElement | ShadowRoot |
         DocumentFragment;
-    const keyMap = state.keyMap;
-
-    const itemParts = [];
     let index = -1;
     let currentMarker = part.startNode.nextSibling;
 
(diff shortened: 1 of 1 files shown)


src/lib/repeat.ts
```diff
@@ -43,112 +49,66 @@ export function repeat<T>(
...
-      if (itemParts.length === 0) {
-        clearRange.setStartBefore(clearStart);
-      } else {
-        clearRange.setStartAfter(clearStart);
-      }
-      clearRange.setEndAfter(clearEnd);
-      clearRange.deleteContents();
-      clearRange.detach();  // is this neccessary?
+    if (currentMarker !== part.endNode) {
+      const range = document.createRange();
```
justinfagnani: We realized in lit core that using a Range to clear nodes is really slow. Want to change this to just iterate and remove?

jridgewell: https://jsperf.com/range-deletecontents-vs-loop

It seems creating the range is stupid slow, but either way using a loop is faster.

jridgewell: Seems extraction is slower, too: https://jsperf.com/range-extractcontents-vs-loop

PR #163, opened 2017-11-03
Add removeNodes helper

jridgewell:
We have this reimplemented three times, twice here and once (soon) in the repeat directive.

Should we create a utils module? This probably shouldn't be exported with the main binary, but it'll need to be imported in multiple files.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -624,16 +624,10 @@ export const removeNodes =
     (startNode: Node | null,
      endNode: Node | null,
      container: Node): void => {
-      if (startNode === null) {
-        return;
-      }
       let node = startNode;
       while (node !== endNode) {
-        const n = node.nextSibling;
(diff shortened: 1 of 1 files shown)


src/lit-html.ts
```diff
@@ -624,3 +618,19 @@ export class TemplateInstance {
     return fragment;
   }
 }
+
+/**
+ * Removes nodes, starting from `startNode` (inclusive) to `endNode`
+ * (exclusive), from `container`.
+ */
+export const removeNodes =
```
justinfagnani: Similar to my comment in the SVG PR: https://github.com/PolymerLabs/lit-html/pull/164#discussion_r148941728

I'd change the signature to:

```ts
const removeNodes = (container: Node, start: Node, end: Node | null = null): void => {
```

jridgewell: Done.

PR #164, opened 2017-11-03
Reparent SVGs after initial creation

jridgewell:
No need to do it every time we clone, we can just reparent them after the DOM string is initially parsed.

I created a reusable helper like #163, since this will be useful for the repeat directive, too.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -629,16 +629,10 @@ export const reparentNodes =
      endNode: Node | null,
      container: Node,
      beforeNode: Node | null): void => {
-      if (startNode === null) {
-        return;
-      }
       let node = startNode;
       while (node !== endNode) {
-        const n = node.nextSibling;
-        container.insertBefore(node, beforeNode);
-        if (n === null) {
-          break;
-        }
+        const n = node!.nextSibling;
+        container.insertBefore(node!, beforeNode);
         node = n;
       }
     };
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -155,12 +155,16 @@ export class TemplatePart {
 export class Template {
   parts: TemplatePart[] = [];
   element: HTMLTemplateElement;
-  svg: boolean;
 
   constructor(strings: TemplateStringsArray, svg: boolean = false) {
-    this.svg = svg;
-    this.element = document.createElement('template');
-    this.element.innerHTML = this._getHtml(strings, svg);
+    const element = this.element = document.createElement('template');
+    element.innerHTML = this._getHtml(strings, svg);
+    if (svg) {
+      const svgElement = element.content.firstChild!;
+      reparentNodes(svgElement.firstChild, null, element.content, null);
+      element.content.removeChild(svgElement);
+    }
+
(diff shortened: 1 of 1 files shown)


src/lit-html.ts
```diff
@@ -613,14 +617,25 @@ export class TemplateInstance {
...
   }
 }
+
+/**
+ * Reparents nodes, starting from `startNode` (inclusive) to `endNode`
+ * (exclusive), into another container (could be the same container), before
+ * `beforeNode`. If `beforeNode` is null, it appends the nodes to the
+ * container.
+ */
+export const reparentNodes =
```
justinfagnani: Let's order the arguments to put optional last and use `?` to mark the optional args.

And 🚲🏠: since they're all `Node`s, we can shorten the names:

```ts
const reparentNodes = (container: Node, start: Node, end: Node|null = null, before?: Node) {
```

I added the default value for `end` so that if it's omitted, it's defaulted to `null` to work with `node !== endNode`. Accidentally passing `undefined` would have caused an infinite loop. `beforeNode` is safe as `undefined` though.

src/lit-html.ts
```diff
@@ -157,12 +157,16 @@ export class TemplatePart {
...
   constructor(strings: TemplateStringsArray, svg: boolean = false) {
-    this.svg = svg;
-    this.element = document.createElement('template');
-    this.element.innerHTML = this._getHtml(strings, svg);
+    const element = this.element = document.createElement('template');
+    element.innerHTML = this._getHtml(strings, svg);
+    if (svg) {
+      const svgElement = element.content.firstChild!;
+      reparentNodes(svgElement.firstChild, null, element.content, null);
+      element.content.removeChild(svgElement);
```
justinfagnani: Would there be a benefit to removing the svg element first?

src/lit-html.ts
```diff
@@ -613,14 +617,25 @@ export class TemplateInstance {
...
+ */
+export const reparentNodes =
+    (startNode: Node | null,
+     endNode: Node | null,
+     container: Node,
+     beforeNode: Node | null): void => {
+      let node = startNode;
+      while (node !== endNode) {
+        const n = node!.nextSibling;
+        container.insertBefore(node!, beforeNode);
```
justinfagnani: If we don't yet have a case of using `beforeNode`, I think that `appendChild` is 10-20% faster than `insertBefore` in Chrome due to a couple of additional precondition checks.

src/lit-html.ts
jridgewell: `start` may be `null`, [though](https://github.com/PolymerLabs/lit-html/pull/164/files#diff-7548be97e171131880082c935a962f68R166).

> beforeNode is safe as undefined though.

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) has a specific warning about this.

src/lit-html.ts
jridgewell: It'll be necessary for the `repeat` directive, so I can move the current part before `currentMarker`.

PR #165, opened 2017-11-03
Simplify EventPart

jridgewell:
No need to register the part as a listener, just register the listener.

And fixes bugs with updating the event listener (instead of having to render it null first).

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -107,22 +107,17 @@ export class EventPart implements Part {
 
   setValue(value: any): void {
     const listener = getValue(this, value);
-    if (listener === this._listener) {
+    const previous = this._listener;
+    if (listener === previous) {
       return;
     }
-    if (listener == null) {
-      this.element.removeEventListener(this.eventName, this);
-    } else if (this._listener == null) {
-      this.element.addEventListener(this.eventName, this);
-    }
-    this._listener = listener;
-  }
 
-  handleEvent(event: Event) {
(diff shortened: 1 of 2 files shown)


kaste:
This looks like a perf regression. Consider that we're usually in a render loop. With this patch, inline event handlers will cause a remove / add eventListener dance, don't they?

justinfagnani:
@kaste There is a small change here - previously the EventPart itself was passed to addEventListener, so it was stable, but it's `_listener` property was set with the new user-provided listener on every update. With this PR the intermediate event handler is gone, so instead of property sets each update, add/removeEventListener is indeed called each time.

Admittedly, we didn't measure the perf impact here. My assumption is that inline handlers are pretty bad in this case because you're already creating a new closure each update, and the correct way to do this, which is to create the handler outside of the template, will eliminate thrashing in the code before and after this PR.

We could add back the stable handler though, if this is a problem. This PR was just a nice simplification... did you measure an impact here?

kaste:
I could measure an impact albeit only in benchmarks of course. 

But actually the observation was the other way around. I first observed that you don't do the add/remove dance all the time, and thought That's nice, I just keep my inline handlers here. 😄 

Creating closures itself is fast enough, maybe too much pressure on GC.

justinfagnani:
Yeah, it's possible that closure creation is enough faster than add/RemoveEvent listener that this is a meaningful regression for inline handlers. I'm definitely open to reviving the old approach.

PR #168, opened 2017-11-04
Various improvements

jridgewell:
I'll do an inline review of the changes.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -109,13 +109,11 @@ export function render(
  * An expression marker with embedded unique key to avoid
  * https://github.com/PolymerLabs/lit-html/issues/62
  */
-const attributeMarker = `{{lit-${Math.random()}}}`;
-
-const textMarkerContent = '_-lit-html-_';
-const textMarker = `<!--${textMarkerContent}-->`;
-const attrOrTextRegex = new RegExp(`${attributeMarker}|${textMarker}`);
-const lastAttributeNameRegex =
-    /((?:\w|[.\-_$])+)=(?:[^"']*|(?:["][^"]*)|(?:['][^']*))$/;
+const marker = `{{lit-${String(Math.random()).slice(2)}}}`;
+const nodeMarker = `<!--${marker}-->`;
+const markerRegex = new RegExp(`${marker}|${nodeMarker}`);
+const nonWhitespace = /[^\s]/;
+const lastAttributeNameRegex = /([\w\-.$]+)=(?:[^"']*|"[^"]*|'[^']*)$/;
 
 /**
  * Finds the closing index of the last closed HTML tag.
@@ -194,24 +192,23 @@ export class Template {
         // attributes are not guaranteed to be returned in document order. In
         // particular, Edge/IE can return them out of order, so we cannot assume
         // a correspondance between part index and attribute index.
-
-        // Do a first pass to count attributes that correspond to parts.
-        const attributesWithParts: string[][] = Array.prototype.filter.call(
-            attributes,
-            (attribute: Attr) =>
-                attribute.value.split(attrOrTextRegex).length > 1);
-        // Loop that many times, but don't use loop index for anything.
-        for (let i = 0; i < attributesWithParts.length; i++) {
+        let count = 0;
+        for (let i = 0; i < attributes.length; i++) {
+          if (attributes[i].value.indexOf(marker) >= 0) {
+            count++;
+          }
+        }
+        while (count--) {
           // Get the template literal section leading up to the first
           // expression in this attribute attribute
           const stringForPart = strings[partIndex];
           // Find the attribute name
           const attributeNameInPart =
-              stringForPart.match(lastAttributeNameRegex)![1];
+              lastAttributeNameRegex.exec(stringForPart)![1];
           // Find the corresponding attribute
           const attribute = attributes.getNamedItem(attributeNameInPart);
           const stringsForAttributeValue =
-              attribute.value.split(attrOrTextRegex);
+              attribute.value.split(markerRegex);
           this.parts.push(new TemplatePart(
               'attribute',
               index,
@@ -223,9 +220,9 @@ export class Template {
         }
       } else if (node.nodeType === 3 /* Node.TEXT_NODE */) {
         const nodeValue = node.nodeValue!;
-        const strings = nodeValue.split(attrOrTextRegex);
-        if (strings.length > 1) {
+        if (nodeValue.indexOf(marker) > -1) {
           const parent = node.parentNode!;
+          const strings = nodeValue.split(markerRegex);
           const lastIndex = strings.length - 1;
 
           // We have a part for each match found
@@ -251,7 +248,7 @@ export class Template {
                previousSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
               (nextSibling === null ||
                nextSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
-              nodeValue.trim() === '') {
+              !nonWhitespace.test(nodeValue)) {
             nodesToRemove.push(node);
             currentNode = previousNode;
             index--;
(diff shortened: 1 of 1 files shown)


src/lit-html.ts
```diff
@@ -109,13 +109,11 @@ export function render(
...
  * https://github.com/PolymerLabs/lit-html/issues/62
  */
-const attributeMarker = `{{lit-${Math.random()}}}`;
-
-const textMarkerContent = '_-lit-html-_';
-const textMarker = `<!--${textMarkerContent}-->`;
-const attrOrTextRegex = new RegExp(`${attributeMarker}|${textMarker}`);
-const lastAttributeNameRegex =
-    /((?:\w|[.\-_$])+)=(?:[^"']*|(?:["][^"]*)|(?:['][^']*))$/;
+const marker = `{{lit-${String(Math.random()).slice(2)}}}`;
```
jridgewell: Removes the `0.` (particularly that "."), which is a special character in regexes. They don't really gain us any uniqueness, and just makes the regex work harder.

src/lit-html.ts
```diff
@@ -109,13 +109,11 @@ export function render(
...
+const nodeMarker = `<!--${marker}-->`;
```
jridgewell: Reusing the same marker text in our node marker allows us to use the insanely performant `String#indexOf` to test for markers.

src/lit-html.ts
```diff
@@ -109,13 +109,11 @@ export function render(
...
+const markerRegex = new RegExp(`${marker}|${nodeMarker}`);
+const nonWhitespace = /[^\s]/;
+const lastAttributeNameRegex = /([\w\-.$]+)=(?:[^"']*|"[^"]*|'[^']*)$/;
```
jridgewell: Just cleans up the groups.

src/lit-html.ts
```diff
@@ -194,24 +192,23 @@ export class Template {
...
-
-        // Do a first pass to count attributes that correspond to parts.
-        const attributesWithParts: string[][] = Array.prototype.filter.call(
-            attributes,
-            (attribute: Attr) =>
-                attribute.value.split(attrOrTextRegex).length > 1);
-        // Loop that many times, but don't use loop index for anything.
-        for (let i = 0; i < attributesWithParts.length; i++) {
+        let count = 0;
+        for (let i = 0; i < attributes.length; i++) {
```
jridgewell: A call to `Array#filter` is much slower.

src/lit-html.ts
```diff
@@ -194,24 +192,23 @@ export class Template {
...
+          if (attributes[i].value.indexOf(marker) >= 0) {
```
jridgewell: Gets rid of a useless split, and it's **way** faster.

src/lit-html.ts
```diff
@@ -194,24 +192,23 @@ export class Template {
...
+          }
+        }
+        while (count--) {
           // Get the template literal section leading up to the first
           // expression in this attribute attribute
           const stringForPart = strings[partIndex];
           // Find the attribute name
           const attributeNameInPart =
-              stringForPart.match(lastAttributeNameRegex)![1];
+              lastAttributeNameRegex.exec(stringForPart)![1];
```
jridgewell: Since `Symbol.match` came around, using `String#match` has gotten slower. Instead, just do the thing `#match` does directly, and save a byte. 😄

src/lit-html.ts
```diff
@@ -223,9 +220,9 @@ export class Template {
         }
       } else if (node.nodeType === 3 /* Node.TEXT_NODE */) {
         const nodeValue = node.nodeValue!;
-        const strings = nodeValue.split(attrOrTextRegex);
-        if (strings.length > 1) {
+        if (nodeValue.indexOf(marker) > -1) {
```
jridgewell: Have I mentioned how insanely fast `String#indexOf` is?

This really optimizes for the negative cases (`Text` nodes without any markers in them) which is likely much more common. Splitting with a regex (which is kinda slow) can come after we know for sure we have a hit.

src/lit-html.ts
```diff
@@ -251,21 +248,21 @@ export class Template {
                previousSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
               (nextSibling === null ||
                nextSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
-              nodeValue.trim() === '') {
+              !nonWhitespace.test(nodeValue)) {
```
jridgewell: `String#trim` is absurdly slow. Instead, search for a non-whitespace character.

src/lit-html.ts
```diff
@@ -251,21 +248,21 @@ export class Template {
...
           node.nodeType === 8 /* Node.COMMENT_NODE */ &&
-          node.nodeValue === textMarkerContent) {
+          node.nodeValue === marker) {
         const parent = node.parentNode!;
         // If we don't have a previous node add a marker node.
         // If the previousSibling is removed, because it's another part
         // placholder, or empty text, add a marker node.
-        if (node.previousSibling === null ||
-            node.previousSibling !== previousNode) {
+        const previousSibling = node.previousSibling;
```
jridgewell: Avoids a second DOM hit, _and_ save a few bytes. 😄

src/lit-html.ts
```diff
@@ -598,19 +596,14 @@ export class TemplateInstance {
...
 
-      const parts = this.template.parts;
-      let index = 0;
-      let partIndex = 0;
-      let templatePart = parts[0];
-      let node = walker.nextNode();
-      while (node != null && partIndex < parts.length) {
-        if (index === templatePart.index) {
-          this._parts.push(this._partCallback(this, templatePart, node));
-          templatePart = parts[++partIndex];
```
jridgewell: This would cause an out-of-bounds deopt every call to `#_clone`.

src/lit-html.ts
justinfagnani: If you think of a way to eliminate this Regex too, then 👍

src/lit-html.ts
```diff
@@ -194,24 +192,23 @@ export class Template {
...
-                attribute.value.split(attrOrTextRegex).length > 1);
-        // Loop that many times, but don't use loop index for anything.
-        for (let i = 0; i < attributesWithParts.length; i++) {
+        let count = 0;
+        for (let i = 0; i < attributes.length; i++) {
+          if (attributes[i].value.indexOf(marker) >= 0) {
+            count++;
+          }
+        }
+        while (count--) {
```
justinfagnani: I'd go with `while (count-- > 0)` to avoid the ToBoolean conversion

src/lit-html.ts
justinfagnani: Nice :)

src/lit-html.ts
```diff
@@ -598,19 +596,14 @@ export class TemplateInstance {
...
-      let node = walker.nextNode();
-      while (node != null && partIndex < parts.length) {
-        if (index === templatePart.index) {
-          this._parts.push(this._partCallback(this, templatePart, node));
-          templatePart = parts[++partIndex];
-        } else {
+      let index = -1;
+      for (let i = 0; i < parts.length; i++) {
+        const part = parts[i];
+        while (index < part.index) {
```
justinfagnani: `while (index++ < part.index) {` should work also

richardanaya:
This PR rocks, nice work

src/lit-html.ts
jridgewell: Can't do that, since it'd still increment `index` when `part.index` already matched.

```js
html`<div first=${thisWouldWork} second=${butNowIndexIncremented} third=${andICreateAnInfiniteLoop}></div>`

src/lit-html.ts
jridgewell: I tried. We could do it if we used `<!--marker-->` for both attr and node markers, but then I hit an issue with unquoted attributes ending the node. I was saving that for the next PR. 😉

justinfagnani:
Agreed with @richardanaya!

src/lit-html.ts
justinfagnani: Ah, good point

PR #169, opened 2017-11-05
Record NodePart's endNode index during cloning

### LINKED ISSUE #147: Optimize for NodeParts that are the only child
They don't need start and end nodes.

### LINKED ISSUE #148: Exception when expressions directly follow elements
See https://github.com/PolymerLabs/lit-html/pull/139

jridgewell:
This fixes the issue with `<a>some child</a>${expression}` attachment.
Since the traversal order is 0: `<a>`, 1: `"some child"`, 2:
`<!--marker-->`, we would find the marker and incorrectly record 2 as its location.
But 2 is inside the `<a>` tag, that's no good. This is because we assumed the index
was the previous sibling, but it can actually be the last child of a deep descendant of
the previous node.

Fixes #148.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -167,25 +167,18 @@ export class Template {
     // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
     const walker = document.createTreeWalker(
         content,
-        133 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
-               NodeFilter.SHOW_TEXT */
+        1157 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+               NodeFilter.SHOW_TEXT | NodeFilter.SHOW_DOCUMENT_FRAGMENT */
         ,
         null as any,
         false);
     let index = -1;
     let partIndex = 0;
-    const nodesToRemove: Node[] = [];
-
-    // The actual previous node, accounting for removals: if a node is removed
-    // it will never be the previousNode.
-    let previousNode: Node|undefined;
-    // Used to set previousNode at the top of the loop.
-    let currentNode: Node|undefined;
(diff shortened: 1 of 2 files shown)


src/lit-html.ts
```diff
@@ -167,25 +167,18 @@ export class Template {
     // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
     const walker = document.createTreeWalker(
         content,
-        133 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
-               NodeFilter.SHOW_TEXT */
+        1157 /* NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+               NodeFilter.SHOW_TEXT | NodeFilter.SHOW_DOCUMENT_FRAGMENT */
```
jridgewell: This is so we can backtrack if we're going to remove the first child of the template fragment. This is fine because the template is the only fragment we can ever encounter.

src/lit-html.ts
```diff
@@ -614,18 +596,21 @@ export class TemplateInstance {
           index++;
           walker.nextNode();
         }
-        this._parts.push(this._partCallback(this, part, walker.currentNode));
+        let node = walker.currentNode;
+        if (part.type === 'node') {
+          node = node.previousSibling!;
```
jridgewell: We could instead pass the `endNode` to `NodePart`s?

justinfagnani:
So https://github.com/PolymerLabs/lit-html/issues/147 is about getting ``html`<a>${0}</a>``` to produce no extra text nodes, but the result of this PR is a good thing too :)

jridgewell:
Lol, I meant #148.

jridgewell:
I think I've actually come up with a better algorithm, hold off on this one.

justinfagnani:
I think this PR is obsolete now...

Issue #56, opened 2017-08-18
Attribute parts with value `undefined` set attribute to string "undefined"


cdata:
Please refer to the following reduced repro case and read the console output: https://enshrined-noise.glitch.me/

### What I wanted

```
Element.outerHTML: <div foo></div>
```

### What I got

```
Element.outerHTML: <div foo="undefined"></div>
```

justinfagnani:
After talking, I'm unsure whether or not this is a bug. Calling setAttribute() with a value of `undefined` results in the attribute having the string value `"undefined"`.

The question to answer is how much an attribute binding should behave like setAttribute(). Should attribute bindings be _just_ setAttribute(), or should the try to fix the DOM API somewhat?

@sorvell @kevinpschaaf for opinions

cdata:
Definitely think this was a matter of ergonomics and not correctness for me. Consider this food for thought and not a bug report, because I'm also on the fence about what is the best way.

justinfagnani:
@sorvell @kevinpschaaf interested in your opinion here. Should `undefined` as an attribute value act like `setAttribute(undefined)` (set the attribute to the string `"undefined"`), or should it do something different, like remove attribute, or remove its value?

gdi2290:
`setAttribute` should only accept a string so anything else is type coercion (which should be avoided). If someone wanted undefined they can return a string value of `"undefined"`. If the value returned is `false || undefined || NaN || null` then the attribute should most likely be removed as you're most likely doing this for boolean values. So this usually boils down to an API like 
```html
<div foo="${ check ? '' : undefined }"></div>
```
returning `''` should output
```html
<div foo></div>
```
and not
```html
<div foo=''></div>
```
In some template engines most falsey values are converted to null otherwise you have to deal with  `false || undefined || NaN || null` (`''` is an exception)

Even though we would like to live in a world with an API decoupled from the DOM at the end of the day it's still the browser DOM API.

-from experience working with Angular's templating engine

justinfagnani:
The question for me, especially in lit-html core as opposed to lit-extended, is how much should an attribute binding behave like `setAttribute()`, and how does a full-binding relate to an interpolation. Would it be surprising that `<div foo=${undefined}>` behaves different than `setAttribute('foo', undefined)`, or different than `<div foo=" ${undefined} ">` (note the spaces).

gdi2290:
[...]

PR #170, opened 2017-11-05
Remove attributes with nullish values

### LINKED ISSUE #86: Allow conditional creation of attributes
lit currently doesn't allow you to conditionally create an attribute.

For example, if you would like to toggle 'checked' state on an input.

<input type="checkbox" checked>

There is no way to conditionally toggle the 'checked' attribute.

This may be related to https://github.com/PolymerLabs/lit-html/issues/56 but that seems to expect undefined to not serialize, but does not mention removing attributes.

jridgewell:
Fixes #86.

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -72,16 +72,7 @@ export const extendedPartCallback =
 
 export class PropertyPart extends AttributePart {
   setValue(values: any[], startIndex: number): void {
-    const s = this.strings;
-    let value: any;
-    if (s.length === 2 && s[0] === '' && s[1] === '') {
-      // An expression that occupies the whole attribute value will leave
-      // leading and trailing empty strings.
-      value = getValue(this, values[startIndex]);
-    } else {
-      // Interpolation, so interpolate
-      value = this._interpolate(values, startIndex);
-    }
+    const value = this._interpolate(values, startIndex);
     (this.element as any)[this.name] = value;
   }
 }
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -207,8 +207,7 @@ export class Template {
               lastAttributeNameRegex.exec(stringForPart)![1];
           // Find the corresponding attribute
           const attribute = attributes.getNamedItem(attributeNameInPart);
-          const stringsForAttributeValue =
-              attribute.value.split(markerRegex);
+          const stringsForAttributeValue = attribute.value.split(markerRegex);
           this.parts.push(new TemplatePart(
               'attribute',
               index,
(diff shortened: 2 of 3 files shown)


justinfagnani:
@jridgewell we still have a little discussion to do on the exact semantics we want wrt `null` and `undefined`, like in https://github.com/PolymerLabs/lit-html/issues/56#issuecomment-327623363

> The question for me, especially in lit-html core as opposed to lit-extended, is how much should an attribute binding behave like setAttribute(), and how does a full-binding relate to an interpolation. Would it be surprising that <div foo=${undefined}> behaves different than setAttribute('foo', undefined), or different than <div foo=" ${undefined} "> (note the spaces).

We'll talk about this next week, maybe in #56

jridgewell:
It seems Glimmer follows roughly this. If there attribute has no interpolation, it checks for nully values and removes the attributes if so. If there is interpolation, it filters nully values from the string. So:

```js
html`<div attr=${null}></div>`; // => <div>
html`<div attr="${null}"></div>`; // => <div>
html`<div attr=" ${null}"></div>`; // => <div attr=" null">, glimmer would do " "
html`<div attr="${null} "></div>`; // => <div attr="null ">, glimmer would do " "
html`<div attr=" ${null} "></div>`; // => <div attr=" null ">, glimmer would do "  "

html`<div attr=${false}></div>`; // => <div attr="false">
html`<div attr="${false}"></div>`; // => <div attr="false">
html`<div attr=" ${false}"></div>`; // => <div attr=" false">, glimmer would do " false"
html`<div attr="${false} "></div>`; // => <div attr="false ">, glimmer would do "false "
html`<div attr=" ${false} "></div>`; // => <div attr=" false ">, glimmer would do " false "
```

That all seems pretty good to me.

derappelt:
@jridgewell 
This maybe a solution, but  I think it can be confusing. There may be cases where you want to set an attribute to null. And in this cases attr=${null} or attr="${null}" would not do what you would expect.
What I would like more is something like:
`<div ${(condition) ? "attr" : ""}> or <div ${(condition) ? "attr" : null}>` if something like this even possible

jridgewell:
Attributes are strings, though. If you want `attr="null"`, pass (the string) `"null"` as the value.

> `<div ${(condition) ? "attr" : ""}>`

I have my eye on that (specifically, attribute bindings that aren't the value), just haven't gotten around to it yet.

zandaqo:
In theory, this can be left for a directive to do if we are dead set on keeping lit-html minimalist, but I think the proposed solution here is the way to go. Do we have any ETA on this one?

LarsDenBakker:
Would be great to have some resolution on this point, I use quite a lot of boolean properties and it's inconvenient to use lit templates at the moment :)

justinfagnani:
I'd like to align here with the template instantiation proposal, at least for lit-html core. lit-extended can offer more opinionated choices. I think as an initial step we can easily offer a toggle directive, it's just a few lines of code.

zandaqo:
> I'd like to align here with the template instantiation proposal, at least for lit-html core.

As I understand the current proposal, it features a different syntax, I wonder how lit-html is supposed to fit into it.

> I think as an initial step we can easily offer a toggle directive, it's just a few lines of code.

Should we go with the directive then? At least until that proposal shapes up which can take some time, I assume.

justinfagnani:
> As I understand the current proposal, it features a different syntax, I wonder how lit-html is supposed to fit into it.

This isn't necessarily syntax. If the template instantiation spec uses undefined to remove attributes, then I'd want to use that. If the spec uses some special attribute name, likewise. If it uses syntax inside the expression, then a directive would be the way.

justinfagnani:
@jridgewell after discussing with the team, I think we want to go with a syntactical opt-in, like in #197. We're still figuring out the specific semantics of that before merging.

Thanks for the PR!

PR #171, opened 2017-11-05
Use new helper methods in repeat directive

jridgewell:


src/lib/repeat.ts
```diff
@@ -92,12 +98,7 @@ export function repeat<T>(
 
     // Cleanup
     if (currentMarker !== part.endNode) {
-      const end = part.endNode;
-      for (let node = currentMarker; node !== end;) {
-        const n = node.nextSibling!;
-        container.removeChild(node);
-        node = n;
-      }
+      removeChild(container, currentMarker, part.endNode);
```
justinfagnani: `removeChild` -> `removeNodes`

jridgewell: Lol, fixed.

PR #173, opened 2017-11-05
Added test that seams to loop forever.

sunesimonsen:
I was trying to extend my [property based](https://github.com/sunesimonsen/lit-html-tests) test that earlier found a swap bug, I had a hard time to not get into infinite loops. I think I have reproduced a minimal failing test. But I would like it to be verified.


justinfagnani:
Tests added and bug fixed in #177

PR #174, opened 2017-11-05
 Allow all valid HTML attribute names.

### LINKED ISSUE #172: Support Angular-style attributes
I'm trying to use angular-style attributes for property/event binding using a custom `render` extension, but I run into an error when trying to call `html` with that syntax.

```js
import {html} from './path/to/lit-html/lit-html.js';

html`<some-element [prop]=${5} (click)=${() => alert('clicked!')}></some-element>`;
```

```
  Cannot read property '1' of null
    <unknown> at   new Template at /components/lit-html/lit-html.js:172:0
    <unknown> at         litTag at /components/lit-html/lit-html.js:43:0
    <unknown> at           html at /components/lit-html/lit-html.js:31:0
```

The `lastAttributeNameRegex` used at [`src/lit-html.ts#L209-210`](https://github.com/PolymerLabs/lit-html/blob/eeafa95c820672ec95231afbc5861d1755d46457/src/lit-html.ts#L209-L210) doesn't match because of the `[]` and `()` in the attribute names.

I've seen comments in #168 that this regex might disappear in the future ([link](https://github.com/PolymerLabs/lit-html/pull/168#discussion_r148941842)). Is this future coming soon or would you accept a PR that adds `[]()` to `lastAttributeNameRegex`?

justinfagnani:
Fixes #172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+* Allow all valid HTML attribute names, including emoji and Angular-style
+  (foo)= and [foo]= names.
 
 ## [0.7.0] - 2017-10-06
 
--- a/src/lib/repeat.ts
+++ b/src/lib/repeat.ts
@@ -12,13 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {
-  directive,
-  DirectiveFn,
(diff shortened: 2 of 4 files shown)


justinfagnani:
`npm run lint` which is run from `npm test`

justinfagnani:
which... I forgot to run. CI is calling wct directly, oops.

jridgewell:
It doesn't seem to catch any of the whitespace changes you made?

justinfagnani:
That's just the formatter with `npm run format`. It uses clang-format: https://github.com/PolymerLabs/lit-html/blob/master/package.json#L26

PR #175, opened 2017-11-06
Split lib into module files

jridgewell:
It's getting a little crowded in here.


daKmoR:
@justinfagnani is this (or something similar) going to happen soon?

just so to know it it makes sense to do pull requests to the current codebase or better wait for a restructure?

justinfagnani:
I'm not necessarily opposed to this, but I think I'd like most files to live in lib/ and maybe we should try this again after clearing the deck of some upcoming changes.

justinfagnani:
@jridgewell do you mind if I close this for now and we'll split after 0.8?

PR #176, opened 2017-11-06
Repeat duplicate

jridgewell:
Also optimizes for Object keys, in which case we can use a `WeakMap`,
and skip map cleanups.

Fixes #173.


jridgewell:
Actually, I thought of a better way.

PR #177, opened 2017-11-06
Fix infinite loop with repeated repeat node

jridgewell:
A better fix to #173.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -661,10 +661,6 @@ export const reparentNodes =
      end: Node | null = null,
      before: Node | null = null): void => {
       let node = start;
-      if (before && before === end) {
-        // Prevent a really fun infinite loop.
-        return;
-      }
       while (node !== end) {
         const n = node!.nextSibling;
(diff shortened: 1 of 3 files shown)


sunesimonsen:
@jridgewell nice 👍

src/lit-html.ts
```diff
@@ -661,6 +661,10 @@ export const reparentNodes =
      end: Node | null = null,
      before: Node | null = null): void => {
       let node = start;
+      if (before && before === end) {
+        // Prevent a really fun infinite loop.
```
justinfagnani: I think there are other similar cases that would cause an infinite loop still, like `before` being a preceding sibling of `end`, but a following sibling of `start`. Since we should document not to do that anyway, and this only protected against one case, I wonder if we should move the check into repeat.ts: https://github.com/PolymerLabs/lit-html/blob/master/src/lib/repeat.ts#L77

jridgewell: That's where I had it originally, I'll move it back.

PR #178, opened 2017-11-07
Optimize lastAttributeNameRegex

jridgewell:
A pretty easy optimization to `lastAttributeNameRegex`. Giving it a starting bound improves the performance by 2-3x in Chrome, 3x in Safari, and 5-6x in Firefox.

See https://regex101.com/r/etquoe/2/, and try deleting the space character class at the beginning, and see the match time in the upper right side of the input.

Then try this [very long attribute name](https://regex101.com/r/etquoe/3) on on for size. It's infinitely faster (division by zero error...). And then there's improvement to super long [quoted](https://regex101.com/r/etquoe/4) and [unquoted](https://regex101.com/r/etquoe/6) attribute values.

The space between a tagname and an attribute, or between two attributes, is required by https://www.w3.org/TR/html5/syntax.html#attributes-0.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -134,10 +134,10 @@ const nonWhitespace = /[^\s]/;
  *  * Followed by:
  *    * Any character except space, ('), ("), "<", ">", "=", or
  *    * (") then any non-("), or
- *    * (') then any non-(')
+ *    * (') then lastAttributeNameRegexany non-(')
  */
-const lastAttributeNameRegex =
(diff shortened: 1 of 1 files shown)


jridgewell:
Ping @justinfagnani

jridgewell:
Ping @justinfagnani, a pretty simple change.

justinfagnani:
Hey @jridgewell sorry, I had TPAC and then some travel. Looks good!

PR #179, opened 2017-11-07
Leave whitespace text nodes intact

jridgewell:
Our whitespace stripper doesn't understand CSS's whitespace collapsing rules. For instsance,

```js
html`<span>test<span> <a href="">ing</a>`
```

should read "test ing".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6272,6 +6272,12 @@
       "integrity": "sha1-gIudDlb8Jz2Am6VzOOkpkZoanxo=",
       "dev": true
     },
+    "string_decoder": {
+      "version": "0.10.31",
+      "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-0.10.31.tgz",
+      "integrity": "sha1-YuIDvEF2bGwoyfyEMB2rHFMQ+pQ=",
+      "dev": true
+    },
     "string-width": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/string-width/-/string-width-1.0.2.tgz",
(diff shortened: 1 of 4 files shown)


daKmoR:
fancy enough this with replacement of "for of" loops will make lit-html run at least in IE11 (still failing tests but not a fatal error in simple usecases as it is now)

jridgewell:
Curious. What about the old code broke in IE?

daKmoR:
hmm maybe it's just because I write templates mostly like this 
```
      return html`
        <style>
          h3 {
            color: orange;
          }
        </style>
        <h3>${this.header}</h3>
        <p>some more text</p>
      `;
```

so maybe preserving the whitespace in the start/end now helps IE11...

jridgewell:
@justinfagnani: Bugfix here.

justinfagnani:
@jridgewell I'm a little unsure on what we should do with whitespace stripping overall. I'm open to the PR with the idea that we figure out a way to add back and configure whitespace stripping in the future. It could be a different tag...

cc @kevinpschaaf and @sorvell

jridgewell:
Rebased.

justinfagnani:
> What's the need?

I pinged @kevinpschaaf for use cases in Polymer. Minification and perf (it's faster to clone less nodes, but we don't want users to have to write ugly templates to be as fast as possible) are the reasons for stripping whitespace that I know of, and so in Polymer there's an opt-out. I'm not sure if there's other reasons to do stripping.

justinfagnani:
@kevinpschaaf I'm inclined to merge this for 0.8, since it fixes a real problem with removing meaningful spaces, and then add in an alternate tag which strips whitespace later.

kevinpschaaf:
A separate tag that strips whitespace is probably fine.  Could consider a factory that takes options to generate a tag if there start to be a lot of permutations of options.

justinfagnani:
@kevinpschaaf I was thinking along the same lines

kaste:
I think a relative good heuristic is that we strip whitespace nodes that *begin* with the new line `\n` char. This can drop a lot of nodes the user not intended bc she just wanted to format the code she wrote.

```
html `
  <div>
    <p></p>
    <p></p>
  </div>
`
``` 

Here, ideally we would remove 5 nodes (and keep 3) (Did I count correct? LOL). Keep also in mind that we cannot write

```
return html
`here we go
`
```

bc of ASI. So we really always do

```
return html`
     Here we go
`
```

So we by design always have whitespace at the beginning and probably the end of the literal.

PR #186, opened 2017-11-13
Add prepare script to allow linking or github installs

bgotink:
The prepare script is run after installation and before publish, meaning that
it is run during npm-link and when installing from the github repository,
e.g. when fetching a prereleased version.

This means we can use

```json
{
  "//": "...",
  "devDependencies": {
    "lit-html": "PolymerLabs/lit-html#master",
  }
}
```

and it'll work, rather than leave us with a package containing the typescript sources but no build output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "quicktest": "wct -l chrome -p --npm",
     "checksize": "uglifyjs lit-html.js -mc --toplevel | gzip -9 | wc -c",
     "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i",
-    "lint": "tslint --project ./"
+    "lint": "tslint --project ./",
+    "prepare": "npm run build"
   },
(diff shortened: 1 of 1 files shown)


daKmoR:
I would even go as far as adding a `dist` folder to allow also for usage in non es module applications...
Proposal:
https://github.com/PolymerLabs/lit-html/pull/195

would solve your "problem" as well right?

bgotink:
That's indeed another way to tackle this.

My personal preference is to use a `prepare` script though, as aving built artefacts in the git repo can lead to inconsistencies when someone forgets to rebuild the artefacts in a PR. It'll also lead to conflicts in all other PRs once one PR gets merged. This makes contributing to lit-html more difficult imo, especially for novice git-users, and it introduces extra things that can go wrong in the pull request process.

justinfagnani:
Agreed, this is good for local development, and checking in artifacts can lead to problems. Thanks @bgotink

PR #188, opened 2017-11-14
Failing test for re-rendering repeated items containing duplicates

sunesimonsen:
I ran [my property based test](https://github.com/sunesimonsen/lit-html-tests/blob/master/test/repeat.spec.js#L38) on the latest master and got this result:

screen shot 2017-11-14 at 19 49 46

This PR adds a failing test for the found case.

### DIFF
--- a/src/test/lib/repeat_test.ts
+++ b/src/test/lib/repeat_test.ts
@@ -145,6 +145,19 @@ suite('repeat', () => {
       assert.equal(container.innerHTML, `<li>item: 2</li>`);
     });
 
+    test('can rerender repeated items with no changes', () => {
+      const t = (items: number[]) =>
+          html`${repeat(items, (i) => i, (i: number) => html`
+          <li>item: ${i}</li>`)}`;
+
+      render(t([666, 666]), container);
+      render(t([666, 666]), container);
+
+      assert.equal(
+          container.innerHTML,
+          `<li>item: 666</li><li>item: 666</li>`);
+    })
+
     test('can insert an item at the beginning', () => {
       let items = [1, 2, 3];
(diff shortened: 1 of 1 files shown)


src/test/lib/repeat_test.ts
```diff
@@ -145,6 +145,19 @@ suite('repeat', () => {
...
+      const t = (items: number[]) =>
+          html`${repeat(items, (i) => i, (i: number) => html`
+          <li>item: ${i}</li>`)}`;
+
+      render(t([666, 666]), container);
+      render(t([666, 666]), container);
+
+      assert.equal(
+          container.innerHTML,
+          `<li>item: 666</li><li>item: 666</li>`);
```
jridgewell: This should be just `<li>item: 666</li>`.

sunesimonsen: It should render two items right?

sunesimonsen: It should render two items right?

jridgewell: Nah, it'll find the same part (with id `666`), and update it. See https://github.com/PolymerLabs/lit-html/pull/188/files/cab9000cc4d6a777a47122a3d829c777cd3810a5#diff-df279e375bd22a90dee8ee7e0ff80d13R125.

sunesimonsen:
@jridgewell you are right - I was keying the items by value which is probably not allowed. It would be great if it warned me that I'm using the same key twice, but I guess that would affect performance.

The updated test passes with 10000 iterations so it should definitely be stable under swaps now 👍 

screen shot 2017-11-15 at 09 30 39

PR #195, opened 2017-11-19
feat: add dist for direct consumption

daKmoR:
For easy consumption in applications that do not use es modules yet it would be nice to have a pre build dist. Also it would be nice to have some ready js code when using the git repository directly.

Usage would be
```
// using es modules
<script type="module">
  import { html, render } from '../lit-html/dist/es/lit-extended.js';
  render(html`<p>${content}</p>`, targetNode);
</script>

// using "plain" script tags
<script  src="../lit-html/dist/script/lit-extended.js"></script>
<script>
  Lit.render(Lit.html`<p>${content}</p>`, targetNode);
  // or
  var render = Lit.render;
  var html = Lit.html;
  render(html`<p>${content}</p>`, targetNode);
</script>
```


bgotink:
This PR changes the path to the script from `../lit-html/lit-html.js` to `../lit-html/dist/es/lit-html.js`. Note that the `lit-html` npm package will still contain the `lit-html.js` script, but when installing from github the package will only contain `dist/es/lit-html.js`.
Maybe it would be better to move the output of typescript to a temporary folder and make `lit-html.js` the output of the rollup build? That's how I set up the build in [lit-html-brackets](https://github.com/bgotink/lit-html-brackets).

Having built artefacts in the git repo can lead to inconsistencies when someone forgets to run `gen-dist` in a PR and it'll lead to conflicts in all other PRs once one PR gets merged. This makes contributing to lit-html more difficult imo, especially for novice git-users.
Adding a `prepare` script also allows installing this package from the git repo directly, without this added complexity.

I'm definitely pro using rollup to create one single lit-html ESM, once #175 lands, and to create a build for the pre-ES2015 world.
I personally prefer UMD over IIFE output to allow consumption using CJS/AMD as well, in case someone uses systemjs/requirejs or a workflow with browserify/rollup/webpack and CJS. If we have UMD output we can point the `main` property of the `package.json` to the UMD build and the `module` to the ESM build for improved interop with existing tools.

daKmoR:
Sure the npm release package would need to change as well... probably there should be no "default" lit-html/lit-html.js - you should specifically choose which version you need.

having the fragments not in the repo probably makes it a little harder to as you will also have to install all required build tools e.g. ts and rollup in this case... but yeah then again might be a good tradeoff...

so you say we would actually need 3 versions to satisfy all landscapes? as umd can't be used directly with a script tag in the browser?

```
dist/es/lit-html.js // for modern apps using es modules
dist/umd/lit-html.js // for CJS/AMD
dist/script/lit-html.js // for "plain" script tag (using iife)
```

So I could change this to have these 3 config in rollup. Then use the `prepare` for automatic building if using git directly. 
And the npm package should probably "only" contain the dist folder (still keep dist to keep the dirs for git and npm the same)

sounds like a plan?

bgotink:
UMD builds created by rollup support CJS, AMD and globals, so those builds can be loaded in a regular script tag.

I'm not quite sure what the current best practices are for delivering front-end packages using npm. Angular component libraries tend to output a ES5 UMD build (`main` in `package.json`), a ES5 ESM build (`module`) and a ES2015+ ESM build (`es2015` in `package.json`). Outside of Angular libraries I don't have a clue what the norm is.

---

If a package is installed from github, and that package has a `prepare` script, npm handles everything for you: 

> If the package being installed contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.
> ~ https://docs.npmjs.com/cli/install

justinfagnani:
A few thoughts here:

 * I don't want to move the imports into a folder like `/es`, this disadvantages standard usage and bakes in the format into the path.
 * How do we deal with some libraries importing the JS module version and others import the IIFE version? I don't think we can easily have references to one rewritten to references to the other, so we'll get code duplication in applications.
 * Why IIFE output and not UMD, AMD, SystemJS, CommonJS, etc? Sorry for the rhetorical question - just highlighting that my opinion here is that the consumer should choose the transformation they possibly need to turn standard JS modules into something particular to their environment.
 * Installing from GitHub isn't a supported use case. GitHub is for development and contains only the original source, npm is for distribution. We can think of an alternate distribution mechanism if npm isn't suitable, but I do not want to check in build artifacts.

daKmoR:
- ad 1: totally get it
- ad 2: I guess you should always use ONE import method for your whole app right?
- ad 3: so lit html will only be directly usable with es modules? e.g. if I have an app that uses Html Imports I would need to have a build in one of my components (probably a lit-element)? so the "work" would just be pushed to the user... => probably makes adoption harder but if that is the decision I can live e.g. work around it
- ad 4: would still be nice while developing (and it's only a `"prepare": "npm run build"` which would not have any negative effects?)

justinfagnani:
Item 2 (multiple legit import paths): we're not always in a single-app situation. If lit-html is used by several web components implemented independently by different people, they each may choose a different import path and the app wouldn't be able to do anything about it. That's why my question was phrased in terms of de-duplication. Once an app uses elements with different styles, it'd have to deal with the deduplication itself.

Item 3: Yes, I think it's fine that lit-html is only importable _in source_, as a JS module. It's the only standard and common module format, and all compilers support it at this point. Anyone using CommonJS would already have to be using a bundler at dev time, and the popular bundlers support JS imports. If you use HTML Imports you can convert the script tag that you want to use lit-html form into a script/type=module tag and import lit there. This works with both native HTML Imports in Chrome and the HTML Imports polyfill. Then the tooling just needs to do this right thing for browsers w/o module support. Our tooling in on the way, and the `polymer serve` command already supports this workflow. If there's non-Polymer tooling that works with HTML Imports, we should make sure it works with modules.

Item 4: During development you should be able to use any dev server that compiles modules to a custom loader. Like I mentioned,`polymer serve` compiles modules for some browsers, and I'm sure webpack dev server supports this as well.

daKmoR:
ad 2: puuh yeah if that would happen you are ***cked - so  I would say THAT is the work of the "app" / the user to make sure it does not happen... but I guess if you load webcomponents via html imports (internally load lit-html with a iife) you should not expect a webcomponent you load via es import (internally loads lit-html via es import) to work nicely together... I mean it would be awesome but I doubt it...

ad 3: let's have a very specific "problem"
https://github.com/daKmoR/grain-lit-element/blob/master/external-imports/lit-extended.html
here I load it via "my fork" with an iife... so that works fine with html imports... but it does not if I load it via es module (if the browser does not support es modules THEN it will be converted to require and THAT will mess up all the timings)
I tried already 2 times to mix html imports and es modules but so far it never worked out because of timing issues... if you have a way I would be more then happy to see that... 

Some more details:
Currently I have 2 Versions (one *.html and one *.js I have a simple "converter" for both directions) :
- for HTML Imports
  - https://github.com/daKmoR/grain-lit-element/blob/master/GrainLitElement.html
  - https://github.com/daKmoR/grain-lit-element/blob/master/external-imports/lit-extended.html
- for es module import
  - https://github.com/daKmoR/grain-lit-element/blob/master/GrainLitElement.js
  - https://github.com/daKmoR/grain-lit-element/blob/master/external-imports/lit-extended.js

ad 4: I mean while developing on lit-html (e.g. testing lit-html master in my web component)

PS: again not a real problem as I can work around it if I provide a build version of lit-html in this element... it kinda feels wrong... but maybe it's the way to go...

justinfagnani:
2: The app _can't_ make sure that there's only one canonical import if the library publishes more than one. Like I said, the app could use two third party components where one loads the global build and the other imports the module. The only way for the app to ensure this doesn't happen is to not use any third party libraries, and that's not a restriction or responsibility I want to put on apps, especially since with modules we have a way forward out of the traditional mess: only publish one canonical module. That _guarantees_ the proper import graph.

3: With modules you don't need to create .html wrappers on JS files anymore. The only reason that was necessary is because `<script>` tags do not de-duplicate, but HTML imports do, so by wrapping and only using the HTML import, you get ral de-duping. With `<script type=module>` we _do_ get de-duping, so all you have to do is import the library.

I don't think you need anything in that external-imports/ imports folder. Just import lit-html directly in GrainLitElement.html:

```html
<link rel="import" href="../grain-element/GrainElement.html">
<script type="module">
import {render} from 'lit-html/lit-html.js';
</script>
```

In fact, your library highlights the duplication problem very well. Since grain-lit-element imports lit-html via a global build,if you use it with any other library that imports lit-html.js via module imports you'll get duplication. This duplication might in fact cause bugs, because some `instanceof` checks in lit could fail if you try to compose templates from one lit-html version with another.

4: I don't think any build is necessary to develop on lit-html itself. I don't use one. Just `polymer serve --npm` from within the lit-html directory. Like I mentioned, other dev servers, like WebPack's will compile modules for older browsers, though none likely work with HTML imports.

bgotink:
About 4: While it may not be useful for lit-html itself, allowing installs from github makes it a lot easier for dependants to test unreleased versions (e.g. to test a bugfix) without having to rely on npm-link. This is especially useful when using CI environments like Travis where npm-link doesn't work when caching node_modules.  
Example: an unreleased version of lit-html extension A requires a fix which hasn't been released yet. Rather than wait until a new version of lit-html is released, the dev of A wants to continue working on their new features, preferably with a working Travis CI build.

justinfagnani:
@bgotink you're right that making this easier to consume as a git dependency would help in some cases. I merged #186 which will be good to get going with a fresh git clone or for prepublishing, but I'm not sure how it'll do for git dependencies if devDependencies aren't install. Can't hurt though.

I think that's a bit separate from this issue though, which even if we're not checking in artifacts, still would mean that there's more than one version of the library available for import, even when there's not multiple version of the package in an application.

daKmoR:
@justinfagnani I'm really sorry for all the tough love :p

if I keep everything within `script type="module"` and only use the "elements" within the other components it actually works fine :)
that is way nicer :)

thx for staying tough on this one - I tried mixing script with script module and that where it failed... (combining es module import and html import seem to work out)

thx again :)

daKmoR:
@justinfagnani ohhh one question... any idea when the "transformation" from es module to requirejs that is currently done in the latest polyserve will be available in polymer build?

HitkoDev:
@justinfagnani #186 causes problems when TypeScript ^2.6 is installed in the same project (breaking the whole `npm install` command if lit-html is listed as a dependency), and installing with `--production` flag won't work unless there's a (compatible) TS compiler available. I believe this would be a better solution, as it makes the whole package independent and more suitable for "stable" usage.

justinfagnani:
@hitkodev can you file a new issue for that?

PR #197, opened 2017-11-20
Flag attribute syntax for lit-extended ("?" after attribute name).

brunnerh:
I thought that this might be a nice syntax for the extended usage, though the use cases for when setting a content attribute is desired over just using a DOM property are probably limited.

```html
<input readonly?="${true}"/>  => <input readonly=""/>
<input readonly?="${false}"/> => <input />
```

Well, just a suggestion more than anything else.

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -61,6 +61,11 @@ export const extendedPartCallback =
               return new AttributePart(
                   instance, node as Element, name, templatePart.strings!);
             }
+            if (templatePart.name!.endsWith('?')) {
+              const name = templatePart.name!.slice(0, -1);
+              return new FlagAttributePart(
+                  instance, node as Element, name, templatePart.strings!);
+            }
             return new PropertyPart(
                 instance,
                 node as Element,
@@ -112,5 +117,24 @@ export class EventPart implements Part {
     if (listener != null) {
       this.element.addEventListener(this.eventName, listener);
     }
+  }
+}
+
+export class FlagAttributePart extends AttributePart {
+  setValue(values: any[], startIndex: number): void {
+    const s = this.strings;
+    if (s.length !== 2 || s[0] !== '' || s[1] !== '') {
+      // Has to be single expression
+      throw new Error('Flag attributes may only contain a single expression.');
+    }
+
+    const value = values[startIndex];
+    if (typeof value !== 'boolean')
+      throw new Error('Value has to be boolean.');
+
+    if (value)
+      this.element.setAttribute(this.name, '');
+    else
+      this.element.removeAttribute(this.name);
   }
 }
--- a/src/test/lib/lit-extended_test.ts
+++ b/src/test/lib/lit-extended_test.ts
@@ -51,6 +51,15 @@ suite('lit-extended', () => {
       assert.equal(container.innerHTML, '<div foo="1bar2baz3"></div>');
     });
 
+    test('renders flag attribute', () => {
+      render(html`<div foo?="${true}"></div>`, container);
+      assert.equal(container.innerHTML, '<div foo=""></div>');
+    });
+    test('does not render flag attribute', () => {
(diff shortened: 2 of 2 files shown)


brunnerh:
Signed the CLA 🎉

shutdown256:
@brunnerh nice job you beat me to the PR for the same issue. However my implementation has 2 differences:

1.  it treats null and undefined as false - this behavior is consistent with polymer and feels logical: unresolved/uninitialized expressions will not set (e.g. hidden) + .setAttribute('a', null) sets attribute as flag so why change it?
2.  no special syntax other than $= required. In rare case that you actually want to set attribute as "true" or "false", just use a ? "true" : "false" - I believe (and it it just my opinion) that when most people set attribute to bool they expect default result to be flag.

brunnerh:
@shutdown256 Personally, i prefer it if things are explicit and prevent unwanted behavior as much as possible. In this case, the attribute is explicitly set to handle the case of flags, which is a boolean operation, hence it seems reasonable to throw an error if anything but a boolean is passed it.

If `null` or `undefined` are passed in, that may very well be the cause of an error somewhere else. If you adjust the `$=` behavior to handle the flag case as well, it kind of makes sense to have that bit of black magic in there. The question is whether that is desirable or not and there probably is a case to be made either for either option.

To me the `?=` feels quite intuitive and it makes both intention and expected behavior very clear.

shutdown256:
@brunnerh That is why I mentioned it: to hopefully start some discussion, I feel a lot of people want this behavior in some form and now the question is what it should be.

justinfagnani:
@brunnerh and @shutdown256 - like I mentioned in #204, I think we want to go with the suffix solution presented here. I'm going to consider the specific values we trigger removal on though. I like explicit, but there might be arguments for removing on falsey values, or only `undefined` - there are some weird attributes out there.

What I'm wondering is if there are attributes that need to be able to take a string value _or_ be removed. This PR wouldn't support that because the value has to be a booean.

zandaqo:
> What I'm wondering is if there are attributes that need to be able to take a string value or be removed. This PR wouldn't support that because the value has to be a boolean.

That would a better, more general solution in my opinion. We can remove on falsey values (false/null/undefined) and have specified string value otherwise. E.g.:
```html
<div data-value?="${someValue}"/>  
//=> <div data-value="somevalue"/> // if someValue === 'somevalue'
//=> <div /> // if someValue === false/null/undefined
```

On a related note, can we have this in the core instead of lit-extended? I prefer to handle events in my own framework (which might conflict with lit-extended), and setting properties is done by a one-line directive, so I don't find much use for lit-extended.

justinfagnani:
Ok, so after talking with teammates some more, I'm now leaning to the `?` suffix opting in to boolean-ness based on the truthyness, so that `false`, `0`, `""`, `null`, `undefined`, and `NaN` remove an attribute and all other values set it's value to `""`.

I think any choice here is going to have some weird edge cases for interpreting a value, which might need to be addressed inside the expression. With this choice the weird one is how empty strings are interpreted. A `""` will remove the attribute, while `"a"` would set it's value to `""`. But like I said, other options have their weirdness too.

zandaqo:
@justinfagnani As I said, I'd prefer to set it to optionally supplied value, but whatever. Can we at least have some way to prevent setting attribute in the directive? Right now no matter what a directive does, lit-html sets an attribute, the only way I found around it is to asynchronously remove the attribute after the render, resulting in a directive like this:

```javascript
const toggle = (condition, value = true) => directive((part) => {
  const method = condition ? 'setAttribute' : 'removeAttribute';
  const delay = setTimeout(() => {
    part.element[method](part.name, value);
    clearTimeout(delay);
  }, 0);
  return value;
});
```

That's pretty ugly not to mention inefficient.

justinfagnani:
Thanks for the PR @brunnerh! I commited a similar change in #259 that works against truthy and falsey values.

PR #200, opened 2017-11-27
Rename the word from "template tag" to "tagged template literal"

axross:
```
html`<p>Lorem Ipsum</p>`
```

is called [Tagged Template Literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) generally.

I think we should use general terminology. There was few "template tag" in docs, and I replaced it with "tagged template literal".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [0.7.0] - 2017-10-06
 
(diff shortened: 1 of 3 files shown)


axross:
I signed it!

justinfagnani:
A tagged template literal is the combination of a template tag function applied to a template literal. So ``html`hi``` is a tagged template literal, but `html` in is own is a template tag. I think in the cases you changed in referring to tag functions, but not actual tagged template literals.

PR #202, opened 2017-11-27
Do not use empty text nodes in the template element

### LINKED ISSUE #201: IE: importNode doesn't clone empty text nodes
In IE, the following works:

```ts
render(html`<div>0 ${1} ${2} 3</div>`, container);
assert.equals(container.innerHTML, '<div>0 1 2 3</div>');
```

and the following doesn't

```ts
render(html`<div>${1} ${2}</div>`, container);
assert.equals(container.innerHTML, '<div>1 2</div>');
```

IE11's `document.importNode` removes empty text nodes, which means the cloned template doesn't match the original template anymore, leading to random breakage when rendering lit-html templates.

The following code shows the problem:

```ts
const result = html`<div>${1} ${2}</div>`;

console.log('pre clone:');
const div = result.template.element.content.childNodes[0];
console.log(div.childNodes.length);
for (let i = 0; i < div.childNodes.length; i++) {
  console.log(div.childNodes[i], JSON.stringify(div.childNodes[i].nodeValue));
}

console.log('post clone:');
const clonedDiv = document.importNode(result.template.element.content, true).childNodes[0];
console.log(clonedDiv.childNodes.length);
for (let i = 0; i < clonedDiv.childNodes.length; i++) {
  console.log(clonedDiv.childNodes[i], JSON.stringify(clonedDiv.childNodes[i].nodeValue));
}

```

In Chrome it logs:

image

In IE11 it logs:

image

Note: examples in typescript, which were run transpiled to a IE-compatible javascript.

EDIT: fixed typo in the test code which produced the screenshots: the `nodeValue`s of the original div's children were used instead of the clone's children.

bgotink:
Empty text nodes aren't cloned by IE when using document.importNode.
Putting content in the text ensures that they will be cloned by IE.

This increases the number of passing tests on IE11 from 40 to 93.
One failure remains for table handling, a couple are due to a different ordering of attributes in IE and the other remaining failures are due to missing `Promise`, `Array.from`, `String#startsWith`.

Fixes #201 
This is an alternative to #143 

Open questions:

- The nodes need a unique content. Given that we already have a unique marker, I just reused that instead of creating another unique string. This might be confusing, so maybe for readability's sake it should be a separate string?
- An alternative to the unique string could be Unicode codepoint U+2060 "word joiner", which could be left after cloning rather than having to remove after every clone. I don't know enough about this codepoint to know what the impact is of replacing empty text nodes with nodes containing this character (i.e. are there cases where this behaves differently from an empty text node?), so I opted to not use it.

EDIT: added some open questions

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -260,12 +260,12 @@ export class Template {
           // We keep this current node, but reset its content to the last
           // literal part. We insert new literal nodes before this so that the
           // tree walker keeps its position correctly.
-          node.textContent = strings[lastIndex];
+          node.textContent = strings[lastIndex] || marker;
 
           // Generate a new text node for each literal section
           // These nodes are also used as the markers for node parts
           for (let i = 0; i < lastIndex; i++) {
-            parent.insertBefore(document.createTextNode(strings[i]), node);
+            parent.insertBefore(document.createTextNode(strings[i] || marker), node);
             this.parts.push(new TemplatePart('node', index++));
           }
         } else {
@@ -300,7 +300,7 @@ export class Template {
         const previousSibling = node.previousSibling;
         if (previousSibling === null || previousSibling !== previousNode ||
             previousSibling.nodeType !== Node.TEXT_NODE) {
-          parent.insertBefore(document.createTextNode(''), node);
+          parent.insertBefore(document.createTextNode(marker), node);
         } else {
           index--;
         }
@@ -310,7 +310,7 @@ export class Template {
         // We don't have to check if the next node is going to be removed,
         // because that node will induce a new marker if so.
         if (node.nextSibling === null) {
-          parent.insertBefore(document.createTextNode(''), node);
+          parent.insertBefore(document.createTextNode(marker), node);
         } else {
           index--;
         }
@@ -432,6 +432,15 @@ export class NodePart implements SinglePart {
     this.instance = instance;
     this.startNode = startNode;
     this.endNode = endNode;
+
+    if (startNode.nodeValue === marker) {
+      startNode.nodeValue = '';
+    }
+
+    if (endNode.nodeValue === marker) {
+      endNode.nodeValue = '';
+    }
+
     this._previousValue = undefined;
   }
 
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -257,17 +257,24 @@ export class Template {
           // We have a part for each match found
           partIndex += lastIndex;
 
-          // We keep this current node, but reset its content to the last
-          // literal part. We insert new literal nodes before this so that the
-          // tree walker keeps its position correctly.
-          node.textContent = strings[lastIndex] || marker;
-
           // Generate a new text node for each literal section
           // These nodes are also used as the markers for node parts
           for (let i = 0; i < lastIndex; i++) {
-            parent.insertBefore(document.createTextNode(strings[i] || marker), node);
+            // IE doesn't clone empty text nodes, so use comments instead
+            parent.insertBefore(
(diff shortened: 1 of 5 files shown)


daKmoR:
finally could test it :) works very nicely - I like this approach way more :)
I would also opt in using a different marker?

Are you going to fix all the `Promise, Array.from, String#startsWith` "problems"? they should be pretty easy - or should I help out here? - I think IE11 with 100% passing tests is in near reach :)

bgotink:
I've just opened #211 to fix the remaining issues. Only one failing test remains, caused by a bug in the template polyfill.

daKmoR:
hmm I would opt more into only load polyfills that are really needed... 
e.g. Promise, Template, Symbol (I don't think we need more)

Array.from, String#startsWith code can easily be rewritten (use a tmp array, use indexOf) so they just work out of the box also in IE11

justinfagnani:
Thanks for this PR @bgotink! It's nice how small of a change this is :)

I do find it a little strange to keep the placeholder marker in the template such that it survives cloning. I think it'd be a bit nicer to replace the empty text node with a comment node. Then you don't have to clean up cloned instances.

So here: https://github.com/PolymerLabs/lit-html/pull/202/files#diff-7548be97e171131880082c935a962f68R263 instead of using the marker text, insert a comment:

```ts
        if (nodeValue.indexOf(marker) > -1) {
          const parent = node.parentNode!;
          const strings = nodeValue.split(markerRegex);

          // We have a part for each match found
          partIndex += strings.length - 1;

          // Generate a new text node for each literal section
          // These nodes are also used as the markers for node parts
          for (const s of strings) {
            // IE doesn't clone empty text nodes, so use comments instead
            parent.insertBefore(s === ''
                ? document.createComment(marker)
                : document.createTextNode(s), node);
          }
          nodesToRemove.push(node);
        } else {
```

and in the other places:

```ts
parent.insertBefore(document.createComment(marker), node);
```

bgotink:
Note that all of these comments are visible in the innerHTML of the elements, so this requires changes to 61 tests:

```
expected '<div><!--{{lit-8277999649123744}}-->foo<!--{{lit-8277999649123744}}--></div>' to equal '<div>foo</div>'
```

Using comment nodes with a constant value (e.g. empty comments) instead of the marker should improve testability by removing the randomness from the runtime DOM. It's either this, or export the marker so the tests can use it, or update all of the innerHTML tests to use a regular expression rather than raw string comparison.
I'd rather not write 61 regular expressions and I don't like the idea of exporting the marker because that means outside code can start to fiddle with the workings of lit-html e.g. by adding the marker to templates. As such I propose using empty comment nodes here.

I also get one other error in some tests that I'll continue to investigate tomorrow:

```
Error: Cannot read property 'previousSibling' of null
        NodePart._setText at /components/lit-html/lit-html.js:394
```

---

While I agree that using a comment node is awesome in that it only impacts the setup of the template and has no cost at clone time, I admit I dislike that the resulting DOM is now full of comment nodes (visible in the debugger and in `element.innerHTML`) that aren't in the string passed to the `html` function.  
I don't see any way around this without impact at clone time, so&hellip; ¯\\\_(ツ)\_/¯

bgotink:
Rebased onto master and moved to empty comment nodes instead of text nodes with a marker in them.

This PR now brings the number of test failures on IE down to 1 🎉

@justinfagnani I couldn't use the for-of loop you proposed because that adds one extra NodePart, so I kept the index iteration as is.

justinfagnani:
I think the changes in tests are fine for now. Maybe we can make that nicer in the future. Awesome work!

Issue #192, opened 2017-11-18
Rendering file templates


adaliszk:
I would like to use a separate file for rendering a template: Often the actual HTML+CSS part is way bigger than the actual JavaScript. Separating them makes them more readable and manageable for developers and for tools. A lot of time my template is made from multiple parts which often shared and used in multiple template, it would be nice to able to load a view file, pass the variables and use it as a template and pass it to rendering. Within this views you do need tools to load other views so you can build up your template just as any server side rendered template system.

I know loading an other resource has drawbacks but with HTTP2 you should split your code so you can cache them individually and often when you developing you markuping and styling and not writing functionalities or the other way around.

LarsDenBakker:
You can already write your template as a separate file:

// my-style.js
```
import { html } from './node_modules/lit-html/lit-html.js'
const style = html`<style>
  .msg {
    font-weight: bold;
    font-size: 20px;
  }
</style>`;
export default style;
```

// my-header-template.js
```
import { html } from './node_modules/lit-html/lit-html.js'
const template = html`<div>My Header<div>`;
export default template;
```

// my-msg-template.js
```
import { html } from './node_modules/lit-html/lit-html.js'
const template = msg => html`<div class="msg">${msg}<div>`;
export default template;
```

Then, to use it:

// my-renderer.js
```
import styleTemplate from './my-style.js';
import headerTemplate from './my-header-template.js';
import msgTemplate from './my-msg-template.js';
import { render, html } from './node_modules/lit-html/lit-html.js'

const template = html`
  ${styleTemplate}
  ${headerTemplate}
  ${msgTemplate('Hello world')}
`;
```
Or inside a web component:

// my-component.js
```
import styleTemplate from './my-style.js';
import headerTemplate from './my-header-template.js';
import msgTemplate from './my-msg-template.js';
import { render, html } from './node_modules/lit-html/lit-html.js'

class MyComponent extends SomeLitElement {

  renderCallback() {
    return  html`
      ${styleTemplate}
      ${headerTemplate}
      ${msgTemplate('Hello world')}
    `;
  }

}

customElements.define('my-component', MyComponent);
```

adaliszk:
That's not what I want to do. What I want to achieve is the same with any other server side templating engine where you assign values and you only writing into the HTML the expressions:

// layout.tpl
```html
${file("common/header.tpl")}
<main>
    ${content}
</main>
```

// common/header.tpl
```html
<header>
    <span class="title">${title}</span>
[...]

PR #203, opened 2017-11-27
Add #192 - File rendering directive

adaliszk:
So by adding this directive it opens the possibility to render a file as a template string, the general usage:

```javascript
// somewhere in script
render(html`${file('file.tpl', 'Loading...')}`, document.body);
```

```html
<!-- file.tpl -->
<p>Rendered content from file</p>
```

### DIFF
--- a/demo/render-file.html
+++ b/demo/render-file.html
@@ -1,0 +1,9 @@
+<!doctype html>
+<html>
+<head>
+  <script type="module" src="./render-file.js"></script>
+</head>
+<body>
+<lit-render-file></lit-render-file>
+</body>
+</html>
--- a/demo/render-file.js
+++ b/demo/render-file.js
@@ -1,0 +1,13 @@
+
+import {html, render} from '../lit-html.js';
+import {file} from '../lib/file.js';
+
+export class LitRenderFile extends HTMLElement {
+  constructor() {
+    super();
+    this.attachShadow({mode: 'open'});
+    render(html`${file('render-file.tpl', 'Loading...')}`, this.shadowRoot || this);
+  }
+}
+
+customElements.define('lit-render-file', LitRenderFile);
--- a/demo/render-file.tpl
+++ b/demo/render-file.tpl
@@ -1,0 +1,1 @@
+<h1>Hello file template!</h1>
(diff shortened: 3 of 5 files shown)


adaliszk:
I signed it!

adaliszk:
I did it with my other account as well!

RayBenefield:
@adaliszk does this give the file access to the context at all? And is it possible to add/override scope? For example a partial that has a single element and I want to pass it all of the item's attributes for access?

adaliszk:
@RayBenefield no, this is just the initial extension which allows you to load file content instead writting it in your javascript. For giving access which variable are passed I need to check the code more, usually you are able to access everything whatever the context of the render function. So in your example you just use your template string as a template file and do exactly the same what you would do with string, but you can put your element into a separate file with your attribute assigns and you can reuse that template within multiple ones if the context has the necessary variables, otherwise the js will throw an error.

chicoxyzzy:
In current PR if fetching fails, content doesn't ever changes. IMHO it's better to keep file loading not in directive.

adaliszk:
That's true, I didn't wanted to add a fail case because you would see the error in the console anyways. Thanks for the feedback, I will put a catch and I will print and error message about that. I wanted to use directive because I want to nest file templates and I don't want to build a huge resolver around the library just to render stuff. I guess I could make it separate from the lit-html itself as a separate package :/

chicoxyzzy:
Alternatively your directive could accept `onSucces` and `onError` functions as arguments or return Promise with template file text or an error.

justinfagnani:
@adaliszk if I have this right, this is essentially:

```ts
export const file = (file: string, defaultContent: any) => until(unsafe(fetch(file).then(r => r.text())), defaultContent);
```

_if_ `unsafe` could accept a Promise?

I'm now sure about putting this into the lit-html repo itself. I'd like to limit and make explicit unsafe inclusion of HTML. If the URL referenced has user-generated content, this is a potential XSS vulnerability.

Also, I think people want a declarative format for lit-html templates, and they might confuse this directive with that.

justinfagnani:
@adaliszk I don't think this is something we'd like to ship with lit-html itself, considering the security implications. But the nice thing about directives is that you can publish them independently! I'd support you doing that and we can add a section to the docs that list third party directives.

Thanks!

adaliszk:
@justinfagnani yeah, I will do that because this is a nice feature and it has to have multiple iterations until it's correct for every viewpoints.

PR #204, opened 2017-11-28
Alternative Flag attributes implementation

shutdown256:
Alternative implementation to #197 for lit-extended attributes with following differences:
1.  it treats null and undefined as false - this behavior is consistent with polymer: unresolved/uninitialized expressions will not set (e.g. hidden)
2.  no special syntax other than $= required. In rare case that you actually want to set attribute as "true" or "false", just use a ? "true" : "false"

```html
<div hidden$="${true}"/>  => <div hidden=""/>
<div hidden$="${false}"/> => <div />
<div hidden$="${null}"/> => <div />
<div hidden$="${undefined}"/> => <div />
<div hidden$="${!(undefined)}"/> => <div hidden="" />
<div hidden$="${'true'}"/>  => <div hidden="true"/>
<div hidden$="${'false'}"/>  => <div hidden="false"/>
```

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -58,7 +58,7 @@ export const extendedPartCallback =
             }
             if (templatePart.name!.endsWith('$')) {
               const name = templatePart.name!.slice(0, -1);
-              return new AttributePart(
+              return new ExtendedAttributePart(
                   instance, node as Element, name, templatePart.strings!);
             }
             return new PropertyPart(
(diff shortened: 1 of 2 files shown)


justinfagnani:
I think it might be ok to do the boolean attribute handling in lit-extended first. I'm unsure about the set of values that would remove an attribute. I think that just `undefined` is a better signal. I'd still like to be aligned with the template instantiation proposal, but that's not figured out at this point.

justinfagnani:
Hey @shutdown256  - I've been discussing this with some other Polymer people, and I think that we want to go with a suffix in lit-extended like in #197, and a directive for plain lit. If lit-extended is a bit of Polymer-flavor for lit-html, then it makes sense to use the Polymer convention to be more familiar to devs.

Another nice thing about the suffix/directive is that it's usually the _attribute_ that's boolean and needs to be removed, information which might not be embedded in the _value_.

Thanks for the PR!

Issue #210, opened 2017-12-01
IE 11 support


LarsDenBakker:
I'd really like to start using lit-html. I don't mind it being rough around the edges, and breaking changes coming in the future, but IE11 support is unfortunately a deal breaker for my company. 

I'd like to contribute, but I see there have already been a few attempts at becoming IE11 compatible. Are we close to a solution? Is there an overview of the remaining issues?

bgotink:
I'm not sure the list is complete, but:

- There's an issue with the cloning process (see #201 and proposed fix #202)
- There's an issue with `<tr>`, `<th>` or `<td>` elements as root element in the template because of an issue in the template polyfill (see webcomponents/template#3)
- The tests for lit-extended and some directives don't import the necessary polyfills for IE yet, so the tests fail
- A couple of tests fail because IE sorts attributes differently from other browsers. This is a bug in the tests, not the library.

justinfagnani:
With #202 and #211 we're really close.

I think we might be able to fix the table problem because we're in control of generating the template contents - unlike the template polyfill we get to do things before the parser.

Before that I would be ok with skipping the table tests on IE and making sure we get tests passing to lock in our progress and make sure we don't regress.

LarsDenBakker:
Is there anything that can be done here to help put?

toddoh:
Any updates on this?

smalluban:
There is yet another bug - it is not possible to set `style` attribute. IE11 prevents setting style attribute, which is not proper CSS text (list of CSS properties) and clears attribute placeholder, so template does not know that there was any expression.

```javascript
const template = value => html`<div style="{value}">Sample text</div>`;
render(template('color: red'), ...);
```

Above example in IE11 produces black text.

The second thing is npm package. It is not properly distributed. The compiled version of the project contains code, which is not supported in IE11, for example, arrow functions.

seaneking:
@smalluban just re npm package, pretty sure it's only shipped as an ES module, the assumption being that you'd bundle and transpile it youself for legacy browsers

smalluban:
@seaneking I think it is not a good idea. `module` or `jsnext:main` should be used for esm version of the library. Parcel and common Webpack configuration of babel usually omits `node_modules` and does not transpile external packages. With current solution, users have to change this only for one dependency.

seaneking:
[...]

PR #212, opened 2017-12-03
Disable table test in IE

bgotink:
This test fails due to a bug in the template polyfill. Disabling this test allows us to reach 100% passing tests on IE, which means IE can be added to the CI build.

As mentioned in #210
Together with #202 this change results in a passing build for IE.

### DIFF
--- a/src/test/lit-html_test.ts
+++ b/src/test/lit-html_test.ts
@@ -225,7 +225,12 @@ suite('lit-html', () => {
         assert.equal(container.innerHTML, '<div>foo bar</div>');
       });
 
-      test('renders nested templates within table content', () => {
+      const testSkipForTemplatePolyfill =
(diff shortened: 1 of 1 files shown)


src/test/lit-html_test.ts
```diff
@@ -225,7 +225,13 @@ suite('lit-html', () => {
         assert.equal(container.innerHTML, '<div>foo bar</div>');
       });
 
-      test('renders nested templates within table content', () => {
+      const testSkipIE =
```
justinfagnani: Rather than a UA-sniff, I think it'd be good to check for the template polyfill. HTMLTemplateElement.decorate should be a good property to check.

bgotink: That is indeed better than a UA sniff, thanks! I've updated the PR.

PR #213, opened 2017-12-05
feat: SpreadPart for multiple attributes/properties

daKmoR:
Allows for the following syntax:

```js
let options = {
  'id$': 'my-id',
  'active': true
};

render(html`<my-element my-option="some" ...=${options}></my-element>`, target);
```

Result:
```html
<my-element my-option="some" id="my-id"></my-element>
```
```js
document.querySelector('my-element').active; // true
```

If there is an interest for it I can provide tests, demos and docu.

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -42,6 +42,23 @@ export {html} from '../lit-html.js';
  *
  *     html`<button on-click=${(e)=> this.onClickHandler(e)}>Buy Now</button>`
  *
+ * To set multiple values you can use the spread attribute with `...=${options}`:
+ *   Note: provided spread data will overwrite currently set values
+ *
+ * Example:
+ *
+ *     const options = {
+ *       'id$': 'buy',
+ *       'data-action': 'add-to-cart',
+ *       'active': true
+ *     };
+ *     html`<button class="fanzy" id="my-button" ...=${options}>Buy Now</button>`
+ *
+ * Result:
+ *
+ *     <button class="fanzy" id="buy" data-action="add-to-cart">Buy Now</button>
+ *     document.querySelector('button').active; // true
+ *
  */
 export function render(
     result: TemplateResult, container: Element|DocumentFragment) {
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -61,6 +61,10 @@ export const extendedPartCallback =
               return new AttributePart(
                   instance, node as Element, name, templatePart.strings!);
             }
+            if (templatePart.name!.endsWith('...')) {
+              return new SpreadPart(
+                  instance, node as Element, templatePart.rawName as string, templatePart.strings!);
+            }
             return new PropertyPart(
                 instance,
                 node as Element,
@@ -69,6 +73,21 @@ export const extendedPartCallback =
           }
           return defaultPartCallback(instance, templatePart, node);
         };
+
+export class SpreadPart extends AttributePart {
+  setValue(values: any[], startIndex: number): void {
+    const properties = values[startIndex];
+    Object.keys(properties).forEach((property) => {
+      const value = properties[property];
+      if (property.endsWith('$')) {
+        const attribute = property!.slice(0, -1);
+        this.element.setAttribute(attribute, value);
+      } else {
+        (this.element as any)[property] = value;
+      }
+    });
+  }
+}
 
 export class PropertyPart extends AttributePart {
   setValue(values: any[], startIndex: number): void {
--- a/src/test/lib/lit-extended_test.ts
+++ b/src/test/lib/lit-extended_test.ts
@@ -191,6 +191,21 @@ suite('lit-extended', () => {
       assert.equal(target, undefined);
     });
 
+    test('spread part sets properties and attributes', () => {
+      const options = { foo: '123', 'bar$': '456' };
+      render(html`<div ...=${options}></div>`, container);
+      const div = container.firstChild!;
+      assert.equal((div as any).foo, 123);
+      assert.equal((div as any).getAttribute('bar'), 456);
+    });
+
+    test('spread part overrides current values', () => {
+      const options = { foo: '123', 'bar$': '456' };
+      render(html`<div foo="100" bar$="200" ...=${options}></div>`, container);
+      const div = container.firstChild!;
+      assert.equal((div as any).foo, 123);
+      assert.equal((div as any).getAttribute('bar'), 456);
+    });
 
   });
 });


jridgewell:
`...` is a valid attribute name, however unlikely. Why not allow:

```js
html`
  <my-element my-option="some" ${options}></my-element>
`;

**Comment by daKmoR** [comment]
hmm I would not know what to set options then? a string? `option = 'id="my-id"`?
the "..." spread "attribute" in this case only processes "flat" key, value objects... which seems easier to understand for me.

but then again I was in need of a solution and found this so I'm probably biased :p

also it feels more "lit like" to me... but sure `<my-element my-option="some" ${options}></my-element>` could work just fine... however I'm not really sure how it would be implemented...

**Comment by shutdown256** [comment]
I like it and the spread operator syntax makes sense to me. I don't believe we need 100% backwards compatibility with HTML5 spec in lit-extended - it can be little opiniated.

**Comment by justinfagnani** [comment]
> ... is a valid attribute name, however unlikely. Why not allow:
>
>```js
>html`
>  <my-element my-option="some" ${options}></my-element>
>`;
>```

I'd need to check that this parses correctly and works with our regex scanning.

This:
```js
html`<my-element ${options}>`
```

would generate:
```html
<my-element {{}}>
```

So we could look for attributes named literally `{{}}`, and support only one spread. I think we could get it to work, but the criticism that `...` is a valid attribute name also applies to `{{}}`.

I personally _think_ (but could be convinced otherwise) that `<x-foo ...=${attrs}>` is better ergonomics because the syntax hints what's going on since it _looks_ like JS spread syntax.

jridgewell:
> named literally {{}}

It'd be `{{lit-43247389512}}`, no?

danizord:
>I personally think (but could be convinced otherwise) that <x-foo ...=${attrs}> is better ergonomics because the syntax hints what's going on since it looks like JS spread syntax.

Does `<x-foo ...${attrs}>` work? If so, it looks even more closer to the JS syntax :)

justinfagnani:
@jridgewell 

> It'd be `{{lit-43247389512}}`, no?

Yeah, I was simplifying for the example, and then read my simplification and argued that it's a valid attribute name :) So it's unlikely to collide, and it's still a  known name though, so we can look for it. Interesting...

@danizord 

> Does `<x-foo ...${attrs}>` work? If so, it looks even more closer to the JS syntax :)

I think that could be made to work too, as the attribute name would be `...{{lit-43247389512}}`, which we could look for.

I feel a _little_ weird about those two options though, because they look so little like HTML's attribute syntax. My hunch is that documenting `...` as a special attribute is easier to explain, and probably easier for a first time reader to guess at what's going on.

There's also a question of whether this should allow setting attributes and properties in one object, or whether there should be two special attribute names, like `...` and `...$`. `<x-foo ...$=${attrs}>` is a lot of symbols, but logically it makes sense to me. But then, what about event handlers? This PR doesn't special-case `on-`-prefixed property names, but if it does that for the `$` suffix, it probably should.

I do like that all this sugar is limited to lit-extended so that someone else could make different decisions if they wanted to :) Thanks for kicking this off @daKmoR!

bgotink:
I'd expect `<x-foo ${obj} ${obj2}>` or `<x-foo ...${obj} ...${obj2}>` to work. We can do `{ ...obj, ...obj2 }` in javascript after all. As a reader of that code I don't interpret those spreads as attributes despite the HTML parser disagreeing with me.

The `<x-foo ...=${obj}>` syntax is different as it's clearly an attribute. HTML doesn't support duplicate attributes, so this syntax more clearly shows that there can only be one spread per element imo.

src/lib/lit-extended.ts
```diff
@@ -70,6 +74,21 @@ export const extendedPartCallback =
...
 
+export class SpreadPart extends AttributePart {
+  setValue(values: any[], startIndex: number): void {
+    const properties = values[startIndex];
+    Object.keys(properties).forEach((property) => {
+      const value = properties[property];
+      if (property.endsWith('$')) {
+        const attribute = property!.slice(0, -1);
+        this.element.setAttribute(attribute, value);
+      } else {
```
justinfagnani: This should handle `on-`prefixed names (and `?` suffixed names in the future).

daKmoR: how would we like to do that?
"duplicate code" for setAttribute and assign an property is still ok I guess...

but for event "handling" we probably should use the existing EventPart right? howto best do that?

daKmoR:
added initial tests and "docu"... for "real" demos we currently don't something... how do we want to do that? multiple html files in `/demos/*`?

I also did not generate the docu as that changed a lot... generate docu later with a separate pull request?

bgotink:
I wonder what the expected behaviour is when a property is removed from the object, e.g.

```js
const template = obj => html`<x-foo ...="${obj}"></x-foo>`

render(template({bar$: 'baz', qux: true}), container);

// later on

render(template({}), container);
```

Should this remove the `bar` attribute and the `qux` property or should those be left alone?

daKmoR:
very good question :)

js spread would remove it - if we say render is sort of a assignment
```js
let obj = { bar: 1 };
let a = { ...obj };
obj = {};
a = { ...obj }; // {}
```

current result:

```js
let obj = {bar$: '1'};
render(html`<x-foo ...="${obj}"></x-foo>`, container); //<x-foo bar="1">
obj = {baz$: 2};
render(html`<x-foo ...="${obj}"></x-foo>`, container); //<x-foo bar="1" baz="2">
```
bar value didn't "change" so I guess that is why lit-html simple just does not touch it

2 possible "workaround" I would see

1) force new render
```js
render(html`<x-foo></x-foo>`, container); // reset "element" - is there a better way?
render(html`<x-foo ...="${obj}"></x-foo>`, container); //<x-foo baz="2">
```

2) be very deliberate with your spread obj e.g. do not remove elements from it - "only" mark them as deleted (for example with using "?" and setting it to undefined) [if we go ahead with ?]
```js
obj = {'bar?': undefined, baz$: 2};
render(html`<x-foo ...="${obj}"></x-foo>`, container); //<x-foo baz="2">
```

daKmoR:
Is there anything still needed to make this happen?

I think it adds enough value as an initial implementation with some limitations - maybe just add those + the workarounds to the documentation.

If at one point there is a need for managing object data with the removal of properties and attributes we can implement a more complex solution.

src/lib/lit-extended.ts
bgotink: Maybe it'd be useful to actually create `Part` instances for every property of the object. If we'd use `this.instance._partCallback` for that, it'd support event handling, property setting and attribute setting out of the box, as well as any future parts that are added like the `?` suffix for flags. In case anyone extends lit-extended with other custom parts, those would be supported automatically as well.

Although it would look kinda weird to create parts with hardcoded `['', '']` as strings. Calling `setValue` like this also looks weird:

```ts
for (const property in properties) {
  // create subpart if needed

  this.subParts[property].setValue([properties[property]], 0);
}
```

but I think the benefit of not duplicating code and supporting any future and external `Part`s is nice to have.

justinfagnani:
@daKmoR I'll review this again asap after I get the next release out.

src/lib/lit-extended.ts
justinfagnani: I think @bgotink is right about the approach. It'll reuse the logic in the part callback and naturally gain any new capabilities.

I'd rebase on to the shadycss branch to pick up changes to how part callbacks work.

AStaroverov:
Sorry, but i dont understand, why u dont want use naming for spreadable object like for lit template?
```
var obj = {
  bar: '123', // attribute
 '.baz': [], // property
 '?disabled': true, // boolean attribute
 '@click': console.log // listener
}
```

justinfagnani:
The key for this feature is to implement it without too much code weight. I think we can revisit for 1.1

daniel-nagy:
Landed here after trying to figure out how to pass props through to children.

FaBeyyy:
Would love to see this, this is something that has been bothering me for months.

straversi:
This PR is against an old version of the API, so we've made a new issue for tracking this feature request https://github.com/Polymer/lit-html/issues/923.

We like the convenience of spread, but want to weight that against potential bloat to the package. We'll consider these points when deciding whether to merge this feature.

arpitBhalla:
> Landed here after trying to figure out how to pass props through to children.

I think you can use slots

PR #225, opened 2017-12-11
Better type for directives

### LINKED ISSUE #224: Type errors in `repeat.ts`
Don't know, I'm not really a TypeScripter, but Code squiggles

image

kaste:
Fixes #224 

The patch is not as small as I intended bc I had to propagate the changes to the available directives.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -353,7 +353,7 @@ export const getValue = (part: Part, value: any) => {
 
 export interface DirectiveFn<P=Part> {
   (part: P): void;
-  __litDirective?: boolean;
+  __litDirective?: true;
 }
 
 export const directive = <P=Part>(f: DirectiveFn<P>): DirectiveFn<P> => {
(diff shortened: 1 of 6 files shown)


kaste:
I still think something like 

https://github.com/PolymerLabs/lit-html/blob/5742a9e5e6ae74ee91b0421bca92d90f3e2b0601/src/lit-html.ts#L352-L354

should warn, bc we said directives are functions to `void`. (It should also warn if we return something from a directive.)

src/lit-html.ts
```diff
@@ -356,10 +356,13 @@ export const getValue = (part: Part, value: any) => {
   return value === null ? undefined : value;
 };
 
-export type DirectiveFn<P extends Part = Part> = (part: P) => any;
+export interface DirectiveFn<P=Part> {
+  (part: P): void;
+  __litDirective?: boolean;
```
justinfagnani: You should be able to do: `__litDirective?: true`;

kaste:
Rebased and addressed feedback

justinfagnani:
Thanks @kaste! Sorry this took so long to merge!

PR #226, opened 2017-12-11
Fix `envCachesTemplates` for UglifyJS

### LINKED ISSUE #223: UglifyJS overmangles `envCachesTemplates`
Okay, in

```
const envCachesTemplates =
    ((t) => t() === t())(() => ((s) => s) ``)
```

Out: 

```
const envCachesTemplates=(e=>e()==e())(()=>e=>e``);
```

which will evaluate to `false`. 

image

kaste:
Fixes #223 

image

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -19,8 +19,11 @@
  * TODO(justinfagnani): Run tests compiled to ES5 with both Babel and
  * TypeScript to verify correctness.
  */
-const envCachesTemplates =
-    ((t: any) => t() === t())(() => ((s: TemplateStringsArray) => s) ``);
+const envCachesTemplates = (() => {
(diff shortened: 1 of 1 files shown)


jridgewell:
Ping @justinfagnani. It's a bug fix, and it actually makes `envCachesTemplates` understandable.

stramel:
@jridgewell did you happen to see #233? I'm just curious if that may be a better fix or not?

jridgewell:
Oo, that's even better.

justinfagnani:
Yeah, I think we can remove this now. It required that lit-html and calling code be compiled with the same compiler anyway, which was kind of a Google-internal-only setup.

Thanks for the fix though @kaste!

PR #227, opened 2017-12-11
Better types for render function

kaste:
Avoid any casts within `render`. Dunno, I think it is less confusing.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -72,6 +72,10 @@ export class TemplateResult {
   }
 }
 
+export interface RenderContainer extends Element {
+  __templateInstance?: TemplateInstance
+}
+
 /**
  * Renders a template to a container.
  *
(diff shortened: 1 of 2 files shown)


src/lit-html.ts
```diff
@@ -72,6 +72,10 @@ export class TemplateResult {
   }
 }
 
+export interface RenderContainer extends Element {
```
justinfagnani: This should be `extends ParentNode` to also pickup `DocumentFragment`.

kaste: `extends ParentNode` does not work. `ParentNode` has no `firstChild` nor `appendChild`

justinfagnani:
On a little more thinking, we don't want to expose this type in the signature of `render()`, because TypeScript users of the library would then have to cast their element to `RenderContainer` to call `render()`.

It does make sense to me as an internal type to reduce the casts to `any` and document the expectations, so I opened #330.

Thanks for the nudge in this direction - it's a good idea, and we'll try to eliminate the other casts where we can :)

Issue #219, opened 2017-12-07
Consider a tag factory to apply options to templates


justinfagnani:
As discussed in #179

```js
// tag is probably a bad name
import {tag} from 'lit-html/lit-html.js';

const html = tag({stripWhitespace: true});
render(html`
  <div>
    <h1>whitespace is stripped</h1>
  </div>`,
  document.body);
```

Tags will need their own template cache if they change how templates are prepared.

justinfagnani:
We have a tag factory in 2.0 with the static module. That pattern can be applied to other rendering variations. We should track whitespace separately.

Issue #140, opened 2017-10-19
Consider factoring template preparation into separate methods.


justinfagnani:
The Template constructor is quite long now...

Issue #136, opened 2017-10-16
lit-html is not working with shadyDOM


AndreasGalster:
I'm sure you're aware of this already, just putting it on paper so we can track the future dev :). My lit-html component doesn't render styles properly if I'm turning shadyDOM on. 

# Expected result
All styles, custom properties and @apply mixins work as expected

# Actual result
None of them work properly

daKmoR:
I guess that is not really a problem of shadyDom or shadyCSS as something like this will probably need to be done by the "app" or element itself.

litHtml does not "care" if it renders into a shadowRoot or some innerHtml.

You can see a simple demo here that uses shadowdom and shadycss.
https://github.com/daKmoR/lit-html-demos/blob/master/demo/wc02.html

jamesisaac:
@daKmoR That example works to some degree, but from what I can tell, you're applying `ShadyCSS.prepareTemplate` to the throwaway `this.shadyTemplate`, which isn't used anywhere further. 
 So what this loses out on, is that the ShadyCSS classes (`style-scoped` and the name of the custom element) are not actually applied to all the elements in the subtree.  So, shadow DOM style encapsulation doesn't work, and the scoped styles ShadyCSS applied in the <head> don't do anything?

I had a go at applying `prepareTemplate` to lit-html's internal template, but I realise the fundamental flaw there is that lit-html is remembering the exact position of each TemplatePart, which will be broken if class names are inserted.  Seeing as when a TemplateResult is constructed, it doesn't really have insight into what element it's a part of (and therefore what classnames to add), I'm not sure a way round this.

justinfagnani:
@sorvell is working on this

tjmonsi:
I did something that sorts of hack this to make use of shadyCSS.

```
// in my connectedCallback() 

      if (window.ShadyCSS) {
        var template = document.createElement('template');
        render(this.template(), template.content);
        window.ShadyCSS.prepareTemplate(template, this.constructor.is);
        window.ShadyCSS.styleElement(/** @type {!HTMLElement} */(this));
      }

// and i override this component method
    template () {
      return html``;
    }
```

You guys can check my hacked polymer element-mixin where I removed the dom-module dependency and templating here: https://github.com/tjmonsi/tjmonsi-site/blob/master/src/core/mixins/element-mixin.js

justinfagnani:
This was initially fixed a while back in #249. There are newer bugs and PRs for improved ShadyCSS integration.

PR #228, opened 2017-12-13
Proposal: Make Template extensible

bgotink:
This PR combines #140 and #219 to make the template tag functions more extensible.

Note: This is more a request for comments on the idea of the code than an actual pull request.

There are two parts to this proposal:

- It splits the Template constructor into multiple methods which can be overriden for subclasses. SVG templates are implemented as a subclass instead of `if`s.
- It also exposes a `tag` function which creates tag functions, export `html` and `svg` as created by the `tag` function.

Using this set-up we can easily create a tag function that strips whitespace *:

```ts
import {TemplateConstructor, SvgTemplate, Template, tag} from './lit-html.js';

const nonWhitespace = /[^\s]/;

function StripWhitespaceTemplate(BaseTemplate: TemplateConstructor): TemplateConstructor {
  return class extends BaseTemplate {
    visitText(text: Text, nodesToRemove: Node[]) {
      // Strip whitespace-only nodes, only between elements, or at the
      // beginning or end of elements.
      const previousSibling = text.previousSibling;
      const nextSibling = text.nextSibling;
      if ((previousSibling === null ||
           previousSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
          (nextSibling === null ||
           nextSibling.nodeType === 1 /* Node.ELEMENT_NODE */) &&
          !nonWhitespace.test(text.nodeValue)) {
        nodesToRemove.push(text.node);
        this.currentNode = this.previousNode;
        this.index--;
      } else {
        super.visitText(text, nodesToRemove);
      }
    }
  };
}

export const strippedHtml = tag({Template: StripWhitespaceTemplate(Template)});

export const strippedSvg = tag({Template: StripWhitespaceTemplate(SvgTemplate(Template))});
```

Or we can use this to add support for ShadyDOM (#136) *

```ts
import {TemplateConstructor, Template, tag, html} from './lit-html.js';

function ShadyTemplate(tagName: string, BaseTemplate: TemplateConstructor): TemplateConstructor {
  return class extends BaseTemplate {
    getTemplate(strings: TemplateStringsArray): HTMLTemplateElement {
      const template = super.getTemplate(strings);
      ShadyCSS.prepareTemplate(template, tagName);
      return template;
    }
  }
}

export const createHtml = typeof ShadyCSS !== 'object'
        ? () => html 
        : (tagName: string) => tag({Template: ShadyTemplate(tagName, Template)});

export class MyElement extends HTMLElement {
  protected html: typeof html;

  constructor() {
    super();

    this.html = createHtml(this.localName);
  }

  protected template(): TemplateResult {
    return this.html`
      <style>.error { color: red; }</style>
      <h1 class="error">Override this function</h1>
    `;
  };

  public render(): void {
    render(this.template(), this.shadowRoot || this.attachShadow({mode: 'open'}));
    if (typeof ShadyCSS === 'object') {
      ShadyCSS.styleElement(this);
    }
  }
}
```

We could also move the `svg` and `SvgTemplate` functions to a separate file (`lit-svg.js`?), to keep `lit-html.js` as lean and clean as possible.

\* These code samples are quickly written in the GitHub issue textarea, don't expect them to work or even compile.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -22,37 +22,32 @@
 const envCachesTemplates =
     ((t: any) => t() === t())(() => ((s: TemplateStringsArray) => s) ``);
 
-// The first argument to JS template tags retain identity across multiple
-// calls to a tag for the same literal, so we can cache work done per literal
-// in a Map.
-const templates = new Map<TemplateStringsArray|string, Template>();
-const svgTemplates = new Map<TemplateStringsArray|string, Template>();
-
 /**
- * Interprets a template literal as an HTML template that can efficiently
- * render to and update a container.
+ * Options that can be passed on to the tag factory.
  */
-export const html = (strings: TemplateStringsArray, ...values: any[]) =>
-    litTag(strings, values, templates, false);
+export interface TagOptions { Template: TemplateConstructor; }
 
-/**
- * Interprets a template literal as an SVG template that can efficiently
- * render to and update a container.
- */
-export const svg = (strings: TemplateStringsArray, ...values: any[]) =>
-    litTag(strings, values, svgTemplates, true);
+export const tag = (options: TagOptions) => {
+  // The first argument to JS template tags retain identity across multiple
+  // calls to a tag for the same literal, so we can cache work done per
+  // literal in a Map.
+  const templates = new Map<TemplateStringsArray|string, Template>();
+
+  return (strings: TemplateStringsArray, ...values: any[]) =>
+             litTag(strings, values, templates, options);
+};
 
 function litTag(
     strings: TemplateStringsArray,
     values: any[],
     templates: Map<TemplateStringsArray|string, Template>,
-    isSvg: boolean): TemplateResult {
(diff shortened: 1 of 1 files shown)


justinfagnani:
Hey @bgotink 

I have some API changes around the and functions and `render` that are very similar in spirit to this, and needed for ShadyCSS integration and for being able to compose templates from different flavors like plain lit, lit-extended and lit-html-brackets.

It moves responsibility for choosing the partCallback to the tag, so different flavors can be passed to the same render function and still work as intended at the template definition site. It also makes TemplateResult responsible for generating the html string and the template element so these can be customized, as in the SVG case. Finally, render is able to take a hook for caching and generating the template so that ShadyCSS can modify it and cache per scope.

I'll add you to that PR, and if it doesn't cover anything you needed here, let me know. I'd also be great to get feedback from updating lit-html-brackets. I'll close this PR for now as I think it'll be fully covered. Thanks!

bgotink:
Thanks for the feedback!

PR #229, opened 2017-12-13
Switch to using text nodes with a marker instead of empty comment nodes

### LINKED ISSUE #214: Comment nodes inside textContent
After https://github.com/PolymerLabs/lit-html/pull/202, we can get comment nodes inside our text content, which isn't great:

```js
render(html`<style>${''}</style>`, container);
container.innerHTML; // => '<style><!----><!----></style>'
```

Test case:
```js
      test('renders an attribute after empty style node binding', () => {
        render(html`
            <style>${''}</style>
            <a href="/buy/${'foo'}"></a>
          `, container);
        assert.equal(container.innerHTML, `
            <style></style>
            <a href="/buy/foo"></a>
          `);
      });
```

bgotink:
Comment nodes are problematic when inserted into textContent, e.g. a `<style>`
element. This commit changes to using a text node with a marker content rather
than an empty comment. The markers are removed at clone time.

All of this is implemented using a platform check to ensure that we only do
this on platforms that don't clone empty text nodes (IE).

Fixes #214

Note: not yet tested on IE because I can't access any Windows machine/vm atm.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -644,18 +644,6 @@ export class TemplateInstance {
     const parts = this.template.parts;
 
     if (parts.length > 0) {
-      if (envDoesntCloneEmptyTextNodes) {
-        // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be
-        // null
-        const textWalker = document.createTreeWalker(
-            fragment, 4 /* NodeFilter.SHOW_TEXT */, null as any, false);
-        while (textWalker.nextNode()) {
-          if (textWalker.currentNode.nodeValue === marker) {
-            textWalker.currentNode.nodeValue = '';
-          }
-        }
-      }
-
       // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be
       // null
       const walker = document.createTreeWalker(
@@ -667,13 +655,27 @@ export class TemplateInstance {
           false);
 
       let index = -1;
+      let currentNode: Node;
       for (let i = 0; i < parts.length; i++) {
         const part = parts[i];
         while (index < part.index) {
           index++;
-          walker.nextNode();
+          currentNode = walker.nextNode();
+
+          if (envDoesntCloneEmptyTextNodes &&
+              currentNode.nodeType === 3 /* Node.TEXT_NODE */ &&
+              currentNode.nodeValue === marker) {
+            currentNode.nodeValue = '';
+          }
         }
+
         this._parts.push(this._partCallback(this, part, walker.currentNode));
+      }
+
+      if (envDoesntCloneEmptyTextNodes && (currentNode = walker.nextNode()) &&
+          currentNode.nodeType === 3 /* Node.TEXT_NODE */ &&
+          currentNode.nodeValue === marker) {
+        currentNode.nodeValue = '';
       }
     }
     return fragment;
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -151,6 +151,13 @@ function findTagClose(str: string): number {
   return open > -1 ? str.length : close;
 }
 
+const envDoesntCloneEmptyTextNodes = (() => {
+  const container = document.createElement('template');
+  container.content.appendChild(document.createTextNode(''));
+
+  return !document.importNode(container.content, true).childNodes.length;
+})();
+
 /**
  * A placeholder for a dynamic expression in an HTML template.
  *
@@ -259,17 +266,24 @@ export class Template {
         // We have a part for each match found
         partIndex += lastIndex;
 
-        // We keep this current node, but reset its content to the last
-        // literal part. We insert new literal nodes before this so that the
-        // tree walker keeps its position correctly.
-        node.textContent = strings[lastIndex];
-
         // Generate a new text node for each literal section
         // These nodes are also used as the markers for node parts
         for (let i = 0; i < lastIndex; i++) {
-          parent.insertBefore(document.createTextNode(strings[i]), node);
+          // IE doesn't clone empty text nodes, so use comments instead
+          previousNode = document.createTextNode(
+              envDoesntCloneEmptyTextNodes && strings[i] === '' ? marker :
+                                                                  strings[i]);
+          parent.insertBefore(previousNode, node);
           this.parts.push(new TemplatePart('node', index++));
         }
+
+        parent.insertBefore(
+            document.createTextNode(
+                envDoesntCloneEmptyTextNodes && strings[lastIndex] === '' ?
+                    marker :
+                    strings[lastIndex]),
+            node);
+        nodesToRemove.push(node);
       } else if (
           node.nodeType === 8 /* Node.COMMENT_NODE */ &&
           node.nodeValue === marker) {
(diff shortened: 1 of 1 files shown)


daKmoR:
I do get some errors on IE11 :/
ie11result

any way I can help out?

bgotink:
I've tested the code with the `envDoesntCloneEmptyTextNodes` forced to true on Chrome and others, so I don't think that the changed runtime code is a problem. Perhaps IE clones empty text nodes if they're the only child? That would make the check return a faulty result on IE which leads to lit-html using empty text nodes in the template itself.

I'll probably be able to access a Windows machine to verify this assumption tomorrow if I find the time.

daKmoR:
running this in IE11
```
var envDoesntCloneEmptyTextNodes = function() {
 var container = document.createElement('template');
 container.content.appendChild(document.createTextNode(''));
 return !document.importNode(container.content, true).childNodes.length;
}();
console.log(envDoesntCloneEmptyTextNodes); // true
```

so that should be fine...

bgotink:
Alright, thanks for checking! In that case I'll have to investigate some more. I didn't get the chance to boot a windows VM today, hopefully I'll find the time this weekend.

bgotink:
Apparently TreeWalkers skip empty text nodes in Internet Explorer, so the approach with two TreeWalkers to keep the impact on non-IE platforms low won't work.

```ts
(() => {
  const container = document.createElement('template');
  container.content.appendChild(document.createTextNode(''));
  container.content.appendChild(document.createTextNode(''));

  const walker = document.createTreeWalker(container.content, 4 /* NodeFilter.SHOW_TEXT */, null as any, false);
  let index = 0;

  console.log('walker :');
  while (walker.nextNode()) {
    const {currentNode} = walker;
    console.log(currentNode, 'has index', index++);
  }

  console.log('end');
})();
```

Console output in Chrome:

image

Console output in IE:

[I'll upload a screenshot from a windows machine in an edit to this post]

I'm not sure how to continue. I'd prefer to keep the impact on browsers-other-than-IE as low as possible and one if-check per clone was the best I could come up with.
Alternatives are

- Check `startNode` and `endNode` in the `NodePart` constructor. That's what the #202 PR did originally. This means non-IE browsers do a lot of extra work and it creates extra useless work, because not all `NodePart`s are created as part of the cloning process: `repeat` and `NodePart#_setIterable` create more instances.
- Add an if-check in the while in the for in the clone function. This also means non-IE browsers execute a lot of extra if-checks they don't need.
- Add an if-check around the for in the clone function to create two for-loops: one for non-IE browsers and one for IE. This approach leaves us at one extra if-check per clone for non-IE browsers but it duplicates code.

kaste:
Patch importNode. Something like this: https://github.com/kaste/maik.h/blob/master/src/dom-utils.js#L31

bgotink:
Patching importNode alone would not be enough, we'd need to drop the treewalker and do our own DOM traversal as well.

daKmoR:
With your second option you mean something like this 
```js
        while (index < part.index) {
          index++;
          walker.nextNode();
          if (envDoesntCloneEmptyTextNodes && walker.currentNode.nodeValue === marker) {
            walker.currentNode.nodeValue = '';
          }
        }
```
at ~https://github.com/PolymerLabs/lit-html/blob/master/src/lit-html.ts#L646 right?

I mean that just works then right? so full support everywhere and you get the markers only if needed... should be worth the (presumably small) impact...

bgotink:
Yeah that would be the second option. We'd need one extra check after the outer loop because the last Part might be followed by a text node with the marker

daKmoR:
do you care to update your pull request?
or do you want a pull request for your pull request :p

bgotink:
@daKmoR updated :)

I'm on a new computer and the modern.ie download links are broken so yet again I cannot test this on IE. Tests run fine on safari/firefox/chrome.

daKmoR:
@bgotink works almost perfectly on my IE11... just some small fixes for "for of loops"
https://github.com/bgotink/lit-html/pull/1

but with this, it works super nice :)
IE11 works nicely - no comment nodes or anything - yeah!!!
Chrome, Firefox looks like before the comment change and works nicely :)

bgotink:
These for-of loops are not an issue when I tested on IE before, shouldn't wct transpile those to for-in or regular for-loops?

Performance-wise there's no clear winner when comparing for-of/for/Array#forEach in different browsers ([jsperf](https://jsperf.com/for-of-vs-other-loops)) so perhaps moving to another loop would be feasible.

daKmoR:
yeah Test are all Green even on IE 11 without the change of the for of
loops...
but when I use it even in a simple webcomponent I do get the error on IE11
when using latest polymer serve...
once change it does work just fine on IE11...

bgotink:
Hmm that's weird&hellip; That would mean it transpiles the lit-html code (otherwise it obviously wouldn't work because of the ESM exports, arrow functions etc) but it keeps the for-of loops intact.
Maybe it'd be best to file an issue so this can be investigated and discussed?

daKmoR:
I meant it more like this...
for of loops do get transpiled on IE11 but there is an error thrown "Object expected"...

source:
```
    // Remove text binding nodes after the walk to not disturb the TreeWalker
    for (const n of nodesToRemove) {
      n.parentNode!.removeChild(n);
    }
```

IE11 Transpiled:
```
            // Remove text binding nodes after the walk to not disturb the TreeWalker
            var _iteratorNormalCompletion = true;
            var _didIteratorError = false;
            var _iteratorError = undefined;

            try {
                for (var _iterator = nodesToRemove[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
                    var n = _step.value;

                    n.parentNode.removeChild(n);
                }
            } catch (err) {
                _didIteratorError = true;
                _iteratorError = err;
            } finally {
                try {
                    if (!_iteratorNormalCompletion && _iterator.return) {
                        _iterator.return();
                    }
                } finally {
                    if (_didIteratorError) {
                        throw _iteratorError; //error thrown here "Object expected"
                    }
                }
            }
```


But if I replace the loop with
```
    nodesToRemove.forEach((n) => {
        n.parentNode!.removeChild(n);
    });
```

it work everywhere + IE11...

so I don't think the problem is within polymer serve...

bgotink:
Rebased unto master

justinfagnani:
Hey everyone, I'm taking a look at IE11 again.

Just a heads up: I wanted to explore some options and started from #202 as a base while trying to keep style and script bindings working properly. I like that because once we use comment nodes for placeholders, we can start to do less work when prepping the template - just leave the comment nodes in there as parsed.

So far I think a workable solution is somewhat like this PR, but use comment nodes in the templates, then swap them for empty text nodes on clone. We can limit the swapping to only in STYLE and SCRIPT elements too.

justinfagnani:
I got it working in #267, including a test for the style binding that was breaking under #202. Closing this one now. Thanks for all your work on this again @bgotink!

jpierson:
I'm seeing an error in the place where @daKmoR had [described above](https://github.com/Polymer/lit-html/pull/229#issuecomment-356448243). I'm using lit-html `0.11.0-dev.1` and trying to get what I had working otherwise working in IE11.

PR #239, opened 2018-01-03
Add previous value optimizations to AttributePart and PropertyPart

### LINKED ISSUE #199: AttributePart and PropertyPart need previous value optimizations
The dirty checking that NodePart does for primitive values should be ported to AttributePart and PropertyPart.

kenchris:
This ports the dirty checking code from NodePart into AttributePart
and PropertyPart as well.

Fixes #199

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -74,6 +74,9 @@ export class PropertyPart extends AttributePart {
   setValue(values: any[], startIndex: number): void {
     const s = this.strings;
     let value: any;
+    if (this._equalToPreviousValues(values, startIndex)) {
+      return;
+    }
     if (s.length === 2 && s[0] === '' && s[1] === '') {
       // An expression that occupies the whole attribute value will leave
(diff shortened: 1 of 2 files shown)


src/lib/lit-extended.ts
```diff
@@ -77,12 +77,20 @@ export class PropertyPart extends AttributePart {
     if (s.length === 2 && s[0] === '' && s[1] === '') {
       // An expression that occupies the whole attribute value will leave
       // leading and trailing empty strings.
+      if (values[startIndex] === this._previousValues[startIndex]) {
+        return;
```
justinfagnani: In NodePart we only do this for primitives, in case a nested value changed. Since complex objects can flow to here, and could have a different ToString value even if referentially identical, we need the same check.

justinfagnani:
@kenchris can you rebase to master to pick up the fixed wct config for travis?

kenchris:
Sure, done

justinfagnani:
ok, looks good!

PR #241, opened 2018-01-06
Keep fragment in the template's document during clone

### LINKED ISSUE #231: Error in shady DOM when rendering a template with shadow root and slotted content
On browsers that don't support native shadow dom (ex. firefox) there is an error when rendering a template that contains a shadow root with slotted content which is followed by a sub template.

I couldn't easily get a proper demo set up for firefox as it doesn't support native es modules. However forcing the shady dom polyfill on chrome also throws an error:

See: http://plnkr.co/edit/ofiGJR5T0ewsvPNrQluG?p=preview for a live example.

I created three template, two of which work and one that fails. It demonstrates that the template works fine if there is no slotted content:
```
    render(html`      
      <custom-element-with-shadowroot>
      </custom-element-with-shadowroot> 
      
      ${html`<h3>sub template</h3>`}
    `, document.body);
```

Or if the sub template precedes the shadow root:

```
    render(html`      
      ${html`<h3>sub template</h3>`}

      <custom-element-with-shadowroot>
        <h1>Slotted content</h1>
      </custom-element-with-shadowroot> 
    `, document.body);
```

But breaks when the sub template follows the shadow root:

```
    render(html`      
      <custom-element-with-shadowroot>
        <h1>Slotted content</h1>
      </custom-element-with-shadowroot> 
      
      ${html`<h3>sub template</h3>`}
    `, document.body);
```

bgotink:
The template's document doesn't have a browsing context, so custom elements aren't upgraded. This change defers upgrading any custom elements inside the template until after the treewalker that links the Parts to the DOM.
This is important because custom elements can modify their DOM at upgrade time (e.g. when using ShadyDOM).

Fixes #231

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -629,7 +629,7 @@ export class TemplateInstance {
     // Clone the node, rather than importing it, to keep the fragment in the
     // template's document. This leaves the fragment inert so custom elements
     // won't upgrade until after the main document adopts the node.
-    const fragment = this.template.element.content.cloneNode(true);
+    const fragment = this.template.element.content.cloneNode(true) as DocumentFragment;
     const parts = this.template.parts;
 
     if (parts.length > 0) {
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -626,7 +626,10 @@ export class TemplateInstance {
   }
 
   _clone(): DocumentFragment {
-    const fragment = document.importNode(this.template.element.content, true);
+    // Clone the node, rather than importing it, to keep the fragment in the
+    // template's document. This leaves the fragment inert so custom elements
+    // won't upgrade until after the main document adopts the node.
+    const fragment = this.template.element.content.cloneNode(true);
     const parts = this.template.parts;
 
     if (parts.length > 0) {


justinfagnani:
@bgotink I think that #231 should also be fixed by proper ShadyCSS support in #249.

This PR still seems potentially more correct, but I do worry that there's a cost to creating a node in the template's document, then changing the document before upgrade. I'd like to measure that.

bgotink:
I'm not sure that #249 will fix the problem in #231: custom elements that create a ShadyDOM shadowRoot in their constructor still modify the DOM tree before the walker links the nodes to the `Part`s, resulting in a mismatch and vague errors in the resulting DOM.

justinfagnani:
I merged into to master to make sure that the tests are still green: https://travis-ci.org/Polymer/lit-html/builds/380972476

src/lit-html.ts
```diff
@@ -626,7 +626,10 @@ export class TemplateInstance {
   }
 
   _clone(): DocumentFragment {
-    const fragment = document.importNode(this.template.element.content, true);
+    // Clone the node, rather than importing it, to keep the fragment in the
+    // template's document. This leaves the fragment inert so custom elements
+    // won't upgrade until after the main document adopts the node.
```
jridgewell: When does the main document adopt the clone? This seems like it will always leave the elements inert.

bgotink: Nodes are adopted by a document when they get added somewhere in that document's node tree. In the case of lit-html that is the `render` function at

https://github.com/Polymer/lit-html/blob/f4e8db4509d41d78b2d18563c0b0831d62f0816f/src/lit-html.ts#L184

PR #244, opened 2018-01-07
handle boolean attributes

### LINKED ISSUE #236: Boolean attributes are rendered incorrectly
When using boolean attributes, [the spec](http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes) states:
> [2.4.2. Boolean attributes](http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes)
> A number of attributes are boolean attributes. The presence of a boolean attribute on an element 
> represents the true value, and the absence of the attribute represents the false value.
>
> If the attribute is present, its value must either be the empty string or a value that is an ASCII case-
> insensitive match for the attribute’s canonical name, with no leading or trailing white space.

`lit-html` currently renders boolean attributes as `enabled="true"` or `enabled="false"` instead of using `enabled` when `true` and removing the attribute when `false`.

This could be resolved by updating `Attribute.setValue()` to something like:
```javascript
async setValue(values, startIndex) {
  const text = await this._interpolate(values, startIndex);
  if (text === 'false') {
    this.element.removeAttribute( this.name );

  } else if (text === 'true') {
    this.element.setAttribute(this.name, '');

  } else {
    this.element.setAttribute(this.name, text);

  }
```

43081j:
Fixes #236 

---

Does this seem like a stable enough fix @justinfagnani?

Basically we're checking if the remaining values is of length `1` and that the value is a bool. Then toggling the attribute instead of setting it etc.

Is it ok to rely on that length check? or is there a better way?

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -417,7 +417,20 @@ export class AttributePart implements MultiPart {
     return text + strings[l];
   }
 
+  protected _setBooleanValue(value: boolean) {
+    if (value) {
+      this.element.setAttribute(this.name, this.name);
+    } else {
+      this.element.removeAttribute(this.name);
+    }
+  }
+
   setValue(values: any[], startIndex: number): void {
+    if (this.strings.length === 2 && typeof values[startIndex] === 'boolean') {
+      this._setBooleanValue(values[startIndex]);
+      return;
+    }
+
     const text = this._interpolate(values, startIndex);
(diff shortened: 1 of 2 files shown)


justinfagnani:
@43081j I think I want to avoid value-type based tests for boolean attributes. I'd rather have an explicit syntax. #197 looks like the way to go to me.

Issue #222, opened 2017-12-10
unsafeHTML gets rerendered every time


daKmoR:
hey,

just noticed

if you have something like this
```js
const foo = '<h2>foo</h2>';
render(html`${unsafeHTML(foo)}`, container);

// ... later
render(html`${unsafeHTML(foo)}`, container);
```

this will fully update the h2 dom node every time - even though it's not needed.

easily fixed with
```ts
export const unsafeHTML = (value: any) => directive((part: NodePart) => {
  if (part.previousValue !== value) {
    const tmp = document.createElement('template');
    tmp.innerHTML = value;
    part.setValue(document.importNode(tmp.content, true));
    part.previousValue = value;
  }
});
```

is that the correct place to handle previousValue?
if so I can provide PR with tests :)

PR #247, opened 2018-01-12
fix: unchanged unsafeHTML code should not trigger dom update

daKmoR:
fixes: #222

### DIFF
--- a/src/lib/unsafe-html.ts
+++ b/src/lib/unsafe-html.ts
@@ -22,7 +22,10 @@ import {directive, NodePart} from '../lit-html.js';
  * vulnerabilities.
  */
 export const unsafeHTML = (value: any) => directive((part: NodePart) => {
-  const tmp = document.createElement('template');
-  tmp.innerHTML = value;
-  part.setValue(document.importNode(tmp.content, true));
+  if (part._previousValue !== value) {
+    const tmp = document.createElement('template');
+    tmp.innerHTML = value;
+    part.setValue(document.importNode(tmp.content, true));
+    part._previousValue = value;
+  }
 });


justinfagnani:
@daKmoR see my comments to @kenchris on #239: https://github.com/Polymer/lit-html/pull/239#discussion_r161332076

When we do short-cutting optimizations, we need to make sure that we only do it for primitive values, otherwise we could miss deep object property changes that would change the result of the ToString operation here.

The previous value check in NodePart is here: https://github.com/Polymer/lit-html/blob/master/src/lit-html.ts#L439

daKmoR:
so the problem is that even though we use 
`part.setValue(document.importNode(tmp.content, true));``
with all it's check it still does an rerender every time

probably because `document.importNode()` always returns a "fresh" copy...
```
    _setNode(value) {
        console.log('compare', this._previousValue); //#document-fragment (same content)
        console.log('value', value); //#document-fragment (same content)
        console.log('result', this._previousValue === value); //false even with same html content
        if (this._previousValue === value) {
            return;
        }
        this.clear();
        this._insert(value);
        this._previousValue = value;
    }
```
so yes indeed... the same but different...

so we could add `if (value instanceof DocumentFragment) {` and compare the "raw" html string.... but getting the "raw" html value from a DocumentFragment seems to be tough - e.g. not performant? see
```
var div = document.createElement('div');
div.appendChild( fragment.cloneNode(true) );
var html = div.innerHTML;
```

so as we probably already have the raw string within unsafeHTML... seems kind of a bad idea...

so maybe using a different variable e.g. `part._rawUnsafeHtmlpreviousValue = value;` and check it (maybe also only if it's a string?)

justinfagnani:
@daKmoR to revive this PR, we can do the check here: https://github.com/Polymer/lit-html/pull/247/files#diff-022f841b78069df937443c1ba0808659R25

But _only_ if it's a primitive value according to this check: https://github.com/Polymer/lit-html/blob/master/src/lit-html.ts#L436

Otherwise we don't know ahead of time that the toString value will be the same. THis will catch most cases of passing a string of HTML to `unsafeHTML`

justinfagnani:
@daKmoR I made a new PR at #384 that does the dirty check only for primitive values, and should take care of the main use cases with `unsafeHTML`. Closing this in favor of that one, thanks for the initial work!

PR #249, opened 2018-01-13
ShadyCSS Support

justinfagnani:
Ready for review now.

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -1,0 +1,52 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {render as baseRender, Template, templateCaches, TemplateResult} from '../lit-html.js';
+
+export {html, svg, TemplateResult} from '../lit-html.js';
+
+declare global {
+  interface Window {
+    ShadyCSS: any;
+  }
+}
+
+const shadyTemplateFactory = (scopeName: string) =>
+    (result: TemplateResult) => {
+      const cacheKey = `${result.type}--${scopeName}`;
+      let templateCache = templateCaches.get(cacheKey);
+      if (templateCache === undefined) {
+        templateCache = new Map<TemplateStringsArray, Template>();
+        templateCaches.set(cacheKey, templateCache);
+      }
+      let template = templateCache.get(result.strings);
+      if (template === undefined) {
+        const element = result.getTemplateElement();
+
+        if (typeof window.ShadyCSS === 'object') {
+          window.ShadyCSS.prepareTemplate(element, scopeName);
+        }
+
+        template = new Template(result, element);
+        templateCache.set(result.strings, template);
+      }
+      return template;
+    };
+
+export function render(
+    result: TemplateResult,
+    container: Element|DocumentFragment,
+    scopeName: string) {
+  return baseRender(result, container, shadyTemplateFactory(scopeName));
+}
--- a/src/test/lib/shady-render_test.ts
+++ b/src/test/lib/shady-render_test.ts
@@ -1,0 +1,42 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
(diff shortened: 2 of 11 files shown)


src/lib/shady-render.ts
```diff
@@ -0,0 +1,52 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {render as baseRender, Template, templateCaches, TemplateResult} from '../lit-html.js';
+
+export {html, svg, TemplateResult} from '../lit-html.js';
```
sorvell: Why export `html` and `svg` here?

src/lib/shady-render.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (templateCache === undefined) {
+        templateCache = new Map<TemplateStringsArray, Template>();
+        templateCaches.set(cacheKey, templateCache);
+      }
+      let template = templateCache.get(result.strings);
+      if (template === undefined) {
+        const element = result.getTemplateElement();
+
+        if (typeof window.ShadyCSS === 'object') {
+          window.ShadyCSS.prepareTemplate(element, scopeName);
```
sorvell: Currently ShadyCSS expects this to be called one time only per scope name. Multiple calls should result in correct static scoping (although lit should add a test for this), but we expect custom property shimming (needed in IE) to be broken. We'll add an issue to ShadyCSS to work on an issue for this.

src/test/lib/shady-render_test.ts
```diff
@@ -0,0 +1,42 @@
...
+/// <reference path="../../node_modules/@types/chai/index.d.ts" />
+
+const assert = chai.assert;
+
+suite('shady-render', () => {
+
+  test('prepares templates with ShadyCSS', () => {
+    const container = document.createElement('div');
+    render(html`
+      <style>
```
sorvell: Let's add a test that includes multiple style elements and one that uses nested templates.

src/lib/shady-render.ts
justinfagnani: Just for convenience. This is everything that would be needed for the base class. I can leave it out.

sorvell: Gotcha... eh, I think it's ok. Wanted to make sure it was not a mistake.

src/lib/shady-render.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        template = new Template(result, element);
+        templateCache.set(result.strings, template);
+      }
+      return template;
+    };
+
+export function render(
```
NeoLegends: How does this play with lit-extended? Seems like with this you can choose between either lit-extended or shady dom..?

src/lib/shady-render.ts
```diff
@@ -0,0 +1,52 @@
...
+    result: TemplateResult,
+    container: Element|DocumentFragment,
+    scopeName: string) {
+  return baseRender(result, container, shadyTemplateFactory(scopeName));
```
NeoLegends: How does this play with lit-extended? Seems like with this you can choose between either lit-extended or shady dom..?

bgotink: lit-extended now creates and exports its own `html` and `svg` template literals instead of its own `render` function, so this should work fine.

Example usage:

```js
import {html} from 'path/to/lit-html/lit-html.js';
import {render} from 'path/to/lit-html/lib/shady-render.js';

render(html`<div>test</div>`, container, 'custom-element-name');
```

NeoLegends: Ahh great! I always thought having the instantiation (and thus also the template syntax) depend on the render and not the template function was an API wart. Good thing thats gone now!

justinfagnani: Yeah, me too. Not sure why I didn't do it this way originally. 🤷‍♂️

@bgotink this will require some changes for you, but should be straight-forward. WDYT?

bgotink: Should be straightforward enough, similar to the changes made to `lit-extended.js`. I haven't tried it yet though.

daKmoR: that sure seems like a better approach :) looking forward to it :)

NeoLegends:
Is there anything holding this back from being merged?

src/test/lib/shady-render_test.ts
justinfagnani: Done. It seems to actually do the right thing in Chrome at least.

justinfagnani:
@NeoLegends nothing other than the test I just added. I think we'll get it merged very soon.

AndreasGalster:
FYI @apply shim would be nice to have next. I think it's one of the few things holding lit back to be usable in production.

I'm well aware that it's not considered a long-term solution but many apps still rely heavily on @apply mixins

justinfagnani:
@sorvell PTAL

stramel:
🎉

NeoLegends:
Obligatory question: When do you intend to publish this to npm 😅?

AndreasGalster:
Also, I think some docs in the readme are still needed how to make it work. Does this work out of the box, do I have to use the render function from shady-css instead? Does this do conditional polyfilling, etc.?

PR #250, opened 2018-01-14
Run travis tests with headless chrome

bgotink:
Chromedriver has issues with Travis when running a non-headless chrome, so run headless chrome instead.

This is probably a temporary workaround, but it _should_ get the build green again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
 before_script:
 - npm run build
 script:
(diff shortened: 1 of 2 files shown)


bgotink:
At least the build fails with a different message now 😅

> Chrome failed to start: crashed

justinfagnani:
hey @bgotink I must have missed this when I submitted a similar fix. Tests are working now. Thanks!

PR #259, opened 2018-01-28
Add boolean attribute support to lit-extended.

justinfagnani:
This adds support for explicit syntax for boolean attributes to lit-extended with the `?` attribute name suffix.

This is basically the same as #197 but accepts all values and removes and attribute if it's falsey.


CHANGELOG.md
```diff
@@ -10,7 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
+* lit-extended: Attribute bindings with a `?` suffix on the name now act as boolean
+  attributes. The attribute will be removed for falsey values and set to `''` for
+  truthy values.
```
kenchris: Maybe make it clear that this follows the HTML behavior:

From the HTML standard:

The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.

src/test/lib/lit-extended_test.ts
```diff
@@ -51,6 +51,35 @@ suite('lit-extended', () => {
       assert.equal(container.innerHTML, '<div foo="1bar2baz3"></div>');
     });
 
+    test('renders a boolean attribute as an empty string when truthy', () => {
+      let t = (value: any) => html`<div foo?="${value}"></div>`;
+
+      render(t(true), container);
+      assert.equal(container.innerHTML, '<div foo=""></div>');
```
kenchris: in HTML <div foo=""> is the same as <div foo> or <div foo=''>.

Wouldn't going for <div foo> seem more logical?

kenchris: Even if I set `<div foo="">` manually in devtools, it turns into `<div foo>`

zandaqo:
So, does this mean boolean/flag attributes aren't coming to the lit-core? 

Using lit-extended solely for boolean attributes would be a bit of an overkill for me, is there a better way to get the same result with a directive?

src/test/lib/lit-extended_test.ts
justinfagnani: I think this is the only way to do it from DOM APIs.

justinfagnani:
@zandaqo the problem is finding the syntactic space to indicate that something is a boolean attribute binding. We can do this in lit-extended because it's not _just_ HTML syntax like lit-core. Same problem with value-based semantics. `setAttribute` doesn't remove an attribute when passed `undefined`, so how does that jive with an attribute binding being declarative syntax for `setAttribute`?

I think for lit-core the solution is a directive. We can attack that next.

src/lib/lit-extended.ts
```diff
@@ -75,6 +80,23 @@ export const extendedPartCallback =
           return defaultPartCallback(instance, templatePart, node);
         };
 
+export class BooleanAttributePart extends AttributePart {
+  setValue(values: any[], startIndex: number): void {
+    const s = this.strings;
+    if (s.length === 2 && s[0] === '' && s[1] === '') {
```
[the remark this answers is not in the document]
justinfagnani: done

PR #260, opened 2018-01-29
Implement a repeat version using the array-splice algorithm

kenchris:
This is to get some feedback on how to make this work with the KeyFn's

### DIFF
--- a/src/lib/array-splice.ts
+++ b/src/lib/array-splice.ts
@@ -1,0 +1,284 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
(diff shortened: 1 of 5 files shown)


kenchris:
Some improvements can be made to the key caching, but what worries me the most is how to support Iterables, as I will have to do one loop, just to get the keys and to get the values.

Ideas @justinfagnani  ?

kenchris:
Works now and passes the tests

justinfagnani:
@kenchris thanks a lot for this, and sorry for the unusually large delay. The Polymer core team had said that the array-splice algorithm from Polymer wasn't ideal, and recently @kevinpschaaf has take up rewriting `repeat()` to use a similar algorithm as Vue and snabdom. I'll close this PR in light of that new work.

Thanks again!

PR #263, opened 2018-02-02
Fix 'every' typo

jordanthomas:
I believe this should be "ever" not "every".


jordanthomas:
I don't think I'll sign this CLA thing, given the change in question. If it's required to merge just consider this PR a friendly heads up. :)

justinfagnani:
Thanks!

PR #267, opened 2018-02-04
IE11 support

justinfagnani:
This PR uses comment nodes in templates, but replaces them with text nodes during initial cloning to continue to support all the previous binding positions. That's why tests don't have to change much compared to #202.

I added a test of a binding that requires new placeholder node generation in style tags to ensure that style bindings still work.

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -744,17 +744,6 @@ export class TemplateInstance {
           walker.nextNode();
         }
         let node = walker.currentNode;
-        // Swap the comment placeholders for empty text nodes
-        // The nextSibling of the part placeholder is the part endNode
-        if (node.nextSibling !== null && node.nextSibling!.nodeType === 8) {
-          node.parentNode!.replaceChild(document.createTextNode(''), node.nextSibling!);
-        }
-        // The part placeholder is the part startNode
-        if (node.nodeType === 8) {
-          node = document.createTextNode('');
-          walker.currentNode.parentNode!.replaceChild(node, walker.currentNode);
-          walker.currentNode = node;
-        }
         this._parts.push(this._partCallback(this, part, node));
       }
     }
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -337,17 +337,23 @@ export class Template {
         // We have a part for each match found
         partIndex += lastIndex;
 
-        // We keep this current node, but reset its content to the last
-        // literal part. We insert new literal nodes before this so that the
-        // tree walker keeps its position correctly.
-        node.textContent = strings[lastIndex];
-
         // Generate a new text node for each literal section
         // These nodes are also used as the markers for node parts
         for (let i = 0; i < lastIndex; i++) {
-          parent.insertBefore(document.createTextNode(strings[i]), node);
+          console.log(parent.nodeName);
+          parent.insertBefore(
+              (strings[i] === '')
+                  ? document.createComment('')
+                  : document.createTextNode(strings[i]),
+              node);
           this.parts.push(new TemplatePart('node', index++));
         }
+        parent.insertBefore(
+            strings[lastIndex] === '' ?
+                document.createComment('') :
+                document.createTextNode(strings[lastIndex]),
+            node);
+        nodesToRemove.push(node);
       } else if (
           node.nodeType === 8 /* Node.COMMENT_NODE */ &&
           node.nodeValue === marker) {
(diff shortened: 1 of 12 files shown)


src/lit-html.ts
```diff
@@ -738,7 +744,19 @@ export class TemplateInstance {
           index++;
           walker.nextNode();
         }
-        this._parts.push(this._partCallback(this, part, walker.currentNode));
+        let node = walker.currentNode;
+        // Swap the comment placeholders for empty text nodes
+        // The nextSibling of the part placeholder is the part endNode
+        if (node.nextSibling !== null && node.nextSibling!.nodeType === 8) {
```
bgotink: This also removes comments placed by the library user

```ts
    test('preserves comments', () => {
      const tpl = html`<!-- test -->${'foo'}<!-- test -->`;
      render(tpl, container);
      assert.equal(container.innerHTML, '<!-- test -->foo<!-- test -->');
    });
```

```
chrome 63                Beginning tests via http://localhost:8081/components/lit-html/generated-index.html?cli_browser_id=0
chrome failed to maximize
chrome 63                ✖ test/ » lit-html » html » preserves comments

  expected '<!-- test -->foo' to equal '<!-- test -->foo<!-- test -->'
    <unknown> at   Context.test at lit-html_test.js:58:0

chrome 63                Tests failed: 1 failed tests
```

src/lit-html.ts
```diff
@@ -738,7 +744,19 @@ export class TemplateInstance {
...
+          node.parentNode!.replaceChild(document.createTextNode(''), node.nextSibling!);
+        }
+        // The part placeholder is the part startNode
+        if (node.nodeType === 8) {
```
bgotink: Throughout the library we tend to put `/* Node.COMMENT_NODE */` comments behind these magic numbers.
Maybe it'd be nice to look into a `const enum`? That results in the magic numbers in javascript but it keeps them out of the typescript.

```ts
const enum NodeType {
  COMMENT_NODE = 8
}

if (node.nodeType === NodeType.COMMENT_NODE) {
  // foo
}
```

becomes

```js
if (node.nodeType === 8 /* COMMENT_NODE */) {
    // foo
}
```

src/lit-html.ts
```diff
@@ -337,17 +337,23 @@ export class Template {
...
-        // We keep this current node, but reset its content to the last
-        // literal part. We insert new literal nodes before this so that the
-        // tree walker keeps its position correctly.
-        node.textContent = strings[lastIndex];
-
         // Generate a new text node for each literal section
         // These nodes are also used as the markers for node parts
         for (let i = 0; i < lastIndex; i++) {
-          parent.insertBefore(document.createTextNode(strings[i]), node);
+          console.log(parent.nodeName);
```
daKmoR: missed console.log

LarsDenBakker:
Looks promising, what can be done to help this move along? Do we need more test cases?

itssumitrai:
any way to help in order to get this branch merged ?

src/lit-html.ts
```diff
@@ -337,17 +337,23 @@ export class Template {
...
+          parent.insertBefore(
+              (strings[i] === '')
+                  ? document.createComment('')
+                  : document.createTextNode(strings[i]),
+              node);
           this.parts.push(new TemplatePart('node', index++));
         }
+        parent.insertBefore(
+            strings[lastIndex] === '' ?
+                document.createComment('') :
```
sorvell: Can we consider using comment nodes only if required? (ideally with feature detection, but if necessary via a special setting for IE11).

src/lit-html.ts
```diff
@@ -738,7 +744,19 @@ export class TemplateInstance {
           index++;
           walker.nextNode();
         }
-        this._parts.push(this._partCallback(this, part, walker.currentNode));
+        let node = walker.currentNode;
+        // Swap the comment placeholders for empty text nodes
+        // The nextSibling of the part placeholder is the part endNode
+        if (node.nextSibling !== null && node.nextSibling!.nodeType === 8) {
+          node.parentNode!.replaceChild(document.createTextNode(''), node.nextSibling!);
```
sorvell: Replacing each of these on every stamping seems unfortunate, but probably needs some perf gut checking. If this is an issue, it can be isolated to just where the issue occurs (IE11?)

stramel:
Awesome work y'all! :tada:

jridgewell:
@justinfagnani: I don't see how this is any different than #202, which caused #214.

PR #268, opened 2018-02-05
Add <title> to the docs

limhenry:


justinfagnani:
These docs are generated in tools/src/site.ts. We'll want to make the change here: https://github.com/Polymer/lit-html/blob/master/tools/src/site.ts#L98

limhenry:
oops. sure!

justinfagnani:
I added this to the site generator at: https://github.com/Polymer/lit-html/pull/329

PR #276, opened 2018-02-13
implement a $NO_OP return value  that indicates already from the comp…

fopsdev:
…onent that nothing was changed

eg. if the application using lit-html already uses a caching layer

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -413,7 +413,7 @@ export const getValue = (part: Part, value: any) => {
   // so we convert it to undefined
   if (isDirective(value)) {
     value = value(part);
-    return directiveValue;
+    return noChange;
   }
   return value === null ? undefined : value;
 };
@@ -435,7 +435,7 @@ const isDirective = (o: any) =>
  * A sentinel value that signals that a value was handled by a directive and
  * should not be written to the DOM.
  */
-export const directiveValue = {};
+export const noChange = {};
 
 const isPrimitiveValue = (value: any) => value === null ||
     !(typeof value === 'object' || typeof value === 'function');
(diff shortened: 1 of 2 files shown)


fopsdev:
I signed it!

NeoLegends:
Why don't you just not call `render` if your data hasn't changed? Shouldn't that result in the same?

fopsdev:
@NeoLegends 
because there are components up the tree which needs to be rendered
```
A
|-B
```
lets say render starts with A which uses B but only B can be cached.

NeoLegends:
Ah, I understand. B cannot `render` itself (instead of being rendered by A)?

src/lit-html.ts
```diff
@@ -528,6 +528,11 @@ export class NodePart implements SinglePart {
 
   setValue(value: any): void {
     value = getValue(this, value);
+    if (value === '$NO_OP') {
```
justinfagnani: Rather than a magic string, I'd like to export a special value, like we do for `directiveValue`. In fact, `directiveValue` might be able to serve as this value already. Maybe it just needs a new name.

NeoLegends: What about a symbol?

justinfagnani: It doesn't really matter - any unique object will work - except that IE11 doesn't have Symbols.

fopsdev: yeah looks like directiveValue is handling it as well :)
what about renaming directiveValue to alreadyHandled?

justinfagnani: I think something like `noChange` might be a better name.

fopsdev: fine for me... lemme know if i should create a new pr

ruphin: I was running into the same issue today, and wasn't able to find the `directiveValue` solution. Changing the name to `noChange` sounds good to me

fopsdev:
created a new pr: https://github.com/Polymer/lit-html/pull/357

PR #284, opened 2018-03-01
improve readme code smaple

uxitten:


README.md
```diff
@@ -104,13 +104,13 @@ This should make the approach generally fast and small. Actual science and optim
 Anything coercible to strings are supported:
 
 ```javascript
-const render = () => html`foo is ${foo}`;
```
justinfagnani: Our style is to always use parens for arrow functions.

README.md
```diff
@@ -104,13 +104,13 @@ This should make the approach generally fast and small. Actual science and optim
...
+const render = foo => html`foo is ${foo}`;
 ```
 
 ### Attribute-value Expressions
 
 ```javascript
-const render = () => html`<div class="${blue}"></div>`;
```
justinfagnani: I'm not sure the changes to add parameters are necessary... The expressions are just JS that can close over variables in outer scopes. These snippets aren't mean to be complete.

README.md
uxitten: ok but write argument name in parens

uxitten: (foo) => ...

justinfagnani: These aren't necessarily parameters, they could be closed over variables from outer scopes.

uxitten: But the user should guess this

Issue #295, opened 2018-03-20
MS Edge: NamedNodeMap.getNamedItem('some-attr$') throws


keanulee:
On Edge, `NamedNodeMap.getNamedItem('some-attr$')` throws an `InvalidCharacterError` (suspect the `$`, which is used for attribute binding). This is used here - https://github.com/Polymer/lit-html/blob/master/src/lit-html.ts#L316

valdrinkoshi:
(Including the sha to have the line rendered inline)

https://github.com/Polymer/lit-html/blob/54afddcfa7de802181e502f3b69599aa0507efb8/src/lit-html.ts#L316

justinfagnani:
Along with #302, there's a possible fix here: https://github.com/Polymer/lit-html/pull/344/files#diff-7548be97e171131880082c935a962f68R326

We need a regression test though.

csvn:
Should not be hard to reproduce for testing. The below code does not work in Edge 16. (https://jsfiddle.net/nf4w7o92/7/)

```html
<script type="module">
import { html, render } from 'https://unpkg.com/lit-html@0.10.2/lib/lit-extended.js';

render(html`<h2 title$="${'test'}">Should render without error</h2>`, document.body);
</script>
```

bgotink:
The tests contain scenarios where attributes contain `$`, e.g.

https://github.com/Polymer/lit-html/blob/1a993e74411981e623c6d66cc2fb81ff279932f5/src/test/lib/lit-extended_test.ts#L41-L44

The jsfiddle works fine in Edge 17 (see screenshot below). The tests work fine in Edge 15.
Maybe this bug only affects Edge 16? A regression test might be as simple as adding `windows 10/microsoftedge@16` to the list of browsers to test in travis.

![2018-06-24 11 22 37](https://user-images.githubusercontent.com/821510/41817729-04da7c9c-77a1-11e8-8e60-7e1531377a33.jpg)

jridgewell:
@justinfagnani: The issue here is Edge doesn't accept invalid XML attr names, which follow different rules than HTML's attrs:

- Name: https://www.w3.org/TR/xml/#NT-Name
  - NameStartChar: https://www.w3.org/TR/xml/#NT-NameStartChar
  - NameChar: https://www.w3.org/TR/xml/#NT-NameChar

[...]

PR #302, opened 2018-04-05
fixes invalidCharacter error in Edge

filaraujo:
Background: 

attributes created with special attribute character will not be processed in Edge. This commit will allow attempt to `attributes.getNamedItem` and if that fails fall back to `node.getAttribute` instead.

fixes issue #295

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -312,8 +312,20 @@ export class Template {
           // Find the attribute name
           const attributeNameInPart =
               lastAttributeNameRegex.exec(stringForPart)![1];
-          // Find the corresponding attribute
-          const attribute = attributes.getNamedItem(attributeNameInPart);
+
+          let attribute: { name: string, value: any };
+
+          try {
+            // Find the corresponding attribute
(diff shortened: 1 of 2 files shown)


justinfagnani:
@filaraujo our Edge tests are passing. Do you have an example that is failing for you?

filaraujo:
ill have to reproduce this again - it was happening locally on a project. I'll update the pr with tests if i can reproduce it again

justinfagnani:
@filaraujo @MartinsThiago any reproduction would be great. Like I said, we have tests that use the `?` suffix and they are passing on Edge. Before merging this change I'd want to see a failing test that passes with the change.

justinfagnani:
As part of another PR, I have a change that should fix by building up a map of bound attributes: https://github.com/Polymer/lit-html/pull/344/files#diff-7548be97e171131880082c935a962f68R326

I like this a little better than causing exceptions, so I'm going to close this PR and we can extract that fix from the other. We still need a failing test though.

PR #307, opened 2018-04-06
Makes EventPart more efficiently update

### LINKED ISSUE #306: [lit-extended] EventPart is inefficiently updated for a common case
Consider rendering simple content when an event listener defined inline like this:

`<h3 on-click="${(e) => console.log(e.target.textContent)}">Hi</h3>`

This seems pretty straightforward but as `EventPart` is currently implemented, whenever this template is re-rendered, the event listener on the `h3` is removed and re-added. This is because `EventPart` sees this as a new listener and provokes a removal of the old listener and an add of the new listener.

While the most efficient thing is to generate this function once outside of rendering instead of generating it once per render, this is significantly less expensive than removing and adding the event listener when this happens so it'd be great if this weren't the required way to address this issue.

Instead, `EventPart` could install the listener 1x using a listener that delegates to the user's method. Then the user's method could be easily changed with a simple property set.

sorvell:
Fixes #306. Instead of calling add/removeEventListener whenever an EventPart's listener changes, installs the listener 1x and stores/calls the user's method, updating it via a simple property set when needed.

Note, this uses a previously committed method of updating the EventPart event listener.

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -144,17 +144,22 @@ export class EventPart implements Part {
 
   setValue(value: any): void {
     const listener = getValue(this, value);
-    const previous = this._listener;
-    if (listener === previous) {
+    if (listener === this._listener) {
       return;
     }
-
-    this._listener = listener;
-    if (previous != null) {
-      this.element.removeEventListener(this.eventName, previous);
+    if (listener == null) {
+      this.element.removeEventListener(this.eventName, this);
+    } else if (this._listener == null) {
+      this.element.addEventListener(this.eventName, this);
     }
-    if (listener != null) {
-      this.element.addEventListener(this.eventName, listener);
+    this._listener = listener;
+  }
+
+  handleEvent(event: Event) {
+    if (typeof this._listener === 'function') {
+      this._listener.call(this.element, event);
+    } else if (typeof this._listener.handleEvent === 'function') {
+      this._listener.handleEvent(event);
     }
   }
 }
--- a/src/test/lib/lit-extended_test.ts
+++ b/src/test/lib/lit-extended_test.ts
@@ -203,6 +203,36 @@ suite('lit-extended', () => {
       assert.equal(count2, 1);
     });
 
+    test('allows updating event listener without extra calls to remove/addEventListener', () => {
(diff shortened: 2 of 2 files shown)


src/test/lib/lit-extended_test.ts
```diff
@@ -203,6 +203,36 @@ suite('lit-extended', () => {
       assert.equal(count2, 1);
     });
 
+    test('allows updating event listener without extra calls to remove/addEventListener', () => {
+      let listener: Function|null;
+      const t = () => html`<div on-click=${listener}></div>`;
```
justinfagnani: I think the test would be a little easier to follow if the listener was just a parameter to `t` rather than a variable.

src/lib/lit-extended.ts
```diff
@@ -144,17 +144,22 @@ export class EventPart implements Part {
...
+    } else if (this._listener == null) {
+      this.element.addEventListener(this.eventName, this);
     }
-    if (listener != null) {
-      this.element.addEventListener(this.eventName, listener);
+    this._listener = listener;
+  }
+
+  handleEvent(event: Event) {
+    if (typeof this._listener === 'function') {
```
justinfagnani: It's possible that the listener could be an object with a `handleEvent` method. I see I didn't handle this in my previous implementation either...

src/test/lib/lit-extended_test.ts
```diff
@@ -203,6 +203,36 @@ suite('lit-extended', () => {
...
+      render(t(), container);
+      const div = container.firstChild as HTMLElement;
+      let addCount = 0;
+      let removeCount = 0;
+      div.addEventListener = () => addCount++;
+      div.removeEventListener = () => removeCount++;
+      listener = () => {};
```
justinfagnani: Since we're interceding the handler, it'd be good to test that the listener is called.

sorvell: There's [another test](https://github.com/Polymer/lit-html/blob/fix-306/src/test/lib/lit-extended_test.ts#L188) for changing the event listener. This one was just to ensure that changing it doesn't spam add/removeEventListener. wdyt?

PR #310, opened 2018-04-11
IE11 Support

filaraujo:
IE11 does not import empty text nodes. This adds text node markers and then removes them as part of the content cloning and importing. #201 

I understand there is already an IE11 Support #267 PR opened, however this one uses the already existing marker system in lit-html does not have the side effect of removing comments.


justinfagnani:
We just merged #267 for IE11 support

Thanks for this though!

PR #313, opened 2018-05-04
Test that comments aren't inserted into textContent

jridgewell:
Reopens #214.
Re: https://github.com/Polymer/lit-html/pull/267#issuecomment-386487510

### DIFF
--- a/src/test/lit-html_test.ts
+++ b/src/test/lit-html_test.ts
@@ -489,6 +489,11 @@ suite('lit-html', () => {
           `);
       });
 
+      test('renders no comments inside textContent', () => {
+        render(html`<style>${''}</style>`, container);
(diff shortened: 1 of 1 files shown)


justinfagnani:
@jridgewell I'm adding in this test, but checking `.textContent` instead of `.innerHTML`, and it passes. CSS engines depend on the `.textContent` of `<style>`, so the presence of comment nodes is fine.

Granted, other tools or libraries which check innerHTML on styles may hit problems, but they should be looking at textContent.

jridgewell:
👍

jridgewell:
Updated the test.

justinfagnani:
@jridgewell I submitted already in #314, thanks!

PR #319, opened 2018-05-10
Add badges

FluorescentHallucinogen:


web-padawan:
The badges are pointing to the incorrect repo. This is `lit-html`, not `lit-element` and I doubt that webcomponents.org badge should be present here.

PR #328, opened 2018-05-19
Update dependencies

justinfagnani:
This removes some security warnings in `npm audit`, but not all. The rest come from still outdated transitive dependencies in tools.


justinfagnani:
Merged TBR to try to remove security warnings.

PR #330, opened 2018-05-19
Add a type for template containers.

justinfagnani:
This reduces the casts to `any` a little. Supersedes #227

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -140,6 +140,10 @@ export function defaultTemplateFactory(result: TemplateResult) {
   return template;
 }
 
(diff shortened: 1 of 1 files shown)


src/lit-html.ts
```diff
@@ -159,7 +163,7 @@ export function render(
     container: Element|DocumentFragment,
```
rictic: Why not make TemplateContainer the type of `container` here? It'll still be legal to pass in any Element or DocumentFragment

justinfagnani: Ah, that's true because the property is optional. Still, this is an internal detail of the render function. I don't want the type to change if we used a WeakMap instead of an expando property.

PR #337, opened 2018-05-22
Shady render fix

### LINKED ISSUE #326: shady-render does not work with ApplyShim or the CSS Custom Property shim (IE11)
There are 2 related issues here:
1. `styleElement` is currently never called and is required for the ApplyShim or Custom Property shim (IE11) to work correctly
1. Calling `prepareTemplate` multiple times as `shady-render` does is incompatible with `styleElement`

So, to address this, we'll need to call `prepareTemplate` only once per element type and figure out when/how to call `styleElement` within rendering.

sorvell:
Fixes #326 by imposing a limitation that styles for a scope name can only be supplied in the initial render of that shadowRoot. This is required because `ShadyCSS` can only prepare styles for a scope once to properly track styles in that scope and process custom properties when polyfilled and @apply.

Styles are processed based on initial scope rendering, but templated DOM within a scope will always be correctly given a scoping class. 

Note that styles in the first render of a scope can contain lit parts. This is useful for one-time bindings.

**Warning:** merge only after [this ShadyCSS PR](https://github.com/webcomponents/shadycss/pull/183) is merged and released.

### DIFF
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -85,7 +85,7 @@ const countNodes = (node: Node) => {
     count++;
   }
   return count;
-}
+};
 
 const nextActiveIndexInTemplateParts = (parts: TemplatePart[], startIndex: number = -1) => {
   for (let i = startIndex + 1; i < parts.length; i++) {
@@ -95,7 +95,7 @@ const nextActiveIndexInTemplateParts = (parts: TemplatePart[], startIndex: numbe
     }
   }
   return -1;
-}
+};
 
 /**
  * Inserts the given node into the Template, optionally before the given
@@ -110,8 +110,6 @@ export function insertNodeIntoTemplate(
   if (refNode === null || refNode === undefined) {
     content.appendChild(node);
     return;
-  } else if (!content.contains(refNode)) {
-    throw 'Reference node must be inside template.';
   }
   const walker = document.createTreeWalker(
       content,
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -1,0 +1,82 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Template, TemplatePart} from '../lit-html.js';
+
+export function removeNodesFromTemplate(template: Template, nodes: Node[]) {
+  const {element: {content}, parts} = template;
+  const walker = document.createTreeWalker(
+    content,
+    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+        NodeFilter.SHOW_TEXT,
+    null as any,
+    false);
+  let nodesIndex = 0;
+  let nodeToFind = nodes[nodesIndex];
+  let partIndex = 0;
+  let part = parts[partIndex];
+  let partDelta = 0;
+  let disableUntil = 0;
+  const nodesToRemove: Node[] = [];
+  let index = -1;
+  while (walker.nextNode()) {
+    index++;
+    const node = walker.currentNode as Element;
+    if (node === nodeToFind) {
+      nodeToFind = nodes[++nodesIndex];
+      const removeCount = node.childNodes.length + 1;
+      disableUntil = index + removeCount;
+      partDelta -= removeCount;
+      nodesToRemove.push(node);
+    }
+    // adjust part indices or disable parts.
+    while (part && part.index === index) {
+      if (index < disableUntil) {
+        part.index = -1;
+      } else {
+        part.index += partDelta;
+      }
+      part = parts[++partIndex];
+    }
+  }
+  nodesToRemove.forEach((n) => n.parentNode!.removeChild(n));
+}
+
+export function insertNodeIntoTemplate(template: Template, node: Node, refNode: Node|null = null) {
+  const {element: {content}, parts} = template;
+  const walker = document.createTreeWalker(
+    content,
+    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+        NodeFilter.SHOW_TEXT,
+    null as any,
+    false);
+  let partIndex = 0;
+  let part = parts[partIndex];
+  let partDelta = 0;
+  let index = -1;
+  while (walker.nextNode()) {
+    index++;
+    const currentNode = walker.currentNode as Element;
+    if (currentNode === refNode) {
+      content.insertBefore(node, refNode);
+      partDelta = 1 + node.childNodes.length;
+    }
+    while (part && (part.index === index || part.index < 0)) {
+      if (part.index >= 0) {
+        part.index += partDelta;
+      }
+      part = parts[++partIndex];
+    }
+  }
+}
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {Template, TemplatePart} from '../lit-html.js';
+import {Template} from '../lit-html.js';
 
 export function removeNodesFromTemplate(template: Template, nodes: Node[]) {
   const {element: {content}, parts} = template;
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -14,7 +14,12 @@
 
 import {Template} from '../lit-html.js';
 
-export function removeNodesFromTemplate(template: Template, nodes: Node[]) {
+/**
+ * Removes the list of nodes from a Template safely. In addition to removing
+ * nodes from the Template, the Template part indices are updated to match
+ * the mutated Template DOM.
+ */
+export function removeNodesFromTemplate(template: Template, nodesToRemove: Set<Node>) {
   const {element: {content}, parts} = template;
   const walker = document.createTreeWalker(
     content,
@@ -22,57 +27,61 @@ export function removeNodesFromTemplate(template: Template, nodes: Node[]) {
         NodeFilter.SHOW_TEXT,
     null as any,
     false);
-  let nodesIndex = 0;
-  let nodeToFind = nodes[nodesIndex];
   let partIndex = 0;
-  let part = parts[partIndex];
-  let partDelta = 0;
-  let disableUntil = 0;
-  const nodesToRemove: Node[] = [];
-  let index = -1;
+  let part = parts[0];
+  let nodeIndex = -1;
+  let removeCount = 0;
+  const nodesToRemoveInTemplate = [];
+  // TODO(sorvell): fix type
+  const removalStack: any[] = [];
   while (walker.nextNode()) {
-    index++;
+    nodeIndex++;
     const node = walker.currentNode as Element;
-    if (node === nodeToFind) {
-      nodeToFind = nodes[++nodesIndex];
-      const removeCount = node.childNodes.length + 1;
-      disableUntil = index + removeCount;
-      partDelta -= removeCount;
-      nodesToRemove.push(node);
+    // maybe pop the stack
+    if (walker.previousSibling === removalStack[removalStack.length - 1]) {
+      removalStack.pop();
     }
-    // adjust part indices or disable parts.
-    while (part && part.index === index) {
-      if (index < disableUntil) {
-        part.index = -1;
-      } else {
-        part.index += partDelta;
-      }
+    // track nodes we're removing
+    if (nodesToRemove.has(node)) {
+      removalStack.push(node);
+      nodesToRemoveInTemplate.push(node);
+    }
+    if (removalStack.length > 0) {
+      removeCount++;
+    }
+    while (part && part.index === nodeIndex) {
+      part.index = removalStack.length ? -1 : part.index - removeCount;
       part = parts[++partIndex];
     }
   }
-  nodesToRemove.forEach((n) => n.parentNode!.removeChild(n));
+  nodesToRemoveInTemplate.forEach((n) => n.parentNode!.removeChild(n));
 }
 
-export function insertNodeIntoTemplate(template: Template, node: Node, refNode: Node|null = null) {
+/**
+ * Inserts the given node into the Template, optionally before the given
+ * refNode. In addition to inserting the node into the Template, the Template
+ * part indices are updated to match the mutated Template DOM.
+ */
+export function insertNodeIntoTemplate(
+    template: Template, node: Node, refNode: Node|null = null) {
   const {element: {content}, parts} = template;
   const walker = document.createTreeWalker(
-    content,
-    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
-        NodeFilter.SHOW_TEXT,
-    null as any,
-    false);
+      content,
+      NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT,
+      null as any,
+      false);
   let partIndex = 0;
   let part = parts[partIndex];
   let partDelta = 0;
-  let index = -1;
+  let walkerIndex = -1;
   while (walker.nextNode()) {
-    index++;
+    walkerIndex++;
     const currentNode = walker.currentNode as Element;
     if (currentNode === refNode) {
       content.insertBefore(node, refNode);
       partDelta = 1 + node.childNodes.length;
     }
-    while (part && (part.index === index || part.index < 0)) {
+    while (part && (part.index === walkerIndex || part.index < 0)) {
       if (part.index >= 0) {
         part.index += partDelta;
       }
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -12,7 +12,10 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {Template} from '../lit-html.js';
+import {Template, isTemplatePartActive} from '../lit-html.js';
+
+const walkerNodeFilter = NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+NodeFilter.SHOW_TEXT;
 
 /**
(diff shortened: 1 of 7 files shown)


src/lib/shady-render.ts
```diff
@@ -22,31 +23,83 @@ declare global {
...
+            // insert style into rendered fragment
+            fragment.insertBefore(style, fragment.firstChild);
+            // insert into lit-template (for subsequent renders)
+            const template = templateFactory(result);
+            insertNodeIntoTemplate(template, style.cloneNode(true),
+                template.element.content.firstChild);
+          }
+        }
+      }
+      window.ShadyCSS.styleElement(host);
```
smalluban: For the first time calling `ShadyCSS.styleElement()` is ok, but if a change is made to the template and value of CSS custom property is changed, `ShadyCSS.styleSubtree()` has to be called to update nested elements using those CSS custom properties.

For example: 
```javascript
template = (active) => html`
  <style>
     my-button {
       --custom-color: red;
     }
     my-button.active {
       --custom-color: blue;
     }
  </style>
  <my-button class="${active ? 'active' : ''}"></my-button>
`;
```
Without calling `styleSubtree`, after change `active` to `true` in IE11 `<my-button>` will be still red.

abdonrd:
Any plan to replace the `@apply` with some `::part` and `::theme` polyfill?
I think new elements with LitElement should avoid deprecated things like `@apply`.

src/lib/shady-render.ts
sorvell: There's not a spot to do this in lit-html, but it does deserve documentation since it is necessary. It will be handled separately in LitElement.

smalluban: But there is a spot... You actually put `window.ShadyCSS.styleElement(host)` here. It has to be simply changed to `styleSubtree` to update all nested elements as well. I think it is very low-level API and LitElement using lit-html won't be able to do anything with that.

sorvell: The call to `styleElement` is only made when a lit `TemplateInstance` is created, not updated. Typically this is first render only, and this is done for efficiency. Therefore, changing this to `styleSubtree` wouldn't help with the case outlined here.

In general, ShadyCSS [requires](https://github.com/webcomponents/shadycss#usage) calling `ShadyCSS.styleSubtree` when dynamic updates that affect custom properties are made. This is specifically for the custom property shim which is only needed on IE11.

Calling `ShadyCSS.styleSubtree` is not cheap and doing so on every call to `shady-render` would have a pretty negative performance impact. For this reason, it seems better **not** to handle this case generally in `shady-render` and instead require users to call `ShadyCSS.styleSubtree` when needed. This is not ideal but seems like the best tradeoff.

src/lib/shady-render.ts
```diff
@@ -22,31 +23,83 @@ declare global {
...
 
 export function render(
     result: TemplateResult,
     container: Element|DocumentFragment,
     scopeName: string) {
-  return baseRender(result, container, shadyTemplateFactory(scopeName));
+  const host = hostForNode(container);
+  if (host && typeof window.ShadyCSS === 'object') {
+    const templateFactory = shadyTemplateFactory(scopeName);
+    const renderer = (container: Element|DocumentFragment, fragment: DocumentFragment) => {
```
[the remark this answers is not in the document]
justinfagnani: Talked offline, and it seems better to copy & paste the contents of render() for now, rather than expose an unstable API in the form of the renderDOM callback

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Template} from '../lit-html.js';
+
+export function removeNodesFromTemplate(template: Template, nodes: Node[]) {
```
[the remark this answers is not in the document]
sorvell: done

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+  const {element: {content}, parts} = template;
+  const walker = document.createTreeWalker(
+    content,
+    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+        NodeFilter.SHOW_TEXT,
+    null as any,
+    false);
+  let nodesIndex = 0;
```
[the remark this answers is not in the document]
sorvell: refactored

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+  let nodeToFind = nodes[nodesIndex];
```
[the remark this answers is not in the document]
sorvell: refactored

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+  let partIndex = 0;
+  let part = parts[partIndex];
+  let partDelta = 0;
+  let disableUntil = 0;
```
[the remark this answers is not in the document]
sorvell: refactored

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+  const walker = document.createTreeWalker(
+    content,
+    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT |
+        NodeFilter.SHOW_TEXT,
+    null as any,
+    false);
+  let nodesIndex = 0;
+  let nodeToFind = nodes[nodesIndex];
+  let partIndex = 0;
+  let part = parts[partIndex];
```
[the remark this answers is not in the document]
sorvell: refactored

src/lib/modify-template.ts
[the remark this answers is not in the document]
sorvell: refactored

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+  let partDelta = 0;
+  let disableUntil = 0;
+  const nodesToRemove: Node[] = [];
+  let index = -1;
+  while (walker.nextNode()) {
+    index++;
+    const node = walker.currentNode as Element;
+    if (node === nodeToFind) {
+      nodeToFind = nodes[++nodesIndex];
+      const removeCount = node.childNodes.length + 1;
```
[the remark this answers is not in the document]
sorvell: refactored to fix but needs tests.

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+      disableUntil = index + removeCount;
+      partDelta -= removeCount;
+      nodesToRemove.push(node);
+    }
+    // adjust part indices or disable parts.
+    while (part && part.index === index) {
+      if (index < disableUntil) {
+        part.index = -1;
```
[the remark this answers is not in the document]
sorvell: Added a specific method and comment.

src/lib/modify-template.ts
```diff
@@ -0,0 +1,82 @@
...
+    null as any,
+    false);
+  let nodesIndex = 0;
+  let nodeToFind = nodes[nodesIndex];
+  let partIndex = 0;
+  let part = parts[partIndex];
+  let partDelta = 0;
+  let disableUntil = 0;
+  const nodesToRemove: Node[] = [];
+  let index = -1;
```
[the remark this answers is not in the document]
sorvell: renamed

src/lib/shady-render.ts
```diff
@@ -22,31 +23,83 @@ declare global {
...
+}
 
-        template = new Template(result, element);
-        templateCache.set(result.strings, template);
-      }
-      return template;
-    };
+const shadyRenderSet = new Set<string>();
+
+function hostForNode(node: Node) {
```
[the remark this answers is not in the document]
sorvell: refactored

src/lib/shady-render.ts
```diff
@@ -22,31 +23,83 @@ declare global {
...
+  return node.nodeType === Node.DOCUMENT_FRAGMENT_NODE && (node as ShadowRoot).host
```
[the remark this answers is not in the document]
sorvell: done

src/lib/shady-render.ts
```diff
@@ -22,31 +23,83 @@ declare global {
...
+}
 
 export function render(
     result: TemplateResult,
     container: Element|DocumentFragment,
     scopeName: string) {
-  return baseRender(result, container, shadyTemplateFactory(scopeName));
+  const host = hostForNode(container);
+  if (host && typeof window.ShadyCSS === 'object') {
```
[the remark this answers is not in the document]
sorvell: done

src/lib/shady-render.ts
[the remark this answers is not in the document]
sorvell: done

src/lib/modify-template.ts
```diff
@@ -23,40 +26,57 @@ export function removeNodesFromTemplate(template: Template, nodesToRemove: Set<N
...
+  let currentRemovingNode: Node|null = null;
   while (walker.nextNode()) {
     nodeIndex++;
     const node = walker.currentNode as Element;
-    // maybe pop the stack
-    if (walker.previousSibling === removalStack[removalStack.length - 1]) {
-      removalStack.pop();
+    // end removal if stepped past the removing node
+    if (node.previousSibling === currentRemovingNode) {
+      currentRemovingNode = null;
```
justinfagnani: I think this needs to be the next node in the stack, in the case of nested removal.

sorvell: Added a little comment description at the top of that function. Don't think we care if it's nested since we only need to know when we're ending the removal state of a non-nested node:
* any part inside the removing node gets deactivated
* any part outside a non-nested removing node has its index adjusted

justinfagnani: Ah, that's right.

src/lib/modify-template.ts
```diff
@@ -0,0 +1,129 @@
...
+  let activeParts = parts.filter((part) => isTemplatePartActive(part));
+  let part = activeParts[partIndex];
+  let insertCount = 0;
+  let walkerIndex = -1;
+  while (walker.nextNode()) {
+    walkerIndex++;
+    const walkerNode = walker.currentNode as Element;
+    if (walkerNode === refNode) {
+      refNode.parentNode!.insertBefore(node, refNode);
+      insertCount = countNodes(node);
```
justinfagnani: I think you can reuse the same walker. Not sure if that'd be faster, but it might be a bit smaller:

```ts
walker.previousSibling();
while (walker.nextNode() !== refNode) {
  insertCount++;
}
```

src/lib/modify-template.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    while (part !== undefined && part.index === walkerIndex) {
```
justinfagnani: Once the node is inserted we don't have to walk the tree anymore, but can just modify the remaining part indices.

src/lib/modify-template.ts
```diff
@@ -0,0 +1,129 @@
...
+    throw 'Reference node must be inside template.';
+  }
+  const walker = document.createTreeWalker(
+      content,
+      walkerNodeFilter,
+      null as any,
+      false);
+  let partIndex = 0;
+  // only need to modify active parts.
+  let activeParts = parts.filter((part) => isTemplatePartActive(part));
```
justinfagnani: This check could be done in the while loop at line 124 and avoid creating another array.

src/lib/shady-render.ts
```diff
@@ -12,19 +12,29 @@
...
 export {html, svg, TemplateResult} from '../lit-html.js';
 
 declare global {
   interface Window {
     ShadyCSS: any;
   }
+  class ShadowRoot {}
 }
 
+const getTemplateKey = (type: string, scopeName: string) =>
```
justinfagnani: Consider a different name: `templateKey` sounds like it'd be based on the template contents... this is a key used to look up a cache.

src/lib/shady-render.ts
```diff
@@ -33,20 +43,106 @@ const shadyTemplateFactory = (scopeName: string) =>
...
         templateCache.set(result.strings, template);
       }
       return template;
     };
 
+
+const TEMPLATE_TYPES = ['html', 'svg'];
+
+/**
+ * Removes all style elements from Templates for hte given scopeName
```
justinfagnani: hte -> the

src/lib/shady-render.ts
```diff
@@ -33,20 +43,106 @@ const shadyTemplateFactory = (scopeName: string) =>
...
+      // only scope element template once per scope name
+      if (!shadyRenderSet.has(scopeName)) {
+        shadyRenderSet.add(scopeName);
+        const styleTemplate = document.createElement('template');
+        Array.from(fragment.querySelectorAll('style')).forEach((s: Element) => {
+          styleTemplate.content.appendChild(s);
+        });
+        window.ShadyCSS.prepareTemplateStyles(styleTemplate, scopeName);
+        // fix templates.
+        removeStylesFromLitTemplates(scopeName);
```
justinfagnani: Can you expand the comment to describes why the template cache contains all the templates we need to modify by this point?

src/lib/shady-render.ts
```diff
@@ -33,20 +43,106 @@ const shadyTemplateFactory = (scopeName: string) =>
...
+ * For the given scope name, ensures that ShadyCSS style scoping is performed.
+ * This is done just once per scope name so the fragment and template cannot
+ * be modified.
+ * (1) extracts styles from the rendered fragment and hands them to ShadyCSS
+ * to be scoped and appended to the document
+ * (2) removes style elements from all lit-html Templates for this scope name.
+ */
+const ensureStylesScoped =
+    (fragment: DocumentFragment, template: Template, scopeName: string) => {
+      // only scope element template once per scope name
```
justinfagnani: style super-nit: Let's standardize on capitalizing the comments

src/lib/shady-render.ts
```diff
@@ -33,20 +43,106 @@ const shadyTemplateFactory = (scopeName: string) =>
...
+      if (!shadyRenderSet.has(scopeName)) {
+        shadyRenderSet.add(scopeName);
+        const styleTemplate = document.createElement('template');
+        Array.from(fragment.querySelectorAll('style')).forEach((s: Element) => {
+          styleTemplate.content.appendChild(s);
+        });
+        window.ShadyCSS.prepareTemplateStyles(styleTemplate, scopeName);
+        // fix templates.
```
justinfagnani: nit: no period needed for non sentence

src/lib/shady-render.ts
```diff
@@ -33,20 +43,106 @@ const shadyTemplateFactory = (scopeName: string) =>
...
+        removeStylesFromLitTemplates(scopeName);
+        // ApplyShim case.
+        if (window.ShadyCSS.nativeShadow) {
+          const style = styleTemplate.content.querySelector('style');
+          if (style) {
```
justinfagnani: here and a couple of other places: I prefer explicit checks, like `style !== null`

src/test/lib/shady-render_test.ts
```diff
@@ -22,50 +22,144 @@ const assert = chai.assert;
...
           span {
-            color: blue;
+            border: 5px solid tomato;
           }
         </style>
         <span>Testing...</span>
       `}
-    `, container, 'scope-2');
+    `, container.shadowRoot as DocumentFragment, 'scope-2');
+    const div = (container.shadowRoot  as DocumentFragment).querySelector('div');
```
justinfagnani: the cast to DocumentFragment shouldn't be necessary

src/test/lib/shady-render_test.ts
```diff
@@ -22,50 +22,144 @@ const assert = chai.assert;
...
+    assert.equal(getComputedStyle(div as HTMLElement).getPropertyValue('border-top-width').trim(), '4px');
```
justinfagnani: the cast to `HTMLElement` shouldn't be necessary either, as `.querySelector('div')` is typed to return a `HTMLDivElement`

src/test/lib/shady-render_test.ts
```diff
@@ -22,50 +22,144 @@ const assert = chai.assert;
...
+        <style>
+          div {
+            border: ${border};
+          }
+        </style>
+        <div>Testing...</div>
+      `, container.shadowRoot as DocumentFragment, 'scope-3');
+    }
+    renderTemplate('1px solid black');
+    const div = (container.shadowRoot  as DocumentFragment).querySelector('div');
```
justinfagnani: some extra spaces in here, make sure to run `npm run format`

test/index.html
```diff
@@ -13,9 +13,12 @@
     <script type="module" src="./lib/unsafe-html_test.js"></script>
     <script type="module" src="./lib/async-append_test.js"></script>
     <script type="module" src="./lib/async-replace_test.js"></script>
+    <script type="module" src="./lib/modify-template_test.js"></script>
     <script>
       WCT.loadSuites([
         'shady.html',
+        'shady.html?shadydom=true',
```
justinfagnani: I keep forgetting about this... I wonder if we should do this for all the tests?

src/lib/modify-template.ts
sorvell: Since you can't change the walker root node, this doesn't seem worth trying since there'd be more code to re-use the same walker.

src/lib/modify-template.ts
sorvell: done

src/lib/modify-template.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/lib/shady-render.ts
sorvell: done

src/test/lib/shady-render_test.ts
sorvell: done

src/test/lib/shady-render_test.ts
sorvell: done

src/test/lib/shady-render_test.ts
sorvell: Had trouble with this: running the formatter resulted in a lot of diffs.

test/index.html
sorvell: let's revisit later.

PR #340, opened 2018-05-23
Update README.md

newyork-anthonyng:


newyork-anthonyng:
I signed it!

docs-src/guide/03-how-it-works.md
```diff
@@ -25,15 +25,15 @@ tag`My name is ${name}.`
 
 Tags are functions that take the literal strings of the template and values of the embedded expressions, and return a new value. This can be any kind of value, not just strings. lit-html returns an object representing the template, called a `TemplateResult`.
 
-The key features of template tags that lit-html utilizes to make updates fast is that the object holding the literals strings of the template is _exactly_ the same for every call to the tag for a particular template.
+The key feature of template tags that lit-html utilizes to make updates fast is that the object holding the literals string of the template is _exactly_ the same for every call to the tag for a particular template.
```
justinfagnani: "strings" should actually be plural here, as there's more than one, and it's a possessive, so: "literal's strings"

newyork-anthonyng: Updated 👍

PR #357, opened 2018-06-07
change directiveValue to noChange

fopsdev:
replaces:
https://github.com/Polymer/lit-html/pull/276


fopsdev:
screwed up lineendings...

PR #358, opened 2018-06-07
changed directiveValue to noChange. so returning a noChange in a comp…

fopsdev:
…onent will force lit to do nothing (more detailed explanation in https://github.com/Polymer/lit-html/pull/276)

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -437,6 +437,11 @@ const isDirective = (o: any) =>
  */
 export const noChange = {};
 
+/**
+ * @deprecated Use `noChange` instead.
+ */
+export { noChange as directiveValue };
+
 const isPrimitiveValue = (value: any) => value === null ||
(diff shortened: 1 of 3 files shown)


src/lit-html.ts
```diff
@@ -435,7 +435,7 @@ const isDirective = (o: any) =>
  * A sentinel value that signals that a value was handled by a directive and
  * should not be written to the DOM.
  */
-export const directiveValue = {};
+export const noChange = {};
```
justinfagnani: Add:

```ts
/**
 * @deprecated Use `noChange` instead.
 */
export {noChange as directiveValue};
```

so this isn't a breaking change

fopsdev: yeah makes sense, done

justinfagnani:
@fopsdev can you also add an entry to the changrlog?

PR #376, opened 2018-06-21
Handle case-sensitive attributes like viewBox.

### LINKED ISSUE #374: `viewBox` not recognized in lit-extended
When using the `$=` syntax to dynamically update the `viewBox` attribute of an SVG, the following error appears in the console and the DOM does not render:

> Uncaught TypeError: Cannot read property 'value' of null

[This Plunkr](https://next.plnkr.co/edit/CEjDaef5AR2I1Wra) concisely demonstrates the issue.

For reference, here is the SVG snippet:

```
<svg viewBox$="0 0 ${viewBoxSize} ${viewBoxSize}" width$="100px" height$="100px">
    ${ svg`<rect x="0" y="0" width$="${viewBoxSize}" height$="${viewBoxSize}"/>` }
</svg>
```

and the import for lit-extended:

```
import {html, svg, render} from 'https://unpkg.com/lit-html@0.10.2/lib/lit-extended.js';
```

This is related to #275 but as can be seen by the Plunkr it is a different issue.

justinfagnani:
Fixes #374

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -259,7 +259,14 @@ function findTagClose(str: string): number {
 export class TemplatePart {
   constructor(
       public type: string, public index: number, public name?: string,
-      public rawName?: string, public strings?: string[]) {
+      public strings?: string[]) {
+  }
+
+  /**
+   * @deprecated Use `name`, which is now also case-preserving.
+   */
+  get rawName(): string|undefined {
+    return this.name;
   }
 }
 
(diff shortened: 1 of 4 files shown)


intcreator:
Looking good so far!

src/test/lit-html_test.ts
```diff
@@ -324,6 +315,14 @@ suite('lit-html', () => {
         assert.equal(stripExpressionDelimeters(container.innerHTML), '<div foo="bar"></div>');
       });
 
+      test('renders a case-sensitive attribute', () => {
+        const size = 100;
+        render(html`<svg viewBox="0 0 ${size} ${size}"></svg>`, container);
+        assert.equal(
+            stripExpressionDelimeters(container.innerHTML),
+            '<svg viewBox="0 0 100 100"></svg>');
+      });
```
kevinpschaaf: For Edge test to pass, apparently need to strip the extra `xmlns` attribute when asserting.  And they helpfully render the innerHTML as a self-closing tag, apparently.  Sigh.
```
expected '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" />' to equal '<svg viewBox="0 0 100 100"></svg>'
```
Maybe just change it to 
```js
assert.match(stripExpressionDelimeters(container.innerHTML), /viewBox="0 0 100 100"/);
```

src/lit-html.ts
```diff
@@ -320,16 +327,21 @@ export class Template {
...
               lastAttributeNameRegex.exec(stringForPart)![1];
           // Find the corresponding attribute
-          // TODO(justinfagnani): remove non-null assertion
-          const attribute = attributes.getNamedItem(attributeNameInPart)!;
-          const stringsForAttributeValue = attribute.value.split(markerRegex);
+          // If the attribute name contains special characters, lower-case it
+          // so that on XML nodes with case-sensitive getAttribute() we can
+          // still find the attribute, which will have been lower-cased by
+          // the parser.
+          const attributeLookupName = /^[a-zA-z-]*$/.test(attributeNameInPart) ?
```
jridgewell: The `A-z` looks like a bug.

justinfagnani: Definitely. Nice catch

Issue #375, opened 2018-06-20
On Edge/IE11 input is stamping the expression placeholder as the value 


notwaldorf:
In a `LitElement`, if you are using an `<input value="${props._value}">` in your template and `props._value` is `undefined`, then Edge/IE stamp the expression placeholder instead of something good:
<img width="186" alt="screen shot 2018-06-20 at 12 51 14 pm" src="https://user-images.githubusercontent.com/1369170/41681167-aa611dcc-7488-11e8-86bd-740834c81037.png">

If you set a default value for `_value` in the element's constructor then it's fine.

Repro: https://lit-element-example-6nhwjs.stackblitz.io
(Works fine on Chrome/FF/Safari, borked on Edge/IE11)

bgotink:
```js
const tpl = document.createElement('template');
tpl.innerHTML = '<input value="foo">';
const input = document.importNode(tpl.content, true).firstElementChild;

input.removeAttribute('value');
console.log('value is now:');
console.log(JSON.stringify(input.value));
```

This logs the value as `""` for Chrome and `"foo"` for Edge. This explains why you get the expected result in your example in Chrome etc, but this actually masks a bug in `lit-html`:

Setting the `_value` to empty string in the constructor means the first render will call the `value` property setter on the input element. When this initial `_value` is `undefined`, this setter is not called because the `_equalToPreviousValues` of `AttributePart` returns `true`.
This is wrong, on a first render `_equalToPreviousValues` should never be `true`.

Repro in @notwaldorf's example: https://stackblitz.com/edit/lit-element-example-sqqeig. Note how the setter never gets called unless you uncomment the `this._value = '';` in the constructor.

Another result of the fact that `_equalToPreviousValues` returns `true` if the initial value is `undefined` is the following unexpected behaviour:

```js
import {render, html} from 'lit-html';

render(html`
  <div test="${undefined} is ${undefined}"></div>
`, document.body);

console.log(document.body.innerHTML);
```

logs

```
<div></div>
```

Repro: https://stackblitz.com/edit/lit-element-example-fqhsbb

justinfagnani:
I couldn't repro this anymore, so I think it was fixed by #377, but I added a test in #448

PR #377, opened 2018-06-24
Render initial undefined values

bgotink:
As mentioned in #375, if all values for an AttributePart or PropertyPart are `undefined`, the initial render will not set the attribute/property. See [this comment](https://github.com/Polymer/lit-html/issues/375#issuecomment-399745122) for an explanation as to why that's wrong.

Instead of adding an extra property to check for the initial render and introducing logic to update the property on the first render I simply added a check on the length of `this._previousValue`, as that array can only be empty for the initial render (the property is set to an empty array in the constructor; `setValue` always sets the property to a non-empty array).

### DIFF
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -501,6 +501,11 @@ export class AttributePart implements MultiPart {
   }
 
   protected _equalToPreviousValues(values: any[], startIndex: number) {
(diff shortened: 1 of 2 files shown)


justinfagnani:
Awesome fix, thanks!

PR #384, opened 2018-06-26
Dirty check primitive values passed to unsafeHTML()

### LINKED ISSUE #222: unsafeHTML gets rerendered every time

justinfagnani:
Fixes #222 and supersedes #247 

cc @daKmoR

### DIFF
--- a/src/lib/unsafe-html.ts
+++ b/src/lib/unsafe-html.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {directive, DirectiveFn, NodePart} from '../lit-html.js';
+import {directive, DirectiveFn, NodePart, _isPrimitiveValue} from '../lit-html.js';
 
 /**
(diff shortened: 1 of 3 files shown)


daKmoR:
uh very nice :) thank you :)

PR #385, opened 2018-06-26
Update values after upgrading custom elements

bgotink:


justinfagnani:
Awesome, this was next on my list :)

There's one more test that I'd like to add from user bug reports - make sure that property setters are called for custom elements inside of nested templates add _after_ the initial render.

You can use the same custom element class in your test, but do something like this:

```ts
const t = (value: any) => html`${value}`;
render(t('A'), container);

const fragment = html`<x-test-uses-property-setters value=${'foo'}></x-test-uses-property-setters>`;
render(t(fragment), container);
// asserts
```

bgotink:
Alright! I've added the extra test.

These tests only run in browsers where custom elements are natively supported, it is probably interesting to test them with the customElements shim as well. I don't have time for this atm, but I'd be willing to look into that later this week.

PR #386, opened 2018-06-27
Add test for dirty checking Node values.

### LINKED ISSUE #142: Add a test that node values do not get removed and reattached
Basically a test for #129

justinfagnani:
Fixes #142


justinfagnani:
ah, yeah, hopefully because cause IE doesn't have Promises. I changed the await so that hopefully we trigger the Promise polyfill to wait the right amount of time.

PR #393, opened 2018-07-09
Format

justinfagnani:


src/lit-html.ts
```diff
@@ -161,8 +161,7 @@ export type TemplateContainer = (Element|DocumentFragment)&{
 export function render(
     result: TemplateResult,
     container: Element|DocumentFragment,
-    templateFactory: TemplateFactory = defaultTemplateFactory
-  ) {
+    templateFactory: TemplateFactory = defaultTemplateFactory) {
```
bigopon: @justinfagnani May I know why this change ? It's trivial but nice to know what the argument is

justinfagnani: It's just how the formatter works. I don't argue with the formatter - the whole point is to completely remove formatting disagreements.

If there's a particularly egregious case around line breaks, it can usually be solved with a well-placed comment. Other than that, we take the bad with the good.

PR #398, opened 2018-07-11
New syntax and main module

### LINKED ISSUE #399: Change attribute/property binding syntax to require opt-in to properties.
The current syntax of lit-extended sets properties by default and requires opting out of properties to set attributes with the `$` suffix.

This has a number of downsides:

- The default of setting properties is different from plain HTML
- When you add a binding to an attribute it changes from an attribute to a properties
- There's no syntax indication that some behavior is different from HTML
- Case-sensitive attributes like `viewBox` are harder to implement
- The syntax isn't a super-set of lit-html core

This issue is to document the rationale for changing the syntax to set attributes by default and require opt-in to properties with a `.` prefix.

## Compison

Current syntax:

```js
html`<input id$=${id} value=${value} required?=${required} on-change=${onChange}>`
```

Proposed syntax:
```js
html`<input id=${id} .value=${value} required?=${required} on-change=${onChange}>`
```

And for consistency, we could go further and make all binding modifiers be single-symbol prefixes, like:
```js
html`<input id=${id} .value=${value} ?required=${required} @change=${onChange}>`
```

justinfagnani:
This PR adds a new syntax that's like lit-extended, but defaults to setting attributes and requires opting-in to setting properties.

Properties are set with a `.` prefix:

```js
html`<input .value=${value}>`
```

The new syntax is the main module of the package now. The unopinionated core has been moved to core.js and the new syntax is in lit-html.js

lit-extended is deprecated and will be removed before 1.0.

Fixes #399


nicolasparada:
I still think lit-html should try to set properties falling back to attributes without any special syntax. Same behavior as preact.

justinfagnani:
Let's keep discussion of the options to #399

docs-src/guide/01-getting-started.md
```diff
@@ -30,19 +30,19 @@ Modules are increasingly supported in JavaScript environments and are shipping i
...
 
 ```js
-import {html, render} from './node_modules/lit-html/lib/lit-extended.js';
+import {html, render} from './node_modules/lit-html/lit-html.js';
 ```
 
 The path to use depends on where you've installed lit-html to. Browsers only support importing other modules by path, not by package name, so without other tools involved, you'll have to use paths.
 
-If you use some tool than can convert package names into paths, then you can import by path:
+If you use a tool than converts package names into paths, then you can import by path:
```
bgotink: typo: `tha{n => t}`

docs/guide/getting-started.html
```diff
@@ -3,7 +3,6 @@
 
   <html>
     <head>
-      <title>Getting Started</title>
```
bgotink: All the titles are gone, is that on purpose? (I honestly don't know how the documentation works, this just caught my eye)

docs-src/guide/02-writing-templates.md
```diff
@@ -54,16 +47,41 @@ Follow these rules for well-formed templates:
 
 ## Binding Types
 
-Expressions can occur only in attribute value or text content positions, but in lit-extended a few different binding types use expressions in attribute value positions. Each binding types supports different types of values. In total, theses are the types of bindings:
-
- * Attribute bindings: All values are converted to strings
- * Property bindings: Any type of value
- * Event handler bindings: Event handler functions only
- * Text content bindings: TemplateResults, Nodes. Other types converted to strings.
+Expressions can occur text content or in attribute value positions.
```
sorvell: Expressions can occur *in* text content

src/lit-html.ts
```diff
@@ -12,846 +12,156 @@
...
-};
-
-/**
- * Renders a template to a container.
+ * Properties are set by prefixing an attribute name with `.`.
  *
- * To update a container with new values, reevaluate the template literal and
- * call `render` with the new result.
+ * Attribute names in lit-html templates preserve case, so properties are case
+ * sensitive. If an
```
sorvell: formatting

src/lit-html.ts
```diff
@@ -12,846 +12,156 @@
...
- * against the string literal directly preceding the expression, assuming that
- * the expression is in an attribute-value position.
+ * Example:
  *
- * See attributes in the HTML spec:
- * https://www.w3.org/TR/html5/syntax.html#attributes-0
+ *     html`<input .value=${value}>`
  *
- * "\0-\x1F\x7F-\x9F" are Unicode control characters
+ * Event handlers are sset by prefixing an attribute name with `on-`.
```
sorvell: typo: sset => set

Change `on-` to `@` right?

src/lit-html.ts
```diff
@@ -12,846 +12,156 @@
...
- */
-function findTagClose(str: string): number {
-  const close = str.lastIndexOf('>');
-  const open = str.indexOf('<', close + 1);
-  return open > -1 ? str.length : close;
-}
-
-/**
- * A placeholder for a dynamic expression in an HTML template.
+ *     html`<button on-click=${(e)=> this.onClickHandler(e)}>Buy Now</button>`
```
sorvell: `on-` to `@`

src/lit-html.ts
```diff
@@ -12,846 +12,156 @@
...
  *
- * There are two built-in part types: AttributePart and NodePart. NodeParts
- * always represent a single dynamic expression, while AttributeParts may
- * represent as many expressions are contained in the attribute.
+ * Boolean attributes are set by appending `?` to an attribute name.
```
sorvell: Change appending to prepending

src/lit-html.ts
```diff
@@ -12,846 +12,156 @@
...
- * A Template's parts are mutable, so parts can be replaced or modified
- * (possibly to implement different template semantics). The contract is that
- * parts can only be replaced, not removed, added or reordered, and parts must
- * always consume the correct number of values in their `update()` method.
+ * Example:
  *
- * TODO(justinfagnani): That requirement is a little fragile. A
- * TemplateInstance could instead be more careful about which values it gives
- * to Part.update().
+ *     html`<input type="checkbox" checked?=${value}>`
```
sorvell: `checked?` to `?checked`

src/lit-html.ts

src/lit-html.ts

src/lit-html.ts

src/lit-html.ts

src/lit-html.ts

justinfagnani:
@sorvell PTAL

geraldfullam:
Why did you [remove so much documentation from the README.md](https://github.com/Polymer/lit-html/pull/398/commits/c7b642333a8c98af254eac2f8cb06d2e9098e7f2)? (See deletions from Lines 61–427).

PR #400, opened 2018-07-14
Refactor AttributePart to only have a single value

justinfagnani:
This PR removes the SinglePart/MultiPart distinction. AttributeParts now only have a single value, but are associated with an AttributeCommitter that commits multiple values to an attribute.

This PR is based on #398.

The attribute handling code is a bit simpler in this factoring, though the "fast" path for attribute with a single value isn't implemented here, but it is implemented for PropertyCommitter. Not sure if this matters much.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -478,6 +478,10 @@ export interface Part {
   commit(): void;
 }
 
+/**
+ * Sets attribute values for AttributeParts, so that the value is only set once
+ * even if there are multiple parts for an attribute.
+ */
 export class AttributeCommitter {
   element: Element;
   name: string;
@@ -491,8 +495,15 @@ export class AttributeCommitter {
     this.strings = strings;
     this.parts = [];
     for (let i = 0; i < strings.length - 1; i++) {
-      this.parts[i] = new AttributePart(this);
+      this.parts[i] = this._createPart();
     }
+  }
+
+  /**
+   * Creates a single part. Override this to create a differnt type of part.
+   */
+  protected _createPart(): AttributePart {
+    return new AttributePart(this);
   }
 
   protected _getValue(): any {
--- a/src/core.ts
+++ b/src/core.ts
@@ -429,12 +429,12 @@ export class Template {
  * with certain DOM APIs, like textContent.
  */
 export const getValue = (part: Part, value: any) => {
-  // `null` as the value of a Text node will render the string 'null'
-  // so we convert it to undefined
   if (isDirective(value)) {
     value = value(part);
     return noChange;
   }
+  // `null` as the value of a Text node will render the string 'null'
+  // so we convert it to undefined
   return value === null ? undefined : value;
 };
 
@@ -580,7 +580,15 @@ export class NodePart implements Part {
   }
 
   setValue(value: any): void {
-    value = getValue(this, value);
+    if (isDirective(value)) {
+      this._pendingValue = noChange;
+      value(this);
+      return;
+    }
+    if (value === noChange) {
+      return;
+    }
+    value = value === null ? undefined : value;
     if (_isPrimitiveValue(value) && value === this._value) {
       value = noChange;
     }
--- a/src/core.ts
+++ b/src/core.ts
@@ -69,8 +69,8 @@ export class TemplateResult {
       // "...>...": text position. open === -1, close > -1
       // "...<...": attribute position. open > -1
       // "...": no change. open === -1, close === -1
-      isTextBinding = (close > -1 || isTextBinding) &&
-          s.indexOf('<', close + 1) === -1;
+      isTextBinding =
+          (close > -1 || isTextBinding) && s.indexOf('<', close + 1) === -1;
       html += isTextBinding ? nodeMarker : marker;
     }
     html += this.strings[l];
@@ -542,26 +542,25 @@ export class AttributeCommitter {
 export class AttributePart implements Part {
   committer: AttributeCommitter;
   _value: any = undefined;
+  _pendingValue: any = undefined;
 
   constructor(comitter: AttributeCommitter) {
     this.committer = comitter;
   }
 
   setValue(value: any): void {
-    value = getValue(this, value);
-    if (value === noChange) {
-      return;
-    }
-    if (_isPrimitiveValue(value)) {
-      if (value === this._value) {
-        return;
-      }
-    }
-    this._value = value;
-    this.committer.dirty = true;
+    this._pendingValue = value;
   }
 
   commit() {
+    let value = getValue(this, this._pendingValue);
+    if (value === noChange) {
+      return;
+    }
+    if (!_isPrimitiveValue(value) || value !== this._value) {
+      this._value = value;
+      this.committer.dirty = true;
+    }
     this.committer.commit();
   }
 }
--- a/src/core.ts
+++ b/src/core.ts
@@ -421,23 +421,6 @@ export class Template {
   }
 }
 
-/**
- * Returns a value ready to be inserted into a Part from a user-provided value.
- *
- * If the user value is a directive, this invokes the directive with the given
- * part. If the value is null, it's converted to undefined to work better
- * with certain DOM APIs, like textContent.
- */
-// export const getValue = (part: Part, value: any) => {
-//   if (isDirective(value)) {
-//     value = value(part);
-//     return noChange;
-//   }
-//   // `null` as the value of a Text node will render the string 'null'
-//   // so we convert it to undefined
-//   return value === null ? undefined : value;
-// };
-
 export interface DirectiveFn<P = Part> {
   (part: P): void;
   __litDirective?: true;
@@ -457,12 +440,7 @@ export const isDirective = (o: any) =>
  */
 export const noChange = {};
 
-/**
- * @deprecated Use `noChange` instead.
- */
-export {noChange as directiveValue};
-
-export const _isPrimitiveValue = (value: any) =>
+export const isPrimitive = (value: any) =>
     (value === null ||
      !(typeof value === 'object' || typeof value === 'function'));
 
@@ -471,7 +449,7 @@ export const _isPrimitiveValue = (value: any) =>
  * by lit-html.
  */
 export interface Part {
-  _value: any;
+  value: any;
 
   /**
    * Sets the current part value, but does not write it to the DOM.
@@ -522,7 +500,7 @@ export class AttributeCommitter {
       text += strings[i];
       const part = this.parts[i];
       if (part !== undefined) {
-        const v = part._value;
+        const v = part.value;
         if (v != null &&
             (Array.isArray(v) || typeof v !== 'string' && v[Symbol.iterator])) {
           for (const t of v) {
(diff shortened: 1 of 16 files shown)


src/core.ts
```diff
@@ -717,24 +709,28 @@ export class NodePart implements SinglePart {
...
-            instance, node as Element, templatePart.name!, templatePart.strings!
-        );
+        const comitter = new AttributeCommitter(node as Element, templatePart.name!, templatePart.strings!);
+        return comitter.parts;
       } else if (templatePart.type === 'node') {
-        return new NodePart(instance, node, node.nextSibling!);
+        return [new NodePart(instance, node, node.nextSibling!)];
       }
-      throw new Error(`Unknown part type ${templatePart.type}`);
+      throw [new Error(`Unknown part type ${templatePart.type}`)];
```
bgotink: Is this intentional?

src/core.ts
```diff
@@ -757,16 +753,16 @@ export class TemplateInstance {
...
-      if (!part) {
-        valueIndex++;
-      } else if (part.size === undefined) {
-        (part as SinglePart).setValue(values[valueIndex]);
-        valueIndex++;
-      } else {
-        (part as MultiPart).setValue(values, valueIndex);
-        valueIndex += part.size;
+      if (part !== undefined) {
+        part.commit();
```
bgotink: Won't the current setup lead to multiple commits for some attributes?

e.g.

```js
const template = (one, two) => html`<my-element some-attr="${one} ${two}"></my-element>`;

render(template('lorem', 'ipsum'), container);

// later on
render(template('dolor', 'sit'), container);
```

Won't that change the `some-attribute` to `"dolor sit"` twice? This loop commits every part, and both `AttributePart`s in the attribute commit their `AttributeCommitter` because they're both dirty.

A possible solution would be to mark all parts of an `AttributeCommitter` as not dirty when running `AttributeCommitter#commit()`

src/core.ts
justinfagnani: Yeah. I'll add some more comments, but now there's not a 1-1 correspondence between "template parts" and `Part`s. Maybe there's some further refinement there, but we need something with knowledge of a whole attribute.

src/core.ts
justinfagnani: This was a mistake, I was switching patterns and forgot to make committing only run once. I'll fix asap :)

justinfagnani: Fixed and added a test

justinfagnani:
@sorvell this is ready for review now. I like the shape of the API now, with the small exception of PartCallback returning an array of Parts. I didn't yet explore returning multiple TemplateParts for attributes. I can look into that or not... I don't think there will be _that_ many PartCallback implementations our there.

core.ts gets a _tiny_ bit smaller with this factoring, by about ~40 bytes.

src/lit-html.ts
```diff
@@ -57,26 +57,28 @@ export const svg = (strings: TemplateStringsArray, ...values: any[]) =>
...
+        Part[] => {
+          if (templatePart.type === 'attribute') {
+            const name = templatePart.name!;
+            const prefix = name[0];
+            if (prefix === '.') {
+              const comitter = new PropertyCommitter(
+                  node as Element,
+                  templatePart.name!.slice(1),
+                  templatePart.strings!);
+              return comitter.parts;
```
sorvell: It seems somewhat weird that you make `AttributePart`s for properties, but I'm not sure this is worth addressing, wdyt?

src/core.ts
```diff
@@ -470,112 +470,109 @@ export const noChange = {};
...
-      if (v && v !== noChange &&
-          (Array.isArray(v) || typeof v !== 'string' && v[Symbol.iterator])) {
-        for (const t of v) {
-          // TODO: we need to recursively call getValue into iterables...
-          text += t;
+      const part = this.parts[i];
+      if (part !== undefined) {
+        const v = part._value;
+        if (v != null &&
+            (Array.isArray(v) || typeof v !== 'string' && v[Symbol.iterator])) {
```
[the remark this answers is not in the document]
justinfagnani: I think this TODO was mostly incorrect, and fully incorrect now that this code is in AttributeCommitter. The values here are after AttributePart has already handled them, passing them to getValue. Directive handling is the biggest thing we get out of calling getValue, and arrays of directive don't make sense, since there's no individual part to pass to the directive. TemplateResults don't make sense for attribute parts - using them here would be essentially toString'ing a DOM tree into an attribute.

src/core.ts
```diff
@@ -757,16 +757,16 @@ export class TemplateInstance {
   }
 
   update(values: any[]) {
-    let valueIndex = 0;
+    let i = 0;
+    for (const part of this._parts) {
+      if (part !== undefined) {
+        part.setValue(values[i]);
```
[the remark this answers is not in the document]
justinfagnani: Yes, right now all attribute changes will apply before children changes. That shouldn't affect any well behaved elements, and won't affect anything that uses batching. I guess we can change this so it goes in DOM order.

src/lit-html.ts
justinfagnani: I actually do this we should add a trivial PropertyPart subclass of AttributePart just in case there's a need for a difference in behavior in the future and we want that change to be less likely to break anyone.

I'll add this.

src/core.ts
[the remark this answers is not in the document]
justinfagnani: Actually, it looks a bit tricky to defer the work in `_setValue()`

justinfagnani:
@sorvell added PropertyPart. PTAL

src/core.ts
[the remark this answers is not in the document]
sorvell: Still concerns me a bit. For example, events happen in `commit` but nodes are done in `setValue` an event listener installed on a parent wouldn't see an event fired by a child custom element at connect time.

src/core.ts
bgotink: It just seems weird to me to be throwing an `Array<Error>`, especially because there's no special handling of this array anywhere in lit-html itself.

justinfagnani: oops, I didn't even see that it was an array of Error, I read this as a line returning multiple parts! Thanks!

src/lit-html.ts
```diff
@@ -85,79 +87,122 @@ export const partCallback =
...
       return;
     }
-    if (s.length === 2 && s[0] === '' && s[1] === '') {
-      // An expression that occupies the whole attribute value will leave
-      // leading and trailing empty strings.
-      value = getValue(this, values[startIndex]);
-    } else {
-      // Interpolation, so interpolate
-      value = this._interpolate(values, startIndex);
+    this._value = !!value;
```
keanulee: This line should be after the dirty check.

src/lit-html.ts
justinfagnani: This is done btw

src/lit-html.ts
```diff
@@ -85,79 +87,135 @@ export const partCallback =
...
+          'Boolean attributes can only contain a single expression');
+    }
+    this.element = element;
+    this.name = name;
+    this.strings = strings;
+  }
+
+  setValue(value: any): void {
+    this._pendingValue = value;
+    // value = !!value;
```
justinfagnani: will remove

src/core.ts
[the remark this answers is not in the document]
sorvell: Note, this has been addressed since NodePart now applies changes to DOM in `commit`.

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
-    for (let i = startIndex; i < startIndex + this.size; i++) {
-      if (this._previousValues[i] !== values[i] ||
-          !_isPrimitiveValue(values[i])) {
-        return false;
+  }
+}
+
+export class AttributePart implements Part {
+  committer: AttributeCommitter;
+  _value: any = undefined;
```
sorvell: Seems wrong that this is `_value` since directives can and do access it. (Same for NodePart).

src/lit-html.ts
sorvell: remove cruft

src/core.ts
```diff
@@ -428,15 +428,15 @@ export class Template {
...
-export const getValue = (part: Part, value: any) => {
-  // `null` as the value of a Text node will render the string 'null'
-  // so we convert it to undefined
-  if (isDirective(value)) {
-    value = value(part);
-    return noChange;
-  }
-  return value === null ? undefined : value;
-};
+// export const getValue = (part: Part, value: any) => {
```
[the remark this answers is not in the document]
justinfagnani: done

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
  */
 export {noChange as directiveValue};
 
-export const _isPrimitiveValue = (value: any) => value === null ||
-    !(typeof value === 'object' || typeof value === 'function');
+export const _isPrimitiveValue = (value: any) =>
```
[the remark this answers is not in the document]
justinfagnani: Renamed to `isPrimitive`

src/core.ts
justinfagnani: Renamed to `value`

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
+export class AttributePart implements Part {
+  committer: AttributeCommitter;
+  _value: any = undefined;
+
+  constructor(comitter: AttributeCommitter) {
+    this.committer = comitter;
+  }
+
+  setValue(value: any): void {
+    value = value === null ? undefined : value;
```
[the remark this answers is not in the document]
justinfagnani: This was inlined from `getValue()`, but I think it's only necessary on NodePart. Testing without it.

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
+    if (value !== noChange || !_isPrimitiveValue(value) || value !== this._value) {
+      this._value = value;
+      if (!isDirective(value)) {
```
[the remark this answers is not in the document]
justinfagnani: done

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
-      // leading and trailing empty strings.
-      value = getValue(this, values[startIndex]);
-      if (Array.isArray(value)) {
-        value = value.join('');
-      }
-    } else {
-      value = this._interpolate(values, startIndex);
+  commit() {
+    // console.log('commit', this._value);
+    while (isDirective(this._value)) {
```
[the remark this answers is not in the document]
justinfagnani: done

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
+    this._pendingValue = value;
+  }
+
+  commit() {
+    while (isDirective(this._pendingValue)) {
+      const directive = this._pendingValue;
+      this._pendingValue = noChange;
+      directive(this);
+    }
+    const value = this._pendingValue === null ? undefined : this._pendingValue;
```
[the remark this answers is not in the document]
justinfagnani: `textContent = null` would render "null", so this conversion was done in `getValue()` previous. Moved into `_setText()`

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
     if (value === noChange) {
       return;
     }
     if (_isPrimitiveValue(value)) {
-      // Handle primitive values
-      // If the value didn't change, do nothing
-      if (value === this._previousValue) {
-        return;
+      if (value !== this._value) {
+        this._setText(value);
```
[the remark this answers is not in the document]
justinfagnani: Changed

src/lit-html.ts

src/core.ts
```diff
@@ -462,126 +462,153 @@ export const noChange = {};
...
-    this._previousValue = undefined;
   }
 
   setValue(value: any): void {
-    value = getValue(this, value);
+    this._pendingValue = value;
+  }
+
+  commit() {
+    while (isDirective(this._pendingValue)) {
```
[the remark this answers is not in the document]
justinfagnani: Possibly a little, maybe on the base class. I'd like to do another refactoring attempt later as well.

PR #402, opened 2018-07-15
WIP:  Work with Attr nodes to fix invalid attr names

jridgewell:
This is an attempt to fix getting/setting attributes with invalid XML names, eg `class$` or `.test`.

Needs more work and manual testing... But IE seems to be fine parsing these attrs and then allowing manipulation through the `Attr` node.

Fixes https://github.com/Polymer/lit-html/issues/295.

### DIFF
--- a/src/lib/lit-extended.ts
+++ b/src/lib/lit-extended.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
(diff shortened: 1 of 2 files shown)


justinfagnani:
@jridgewell in #344 I tried a different technique with attributes: to build up a Map of name->attributes for the bound attributes. This should allow us to get around any restrictions on  getAttribute() or getNamedItem(): https://github.com/Polymer/lit-html/pull/344/files#diff-7548be97e171131880082c935a962f68R325

PR #404, opened 2018-07-16
Handle forms as Node instead of iterable

### LINKED ISSUE #390: Iterable element handling is broken
The `<form>` element does not handled as a `Node`, because it is also iterable.

The problem seems to be here: https://github.com/Polymer/lit-html/blob/v0.10.2/src/lit-html.ts#L563

bgotink:
Fixes #390


justinfagnani:
Awesome! Thanks @bgotink

PR #405, opened 2018-07-19
Replace PartCallback with TemplateProcessor

justinfagnani:
Also changes from passing a TemplateInstance down a tree of NodeParts to just passing the TemplateFactory, and _not_ passing down the TemplateInstance's partCallback, since that needs to come from the TemplateResults being rendered anyway. This was a bug before.


justinfagnani:
cc @bgotink

src/core.ts
```diff
@@ -377,7 +362,7 @@ export class Template {
                 (strings[i] === '') ? document.createComment('') :
                                       document.createTextNode(strings[i]),
                 node);
-            this.parts.push(new TemplatePart('node', index++));
+            this.parts.push({type: 'node', index: index++});
```
sorvell: What does removing the `TemplatePart` class do to code size?

src/core.ts
```diff
@@ -631,13 +616,13 @@ export class NodePart implements Part {
...
-    const template = this.instance._getTemplate(value);
+    const template = this.templateFactory(value);
     let instance: TemplateInstance;
     if (this._value && this._value.template === template) {
       instance = this._value;
     } else {
-      instance = new TemplateInstance(
-          template, this.instance._partCallback, this.instance._getTemplate);
+      instance =
+          new TemplateInstance(template, value.processor, this.templateFactory);
```
sorvell: Maybe add a comment about using the `processor` from the `value` since this was an issue before?

src/core.ts
```diff
@@ -32,6 +32,23 @@ export const html = (strings: TemplateStringsArray, ...values: any[]) =>
 export const svg = (strings: TemplateStringsArray, ...values: any[]) =>
     new SVGTemplateResult(strings, values, 'svg');
 
+/**
+ * Creates Parts when a template is instantiated.
+ */
+export class TemplateProcessor {
```
sorvell: Document these methods?

src/core.ts
justinfagnani: I don't remember the delta on this specific part, but it was smaller by a few bytes. I like not using a class for this internal storage too. Object literals are minutely faster.

justinfagnani:
This TemplateProcessor has a different API, as far as we know, but we don't really know the Template Instantiation API yet. I think the two slot into the same concept though in terms of customizing template behavior. We just don't know which set of callbacks the spec will eventually have.

src/core.ts
justinfagnani: ack

src/core.ts
justinfagnani: ack

justinfagnani:
@sorvell PTAL

PR #409, opened 2018-07-21
quick fix for typo in docs. Fixes #408

### LINKED ISSUE #408: Typo in docs
There's a typo in the docs in the guide. Just creating this issue to track the little PR I made to fix it.

In 01-getting-started.md, changing the last paragraph to:

```
The only true solution is to have one canonical version of a library that all other libraries import. Since modules support is rolling out to browsers now, and modules are very well supported by tools, it makes sense for that format to be modules.
```

claytongulick:


jsilvermist:
A problem with this way of saying it is that it will be out of date and requiring another update as soon as the latest version is released. Maybe it would be better to instead call out which version it applies/doesn't apply to, without directly referring to what's currently released?

claytongulick:
@jsilvermist I agree - but I wasn't 100% sure what the next version is going to be - is it going to be 1.0?

jsilvermist:
Not sure, but I was assuming 0.11.0 from previous conversations I've seen regarding the changes.

justinfagnani:
I committed the typo fix separately.

Thanks for pointing it out @claytongulick!

PR #413, opened 2018-07-25
Hoist template preparers to avoid excessive closures

jridgewell:
Instead of recreating the `_prepareTemplate` and `_prepareInstance`
helpers on every template creation, this hoists them to prototype
functions and passes in the necessary params. It's a bit verbose, but it
should help.

Also, it avoids recreating the `TreeWalker` on in every nested
`<template>`. Any constructor that touches WebIDL is dog slow. Reuse the
same instance all the way down the recursion.

Try viewing this with [whitespace changes hidden](https://github.com/Polymer/lit-html/pull/413/files?w=1).

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -281,151 +281,162 @@ export class Template {
 
   constructor(result: TemplateResult, element: HTMLTemplateElement) {
     this.element = element;
-    let index = -1;
-    let partIndex = 0;
     const nodesToRemove: Node[] = [];
+    // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be
+    // null
+    const walker = document.createTreeWalker(
+        element.content,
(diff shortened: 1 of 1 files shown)


justinfagnani:
@jridgewell I thought I had replied to this already, sorry!

I'm a little unsure about this change. Nested templates should be a pretty-rare edge case that we don't need to aggressively optimize for, and we're trying to save every byte now, so the extra verbosity might cost there. On the other hand, making these prototype methods means we don't create the closure even on the top-level/non-nested cases. I do wish I had my benchmark runner in place already.

I just submitted a massive refactoring though (which should make you happy I hope :) ), so this will need to be rebased at the least.

justinfagnani:
Going to close this now as pretty far out of date with the current code. We very well may want to do this refactor in the future though.

PR #415, opened 2018-07-27
Add NodePart attachment methods

justinfagnani:
Adds `NodePart#insertAfterNode`, `NodePart#appendIntoPart`, `NodePart#insertAfterPart` to abstract away a little of the marker node management. Cleans up some of the code in `NodePart#_setIterable`.

This brings the size of core.js to just under 2.5k

Note that I didn't put much work into cleaning up repeat(), but left a comment explaining why.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -53,12 +53,10 @@ export class TemplateProcessor {
 
   /**
    * Create parts for a text-position binding.
-   *
-   * @param node The previous sibling of the binding
    * @param templateFactory
    */
-  handleTextExpression(node: Node, templateFactory: TemplateFactory) {
-    return new NodePart(node, node.nextSibling!, templateFactory);
+  handleTextExpression(templateFactory: TemplateFactory) {
+    return new NodePart(templateFactory);
   }
 }
 
@@ -221,6 +219,10 @@ export function render(
   instance.update(result.values);
 }
 
+// Allows `document.createComment('')` to be renamed for a
+// small manual size-savings.
+const createMarker = () => document.createComment('');
+
 /**
  * An expression marker with embedded unique key to avoid collision with
  * possible text in templates.
@@ -379,14 +381,14 @@ export class Template {
           // These nodes are also used as the markers for node parts
           for (let i = 0; i < lastIndex; i++) {
             parent.insertBefore(
-                (strings[i] === '') ? document.createComment('') :
+                (strings[i] === '') ? createMarker() :
                                       document.createTextNode(strings[i]),
                 node);
             this.parts.push({type: 'node', index: index++});
           }
           parent.insertBefore(
               strings[lastIndex] === '' ?
-                  document.createComment('') :
+                  createMarker() :
                   document.createTextNode(strings[lastIndex]),
               node);
           nodesToRemove.push(node);
@@ -407,7 +409,7 @@ export class Template {
           const previousSibling = node.previousSibling;
           if (previousSibling === null || previousSibling !== previousNode ||
               previousSibling.nodeType !== Node.TEXT_NODE) {
-            parent.insertBefore(document.createComment(''), node);
+            parent.insertBefore(createMarker(), node);
           } else {
             index--;
           }
@@ -417,7 +419,7 @@ export class Template {
           // We don't have to check if the next node is going to be removed,
           // because that node will induce a new marker if so.
           if (node.nextSibling === null) {
-            parent.insertBefore(document.createComment(''), node);
+            parent.insertBefore(createMarker(), node);
           } else {
             index--;
           }
@@ -572,16 +574,46 @@ export class AttributePart implements Part {
 
 export class NodePart implements Part {
   templateFactory: TemplateFactory;
-  startNode: Node;
-  endNode: Node;
+  startNode!: Node;
+  endNode!: Node;
   value: any = undefined;
   _pendingValue: any = undefined;
 
-  constructor(
-      startNode: Node, endNode: Node, templateFactory: TemplateFactory) {
+  constructor(templateFactory: TemplateFactory) {
     this.templateFactory = templateFactory;
-    this.startNode = startNode;
-    this.endNode = endNode;
+  }
+
+  /**
+   * Inserts this part between `ref` and `ref`'s next sibling. Both `ref` and
+   * its next sibling must be static, unchanging nodes such as those that appear
+   * in a literal section of a template.
+   *
+   * This part must be empty, as its contents are not automatically moved.
+   */
+  insertAfterNode(ref: Node) {
+    this.startNode = ref;
+    this.endNode = ref.nextSibling!;
+  }
+
+  /**
+   * Appends this part into a parent part.
+   *
+   * This part must be empty, as its contents are not automatically moved.
+   */
+  appendIntoPart(part: NodePart) {
+    part._insert(this.startNode = createMarker());
+    part._insert(this.endNode = createMarker());
+  }
+
+  /**
+   * Appends this part after `ref`
+   *
+   * This part must be empty, as its contents are not automatically moved.
+   */
+  insertAfterPart(ref: NodePart) {
+    ref._insert(this.startNode = createMarker());
+    this.endNode = ref.endNode;
+    ref.endNode = this.startNode;
   }
 
   setValue(value: any): void {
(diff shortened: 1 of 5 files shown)


src/core.ts
```diff
@@ -424,7 +428,7 @@ export class Template {
 export const getValue = (part: Part, value: any) => {
   // `null` as the value of a Text node will render the string 'null'
   // so we convert it to undefined
-  if (isDirective(value)) {
+  if (value != null && value.__litDirective === true) {
```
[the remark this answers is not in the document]
sorvell: A directive doesn't have to be a function anymore?

src/core.ts
```diff
@@ -42,8 +42,8 @@ export class TemplateProcessor {
     return comitter.parts;
   }
 
-  handleTextExpression(node: Node, templateFactory: TemplateFactory) {
-    return new NodePart(node, node.nextSibling!, templateFactory);
+  handleTextExpression(templateFactory: TemplateFactory) {
```
sorvell: Must we remove the node argument here? Without that does this server any purpose? What can you do?

src/core.ts
```diff
@@ -774,8 +795,9 @@ export class TemplateInstance {
           partIndex++;
         } else if (nodeIndex === part.index) {
           if (part.type === 'node') {
-            this._parts.push(
-                this.processor.handleTextExpression(node, this._getTemplate));
+            const part = this.processor.handleTextExpression(this._getTemplate);
+            part.insertAfterNode(node);
```
sorvell: Could we make `handleTextExpression` responsible for this?

src/test/core_test.ts
```diff
@@ -1009,11 +1009,13 @@ suite('Core', () => {
 
     setup(() => {
       container = document.createElement('div');
-      startNode = document.createTextNode('');
-      endNode = document.createTextNode('');
+      startNode = document.createComment('');
```
sorvell: Is this change important?

src/test/core_test.ts
```diff
@@ -1051,6 +1053,8 @@ suite('Core', () => {
 
       test('accepts arrays', () => {
         part.setValue([1, 2, 3]);
+        console.log(Array.from(container.childNodes)
```
sorvell: remove logging

test/index.html
```diff
@@ -15,7 +15,7 @@
     <script type="module" src="./lib/async-append_test.js"></script>
     <script type="module" src="./lib/async-replace_test.js"></script>
     <script type="module" src="./lib/modify-template_test.js"></script>
-    <script type="module">
+    <!-- <script type="module">
```
sorvell: replace these tests?

src/test/core_test.ts
```diff
@@ -1254,6 +1264,58 @@ suite('Core', () => {
...
+            assert.equal(testPart.endNode, testEndNode);
+            const text = document.createTextNode('');
+            testPart.setValue(text);
+            assert.deepEqual(
+                Array.from(container.childNodes),
+                [startNode, endNode, text, testEndNode]);
+          });
+    });
+
+    suite('appendIntoPart', () => {
```
sorvell: Test that setting the parent part still works as expected.

src/test/core_test.ts
```diff
@@ -1254,6 +1264,58 @@ suite('Core', () => {
...
+              startNode,
+              testPart.startNode,
+              text,
+              testPart.endNode,
+              endNode
+            ]);
+          });
+    });
+
+    suite('insertAfterPart', () => {
```
sorvell: Test that setting the previous part still works as expected.

src/core.ts
```diff
@@ -550,15 +551,45 @@ export class AttributePart implements Part {
...
+    this.startNode = ref;
+    this.endNode = ref.nextSibling!;
+  }
+
+  /**
+   * Appends this part into a parent part.
+   *
+   * This part must be empty, as its contents are not automatically moved.
+   */
+  appendIntoPart(part: NodePart) {
```
sorvell: Definitely not a slam dunk, but these could be more like their element counterparts as `appendPart`, `insertPartBefore`... wdyt?

src/core.ts
```diff
@@ -636,53 +667,43 @@ export class NodePart implements Part {
...
       partIndex++;
     }
 
-    if (partIndex === 0) {
-      this.clear();
-      this._value = undefined;
-    } else if (partIndex < itemParts.length) {
-      const lastPart = itemParts[partIndex - 1];
-      // Truncate the parts array so _previousValue reflects the current state
+    if (partIndex < itemParts.length) {
```
sorvell: Can you explain why the other part of the conditional went away?

src/core.ts
```diff
@@ -201,6 +201,10 @@ export function render(
   instance.update(result.values);
 }
 
+// Allows `document.createComment('')` to be renamed for a
+// small manual size-savings.
+const createMarker = () => document.createComment('');
```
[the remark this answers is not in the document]
justinfagnani: I hope this is trivial to inline by VMs. Closure will at least.

src/core.ts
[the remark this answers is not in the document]
justinfagnani: I felt the multiple checks were redundant. `value.__litDirective` should be a good enough signal.

src/core.ts
justinfagnani: Many we can leave it but not use it? This is just separating part creation from attachment.

src/core.ts
justinfagnani: I like `handleTextExpression` not knowing about the assumption that we're attaching to a marker node like this.

test/index.html
justinfagnani: oops

src/core.ts
justinfagnani: I went around on the names a bit and I still think these are the most clear.

src/core.ts
justinfagnani: I just realized that I could inline the conditional into the call to `clear()`

justinfagnani:
@sorvell PTAL

src/lib/async-replace.ts
```diff
@@ -55,6 +53,7 @@ export const asyncReplace =
             // previous value display until we can replace it.
             if (i === 0) {
               part.clear();
+              itemPart.appendIntoPart(part);
```
sorvell: Why wait to append this?

src/core.ts
```diff
@@ -668,54 +700,44 @@ export class NodePart implements Part {
...
       partIndex++;
     }
 
-    if (partIndex === 0) {
-      this.clear();
-      this.value = undefined;
-    } else if (partIndex < itemParts.length) {
-      const lastPart = itemParts[partIndex - 1];
-      // Truncate the parts array so _previousValue reflects the current state
+    if (partIndex < itemParts.length) {
```
sorvell: Can you explain how the previous code is equivalent to this? it's a little unclear.

src/lib/async-replace.ts
justinfagnani: It's no longer sharing the start and end node with the parent part, so if we clear the parent after appending, we'll nuke this part. I think this is easier to understand than the previous sharing of markers. It's a little more intuitive that if a part contains a child part and you clear the parent, you'll remove the child.

src/core.ts
justinfagnani: the case where `partIndex < itemParts.length` already covered `partIndex === 0`, and the critical operation there was to clear. `clear()`'s argument is optional, so `itemPart && itemPart!.endNode` will === `undefined` when `partIndex === 0` and clear the whole part

sorvell: ok, makes sense.

src/lib/async-replace.ts
sorvell: Are there more nodes now?

sorvell: And is it worth documenting (at least internally) how the marker nodes are setup between parts?

justinfagnani: This is one reason for always using the attachment methods rather than having calling code manipulate the markers directly. Between insert/append/clear, most code should not have to care about them.

justinfagnani: and I do think the "contains" semantic between parent and child parts is way more understandable here. It would have been weird if you cleared a parent and a single child part wasn't removed, or if you had multiple parts they were just broken.

daKmoR:
guys I just wanted to thank you that you do these discussions in the "open" - helps me understand the changes and as I would have had similar question it's nice that they are already answered :)

thxxx :)

PR #416, opened 2018-07-27
Simplify close-tag finding logic.

justinfagnani:
Saves 18 bytes ;)

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -61,11 +61,16 @@ export class TemplateResult {
     for (let i = 0; i < l; i++) {
       const s = this.strings[i];
       html += s;
-      // We're in a text position if the previous string closed its tags.
-      // If it doesn't have any tags, then we use the previous text position
-      // state.
-      const closing = findTagClose(s);
-      isTextBinding = closing > -1 ? closing < s.length : isTextBinding;
+      const close = s.lastIndexOf('>');
+      const open = s.indexOf('<', close + 1);
+      // We're in a text position if the previous string closed its last tag, an
+      // attribute position if the string opened an unclosed tag, and unchanged
+      // if the string had no brackets at all:
+      //
+      // "...>...": text position. open === -1, close > -1
+      // "...<...": attribute position. open > -1
+      // "...": no change. open === -1, close === -1
+      isTextBinding = open === -1 && (close > -1 || isTextBinding);
       html += isTextBinding ? nodeMarker : marker;
     }
     html += this.strings[l];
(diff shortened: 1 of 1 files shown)


src/core.ts
```diff
@@ -64,8 +64,9 @@ export class TemplateResult {
       // We're in a text position if the previous string closed its tags.
       // If it doesn't have any tags, then we use the previous text position
       // state.
-      const closing = findTagClose(s);
-      isTextBinding = closing > -1 ? closing < s.length : isTextBinding;
+      const close = s.lastIndexOf('>');
+      const open = s.indexOf('<', close + 1);
+      isTextBinding = open === -1 && (close > -1 || isTextBinding);
```
sorvell: Let's add a little comment visualization if possible. Not sure this helps but something like:
```
// when string ends with:
// text: ... > ... 
// not text: ... < ...
// take last value: ... or ... < ...
```

sorvell: Does flipping it around work and make it smaller?
`isAttributeBinding = open > -1 || isAttributeBinding;`

justinfagnani: I made a truth table and did the logical reduction, so I think this is minimal.

justinfagnani: Updated the comment

justinfagnani:
@sorvell PTAL

src/core.ts
```diff
@@ -61,11 +61,16 @@ export class TemplateResult {
...
+      const close = s.lastIndexOf('>');
+      const open = s.indexOf('<', close + 1);
+      // We're in a text position if the previous string closed its last tag, an
+      // attribute position if the string opened an unclosed tag, and unchanged
+      // if the string had no bracets at all:
+      //
+      // "...>...": text position. open === -1, close > -1
+      // "...<...": attribute position. open > -1
+      // "...": no change. open === -1, close === -1
+      isTextBinding = open === -1 && (close > -1 || isTextBinding);
```
jridgewell: If we want to get fancy, we can avoid the forward scan in some cases:

```js
const close = s.lastIndexOf('>');
if (close === -1) {
  if (isTextBinding) {
    isTextBinding = s.indexOf('<', close + 1) === -1
  } else {
    // Woop. We were in an attributes, and we didn't close it.
    // No need for forward scan.
    isTextBinding = false;
  }
} else {
  isTextBinding = s.indexOf('<', close + 1) === -1;
}

// Compressed is just a flip of what you already wrote.
const close = s.lastIndexOf('>');
isTextBinding = (close > -1 || isTextBinding) && s.indexOf('<', close + 1) === -1

Comment (justinfagnani) [reply src/core.ts]: There's a runtime perf / file size balancing game we're playing right now. I'm hoping that the `close` index is generally close enough to the end of the string that both the backward and forward scan is cheap, and we can justify saving a few bytes here.

Comment (jridgewell) [reply src/core.ts]: This compresses better than the new code, should save 2 bytes. 😉

```js
// Yours:
var b=a.lastIndexOf(">");c=-1===a.indexOf("<",b+1)&&(-1<b||c)

// Mine:
var b=a.lastIndexOf(">");c=(-1<b||c)&&a.indexOf("<",b+1)
```

It might have gotten lost in the code I posted (edited it to add more new lines), but the compressed (not if-statement) code is at the bottom:

```js
// Compressed is just a flip of what you already wrote.
const close = s.lastIndexOf('>');
isTextBinding = (close > -1 || isTextBinding) && s.indexOf('<', close + 1) === -1
```

That one's faster _and_ smaller.

jridgewell: Edit, I forgot the comparison in my compiled output. Updated the comment.

PR #421, opened 2018-08-02
Rework `repeat` to support reparenting nodes with a custom keymap

gkjohnson:
This PR adds a fourth, optional argument to the `repeat` directive so you can provide shared a keyMap and reparent nodes with the same ids across different containers. The `cleanMap` function has also been modified and exposed so the map can be cleaned manually when all the rendering that uses the shared elements has been done.

#### Uses
- If a repeated template is particularly large and needs to be moved to a new container, (I believe) using lit to do that at the moment results in the old element being disposed while creating that large dom tree _again_ to recreate it in the new spot, which is unnecessarily slow. This allows the expensive element to just be re-parented.
- If the repeated data has any internal state then it will be lost if it's moved to a new container. That means that manually added styles, internal web-component state, or canvas draw contexts will be invalid or lost when the data is rerendered with new dom in another container.

I've run into these issues in the past with drag-and-drop interfaces and moving data representations and views between columns and it can be _really_ hard to work around cleanly if the library doesn't support it well.

#### Example Use
```js
import { render, html } in '.../lit-html.js';
import { repeat, cleanMap } in '.../lib/repeat.js';

const customMap = new Map();
const sharedTemplate = o => html`...${ o.value }...`;

render(html`
   ...
   ${ repeat(items, o => o.id, sharedTemplate) }
   ${ repeat(items2, o => o.id, sharedTemplate) }
   ...
`, document.body)

cleanMap(customMap);
```

And an html file showing the new behavior:
[shared-repeat-example.zip](https://github.com/Polymer/lit-html/files/2251886/shared-repeat-example.zip)

Note that the element stays red after being moved.

#### Things to keep in mind
- If `cleanMap` is not used appropriately or ever then you could get a memory leak but that's only the case when using this advanced functionality.
- For this to work properly the template used in the repeat must be from the same function -- copy and pasting the exact same string between the repeat functions will not work. I imagine it's a quirk in the way that template tag literals operate so I don't know if anything can be done about it:
```js
// good
const func = o => html`<li>${ o.val }</li>`
html`
${ repeat(items, o => o.id, func, customMap) }
${ repeat(items2, o => o.id, func, customMap) }
`

// bad
html`
...
${ repeat(items, o => o.id,  html`<li>${ o.val }</li>`, customMap) }
${ repeat(items2, o => o.id,  html`<li>${ o.val }</li>`, customMap) }
...
`
```

Maybe people have some thoughts on how to address those quirks, but at the least I'd think that good documentation and best practices would be enough.

Thanks!

### DIFF
--- a/src/lib/repeat.ts
+++ b/src/lib/repeat.ts
@@ -19,10 +19,12 @@ export type ItemTemplate<T> = (item: T, index: number) => any;
 
 const keyMapCache = new WeakMap<NodePart, Map<any, NodePart>>();
 
-function cleanMap(part: NodePart, key: any, map: Map<any, NodePart>) {
-  if (!part.startNode.parentNode) {
-    map.delete(key);
-  }
+export function cleanMap(keyMap: Map<any, NodePart>) {
+  keyMap.forEach((part: NodePart, key: any, map: Map<any, NodePart>) => {
+    if (!part.startNode.parentNode) {
+      map.delete(key);
+    }
+  });
 }
 
 export function repeat<T>(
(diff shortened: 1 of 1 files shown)


gkjohnson:
I signed it!

aadamsx:
I reviewed the code, this is a good PR and should be accepted.  Seems useful.  If not accepted, then an explination as to why should be given out of respect for peoples time and effort. :)

justinfagnani:
@gkjohnson: @kevinpschaaf is rewriting repeat() to use a similar algorithm to Vue and snabdom, and might or might not include key map sharing. I can't remember right now. So just FYI that this PR probably won't apply on that. Thanks for working this out!

gkjohnson:
Thanks for the update! It sounds like the repeat function is going to get a bit more complex. If this feature isn't otherwise supported in the rewrite I can take a look at re-adding it. If you remember ping me in the PR and I'll gladly take a look!

Thanks!

justinfagnani:
@gkjohnson the new repeat() landed a little while back. I'll close this for now, and we can discuss in an issue how cache sharing might work with the new algorithm.

PotterDai:
Hi, I have encountered a similar situation . @justinfagnani has this feature (shared keymap for reparenting between different containers) been included in the new version of repeat? Thank you very much! = )

PR #431, opened 2018-08-10
Update index.html

cvializ:
Correct `TempalteResult` to `TemplateResult`


cvializ:
Ah my mistake, I'll update that file instead. I was too quick with the drive-by PR 😄

PR #436, opened 2018-08-18
Split implementation into smaller modules. Remove core/lit-html distinction.

### LINKED ISSUE #434: Merge core.js and lit-html.js

justinfagnani:
This PR is much easier to review commit-by-commit. For the most part I've tried not to update any code, comments, lets-labels, etc., so that it's just code movement. Where I've moved tests I've sometimes created new suites though.

All tests should pass at every commit.

Fixes #434


src/test/lit-html_test.ts
```diff
@@ -122,7 +121,7 @@ suite('lit-html', () => {
       assert.equal(fooDiv, barDiv);
     });
 
-    test(
+    test.skip(
```
sorvell: Let's bring this test back.

rollup.config.js
```diff
@@ -0,0 +1,26 @@
...
+import {terser} from 'rollup-plugin-terser';
+
+export default {
+  input: 'index.js',
+  output: {
+    file: 'index.bundled.js',
+    format: 'esm',
+  },
+  onwarn(warning) {
+    if (warning.code !== 'CIRCULAR_DEPENDENCY') {
```
justinfagnani: @rictic does Closure support circular dependencies?

rollup.config.js
```diff
@@ -0,0 +1,26 @@
+import filesize from 'rollup-plugin-filesize';
```
justinfagnani: I'll add a license

rollup.config.js
rictic: It supports some simple cases I tested, but I don't know how firm their guarantees of correctness are. We might want to test it out

PR #438, opened 2018-08-18
Add guard directive

LarsDenBakker:
This adds a guard directive. I know there is a lot of refactoring going on in https://github.com/Polymer/lit-html/pull/436, but with this PR this can already be reviewed. I can rebase later.

The implementation is really simple, it just dirty checks the expression and if it changed since the previous render it sets the value. 

Dirty checking for objects is done on identity/pointer, so immutable data patterns need to be used for updating. Unlike the polymer data system, there is no 'escape' like doing `this.set('foo.bar')` or `notifyPath()`. We could add a 'force' parameter, but maybe see if we can get away without that first.

### DIFF
--- a/src/directives/guard.ts
+++ b/src/directives/guard.ts
@@ -1,0 +1,46 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
@justinfagnani Thinking about this more, should we make the value lazy as well? Then you could do things like:

```
html`${guard(user,  () => html`${user.someProperty}`}`
```

Knowing that the template will not be evaluated until the user object is defined, allowing you safely access it.

ruphin:
The implementation you submitted is already lazy, or am I missing something?

LarsDenBakker:
Heh, you're right. Disregard what I said :)

justinfagnani:
@LarsDenBakker #436 is submitted, so please rebase to master now.

PR #440, opened 2018-08-21
Add ifDefined() directive

justinfagnani:
For AttributeParts, sets the attribute if the value is defined and removes the attribute if the value is undefined.

For other part types, this directive is a no-op.

Example:

```js
const render = () => html`
  <div class=${ifDefined(className)}></div>
`;
```

### DIFF
--- a/docs-src/guide/02-writing-templates.md
+++ b/docs-src/guide/02-writing-templates.md
@@ -244,7 +244,7 @@ const render = () => html`
 
 lit-html includes a few directives:
 
-#### `repeat(items, keyfn, template)`
+### `repeat(items, keyfn, template)`
 
 A loop that supports efficient re-ordering by using item keys.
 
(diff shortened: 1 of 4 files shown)


LarsDenBakker:
Fixes https://github.com/Polymer/lit-html/issues/437 :)

justinfagnani:
@sorvell I considered `ifExists` that would check for `null` and `undefined`. I don't think a falsey check is a good idea. Falsy is too loose, `0` and `''` fall into it.

justinfagnani:
I think `isDefined` is fine, personally. Missing properties are `undefined`, non-present keys in Maps, too. Anything more complicated would be better served with a predicate. We could add a second argument to this directive for a predicate function.

ruphin:
Is the desired behaviour when you have multiple conditionals that it only removes the attribute if both are not defined? I think that's what the current implementation does.

```javascript
html`<div class="${ifDefined(className)} ${ifDefined(otherClassName}"></div>`
```

Maybe add a test for that case.

ernsheong:
This does fix the unintentional attribute but at the cost of making lit-html less developer friendly. Can we default to not setting the attribute if undefined without the need for this directive?

justinfagnani:
@sorvell the most important thing here is the correct logic for setting and removing the attribute. Users can transform custom values they want to force a remove into undefined to reuse this directive. They can easily make a custom directive based on that too.

justinfagnani:
Got a verbal LGTM from @sorvell. Merging.

PR #441, opened 2018-08-21
Fix bug in AttributePart setValue

ruphin:
There's a bug in your `setValue()` function in `AttributePart`

This statement always evaluates to true, which made the conditional always evaluate the true branch:

```
value !== noChange || !isPrimitive(value)
```

Clearly the intent is to do nothing if `value === noChange` (or if ` isPrimitive(value) && value === this.value`)

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -118,7 +118,7 @@ export class AttributePart implements Part {
   }
 
   setValue(value: any): void {
-    if (value !== noChange || !isPrimitive(value) || value !== this.value) {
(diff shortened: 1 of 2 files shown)


justinfagnani:
Nice catch. We need a regression test for this though.

ruphin:
I added a test, but I'm not able to build the TS so I had to write it blind. It may fail, but it should be easy to fix for someone who can actually build and run the tests.

justinfagnani:
It passed, but does `npm run build`, or `npm test` not work for you?

PR #442, opened 2018-08-21
Upgrade disconnected custom elements before setting properties on them.

justinfagnani:
This keeps us from creating instance properties that alias assessors on disconnected elements.

This technique only works if `customElements.upgrade` is implemented. It's in Chrome, the polyfill, and Safari TP, and Firefox nightly.

There's two other ways we could solve this. First we either:
1. Patch TreeWalker properly in ShadyDOM
2. Don't use TreeWalker to traverse, so that we don't traverse ShadyDOM ShadowRoots created synchronously in constructors.

Then we go back to `document.importNode(template.content, true)` to clone & upgrade in one step.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -13,13 +13,11 @@
  */
 
 
+import {isCEPolyfill} from './dom.js';
 import {Part} from './parts.js';
 import {TemplateFactory} from './template-factory.js';
 import {TemplateProcessor} from './template-processor.js';
 import {isTemplatePartActive, Template} from './template.js';
-
-const isCEPolyfill = window.customElements !== undefined &&
-    (window.customElements as any).polyfillWrapFlushCallback !== undefined;
 
 /**
  * An instance of a `Template` that can be attached to the DOM and updated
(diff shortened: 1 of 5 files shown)


justinfagnani:
Tests are failing in Safari 10 as expected. This shows the problem with this approach...

The other solutions are much more invasive though

ruphin:
I think the proper solution is to defer the responsibility of 'fixing' instance properties to the Custom Element. This is required anyway for Custom Elements to function with deferred upgrades (through e.g. lazy imports).

Any Custom Element baseclass should have this bit of code in the constructor:

```
Object.getOwnPropertyNames(this).forEach(property => {
      const propertyValue = this[property];
      delete this[property];
      this[property] = propertyValue;
    });

**Comment by justinfagnani** [comment]
@sorvell PTAL

Changed to call `document.importNode` in the native custom elements case, where elements cannot spawn children during upgrade, and to force upgrade after instance prep in the polyfill case.

File: src/lib/template-instance.ts [hunk 1]
```diff
@@ -103,6 +108,18 @@ NodeFilter.SHOW_TEXT */
       }
     };
     _prepareInstance(fragment);
+
+    // Since we cloned in the polyfill case, now force an upgrade
+    if (isCEPolyfill) {
+      document.adoptNode(fragment);
+      customElements.upgrade(fragment);
```
sorvell: We should only do this if it's not connected since otherwise it's duplicate work and crawling trees is costly. I think it's fine to merge this and just put a TODO here for now.

justinfagnani: Just fixed it here

Issue #422, opened 2018-08-03
Render null or undefined does not clear render container


LarsDenBakker:
When rendering null or undefined to a container which has previously rendered content, the container is not cleared. 

Example:

```
render(html`Foo`, document.body);
render(null, document.body);
```

Will keep 'foo' rendered, and throw an error because the template is a non-optional property. When using lit-html directly, this is not really a big use case. However it makes sense for web components wanting clear out dom nodes under certain conditions. 

See: https://stackblitz.com/edit/lit-element-example-kz5y68 

Clicking the button should clear the dom.

An example use case for us is that we have a lot of dropdown menus in our application. I want to render them when opened, but when closed I want to remove them from the dom and not pollute the dom tree.

ruphin:
A very simple solution to your problem is to return a value that renders nothing, instead of rendering undefined or null.
```javascript
// Just do this
return html``;
return '';

// Instead of this
return null;
```

Testing for `undefined`/`null` adds run-time overhead and is not necessary because there is an acceptable alternative.

LarsDenBakker:
Thi incurs the cost of creating a template element for rendering nothing. That seems wasteful.

ruphin:
The difference is that checking for `null`/`undefined` creates overhead for _every_ render _everywhere_, whereas the additional overhead for creating an empty template only happens once, and only occurs when rendering the empty template. It also adds more code to the lib, which increases the payload size for every project using it.

```javascript
html``
```
is really cheap. After the first evaluation it basically returns a static object. If you're using this pattern in a lot of places you could even optimise it into a single template with something like

```javascript
export const nothing = () => html``;
```
and using
```javascript
return nothing();
```

The performance you gain from this will be extremely minimal. I expect the performance difference between implementing a check for `null`/`undefined` and using empty template will be significantly less than 1ms. Performance optimisation needs to account payload size in addition to run-time performance, and in this case the trade-off makes very little sense.

It is useful feedback that doing nothing on `null`/`undefined` is confusing, and I think this should be resolved through documentation.

PR #445, opened 2018-08-23
Rewrite render() to use a NodePart

justinfagnani:
This is mostly just to reuse some logic and try to not repeat ourselves in render() and NodePart#_commitTemplateResult.

This has the side-effect of allowing more value types to render(), which I'm so-so on as it doesn't seem like a huge use case, though it does solve #422.

shady-render is tricky as usual. I included a commit that rewrites it, but it has to patch a NodePart :/ I didn't bother making all the shady tests pass because of how ugly that is. An option there would be to let NodePart#_insert be overridable and subclass it with a ShadyNodePart.

### DIFF
--- a/src/lib/render.ts
+++ b/src/lib/render.ts
@@ -12,14 +12,13 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {isCEPolyfill, removeNodes} from './dom.js';
+import {removeNodes} from './dom.js';
+import {NodePart} from './parts.js';
 import {templateFactory as defaultTemplateFactory, TemplateFactory} from './template-factory.js';
-import {TemplateInstance} from './template-instance.js';
 import {TemplateResult} from './template-result.js';
+import {createMarker} from './template.js';
 
-export type TemplateContainer = (Element|DocumentFragment)&{
-  __templateInstance?: TemplateInstance;
-};
+const parts = new WeakMap<Node, NodePart>();
 
 /**
  * Renders a template to a container.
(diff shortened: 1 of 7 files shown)


src/lib/render.ts
```diff
@@ -39,24 +38,17 @@ export function render(
...
-    return;
+  let part = parts.get(container);
+  if (part === undefined) {
+    removeNodes(container, container.firstChild);
+    part = new NodePart(templateFactory);
+    const startNode = createMarker();
+    const endNode = createMarker();
+    container.appendChild(startNode);
+    container.appendChild(endNode);
+    part.insertAfterNode(startNode);
```
sorvell: Should we add `appendIntoNode` to make this the mucking with markers go away?

src/lib/shady-render.ts
```diff
@@ -142,39 +140,37 @@ const ensureStylesScoped =
...
-  const host = container instanceof ShadowRoot ? container.host : undefined;
-
-  // If there's a shadow host, do ShadyCSS scoping...
-  if (host !== undefined && verifyShadyCSSVersion()) {
-    ensureStylesScoped(fragment, template, scopeName);
-    window.ShadyCSS.styleElement(host);
+  let part = parts.get(container);
+  if (part === undefined) {
+    part = new NodePart(templateFactory);
+    (part as any)._insert = function _insert(node: Node) {
```
sorvell: I don't see why not to make this public and make `ShadyNodePart`. Perhaps we should make a render factory or some other way to eliminate the duplicated code in shady-render.

justinfagnani:
Closing in favor of #491

PR #446, opened 2018-08-23
 Don’t break templates when HTML comments have bindings in them

### LINKED ISSUE #411: Dynamic parts inside HTML comments are not recognised
Dynamic parts inside HTML comments are not recognised, causing the following insertion points to misalign.

### Example
The following code renders the value '1' into the `<div>` when it should render '2'.

```javascript
const templateWithComment = html`<!-- ${1} --><div>${2}</div>`;
render(templateWithComment, container)
```

### Live reproduction
https://codepen.io/ruphin/pen/MBpJXv?editors=1010

### Discussion

The issue exists because comment nodes are not parsed for the marker, and only checked against the comment-marker specifically. I see two ways to resolve this problem:

#### Parse comment nodes for the marker

This adds a lot of runtime parsing which is obviously bad. Also need a way to differentiate between the comment marker and an empty comment with the marker in it.

#### Make `findTagClose` emit a special case when we are inside a comment

Probably a better option, but it will cause overhead even when there are no comment nodes which is unfortunate.

justinfagnani:
Fixes #411

### DIFF
--- a/src/lib/template.ts
+++ b/src/lib/template.ts
@@ -127,39 +127,46 @@ export class Template {
                   document.createTextNode(strings[lastIndex]),
               node);
           nodesToRemove.push(node);
-        } else if (
-            node.nodeType === 8 /* Node.COMMENT_NODE */ &&
-            node.nodeValue === marker) {
-          const parent = node.parentNode!;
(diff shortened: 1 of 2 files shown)


ruphin:
I think this won't work for comments with multiple binding parts

PR #447, opened 2018-08-23
Handle Symbol values

### LINKED ISSUE #427: Rendering Symbols breaks NodePart
Symbols are considered 'primitive' values, so NodePart will execute `_setText()` which will effectively execute `document.createTextNode(Symbol())`. 

This throws the following error:
`Uncaught TypeError: Cannot convert a Symbol value to a string`

justinfagnani:
Fixes #427


justinfagnani:
It's because the other option is just crashing. This seems a lot better.

PR #448, opened 2018-08-23
Fix style attribute bindings in IE

### LINKED ISSUE #450: Bindings in style attribute broken in IE
Because IE rewrites the style attribute.

justinfagnani:
Fixes #450

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -14,7 +14,7 @@
 
 import {reparentNodes} from './dom.js';
 import {defaultTemplateProcessor, TemplateProcessor} from './template-processor.js';
-import {marker, nodeMarker} from './template.js';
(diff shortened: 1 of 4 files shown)


ruphin:
Creative solution to the IE11 style problem. I like it :)

justinfagnani:
@sorvell tests passing now, PTAL

src/lib/template-result.ts
```diff
@@ -55,6 +55,13 @@ export class TemplateResult {
       // "...": no change. open === -1, close === -1
       isTextBinding =
           (close > -1 || isTextBinding) && s.indexOf('<', close + 1) === -1;
+
+      // IE does a bad. Work around it.
```
sorvell: A better explanation would help =)

PR #452, opened 2018-08-25
Small directive optimization

ruphin:
Just a small change, I think this variant is a little cleaner since it doesn't use 'private' properties on objects.

This saves 21 bytes in the library size (before minification and compression). It increases runtime memory consumption by a few bytes: it adds one `WeakMap` instance to the memory pool.

I found no measurable performance difference.

Someone who understands TypeScript should probably check the `any` type on the WeakMap key. I'm sure it's not correct, but I couldn't find anything else that would compile. I can update the PR with the correct type.

----

An alternative option is to use Set or WeakSet instead of a WeakMap, but I believe WeakMap performance is slightly better than Set, and WeakSet would require another polyfill for IE11.

The odds of a directive ever leaving memory once created is pretty much zero, so Set is probably the most appropriate to use here. It would probably also save additional bytes. I can benchmark again with Set to verify that there is no impact.

### DIFF
--- a/src/lib/directive.ts
+++ b/src/lib/directive.ts
@@ -14,15 +14,15 @@
 
 import {Part} from './parts.js';
 
+const directives = new WeakMap<any, Boolean>();
+
 export interface Directive<P = Part> {
   (part: P): void;
-  __litDirective?: true;
 }
 
 export const directive = <P = Part>(f: Directive<P>): Directive<P> => {
-  f.__litDirective = true;
+  directives.set(f, true);
   return f;
 };
 
-export const isDirective = (o: any) =>
-    typeof o === 'function' && o.__litDirective === true;
+export const isDirective = (o: any) => directives.has(o);


ruphin:
I ran a bunch of benchmarks on all browsers, and WeakMap is faster than Set in all of them, so this is probably the best we can do.

src/lib/directive.ts
```diff
@@ -14,15 +14,15 @@
 
 import {Part} from './parts.js';
 
+const directives = new WeakMap<any, Boolean>();
```
justinfagnani: Too bad IE11 doesn't have WeakSet :/

justinfagnani:
> The odds of a directive ever leaving memory once created is pretty much zero, so Set is probably the most appropriate to use here.

This actually isn't true. Directives are values and usually will be functions returned from a directive factory, so their lifetime is actually going to be very short. I wonder if a wrapper will be better so that instead of modifying the hidden class of the directive functions, we create a short-lived object in the young generation on generational GCs.

justinfagnani:
> Someone who understands TypeScript should probably check the any type on the WeakMap key. I'm sure it's not correct, but I couldn't find anything else that would compile. I can update the PR with the correct type.

`any` is fine here, as this is just a arbitrary bit we're setting, and `isDirective` would have to change it's parameter type in order to use `Directive`.

The only better typing we can do is to type `isDirective` as a custom type guard so that TS can narrow the type (but this doesn't really matter):

```ts
export const isDirective = (o: any): o is Directive => directives.has(o);
```

src/lib/directive.ts
ruphin: I considered it, but didn't use it for that reason. Also, I don't think using WeakSet vs WeakMap will have an overall performance impact, and using WeakMap may actually be beneficial because it compresses better (as it is used in multiple locations).

PR #455, opened 2018-08-28
Update properties before upgrading custom elements (Fixes #453)

### LINKED ISSUE #453: Custom elements should be connected after being updated
There's currently a difference between `TemplateResult`'s processed via `render` and those handled by `NodePart` in a "nested" `TemplateResult`. The former updates prior to being attached to the DOM, but the latter does not. 

This can cause an issue if a custom element processes state in the `connectedCallback`. Elements made with `Polymer` and `LitElement` (prior to version `0.6`) do this, and when used in a nested `TemplateResult` these elements render with default state and then update with the state set by `lit-html`. This is inconsistent with non-nested `TemplateResult`'s and less efficient. 

The fix is simple, just update the instance before https://github.com/Polymer/lit-html/blob/master/src/lib/parts.ts#L267.

ruphin:
This will set the values in templates before elements are upgraded.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -249,14 +249,13 @@ export class NodePart implements Part {
 
   private _commitTemplateResult(value: TemplateResult): void {
     const template = this.templateFactory(value);
-    let instance: TemplateInstance;
     if (this.value && this.value.template === template) {
-      instance = this.value;
+      this.value.update(value.values);
     } else {
       // Make sure we propagate the template processor from the TemplateResult
       // so that we use it's syntax extension, etc. The template factory comes
       // from the render function so that it can control caching.
-      instance =
+      const instance =
           new TemplateInstance(template, value.processor, this.templateFactory);
       const fragment = instance._clone();
       // Since we cloned in the polyfill case, now force an upgrade
@@ -264,10 +263,10 @@ export class NodePart implements Part {
         document.adoptNode(fragment);
         customElements.upgrade(fragment);
       }
+      instance.update(value.values);
       this._commitNode(fragment);
       this.value = instance;
     }
-    instance.update(value.values);
   }
 
   private _commitIterable(value: any): void {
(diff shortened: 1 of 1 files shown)


src/lib/parts.ts
```diff
@@ -249,16 +249,16 @@ export class NodePart implements Part {
...
-    let instance: TemplateInstance;
     if (this.value && this.value.template === template) {
-      instance = this.value;
+      this.value.update(value.values);
     } else {
       // Make sure we propagate the template processor from the TemplateResult
       // so that we use it's syntax extension, etc. The template factory comes
       // from the render function so that it can control caching.
-      instance =
+      const instance =
```
justinfagnani: This should come after elements are upgraded, right before `this._commitNode(fragment);`. That should make the tests pass.

ruphin: A render test in Firefox is still breaking. I'm looking into it

justinfagnani: We don't run Sauce tests on PRs, so FF is the only configuration that tests the polyfills right now. It's probably also failing on Edge on Sauce.

ruphin:
I found the problem; we were not upgrading fragments in parts that were not connected, but when we update parts before connecting nodes, this causes nested elements to not upgrade.

ruphin:
Looking into this more closely I feel like we can also change `TemplateInstance._clone()` to return a DocumentFragment that has already been imported. I think every consumer of `_clone()` more or less immediately imports the fragment and/or upgrades the elements now, so we get some nice deduplication by incorporating this into `_clone()`

ruphin:
Fixes #453

@justinfagnani ptal :)

justinfagnani:
> Looking into this more closely I feel like we can also change TemplateInstance._clone() to return a DocumentFragment that has already been imported.

That's a good idea :)

ruphin:
I'll make a new PR that implements the suggested change

Issue #451, opened 2018-08-24
_commitPromise is inefficient


ruphin:
There's a problem with `_commitPromise`; because it changes `this.value`, it will cause all forms of "caching" for other commits to break. Because of this things like the current `until` directive implementation are really inefficient.

The `until` directive does this:

```js
part.setValue(defaultContent);
part.commit();
part.setValue(promise);
```

Imagine that `defaultContent` is a `Node`. Right now, the node is removed from DOM and re-inserted every time the directive is processed, because `part.setValue(promise)` changes `part.value` which causes the next `part.setValue(defaultContent)` to fail to detect that it already rendered this node.

This happens regardless if the Promise is resolved or not.

Essentially, `_commitPromise` breaks the `this.value` API by setting it to a value that does not represent the currently rendered value, which then breaks the other commit functions that rely on that API.

It can be solved by using a `this.promise`-like variable and use that instead of `this.value` for promises.

I can make a PR for this.

ruphin:
The solution is actually quite awkward, because you need to somehow know if something has been rendered _after_ rendering the promise, and the only reliable way to do that is by checking if `this.value` has changed. Alternatively, all other `commit` functions must set `this.promise` to `undefined` to make sure the promise is inert when it resolves. Either way

ruphin:
I'm working on a PR for this

ruphin:
I made a PR that fixes the issue with Promises. After looking at issue #272 (Promises for AttributeParts) I realise that it's effectively possible to implement a Directive that wraps Promises, and solves the issue entirely.

So effectively, there is an alternative solution. We can change the `render` / `setValue` API to no longer accept Promises, and support Promise rendering through a Directive instead. Something like this:

```js
import {promise} from 'lit-html/directives/promise.js';

const p = Promise.resolve('value');

render(html`<div>${promise(p)}</div>`, document.body);
```

Downside: less flexibility in rendering, if you throw in a Promise it will render `[object Promise]`.
Upsides: Easier implementation. Fixes the issue for all Part types.

justinfagnani:
Closed by #555

PR #458, opened 2018-08-29
Improve promise performance

ruphin:
This is a possible fix for #451

I added some comments to explain what is going on. Basically, promises are cached in `this._promise` instead of `this.value` inside `NodePart`, so they don't break cached results.

The unfortunate side-effect of this is that each `commit` needs to clear `this._promise` which is a small performance hit (~1-2% last I checked).

Let me know if this is a good direction, and I will clean up the commit, and add some regression tests.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -147,6 +147,7 @@ export class NodePart implements Part {
   startNode!: Node;
   endNode!: Node;
   value: any = undefined;
+  _promise: Promise<any> | undefined = undefined;
   _pendingValue: any = undefined;
 
   constructor(templateFactory: TemplateFactory) {
@@ -212,10 +213,13 @@ export class NodePart implements Part {
       this._commitIterable(value);
     } else if (value.then !== undefined) {
       this._commitPromise(value);
+      // Return so we do not clear this._promise
+      return;
     } else {
       // Fallback, will render the string representation
       this._commitText(value);
     }
+    this._promise = undefined;
   }
 
   private _insert(node: Node) {
@@ -319,9 +323,12 @@ export class NodePart implements Part {
   }
 
   private _commitPromise(value: Promise<any>): void {
-    this.value = value;
+    this._promise = value;
     value.then((v: any) => {
-      if (this.value === value) {
+      if (this._promise === value) {
+         // This is not strictly necessary, but avoids additional commits
+         // if the value is `noChange`.
+        this._promise = undefined;
         this.setValue(v);
         this.commit();
       }


src/lib/parts.ts
```diff
@@ -319,9 +323,12 @@ export class NodePart implements Part {
   }
 
   private _commitPromise(value: Promise<any>): void {
-    this.value = value;
+    this._promise = value;
     value.then((v: any) => {
-      if (this.value === value) {
+      if (this._promise === value) {
+         // This is not strictly necessary, but avoids additional commits
+         // if the value is `noChange`.
```
justinfagnani: On a re-render after the promise resolves, won't this keep us from checking the Promise identity and make us always call .then()? It seems like we want to only call .then() if the identity of the Promise has changed - treating it almost like a primitive.

ruphin: After the promise resolves, a regular call to `commit(<value from promise>)` will always clear out `this._promise`. That line is there to make sure this also happens when `<value from promise>` is `noChange`, because that is the only possible value skips the reset and returns early (other than a directive).

We have to reset `this._promise` when we render another value, to avoid the following problem sequence:

1. User commits Promise A -> nothing renders
2. User commits some other value -> other value renders
3. Promise A resolves -> result of Promise A renders

Essentially, if we don't reset `this._promise`, unresolved promises will trigger renders after the user rendered something else, which is clearly not desirable.

Because Promises no longer affect `this.value`, we can freely call `then()` on the promise and commit the result, because if the promise already resolved and the value does not change, then the commit will also be cached.

I looked into other ways of solving the problem, but this is the one of the few that avoids all the race conditions that can happen (there are many) between different renders and promises resolving. It's also easily the least complex solution.

One optimisation that can be made is an additional `if (this._promise !== value)` before `value.then`, which ensures that we only queue one `then` if we render multiple times before the promise resolves, but the current construction makes sure that any additional queued `then`s are aborted immediately with `if (this._promise === value)` because the first `then` to resolve will reset `this._promise`. On IE11 where you need Polyfills for Promises, this additional check will probably be beneficial, but on platforms that support Promises it costs one additional equality check, plus the extra bytes.

ruphin: After the promise resolves, a regular call to commit(<value from promise>) will always clear out this._promise. That line is there to make sure this also happens when <value from promise> is noChange, because that is the only possible value skips the reset and returns early (other than a directive).

We have to reset this._promise when we render another value, to avoid the following problem sequence:

User commits Promise A -> nothing renders
User commits some other value -> other value renders
Promise A resolves -> result of Promise A renders
Essentially, if we don't reset this._promise, unresolved promises will trigger renders after the user rendered something else, which is clearly not desirable.

Because Promises no longer affect this.value, we can freely call then() on the promise and commit the result, because if the promise already resolved and the value does not change, then the commit will also be cached.

I looked into other ways of solving the problem, but this is the one of the few that avoids all the race conditions that can happen (there are many) between different renders and promises resolving. It's also easily the least complex solution.

One optimisation that can be made is an additional if (this._promise !== value) before value.then, which ensures that we only queue one then if we render multiple times before the promise resolves, but the current construction makes sure that any additional queued thens are aborted immediately with if (this._promise === value) because the first then to resolve will reset this._promise. On IE11 where you need Polyfills for Promises, this additional check will probably be beneficial, but on platforms that support Promises it costs one additional equality check, plus the extra bytes.

justinfagnani:
@ruphin please check out issue #550. If we remove Promises all this can go into the `until` directive.

ruphin:
I commented on that issue. I don't think it is possible to solve the problem like this, unless you are willing to concede that after rendering anything with `until`, you cannot render a non-`until` value without breaking.

NodePart _must_ set or remove some property on every render in order for the Promise resolver to know if it has been superseded or not. This is a fundamental problem, and moving the Promise resolver logic to a directive does not change this.

justinfagnani:
Closed by #555

Issue #462, opened 2018-08-30
Version 0.11 fails on very old Chrome versions


sorvell:
Old Chrome versions do not support calling `has` with a primitive value on `WeakMap`. To fix this we need a `typeof` check at https://github.com/Polymer/lit-html/blob/master/src/lib/directive.ts#L28.

PR #463, opened 2018-08-30
Update sauce browser testing list

azakus:
Add Chrome 41, Edge 17, and Safari 11


azakus:
Chrome 41 is a known fail because of #462

PR #466, opened 2018-08-31
Rough template polyfill implementation

zeroliu:


src/lib/template-result.ts
```diff
@@ -68,8 +74,19 @@ export class TemplateResult {
   }
 
   getTemplateElement(): HTMLTemplateElement {
-    const template = document.createElement('template');
-    template.innerHTML = this.getHTML();
+    let template;
+    if (needTemplate) {
+      template = contentDoc.createElement('div') as any;
```
justinfagnani: Can we still use "template" as the tag name? That'll get rid of the need for the cast too.

src/lib/template-result.ts
```diff
@@ -68,8 +74,19 @@ export class TemplateResult {
...
+      template.content = contentDoc.createDocumentFragment();
+      const body = contentDoc.body;
+      body.innerHTML = this.getHTML();
+      while (body.firstChild) {
```
justinfagnani: we have a reparentNodes function here: https://github.com/Polymer/lit-html/blob/master/src/lib/dom.ts#L24

src/lib/template-result.ts
```diff
@@ -16,6 +16,12 @@ import {reparentNodes} from './dom.js';
 import {defaultTemplateProcessor, TemplateProcessor} from './template-processor.js';
 import {lastAttributeNameRegex, marker, nodeMarker, rewritesStyleAttribute} from './template.js';
 
+const needTemplate = typeof HTMLTemplateElement === 'undefined';
+const contentDoc = document.implementation.createDocument(
```
justinfagnani: Do this setup work only when needed

src/lib/template-result.ts
zeroliu: Using template tag name might make testing a bit complicated. After deleting HTMLTemplateElement from window, `document.createElement('template')` still returns a proper HTMLTemplateElement. I am not confident if writing tests like this would be able to represent the real world scenarios. Any ideas on how to cleanly remove template element from the browser?

src/lib/template-result.ts
zeroliu: done

src/lib/template-result.ts
zeroliu: Done

zeroliu:
Need to fix a bug in webcomponentjs first:
https://github.com/webcomponents/webcomponentsjs/pull/1001

zeroliu:
Our application covers a great amount of users who are using browsers that do not support template element (mobile web, TV browsers). To make lit-html work on those environment, we can either load a polyfill or change lit-html to make template optional.

The template polyfill is overkill compared to what lit-html actually needs. I suspect it would perform well on low end devices. @justinfagnani and I looked at the second option and found that we can achieve it with very minimal changes to the core code. The fact that developer no longer needs to worry whether the browser supports template natively is a huge win compared to a small amount of binary size increase. I believe it can benefit all developers and make lit-html more accessible.

I would argue that lit-html is not designed around template but tagged template literal. I see template element as an extra performance boost but in terms of functionality, it can be replaced by anything that can hold DOM elements that do no render when page is loaded.

web-padawan:
@zeroliu the way I see it, lit-html was designed to move platform forwards and there is a Template Instantiation proposal (w3c/webcomponents#747) which we expect to be implemented at some point.
There are other libraries built on top of using tagged literals, e.g. hyperHTML which are not expected to serve that purpose but rather designed for better support for older browsers-

I generally agree that using good old DocumentFragment is an understandable option, but again it seems better to keep it as an opt-in solution (should be as simple as change 1 import).

ruphin:
I understand the desire to reduce the amount of bytes required to be sent to consumers, but by building in a lightweight template polyfill inside lit-html, we end up shipping extra bytes to all users who DO have an environment that supports template (which is already most of them, should at some point be all of them).

As @justinfagnani mentioned, you probably won't see much of a performance gain by using this over the existing template polyfill, as it does pretty much the same thing as this.

If you want to reduce the number of bytes shipped, perhaps a solution is to make an alternative template polyfill that is more lightweight and only does what lit-html needs.

CaptainCodeman:
Including polyfills with a library makes it convenient to use but can be a real PITA if multiple packages end up shipping different versions of the same thing. If nothing else, it's unnecessary bloat and extra bytes being served to users of browsers that are _not_ the problem.

This is a great article about [polyfill best practices](https://www.w3.org/2001/tag/doc/polyfills/).

LT;DR; keep them separate - document what features the package requires and if you want to support a browser that doesn't have those, you can load them just for that browser. [polyfill.io](https://polyfill.io/v2/docs/) is often a great option for that.

Polyfills should just 'melt away'. If they are built into packages they become difficult to remove.

justinfagnani:
This isn't necessarily baking a polyfill into the package, but taking an alternate route in case `<template>` isn't supported, which happens to be some of what the polyfill does. In the presence of a polyfill, this code wouldn't be executed.

justinfagnani:
I think it seems reasonable to separate this into a micro-polyfill that does just what lit-html needs and include it in the lit-html package.

b-strauss:
Shouldn't this then be also removed from lit-html and included into the micro-polyfill?
https://github.com/Polymer/lit-html/blob/b6da58d575ddb37e5aa06afca30bc2ac3f3e6f52/src/directives/classMap.ts#L20

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,69 @@
...
+          this.content.append(body.firstChild);
+        }
+      },
+      configurable: true
+    });
+  };
+
+  defineInnerHTML(PolyfilledHTMLTemplateElement.prototype);
+
+  PolyfilledHTMLTemplateElement.decorate = function(template: any) {
```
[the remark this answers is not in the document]
zeroliu: done

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,69 @@
...
+    // if the template is decorated or not in HTML namespace, return fast.
+    if (
```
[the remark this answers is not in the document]
zeroliu: done

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,69 @@
...
+      template.namespaceURI !== document.documentElement.namespaceURI
+    ) {
+      return;
+    }
+    template.content = contentDoc.createDocumentFragment();
+    let child;
+    while ((child = template.firstChild)) {
+      template.content.appendChild(child);
+    }
+    template.__proto__ = PolyfilledHTMLTemplateElement.prototype;
```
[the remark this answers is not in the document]
zeroliu: done

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,69 @@
...
+  Document.prototype.createElement = function createElement() {
+    let el = capturedCreateElement.apply(this, arguments);
+    if (el.localName === 'template') {
+      el = capturedCreateElement.call(this, 'template');
+      PolyfilledHTMLTemplateElement.decorate(el);
+    }
+    return el;
+  };
+
+  (window as any).HTMLTemplateElement = PolyfilledHTMLTemplateElement;
```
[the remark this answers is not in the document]
zeroliu: done

zeroliu:
My tests throw `Cannot assign to read only property 'content' of object '#<HTMLTemplateElement>'` because even if I delete window.HTMLTemplateElement, createElement('template') would still generate a real template element in browsers that support template. There are a few options for the fix:
- Pass 'div' instead of 'template' to createElement
- Redefine content property getter and setter

If we can come up with a different approach to mimic no template support environment, that would be ideal.

zeroliu:
All 410 tests passed on chrome 69 locally. The test failure on travis looks confusing to me...

zeroliu:
Ping! can we merge this PR? Thanks

polyfills/template_polyfill.d.ts
```diff
@@ -0,0 +1,22 @@
+/**
```
justinfagnani: you need to add
```
/polyfills/
```
to .gitignore

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,59 @@
...
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ *
```
justinfagnani: separate the license and doc comments

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,59 @@
...
+    return;
+  }
+  const contentDoc = document.implementation.createHTMLDocument('template');
+
+  const upgrade = (template: any) => {
+    template.content = contentDoc.createDocumentFragment();
+    defineInnerHTML(template);
+  };
+
+  const defineInnerHTML = (obj: any) => {
```
justinfagnani: Is there much reason to have this as a separate function?

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,59 @@
...
+    Object.defineProperty(obj, 'innerHTML', {
+      set: function(text) {
+        contentDoc.body.innerHTML = text;
+        const template = this as HTMLTemplateElement;
```
justinfagnani: Since you always access `template.content` below, I'd declare this as:

```ts
const content = (this as HTMLTemplateElement).content;
```

src/polyfills/template_polyfill.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+      configurable: true
+    });
+  };
+
+  const capturedCreateElement = Document.prototype.createElement;
+  Document.prototype.createElement = function createElement() {
+    let el = capturedCreateElement.apply(this, arguments);
+    if (el.localName === 'template') {
+      el = capturedCreateElement.call(this, 'div');
```
justinfagnani: I forget: what''s the reason for not creating a "template" element? Can you document it here in a comment?

polyfills/template_polyfill.d.ts
zeroliu: done

src/polyfills/template_polyfill.ts
zeroliu: done

src/polyfills/template_polyfill.ts
zeroliu: done

src/polyfills/template_polyfill.ts
zeroliu: done

src/polyfills/template_polyfill.ts
zeroliu: The reason for using createElement('div') instead of createElement('template') is because deleting window.HTMLTemplateElement won't prevent document.createElement('template') from creating the real template element in chrome. Modifying template.content will then throw an exception and causing tests to fail.

I am glad you brought it up. Having the test affect the production code design doesn't feel right. I don't know if there is a test environment that only runs on browsers that do not support template. Otherwise we have to introduce this workaround or exclude this test.

justinfagnani: I think we can run these tests only on IE. They won't be run on community PRs, but will be on merges to master.

PR #470, opened 2018-09-02
Make PR references in changelog into links

Alexendoo:
Just a small convenience since bare #num issues don't get auto linked in markdown files


justinfagnani:
Thanks @Alexendoo!

PR #472, opened 2018-09-04
Break cycle in dependency graph

justinfagnani:
For compatibility with tools that don't handle cycles.

Note: I broke out more interfaces than strictly necessary (splitting TemplateProcessor would have been enough), but I didn't yet create an interface for NodePart, which would be nice to do soon.

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -13,7 +13,7 @@
  */
 
 import {reparentNodes} from './dom.js';
-import {defaultTemplateProcessor, TemplateProcessor} from './template-processor.js';
+import {TemplateProcessor} from './template-processor.js';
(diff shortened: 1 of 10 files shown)


src/lib/template-result.ts
```diff
@@ -28,7 +28,7 @@ export class TemplateResult {
 
   constructor(
       strings: TemplateStringsArray, values: any[], type: string,
-      processor: TemplateProcessor = defaultTemplateProcessor) {
+      processor: TemplateProcessor) {
```
rictic: Note this breaking change in the CHANGELOG?

justinfagnani: Thanks! Done

PR #477, opened 2018-09-06
Cache ShadyCSS version (Fixes #476)

### LINKED ISSUE #476: Shady Render is checking the polyfill version on every render
`verifyShadyCSSVersion` is called on every render, which makes several polyfill version checks. These checks only need to be done once, and the result can be cached.

ruphin:
This implements a cache for the ShadyCSS version check, which improves ShadyCSS render performance.

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -37,8 +37,7 @@ let compatibleShadyCSSVersion = true;
 
 if (typeof window.ShadyCSS === 'undefined') {
   compatibleShadyCSSVersion = false;
-}
-if (typeof window.ShadyCSS.prepareTemplateDom === 'undefined') {
+} else if (typeof window.ShadyCSS.prepareTemplateDom === 'undefined') {
   console.warn(
     `Incompatible ShadyCSS version detected.` +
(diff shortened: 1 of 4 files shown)


ruphin:
Unfortunately the tests break because they change the ShadyCSS version during runtime. We would need a separate 'html' testfile that loads the correct environment for this test. If you think this is a meaningful change I can implement that, but I'll hold off with implementing until it's approved.

justinfagnani:
Fixes #476

src/lib/shady-render.ts
```diff
@@ -33,17 +33,20 @@ declare global {
 const getTemplateCacheKey = (type: string, scopeName: string) =>
     `${type}--${scopeName}`;
 
-const verifyShadyCSSVersion = () => {
+let verifyShadyCSSVersion = () => {
```
justinfagnani: Let's just change this away from a function to a const boolean. I don't immediately see a reason we shouldn't do the check at boot up.

justinfagnani:
@ruphin I think the change and a separate test are worth it. Thanks!

ruphin:
Sorry for the number of commits, I can only run some basic tests myself so I'm abusing CI to run the test suite. I can squash the commits if needed.

PR #479, opened 2018-09-07
Remove unused variable

ruphin:
Tiny optimization. The `host` variable is only used once so we don't need this duplicate check.


justinfagnani:
Nice find

PR #485, opened 2018-09-09
Add AsyncIterator tracking on async-replace directive

ManRueda:
It solves the problem commented on the issue Polymer/lit-html#484


justinfagnani:
Thanks a lot @ManRueda!

PR #486, opened 2018-09-10
Adds `classMap` and `styleMap` directives

### LINKED ISSUE #467: Add directives for setting css styles and classes
Suggest transferring something like what is currently provided in LitElement

```
const bar = true;
const baz = false;
html`<div class="foo ${classMap({bar, special: baz})}"></div>`;

const margin = '2px';
html`<div style="display: inline; ${styleMap({height: '2px', margin})}"></div>`
```

sorvell:
Fixes #467


src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+      (part.committer.element as HTMLElement).style.cssText =
+          part.committer.strings.reduce((a, c) => `${a} ${c}`);
+      styleMapStatics.add(part);
+    }
+    // remove old styles that no longer apply
+    const oldInfo = styleMapCache.get(part);
+    for (const name in oldInfo) {
+      if (!(name in styleInfo)) {
+        // setting to `null` to "unset" a previously applied value.
+        ((part.committer.element as HTMLElement).style as any)[name] = null;
```
LarsDenBakker: `element.style.foo = null` does not work in IE11 :( you have to use `foo.style.removeProperty(foo)`

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+  directive((part: AttributePart) => {
+    if (!(part instanceof AttributePart) || (part instanceof PropertyPart) ||
+        part.committer.name !== 'style' || part.committer.parts.length > 1) {
+      throw new Error('The `styleMap` directive must be used in the style attribute ' +
+          'and must be the only part in the attribute.');
+    }
+    // handle static styles
+    if (!styleMapStatics.has(part)) {
+      (part.committer.element as HTMLElement).style.cssText =
+          part.committer.strings.reduce((a, c) => `${a} ${c}`);
```
justinfagnani: Use join

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+        // We explicitly want a loose truthy check here because
+        // it seems more convenient that '' and 0 are skipped.
+        part.committer.element.classList.toggle(name, Boolean(classInfo[name]));
+      }
+    }
+    classMapCache.set(part, classInfo);
+});
+
+const styleMapCache = new WeakMap();
+const styleMapStatics = new WeakSet();
```
justinfagnani: comment these

src/directives/css.ts
sorvell: Seems to work in testing. Do you know if there are there certain properties for which this doesn't work? (`removeProperty` doesn't accept the camelCased property name, so would love to avoid that if possible).

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A directive that applies css properties. This must be used in the `style` attribute.
+ * It takes the property names in the `styleInfo` object and applies the property
+ * values as a css style propertes. For example `{backgroundColor: 'red', borderTop: '5px'}`
+ * sets these properties to the element's style.
+ * @param styleInfo {StyleInfo}
+ */
+export const styleMap = (styleInfo: StyleInfo): Directive<AttributePart> =>
```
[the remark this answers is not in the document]
sorvell: Done.

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface StyleInfo {
+  [name: string]: string;
+}
+
+const classMapCache = new WeakMap();
+const classMapStatics = new WeakSet();
+
+/**
+ * A directive that applies css classes. This must be used in the `class` attribute.
```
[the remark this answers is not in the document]
sorvell: Done.

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface ClassInfo {
+  [name: string]: string|boolean|number;
+}
+
+export interface StyleInfo {
+  [name: string]: string;
+}
+
+const classMapCache = new WeakMap();
```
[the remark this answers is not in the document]
sorvell: Done.

src/directives/css.ts
```diff
@@ -0,0 +1,99 @@
...
+  directive((part: AttributePart) => {
+    if (!(part instanceof AttributePart) || (part instanceof PropertyPart) ||
+        part.committer.name !== 'class' || part.committer.parts.length > 1) {
+      throw new Error('The `classMap` directive must be used in the `class` attribute ' +
+          'and must be the only part in the attribute.');
+    }
+    // handle static classes
+    if (!classMapStatics.has(part)) {
+      part.committer.element.className =
+          part.committer.strings.reduce((a, c) => `${a} ${c}`);
```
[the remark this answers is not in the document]
sorvell: Done.

src/directives/css.ts
sorvell: Done.

src/directives/css.ts
sorvell: Done.

src/directives/css.ts
LarsDenBakker: Try this jsbin: http://jsbin.com/kosikijule/edit?html,output the background color should be removed after 2 sec

I notice now that that it works for `height` but not for `backgroundColor`, but it can't be due to casing because I've had this issue a lot with properties like `overflow` when writing css animations.

I understand not wanting to use `removeProperty` as you'd need to convert camel to dash...

frankiefu: You can reset by setting the CSS property to an empty string, that should work everywhere including IE11. e.g.
`element.style.foo = '';`

PR #489, opened 2018-09-11
Make TemplateInstance clone return upgraded fragments

ruphin:
As discussed in #455 

For some reason ShadyRender did not have the necessary code to upgrade the elements, so this possibly fixes a bug there. If this commit does not cause any tests to break then there's clearly some testing surface missing there, because it definitely changes behaviour.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -13,7 +13,7 @@
  */
 
 import {isDirective} from './directive.js';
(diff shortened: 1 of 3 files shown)


justinfagnani:
Tests are passing now (the failure was a flaky FF startup). There's definitely some missing coverage on shady-render, but @sorvell is doing some rework there, maybe he can add coverage from the main tests. We should really re-run all the render tests with shady-render.

PR #491, opened 2018-09-12
Rewrite render() to use NodePart

### LINKED ISSUE #422: Render null or undefined does not clear render container

justinfagnani:
This supersedes #445 

Fixes #422

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -130,6 +130,16 @@ export class NodePart implements Part {
   }
 
   /**
+   * Inserts this part into a container.
+   *
+   * This part must be empty, as its contents are not automatically moved.
+   */
+  appendInto(container: Node) {
+    this.startNode = container.appendChild(createMarker());
+    this.endNode = container.appendChild(createMarker());
+  }
+
+  /**
    * Inserts this part between `ref` and `ref`'s next sibling. Both `ref` and
    * its next sibling must be static, unchanging nodes such as those that appear
    * in a literal section of a template.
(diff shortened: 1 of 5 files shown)


ruphin:
I was looking into this exact patch yesterday. I came up with an implementation very similar to this, although yours looks a bit cleaner. The biggest problem with this change is that each `render(<something>)` now injects two markers that it didn't need before, which affects performance.

I think it should be possible to use a single marker. A NodePart only needs one node to exist (before or after) so it can reference it's parent. If the other node doesn't exist, it can assume that it is the first node (or the last node) in the context. This will require some refactoring of parts in general, but I feel like the number of required markers can be cut down, which will not only affect `render()`, but other instances of parts as well. My own implementation of NodePart does exactly this, and although it adds a bit of complexity, it creates one less marker for most instances of NodePart.

justinfagnani:
@ruphin there are actually a number of places where we can reduce markers. In this case (and a few others) we should be able to get it down to 0 markers. I think I want to get this in and attack markers more holistically.

PR #497, opened 2018-09-13
Cast directly to String

jridgewell:
The typeof check is kind of unnecessary, `String(str)` is just `str`.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -65,10 +65,10 @@ export class AttributeCommitter {
         if (v != null &&
             (Array.isArray(v) || typeof v !== 'string' && v[Symbol.iterator])) {
           for (const t of v) {
-            text += typeof t === 'string' ? t : String(t);
+            text += String(t);
           }
         } else {
-          text += typeof v === 'string' ? v : String(v);
+          text += String(v);
         }
       }
     }
@@ -227,8 +227,7 @@ export class NodePart implements Part {
       // TODO(justinfagnani): Can we just check if this.value is primitive?
       node.textContent = value;
     } else {
-      this._commitNode(document.createTextNode(
-          typeof value === 'string' ? value : String(value)));
+      this._commitNode(document.createTextNode(String(value)));
     }
     this.value = value;
   }


justinfagnani:
The conditional there was based on some results from: https://jsperf.com/string-wrapping-overhead

Which actually showed a biggered difference in Chrome at least, when I ran it before. I should make a new benchmark with the conditional as well.

ruphin:
My old benchmark was mostly just to show that casting is better than passing non-string values to `textContent`.

I created a new benchmark that tests conditionally casting vs always casting in three scenarios:
 - Inputs are mostly strings
 - Inputs are mostly numbers
 - Inputs are equally often strings and numbers

https://jsperf.com/string-wrapping-overhead-again

Results in Chrome seem somewhat random to me, with no significant difference in performance. In Firefox always casting is (significantly) faster. Probably best to save the bytes and just always cast, but just to make sure someone should run the benchmark other browsers too, since it clearly does impact performance, and this is probably one of the hot paths.

jridgewell:
```
  string += typeof value === String ? value : String(value);
```

Should be `typeof value === 'string'`

justinfagnani:
So I realize that `String` and `new String` are quite different, and I was thinking of the differences between string primitives and String objects returned from the String _constructor_. This isn't that and `String(s)` can obviously do a typeof equivalent check internally. So this looks good to me.

Also, jsperf was failing to save my edits to the micro benchmark 😡

ruphin:
Good point about the typeof check, I'll fix it and re-run. Unfortunately
jsperf is super unreliable in editing. :(

ruphin:
I had to create a new benchmark again because jsperf borked. Here is a proper version that also runs in all browsers:

https://jsperf.com/string-wrapping-overhead-fixed/1

Takeaway is that the `String()` cast DOES add a considerable overhead if the source was already a string. Probably because `String('someString')` does not return the same object, but constructs a new String object in memory that is a copy of the input, and returns that.

The condition check also adds some overhead, but less than casting. Ofcouse if the input is mostly non-string values, the performance of the conditional is worse because it's just added overhead, but if the input is mostly strings, the conditional version is significantly faster (~10% on most browsers).

My testing indicates that always casting is the most consistent method, regardless of the type of inputs. When the input is mostly strings, conditionally casting is universally faster. When the ratio of strings vs non-strings becomes too low, the conditional becomes slower. The tipping point is browser dependent.

I estimate the minimum ratio of Strings to make the conditional faster to be around 50% in Chrome, 20% in Edge, and 0% in IE11 (conditional is always faster in IE11). Note that the tests are a little flaky and variance is high, so you have to re-run the tests several times to get a reasonable indicator. I tested Chrome, Edge, and Internet Explorer.

I expect most inputs to be strings, so generally I think performance will be better with the conditional.

justinfagnani:
@ruphin I wanted to come back to this but JS Perf is broken again. Would you mind putting your benchmark up at https://esbench.com/ ?

ruphin:
JSperf deleted all my tests unfortunately. I'm trying to rewrite them from memory on esbench.

ruphin:
I updated the benchmark:
https://esbench.com/bench/5bc87f53f2949800a0f6217d

I think the results are the same as my previous tests. String(<string>) has an overhead cost, so always casting is not 'free'. From the benchmarks, even in scenarios where most of the values are non-string, checking before casting performs nearly equal to always casting, and when ~20% or more of the values are strings, checking before casting becomes significantly more efficient.

ruphin:
I'm also pretty sure there's some memory efficiency in checking before casting, since `String(<string>)` will create a new string object, and it's cheaper to just use the existing reference (I think).

PR #502, opened 2018-09-13
Ensure template given to ShadyCSS is the one rendered by lit-html

sorvell:
This ensures ShadyCSS can update the style in the template that will be rendered by lit-html. ShadyCSS needs to do this, for example, when the set of properties needed in any `@apply` changes.

Fixes https://github.com/Polymer/lit-element/issues/117

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -111,15 +111,19 @@ const styleTemplatesForScope =
     (fragment: DocumentFragment, template: Template, scopeName: string) => {
       shadyRenderSet.add(scopeName);
       // Move styles out of rendered DOM and store.
+      const styles = fragment.querySelectorAll('style');
       const styleFragment = document.createDocumentFragment();
-      Array.from(fragment.querySelectorAll('style')).forEach((s: Element) => {
-        styleFragment.appendChild(s);
-      });
-      // Remove styles from nested templates in this scope and put them
-      // into the "root" template passed in as `template`.
+      for (let i = 0; i < styles.length; i++) {
+        styleFragment.appendChild(styles[i]);
+      }
+      // Remove styles from nested templates in this scope.
       removeStylesFromLitTemplates(scopeName);
+      // And then put them into the "root" template passed in as `template`.
       insertNodeIntoTemplate(
           template, styleFragment, template.element.content.firstChild);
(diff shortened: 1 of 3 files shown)


src/lib/shady-render.ts
```diff
@@ -107,53 +107,48 @@ const shadyRenderSet = new Set<string>();
...
-            // Insert into lit-template (for subsequent renders)
-            insertNodeIntoTemplate(
-                template,
-                style.cloneNode(true),
-                template.element.content.firstChild);
-          }
+      shadyRenderSet.add(scopeName);
+      // Move styles out of rendered DOM and store.
+      const styleFragment = document.createDocumentFragment();
+      Array.from(fragment.querySelectorAll('style')).forEach((s: Element) => {
```
justinfagnani: I'd use a for loop... not a lot of reason to allocate the array.

src/lib/shady-render.ts
```diff
@@ -107,53 +107,48 @@ const shadyRenderSet = new Set<string>();
...
+        styleFragment.appendChild(s);
+      });
+      // Remove styles from nested templates in this scope and put them
+      // into the "root" template passed in as `template`.
+      removeStylesFromLitTemplates(scopeName);
```
justinfagnani: could this be renamed to be more clear that it's moving styles?

src/lib/shady-render.ts
```diff
@@ -107,53 +107,48 @@ const shadyRenderSet = new Set<string>();
...
+      insertNodeIntoTemplate(
```
justinfagnani: Is this the big change here? I'm having a little bit of a hard time seeing the change in behavior.

src/lib/shady-render.ts
sorvell: Done.

src/lib/shady-render.ts
sorvell: It is named correctly, but the comment is for the next two lines. Moved the comments to make more clear.

src/lib/shady-render.ts
sorvell: Added a comment. Yes, the key bit is that we hand `ShadyCSS` a template that will actually be rendered by `lit-html`. This way it can update the style inside it and affect subsequent `lit-html` renderings.

sorvell:
@justinfagnani PTAL

PR #514, opened 2018-09-19
Cache templates in NodePart

### LINKED ISSUE #511: `when` directive: false value should be optional
This would be more convenient if the false argument could be omitted. It's currently r[equired](https://github.com/Polymer/lit-html/blob/master/src/directives/when.ts#L49).

ruphin:
By caching templates in NodePart, we obviate the need for directives like `when`, since switching templates will be efficient by default. It works by creating a cache for rendered TemplateInstances when committing a TemplateResult, and re-using the TemplateInstances when we render the same Template.

Fixes #511

The code is not clean, it's just a prototype/idea to start off a discussion. Do not merge!

### DIFF
--- a/src/directives/when.ts
+++ b/src/directives/when.ts
@@ -12,16 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {directive, Directive, NodePart, reparentNodes} from '../lit-html.js';
-
-interface PartCache {
-  truePart: NodePart;
-  falsePart: NodePart;
-  prevCondition?: boolean;
-  cacheContainer: DocumentFragment;
-}
-
-const partCaches = new WeakMap<NodePart, PartCache>();
+import { directive, Directive, NodePart } from '../lit-html.js';
 
 /**
  * Efficiently switches between two templates based on the given condition. The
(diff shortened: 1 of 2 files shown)


justinfagnani:
I think wanting caching at this place is going to be the exception rather than the rule, and this will lead to more memory overhead.

The cases where a TemplateResult renders to a part are going to be:

- Partials: These won't change and don't need caching.
- Repeats: Usually homogenous, so the same template is already reused in the Iterable case, and caching is handled separately in the repeat() directive case.
- Switches: These will change the Template rendered, but even here only a portion of the cases will want caching.

`Template`s are going to be very long lived, since they're kept in a WeakMap keyed by template literal strings, and it's just not often that those could ever be collected. So once you create the template cache for a part it'll live for the lifetime of an app. This means that we'll essentially cache every branch in every template.

@kevinpschaaf is working on the NodePart attachment/moving APIs, and was thinking of adding a detach method that would move a part's contents into a DocumentFragment. I think this would be easy enough to use from directives like `when()` to avoid needing a built-in cache.

ruphin:
I think switches like `thing ? <some-template> : <other-template>` and variants are more common than you'd think. I've seen a lot of developers use lit-html to build apps and this is a pattern that they reach for _a lot_. Possibly because patterns like that are very common in React-like environments, and it seems harmless enough to do. I think there's value in making that pattern less expensive by default.

The memory usage is indeed an issue, but this is the trade-off. Caching by default means more memory usage, but less thrashing. I certainly think the "average" developer's app becomes faster with this change, but also more memory-consuming. It's probably best to not cache by default and make caching opt-in with a directive like `when`, because the other way around is a lot more difficult.

How would the detach method be different from the `moveNodes` function exported by `dom.js`? I don't see added value unless there's also an `attach()` method that re-attaches whatever was detached.

ruphin:
One option that could improve the memory situation is to not create the cache until you call `clear()` when the current value is a TemplateInstance. That way the cache will not be created in regular static partials and `repeat`.

justinfagnani:
I'm going to close this in favor of the `cache()` directive approach, so that the additional memory overhead will be opt-in.

Issue #145, opened 2017-10-21
Support options for declarative event listeners in lit-extended


justinfagnani:
Passive event listeners are particularly important:  https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

richardanaya:
Just for my own curiosity, why do you think this is something lit-extended should enforce?

justinfagnani:
This isn't about enforcing, just exposing options. Passive event listeners are important on scroll and wheel events for reducing jank, so it'd be nice if declarative event listeners could use the passive option.

richardanaya:
Neat!

bennypowers:
certain events should be passive by default e.g. touch and pointer events

justinfagnani:
I can think of a few syntactic options for passing options to lit:

1. Pass an object with `handleEvent` and `options` properties:

    ```ts
    html`<div on-scroll=${{handleEvent: (e) => doSomething(e), options: {passive: true}}}></div>`;
    ```
  
    **Pros:** Nicely explicit. Familiar by reusing an object with a `handleEvent` as accepted by `addEventListener`
    **Cons:** Verbose. Slight chance of a breaking change if objects with `handleEvent` and `options` properties are already passed to lit.

2. Pass an array:

    ```ts
    html`<div on-scroll=${[(e) => doSomething(e),  {passive: true}]}></div>`;
    ```

    **Pros:** Concise. Kind of like the argument list to `addEventListener`.
    **Cons:** Not obvious? Why is an array passed when a function is expected?

cc @sorvell @kevinpschaaf @azakus

justinfagnani:
And while those are both ok syntaxes, they both those suffer from a problem a creating fresh objects each render and forcing lit-extended to deep compare the options object.

We'll probably have to recommend a pattern of storing the configuration outside of the template, like:

```ts
class MyElement extends LitElement {
  _scrollListener = {
    handleEvent: (e) => this.scroll(e),
    options: {passive: true},
  };

  _render() {
    return html`<div on-scroll=${this._scrollListener}></div>`;
  }
}
```

In this case I think the object form is much easier to read.

bgotink:
Another option would be to pass these options in the attribute name. Possible syntaxes:

```js
html`<div on-scroll:passive=${(e) . => doSomething(e)}></div>`;
html`<div on-scroll.passive=${(e) . => doSomething(e)}></div>`;

[...]

PR #523, opened 2018-09-29
Allow render() to pass an event listener context to be used as the this value for all event listener calls.

justinfagnani:
This PR changes the signature of `render()` and a few other methods to accept an options object with an optional `eventContext` property. The property is used by EventPart as the `this` argument when calling event listeners.

This means that a Web Component class that calls `render()` can pass itself as the event context and event listeners in templates will not have to be bound to receive the component as 'this'.

LitElement can change roughly like this:

```ts
class LitElement {
  // ...
  update(changedProperties: PropertyValues) {
    render(this.render(), this.renderRoot!, {scopeName: this.localName!, eventContext: this});
  }
  // ...
}
```

And then subclasses can write event listeners as simple methods:

```ts
  private _onButtonClick(e) {
    // now called with the correct `this` value
    this.doThing();
  }

  render() {
    return html`
      <button @click=${this._onButtonClick}>Click Me</button>
    `;
  }
}
```

This should improve both ergonomics and memory consumption, as developers don't have to create bound event listener functions for every instance of a component.

## Future Work

When combined with a fix for #145, we can write event listener options directly onto prototype methods, and also do so with a decorator:

```ts
  
  @eventOptions({passive: true})
  private _onScroll(e) {
    //...
  }

  render() {
    return html`
      <div @scroll=${this._onScroll}></div>
    `;
  }
}
```

### DIFF
--- a/docs-src/guide/02-writing-templates.md
+++ b/docs-src/guide/02-writing-templates.md
@@ -16,6 +16,26 @@ html`<h1>Hello World</h1>`
 html`<h1>Hello ${name}</h1>`
 ```
 
+## Rendering Templates
+
+A lit-html template expresion does not cause any DOM to be created or updated. It's only a description of DOM, called a TemplateResult. To actually create or update DOM, you need to pass the TemplateResult to the `render()` function, along with a container to render to:
+
+```ts
+import {html, render} from 'lit-html';
+
+const sayHi = (name) => html`<h1>Hello ${name}</h1>`;
+render(sayHi('Amy'), document.body);
+
+// subsequent renders wiill update the DOM
+render(sayHi('Zoe'), document.body);
+```
+
+### Render Options
+
+`render()` also takes an options argument that allows you to specify:
+* `eventContext`: The `this` value to use when invoking event listeners registered with the `@eventName` syntax.
+* `templateFactory`: The TemplateFactory to use. A TemplateFactory creates a Template from a TemplateResult, typically caching Templates based on their static content. Users won't usually supply their own TemplateFactory, but possible use ShadyCSS-integrated when using `/lib/shady-render.js`.
+
 ## Thinking Functionally
 
 lit-html is ideal for use in a functional approach to describing UIs. If you think of UI as a function of data, commonly expressed as `UI = f(data)`, you can write lit-html templates that mirror this exactly:
@@ -71,7 +91,7 @@ There are a few types of bindings:
     html`<button @click=${(e) => console.log('clicked')}>Click Me</button>`
     ```
 
-## Supported Types
+## Supported Data Types
 
 Each binding type supports different types of values:
 
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -242,8 +242,9 @@ export class NodePart implements Part {
       this.value.update(value.values);
     } else {
       // Make sure we propagate the template processor from the TemplateResult
-      // so that we use it's syntax extension, etc. The template factory comes
-      // from the render function options so that it can control caching.
+      // so that we use its syntax extension, etc. The template factory comes
+      // from the render function options so that it can control template
+      // caching and preprocessing.
       const instance =
           new TemplateInstance(template, value.processor, this.options);
       const fragment = instance._clone();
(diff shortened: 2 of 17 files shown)


kenchris:
Great work! I was about to file an issue regarding event options the other day

src/lib/parts.ts
```diff
@@ -119,14 +119,17 @@ export class AttributePart implements Part {
...
   endNode!: Node;
   value: any = undefined;
   _pendingValue: any = undefined;
 
-  constructor(templateFactory: TemplateFactory) {
-    this.templateFactory = templateFactory;
+  constructor(options: RenderOptions) {
+    this.options = options;
+    // this.templateFactory = options && options.templateFactory ||
+    // defaultTemplateFactory;
```
kevinpschaaf: Remove?  Looks like you ended up making the top-level render always ensure an `options.templateFactory`?

src/lib/render.ts
```diff
@@ -36,11 +37,16 @@ export const parts = new WeakMap<Node, NodePart>();
...
+    options?: Partial<RenderOptions>) {
   let part = parts.get(container);
   if (part === undefined) {
     removeNodes(container, container.firstChild);
-    parts.set(container, part = new NodePart(templateFactory));
+    parts.set(
+        container, part = new NodePart({
+                     eventContext: options && options.eventContext,
+                     templateFactory: options && options.templateFactory ||
+                         defaultTemplateFactory,
```
kevinpschaaf: What was your rationale building up the default options here, vs. having their usage points fall back to defaults if not provided in `options`? Avoiding the need to  undefined-test the usage points? Seems like a good tradeoff for perf, just wanted to confirm since it means this spot needs to know about every option, which is kind of a bummer.

src/lib/shady-render.ts
```diff
@@ -157,12 +158,20 @@ const prepareTemplateStyles =
...
     container: Element|DocumentFragment,
-    scopeName: string) {
+    options: ShadyRenderOptions) {
+  const scopeName = options.scopeName;
   const hasRendered = parts.has(container);
-  litRender(result, container, shadyTemplateFactory(scopeName));
+  litRender(result, container, {
+    eventContext: options.eventContext,
+    templateFactory: shadyTemplateFactory(scopeName),
+  });
```
kevinpschaaf: It seems footgunny to have each `render` need to know about and forward on all the options in existence (TypeScript won't help remind you when you forgot to add a new option here if they're optional, right? or did I miss something?)... could we have the pattern be:
```js
litRender(result, container, {...options, templateFactory: shadyTemplateFactory(options.scopeName)});
```
Is the answer "no", because you need to filter out the extra options the the base render doesn't define to have the type work?

src/lib/render.ts
justinfagnani: Yeah, it was to avoid the checks and be able to set the default template factory here rather than that being determined at the use site. I think I can make this better though with spread so other options pass through.

It occurs to me though that we really never want these options to change, because they're saved into Parts when parts are created. If options changed in-between renders, Parts may have different options. One way around that is to remove options altogether from `render()`, and instead vend a render factory that takes options. Then we don't create the temporary options object either. WDYT?

justinfagnani:
@kevinpschaaf I pushed a commit to allow all options to pass through.

docs-src/guide/02-writing-templates.md
```diff
@@ -16,6 +16,26 @@ html`<h1>Hello World</h1>`
...
+const sayHi = (name) => html`<h1>Hello ${name}</h1>`;
+render(sayHi('Amy'), document.body);
+
+// subsequent renders wiill update the DOM
+render(sayHi('Zoe'), document.body);
+```
+
+### Render Options
+
+`render()` also takes an options argument that allows you to specify:
```
kevinpschaaf: @justinfagnani Add a note that render options should not change between calls to render?  Or at least that some options may only be respected on the first render?

justinfagnani: done

src/lib/render-options.ts
```diff
@@ -0,0 +1,20 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {TemplateFactory} from './template-factory.js';
+
+export interface RenderOptions {
+  templateFactory: TemplateFactory;
```
bashmish: @justinfagnani isn't it optional?

PR #526, opened 2018-10-01
update typescript to 3.1.1

### LINKED ISSUE #524: lit html 0.11.4 doesnt work with TS >= 3.1.1 because of duplicate identifiers
lit html declares the CustomElementRegistry class in template-instance.ts
declare global {
    class CustomElementRegistry {
        upgrade(node: Node): void;
    }
}

tsc also declares this class in 3.1.1 in lib.dom.d.ts so TSC throws a TS2300 Error
`error TS2300: Duplicate identifier 'CustomElementRegistry'.`
maybe remove the declaration and update to TSC => 3.1.1?

43081j:
Fixes #524 .

I don't think we need to provide any backwards compat here because the type declarations should be unchanged pretty much. We only used the removed patch interface internally.


justinfagnani:
@43081j can you add this change to the changelog, since it has an effect on users?

43081j:
yes im just working through a few other interface changes and was trying to figure out why NPM changed the lock to point at an IP rather than a hostname.. bit strange.

43081j:
@justinfagnani how is this? i left some comments in the review also as a few more changes were needed

also tslint failed (not sure why it didn't before?) so i resolved those small issues too.

src/test/lib/parts_test.ts
```diff
@@ -102,16 +102,16 @@ suite('Parts', () => {
         part.setValue([1, 2, 3]);
         part.commit();
         assert.equal(stripExpressionMarkers(container.innerHTML), '123');
-        assert.strictEqual(container.firstChild, startNode);
-        assert.strictEqual(container.lastChild, endNode);
+        assert.strictEqual<Node>(container.firstChild!, startNode);
```
43081j: this is a pattern i had to repeat a few times. its because `firstChild` is now a `ChildNode|null`, not just a `Node`. so `<T>(a: T, b: T)` doesn't work as `ChildNode|null` and `Node` aren't the same.

src/test/lib/render_test.ts
```diff
@@ -1057,7 +1057,7 @@ suite('render()', () => {
       assert.equal(stripExpressionMarkers(container.innerHTML), '<div></div>');
 
       // Wait for mutation callback to be called
-      await new Promise(setTimeout);
+      await new Promise((resolve) => setTimeout(resolve));
```
43081j: Promise became stricter or something, so now the arguments are strongly typed. because setTimeout doesn't take a `reject`, it failed before

PR #530, opened 2018-10-02
Add horizontal padding to main element

### LINKED ISSUE #529: Improve website mobile view
There is currently no padding on the main element which doesn't look good on mobile views.

screen shot 2018-10-02 at 7 20 24 am

Proposed solution: add `padding-right: 20px; `to match with the side nav padding.

screen shot 2018-10-02 at 7 20 50 am

nwarwick:
Fixes #529

### DIFF
--- a/docs-src/index.css
+++ b/docs-src/index.css
@@ -23,6 +23,7 @@ body {
 
 main {
   grid-area: content;
+  padding-right: 20px;
 }
 
 #top-nav {


nwarwick:
Signed the CLA!

justinfagnani:
Thanks a lot @nwarwick, but it looks like we're going to consolidate lit-html, LitElement and PWA Starter Kit docs on the same Jekyll layouts, as you can see in #533.

aadamsx:
PWA Start Kit seems overly complicated for someone just starting out IMO.  When I first reviewed it I was confused by something that was suppose to help me get started.  Mybe having a Level 1, 2, 3 start kits with progressively more tech and advanced concepts would help.

PR #531, opened 2018-10-03
Support event listener options.

### LINKED ISSUE #145: Support options for declarative event listeners in lit-extended

justinfagnani:
Event listeners can be functions or objects with a `handleEvent` method. Listeners are passed as both the listener and options arguments to `addEventListener`/`removeEventListener`, so that the listener can carry event listener options like `capture`, `passive`, and `once`.

```js
const listener = {
  handleEvent(e) {
    console.log('clicked');
  }
  capture: true;
}
html`<button @click=${listener}>Click Me</button>`
```

Fixes #145

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -404,6 +404,32 @@ export class PropertyCommitter extends AttributeCommitter {
 
 export class PropertyPart extends AttributePart {}
 
+declare global {
+  interface EventListenerOptions {
+    capture?: boolean;
+    once?: boolean;
+    passive?: boolean;
+  }
+}
+
+// Detect event listener options support. If the `capture` property is read
+// from the options object, then options are supported. If not, then the thrid
+// argument to add/removeEventListener is interpreted as the boolean capture
+// value so we should only pass the `capture` property.
+let eventOptionsSupported = false;
+
+try {
+  const options = {
+    get capture() {
+      eventOptionsSupported = true;
+      return false;
+    }
+  };
+  window.addEventListener('test', options as any, options);
+  window.removeEventListener('test', options as any, options);
+} catch (_e) {
+}
+
 export class EventPart implements Part {
   element: Element;
   eventName: string;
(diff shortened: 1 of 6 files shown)


justinfagnani:
I expected the IE11 and Chrome 41 tests to fail. I still need to add event listener options support detection as described here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support

justinfagnani:
ok @kevinpschaaf - I got tests passing on all browsers, this is ready to review!

src/lib/parts.ts
```diff
@@ -430,14 +456,32 @@ export class EventPart implements Part {
...
+            (newListener.capture !== oldListener.capture ||
+             newListener.once !== oldListener.once ||
+             newListener.passive !== oldListener.passive);
+    const shouldAddListener =
+        newListener != null && (oldListener == null || shouldRemoveListener);
+
+    if (shouldRemoveListener) {
+      try {
+        // This can throw if the options don't match the options that the
+        // listener was added with. This should never happen.
```
kevinpschaaf: If it should never happen, I wouldn't try/catch; it would be a bug that must be fixed either way, better to let the user know it's a library bug (throw) rather than log ("maybe I did something wrong").

kevinpschaaf: Er, or are you protecting against new options being added?  That might be a good reason to be a little more forgiving...

src/test/lib/parts_test.ts
```diff
@@ -431,4 +431,124 @@ suite('Parts', () => {
       });
     });
   });
+
+  suite('EventPart', () => {
```
kevinpschaaf: Want to add an integration test that uses `@event="{handleEvent: onSomething, capture: true}"` syntax in the template?

src/lib/parts.ts
kevinpschaaf: Oh, but you use a whitelist of properties to copy, so even if the user added new ones, the options passed to aEL would be the same, so I guess not?

src/test/lib/parts_test.ts
justinfagnani: yeah

src/lib/parts.ts
justinfagnani: No, this was just a try/catch because removeEventListener can throw if the arguments don't match a previous add call. There's a small chance that a user could change the options on a reference to the old listener and cause an exception because the options are pull off as needed. I could save to ole options to prevent that possibility.

justinfagnani: I changed to pull off the options so they can't be changed out from under the part.

To address the other part of your comment: this isn't to protect against new options. Ideally we'd like to let those pass through, there's just some difficulties with function objects that require a little bit more code. ie, spreading a method into an object was throwing for me because that was accessing `callee` which throws (I guess it's enumerable, but throws in strict mode? ugh). So we just need to filter some things out. I didn't want to deal with it honestly 🤷‍♂️ I can file an issue...

src/test/lib/parts_test.ts
justinfagnani: done

src/lib/parts.ts
kevinpschaaf: Yeah, we also want the diff (do the old options match the new ones) to determine when to remove to be fast, and supporting an arbitrary list makes that harder (slower).  So... probably ok as is; we can probably rev `lit-html` faster than browsers can ship a new event option. :P

src/test/lib/parts_test.ts
kevinpschaaf: Did you forget to push your changes?  Still only 1 commit on the PR...

justinfagnani: oops, indeed. pushed

PR #537, opened 2018-10-04
Fix formatting

arthurevans:
These raw tags are causing the Type Doc to get cut off.


justinfagnani:
That's unfortunate. Is there a way to mark code in comments that doesn't cause problems?

PR #546, opened 2018-10-08
Update dependencies

abdonrd:
Update the dependencies to the latest versions.


keanulee:
We're generally not updating dependencies in package.json unless we actually require the newer version.

PR #548, opened 2018-10-09
Don't clear() parts in asyncAppend and asyncReplace if the iterable isn't the current value.

### LINKED ISSUE #543: AsyncAppend and AsyncReplace can clear the part even if they are not the current value
In the following code, if the part had changed before the first item in the async iterable, then the part would still be cleared. If these two ifs are swapped it will work better 

        // When we get the first value, clear the part. This lets the previous
        // value display until we can replace it.
        if (i === 0) {
          part.clear();
        }

        // Check to make sure that value is the still the current value of
        // the part, and if not bail because a new value owns this part
        if (part.value !== value) {
          break;
        }

mariusGundersen:
This is a fix to issue #543

### DIFF
--- a/src/directives/async-append.ts
+++ b/src/directives/async-append.ts
@@ -50,16 +50,16 @@ export const asyncAppend = directive(
       let i = 0;
 
       for await (let v of value) {
-        // When we get the first value, clear the part. This lets the
-        // previous value display until we can replace it.
-        if (i === 0) {
-          part.clear();
-        }
-
         // Check to make sure that value is the still the current value of
         // the part, and if not bail because a new value owns this part
         if (part.value !== value) {
           break;
+        }
+
+        // When we get the first value, clear the part. This lets the
+        // previous value display until we can replace it.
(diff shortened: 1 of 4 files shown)


mariusGundersen:
I signed it!

justinfagnani:
Thanks for the fix! We need a test that fails before the change to make sure we don't regress.

mariusGundersen:
I tried to run tests, but I had to install java first, and I didn't really want to do that. Any ways to test without java?

ruphin:
I maintain a fork that can run the tests without Java. If you're interested I can see if I can extract the changes required to make that work.

@justinfagnani is there a reason we run the tests through WCT instead of Karma? I migrated the tests to Karma in my own fork because I can't install Java, and it looks like others are also having the same problem. If we migrate the project to Karma it could facilitate easier test running for all contributors.

mariusGundersen:
I've added test-cases that fail and rebased the fix after the commit with the failing tests

Issue #544, opened 2018-10-06
Asynchronous method in a conditional attribute notation


vdegenne:
I am trying to use the conditional attribute notation with an `async` function, something like that :
```html
<span ?special="${this.asyncMethodThatReturnsTrueOrFalse()}">...</span>
```
However It seems that *lit-html* will always evaluate this expression as `true` even if the method returns `false` ([minimal example](https://glitch.com/edit/#!/lit-html-async-conditional-attribute)).

justinfagnani:
Thanks for the reproduction @vdegenne!

bgotink:
At the moment only `NodePart`s handle promises and other types, `AttributePart`, `PropertyPart`, `EventPart` and `BooleanAttributePart` take the value as-is without special handling for promises, iterables, nodes etc.

In my opinion that shouldn't change. For boolean attributes, changing the behaviour would give unexpected consequences when trying to set an attribute depending on whether a value is truthy or not. That'll work for any value, except for promises (or rather: any value with a `then` property) because for those we'll look at the truthiness of the value they eventually get fulfilled with.  
It's of course true that one can always write

```html
<span ?special=${this.somePromise != null}>
```

which will always work with or without the change. When I see a template with 

```html
<span ?special=${this.something}>
```

I interpret that as 

```js
if (this.something) {
  spanEl.setAttribute('special', '');
} else {
  spanEl.removeAttribute('special');
}
```

---

We could easily update the `until` directive to support other parts and not just `NodePart`s. Right now this directive makes use of the fact that `NodePart` supports promises out of the box, but we could easily update `until` to subscribe to the promise itself.

This feels more correct to me for two reasons:

1. I explicitly declare that I want to use the async value of the promise, not the fact that the promise is present
1. This allows me to give a default value. If `BooleanAttributePart` automatically handles promises, I cannot declare whether I want the attribute to be present or not while the promise is unresolved.

justinfagnani:
Closed by #555

PR #555, opened 2018-10-10
Remove Promise support from core, add async() directive.

### LINKED ISSUE #272: Promised attributes are not rendered properly
Promised attribute are no rendered properly.

Using this code:
```javascript
import { render, html } from "./lit-html.js";

const Main = () => {
  const postTitle = new Promise( (resolve, reject) => {
    setTimeout( () => { resolve('Title') }, 2000 );
  });
  const postClass = new Promise( (resolve, reject) => {
    setTimeout( () => { resolve('good') }, 1000 );
  });

  return html`<h1 class="test ${postClass}">Post: ${postTitle}</h1>`;
};
render(Main(), document.body);
```

Renders:
```html
<h1 class="test [object Promise]">Post: Title</h1>
```

This seems to be caused by the fact that `AttributePart._interpolate()` treats all values as strings.

With the following change to `lit-html.js`, that can be remedied:
```javascript
export class AttributePart {
  ...
  async _interpolate(values, startIndex) {
    ...
      const r = getValue(this, values[startIndex + i]);
      const v = (r instanceof Promise ? await r : r);
    ...
  }
  ...
  async setValue(values, startIndex) {
    ...
      value = await this._interpolate(values, startIndex);
    ...
  }
}
```

### LINKED ISSUE #550: Consider removing built-in Promise support in favor of `until()` directive.
Currently we don't support Promises in attributes (#272, #544). I've been doing some work to support this, but it starts getting pretty complex as we try to answer things like:

- What is the value of a fully-controlled attribute before Promise resolution? Empty string or not present?
- What is the value of an interpolated expression before Promise resolution?
- What about boolean attributes? If the default value is not present, that's the same as defaulting to `false`. Is that correct?
- Do Promises also auto-resolve in property bindings? If so, how would we pass Promises as properties to components that actually _accept_ a promise?

While Promises in Node parts generally have none of these problems, support there sets the expectation that Promises just work in attributes as well. Promises support in NodePart also currently has some bugs due to race conditions. We have outstanding code to fix this, but it adds complexity to core.

So, one solution that limits complexity and keeps support consistent is to remove built-in Promise support from NodePart and require the `until()` directive to use Promises. This means that templates have to be explicitly aware of async values, which is also probably a good thing.

`until()` itself would get more complicated as it can't rely on built-in promise support, and has to track Promise identity and resolved state, but that complexity is pay-for-play.

This is obviously a breaking change, so deciding on this blocks a 1.0 release. I'm curious if anyone in the community has feedback here. Upgrading to a new version shouldn't be very difficult if you know where Promises are used in templates.

justinfagnani:
Removes built-in Promise support. Adds an async() directive as a successor to until() that is required to render Promises. async() handles race conditions, works with attribute, boolean attribute, property and event bindings. It also works as a nested directive value with other directives like repeat().

While this only removes a small amount of current core library code, more core code would have be required to handle attributes, etc., and the race conditions, mainly because current Part factoring has us sharing very little code between Part implementations. It’s possible that we could refactor Parts in the future to share more value-handling code and bring back built-in Promise support in the future. Adding that would not be a breaking change, but we need to remove support before 1.0 since removing later would be a breaking change.

This change keeps directives/until.js to re-export async to ease the transition. It’s deprecated and will be removed before 1.0.

Fixes #550 
Closes #272, #544, #451 and #458


src/directives/async.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const async = (promise: Promise<any>, defaultContent: any = noChange):
+    Directive<Part> => directive((part: Part) => {
+      let promisedValue = promises.get(part);
+
+      if (promisedValue !== undefined && promisedValue.promise === promise) {
+        // Set the default content if the Promise hasn't resolved *even if
+        // we're getting the same Promise* because the default content
+        // could have changed.
+        if (promisedValue.resolved && defaultContent !== noChange) {
```
bgotink: shouldn't this be `!promisedValue.resolved`?

src/directives/async.ts
```diff
@@ -0,0 +1,60 @@
...
+      // The following code only ever runs once per Promise instance
+      promisedValue = {promise, resolved: false};
+      promises.set(part, promisedValue);
+      if (defaultContent !== noChange) {
+        part.setValue(defaultContent);
+      }
+
+      Promise.resolve(promise).then((value: any) => {
+        const currentPromisedValue = promises.get(part);
+        if (currentPromisedValue === undefined ||
```
bgotink: undefined check is not needed because the next check is more strict

src/directives/async.ts
```diff
@@ -0,0 +1,60 @@
...
+            currentPromisedValue !== promisedValue) {
+          return;
+        }
+        promisedValue.value = value;
+        part.setValue(value);
```
bgotink: What if the value of the part has changed by now? `then` callback checks whether a new `async(promise)` was passed but it doesn't check for non-`async(promise)` value

src/directives/async.ts
justinfagnani: Yeah. I'll add a test.

src/directives/async.ts
justinfagnani: This isn't currently supported. There's really no way to tell that this directive has been removed - it has to be used somewhat statically.

bgotink: I though we could solve it by adding a new property to part, `readonly originalValue: any`, which would represent the value that gets set from the template regardless of whether it's a directive or something else entirely. That would allow the `async` directive to check whether it's still the active value in the case of a template like

```ts
html`${async(somePromise, 'temp')}`
```

but it would only work if `until` is used directly inside of the part, not in cases where it's wrapped in another directive, e.g.

```ts
html`${when(value instanceof Promise, () => async(value, 'temp'), () => value)}`
```

src/directives/async.ts
```diff
@@ -0,0 +1,59 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {directive, Directive, noChange, Part} from '../lit-html.js';
+
+interface PromisedValue<T = unknown> {
+  promise: Promise<T>;
+  resolved: boolean;
+  value?: T;
```
bgotink: This property is unused (it's set at line 55 but never read)
Did you mean to add 

```ts
if (promisedValue.resolved) {
  part.setValue(promisedValue.value);
}
```

in the if-check at lines 33-41?

src/directives/async.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+
+      // The following code only ever runs once per Promise instance
+      promisedValue = {promise, resolved: false};
+      promises.set(part, promisedValue);
+      if (defaultContent !== noChange) {
+        part.setValue(defaultContent);
+      }
+
+      Promise.resolve(promise).then((value: any) => {
```
LarsDenBakker: Why is an extra `Promise.resolve` needed? Is that to support non-promise values?

justinfagnani: Yeah, and mostly this is safer than blindly calling `.then()`

src/directives/async.ts
justinfagnani: ah, turns out it's just not needed

src/directives/async.ts
justinfagnani: I think this is something to be designed and solved for directives in general, not just async. Right now there are lots of sharp edges when directives aren't essentially "static" and top-level. Replacing directives can lead to other issues too.

justinfagnani:
@sorvell @kevinpschaaf PTAL

src/directives/async.ts
kenchris: In Canary I have `queueMicrotask` which does the same. Maybe it makes sense to call that when available, as it might have slightly less overhead 

https://www.chromestatus.com/features/5111086432911360

src/directives/async.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (defaultContent !== noChange) {
+        part.setValue(defaultContent);
+      }
+
+      Promise.resolve(promise).then((value: unknown) => {
+        const currentPromisedValue = promises.get(part);
+        if (currentPromisedValue !== promisedValue) {
+          return;
+        }
+        part.setValue(value);
```
kevinpschaaf: Gah, also looks like we're forgetting to set `promisedValue.resolved = true`, right?  Should add a test that catches this case?  Or did I miss something?

src/directives/async.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+        // If the Promise has resolved we simply return, as the value was
+        // updated in the then() callback below.
+        return;
+      }
+
+      // The following code only ever runs once per Promise instance
+      promisedValue = {promise, resolved: false};
+      promises.set(part, promisedValue);
+      if (defaultContent !== noChange) {
```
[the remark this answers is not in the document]
justinfagnani: Used your code. Thanks :)

src/directives/async.ts
justinfagnani: Added test that fails without setting `resolved`

justinfagnani:
@kevinpschaaf the code as currently intended to work should be good now, but I'm not 100% sure on the `until`-style API.

One API share we've talked about is accepting a varargs list of Promises:

```js
html`${async(result1, result2, fallback)}`
```

Where `result1` is prefered over `result2` which is prefered over `fallback`. Any value could be sync or async. The defaultContent parameter would no longer be special, just assumed to be sync or a faster-resolving Promise.

Another potential change is that I think we might want to also support passing a function and invoking it. This will make using async functions easier, since they won't have to be immediate invoked:

```js
<div>
  html`${async(async () => {
    const result = await fetchSomething();
    return html`<h1>${result.title}</h1>`;
  })}`
</div>
```

The problem there is the unfortunate `async(async` run. This makes me question the name...

It also kind of makes me want to keep text-position Promise support in core, and even expand text-position types to invoke functions, since they don't render to anything useful anyway. Then we support this:

```js
<div>
  html`${async () => {
    const result = await fetchSomething();
    return html`<h1>${result.title}</h1>`;
  }}`
</div>
```

But I'm still apprehensive about the usability danger of Promises being supported in text-position but not attribute position. That could be added later too...

justinfagnani:
Coming back around to this. My previous comment had a few problems, mainly that we _don't_ want to support invoking an async function each render. That would produce a new Promise each render and possibly re-render the placeholders while it resolves. The `runAsync()` directive generator pattern is the safe way to do that. This means the `async` name is fine.

I do think it's better to accept varargs of Promise or sync values and have them cascade in priority from left to right. I'll see if that's diable in a simple implementation.

I'll update, rebase and ping when done.

justinfagnani:
@kevinpschaaf PTAL

I got the generalized version working so that `async()` accepts any number of Promises and non-Promises and renders them in priority order. Handling cases where the arguments change over time is not incredibly obvious, but wasn't that bad either, so I added that with hopefully enough comments to explain and test coverage for it.

PR #556, opened 2018-10-10
Don't use export * to workaround rollup bug

keanulee:


Westbrook:
Has this update made it into a release somewhere? Would love to have this available in my build.

kr05:
I just came across this same error. Great to see a solution already implemented, can't wait for the release!

Westbrook:
@justinfagnani is there anyway we could get a bug fix release on this? Even if it wasn't the _right_ answer long term, this proves to be a pretty big blocker to my team moving forward with a number of projects. It seems like the "fix" at the polymer-cli level either isn't working or isn't going to be released, and the conversation at the RollUp level is still ongoing, so any help you could provide at this level via bug fix release or "next" release or what not would be greatly appreciated.

keanulee:
lit-html 0.13.0 now released.

Issue #518, opened 2018-09-23
shady-render @apply is broken in 0.10.1+


mkorenko:
CSS mixins are not applied to Polymer elements:
https://jsfiddle.net/bw03sng5/1
Works in 0.10.0:
https://jsfiddle.net/bw03sng5/2

This affects lit-element 0.6.x
lit-element 0.6.1:
https://jsfiddle.net/mgoqz923/1
Works in lit-element 0.5.2:
https://jsfiddle.net/mgoqz923/2

emilio-martinez:
Check out the [Changelog](https://github.com/Polymer/lit-html/blob/master/CHANGELOG.md). Several fixes have been done around this topic since then.

mkorenko:
@emilio-martinez these fixes don't fix the case above. Check the reproduction, it is based on the latest v0.11.4.

mkorenko:
I did some debugging and diff checking between 0.10.0 and 0.11.4 and here is what I discovered:
shady-render was using prepareTemplate() in shadyTemplateFactory instead of prepareTemplateDom() before, which also called ShadyCSS's prepareTemplateStyles()
If you replace prepareTemplateDom() with prepareTemplate() call - @apply case above works.
But that change was made intentionally, together with it all style processing was moved next to the litRender() call.
And that is the reason why css mixin is not applied:
paper-button, that has "@apply --paper-button;" inside, is rendered by litRender() before the css mixin is processed, so the "@apply --paper-button;" gets ignored and does not transform to "background: var(--paper-button_-_background, transparent);"
Though css mixin gets processed correctly, it happens after <paper-button> is rendered, so the mixin has no effect on <paper-button> as the button doesn't get the appropriate transpiled css variable.

mkorenko:
Actually, I found that replacing
`window.ShadyCSS.styleElement(container.host);`
to
`window.ShadyCSS.styleSubtree(container.host);`
in shady-render fixes the issue.
Also there was a problem in my repro that compatibleShadyCSSVersion was false as ShadyCSS was not in window at first load.
Here is the working example with styleSubtree() call in firstUpdated() hook:
https://jsfiddle.net/mgoqz923/10/

hyperpress:
@apply has been deprecated for awhile now, and we should really be getting away from it. I know it's still supported by Polymer, and not sure what the best alternative is for these use cases. Maybe leads can provide that answer.

mkorenko:
Sorry, didn't have time this week to post here...
[...]

PR #558, opened 2018-10-11
Add @apply tests for polymer-like elements

mkorenko:
Tests for #518


sorvell:
Thanks for this PR. We've added [similar tests](https://github.com/Polymer/lit-html/blob/master/src/test/lib/shady-render-apply_test.ts#L133) in lit-html 0.14.0 and confirmed that #518 has been fixed so going to close this PR.

PR #559, opened 2018-10-11
Fix @apply for polymer-like elements

mkorenko:
Workaround for #518

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -186,14 +186,15 @@ export const render =
         if (!hasRendered) {
           window.ShadyCSS.styleSubtree((container as ShadowRoot).host);
 
-          Array.from(container.querySelectorAll('*')).forEach((childElement) => {
-            if (!childElement.shadowRoot) {
-              return;
-            }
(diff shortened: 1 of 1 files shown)


src/lib/shady-render.ts
```diff
@@ -184,7 +184,17 @@ export const render =
         }
         // Update styling if this is the initial render to this container.
         if (!hasRendered) {
-          window.ShadyCSS.styleElement((container as ShadowRoot).host);
+          window.ShadyCSS.styleSubtree((container as ShadowRoot).host);
+
+          Array.from(container.querySelectorAll('*'))
```
LarsDenBakker: Using `Array.prototype.forEach.call` avoids creating a new array each time

sorvell:
Calling `styleSubtree` like this is pretty inefficient so this isn't a change we can accept.

`Lit-html 0.14.0` and `ShadyCSS 1.7.0` together should fix #518. However, it's important that when rendering a custom-element, it's easiest to ensure the necessary host-> client ordering by rendering in `connectedCallback`.

Issue #552, opened 2018-10-09
Update when directive to handle any number of cases.


justinfagnani:
We have two feature requests related to conditional rendering:

- #511 - the false value should be optional
- A `switch` directive.

We can resolve both of these requests by generalizing `when` to introspect it's arguments to operate in two modes:

### `if` mode

When the second argument is a function or TemplateResult, we treat the condition as a truthy value and render either the second or third argument.

```ts
html`${when(condition,
  () => html`condition is true`,
  () => html`condition is false`
)}`
```

### `switch` mode

When the second argument is an object, we treat the condition as simple value, and the second argument as a map of cases:

```ts
html`${when(value, {
  'one': () => html`value is one`,
  'two': () => html`value is two`,
  default: () => html`value is neither`,
})}`
```

In switch mode, cases can only be keyed by strings, number and symbols. If in the future we want to lift this restriction, we can accept arrays:

```ts
html`${when(value,
  ['one', () => html`value is one`],
  ['two': () => html`value is two`]
)}`
```

But this seems better to leave off for now.

One of the benefits of handling both the `if` and `switch` use cases in one directive is that we don't have to choose two new names, since we can't name a variable `if`, `switch` or `case`. It should also simplify choices.

### Caching

Currently `when` caches the DOM created by the true and false cases. This should increase performance on fast/frequently changing values, but this might not always be the case, and it increases complexity, code size and overhead. To make complexity pay-for-play we should remove caching from the `when` directive and add a separate `cachingWhen` directive. `cachingWhen` can locally be renamed to `when` when imported:

```ts
import {cachingWhen as when} from 'lit-html/directives/caching-when.js';
```

LarsDenBakker:
I had the same ideas when doing the initial implementation. I could take a stab at this if you werent already planning on doing it yourself?

akc42:
It is not clear above what happens in relation to #511 it is it permissible to miss off the third argument? when in If Mode? I have lots of cases where I've had to add 
```
() => html``
``` 
to my `when` directives

Also in switch mode, a missing `default` should also default to effectively an empty render.

One thing I don't understand is why the arguments have to be 
```
() => html`...`
```
I thought this was some sort of esoteric requirement related to caching (although isn't `html` itself effectively a function that returns a `TemplateResult` ?)  if caching isn't happening why

justinfagnani:
@akc42 both of those would be the case.

ruphin:
[...]

PR #560, opened 2018-10-11
Feature/update when

LarsDenBakker:
Fixes https://github.com/Polymer/lit-html/issues/552
Fixed https://github.com/Polymer/lit-html/issues/565

The cacheless `when` is just syntax sugar, and strictly speaking could be implemented as a simple function instead of a directive. For consistency I think it's better to keep it as a directive, we might want to add other things later.

The cached version becomes more complex due to the addition of the caseMap, I tried to keep it as concise as possible and share the same cache system between the if-mode and the case-mode.

### DIFF
--- a/src/directives/when.ts
+++ b/src/directives/when.ts
@@ -12,91 +12,52 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {directive, Directive, NodePart, reparentNodes} from '../lit-html.js';
+import { directive, Directive, NodePart } from '../lit-html.js';
 
-interface PartCache {
-  truePart: NodePart;
-  falsePart: NodePart;
-  prevCondition?: boolean;
-  cacheContainer: DocumentFragment;
-}
-
-const partCaches = new WeakMap<NodePart, PartCache>();
+export type WhenValue = () => any;
+export type CaseMap = { [key: string]: WhenValue };
 
 /**
- * Efficiently switches between two templates based on the given condition. The
- * rendered content is cached, and re-used when switching conditions. Templates
- * are evaluated lazily, so the passed values must be functions.
+ * Directive for handling conditional logic inside templates. One or
+ * two function parameters are treated as if or if/else statements. The
+ * first parameter is checked against truthiness and the associated value
+ * is rendered.
  *
- * While this directive can render any regular part, it makes the most sense
- * when used with TemplateResult since most other values are dirty checked
- * already.
+ * An object is treated as a case switch, where the condition is used as
+ * key to render the value of the matched case on the object. If no condition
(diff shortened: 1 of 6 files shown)


src/directives/when.ts
```diff
@@ -12,91 +12,63 @@
...
  *
- * Example:
+ * An object is treated as a case switch, where the condition is used as
+ * key to render the value of the matched case on the object. If no condition
+ * matches the default case is rendered if present. Keys can be strings and
+ * symbols.
  *
- * ```
- * let checked = false;
+ * Templates are re-instantiated each re-render. For caching nodes between
```
ruphin: This is not exactly correct, templates are not always re-instantiated on re-render. This directive uses the regular lit-html template handling, which re-uses templates by default. Rendering the following two things is effectively the same, caching-wise:

```javascript
when(true, html`<div></div>`)
// Same as
html`<div></div>`
```

Oonly switching between templates can cause template re-instantiation, as old templates will not be cached. This means you only re-instantiate a template when you switch between templates in an A -> B -> A fashion.

src/directives/when.ts
```diff
@@ -12,91 +12,63 @@
...
-            if (nextPart.value) {
-              parentPart.startNode.parentNode!.appendChild(
-                  cache.cacheContainer);
-            }
+  // test whether we are in case or in if/else mode
+  if (typeof trueValueOrCaseMap === 'object') {
+    caseMap = trueValueOrCaseMap;
+  } else {
+    trueValue = trueValueOrCaseMap;
+  }
```
ruphin: It is possible to reduce bytesize and improve performance here, if you are willing to use trueValueOrCaseMap for both cases.

```js
const caseMode = typeof trueValueOrCaseMap === 'object';

if  (caseMode) {
  nextValue = trueValueOrCaseMap[condition] || trueValueOrCaseMap.default;
// etc..
```

It makes the code slightly less readable, so it would require some code comments to clarify things. Maybe it would look better if the parameter has a different name? It's just a suggestion, I'm not sure if it's worth it.

LarsDenBakker: I had it like that initially, but it's much harder to read that way.

src/directives/when.ts
LarsDenBakker: You're right, I will adjust the phrasing.

akc42:
#565 Applies to the `whenWithCaching` part of this pr too

justinfagnani:
@LarsDenBakker thanks for the PR!

We've been talking about the right factoring of if/switch/caching/laziness features, and have some changes in direction here. I'll update #552 so the discussion is in an issue.

PR #569, opened 2018-10-15
[WIP] NodePart API refactor

kevinpschaaf:
_**Work in progress... do not merge yet**_

Refactors the `NodePart` API:
* Makes `NodePart` abstract and separates logic into:
  * `StaticNodePart` - can be attached to a parent _node_, with a static nodes as insertion references, and
  * `DynamicNodePart` - can be attached to a parent _`NodePart`_, with parts as insertion references
* Adds an `attach` API, which hides `startNode`/`endNode` as implementation details of parts 
  * Replaces previous `insert...`, `append...` API's
  * For `DynamicNodePart`, marker nodes are created as needed
  * For `DynamicNodePart`, calling attach at a new location moves any nodes created inside this part
* Adds a `detach` API to `DynamicNodePart`, which removes the part and optionally retains its nodes for future re-attachment

TODO: 
* [ ] Verify performance of improved API is acceptable
* [ ] Consider alternative API options
  * One base class and separate attach/insert API's for static vs. dynamic?
  * "insert before" semantics (to match DOM) instead of "insert after"?
* [ ] Add more tests that stress moving dynamic parts containing dynamic parts
* [ ] Add API docs, improve code comments


justinfagnani:
@kevinpschaaf this needs a rebase. Might be good to do before I start a review.

kevinpschaaf:
Can. Not. Figure. Out. Why. Travis gets a different clang format than locally!

PR #576, opened 2018-10-19
Don’t use `eventContext` for `this` of event listener objects.

justinfagnani:
Fixes https://github.com/Polymer/lit-html/issues/571

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -480,12 +480,11 @@ export class EventPart implements Part {
   }
 
   handleEvent(event: Event) {
-    const listener = (typeof this.value === 'function') ?
-        this.value :
-        (typeof this.value.handleEvent === 'function') ?
-        this.value.handleEvent :
-        () => null;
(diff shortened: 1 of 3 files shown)


src/lib/parts.ts
```diff
@@ -480,12 +480,11 @@ export class EventPart implements Part {
...
   handleEvent(event: Event) {
-    const listener = (typeof this.value === 'function') ?
-        this.value :
-        (typeof this.value.handleEvent === 'function') ?
-        this.value.handleEvent :
-        () => null;
-    listener.call(this.eventContext || this.element, event);
+    if (typeof this.value === 'function') {
+      this.value.call(this.eventContext || this.element, event);
+    } else {
```
kevinpschaaf: Did you intentionally remove the typeof check for `handleEvent`?

```suggestion
    } else if (typeof this.value.handleEvent === 'function') {
```

kevinpschaaf: As discussed irl, change was intentional, LGTM

PR #577, opened 2018-10-19
Switch from wct-browser-legacy to wct-mocha for testing

usergenic:
The prerelease wct-mocha package does away with the dependencies of old WCT client support code that was brought in by the wct-browser-legacy package, including some very outdated/abandoned package versions like a web-ready sinon package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "devDependencies": {
     "@types/chai": "^4.1.0",
     "@types/mocha": "^5.2.0",
+    "@webcomponents/shadycss": "^1.5.2",
     "@webcomponents/webcomponentsjs": "^2.0.4",
     "chai": "^4.1.2",
     "clang-format": "^1.2.4",
@@ -44,8 +45,8 @@
     "typedoc": "^0.12.0",
     "typescript": "^3.1.1",
     "uglify-es": "^3.3.5",
-    "wct-browser-legacy": "^1.0.1",
-    "web-component-tester": "^6.8.0"
+    "wct-mocha": "^1.0.0-pre.2",
+    "web-component-tester": "^6.9.0-pre.1"
   },
   "typings": "lit-html.d.ts",
   "dependencies": {}
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "typescript": "^3.1.1",
     "uglify-es": "^3.3.5",
     "wct-mocha": "^1.0.0",
-    "web-component-tester": "^6.9.0-pre.1"
+    "web-component-tester": "^6.9.0"
   },
   "typings": "lit-html.d.ts",
   "dependencies": {}
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "typedoc": "^0.12.0",
     "typescript": "^3.1.1",
     "uglify-es": "^3.3.5",
-    "wct-mocha": "^1.0.0-pre.2",
+    "wct-mocha": "^1.0.0",
     "web-component-tester": "^6.9.0-pre.1"
   },
   "typings": "lit-html.d.ts",
--- a/test/incompatible-shady.html
+++ b/test/incompatible-shady.html
@@ -9,7 +9,9 @@
         window.ShadyCSS = {shimcssproperties: true};
       }
     </script>
-    <script src="../node_modules/wct-browser-legacy/browser.js"></script>
+    <script src="../node_modules/mocha/mocha.js"></script>
+    <script src="../node_modules/chai/chai.js"></script>
+    <script src="../node_modules/wct-mocha/wct-mocha.js"></script>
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
   "devDependencies": {
     "@types/chai": "^4.1.0",
     "@types/mocha": "^5.2.0",
+    "@webcomponents/shadycss": "^1.5.2",
```
justinfagnani: why do we need this?

test/incompatible-shady.html
```diff
@@ -9,7 +9,9 @@
         window.ShadyCSS = {shimcssproperties: true};
       }
     </script>
-    <script src="../node_modules/wct-browser-legacy/browser.js"></script>
```
justinfagnani: did browser.js include a Symbol polyfill?

web-padawan:
Bug with Symbol polyfill introduced in webcomponents/webcomponentsjs#1005

/CC @ruphin

ruphin:
Most likely some part of wct pulls in and loads webcomponentsjs.

There was a bug in webcomponentsjs version 2.1.0 - 2.1.2 that caused the Promise polyfill to be force loaded even if the browser had native support. The linked PR fixed the issue and has been included in the latest 2.1.3 release, so updating webcomponentsjs should fix the problem.

ruphin:
Ah, it's the IE11 compatibility issue. The problem here is that babel-polyfill forces a Symbol polyfill that is different from the Polyfill used in webcomponentsjs. If webcomponentsjs loads before babel-polyfill, both polyfills will be loaded causing incompatibilities. The solution here is to load babel-polyfill before webcomponentsjs.

We're still looking to create a version of webcomponentsjs that doesn't include any polyfills and is meant to be used alongside babel-polyfill, to avoid problems like these.

package.json
usergenic: Okay it turns out we didn't need this.  My recollection was I added this because I saw a script reference to it, but I must have been mistaken because I don't see that direct reference anywhere.  Removed.

usergenic: Oh wait.  Now I discover why I added this.  Here's the network panel when I'm running tests with WCT:

screen shot 2018-10-22 at 1 17 39 pm

usergenic: With shadycss added as dependency, then it's all 200s:

screen shot 2018-10-22 at 1 43 19 pm

Thoughts? @justinfagnani

test/incompatible-shady.html
usergenic: browser.js didn't directly load a polyfill, but it did load several other scripts directly:
 - `stacky/browser.js`
 - `async/lib/async.js`
 - `lodash/index.js`
 - `mocha/mocha.js`
 - `chai/chai.js`
 - `@polymer/sinonjs/sinon.js`
 - `sinon-chai/lib/sinon-chai.js`
 - `accessibility-developer-tools/dist/js/axs_testing.js`
 - `@polymer/test-fixture/test-fixture.js`

I don't know if the situation described in https://github.com/Polymer/lit-html/pull/577#issuecomment-431556272 is ameliorated coincidentally by the wct-browser-legacy dependency set and loading behavior somehow causing babel-polyfill to load before webcomponentsjs.  Investigating whether this is the root cause and/or whether to fix in wct-mocha or wait for webcomponentsjs update to address.

package.json
usergenic: Oh derp so yeah totally this line loads it https://github.com/Polymer/lit-html/pull/577/files#diff-ba8440caa7003220705be8efef87282bR16

`<script src="../node_modules/@webcomponents/shadycss/apply-shim.min.js"></script>`

test/incompatible-shady.html
web-padawan: From my observations, when webcomponents polyfill is placed before browser.js, error happens. Otherwise, no errors. Have not found a root issue thogh.

usergenic:
Note-to-self: This can be fixed by adjusting the tests to do something other than string equality for the polyfill'd Symbol instance toString()

usergenic:
@justinfagnani the tests are green now.  I `toLowerCase()`'d the result of `Symbol.toString()` in the `assert.include` calls for the three failing tests and this allows tests to pass for either form of the output.

ruphin:
This is maybe not the right place to ask this question, but is there a reason we use WCT to test lit-html? We don't require any of the WCT features (like support for html imports), but WCT requires Selenium which requires Java which adds an external dependency. We could use Karma to run the test suite, and eliminate the Java dependency altogether, which makes it easier for contributors to run the tests.

justinfagnani:
@ruphin we could potentially use Karma. My understanding is that getting it to load modules was somewhat tricky, but @rictic has been working on that. We can also get WCT to use the webdriver package which doesn't require Java. It'd be better to discuss in an issue though.

usergenic:
Just rebased and updated to newly released `wct-mocha@1.0.0` prior to merging.

Issue #575, opened 2018-10-18
Sometimes a child element is re-created in safari instead of just updating the attribute/property


pshihn:
This example is using lit-element but I think applies to lit-html as well:

https://jsbin.com/fereyax/edit?html,console,output

In this example, when the button is pressed the border style of a custom element is changed. The constructor of the child custom element `color-panel` logs a message in the console. 

This happens in various scenarios on iOS Safari, but most reliably reproduced as follows:
* Open the link https://jsbin.com/fereyax 
* click on the button to change the color
* Go to a different tab in Safari
* Come back to the original tab and click the button again
* At this point, the constructor of the child element is called again.

pshihn:
This happens on Desktop Safari as well if you go to a different tab and stay there for 30 seconds - 1 min. Go back to the tab and hit the button -> Child Element gets recreated.

pshihn:
I believe it has something to do with how Safari puts background tabs into hibernation.

justinfagnani:
This probably has to do with the template cache. We store the processed lit-html `Template` objects in a WeakMap keyed by the identity of the TemplateStringsArray that we received as the first argument. I bet that as a page hibernates, those objects are GC'ed, and when the page wakes up new instances are created, cause us to have cache misses.

pshihn:
I have reproduced this without switching tabs on mobile either, but it is much harder to reproduce. Even those scenarios could be explained where GC happens for whatever reason. 
(This bug is really annoying for one particular case I have, where the scroll position gets reset among other things in a virtual list).

justinfagnani:
We can use a Map instead of a WeakMap. It's a memory leak, but only of templates that are never rendered against and their source is collected correctly.

ruphin:
As I mentioned, I have no device that runs Safari. I'll try to get one next
week to run some tests.

pshihn:
> templates that are never rendered

Don't know the code well, but as an outsider: You could kep a WeakMap and a Map. Promote from WeakMap to Map on render.

merlinnot:
I have the exact same issue, it causes a lot of trouble especially with elements which are expected to be placed on a website at most once (e.g. dialog controller), containing Google Maps (resets positions, markers, overlays...) or having on-create animations (e.g. charts).

In my app there is no need to switch tabs, it happens every few seconds (there are some realtime updates).

justinfagnani:
@merlinnot are you also seeing the in just Safari?

merlinnot:
[...]

PR #579, opened 2018-10-21
Use string concatenation for template cache id

ruphin:
This is a temporary solution for #575

[A bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=190756) causes it to not properly re-use the strings array for template literals. Because we use the identity of this object for our Template cache, we need some sort of temporary alternative until the problem is resolved.

The obvious solution is to generate a key by concatenating the strings array into a single string. The downside is that this operation adds overhead to every render in some way.

I see three options with regards to timing of when we generate this key.

- Generate the key every time you render a TemplateResult
This adds no overhead to Templates that are not rendered, but regenerates the key when a single TemplateResult is rendered more than once. This is a rather uncommon occurrence, but there are patterns that trigger this, for example with shared templates:
```js
const sharedStyles = html`<style> ... </style>`
```

- Generate the key when creating a TemplateResult
This guarantees we do not regenerate the key for the same TemplateResult. The downside is that we generate keys for templates that are not actually rendered.

- Generate the key when rendering a TemplateResult, and cache it as a property on the TemplateResult
We can generate the key only when rendering a TemplateResult, and cache the key on a property on the result. That way we avoid having to regenerate the key when we need to render the same TemplateResult multiple times, and we do not generate keys for TemplateResults that are never rendered. It comes at the cost of an extra equality check and property assignment.

I think the third solution is most likely the optimal one in this case, as generating keys is an expensive operation compared to equality checks. This patch implements this solution.

------

Since Safari is the only browser that needs this patch, it is possible that we want to limit the effect to other browsers. We could create a seperate TemplateFactory that implements this patch, and use that TemplateFactory on Safari using some kind of runtime detection. I'm not a huge fan of runtime browser detection, but if the overhead of this patch is significant, it might be worth exploring. I have not tested the performance overhead of this patch, but I plan to do some performance testing this upcoming week, and I will include this patch in some of my tests so we get an idea of the impact.

### DIFF
--- a/src/lib/template-factory.ts
+++ b/src/lib/template-factory.ts
@@ -44,11 +44,14 @@ export type TemplateFactory = (result: TemplateResult) => Template;
 export function templateFactory(result: TemplateResult) {
   let templateCache = templateCaches.get(result.type);
   if (templateCache === undefined) {
-    templateCache = new Map<TemplateStringsArray, Template>();
+    templateCache = {
+      stringsArray: new Map<TemplateStringsArray, Template>(),
+      keyString: new Map<string, Template>()
+    };
     templateCaches.set(result.type, templateCache);
   }
 
-  let template = templateCache.get(result.strings);
+  let template = templateCache.stringsArray.get(result.strings);
   if (template !== undefined) {
     return template;
   }
@@ -58,16 +61,16 @@ export function templateFactory(result: TemplateResult) {
   let key = result.strings.join(marker);
 
   // Check if we already have a Template for this key
-  template = keyedTemplates.get(key);
+  template = templateCache.keyString.get(key);
   if (template === undefined) {
     // If we have not seen this key before, create a new Template
     template = new Template(result, result.getTemplateElement());
     // Cache the Template for this key
-    keyedTemplates.set(key, template);
+    templateCache.keyString.set(key, template);
   }
 
   // Cache all future queries for this TemplateStringsArray
-  templateCache.set(result.strings, template);
+  templateCache.stringsArray.set(result.strings, template);
   return template;
 }
 
@@ -77,11 +80,13 @@ export function templateFactory(result: TemplateResult) {
  * in a Map.
  *
  * Safari currently has a bug which occasionally breaks this behaviour, so we
- * have to check if we processed a template tag already by creating a key that
- * retains identity by joining the TemplateResult.strings with a marker.
+ * need to cache the Template at two levels. We first cache the
+ * TemplateStringsArray, and if that fails, we cache a key constructed by
+ * joining the strings array.
  */
-export const templateCaches =
-    new Map<string, Map<TemplateStringsArray, Template>>();
+export type templateCache = {
+  stringsArray: Map<TemplateStringsArray, Template>;
+  keyString: Map<string, Template>;
+};
 
-// This maps the generated key strings to Templates
-export const keyedTemplates = new Map<string, Template>();
+export const templateCaches = new Map<string, templateCache>();
--- a/src/lib/template-factory.ts
+++ b/src/lib/template-factory.ts
@@ -45,7 +45,7 @@ export function templateFactory(result: TemplateResult) {
   let templateCache = templateCaches.get(result.type);
   if (templateCache === undefined) {
     templateCache = {
-      stringsArray: new Map<TemplateStringsArray, Template>(),
+      stringsArray: new WeakMap<TemplateStringsArray, Template>(),
       keyString: new Map<string, Template>()
     };
     templateCaches.set(result.type, templateCache);
@@ -85,7 +85,7 @@ export function templateFactory(result: TemplateResult) {
  * joining the strings array.
  */
 export type templateCache = {
-  stringsArray: Map<TemplateStringsArray, Template>;
+  stringsArray: WeakMap<TemplateStringsArray, Template>;
   keyString: Map<string, Template>;
 };
 
--- a/src/lib/template-factory.ts
+++ b/src/lib/template-factory.ts
@@ -13,7 +13,7 @@
  */
 
 import {TemplateResult} from './template-result.js';
-import {Template} from './template.js';
+import {marker, Template} from './template.js';
 
 /**
  * A function type that creates a Template from a TemplateResult.
@@ -48,7 +48,7 @@ export function templateFactory(result: TemplateResult) {
     templateCaches.set(result.type, templateCache);
   }
   if (result.key === undefined) {
-    result.key = result.strings.join('');
+    result.key = result.strings.join(marker);
   }
   let template = templateCache.get(result.key);
   if (template === undefined) {
--- a/src/lib/template-factory.ts
+++ b/src/lib/template-factory.ts
@@ -44,27 +44,30 @@ export type TemplateFactory = (result: TemplateResult) => Template;
 export function templateFactory(result: TemplateResult) {
   let templateCache = templateCaches.get(result.type);
   if (templateCache === undefined) {
-    templateCache = new Map<string, Template>();
+    templateCache = new Map<TemplateStringsArray, Template>();
     templateCaches.set(result.type, templateCache);
   }
-  let key = templateKeys.get(result.strings);
-  if (key !== undefined) {
-    // If templateKeys has the key it's guaranteed to exist in the templateCache
-    return templateCache.get(key)!;
+
+  let template = templateCache.get(result.strings);
+  if (template !== undefined) {
+    return template;
   }
 
-  // If the result.strings are new, generate a key and set it in templateKeys
-  key = result.strings.join(marker);
-  templateKeys.set(result.strings, key);
+  // If the TemplateStringsArray is new, generate a key from the strings
+  // This key is shared between all templates with identical content
+  let key = result.strings.join(marker);
 
-  let template = templateCache.get(key);
+  // Check if we already have a Template for this key
+  template = keyedTemplates.get(key);
   if (template === undefined) {
+    // If we have not seen this key before, create a new Template
     template = new Template(result, result.getTemplateElement());
-  } else {
-    // If a template already exists, refresh the key object in the templateCache
-    templateCache.delete(key);
+    // Cache the Template for this key
+    keyedTemplates.set(key, template);
   }
-  templateCache.set(key, template);
+
+  // Cache all future queries for this TemplateStringsArray
+  templateCache.set(result.strings, template);
   return template;
 }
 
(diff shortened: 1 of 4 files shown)


ruphin:
Another option that was discussed in the original issue is to require compiling out the template literals. This option solves the problem without any performance overhead, but it requires that end users compile their own code, as it is not something we can solve on our side. In my opinion this severely limits the usability of this library, and I don't really see this as a feasible solution, but I wanted to mention it here for completeness.

ruphin:
There's a small complication with generating the key by concatenating the strings.
```
['hello', 'world'].join('') === ['hell', 'ow', 'orld'].join('')
```

This leads to Templates with incompatible Part signatures being re-used. We need some sort of unique marker to join the strings with. I think we can just use the existing marker in Template, since it's more or less guaranteed to be unique.

ruphin:
That's quite clever. With that extra caching layer the overall impact is reduced quite a bit, as it only requires one key generation per template literal for compliant browsers. It also reduces the impact on Safari since the TemplateStringArrays are only infrequently reset.

It does add another `Map` lookup, but the performance of that is [significantly better](https://esbench.com/bench/5bce6d354cd7e6009ef61ba2) than having to join.

I was using the existing `marker` from `template.js` for the joiner string, to minimise the chance of collisions. Do you think there's any reason why this isn't a good idea?

justinfagnani:
Thanks :) It still incurs a memory penalty even on spec-compliant platforms, but based on the template size of some complex apps at Google, I think it's ok. Interestingly, this also works around bugs in compilers wrt template strings identity, like: https://github.com/Microsoft/TypeScript/issues/17956

I was originally concerned about still incurring a string comparison cost after Safari's bug kicks in, because the string instances in the template cache are different. This may still be something that shows up in profiles, and if so another tweak is to track if we're creating a new key for an existing template, and if so, delete the entry and re-add it to re-enable reference checking on the strings.

Something like this:

```ts
export const templateCaches = new Map<string, Map<string, Template>>();

export const templateKeys = new WeakMap<TemplateStringsArray, string>();

export function templateFactory(result: TemplateResult) {
  let templateCache = templateCaches.get(result.type);
  if (templateCache === undefined) {
    templateCache = new Map<string, Template>();
    templateCaches.set(result.type, templateCache);
  }
  let newKey = false;
  let key = templateKeys.get(result.strings);
  if (key === undefined) {
    key = result.strings.join('[lit-html-marker]');
    templateKeys.set(result.strings, key);
    newKey = true;
  }
  let template = templateCache.get(key);
  if (template === undefined) {
    template = new Template(result, result.getTemplateElement());
    templateCache.set(key, template);
  } else if (keyKey) {
    // The only way we can create a new key for an existing template is with this bug
    // in template literals. Delete the existing key and re-add to enable fast comparisons
    // until the next time the template strings identity changes
    templateCache.delete(key);
    templateCache.set(key, template);
  }
  return template;
}
```

> I was using the existing marker from template.js for the joiner string, to minimise the chance of collisions. Do you think there's any reason why this isn't a good idea?

No, that's fine. I just added another joiner 'cause I was lazy when typing.

ruphin:
Maybe I'm mistaken, but I don't think the situation you describe ever happens. The only map that gets duplicate entries when a TemplateStringArray is refreshed is `templateKeys`. `templateCache` has a `string` key, which is always the same, even if it has to be regenerated when a TemplateStringArray is refreshed.

We also don't have to worry about clearing the `templateKeys`, because it is a WeakMap, so any TemplateStringArrays that are incorrectly garbage collected by Safari will automatically also disappear here.

ruphin:
One last detail is that when we implement this, templates lose locality. Right now, each template has a unique identifier based on it's source location, regardless of content:

```js
const tag = (stringsArray, ...values) => stringsArray;

const a = () => tag`test${1}test`;
const b = () => tag`test${1}test`;

console.log(a() === a()) // > true
console.log(b() === b()) // > true
console.log(a() === b()) // > false
```

After we make this change, two templates with identical strings will be considered the same. This might actually be beneficial because it means we can reuse Templates, but it's important to note that the behaviour will change.

justinfagnani:
> templateCache has a string key, which is always the same, even if it has to be regenerated when a TemplateStringArray is refreshed.

It's the same string value, but because it's a dynamically allocated string and not a literal, it's unlikely to be interned, so it'll likely be a different string instance. On the map lookup an internal hashcode will need to be generated the first time (but then that's stored), but when two keys have the same hashcode, they then need to be checked for equality which will be a string comparison. If the strings were the same instance, then a reference comparison would return true before needing to do the string comparison.

By replacing the key in the `templateCache` we'd throw away the old string instance and be able to rely on reference comparison.

Granted, this is based on knowledge of VM internals and some assumptions (good ones, I think). I'd say we should measure, but this difference will only show up in Safari after the bug has hit. It might be cumbersome to measure only when this happens...

justinfagnani:
> One last detail is that when we implement this, templates lose locality

We don't rely on locality, since Edge/Chakra doesn't implement it correctly. We'll have to wait for that fix and the fix for this to make their way through browsers.

ruphin:
I see what you mean. I'll try to make a benchmark to see what the performance difference is for Map access.

ruphin:
Here's a benchmark: https://esbench.com/bench/5bcfbdcb4cd7e6009ef61ba8

It does indeed show that the comparison is orders of magnitude faster when the objects are the same. I'll implement a change that takes this into account.

justinfagnani:
@ruphin the esbench links are empty for me. I wonder if you have to save them? Thanks for the benchmarks!

ruphin:
Fairly certain that it works, it loads when I open it in incognito. You have to go to the 'results' tab, I can't link to that directly.

justinfagnani:
Ah, the info page looked like a modal dialog, because all the controls are styled to look disabled.

ruphin:
Stuck on a TypeScript issue. I implemented the algorithm with an optimal decision tree, but Typescript will complain that technically `template` can be `undefined`, because `templateCache.get(key)` has the `Template | undefined` type.

I'm not a TypeScript expert so I don't know if there's a way around this. If there isn't I'll use your implementation instead.

```js
export function templateFactory(result: TemplateResult) {
  let templateCache = templateCaches.get(result.type);
  if (templateCache === undefined) {
    templateCache = new Map<string, Template>();
    templateCaches.set(result.type, templateCache);
  }
  let key = templateKeys.get(result.strings);
  let template;
  if (key === undefined) {
    key = result.strings.join(marker);
    templateKeys.set(result.strings, key);
    template = templateCache.get(key);
    if (template === undefined) {
      template = new Template(result, result.getTemplateElement());
    } else {
      templateCache.delete(key);
    }
    templateCache.set(key, template);
  } else {
    template = templateCache.get(key);
  }
  return template;
}
```

ruphin:
I rebased the PR onto master. I also simplified the routine slightly to make it more understandable, and added some explanatory comments.

ruphin:
All the tests are passing, but Selenium is broken somehow, I have no idea what's causing it

src/lib/template-factory.ts
```diff
@@ -44,19 +44,40 @@ export type TemplateFactory = (result: TemplateResult) => Template;
...
+  key = result.strings.join(marker);
+  templateKeys.set(result.strings, key);
+
+  let template = templateCache.get(key);
   if (template === undefined) {
     template = new Template(result, result.getTemplateElement());
-    templateCache.set(result.strings, template);
+  } else {
+    // If a template already exists, refresh the key object in the templateCache
+    templateCache.delete(key);
```
pshihn: Don't think you need this `delete` because you are setting it again which would overwrite the previous value.

justinfagnani: I'm actually unsure here. We're setting to a value which is equal to, but not identical to, the previous value. I bet that Map implementations are free to replace the key or not in that case, as aside from the perf we're trying to eek out, there shouldn't be an observable difference.

justinfagnani:
@ruphin selenium is broken because of the Firefox update. Fixing that independently.

src/lib/template-factory.ts
```diff
@@ -44,19 +44,40 @@ export type TemplateFactory = (result: TemplateResult) => Template;
...
   }
+  templateCache.set(key, template);
```
justinfagnani: We only need to delete/reset when we generate a new key. I think it's fine to add a variable to track that.

src/lib/template-factory.ts
ruphin: I already tested this. The delete is necessary to force the key to update. The benchmark I linked above tests a version that sets after deleting, and without deleting, and the version that does not delete does not update the key object.

src/lib/template-factory.ts
ruphin: This set only happens when a new key is generated. There is an early return up above that happens when the key exists. It was originally structured like an if-else tree, but most of the code in this project seems to prefer early returns so I followed that style.

ruphin: I could add some more comments to explain the routine if necessary. I was worried I'd add too many comments if I explained it more thoroughly so I kept the comments to what I thought was the minimum necessary

ruphin:
I think there may be a problem with this approach. When alternating between different templates with the same string content, they will each get a different key object, but with the same value. The current algorithm would store only last used key object in the `templateCache` map. Pseudocode example:

```js
const a = html`hello`;
const b = html`hello`;

render(a) // New template - generate key for `a` and set it in `templateCache`
render(b) // New template - generate key for `b` and set it in `templateCache`
render(a) // Existing template - use key for `a` in `templateCache` lookup, which only has they key for `b`
```

One possible solution to this is to store the same key for both templates in `templateKeys`. It would take an additional key set and lookup during the "new template" routine, so it adds some cost to each initial template, but makes things faster for consecutive renders where this scenario happens.

ruphin:
I refactored the lookup routine to caching the Template at two levels. First at object level for the TemplateStringsArray, and if we don't find a match there, we have a cache at string equality level for the joined strings array. If we don't find a match there, we create a new Template. The nice part about this is that all caches map to the Template directly, so there's no longer a double cache lookup for objects that hit at the first cache level.

It fixes the problem I described above, and unless I made any mistakes, this version should also outperform the previous algorithm. I think it's more simple, so I strongly prefer it over the last iteration. I don't see any remaining issues with this.

@justinfagnani PTAL

jyounce:
Hi guys,
Thanks for working on this, I have been following this thread everyday. Just wondering when it is going to be available in a release for I have some components that do not work in safari without it.

src/lib/template-result.ts
```diff
@@ -22,6 +22,7 @@ import {lastAttributeNameRegex, marker, nodeMarker, rewritesStyleAttribute} from
  */
 export class TemplateResult {
   strings: TemplateStringsArray;
+  key?: string;
```
justinfagnani: Looks like this isn't used

Issue #582, opened 2018-10-23
HTMLTemplate as a supported Data Type


paulvarache:
Both `TemplateResult`s and `HTMLElement`s are supported as data types.
I am wondering if `HTMLTemplate`s should also be supported.

Example:
```js
        import { html, render } from './node_modules/lit-html/lit-html.js';

        const htmlTpl = document.createElement('template');

        htmlTpl.innerHTML = '<div>42</div>';
        
        const tpl = html`
            <div>${htmlTpl}</div>
        `;

        render(tpl, document.body);
```

I can work my way around this by using `htmlTpl.cloneNode(true).content`, and could make a directive that does that for me, but as lit is powered by HTMLTemplate is was wondering if it would make sense to support it as a Data type

justinfagnani:
Interesting. Can you tell me your use case?

paulvarache:
My use case is as follow:

I am creating a shared svg icons library. Each icon is a template with the svg element inside. This is to share across multiple projects using different libraries/frameworks. Using HTMLTemplateElements allows these icons to be used everywhere without depending on any library.

If I have these icons:
```js
import { svg } from './svg-string-to-template.js';
export const logout = svg`<svg xmlns="...">...</svg>`;
export const login = svg`<svg xmlns="...">...</svg>`;
```
With Polymer I can do:

```js
import { logout } from './icons.js';

class MyEl extends PolymerElement {
    static get template() {
        return html`
            ...
            ${logout}
            ...
        `;
    }
}
```
As I said before, I can definitely work with lit using a directive. I am wondering though what would be the performance implications of adding one more data type.
Especially if it is more rarely used.

justinfagnani:
Seems reasonable. It wouldn't be much of a perf hit.

paulvarache:
I will make a PR then

ruphin:
As I mentioned in #583 I think this feature can easily be implemented with a directive:

```js
const templateMap = new WeakMap();

const fromTemplate = directive(template => part => {
  if (templateMap.get(part) === template) {
    return;
  }
  templateMap.set(part, template)
  const content = template.cloneNode(true).content;
  part.setValue(content);
  part.commit();
}

// Usage:
const myTemplate = document.createElement('template');
myTemplate.innerHTML = '<p></p>'

html`<div>${ fromTemplate(myTemplate) }</div>`;
```

Since this is not a common use case, I don't see why we should build in support directly in lit-html core, because it will cause overhead to all other renders.

[...]

PR #583, opened 2018-10-23
Template data type

paulvarache:
Related to #582 
 - Add HTMLTemplateElement as data type
 - Add test
 - Update documentation

### DIFF
--- a/docs/_guide/02-writing-templates.md
+++ b/docs/_guide/02-writing-templates.md
@@ -150,6 +150,19 @@ const page = html`
 `;
 ```
 
+### HTML Template
+
+Similarly HTML Template can be used. They will be cloned and their content will be attached to the DOM:
+
+```js
+const template = document.createElement('template');
+template.innerHTML = '<div></div>'
+const page = html`
+  ${template}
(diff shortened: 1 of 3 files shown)


paulvarache:
I signed it!

ruphin:
This adds an additional `instanceof` check to every Part commit. Is there a reason why this feature can't be implemented with a directive? I thought we wanted to keep the core as lightweight as possible, and this adds no additional functionality.

ruphin:
This for example:

```js
const templateMap = new WeakMap();

const fromTemplate = directive(template => part => {
  if (templateMap.get(part) === template) {
    return;
  }
  templateMap.set(part, template)
  const content = template.cloneNode(true).content;
  part.setValue(content);
  part.commit();
}

// Usage:
const myTemplate = document.createElement('template');
myTemplate.innerHTML = '<p></p>'

html`<div>${ fromTemplate(myTemplate) }</div>`;
```

justinfagnani:
Thanks for the PR, but we did decide to not handle `<template>` specially. We can handle templates in directives.

Issue #496, opened 2018-09-13
Caching or update problem with boolean attribute "checked"


annolangen:
I am attempting a lit-html variant of the todomvc app (See todomvc.com for more info). Every call to "show" from controller.js calls render on the node "todoList" as a function of the Javascript object "model". After deleting a todo item from the list the boolean attribute "checked" is incorrect at the list position of the deleted list item. A page reload fixes the problem and illustrates that the checked attribute before the reload was not reflecting the todo's completed field.  

Detailed instructions to reproduce.  Observed in Opera 55.0.2994.44 and Chrome 68.0.3440.106
- Bring up attached app
- Create two todo items, say, "a" and "b"
- Mark the first item complete by clicking the circle next to "a"
- Click "Clear completed"
- I see the single todo item "b", but its input toggle is displaying as completed! 
- Notice that all other indications show that the model has no completed todo items. In particular, the todo title lacks the strike through style that is controlled by the completed field. 
- Reloading, as well as routing to other filters, restore the correct display

This looks like a caching or update problem to me. It looks like the update algorithm is reusing the DOM list item but the boolean attribute "checked" is not updated to false.

[todomvc.zip](https://github.com/Polymer/lit-html/files/2377356/todomvc.zip)

ruphin:
Lit-html does indeed re-use the DOM structure, but it will update the dynamic values.

If you inspect the DOM, you will see that lit-html works properly: the 'checked' attribute is correctly set. I have no idea why Chrome at least doesn't correctly render the label background image. It looks like a CSS bug to me. I'm not sure why the CSS for this app uses the sibling selector instead of descendant from `.completed`, as you can see the strikethrough does update correctly, and it depends on the same variable.

ruphin:
Ah, I see the problem. It's an issue with how Chrome handles the `checked` attribute on an input. When you create an input element, it checks the `checked` value to determine it's initial value. However, after the initial render, changing the `checked` attribute does not change the checked state of the input. If you want to change the `checked` state, you have to set the `checked` _property_, and not change the attribute.

In short:  use `.checked=${completed}`

annolangen:
Your proposed solution works. Clearly, I have something to learn about finer distinctions of properties and attributes.

annolangen:
[...]

PR #584, opened 2018-10-24
UpdateWritingTemplate

zachcd:
See #496 

The checked boolean attribute has problems in Google Chrome as the property of the element is favored by the engine over the attribute.
I recommend changing the docs to use disabled as the Boolean attribute example.

### DIFF
--- a/docs/_guide/02-writing-templates.md
+++ b/docs/_guide/02-writing-templates.md
@@ -84,7 +84,7 @@ There are a few types of bindings:
     ```
   * Boolean Attribute:
     ```js
-    html`<button type="submit" ?disabled=${disabled}>`
+    html`<input ?disabled=${disabled}>`
     ```
   * Property:
     ```js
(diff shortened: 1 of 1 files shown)


zachcd:
I signed it!

docs/_guide/02-writing-templates.md
```diff
@@ -84,7 +84,7 @@ There are a few types of bindings:
     ```
   * Boolean Attribute:
     ```js
-    html`<input type="checkbox" ?checked=${checked}>`
+    html`<button type="submit" ?disabled=${disabled}>`
```
keanulee: Should include closing tag, or stick with self-closing `<input>` instead.

zachcd: sure, I'll swap it to a disabled input

keanulee:
LGTM. cc @arthurevans for this docs update.

PR #589, opened 2018-10-25
Fix the 3 tests which fail on IE for string comparison for Symbol.toString() due to polyfill updates

usergenic:
After updating just the package-lock, there were tests failing for Symbol string comparison due to the polyfill change.  Fixed those here, separate from the wct-mocha branch where this was first encountered.


usergenic:
Closing as these changes went in as part of a larger PR https://github.com/Polymer/lit-html/pull/577.

Issue #527, opened 2018-10-02
Removed handler for parts/directives


mariusGundersen:
I made an attempt to implement a directive for observables, similar to the async-replace observable, and afaict there is no way to unsubscribe. That is, if the asyncReplace directive is nested inside a conditional, then it will continue to run even when the condition is false. This can have negative effects on performance and memory, as the directive and async iterable are not garbage collected and useless work is done.

Have one missed something, or is there a way to be notified when a part is (transitively) removed from the DOM?

mariusGundersen:
I've looked into this a bit (and only a bit!) and I think it can be done. The strategy is have in mind is to have a callback that can be called when the part is removed. This isn't enough though, as it's possible to detach not just leaf nodes but whole sub trees. Therefore a part needs a list of children that it will notify when it's detached, so the call happens recursively.

So the commit method would have to check not only what type the new value is but also what the old value is, and then call the detached handler method on it if it is a part or list of parts.

mariusGundersen:
Ok, I've looked into it a lot more, and I've found that there is a problem with how directives currently work. The problem is that a directive is given the part and can manipulate it directly by calling `setValue(value)` and `commit()` (which is confusing, see #534). The problem is that there is no way to know if the parent element is replacing the part or if it's the directive that does it, so there is no way to know if the currently applied directive should be unsubscribed or not.

I think directives should not be be given the part, but instead something that wraps the part, so that the lifetime of the directive can be tracked. This will probably help with #534 as well. It could be given either an object with methods, or maybe just a function that takes the new value as parameter.

justinfagnani:
This is a dupe of #283 

What do you have in mind with something that wraps Part?

My current thought is to maintain a stack of Parts/TemplateResults while rendering, and add a `onDisconnect()` callback/Promise on a Part that will walk the stack and add the callback to each Part being rendered. That way we don't have to send signals down a tree on disconnect, or maintain parent references, so the feature is pay-for-play.

mariusGundersen:
Oh, I see many of the same things have been discussed there, but it's a long time and no result yet. It might also be that some things have changed since then.

[...]

PR #591, opened 2018-10-28
Reimplement directives

mariusGundersen:
This is a reimplementation of how directives work with the goal of being able to track when a directive is detached in a future pull-request, as described in #527.

The core of this pull-request is the change of the `while(isDirective(value)){}` loop into a recursive call (possible since tail recursion is equivalent to iteration). That is, instead of storing only the final value in the part a stack of parts is created with the value stored in the topmost part on the stack. This results in the directive function being called with a new fresh Part and the directive is free to manipulate it however it likes. This means that the `setValue()` and `commit()` methods of Part of a Template can only be called by the Template (in the `update()` method), directives are given their own Parts to work with.

To make this possible the `directive()` function is replaced with a `createDirective` function. This function returns a `DirectiveResult` which the Part converts to a `DirectiveInstance` similar to how the ``html`...``` method returns a `TemplateResult` that is converted into a `DirectiveInstance`. The `createDirective` function makes it possible to keep state per Part using closures instead of WeakMaps and makes it simple to differentiate when a directive is created and when it is merely updated:

```js
// Just a dummy directive example
const myDirective = createDirective(part => {
  // This is where to store state for this directive
  // because this closure is reused for the part
  // every render
  let previousValue, previousMapper;
  return (value, expensiveMapper) => {
    // This is called every time the part is rendered,
    // potentially with a new value
    if(value !== previousValue || expensiveMapper !== previousMapper){
      // Almost everything calls commit right after setValue, so the commitValue method
      // has been introduced to make it simpler
      part.commitValue(expensiveMapper(value));
      previousValue = value;
      previousMapper = expensiveMapper;
    }
  }
});
``` 

There is also a way for a directive to be notified of when it is detached from the DOM, so it is easier to work with directives that asynchronously modifies a part:

```js
const tick = createDirective(part => {
  let i = 0;
  const intval = setInterval(() => part.commitValue(i++ % 2 ? 'tick' : 'tock'), 1000);

  // The lambda is called when the part is detached, clearing the interval
  part.onDetach(() => clearInterval(intval));

  // This directive doesn't take a value, so return an empty function.
  return () => {};
});
```

In this pull-request the onDetach event is quite limited and only called when the Part is directly removed, but it should be fairly easy to expand this to handle any Part being removed in the tree structure of Parts, TemplateInstances and DirectiveInstances.


mariusGundersen:
I've changed the api slightly by removing the `onDetach()` method from the Part and moving the handling of detachment to the `DirectiveInstance` instead. The new api allows an object with two methods, `update` and `detach`, to be returned when the directive instance is created:

```js
const myDirective = createDirective(part => {
  // return an object with both an update and detach callback
  return {
    update(value) {
      // this is called every time the template is rendered
    },
    detach() {
      // this is called when the directive is detached from the part
    }
  };
});
```

mariusGundersen:
I am unfortunately unable to run `npm format` beceause it assumes a unix machine and doesn't work on a Windows machine. Running tests is also difficult because they requrie Java. This is an unfortunate obstacle to contributing this project.

justinfagnani:
@mariusGundersen from a glance this PR seems like it aligns with some ideas we've had for a directive redesign, but honestly it's quite a large change to review without previous agreement on the design. A code review isn't the best place to review design. I'd rather discuss the shape of the solution before a lot of code is written.

We've been considering aligning directives to templates a little more, including introducing DirectiveResult, and various ways of abstracting state for directives. We need to resolve this before settling on an option.

mariusGundersen:
Sure, no problem, I didn't make this pull request expecting it to be merged directly, but more as a proposal and a basis for discussion. I have made it so as to keep the current api as stable as possible, but if that's not a requirement then there are potential for bigger changes.

What's your preferred way to discuss the design? Is there an open issue I can have a read through?

justinfagnani:
@mariusGundersen thanks for understanding. We have a few related issues open, but nothing comprehensive that touches on the three big areas I see right now: state management, and composition and the disconnect feature. Most of the team discussions have been in person, and not fully captured to issues, so I apologize for that.

Let me open an issue tonight with a summary of things.

justinfagnani:
Closing as out of date and superseded by 2.0

PR #593, opened 2018-10-28
Fixed `when` directive example syntax errors.

### LINKED ISSUE #592: `when` directive example doesn't work
In the documentation for directives there is an example for the `when` directive. The code example doesn't work because the html template was written incorrectly.

The broken part of the code example:

```js
html`
  when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not	
checked`);
```

[Link to the file with issue](https://github.com/Polymer/lit-html/blob/master/docs/_guide/02-writing-templates.md#directives)

ChrisBrownie55:
Fixes #592 

The `when` directive had errors in it and was not functional, to fix this I interpolated the `when` directive.

### DIFF
--- a/docs/_guide/02-writing-templates.md
+++ b/docs/_guide/02-writing-templates.md
@@ -304,9 +304,7 @@ import {when} from 'lit-html/directives/when';
 
 let checked = false;
 
-html`
-  when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not
-checked`);
+const render = checked => html`<div>${when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not checked`)}</div>`;
 ```
 
 ### `repeat(items, keyfn, template)`


docs/_guide/02-writing-templates.md
```diff
@@ -304,9 +304,7 @@ import {when} from 'lit-html/directives/when';
 
 let checked = false;
 
-html`
-  when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not
-checked`);
+const render = checked => html`<div>${when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not checked`)}</div>`;
```
pinkhominid:
```suggestion
html`
  ${when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not checked`)}
`;
```

pinkhominid:
I noticed this bug in the examples as well and was going to submit a fix, but found yours. Can I suggest a simplification that is more in line with the other examples?
```js
html`
  ${when(checked, () => html`Checkmark is checked`, () => html`Checkmark is not checked`)}
`;
```

PR #596, opened 2018-10-29
Fixed directive examples bad code

### LINKED ISSUE #595: Incorrect directive examples
In the directives section of the documentation, the first two examples show the use of directives incorrectly. Use of the code in the documentation would **result in the rendering of a function to the screen**.

Here is the bad code:

```js
const hello = directive(() => (part) => {
  part.setValue('Hello');
});
```

```js
const safe = directive((f) => (part) => {
  try {
    return f();
  } catch (e) {
    console.error(e);
  }
});
```

ChrisBrownie55:
This fixes #595 

This changes the first two directives examples to be functional by **moving the first arrow function outside of the call to `directive`**, therefore, no longer rendering the second arrow function to the screen but rather executing it.

### DIFF
--- a/docs/_guide/02-writing-templates.md
+++ b/docs/_guide/02-writing-templates.md
@@ -249,7 +249,7 @@ Directives are functions that accept some arguments for values and configuration
 ```javascript
 import {directive, html} from 'lit-html';
 
-const hello = directive(() => (part) => {
+const hello = () => directive(part => {
   part.setValue('Hello');
 });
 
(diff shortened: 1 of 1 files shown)


csvn:
Actually @ChrisBrownie55, I think the documentation is correct. It just has not been released yet. It was changed in this commit bcc87cf6488fa414e5361a9d46d992f01c11517a

ChrisBrownie55:
@csvn Oh my bad, sorry didn't realize, I was reading the docs on how to create a custom directive and it just wasn't working. Are there anyways to show the docs for the current version? Or on the docs page specify which versions work which way.

csvn:
Sorry, not as far as I know. I think a new release is pretty close, and there also [seems to be progress towards v1.0 too](https://github.com/Polymer/lit-html/pull/555). The API should be more stable after v1.0 is released.

PR #607, opened 2018-11-02
Fix typedoc generation

keanulee:
This is independent of the other docs stuff I'm working on, so making this a separate PR. Please focus review on package/-lock.json.

- Use `--ignoreCompilerErrors` option to typedoc because it complains about the following compiler error, even though it doesn't show with `tsc`:
```
Using TypeScript 3.1.1 from /Users/keanulee/Code/Polymer/lit-html/node_modules/typescript/lib
Error: Option 'declarationMap' cannot be specified without specifying option 'declaration' or option 'composite'.
```
- Add Google Analytics ID


arthurevans:
Ping?

justinfagnani:
The compile error is weird, since we do have `declaration: true` in the tsconfig. I assume that tsdoc is generating its own tsconfig from ours and not copying both settings.

keanulee:
Updated docs now deployed to https://lit-html.polymer-project.org/api/index.html

PR #609, opened 2018-11-03
Failing test cases 

mariusGundersen:
In anticipation of changes to how directives work I've started to collect failing test-cases that can be used to verify that the new solution works correctly. These cases will fail now, so obviously don't merge this.


justinfagnani:
Closing since we completely changed directive APIs in lit-html 2.0.

PR #615, opened 2018-11-08
Fix directive examples in docs

bashmish:
Seems like the documentation has incorrect examples on how to define a directive.

I tried to do it according to the docs and saw the stringified function in the rendered DOM. Then I looked the implementation of existing ones in the core and made mine in the same way (https://codepen.io/bashmish/pen/oQxBYZ) and it started working. I fixed examples accordingly.

### DIFF
--- a/docs/_guide/02-writing-templates.md
+++ b/docs/_guide/02-writing-templates.md
@@ -249,7 +249,7 @@ Directives are functions that accept some arguments for values and configuration
 ```javascript
 import {directive, html} from 'lit-html';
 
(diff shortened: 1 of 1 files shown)


bashmish:
Looks like this is related to this https://github.com/Polymer/lit-html/issues/604 and this https://github.com/Polymer/lit-html/pull/562

Feel free to close this PR if you think it is not the best moment to fix examples in docs back to the old syntax.

bashmish:
Closing since 0.13 is released now.

PR #624, opened 2018-11-13
Docs redesign

keanulee:
https://lit-html.polymer-project.org/


raydaly:
In lit-html/04-concepts.md there is an awkward sentence under Thinking Functionally: "This leads to model that's"

keanulee:
cc @arthurevans ^

PR #631, opened 2018-11-15
Fixes #630

### LINKED ISSUE #630: Self closing tags
As far as i can see we can't use self closing tags (<custom-element />) in HTML5. I tried using XHTML instead but that have problems with <script type="module"> so i was wondering is there any way we can do this while processing template inside html template literal ? I heard hyperHTML already doing this as you can see at [hyperHTML referance](https://github.com/WebReflection/hyperHTML/blob/7509fcac074217fbc1bcd01d72b0cbd2a122d7fe/cjs/hyper/render.js#L65) .

kaaninel:
Inside getHTML function we replace <tag /> to <tag></tag>

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -71,7 +71,8 @@ export class TemplateResult {
       let signatureValid = false;
       while (!signatureValid) {
         const signature = x.substr(sIndex + 1);
-        const qMatch = signature.match(/"/g); // make sure we are not inside of an attribute
+        const qMatch = signature.match(
+            /"/g);  // make sure we are not inside of an attribute
         if (qMatch) {
           if (qMatch.length % 2 === 0) {
             signatureValid = true;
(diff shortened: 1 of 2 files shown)


kaaninel:
But that's not developer friendly and a real bummer for using html without a pre-processor. This pull-request aims to get rid of this need so we can use lit only.

web-padawan:
By supporting self-closing syntax for custom elrments, we could confuse users trying to use them also outside of `lit-html`.

ruphin:
Discussion regarding the desirability of a PR should take place in the relevant issue (#630), and not the PR. This space is reserved for feedback on implementation details.

justinfagnani:
I believe there's a much simpler way of doing this with `String.replace`[1]. Even so, as discussed in #630 I'd rather leave this up to third-party extensions, which can define a new `html` tag with a simple subclass of `TemplateResult`.

[1]: See an example here: https://github.com/developit/htm/blob/c3182e44049301d7a9cc178576c3a2d281ecef3c/src/index.mjs#L38

PR #633, opened 2018-11-16
Reexport of TemplateProcessor and RenderOptions...

rivron:
...to ease the definition of new template processors.


justinfagnani:
Thanks @rivron!

Issue #635, opened 2018-11-17
[suggestion] A way to be intercept/delay node removal - `unmount` hook


prasannavl:
I've been using `lit-html` for a while, it manages to solve most uses cases with the well designed and thoughtful use of `directives` which are quite fantastic since you don't have resort to building abstractions on top to solve things, that are better solved in the lower layers (like what `react-transition-group` does because of the lack of the ability to do it in lower layers).

But doing so, one of the most common need that I found, that couldn't be quite easily addressed without writing a good amount of abstraction, was for doing something just before node removal, and scheduling node removal right after. 

For instance, `repeat` directive lets you easily manage lists. You can do an onappear effects/animations easily by abstracting them into their own component and handling it in the lower components. But currently, there's no way to detect an onremove, since lit-html just removes the nodes right away (and all that a DOM gives you would be a diconnected hook, which happens after the fact, and not before), there's nothing more you can do and you're force to handle a full fledged view models in the top layers. 

It would be great, if there was a way to detect nodes just before removal, especially while handling lists with `repeat` and the likes -- even better if it can do so with optional promise resolutions. I think having that will help solve a large class of problems a lot more elegantly.

I suspect the ideal approach would be a directive that can work in combination with repeat, providing a hook into remove, but I also think that might be difficult looking at the way `repeat` is implemented -- so perhaps an independent directive with a removal hook.

ruphin:
This same issue is also discussed in #527

prasannavl:
Ah, somehow I seem to have missed the dupes. Sorry about that. Closing this.
Also, related: https://github.com/Polymer/lit-html/issues/283

PR #636, opened 2018-11-17
impl for removal notification for repeat directive

prasannavl:
A very "naive" implementation of `onRemove` just to get the ball rolling. It currently has the following limitations:

- API shape has be thought through.
- ~`markerNodes` are also notified of removal, which is probably is better filtered off.~
- ~The `commit` callback probably should have an better behavior where deferring it is opt in, by calling something, or setting a flag -- similar to `event.preventDefault`, etc.~
- This can potentially integrate in 3 different ways: 
    - A separate directive entirely: Which means, abstracting the inner details of `repeat` to be more portable so that it can be reused.
    - Just make `repeat` have simpler `onremove` handler as per the initial stage of this commit, taking in node and `commit` functions as arguments
    - Make this fit into the context of `lit-html` in a larger scale, by exploring where else an `onremove` fits in context, and refactor the whole project (== lot of work ==)
 
Other things to do:
- ~refactor and cleanup - currently code is just duplicated - eg, `removeNode` .~
- ~linting fails due to formatting (causing test to fail)~

Please let me know if fits into your direction for the project. If it does, we can start ironing out the details, and work on this commit. That being said, it works as expected. 

Related issue: #635

### DIFF
--- a/src/directives/repeat.ts
+++ b/src/directives/repeat.ts
@@ -59,7 +59,7 @@ export class NodeRemovalToken {
   }
 
   commit() {
-    commitRemoveNode(this.container, this.target);
+    commitRemoveNode(this.target, this.container);
   }
 }
 
--- a/src/directives/repeat.ts
+++ b/src/directives/repeat.ts
@@ -62,7 +62,7 @@ export class NodeRemovalToken {
   commit() {
     if (this._isCommitted)
       return;
-    commitRemoveNode(this.container, this.target);
+    commitRemoveNode(this.target, this.container);
     this._isCommitted = true;
   }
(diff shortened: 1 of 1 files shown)


prasannavl:
Actually, this impl now, as it stands is incorrect. It leaves out the comment nodes. I'm still trying to understand the internals of lit-html. Any pointers would be great.

prasannavl:
I'm closing this, since I personally workaround this by implementing a custom component that provides notification - as that was way cleaner than meddling with `lit-html`. I suppose for on-removal, a larger design change is imperative to do it cleanly.

PR #638, opened 2018-11-18
Make it possible to extend the values that a NodePart can render

mariusGundersen:
Summary: This makes it possible to extend the types of values that a NodePart can render.

Instead of having a long `if{}else if{...}else{}` statement in the NodePart, pull all the code out and supply an array of tests and insert methods as part of the options. This way lit-html can be extended, through the options passed to the render function, so that it can handle new types of values passed to NodeParts. For example, if the [Promise support is moved to a directive](https://github.com/Polymer/lit-html/pull/555), this pull-request will make it possible to reapply the built-in support if that is what I want for my project. Or, I can add other value handlers, for example, this handler will serialize an object to a json string:

```ts
export const jsonHandler: NodePartValueHandler<Object> = {
  test(value: any): value is Object {
    return typeof value === 'object';
  },
  insert(part: NodePart, value: Object) {
    primitiveHandler.insert(part, JSON.stringify(value));
  }
}
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,5 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "typescript.referencesCodeLens.enabled": true,
+  "typescript.implementationsCodeLens.enabled": true
 }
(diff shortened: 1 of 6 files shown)


justinfagnani:
We really need to discuss major changes like this in issues before making a PR.

In this case, I'm not sure customization at the render point is the right semantics here. It seems like the template definition itself would be the safest place to extend with value handlers, because it has a better idea of what values to expect. I'm also a little concerned with the overhead from searching though a list on the hot path, rather than a fairly straightforward conditional cascade.

justinfagnani:
Closing this as out of date to the lit-next work.

PR #637, opened 2018-11-18
Fix Docs url

newbornfrontender:

PR #639, opened 2018-11-18
Fix broken documentation link

vinkla:
The documentation link in the `README.md` file is no longer working.


vinkla:
This is already being fixed in #637. Sorry for the noice!

PR #642, opened 2018-11-19
Handle CSS variables in styleMap

### LINKED ISSUE #632: Can't set CSS variables with styleMap directive
The styleMap directive does not work with custom CSS properties. Box 2 should be the same size as Box 1 when you click update.

https://codepen.io/osublake/pen/jQwVyQ

justinfagnani:
Fixes #632

### DIFF
--- a/src/directives/style-map.ts
+++ b/src/directives/style-map.ts
@@ -33,11 +33,20 @@ const styleMapCache = new WeakMap();
 const styleMapStatics = new WeakMap();
 
 /**
- * A directive that applies CSS properties. This must be used in the `style`
- * attribute and must be the only part used in the attribute. It takes the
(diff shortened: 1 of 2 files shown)


src/directives/styleMap.ts
```diff
@@ -60,6 +60,13 @@ export const styleMap = directive((styleInfo: StyleInfo) => (part: Part) => {
       ((part.committer.element as HTMLElement).style as any)[name] = null;
```
kevinpschaaf: I think you need the same `--` check here for removal as well.

While you're at it, seems like the `part.committer.element as HTMLElement).style` expression can be pulled out earlier and used for both places.

justinfagnani:
@kevinpschaaf PTAL

PR #644, opened 2018-11-20
Rename classMap and styleMap to kebab-case

### LINKED ISSUE #641: styleMap.js and classMap.js should be renamed to use kebab-case
These are the only files that use camelCase.

stramel:
Fixes #641 

Feel free to modify the Changelog update or drop it. I wasn't sure if I needed to do that and run gen-docs or not


justinfagnani:
Thanks @stramel!

PR #557, opened 2018-10-10
Fix example code for when() directive

arthurevans:

PR #647, opened 2018-11-21
Remove the when() directive

### LINKED ISSUE #511: `when` directive: false value should be optional

### LINKED ISSUE #552: Update when directive to handle any number of cases.

### LINKED ISSUE #565: when Directive puts its node in the wrong place when taken from the cache.
I have an app with various icons in the toolbar.  I only render any icons when the user is logged on and I only render a "reply" icon (used to return up a level) with the top level is not there.  These are both controlled using the `when` directive.
The first time I logon,  and then go below the top level the icons appear on the correct order, but the second time I activate the reply icon it has moved to the end of the list of icons.

I created a demo at home and proved it fails the same way in a cut down version.  Tried to load to stackblitz and it seems to have got stuck with dependencies. It should be here https://stackblitz.com/edit/js-zt2cnc

However the core of the problem is demonstrated in a single file - which is replicated below
```
import { LitElement, html } from '@polymer/lit-element/lit-element.js';
import {when} from 'lit-html/directives/when.js';

import '@material/mwc-icon/mwc-icon.js';

const appTitle = 'Test System';

class PasApp extends LitElement {
  static get properties() {
    return  {
      pageTitle: {type: String},
      isLoggedOn: {type: Boolean}
    };
  }
  constructor() {
    super();
    this.pageTitle = appTitle;
    this.isLoggedOn = false;
  }
  connectedCallback() {
    super.connectedCallback();
    this.removeAttribute('unresolved');
  }
  disconnectedCallback() {
    super.disconnectedCallback();
  }
  render() {
    return html`
      <style>
        :host {
          display: block;
        }
        .flexcontainer {
          display: flex;
          flex-direction: row;
        }
        .filler {
          flex: 1;
          text-align: center;
          align-self: flex-start;
        }
        mwc-icon {
          cursor: pointer;
        }

      </style>
      <div class="flexcontainer">
      <div class="filler">${this.pageTitle}</div>
        ${when(this.isLoggedOn, () => html`
          <mwc-icon @click="${this._resetTitle}">today</mwc-icon>
          <mwc-icon @click="${this._changeTitle}">date_range</mwc-icon>
          ${when(this.pageTitle !== appTitle, () => html`
            <mwc-icon @click="${this._logoff}">reply</mwc-icon>
          `, () => html``)}
          <mwc-icon id="home" @click="${this._gotoHome}">home</mwc-icon>
        `, () => html``)}
        <mwc-icon @click="${this._logon}">menu</mwc-icon>
      </div>
    `;
  }
  _changeTitle(e) {
    e.stopPropagation();
    this.pageTitle = 'Some Different Text'
  }
  _logoff(e) {
    e.stopPropagation();
    this.isLoggedOn = false;
  }
  _logon(e) {
    e.stopPropagation();
    this.isLoggedOn = true;
  }
  _resetTitle(e) {
    e.stopPropagation();
    this.pageTitle = appTitle;
  }
}
customElements.define('test-app',PasApp);
```

### LINKED ISSUE #592: `when` directive example doesn't work

justinfagnani:
Resolves #552 after #646 adds the `cache()` directive

Closes #608
Closes #593
Closes #560
Closes #557
Closes #592
Closes #565
Closes #552
Closes #511


justinfagnani:
Closes #608 
Closes #593 
Closes #560 
Closes #557 
Closes #592 
Closes #565 
Closes #552  
Closes #511

TimvdLippe:
@justinfagnani If you want to automatically close all these issues, they should be added in the PR description. I have updated the description to include them :smile:

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Added -->
 ### Changed
 * [Breaking] Renamed `classMap` and `styleMap` files to kebab-case. ([#644](https://github.com/Polymer/lit-html/pull/644))
-<!-- ### Removed -->
+### Removed
+* [Breaking] Removed the when() directive.
```
kevinpschaaf:
```suggestion
* [Breaking] Removed the when() directive.  Users may achieve similar behavior by wrapping a ternary with the `cache` directive.
```

PR #657, opened 2018-11-25
Add runAsync() directive

### LINKED ISSUE #649: Add runAsync() directive factory.
This will allow the creation of directives that spawn async work and render templates based on the state of the work.

See prototype here: https://github.com/PolymerLabs/async-demos/blob/master/packages/await-io/src/run-async.ts

justinfagnani:
Adds a `runAsync` directive.

This directive takes a key object, task function, and a set of templates to render based on the state of the task (initial, pending, success and failure).

The API is slightly different than what I referenced in #649: This is a plain directive, and not a factory. A factory can be created by currying this directive, or a directive tied to a specific task function can be created via composition. The templates have also moved to a object with named properties.

Fixes #649

### DIFF
--- a/src/directives/run-async.ts
+++ b/src/directives/run-async.ts
@@ -27,6 +27,9 @@ const hasAbortController = typeof AbortController === 'function';
 
 const runs = new WeakMap<Part, AsyncRunState>();
 
+export type Task<K> = (key: K, options: {signal?: AbortSignal}) =>
+    Promise<unknown>;
+
 /**
  * Runs an async function whenever the key changes, and calls one of several
  * lit-html template functions depending on the state of the async call:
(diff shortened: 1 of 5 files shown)


src/test/directives/run-async_test.ts
```diff
@@ -0,0 +1,167 @@
...
+    assert.equal(stripExpressionMarkers(container.innerHTML), 'Success: test');
+  });
+
+  test('renders pending then initial templates', async () => {
+    renderTest('test', async (_: string) => {
+      throw new InitialStateError();
+    });
+    assert.equal(stripExpressionMarkers(container.innerHTML), 'Pending');
+    await 0;
+    assert.equal(stripExpressionMarkers(container.innerHTML), 'Initial');
```
kevinpschaaf: Make `runAsync` wait a microtask after running the async fn to get a shot at rendering initial before pending?

src/test/directives/run-async_test.ts
```diff
@@ -0,0 +1,167 @@
...
+import {render} from '../../lib/render.js';
+import {html} from '../../lit-html.js';
+import {stripExpressionMarkers} from '../test-utils/strip-markers.js';
+
+const assert = chai.assert;
+
+const hasAbortController = typeof AbortController === 'function';
+const testIfHasAbortController = hasAbortController ? test : test.skip;
+
+suite('runAsync', () => {
```
kevinpschaaf: Add tests:
* Multiple dependencies
* Add use of props from render closure in template result functions, and ensure they re-evaluate correctly each render
* Ensure `pending-state` event is not fired in initial state

src/directives/run-async.ts
```diff
@@ -0,0 +1,153 @@
...
+    const abortSignal =
+        hasAbortController ? abortController!.signal : undefined;
+    let resolvePending!: () => void;
+    let rejectPending!: (e: Error) => void;
+    const pendingPromise = new Promise((res, rej) => {
+      resolvePending = res;
+      rejectPending = rej;
+    });
+    const promise = task(key, {signal: abortSignal});
+    // The state is immediately 'pending', since the function has been
```
Westbrook: Does this mean that even if the `promise` can't be run with the data available it will take always take two renders to get to `initial`? One pass for `pending` and then a second (assuming `throw new InitialStateError()`) to trigger `initial`. If so, that seems confusing to end users.

This may be conflating the responsibilities of `lit-html` and a component system (i.e. `LitElement) that relies on it, but I would expect more mainstream usage of this to rely on a key that powered the promise in someway so an initial call where `key === undefined` or `key === ''` would provide the `initial` render template in a single render pass.

In this case, would you expect there to be additional wrapping of or an alternative to this directive to achieve that functionality? or, would it make sense to include that by default?

src/directives/run-async.ts
```diff
@@ -0,0 +1,153 @@
...
+              composed: true,
+              bubbles: true,
+              detail: {promise: pendingPromise}
+            }));
+      }
+    })();
+  }
+
+  // If the promise has not yet resolved, set/update the defaultContent
+  if ((currentRunState === undefined || currentRunState.state === 'pending') &&
```
dbatiste: Hi @justinfagnani : I am experimenting with this a bit since you mentioned it last week.  I notice that the pending state is not applied on first key/dependency change after a previous success state - it remains in the success state.  Another key change is required for the pending state to be applied.  

I could be doing something incorrectly, but I think : on line#93 `runs.set(part, runState);` is called to set a new pending state, but `currentRunState` is still the previous success state, so `pending()` is not called here.  Once the directive is run again (ex. second key-stroke in the search demo), `currentRunState` is initialized with the most recent pending state, and then `pending()` is called here.  It is perhaps not super noticeable when there are many changes (like `input` event).  The absent intermediate pending state was fairly obvious in my case where there aren't multiple key changes quickly in succession.

Westbrook: I would agree with this. I wonder what the desired benefits of this test being anything more than `if (typeof pending === 'function') {`? I've been playing with the code shaped this way at https://stackblitz.com/edit/lit-element-run-async-2?file=run-async.ts

src/directives/run-async.ts
```diff
@@ -0,0 +1,153 @@
...
+              part.commit();
+            }
+          }
+        });
+
+    (async () => {
+      // Wait a microtask for the initial render of the Part to complete
+      await 0;
+      const currentRunState = runs.get(part);
+      if (currentRunState === runState && currentRunState.state === 'pending') {
```
Westbrook: If you only dispatch a promise when `currentRunState.state === 'pending'` an action that immediate throws `new InitialStateError()` won't be provided the `pendingPromise` reference to manage the reject that happens on [line 108](https://github.com/Polymer/lit-html/pull/657/files#diff-47605b25f29d88148b59b1949601eb08R108). Should this be updated to:
```suggestion
      if (currentRunState === runState && (currentRunState.state === 'pending' || currentRunState.state === 'initial')) {
```

justinfagnani:
Closing in favor of reactive controllers in lit-next.

PR #666, opened 2018-11-29
Update guard to handle multiple checked values and undefined

### LINKED ISSUE #648: Update guard() to support multiple values.
The `guard` name has been confusing for users. `memo` seems to be common enough with it's addition to React, that it should be a bit better.

We also want to support multiple checked values.

justinfagnani:
Fixes #648

### DIFF
--- a/src/directives/guard.ts
+++ b/src/directives/guard.ts
@@ -14,34 +14,61 @@
 
 import {directive, Part} from '../lit-html.js';
 
-const previousExpressions = new WeakMap<Part, any>();
+const previousValues = new WeakMap<Part, unknown>();
 
 /**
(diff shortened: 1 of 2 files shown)


src/directives/guard.ts
```diff
@@ -14,34 +14,61 @@
...
+      } else if (
+          previousValue === value &&
+          (value !== undefined || previousValues.has(part))) {
+        // Dirty-check non-arrays by identity
         return;
       }
 
-      part.setValue(valueFn());
-      previousExpressions.set(part, expression);
+      part.setValue(f());
```
kevinpschaaf: Pass deps to the function

PR #673, opened 2018-11-30
Add nothing sentinel

### LINKED ISSUE #652: Add `nothing` sentinel value
Rendering `nothing` to a Part will cause it to clear. Useful to make sure that elements with ShadowRoots can trigger `<slot></slot>` fallback content:

```js
import {html, nothing} from 'lit-html';

html`<my-element>${condition ? 'something' : nothing}</my-element>`
```

ruphin:
Fixes #652

Also added a small test for the `noChange` sentinel.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -14,7 +14,7 @@
 
 import {isDirective} from './directive.js';
 import {removeNodes} from './dom.js';
-import {noChange, Part} from './part.js';
+import {noChange, nothing, Part} from './part.js';
 import {RenderOptions} from './render-options.js';
 import {TemplateInstance} from './template-instance.js';
 import {TemplateResult} from './template-result.js';
@@ -196,6 +196,9 @@ export class NodePart implements Part {
       this._commitNode(value);
     } else if (Array.isArray(value) || value[Symbol.iterator]) {
       this._commitIterable(value);
+    } else if (value === nothing) {
+      this.value = undefined;
+      this.clear();
     } else {
       // Fallback, will render the string representation
       this._commitText(value);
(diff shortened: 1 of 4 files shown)


src/lib/parts.ts
```diff
@@ -196,6 +196,9 @@ export class NodePart implements Part {
       this._commitNode(value);
     } else if (Array.isArray(value) || value[Symbol.iterator]) {
       this._commitIterable(value);
+    } else if (value === nothing) {
+      this.value = undefined;
```
justinfagnani: keeping `nothing` as the value would make it cleat that the part is empty

ruphin: I opted to set to undefined because that's the initial state, which seemed reasonable to me. 

If we set the value to `nothing` we can test against it, but I think calling `clear` on an empty NodePart is almost free anyway, so I'd rather just always clear and reduce bytesize.

Now that I think about it, `value` probably compresses better than `undefined`, and it doesn't really make any difference what we use as long as we override the previous value, so I'll change it to save some more bytes.

ruphin:
Well this is awkward, calling `clear()` doesn't remove the markers. I'm going to have to rethink how to do this.

justinfagnani:
`clear()` should not remove the markers. They're needed for future updates.

ruphin:
I was worried that the markers would prevent rendering the fallback content. Turns out they don't: https://codepen.io/ruphin/pen/bQZdYN?editors=1010

That means this implementation will work. I will update the tests to verify that no non-comment nodes remain after rendering `nothing`.

src/lib/parts.ts
```diff
@@ -196,6 +196,9 @@ export class NodePart implements Part {
       this._commitNode(value);
     } else if (Array.isArray(value) || value[Symbol.iterator]) {
       this._commitIterable(value);
+    } else if (value === nothing) {
+      this.value = nothing;
```
stramel: If value is nothing do we need to set value to nothing?

PR #674, opened 2018-12-01
Make API docs slightly more usable

justinfagnani:
This is an improvement, but not a great final end point.

I am concerned about the stabilities of URLs, so we should figure out a strategy soon.


docs/_api-readme.md
```diff
@@ -0,0 +1,27 @@
...
+- [render](./globals.html#render)
+
+## Directives
+Import from individual modules in the `lit-html/directives/` folder:
+
+```js
+import {repeat} from 'lit-html/directives/repeat.js';
+```
+
+- [asyncAppend](./globals.html#asyncAppend)
```
arthurevans: Not sure why, but the actual anchors TypeDoc is generating are lowercase, so these links are broken for camelCase directives (asyncReplace, classMap) although they work fine for the others (guard, repeat).

arthurevans:
I think I'm going to merge this and fix links locally so I can publish...?

PR #675, opened 2018-12-01
Update directive docs

### LINKED ISSUE #712: Update directives docs in template reference
Make sure list is up-to-date and examples, text reflect latests changes.

I have sections ready for styleMap and classMap, in case anyone else takes this on.

arthurevans:
Expanded to include the rest of the directives.

Fixes #712.


docs/_guide/03-template-reference.md
```diff
@@ -309,9 +309,14 @@ const myTemplate = () => html`
 
 ### guard
 
-`guard(expression, valueFn)`
+`guard(expressions, valueFn)`
```
justinfagnani: I think `values` makes more sense as the first argument, as the function doesn't actually receive expressions, but the results of those expressions as values. I just called the parameters `value` ('cause it can be just one object) and `f` in the code.

docs/_guide/03-template-reference.md
```diff
@@ -309,9 +309,14 @@ const myTemplate = () => html`
...
+
+Avoids re-evaluating an expensive template function (`valueFn`) unless one of the identified expressions changes identity. Returns the value of `valueFn`, which may be cached.
+
+The `expressions` argument can either be a single (non-array) expression, or an array of multiple expressions to monitor. 
+
+The `guard` directive caches the last-known value of `valueFn`, and only re-evaluates `valueFn` if the identity of any of the expressions changes (for example when a primitive changes value or when an object reference changes).
```
justinfagnani: It doesn't actually cache the last result of `f`, it just doesn't update the DOM if it doesn't call `f`

arthurevans: OK. But it's caching the generated DOM, then? Not sure whether it's returning nodes or a TemplateResult here...

arthurevans: Oh, or... It's a directive, it just doesn't call setValue?

docs/_guide/03-template-reference.md
arthurevans: OK. Would it be correct to call the function a template function? Maybe just call it templateFunction here?

docs/_guide/03-template-reference.md
ruphin: Yes; if the `expressions` don't change, it just does nothing. Only if they change, it will evaluate the value function and update the DOM with the resulting values.

docs/_guide/03-template-reference.md
ruphin: The second argument (`valueFn`) is a function that returns any arbitrary value that lit-html can render, which includes templates, but also strings, arrays, etc. I think we generally use `value` to describe this type of object throughout lit-html.

arthurevans: Thanks @ruphin. Yep, I was just checking on this.

`guard(value, valueFn)` is kind of confusing, but if we try to make it less vague, we end up with `guard(valueOrValuesToMonitor, functionThatReturnsSomethingToRender)` which is just ridiculous ...

kevinpschaaf: I like `dependencies` as the first argument, fwiw.  They are what the `valueFn` depends on.

arthurevans:
Updated other directive docs, too. Staged here:

https://updated-directive-docs-dot-polymer-lit-html.appspot.com/guide/template-reference#built-in-directives

kenchris:
Some of the code samples are missing syntx highlighting

kenchris:
screenshot_20181221-215808

docs/_guide/03-template-reference.md
```diff
@@ -254,44 +254,126 @@ Directives are functions that can extend lit-html by customizing the way a bindi
...
+)}`
+```
+
+When lit-html re-renders a template, it only updates the modified portions: it doesn't create or remove any more DOM than it needs to. But when you switch from one template to another, lit-html needs to remove the old DOM and render a new DOM tree. 
+
+The `cache` directive caches the generated DOM for a given binding and input template. In the example above, it would cache the DOM for both the  `summaryView` and `detailView` templates. When you switch from one view to another, lit-html just needs to swap in the cached version of the new view, and and update it with the latest data.
+
+### classMap
+
+`class=${classMap(classObj)}`
```
justinfagnani: maybe add the tag to the example too

docs/_guide/03-template-reference.md
```diff
@@ -254,44 +254,126 @@ Directives are functions that can extend lit-html by customizing the way a bindi
...
 ```
 
-If no `keyFn` is provided, `repeat` will perform similar to a simple map of
-items to values, and DOM will be reused against potentially different items.
 
 ### ifDefined
 
 `ifDefined(value)`
 
+Location: attribute bindings
```
justinfagnani: might be nice to update the example to show that this _can_ be used as a partial value of an attribute. It's pretty nice for URLs, so maybe something like ``

docs/_guide/03-template-reference.md
```diff
@@ -320,87 +414,114 @@ import { guard } from 'lit-html/directives/guard';
...
-In this case, items are mapped over only when the array reference changes.
+In this case, the `immutableItems` array is mapped over only when the array reference changes.
 
-### until
-
-`until(...values)`
+### repeat 
 
-Renders one of a series of values, including Promises, to a Part.
+`repeat(items, keyfn, template)`
```
justinfagnani: We should show two signatures here, adding: `repeat(items, template)` without the keyFn

docs/_guide/03-template-reference.md
```diff
@@ -320,87 +414,114 @@ import { guard } from 'lit-html/directives/guard';
...
+### until
+
+`until(...values)`
+
+Location: any
+
+Renders placeholder content until the final content is available. 
+
+Takes a series of values, including Promises. Values are rendered in priority order, 
+with the first argument having thehighest priority and the last argument having the
```
justinfagnani:
```suggestion
 with the first argument having the highest priority and the last argument having the 
```

docs/_guide/03-template-reference.md
```diff
@@ -320,87 +414,114 @@ import { guard } from 'lit-html/directives/guard';
...
+
+`until(...values)`
+
+Location: any
+
+Renders placeholder content until the final content is available. 
+
+Takes a series of values, including Promises. Values are rendered in priority order, 
+with the first argument having thehighest priority and the last argument having the 
+lowest priority. If a value is a Promise, a lower-priority values will be rendered until it resolves.
```
justinfagnani:
```suggestion
 lowest priority. If a value is a Promise, a lower-priority value will be rendered until it resolves.
```

PR #679, opened 2018-12-02
typo "create" not "creat"

Lonniebiz:
If a comment has a typo, should I bother to correct it? Or is such a small modification more trouble than its worth to approve? Please advise.


justinfagnani:
This got fixed in #690. Thanks for the PR!

PR #686, opened 2018-12-08
Add instructions for working with the site

arthurevans:


keanulee:
I would suggest putting docs instructions in docs/README.md instead, but I'll leave that for someone else to decide.

arthurevans:
Works for me...

arthurevans:
Addressed this in #693, instead.

PR #693, opened 2018-12-12
Better API docs

### LINKED ISSUE #684: API doc for shady-render.render hides regular render docs.
Currently, the API docs for `render` show the first sentence from the `render` JSDoc, followed by the JSDoc for shady-render.render.

arthurevans:
Fixes some of the existing issues with API docs:

- Uses typdoc-plugin-external-module-name to consolidate multiple sub-modules into the single entry-point modules (lit-html, shady-render).

- Removes API doc for directives, which is just messed up at this point, points to template reference instead.

- Reverts to module mode, which prevents shady-dom.render from overwriting the docs for lit-html.render. (Fixes #684)

- Uses the minimal template, which places the navigation links on the left (more like the guides), creates one page per module, and seems slightly cleaner.

Staged here: https://20181212t125505-dot-polymer-lit-html.appspot.com/api/index.html


package.json
```diff
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "tsc",
     "checksize": "rollup -c ; cat lit-html.bundled.js | gzip -9 | wc -c ; rm lit-html.bundled.js",
-    "gen-docs": "typedoc --readme ./docs/_api-readme.md --mode file --ignoreCompilerErrors --excludeNotExported --excludePrivate --out ./docs/api --gaID UA-39334307-23 ./src/lit-html.ts ./src/lib/*.ts ./src/directives/*.ts",
```
keanulee: Maybe we should move this script to docs/package.json

package.json
```diff
@@ -44,6 +44,7 @@
     "rollup-plugin-terser": "^1.0.1",
     "tslint": "^5.11.0",
     "typedoc": "^0.13.0",
+    "typedoc-plugin-external-module-name": "^1.1.3",
```
keanulee: Consider moving these deps to docs/package.json

src/lit-html.ts
```diff
@@ -12,6 +12,21 @@
...
+  * 
+  * -  [[html]] 
+  * -  [[svg]] 
+  * -  [[render]]
+  *   
+  * @module lit-html
+  * @preferred
+  */
+
+/** ignore this comment, which keeps typedoc from misplacing the module docs. */
```
keanulee: "Do not remove this comment - it keeps typedoc ..."

docs/_api/copy_api.sh
```diff
@@ -0,0 +1,13 @@
+#
```
keanulee:
```
# This script copies the TypeScript source into a staging directory for Typedoc
# and removes the license header for typedoc-plugin-external-module-name.
```

package.json
arthurevans: I encountered some problems when I tried to do so before, but I think I've figured them out. Do you know if anything else relies on this gen-docs script, or can I remove it from the top-level package.json entirely?

src/lit-html.ts
arthurevans: 👍

docs/_api/copy_api.sh
arthurevans: 👍

arthurevans:
Thanks @keanulee. Will fix formatting.

I can't find the package.json changes you mentioned. Do you mean the devDependencies?

keanulee:
package-lock, the auto-generated changes from npm. Do you have node 8 (or something not node 10)?

keanulee:
I think you need to `git merge master` and re-run `npm run format` to make the travis-ci/pr pass.

PR #696, opened 2018-12-14
Cleanup Template Preperation

jridgewell:
By storing the `endNode`'s index (instead of the `startNode`), it becomes easier to reason about the `COMMENT_NODE` path.

The only reason it's complicated now is because we attempt to remove the node, provided there's some `nextSibling`. If we removed that, `COMMENT_NODE` would become dead simple (at the expensive of a unnecessary extra node).

### DIFF
--- a/src/lib/template.ts
+++ b/src/lib/template.ts
@@ -58,7 +58,10 @@ export class Template {
           133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */,
           null as any,
           false);
-      let lastStartIndex = 0;
+      // Keeps track of the last index associated with a part. We try to delete
+      // unnecessary nodes, but we never want to associate two different parts
+      // to the same index. They must have a constant node between.
+      let lastPartIndex = 0;
       while (walker.nextNode()) {
         index++;
         const node = walker.currentNode as Element;
@@ -105,8 +108,6 @@ export class Template {
             const parent = node.parentNode!;
             const strings = nodeValue.split(markerRegex);
             const lastIndex = strings.length - 1;
-            // We have a part for each match found
-            partIndex += lastIndex;
             // Generate a new text node for each literal section
             // These nodes are also used as the markers for node parts
             for (let i = 0; i < lastIndex; i++) {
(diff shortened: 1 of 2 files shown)


src/polyfills/template_polyfill.ts
```diff
@@ -46,7 +46,8 @@ export const initTemplatePolyfill = (forced = false) => {
   };
 
   const capturedCreateElement = Document.prototype.createElement;
-  Document.prototype.createElement = function createElement(tagName: string, options?: ElementCreationOptions) {
+  Document.prototype.createElement = function createElement(
```
justinfagnani: you can revert this

src/lib/template.ts
```diff
@@ -104,61 +99,49 @@ export class Template {
...
-            // If we don't have a nextSibling add a marker node.
-            // We don't have to check if the next node is going to be removed,
-            // because that node will induce a new marker if so.
+            lastStartIndex = index;
+            this.parts.push({type: 'node', index});
+            // If we don't have a nextSibling, keep this node.
+            // Else, remove it to save memory.
             if (node.nextSibling === null) {
-              parent.insertBefore(createMarker(), node);
+              node.nodeValue = '';
```
justinfagnani: what's the reason to clear the node value?

src/lib/template.ts
```diff
@@ -104,61 +99,49 @@ export class Template {
...
             }
-            this.parts.push({type: 'node', index: index++});
-            nodesToRemove.push(node);
-            // If we don't have a nextSibling add a marker node.
-            // We don't have to check if the next node is going to be removed,
-            // because that node will induce a new marker if so.
+            lastStartIndex = index;
+            this.parts.push({type: 'node', index});
+            // If we don't have a nextSibling, keep this node.
+            // Else, remove it to save memory.
```
justinfagnani: We remove the node to reduce clone costs as well, so maybe generalize the reason.

src/lib/template.ts
```diff
@@ -56,15 +56,10 @@ export class Template {
           ,
           null as any,
           false);
-      // The actual previous node, accounting for removals: if a node is removed
-      // it will never be the previousNode.
-      let previousNode: Node|undefined;
-      // Used to set previousNode at the top of the loop.
-      let currentNode: Node|undefined;
+      let lastStartIndex = 0;
```
justinfagnani: add comment

src/lib/template.ts
jridgewell: This is what `createMarker` would do, but we can save a node instance and removal by clearing the value.

PR #703, opened 2018-12-16
Fix unsafeHTML

### LINKED ISSUE #702: Alternating unsafeHTML and regular string not working
This is either a tiny tiny bug or just something I shouldn't do! :) 

Let say I have this: `<p>${ this.getMessage() }</p>` and what it returns maybe unsafeHTML() result or a string, then changing from unsafeHTML to a string, the rendered template doesn't get updated, the result from unsafeHTML is stuck.

Is it something I should avoid? I can also return normal strings as unsafeHTML which works, but probably have a cost I don't know (**all** localized strings in my apps use this principle).

ruphin:
This fixes #702 - and adds a regression test for this case.

Unfortunately fixing the problem requires keeping a cache for previously rendered `HTMLTemplateElements` and `DocumentFragments`, so we can compare vs the current `Part.value`. Without this there is no way to check if the part rendered another value after the previous unsafeHTML render. This means there is some additional memory overhead, but I don't see how it can be avoided.

I added a cache for `HTMLTemplateElement` to ensure the same input string is never parsed twice. Previously, this directive would build duplicate templates if the same input was used across multiple instances. This will generally speed up execution, depending on how often similar strings are used. This also obviates the need for an 'isPrimitive' check.

### DIFF
--- a/src/directives/unsafe-html.ts
+++ b/src/directives/unsafe-html.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import { directive, NodePart, Part } from '../lit-html.js';
+import {directive, NodePart, Part} from '../lit-html.js';
 
 /**
  * Renders the result as HTML, rather than text.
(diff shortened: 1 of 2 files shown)


justinfagnani:
Oops... I tried to push a change I med to this PR, but I accidentally brought in too many commits. Sorry. I'll clean it up

justinfagnani:
Interesting. I think we should open a unsafeHTML performance issue to discuss this.

My first instinct is that re-rendering the _same_ HTML across alternating updates isn't likely to be a common case, and someone doing that amount of perf-critical low-level HTML management isn't really taking advantage of lit-html at that point and should do something custom like write their own directive or hand lit-html a DOM node that they manually manage.

PR #704, opened 2018-12-16
Fix formatting error

ruphin:
This fixes a formatting error on `master`. It's causing CI on master to fail, which means CI for all PRs also fails at the moment.


justinfagnani:
Looks like this was already fixed in #693

PR #718, opened 2018-12-20
fixes#717 first try using closure

osdevisnot:
Currently I am getting these numbers:

image


justinfagnani:
This is a worse number. Doesn't seem like worth doing.

PR #719, opened 2018-12-20
Use data property

jridgewell:
Smaller, and little faster.

Unfortunately, it dirties the type checking a bit. It'd be nice if typescript could detect `nodeType === 3` as a type guard for `Text`... Right now the best we could do is [user defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) but that's too big.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -230,7 +230,7 @@ export class NodePart implements Part {
       // If we only have a single text node between the markers, we can just
       // set its value, rather than replacing it.
       // TODO(justinfagnani): Can we just check if this.value is primitive?
-      node.textContent = value;
+      (node as Text).data = value;
     } else {
       this._commitNode(document.createTextNode(
           typeof value === 'string' ? value : String(value)));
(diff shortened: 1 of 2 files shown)


justinfagnani:
Is `.data` really faster? Interesting...

justinfagnani:
@jridgewell where's `.data` documented? I can't find anything on MDN.

jridgewell:
> Is .data really faster? Interesting...

http://jsbench.github.io/#b266690cc58d684e9f9eb7de3158437e. It's because `CharacterData` holds the `.data`, while `Node` holds the `.nodeValue`, so we'll be one step closer in the prototype chain.

> where's .data documented? I can't find anything on MDN.

MDN has poor documentation for DOM things. It's spec'd at https://dom.spec.whatwg.org/#characterdata, and both `Text` and `Comment` inherit from `CharacterData`

justinfagnani:
@jridgewell seems like `.data` is _much_ faster than `.textContent` too. Nice :)

I merged and formatted your branch, so it's ready to go, but I need permission to push.

justinfagnani:
I went and merged this manually. Thanks!

PR #720, opened 2018-12-20
Don't set value after committing the node

anonrose:
When you [commit the node](https://github.com/polymer/lit-html/blob/c173c59ca612df0ad6767428a964861ef684edef/src/lib/parts.ts#L216) it checks against `this.value` and will assign it if it's not already that value so there's no need to do it again.


anonrose:
I signed it!

PR #728, opened 2018-12-26
optimize TemplateResult getHTML

sheweichun:


Lodin:
Array solution is way slower than string concatenation (almost twice in Chrome):
http://jsbench.github.io/#5b3e080f0d0b693284ef9b5a8e4fed86

ruphin:
Nice catch

sheweichun:
I tried it,you are right

PR #731, opened 2018-12-29
Minor refactoring

vicb:
- uses the TS `constructor(public prop: Type)` shortcuts,
- drops file extension in import (seems like formatting re-ordered some of the import after that).

### DIFF
--- a/src/directives/async-append.ts
+++ b/src/directives/async-append.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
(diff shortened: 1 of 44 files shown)


vicb:
Sure, thanks for the info.
It might make sense to enforce those with tslint rules ?

justinfagnani:
Added a tslint rule for constructor parameter initializers, but couldn't find one for import specifier extensions. I updated CONTRIBUTING.md to mention both of these.

Thanks @vicb!

PR #741, opened 2019-01-10
Remove all uses of "any" outside of test code.

rictic:


src/directives/style-map.ts
```diff
@@ -70,7 +70,8 @@ export const styleMap = directive((styleInfo: StyleInfo) => (part: Part) => {
   for (const name in oldInfo) {
     if (!(name in styleInfo)) {
       if (name.indexOf('-') === -1) {
-        (style as any)[name] = null;
+        // tslint:disable-next-line:no-any
+        (style as any)[name as keyof StyleInfo] = null;
```
justinfagnani: If you cast style to any, you shouldn't need to cast name as well. Doing it for docs?

src/directives/unsafe-html.ts
```diff
@@ -47,7 +47,7 @@ export const unsafeHTML = directive((value: any) => (part: Part): void => {
   }
 
   const template = document.createElement('template');
-  template.innerHTML = value;
+  template.innerHTML = String(value);
```
justinfagnani: I'd cast instead. The `innerHTML` setter does toString if necessary.

src/directives/until.ts
```diff
@@ -27,6 +27,11 @@ interface AsyncState {
 
 const _state = new WeakMap<Part, AsyncState>();
 
+function isThenable(value: unknown): value is PromiseLike<unknown> {
```
justinfagnani: I pretty uniformly prefer const arrow functions

justinfagnani: This might be extra bytes for nicer type guards. What's the size difference (`npm run checksize`)?

src/lib/directive.ts
```diff
@@ -42,12 +45,14 @@ export type DirectiveFn = (part: Part) => void;
...
+    ((...args: unknown[]) => {
       const d = f(...args);
       directives.set(d, true);
       return d;
-    }) as unknown as F;
+    }) as F;
 
-export const isDirective = (o: any) =>
-    typeof o === 'function' && directives.has(o);
+export function isDirective (o: unknown): o is DirectiveFn {
```
justinfagnani: why not const arrow function?

src/lib/parts.ts
```diff
@@ -84,20 +91,20 @@ export class AttributeCommitter {
   commit(): void {
     if (this.dirty) {
       this.dirty = false;
-      this.element.setAttribute(this.name, this._getValue());
+      this.element.setAttribute(this.name, String(this._getValue()));
```
justinfagnani: don't wrap in `String()`. has a cost for values that are already strings

src/directives/style-map.ts
rictic: True, done

src/directives/unsafe-html.ts
rictic: Done

src/directives/until.ts
rictic: 8 bytes growth total from this PR. 3218 -> 3226

It looks like this file isn't tracked by that script though, as it's a directive, and that script just tracks the main lit-element.js file.

rictic: Interesting. const arrow functions produce slightly smaller bundled output.

I'll meet you halfway: const arrow functions with braces. Without the braces it's slightly harder to skim, and it seems to have a 0 byte effect on the output.

stramel: I was wondering if there would be a size reduction in a `isFunction` helper?

src/directives/async-replace.ts
```diff
@@ -33,7 +33,7 @@ import {directive, NodePart, Part} from '../lit-html.js';
  *     value. Useful for generating templates for each item in the iterable.
  */
 export const asyncReplace = directive(
-    <T>(value: AsyncIterable<T>, mapper?: (v: T, index?: number) => any) =>
+    <T>(value: AsyncIterable<T>, mapper?: (v: T, index?: number) => T) =>
```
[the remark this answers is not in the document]
rictic: Done

src/lib/parts.ts
rictic: True, done

src/directives/async-append.ts
```diff
@@ -33,7 +33,7 @@ import {createMarker, directive, NodePart, Part} from '../lit-html.js';
  */
 export const asyncAppend = directive(
     <T>(value: AsyncIterable<T>,
-        mapper?: (v: T, index?: number) => any) => async (part: Part) => {
+        mapper?: (v: T, index?: number) => T) => async (part: Part) => {
```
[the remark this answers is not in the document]
rictic: Fair. I spent a bit longer trying to be more precise until I realized that the only reason we care about this type is for consistency between the async iterable and the mapping funciton. Done.

src/directives/async-replace.ts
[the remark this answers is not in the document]
rictic: Done

src/directives/until.ts
rictic: Maybe? Need to get a size check implemented for all the code, including directives

rictic: Now just 3 bytes growth: 3221

justinfagnani: @stramel there's also a speed tradeoff to balance against byte size. An helper that wraps `typeof` requires a function call or the inliner to kick in on some pretty hot code paths.

stramel: Ah, good point. It will be easier to test the trade-offs once benchmarking and size-checks are both built out.

src/directives/async-append.ts
```diff
@@ -67,7 +67,9 @@ export const asyncAppend = directive(
         // we accept a mapper function. This is especially convenient for
         // rendering a template for each item.
         if (mapper !== undefined) {
-          v = mapper(v, i);
+          // This is safe because T must otherwise be treated as unknown by
+          // the rest of the system.
+          v = mapper(v, i) as T;
```
justinfagnani: I really wish TypeScript had a way to re-declare the type of a variable for these purposes.

src/lib/parts.ts
```diff
@@ -24,9 +24,21 @@ import {TemplateInstance} from './template-instance.js';
...
+// https://tc39.github.io/ecma262/#sec-typeof-operator
+export type Primitive = null|undefined|boolean|number|string|Symbol|bigint;
+export const isPrimitive = (value: unknown): value is Primitive => {
+  return (
+      value === null ||
+      !(typeof value === 'object' || typeof value === 'function'));
+};
+
+const isIterable = (value: unknown): value is Iterable<unknown> => {
+  return !!(
```
justinfagnani: why the `!!`?

src/lib/parts.ts
```diff
@@ -56,7 +68,7 @@ export class AttributeCommitter {
     return new AttributePart(this);
   }
 
-  protected _getValue(): any {
+  protected _getValue(): unknown {
```
justinfagnani: `unknown` should just be `string`

src/lib/parts.ts
```diff
@@ -66,8 +78,7 @@ export class AttributeCommitter {
       const part = this.parts[i];
       if (part !== undefined) {
         const v = part.value;
-        if (v != null &&
-            (Array.isArray(v) || typeof v !== 'string' && v[Symbol.iterator])) {
+        if (isIterable(v)) {
```
justinfagnani: this is a potentially hot code path... is this just for readability?

src/lib/parts.ts
```diff
@@ -240,7 +251,7 @@ export class NodePart implements Part {
 
   private _commitTemplateResult(value: TemplateResult): void {
     const template = this.options.templateFactory(value);
-    if (this.value && this.value.template === template) {
+    if (this.value instanceof TemplateInstance) {
```
justinfagnani: This shouldn't be safe to do. Why'd you remove the other check? Is a test failing? If not we should add one.

src/lib/parts.ts
```diff
@@ -198,7 +209,7 @@ export class NodePart implements Part {
       this._commitTemplateResult(value);
     } else if (value instanceof Node) {
       this._commitNode(value);
-    } else if (Array.isArray(value) || value[Symbol.iterator]) {
+    } else if (isIterable(value)) {
```
justinfagnani: `isIterable` does more checks than what it's replacing. By this point in the `if` chain we know the value is not `null` and not a string.

src/lib/parts.ts
```diff
@@ -412,18 +424,25 @@ try {
...
   };
+  // tslint:disable-next-line:no-any
   window.addEventListener('test', options as any, options);
+  // tslint:disable-next-line:no-any
   window.removeEventListener('test', options as any, options);
 } catch (_e) {
 }
 
+
+type EventHandler = ((this: EventTarget, event: Event) => unknown)|{
```
justinfagnani: isn't there a type in lib.dom.ts for this already?

src/directives/async-append.ts
rictic: Same.

src/lib/parts.ts
rictic: Fair point. Reverted.

src/lib/parts.ts
rictic: Not in a resin context.

src/lib/parts.ts
rictic: Fair, reverted.

src/lib/parts.ts
rictic: Good point. Turns out it's EventListenerOrEventListenerObject

src/lib/parts.ts
rictic: To ensure that it's a boolean. Reverted in any case.

PR #749, opened 2019-01-17
Mark fields as readonly where possible. Likewise use ReadonlyArray.

rictic:
Conceptually with the no-any branch, but split out for clarity, and because that branch had gotten big enough.

### DIFF
--- a/src/lib/template-factory.ts
+++ b/src/lib/template-factory.ts
@@ -89,8 +89,9 @@ export function templateFactory(result: TemplateResult) {
  * joining the strings array.
  */
 export type templateCache = {
-  stringsArray: WeakMap<TemplateStringsArray, Template>;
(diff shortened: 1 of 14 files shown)


src/lib/template-factory.ts
```diff
@@ -89,8 +89,9 @@ export function templateFactory(result: TemplateResult) {
  * joining the strings array.
  */
 export type templateCache = {
-  stringsArray: WeakMap<TemplateStringsArray, Template>;
-  keyString: Map<string, Template>;
+  readonly stringsArray: WeakMap<TemplateStringsArray, Template>; readonly keyString: Map<
```
justinfagnani: wtf. add a comment at the end of the first line?

rictic: Huh, yeah. Fixed.

PR #750, opened 2019-01-18
Enable a resin-like sanitization mechanism.

rictic:
This adds a global DOM sanitization function which is settable via the setSanitizeDOMValue call.

If set, this function is called whenever lit-html would write an attribute, property, or text to the dom, with all of the information necessary to determine if the write is safe.

This API is analogous to the Polymer function of the same name, and it pairs nicely with https://github.com/Polymer/polymer-resin

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -74,6 +74,16 @@ export type DOMSanitizer =
  */
 let sanitizeDOMValue: DOMSanitizer|undefined;
 
+/** Sets the global DOM sanitization callback. */
+export const __testOnlySetSanitizeDOMValueExperimentalMayChangeWithoutWarning = (newSanitizer: DOMSanitizer) => {
+  if (sanitizeDOMValue !== undefined) {
+    throw new Error(
+        `Attempted to overwrite existing lit-html security policy.` +
+        ` setSanitizeDOMValue should be called at most once.`);
+  }
+  sanitizeDOMValue = newSanitizer;
+};
+
 export const __testOnlyClearSanitizerDoNotCallOrElse = () => {
   sanitizeDOMValue = undefined;
 };
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -134,7 +134,7 @@ export class AttributeCommitter {
     // it together into a string without calling String() on the array.
     if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0] !== undefined) {
       const v = parts[0].value;
-      if (!Array.isArray(v)) {
+      if (!isIterable(v)) {
         return v;
       }
     }
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -74,16 +74,6 @@ export type DOMSanitizer =
  */
 let sanitizeDOMValue: DOMSanitizer|undefined;
 
-/** Sets the global DOM sanitization callback. */
-export const setSanitizeDOMValue = (newSanitizer: DOMSanitizer) => {
-  if (sanitizeDOMValue !== undefined) {
-    throw new Error(
-        `Attempted to overwrite existing lit-html security policy.` +
-        ` setSanitizeDOMValue should be called at most once.`);
-  }
-  sanitizeDOMValue = newSanitizer;
-};
-
 export const __testOnlyClearSanitizerDoNotCallOrElse = () => {
   sanitizeDOMValue = undefined;
 };
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -122,7 +122,7 @@ export class AttributeCommitter {
     // String(value)
     // The exception is if v is an array, in which case we do want to smash
     // it together into a string without calling String() on the array.
-    if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0]) {
+    if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0] !== undefined) {
       const v = parts[0].value;
       if (!Array.isArray(v)) {
         return v;
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -47,9 +47,7 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
  * attribute, or text. `node` is the node where the value is being inserted.
  */
 export type DomSanitizer =
-    (value: unknown,
-     name: string,
-     type: ('property'|'attribute'),
+    (value: unknown, name: string, type: ('property'|'attribute'|'text'),
      node: Node) => unknown;
 
 
@@ -108,15 +106,12 @@ export class AttributeCommitter {
     // String(value)
     if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0]) {
       const v = parts[0].value;
-      if (Array.isArray(v)) {
-        if (v.length === 1) {
-          return v[0];
-        }
+      if (Array.isArray(v) && v.length === 1) {
+        return v[0];
       } else {
         return v;
       }
     }
-
     let text = '';
 
     for (let i = 0; i < l; i++) {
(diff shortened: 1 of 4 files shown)


src/lib/parts.ts
```diff
@@ -58,12 +85,33 @@ export class AttributeCommitter {
...
+    // then we don't want to coerce the attribute value into one long
+    // string. Instead we want to just return the value itself directly,
+    // so that sanitizeDOMValue can get the actual value rather than
+    // String(value)
+    if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0]) {
+      const v = parts[0].value;
+      if (Array.isArray(v)) {
+        if (v.length === 1) {
+          return v[0];
+        }
```
justinfagnani: There's a fall-though here that I want to make sure is intentional. Can you comment it?

src/lib/parts.ts
```diff
@@ -58,12 +85,33 @@ export class AttributeCommitter {
...
+    // but not
+    //    attr="${foo} ${bar}" or attr="${foo} baz"
+    // then we don't want to coerce the attribute value into one long
+    // string. Instead we want to just return the value itself directly,
+    // so that sanitizeDOMValue can get the actual value rather than
+    // String(value)
+    if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0]) {
+      const v = parts[0].value;
+      if (Array.isArray(v)) {
+        if (v.length === 1) {
```
justinfagnani: Why special-case single-element arrays here? Add comment if you intend to keep, but I don't immediately see the rationale.

src/lib/parts.ts
```diff
@@ -58,12 +85,33 @@ export class AttributeCommitter {
...
+          return v[0];
+        }
+      } else {
+        return v;
+      }
+    }
+
     let text = '';
 
     for (let i = 0; i < l; i++) {
```
justinfagnani: Maybe add a TODO, and maybe we should loop in Mike, but I feel like we could ideally do some kind of taint-tracking here. If all the values are safe, then I think the result can be safe. Since we don't actually have safe-types to check yet, maybe we can accomplish this via the sanitizer API? Or... maybe that's too much for now.

justinfagnani: What I'm thinking is that the sanitizer gets both an array of strings and values and can do things like return them untransformed if all values are safe types, validate that the joined string wouldn't start with `'javascript:'`, etc.

rictic:
Update on this PR, I'm working through the last few issues with @mikesamuel and the rest of the security team before releasing, to be sure that we have a final API before this goes out.

justinfagnani:
@rictic can you pull in the latest change do disallow nested templates in <style> and <script>?

src/lib/parts.ts
rictic: A safe value doesn't necessarily mean from a trusted source, it could be a value from an untrusted source that has been validated as being safe for a given use case. It's not obvious to me that concatenating safe values produces a safe result.

I think instead what you'd do there is you'd concatenate your inputs first, and then call SafeUrl.sanitize() to get your SafeUrl, which would check that the combined input doesn't start with `javascript:` etc. Then you'd pass your SafeUrl to render. 

The different safe types have different options for construction, e.g. TrustedResourceUrl has a few ways to compose and create a trusted url, but it's even more constrained than SafeUrl because it has higher security requirements.

src/lib/parts.ts
justinfagnani: That behavior was there to match this case:

```html
html`<div>${[]1,2,3}</div>`
```

Where there's no "joiner" between the items. setAttribute would do `"1,2,3"` though, so I'd be open to changing this in a major version.

src/lib/parts.ts
justinfagnani: > it could be a value from an untrusted source that has been validated as being safe for a given use case

I thought the point of contextual auto-escaping is that you can't determine if something is safe outside of its context? In this case the values shouldn't be marked as safe independently of their use site. The sanitizer should be able to check that we're not constructing the attribute value `"javascript:doBadThing()"` from  `java${userValue}`, but it shouldn't do the same check on `java${safeValue}` right?

src/lib/parts.ts
```diff
@@ -249,11 +322,22 @@ export class NodePart implements Part {
         node.nodeType === 3 /* Node.TEXT_NODE */) {
       // If we only have a single text node between the markers, we can just
       // set its value, rather than replacing it.
-      // TODO(justinfagnani): Can we just check if this.value is primitive?
-      (node as Text).data = value as string;
+      if (sanitizeDOMValue) {
+        value = String(sanitizeDOMValue(value, 'data', 'property', node));
```
justinfagnani: why `'data'` here and `'textContent'` below?

daKmoR:
long time no see :grimacing: is this still being worked on? or should we come up with our own sanitize solution?

rictic:
It's still being worked on, just juggling a lot of concerns at the moment.

I pushed a new commit that no longer exposes the setSanitizeDOMValue so we can reduce the delta with the downstream version inside google while we iterate on the API.

I'd also be OK with exposing a function like setSanitizeDOMValueExperimentalMayChange if it's blocking you, with the understanding that we might make breaking changes without a major breaking release to its API.

src/lib/parts.ts
rictic: Switched to data in both cases

src/lib/parts.ts
rictic: It should absolutely do the same check on `java${safeValue}` because the safeValue hasn't been constructed to be safe in any possible configuration. A safe value isn't one that's from a trustworthy source, it's one that was safe at the time it was constructed.

As another example, consider [TrustedResourceUrl](https://google.github.io/closure-library/api/goog.html.TrustedResourceUrl.html), which is a type for a URL that may execute code. Its constructors ensure that it is only constructed in limited ways (e.g. the domain must come from a constant string in source code, with the option of structured parameterization of the path, URL params, etc)
j
As a result, this is safe: `<script src=${trustredResourceUrl}>` but `<script src="http://evil.example.com/${trustedResourceUrl}>` is not safe, and must be blocked.

src/lib/parts.ts
```diff
@@ -68,12 +119,30 @@ export class AttributeCommitter {
...
+    //    attr="${foo} ${bar}" or attr="${foo} baz"
+    // then we don't want to coerce the attribute value into one long
+    // string. Instead we want to just return the value itself directly,
+    // so that sanitizeDOMValue can get the actual value rather than
+    // String(value)
+    // The exception is if v is an array, in which case we do want to smash
+    // it together into a string without calling String() on the array.
+    if (l === 1 && strings[0] === '' && strings[1] === '' && parts[0]) {
+      const v = parts[0].value;
+      if (!Array.isArray(v)) {
```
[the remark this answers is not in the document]
rictic: Switched the check to check for iterables, not just arrays

src/lib/parts.ts
```diff
@@ -93,7 +164,11 @@ export class AttributeCommitter {
   commit(): void {
     if (this.dirty) {
       this.dirty = false;
-      this.element.setAttribute(this.name, this._getValue() as string);
+      let value = this._getValue();
+      if (sanitizeDOMValue) {
+        value = sanitizeDOMValue(value, this.name, 'attribute', this.element);
+      }
+      this.element.setAttribute(this.name, String(value));
```
justinfagnani: Remove the `String()` wrapper. `setAttribute()` already does toString coercion (the dom.d.ts typing is wrong).

src/lib/parts.ts
```diff
@@ -245,18 +320,26 @@ export class NodePart implements Part {
...
     } else {
-      this.__commitNode(document.createTextNode(valueAsString));
+      // When setting text content, for security purposes it matters a lot what
+      // the parent is. For example, <style> and <script> need to be handled
+      // with care, while <span> does not. So first we need to put a text node
+      // into the document, then we can sanitize its contentx.
+      const textNode = document.createTextNode('');
+      this.__commitNode(textNode);
+      if (sanitizeDOMValue) {
+        value = String(
```
justinfagnani: Remove the String() wrapper

justinfagnani: Note that by mutating the `value` variable here, we're going to save the sanitized value as the previous value: 

https://github.com/Polymer/lit-html/pull/750/files#diff-63a9df26ecde05415a902b5b5dc5e266R344

This means when a sanitizer changes a value that we'll always re-render this part. Can we say that sanitizers have to be pure functions so we can remember the pre-sanitized value and skip the whole operation if the input value is the same?

src/lib/parts.ts
rictic: Needed for Symbol unfortunately.

```typescript
'' + Symbol(); // throws error
'' + String(Symbol())
```

src/lib/parts.ts
rictic: Done

rictic: Good point, that's a reasonable assumption. Separated out the rendered value from `this.value`

rictic:
Internal presubmit is green, this is about to land :)

PR #753, opened 2019-01-19
[docs] Add a search box

keanulee:
Staged at https://20190118t161039-dot-polymer-lit-html.appspot.com. Should be pretty simple to apply the same changes to the other product sites.


arthurevans:
This is super cool. Tested, works great. 

I have one request: can we add analytics? It's really helpful to us to see what people are looking for.

We've been tracking this with events on the polymer-library site, but I've just realized that we can integrate with GA's built in site search tracking if we record it as a fake pageview, like this:

```js
  window.recordSearch = function(query) {
    gtag('config', '{{site.google_analytics}}', {
      'page_path': '/search?q=' + query
    });
  }
```

If we later add in-page search results, we can remove this code entirely, since the search tracking is based entirely on the pageview URL.

Also, qq on the query handling: should we be wrapping the query with `encodeURIComponent` like we did on the polymer-library site:

https://github.com/Polymer/polymer-library-docs/blob/master/app/elements/pw-shell.js#L285

The current code is just copypasta from a previous version, so it's unclear to me whether it's needed or not. It seems like there could be issues with search terms, but I have tried some with spaces, etc., and I haven't broken anything yet...

(Of course, I also only tried on Chrome.)

arthurevans:
Thanks Keanu. I've merged Kate's API doc theme change. Can you pull the change you made for lit-element into this PR?

 https://github.com/Polymer/lit-element/pull/464/commits/6e51e69477fa5707c3023cea0b7816b8285ca819

keanulee:
Done, PTAL @arthurevans

keanulee:
Ok, I added `encodeURIComponent` and analytics. To make sure I can used the `ga()` function in Jekyll and TypeDoc output, I switched the Jekyll analytics template to [analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/) instead of [gtag.js](https://developers.google.com/analytics/devguides/collection/gtagjs/). This should be fine since gtag.js just loads analytics.js behind the scenes and I don't think we use the other parts of gtag.js:

> The global site tag (gtag.js) is a JavaScript tagging framework and API that allows you to send event data to Google Analytics, Google Ads, and Google Marketing Platform

Issue #710, opened 2018-12-19
Add docs on rendering templates


arthurevans:

fernandopasik:
I think this was supposed to be closed by #754

arthurevans:
Thanks @fernandopasik! You're right, this should have been closed.

PR #754, opened 2019-01-19
Added Styling & Rendering sections.

### LINKED ISSUE #711: Add docs on styling concerns

arthurevans:
Fixes #711 and #710.

Staged at https://styling-updates-dot-polymer-lit-html.appspot.com


arthurevans:
Sorry for the very late arrival. Something to read on the plane.

Tomorrow: Tooling!

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
+---
+layout: post
+title: Styling templates
+slug: styling-templates
+---
+
+lit-html just renders HTML; you can style HTML created with lit-html like any other HTML, using CSS. How you apply styles to lit-html templates depends on how you're using it—for example, if you're using lit-html inside a component system like LitElement, you can follow the patterns used by that component system.
```
justinfagnani:
```suggestion
lit-html focuses on one thing: rendering HTML. How you apply styles to the HTML lit-html creates depends on how you're using it—for example, if you're using lit-html inside a component system like LitElement, you can follow the patterns used by that component system.
```

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+
+In most cases you should **not** use lit-html bindings inside a style sheet. It's more efficient to use lit-html bindings to manipulate the `class` and `style` attributes.
```
justinfagnani: Seems like a little jump from the intro para to here, and there's a "Don't use bindings in stylesheets" section below. Does this belong there?

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+In most cases you should **not** use lit-html bindings inside a style sheet. It's more efficient to use lit-html bindings to manipulate the `class` and `style` attributes.  
+
+lit-html provides two directives for manipulating an element's `class` and `style` attributes:
+
+*   [`classMap`](template-reference#classmap) sets classes on an element based on the properties of an object.
+*   [`styleMap`](template-reference#stylemap) sets the styles on an element based on a map of style properties and values.
+
+See [classMap](template-reference#classmap) and [styleMap](template-reference#stylemap) in the Template syntax reference.
+
+There are also some special considerations when using lit-html to render into a shadow root.
```
justinfagnani: I think rather than calling Shadow DOM as special, maybe divide the world into two: you're either going to be using Shadow DOM or not.

If not, you're rendering to the main doc and you should use whatever system you already do. A simple stylesheet or various CSS module / CSS-in-JS systems. If you use those you might need to bind class names (it's common for a CSS module system to have you import class names to put on elements), and that's easy in lit-html. For more information see your CSS libraries docs, etc.

If you are using Shadow DOM (recommended), then... (what you have below)

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+  <style>
+    :host { ... } 
+    .test { ... }
+  </style> 
+  <div class="test">...</div> 
+`;
+```
+
+
+### Don't use bindings in stylesheets
```
justinfagnani: I would consider renaming this "Binding in Stylesheets"

Then recommending that in most cases you don't. In native Shadow DOM it's not that bad if the stylesheet is very small, for instance:

```js
html`
<style>
:host {
  --my-special-color: ${specialColor}
}
</style>
`
```

Although, if you're templated to do this, `styleMap` is probably better, so you could show that there.

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+layout: post
+title: Styling templates
+slug: styling-templates
+---
+
+lit-html just renders HTML; you can style HTML created with lit-html like any other HTML, using CSS. How you apply styles to lit-html templates depends on how you're using it—for example, if you're using lit-html inside a component system like LitElement, you can follow the patterns used by that component system.
+
+In most cases you should **not** use lit-html bindings inside a style sheet. It's more efficient to use lit-html bindings to manipulate the `class` and `style` attributes.  
+
+lit-html provides two directives for manipulating an element's `class` and `style` attributes:
```
justinfagnani: I think styleMap and classMap deserve a heading here. Maybe: "Styling individual Elements with classMap and styleMap". An example could be useful too.

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+*   [`classMap`](template-reference#classmap) sets classes on an element based on the properties of an object.
+*   [`styleMap`](template-reference#stylemap) sets the styles on an element based on a map of style properties and values.
+
+See [classMap](template-reference#classmap) and [styleMap](template-reference#stylemap) in the Template syntax reference.
+
+There are also some special considerations when using lit-html to render into a shadow root.
+
+## Rendering in shadow DOM
+
+When rendering into a shadow root, you usually want to add a stylesheet inside the shadow root to the template, to you can style the contents of the shadow root.
```
justinfagnani: I'd call out that inline styles are perfectly OK in ShadowRoots because browsers optimize this case, then maybe mention that using `adoptedStylesheets` can be even better, and link to LitElement docs (do we have docs on that yet?)

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+*   Use bindings in the `class` and `style` attributes to control the styling of child elements.
+
+	
+
+
+### Polyfilled shadow DOM: ShadyDOM and ShadyCSS
+
+If you're using shadow DOM, you'll probably need to use polyfills to support older browsers that don't implement shadow DOM natively. [ShadyDOM](https://github.com/webcomponents/shadydom) and [ShadyCSS](https://github.com/webcomponents/shadycss) are polyfills, or shims, that emulate shadow DOM isolation and style scoping. 
+
+The lit-html `shady-render` module provides necessary integration with the shady CSS shim. If you're writing your own custom element base class that uses lit-html and shadow DOM, you'll need to use `shady-render` and also take some steps on your own.
```
justinfagnani:
```suggestion
The lit-html `shady-render` module provides necessary integration with the ShadyCSS shim. If you're writing your own Custom Element base class that uses lit-html and Shadow DOM, you'll need to use `shady-render` and also take some steps on your own to enable it.
```

docs/_guide/03-styling-templates.md
```diff
@@ -0,0 +1,94 @@
...
+
+The [ShadyCSS README](https://github.com/webcomponents/shadycss#usage) provides some directions for using shady CSS. When using it with `lit-html`:
+
+*   Import `render` and `TemplateResult` from the `shady-render` library.
+*   You **don't** need to call `ShadyCSS.prepareTemplate`.  Instead pass the scope name as a render option. For custom elements, use the element name as a scope name. For example:
+
+    ```js
+    import { render, TemplateResult } from 'lit-html/lib/shady-render';
+
+    render(this.myTemplate(), this.shadowRoot, { scopeName: this.tagName.toLowerCase() });
```
justinfagnani: The `this` is kind of floating here, maybe put a class around the example and put the call in a dummy example method?

docs/_guide/04-rendering-templates.md
```diff
@@ -0,0 +1,30 @@
...
+
+// subsequent renders will update the DOM
+render(sayHi('Zoe'), document.body);
+```
+
+## Render Options
+
+The `render` method also takes an `options` argument that allows you to specify:
+
+*   `eventContext`: The `this` value to use when invoking event listeners registered with the `@eventName` syntax.
```
justinfagnani: eventContext only effects listener functions. Listener objects with a `handleEvent` method are always called with the listener object as the `this` value.

docs/_guide/04-rendering-templates.md
```diff
@@ -0,0 +1,30 @@
...
+
+*   `templateFactory`: The `TemplateFactory` to use. A `TemplateFactory` creates a template element from a `TemplateResult`, typically caching templates based on their static content. Users won't usually supply their own `TemplateFactory`, but libraries that use lit-html may implement custom template factories to customize template handling.
```
justinfagnani: maybe add "this is an advanced option"

docs/_guide/03-styling-templates.md
arthurevans: We don't capitalize shadow DOM or custom elements in docs.

justinfagnani: We do here: https://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-201/
And MDN does in their text (but not in the title?): https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

docs/_guide/03-styling-templates.md
arthurevans: I was trying to avoid writing an entire custom element here, because I know there are other wrinkles we're missing... I dunno... Maybe if we make it clear there are lots of things missing from the sample?

docs/_guide/03-styling-templates.md
arthurevans: Yeah, I hated to add a section here because it felt really repetitive with the section in the reference. 

That example is already incredibly stupid, and I don't really want to repeat it ... Would love suggestions for better examples here.

katejeffreys: https://github.com/katejeffreys/classmaps/blob/master/index.js has a couple i was playing with

docs/_guide/03-styling-templates.md
arthurevans: OK, here you're using "inline style" to refer to a style sheet in a shadow root, right?

docs/_guide/03-styling-templates.md
arthurevans: OK, added just the skeleton of a class here.

docs/_guide/04-rendering-templates.md
arthurevans: Fixed, thanks.

docs/_guide/03-styling-templates.md
arthurevans: OK, fixed.

arthurevans:
Addressed feedback and updated staged version at: https://styling-updates-dot-polymer-lit-html.appspot.com

Think I'm ready to merge this, but going to leave it open for another hour or two in case anyone wants to review staged version before I merge.

PR #780, opened 2019-02-01
Update some Part and directive-related API docs.

justinfagnani:


src/lib/parts.ts
```diff
@@ -128,6 +132,14 @@ export class AttributePart implements Part {
   }
 }
 
+/**
+ * A Part that controls a location within a Node tree. Like a Range, NodePart
+ * has start and end locations and can set and update the Nodes between those
+ * locations.
+ *
+ * NodeParts support several value types: primitives, Nodes, TemplateResults,
```
arthurevans: Don't they also support other NodeParts? I thought that was how repeat was implemented.

src/lib/directive.ts
```diff
@@ -26,26 +26,45 @@ export type DirectiveFactory = (...args: any[]) => object;
...
- * during template rendering, rather than passing as a value.
+ * Brands a function as a directive factory function so that lit-html will call
+ * the function during template rendering, rather than passing as a value.
+ *
+ * A _directive_ is a function that takes a Part as an argument. It has the
+ * signature: `(part: Part) => void`.
+ *
+ * A directive _factory_ is a function that takes arguments for data and
+ * configuration and returns a directive. Users of directive usually refer to
+ * the directive factory as the directive. For example, "The repeat directive".
```
arthurevans: Not a comment for the present, but this paragraph (and this naming scheme) makes for really confusing reading. We should discuss whether there's some way to make this less confusing. Maybe there isn't, dunno.

src/lib/parts.ts
arthurevans:
```suggestion
 * NodeParts support several value types: primitives, Nodes, TemplateResults, and other NodeParts,
```

src/lib/parts.ts
```diff
@@ -128,6 +132,14 @@ export class AttributePart implements Part {
...
+ * and arrays and iterables of those.
```
arthurevans:
```suggestion
 * as well as arrays and iterables of those types.
```

src/lib/parts.ts
justinfagnani: A nested NodePart isn't a value set into another NodePart with setValue, it's actually just placed into the correct DOM position.

justinfagnani: this isn't quite correct, see other comment

Issue #756, opened 2019-01-22
<!-- ${crash} -->


gertcuykens:
Please add some more unit test for `<!-- ${crash} -->`
```js
const crash = "test"
const template = () => html`<!-- ${crash} -->`
```

justinfagnani:
I added another test for this and there's no crash. Please re-open with a repo if it's still an issue for you.

gertcuykens:
Hold it :P https://jsbin.com/wilalus/edit?html,output

True it doesn't crash anymore but it's not working like it supposed to be working right? See `--> error`

justinfagnani:
It might be very difficult to do better than we are now. First priority is to not crash, second is to not already subsequent bindings. After that, we'll do our best, but it's lower priority.

I'll see if I can adjust the literal scan in TemplateResult to not insert comments in this case.

gertcuykens:
Ok thanks, just make sure it's documented somewhere.

PR #781, opened 2019-02-01
Add another test for comments with bindings

justinfagnani:
Should cover #756


gertcuykens:
Hold it :P https://jsbin.com/wilalus/edit?html,output

True it doesn't crash anymore but it's not working like it supposed to be working right? See example `--> error`

Issue #784, opened 2019-02-03
Add documentation for nothing


Christian24:
It would be pretty cool if there was documentation for `nothing` sentinel value. Maybe it could go into https://github.com/Polymer/lit-html/blob/master/docs/_guide/05-template-reference.md or https://github.com/Polymer/lit-html/blob/master/docs/_guide/02-writing-templates.md with an example.

KeithHenry:
How is `nothing` better than `null` or `undefined`?

Where should we `import` `nothing` from?

The proposed doc changes don't say.

ruphin:
`nothing` has a very specific purpose. If you have some custom element `<my-element>` and that custom element has a slot with fallback content, the fallback content will not appear if you render any non-comment nodes as children of that element. (this includes empty TextNodes)

```js
// This will usually render an empty TextNode inside the element if content is falsy,
// so the slot fallback content will not appear
html`<my-element>${ content ? content : "" }</my-element>`
html`<my-element>${ content ? content : null }</my-element>`
html`<my-element>${ content ? content : undefined }</my-element>`

// This will remove all non-comment child nodes when content is falsy,
// so the slot fallback content will appear
html`<my-element>${ content ? content : nothing }</my-element>`
```

Other than this, it has no real purpose and you may as well use `""` (emptystring) or `undefined` or `null`.

`nothing` is a direct export of `lit-html.js`:

```js
import { nothing } from "https://unpkg.com/lit-html/lit-html.js?module";
```

Christian24:
@KeithHenry Thanks for the feedback. I did leave out the import on purpose, because some examples seem to have them while others don't. I added the import now. I also made some efforts to explain what @ruphin has explained here, but probably not as well. 

@ruphin Thanks for the context. I do wonder though if it is a good best practice to use it though? I mean `nothing` could still be seen as a uniform way of rendering '', null or undefined.

ruphin:
It really depends on a bunch of factors, like the context where it is being used. If it is being used in a switch style statement and the other part of the statement is a primitive like a String or a Number, then technically using `""` will have slightly better performance than `undefined` or `null`, which in turn have equal or slightly better performance than `nothing`. If the other part of the statement is a template or Node or some other complex object, then `nothing` may have slightly better performance than the others. Aditionally, `""`, `undefined`, and `null` may cause unexpected behaviours, but only in the above mentioned scenario.

[...]

PR #785, opened 2019-02-03
Document how to render no content

Christian24:
This adds a draft of documentation for `nothing` in response to #784. Is there any other benefit to using `nothing` over '' other than it being more descriptive?


justinfagnani:
meta feedback: "Draft of fixing #784" isn't a very useful title. It's hard to know what this is about without clicking around. "Document the nothing sentinel" would be much better.

Christian24:
@justinfagnani I agree. Sorry, might have been to quick naming this one :)

Christian24:
Based on the discussion in #784. I added more information on how to render no content with empty strings, `null`, `undefined` and `nothing`. 

It's certainly not the best, but I do hope it is at least a first draft.

docs/_guide/05-template-reference.md
```diff
@@ -532,4 +532,75 @@ const content = fetch('./content.txt').then(r => r.text());
...
+handle the two equally:
+
+````js
+document.getElementById('container').innerHTML = null;`
+````
+will produce an empty div while
+````js
+document.getElementById('container').innerHTML = undefined;
+````
+produces a div with the text undefined in it. Also an attribute binding
```
KeithHenry: I think a better example (using lit) here might be that `<input value=${thing ? thing.prop : undefined}>` renders `'undefined'` as the value.

In this case `''` renders `value=""` and you can use `ifDefined` to not include the `value` attribute at all.

Christian24: I did add a new example with `undefined` as attribute value, but kept the old one too, as null does render no content when set through innerHTML and when setting the attribute to null it sets the text to null which is a difference in behavior the user might not expect. 

Also added a comment about `ifDefined` I have a feeling this might be more conditional rendering already.

arthurevans:
I think the format of this content is better suited to the guide (writing templates) than to the reference. WDYT @justinfagnani ?

Christian24:
@arthurevans I do agree. It did grow significantly, I will move it.

Christian24:
I did move them to the end of the composing templates section.

justinfagnani:
Thanks for the PR @Christian24 

I think this new test is fairly long for the concept it's discussion. I worry that it makes the idea of rendering nothing seem more complicated than it is. I wonder how we could trim this down?

arthurevans:
Sorry, meant to add this to my review queue.

Christian24:
> Thanks for the PR @Christian24
> 
> I think this new test is fairly long for the concept it's discussion. I worry that it makes the idea of rendering nothing seem more complicated than it is. I wonder how we could trim this down?

I do feel your pain. :) I did make it this lengthy, because of the discussion in #784. I think there's a desire for this kind of documentation, but I do understand and agree that this sounds a lot more complicated than it should be. I have a couple ideas to address this:

Slim the rendering no content section down to just`''` and `nothing`. Maybe state that `''` is considered a best practice? 

Add a new Further reading page, where the section with null and undefined goes and just link that from the rendering nothing section? What do you think, @justinfagnani?

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
 When lit-html re-renders a template, it only updates the modified portions: it doesn't create or remove any more DOM than it needs to. But when you switch from one template to another, lit-html needs to remove the old DOM and render a new DOM tree. 
 
 The `cache` directive caches the generated DOM for a given binding and input template. In the example above, it would cache the DOM for both the  `summaryView` and `detailView` templates. When you switch from one view to another, lit-html just needs to swap in the cached version of the new view, and update it with the latest data.
+### Rendering nothing
+Very often you may conditionally render something, but in the other case
```
arthurevans: This intro sort of says the same thing twice. Suggest:

Sometimes, you may want to render nothing at all. For example, you might want to render nothing when a given condition is false.

The values `undefined`, `null` and the empty string (`''`) in a text binding all render an empty text node. In most cases, that's exactly what you want.

In some cases you may need to ensure that a binding does not produce a text node at all. This is true when dealing with shadow DOM slots that have fallback content.

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+you do not want to render anything at all. So, if you are using ternary 
+operators in one case you might return your template and in the other you
+want to return no content lit-html leaves you with a couple options:
+* an empty string (`''`)
+* null
+* undefined
+* nothing sentinel
+#### An empty string
```
arthurevans: Don't really need these sections. Could move the example up to the intro section if you like.

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+
+````js
+${user.isAdmin
+      ? html`<button>DELETE</button>`
+      : ''
+  }
+````
+This does the trick and works great and efficiently in most cases. It does
+not work with slot fallback content.
+#### null or undefined
```
arthurevans: I would cut this section entirely. It's not talking about lit, it's talking about DOM APIs. `innerHTML = undefined` might render `undefined`, but ```html`${undefined}` ``` does not.

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+  ${user.isAdmin
+      ? html`<button>DELETE</button>`
+      : nothing
+  }
+  </shadow-element>
+`;
+``` 
+If the user is logged in, the DELETE-button will be rendered. If the user is not logged in the part will clear,
+therefore the slot is empty and its fallback content "Sorry, no content available. I am just fallback content" will be rendered.
+#### Performance
```
arthurevans: If we don't have anything definitive to say about performance, I don't think we need a section on it.

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+`nothing` is a sentinel value provided by lit-html. It really does render nothing, because
+it clears the Part.
+
+Why is this important? Imagine you have a shadow DOM enabled custom element, `shadow-element`, that uses a slot. 
+The template looks like this:
+```js
+html`<slot>Sorry, no content available. I am just fallback content</slot>`;
+``` 
+The slot defines fallback content for when there is no content defined to be put in the slot. 
+So, extending on our first example:
```
arthurevans: our first => the previous

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+```js
+import {nothing} from 'lit-html';
+
+
+html`<shadow-element>
```
arthurevans: So this example actually _won't_ work as described, because there's whitespace inside the `<shadow-element>`. Those whitespaces turn into text nodes.

So it would have to be:

```
const myTemplate = () =>  html`
      <shadow-element>${user.isAdmin
        ? html`<button>DELETE</button>`
        : nothing
      }</shadow-element>
`;
```

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+
+Why is this important? Imagine you have a shadow DOM enabled custom element, `shadow-element`, that uses a slot. 
+The template looks like this:
+```js
+html`<slot>Sorry, no content available. I am just fallback content</slot>`;
+``` 
+The slot defines fallback content for when there is no content defined to be put in the slot. 
+So, extending on our first example:
+```js
+import {nothing} from 'lit-html';
```
arthurevans: `import {html, nothing} from 'lit-html';`

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+
+html`<shadow-element>
+  ${user.isAdmin
+      ? html`<button>DELETE</button>`
+      : nothing
+  }
+  </shadow-element>
+`;
+``` 
+If the user is logged in, the DELETE-button will be rendered. If the user is not logged in the part will clear,
```
arthurevans: DELETE button is rendered

docs/_guide/02-writing-templates.md
arthurevans: ... and then we can skip straight to the section on slots

docs/_guide/02-writing-templates.md
arthurevans: the part will clear => "nothing is rendered inside the `<shadow-button>`, "

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
...
+therefore the slot is empty and its fallback content "Sorry, no content available. I am just fallback content" will be rendered.
```
arthurevans: will be rendered => is rendered

arthurevans: Maybe add:

```
**Whitespace creates text nodes.** For the example to work, the text binding inside `<shadow-element>` must be the **entire** contents of `<shadow-element>`. Any whitespace outside of the binding delimiters adds static text nodes to the template, suppressing the fallback content. (Whitespace _inside_ the binding delimiters is fine.)
{.alert .alert-info}
```

docs/_guide/02-writing-templates.md
```diff
@@ -286,4 +286,83 @@ html`${cache(data.showDetails
 When lit-html re-renders a template, it only updates the modified portions: it doesn't create or remove any more DOM than it needs to. But when you switch from one template to another, lit-html needs to remove the old DOM and render a new DOM tree. 
 
 The `cache` directive caches the generated DOM for a given binding and input template. In the example above, it would cache the DOM for both the  `summaryView` and `detailView` templates. When you switch from one view to another, lit-html just needs to swap in the cached version of the new view, and update it with the latest data.
+### Rendering nothing
```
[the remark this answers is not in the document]
Christian24: The discussion on attribute bindings came from review feedback. However, I agree it is not directly related to actual content, so I think it is okay to remove it. 

I did move the entire section above caching template results now. That was my bad.

docs/_guide/02-writing-templates.md
Christian24: I did stick to your wording here.

docs/_guide/02-writing-templates.md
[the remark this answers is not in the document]
Christian24: I did move it above caching template results now

docs/_guide/02-writing-templates.md
Christian24: Used your text suggestion.

docs/_guide/02-writing-templates.md
Christian24: see my comments below.

docs/_guide/02-writing-templates.md
Christian24: This is again also based on feedback. While it is talking about the way DOM handles this, I think most people might be unaware of this. So, maybe in the intro we should promote the use of empty strings instead of `null` or `undefined`.

docs/_guide/02-writing-templates.md
Christian24: Removed.

docs/_guide/02-writing-templates.md
Christian24: Thanks. Good catch.

docs/_guide/02-writing-templates.md
Christian24: `shadow-element`. Otherwise incorporated your feedback.

docs/_guide/02-writing-templates.md
Christian24: I did add a couple other imports along the way. Is there a general rule when imports are part of the examples and when not?

docs/_guide/02-writing-templates.md
Christian24: What does the `{.alert .alert-info}` do? Otherwise I added the info at the bottom.

justinfagnani:
Thanks for the updates! Seems fine on my end.

docs/_guide/02-writing-templates.md
Christian24: I added a small disclaimer saying there are inconsistencies. What do you think @arthurevans?

arthurevans:
Thanks @Christian24 can you resolve conflicts so I can go ahead and merge?

I might tweak some of the wording, but at this point I'd like to just get this merged & iterate. Thanks!

Christian24:
@arthurevans I did resolve the conflicts. Sorry, I did not notice the conflicts before.

arthurevans:
Thanks!

PR #800, opened 2019-02-09
It's released, already!

arthurevans:


justinfagnani:
I think removing the section entirely as in #806 makes sense to me

PR #806, opened 2019-02-12
Remove status disclaimer now 1.0.0 has shipped

benmccallum:
I removed the section entirely but if you want to reword instead given latest status feel free.


benmccallum:
I signed it!

web-padawan:
Same as #800 but the removal makes more sense to me 👍

benmccallum:
Whoops, sorry should've had a quick look in the active PR list. My bad.

PR #813, opened 2019-02-16
Revamp template polyfill

jridgewell:
Does a few things:

- Minimal poyfills no longer collide
  - Since lit doesn't create a HTMLTemplateElement class, checking for its existence doesn't prevent a second polyfilling (imagine two lit-html versions being included on the page)
- Makes `content` non-writable
- Makes `innerHTML` enumerable
  - Element class methods/accessors are, strangely, all enumerable
- Uses `removeNodes` and `reparentNodes` helpers
- Reuses the original template element when upgrading it, instead of creating a div

### DIFF
--- a/src/polyfills/template_polyfill.ts
+++ b/src/polyfills/template_polyfill.ts
@@ -24,6 +24,10 @@ import {removeNodes, reparentNodes} from '../lib/dom.js';
  * polyfill: https://github.com/webcomponents/template
  */
 export const initTemplatePolyfill = (forced = false) => {
+  // Minimal polyfills (like this one) may provide only a subset of Template's
+  // functionality. So, we explicitly check that at least content is present to
+  // prevent installing patching with multiple polyfills, which might happen if
+  // multiple versions of lit-html were included on a page.
   if (!forced && 'content' in document.createElement('template')) {
     return;
   }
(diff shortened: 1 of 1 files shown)


src/polyfills/template_polyfill.ts
```diff
@@ -22,37 +24,42 @@
  * polyfill: https://github.com/webcomponents/template
  */
 export const initTemplatePolyfill = (forced = false) => {
-  if (typeof HTMLTemplateElement !== 'undefined' && !forced) {
+  if (!forced && 'content' in document.createElement('template')) {
```
justinfagnani: Can you document `'content' in document.createElement('template')`, specifically mentioning the two versions of lit-html issue.

jridgewell: Done.

justinfagnani:
@zeroliu can you check that this still works for you?

@jridgewell one thing about creating a `template` element instead of a `div` is that it make it hard to test these code paths on browsers with native template. See https://github.com/Polymer/lit-html/pull/466/files/9e10a278cdae3454948d04ca239367c8c314684a#r216080403

jridgewell:
> one thing about creating a `template` element instead of a `div` is that it make it hard to test these code paths on browsers with native template.

That's specifically about non-writability of the `template.content` property, so the old code's `template.content = fragment` wouldn't work. But since `content` is an accessor on the prototype, we're able to `Object.defineProperty` to override without any issue. We could further verify our polyfill is running by asserting that `template.innerHtml === undefined`, if you want.

justinfagnani:
Good points. Thanks!

PR #815, opened 2019-02-16
classMap syntax punctuation

Westbrook:
Adds some missing punctuation of the `classMap` documentation.


keanulee:
Thanks @Westbrook

PR #822, opened 2019-02-21
Update spelling mistake in import

kevinsimper:
I copy and pasted it and didn't notice the spelling mistake at first, made me upgrade to the latest lit-element from @polymer/lit-element 👍 😄


keanulee:
Thanks @kevinsimper

PR #824, opened 2019-02-24
Use duck-typing to test if container is ShadowRoot

jridgewell:
The easiest difference to detect between a `ShadowRoot` and a `DocumentFragment` is the `host` element. Conveniently, this works in both v0 and v1.

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -15,9 +15,3 @@ interface Window {
   ShadyCSS?: ShadyCSS;
   ShadyDOM?: ShadyDOM;
 }
-
-/** Allows code to check `instanceof ShadowRoot`. */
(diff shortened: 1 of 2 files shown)


justinfagnani:
Is `container.nodeType === 11 && !!(container as ShadowRoot).host` faster than `container instanceof ShadowRoot`?

It doesn't seem shorter or easier.

jridgewell:
[Considerably faster](https://jsbench.github.io/#1914f82b6eb69ec8bdd05e0537ac5610), `instanceof` is the slowest operator in JS.

But my actual reason for opening it is because `instanceof` suffers from cross realm issues. I try to avoid it whenever possible.

justinfagnani:
Thanks for the benchmarks!

PR #825, opened 2019-02-24
Reuse a single TreeWalker in nested templates

jridgewell:
Creating `TreeWalker`s isn't very optimized, so reusing a single one in the nested templates can speed it up a bit.

https://jsbench.github.io/#d0363ded1753dec3e060e61e1794896b

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -68,14 +68,14 @@ export class TemplateInstance {
     const parts = this.template.parts;
     let partIndex = 0;
     let nodeIndex = 0;
+    // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
+    const walker = document.createTreeWalker(
+        document,
+        133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */,
+        null,
+        false);
     const _prepareInstance = (fragment: DocumentFragment) => {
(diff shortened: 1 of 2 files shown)


justinfagnani:
I don't care that much about optimizing nested templates, since that should be a pretty rare case and we added support mainly for correctness so lit-html templates didn't blow up with nested templates... 

_But_, if reusing a TreeWalker is that much faster, what if we reused one across the all templates? That could have an impact on real-world code (though I do think we're at the point where DOM mutation operations are dominating costs already).

The `Template` constructor isn't re-entrant. `TemplateResult#_clone` shouldn't be either, but I'd want some tests of scenarios like templates with custom elements that synchronously create ShadowRoots under the polyfill, etc.

jridgewell:
> _But_, if reusing a TreeWalker is that much faster, what if we reused one across the all templates? That could have an impact on real-world code

That was the end-goal. 😉

> The `Template` constructor isn't re-entrant. `TemplateResult#_clone` shouldn't be either, but I'd want some tests of scenarios like templates with custom elements that synchronously create ShadowRoots under the polyfill, etc.

Unfortunately, we're not there yet. https://github.com/Polymer/lit-html/pull/831 has to come first, then we need to fix the reentrancy behavior in `insertNodeIntoTemplate`, which can fire upgrade/`connectedCallback` from the `insertBefore`.

PR #826, opened 2019-02-24
Cache length of non-array iterables

jridgewell:
We can marginally speed up for-loops over non-arrays (in this case, c++
backed `NamedNodeMap` and `NodeList`).

NamedNodeMap: https://jsbench.github.io/#e730df94c251a459f9213abdafe1c34c
NodeList: https://jsbench.github.io/#f638cacc866a1b2d6e517e6cfa900d6b

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -129,8 +129,9 @@ const prepareTemplateStyles =
       shadyRenderSet.add(scopeName);
       // Move styles out of rendered DOM and store.
       const styles = renderedDOM.querySelectorAll('style');
+      const {length} = styles;
       // If there are no styles, skip unnecessary work
-      if (styles.length === 0) {
+      if (length === 0) {
         // Ensure prepareTemplateStyles is called to support adding
         // styles via `prepareAdoptedCssText` since that requires that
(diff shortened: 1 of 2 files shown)


stramel:
Wow, on my mobile browser I am averaging ~4.5-5k ops/sec on`NodeList` and with `NamedNodeMap`

~5.5mil ops/sec with no attrs
~400k ops/sec with some attrs
~35k ops/sec with lots of attrs

bigopon:
The cached version doesn't always perform best on my run. Is it expected?
image

jridgewell:
The `NodeList` perf tests are to show that cached-length iteration is always faster than uncached-length iteration (both live and static `NodeList`s). There's a more advanced (and downright uglier) pattern we can use to hit a fast path in Chrome, but it's only faster in Chrome.

bigopon:
I see now. I mapped the tests.

justinfagnani:
It does seem like in the NamedNodeMap tests the cached version isn't always faster. Is the variance that large?

jridgewell:
It's consistently 5-15% faster for me?

justinfagnani:
Oops, I read 5 and 6 backwards...

PR #827, opened 2019-02-24
Optimize TemplateInstance cloning

jridgewell:
I'm assuming it's very common to have a large number of nodes in between the template's parts. If so, creating a small inner loop to advance the tree walker should make everything quicker.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -20,7 +20,7 @@ import {isCEPolyfill} from './dom.js';
 import {Part} from './part.js';
 import {RenderOptions} from './render-options.js';
 import {TemplateProcessor} from './template-processor.js';
-import {isTemplatePartActive, Template} from './template.js';
+import {isTemplatePartActive, Template, TemplatePart} from './template.js';
 
 /**
  * An instance of a `Template` that can be attached to the DOM and updated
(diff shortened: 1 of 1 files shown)


justinfagnani:
@aomarks can you take a look at this one in the benchmark harness?

jridgewell:
Seems the same on Chrome, and an improvement on Safari: https://jsbench.github.io/#46a254d99d01279068126cadd6148637

src/lib/template-instance.ts
```diff
@@ -78,34 +79,36 @@ export class TemplateInstance {
...
+        // on a single element), so this loop may not run at all.
+        while (nodeIndex < part.index) {
           nodeIndex++;
-          if (node.nodeName === 'TEMPLATE') {
-            _prepareInstance((node as HTMLTemplateElement).content);
+          if (node!.nodeName === 'TEMPLATE') {
+            _prepareInstance((node as HTMLTemplateElement).content)!;
           }
-          node = walker.nextNode();
+          if (!(node = walker.nextNode())) {
```
justinfagnani:
```suggestion
          if ((node = walker.nextNode()) !=== null) {
```

I prefer direct comparisons rather than boolean conversions.

src/lib/template-instance.ts
```diff
@@ -68,6 +68,7 @@ export class TemplateInstance {
     const parts = this.template.parts;
     let partIndex = 0;
     let nodeIndex = 0;
+    let part: TemplatePart;
```
justinfagnani: Why hoist this declaration? It's not used outside the loop.

jridgewell: Because nested template elements can advance the current part.

justinfagnani: 👍

PR #828, opened 2019-02-24
 Use a stack to create walk nested templates

jridgewell:
On the heels of #825, this unrolls our recursive template processor into a stack based one.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -62,54 +62,57 @@ export class TemplateInstance {
     const fragment =
         this.template.element.content.cloneNode(true) as DocumentFragment;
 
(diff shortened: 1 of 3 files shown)


justinfagnani:
What's the pre and post PR code size?

As I mentioned in #825 I really don't care about optimizing the nested template case, so this is interesting to me if it improves the non-nested case (by eliminating the closure as it does) or reduces code size. Or if you have a specific use case for nested templates :)

jridgewell:
30b larger, with marginal improvements in no-nested-templates case and large improvements in nested-templates case. https://jsbench.github.io/#e5870d1cf2df2701c83f0e17f0f3c067

PR #830, opened 2019-02-25
Check if value is primitive in AttributeCommitter

### LINKED ISSUE #807: Trouble setting a numeric attribute - weird case
# Description

I'm having trouble setting a numeric attribute using lit-html

The trouble is that on the page there's some 3rd party code that adds an iterator to numbers which returns values from 0 to x, where x is the number.

`AttributeCommitter` does not have a special case for dealing with numbers. It does for strings, arrays. Rest gets translated to string using iterator using https://github.com/Polymer/lit-html/blob/79f4e5764355a87b7faf2992748f34af3523ac80/src/lib/parts.ts#L78

So if I set the attribute to `5`, the attribute gets set to `01234` because of this third party iterator logic. 

I was wondering if lit-html should be able to deal with number and booleans directly rather than falling back to iterators

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

jridgewell:
Also adds an `isIterable` type guard.

Fixes #807.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    Unreleased section, uncommenting the header as necessary.
 -->
 
-<!-- ## Unreleased -->
+## Unreleased
+
 <!-- ### Changed -->
 <!-- ### Added -->
 <!-- ### Removed -->
-<!-- ### Fixed -->
+### Fixed
+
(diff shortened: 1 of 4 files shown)


ruphin:
LGTM

jridgewell:
Done.

justinfagnani:
Thanks!

PR #831, opened 2019-02-25
Adopt+upgrade template fragments after processing for parts

### LINKED ISSUE #561: Template Result injected into wrong part of dom
# Bug

When combined with custom elements, lit-html injects the template result into the wrong element on the page. Tested on Chrome 69.0.3497.100 and Firefox 63.0b13 Windows 10 64-bit.

## Steps to replicate:
[repo](https://github.com/Toxyxer/invalid-dom-injection)
```
npm i
npm start
```
Open file dist/index.html in a browser supporting custom elements.

jridgewell:
Importing the node can cause any custom elements to upgrade, and non-compliant custom elements can cause DOM mutations.

If this happens before we process the template parts, our `part.index`s walking will be broken. So, it has to wait till afterwards.

But we still have to return upgraded elements before the template part's interact with the DOM. Luckily, we just adopt and upgrade them after the processing.

Fixes #561.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -16,7 +16,6 @@
  * @module lit-html
  */
 
-import {isCEPolyfill} from './dom.js';
 import {Part} from './part.js';
 import {RenderOptions} from './render-options.js';
 import {TemplateProcessor} from './template-processor.js';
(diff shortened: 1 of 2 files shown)


ruphin:
The original issue outlined in #561 is that we don't know if the clone-then-adopt approach has a significant [performance impact](https://github.com/Polymer/lit-html/issues/561#issuecomment-430003288) compared to the direct import we have now.

This change does simplify Custom Element handling and fixes several edge cases, but we should get some benchmarks and measure the performance impact before making this change.

jridgewell:
https://jsbench.github.io/#7949baf66c5f9a0a49f478c2a3ee927e

I can't really determine if it's faster or slower, the stddev is really high. Might be due to GC?

justinfagnani:
This is the more correct ordering, so ultimately I'd like to take this change.

On the micro benchmark, it seems clear that in the presence of custom elements, the difference is negligible, and neither benchmark have real script work in between steps, which might reduce the difference even further into noise.

PR #832, opened 2019-02-25
Cache only on the TemplateStringsArray

jridgewell:
This is a refactoring of #579.

The `keyString` caching from #579 permantly caches every `Template`, which holds its `<template>`. This isn't ideal. So, we can't use the `TemplateStringsArray` string values as a caching key.

The issue here is that we want to prevent expensive computations as much as possible. With rendering (and re-rendering when changing to a new template), there are two big computations:

1. `<template>` creation and finding `TemplatePart`s
2. Thrashing a DOM tree

`#1` is due to creating a new `Template`. `#2` is due to creating a `TemplateInstance` from that `Template` and committing it to a container, which could have this exact DOM tree already rendered.

The current code uses the identity equality of `TemplateStringsArray` to get the same `Template` every time, preventing `#1`. It then uses the identity equality of the `Template` instance to avoid a `#2`'s thrash.

- - -

This PR works around these issues by deferring the heavy computations until as late as possible.

First, I've made `Template` instances cheap to create. They now hold onto the `TemplateResult`, and do no further work until necessary. This allows me to only cache based on the `TemplateStringsArray`. Even if Safari forgets to properly cache it, creating a new `Template` doesn't hurt much.

Second, I've changed the re-render check to use deep equality of the `Template`. Because the `Template` now holds onto the `TemplateStringsArray`, I can just grab those and check if they're identity equal (good browsers) or string value equal (Safari).

If we determine the two `Template`s are equal, we can prevent the trash just like normal. If they aren't equal, we then do the two heavy computations.

This leads to an almost ideal caching situation. Unfortunately:

- a `TemplateInstance` holds the `Template`
- a `Template` holds the `TemplateResult`
- a `TemplateResult` holds the `TemplateStringsArray`
- a `TemplateStringsArray` weakly holds the `Template` (via the `WeakMap`)

So we still have a retain cycle. But, once a `NodePart` rerenders with any other value, that cycle is broken. That's not too bad, and it's the best we can do.

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -100,3 +100,22 @@ export class SVGTemplateResult extends TemplateResult {
     return template;
   }
 }
+
+export const templateResultsEqual =
+    (a: TemplateResult, b: TemplateResult): boolean => {
+      if (a === b) {
+        return true;
+      }
+      // Some environments (:ahem: Safari) are buggy...
+      const aStrings = a.strings;
+      const bStrings = b.strings;
+      if (aStrings.length !== bStrings.length) {
+        return false;
+      }
+      for (let i = 0; i < aStrings.length; i++) {
+        if (aStrings[i] !== bStrings[i]) {
+          return false;
+        }
+      }
+      return true;
+    };
--- a/src/lib/template.ts
+++ b/src/lib/template.ts
@@ -41,11 +41,32 @@ export const boundAttributeSuffix = '$lit$';
  * An updateable Template that tracks the location of dynamic parts.
  */
 export class Template {
-  parts: TemplatePart[] = [];
-  element: HTMLTemplateElement;
+  private _parts: TemplatePart[]|undefined = undefined;
+  private _element: HTMLTemplateElement|undefined = undefined;
+  result: TemplateResult;
 
-  constructor(result: TemplateResult, element: HTMLTemplateElement) {
-    this.element = element;
+  constructor(result: TemplateResult) {
+    this.result = result;
+  }
+
+  get element(): HTMLTemplateElement {
+    if (!this._element) {
+      this._process();
+    }
+    return this._element!;
+  }
+
+  get parts(): TemplatePart[] {
+    if (!this._parts) {
+      this._process();
+    }
+    return this._parts!;
+  }
+
+  private _process(): void {
+    const parts = this._parts = [] as TemplatePart[];
+    this._element = this.result.getTemplateElement();
+
     let index = -1;
     let partIndex = 0;
     const nodesToRemove: Node[] = [];
@@ -82,7 +103,7 @@ export class Template {
             while (count-- > 0) {
               // Get the template literal section leading up to the first
               // expression in this attribute
-              const stringForPart = result.strings[partIndex];
+              const stringForPart = this.result.strings[partIndex];
               // Find the attribute name
               const name = lastAttributeNameRegex.exec(stringForPart)![2];
               // Find the corresponding attribute
@@ -95,7 +116,7 @@ export class Template {
               const attributeValue =
                   (node as Element).getAttribute(attributeLookupName)!;
               const strings = attributeValue.split(markerRegex);
-              this.parts.push({type: 'attribute', index, name, strings});
+              parts.push({type: 'attribute', index, name, strings});
               (node as Element).removeAttribute(attributeLookupName);
               partIndex += strings.length - 1;
             }
@@ -116,7 +137,7 @@ export class Template {
                   (strings[i] === '') ? createMarker() :
                                         document.createTextNode(strings[i]),
                   node);
-              this.parts.push({type: 'node', index: ++index});
+              parts.push({type: 'node', index: ++index});
             }
             // If there's no text, we must insert a comment to mark our place.
             // Else, we can trust it will stick around after cloning.
@@ -141,7 +162,7 @@ export class Template {
               parent.insertBefore(createMarker(), node);
             }
             lastPartIndex = index;
-            this.parts.push({type: 'node', index});
+            parts.push({type: 'node', index});
             // If we don't have a nextSibling, keep this node so we have an end.
             // Else, we can remove it to save future costs.
             if (node.nextSibling === null) {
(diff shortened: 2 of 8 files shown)


src/lib/template-result.ts
```diff
@@ -100,3 +100,22 @@ export class SVGTemplateResult extends TemplateResult {
     return template;
   }
 }
+
+export const templateResultsEqual =
+    (a: TemplateResult, b: TemplateResult): boolean => {
+      if (a === b) {
```
ruphin: This should be `a.strings === b.strings`.

Or better, `aStrings === bStrings` and placed after the constant definitions below.

ruphin:
I wrote a basic test that checks how WeakMap works with TemplateStringsArrays: https://codepen.io/ruphin/pen/wOwZoY?editors=0011

As far as I can tell, it never clears the key/value pair from the map. Because this refactor still keeps a map of TemplateStringsArrays to Templates, the Templates are still never cleared from memory, which means this refactor effectively does nothing.

ruphin:
From a logical perspective, I believe the original implementation was already optimal. In all cases (new template, existing template and equal to previous template, and existing template and not equal to previous template) it performs the minimum number of operations required (where the relevant operations are object equality check, strings equality check, and template instantiation).

The real question is if we want to actively purge from cache any templates that are not currently rendered.

jridgewell:
> The functional change of this PR is that templates are no longer permanently cached, but are cached only when they are currently rendered in some part.

Not at all, they're still cached in the `templateCache` via a tagged template literals' reused `TemplateStringsArray`. In compliant browsers, you'll always get the same exact `Template` instance.

This change is specifically for Safari. When it's `TemplateStringsArray` is purged (in error), it'll generate a new `Template`. So this PR makes `Templates` cheep to create. Then the deep equal check will prevent a re-render.

If we were to go from template A to a completely different template B, then everything acts as it did before. And going back to template A doesn't cause a new `Template` to be created (as long as the browser caches the `TemplateStringsArray`), and doesn't create a new `<template>` element or process the template to find `TemplatePart`s.

Ie, after the first render of template A, its processing is done. Same for after the first render of template B. When going from A to B, only the DOM rendering happens. No other heavy computations. And in Safari, when it goes from template A to template A2 (because purged the strings), the template A2 isn't reprocessed and the DOM isn't re-rendered.

> I wrote a basic test that checks how WeakMap works with TemplateStringsArrays: https://codepen.io/ruphin/pen/wOwZoY?editors=0011

This test is flawed, you're never releasing the tagged template itself. You have to set `getTemplateStringsArray` to null or place the entire code in a scope that is itself cleaned up. Try on https://output.jsbin.com/wobolap/12/quiet.

screenshot 2019-02-25 10 34 01

> Also, I'm not sure if this is the behavior we want. With normal use it is expected that some templates will appear/disappear regularly, and this change would purge those templates from cache causing them to be parsed/built more than once.

Only if the rendering function which holds the tagged template literal is purged. That's how the old design before #579 worked.

> In all cases (new template, existing template and equal to previous template, and existing template and not equal to previous template) it performs the minimum number of operations required (where the relevant operations are object equality check, strings equality check, and template instantiation)

Equality checks are cheap, and a deep array equality check is relatively cheap too. But when compared to walking a template and rendering, they don't matter in the slightest.

ruphin:
> This test is flawed, you're never releasing the tagged template itself. You have to set `getTemplateStringsArray` to null or place the entire code in a scope that is itself cleaned up. Try on https://output.jsbin.com/wobolap/12/quiet.

I don't quite understand. What I'm seeing is that the key still exists after the timeout: `TemplateStringsArray after Timeout true`. You are getting the same result judging from the screenshot you posted. Am I interpreting those results wrong?

> This change is specifically for Safari. When it's `TemplateStringsArray` is purged (in error), it'll generate a new `Template`.

It doesn't. There are two levels of caching, one with the TemplateStringsArray and one with the joined Strings. If Safari purges the TemplateStringsArray objects, the second cache will hit and use the cached Template.

Can you describe a specific scenario where the old implementation does not correctly use a cached Template, or performs more than necessary equality checks? I can't think of one.

jridgewell:
> I don't quite understand. What I'm seeing is that the key still exists after the timeout: `TemplateStringsArray after Timeout true`. You are getting the same result judging from the screenshot you posted. Am I interpreting those results wrong?

You're confusing two different parts. Of course `TemplateStringsArray after Timeout true` is the correct behavior. But that's because you still have a reference to the tagged template via the `getTemplateStringsArray` function. `TemplateStringsArray` are cached based on the source text location, and holding onto the function holds onto the source text location.

Once you free the function though (either by it's scope being reclaimed or by manually nulling the function reference), it will be freed. My results are showing the 78kb from the `TemplateStringsArray` being freed.

> If Safari purges the TemplateStringsArray objects, the second cache will hit and use the cached Template
> Can you describe a specific scenario where the old implementation does not correctly use a cached Template, or performs more than necessary equality checks?

In the current code, that is true, it will perform no unnecessary checks. The issue is that it does this by _permanently_ (for the lifetime of the page) holding onto that memory.

This PR gets the same overall benefits (no expensive computations) without permanently holding onto that memory.

src/lib/template-result.ts
justinfagnani: This is partially correct, depending on the intent of the function. If it's as you'd expect from the name, we want to check for identity first. Then you're right in that we do want to also check for `a.strings === b.strings`.

But I suspect the naming isn't quite right here...
```suggestion
      if (a === b) {
```
```suggestion
      if (a === b) {
```
```suggestion
      if (a === b) {
```
```suggestion
      if (a === b) {
```

src/lib/template.ts
```diff
@@ -41,11 +41,32 @@ export const boundAttributeSuffix = '$lit$';
  * An updateable Template that tracks the location of dynamic parts.
  */
 export class Template {
-  parts: TemplatePart[] = [];
-  element: HTMLTemplateElement;
+  private _parts: TemplatePart[]|undefined = undefined;
+  private _element: HTMLTemplateElement|undefined = undefined;
+  result: TemplateResult;
```
justinfagnani: I don't like that a template is holding on to a result that created it. TemplateResults should be fairly transient, and this will also hold onto the values a Template was first rendered with.

justinfagnani:
A couple of notes:
- This adds 74 bytes to master
- I'm not sure it matters too much right now that we hold on to templates and template strings. Code is pretty rarely unloaded I believe, and that's actually the behavior from before the template literal revision to the ES spec. Edge still hasn't implemented the revision that would allow for tempalte strings to be GC'ed.

ruphin:
Overall I'm not a fan of this change. None of the codebases or applications that I have seen ever drop references to templates, so this change would not make any difference other than open up the opportunity for developers to explicitly remove templates from cache, but I don't think there is any demand for this.

Given the temporary nature of the issue, I prefer the current architecture.

PR #845, opened 2019-03-02
Fix rendering comment binding followed by attribute

### LINKED ISSUE #613: lack of suffix ! handling in template.js causing errors
> The syntax is suffix !: **identifier!** removes null and undefined from the type of identifier  (TypeScript)

without this part of logic in the generated  template.js, it's causing errors...

line 88 in **template.ts**
`const name = lastAttributeNameRegex.exec(stringForPart)![2];`

v.s.

line 74 in **template.js**
`const name = lastAttributeNameRegex.exec(stringForPart)[2];`

and

line 102 in **template.ts**
`const attributeValue = node.getAttribute(attributeLookupName)!;

const strings = attributeValue.split(markerRegex);`

v.s.

line 87 in **template.js**
`const attributeValue = node.getAttribute(attributeLookupName);

const strings = attributeValue.split(markerRegex);`

jridgewell:
We forgot to update the `partIndex`, so when the attribute tried to
execute the `lastAttributeNameRegex` on the last string (determined by
the `partIndex`), it would fail to match anything.

I'm guessing this will fix #613.

### DIFF
--- a/src/lib/template.ts
+++ b/src/lib/template.ts
@@ -181,6 +181,7 @@ export class Template {
             // TODO (justinfagnani): consider whether it's even worth it to
             // make bindings in comments work
             this.parts.push({type: 'node', index: -1});
+            partIndex++;
           }
         }
       }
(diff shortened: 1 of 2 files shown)


jridgewell:
Rebased.

PR #846, opened 2019-03-02
Remove unnecessary type assertions

rictic:
This is the result of turning on the no-unnecessary-type-assertion tslint pass and running `--fix`, then fixing up a few tests using `Deferred` and adding an additional assertion in places where TypeScript couldn't infer that a variable was always assigned before being read.

I like this lint pass because it makes the code a bit easier to read and understand. I will admit though that I'm proposing this change because we use this lint pass downstream, and it's annoying seeing the lint warnings.

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -321,7 +321,7 @@ export class NodePart implements Part {
     if (partIndex < itemParts.length) {
       // Truncate the parts array so _value reflects the current state
       itemParts.length = partIndex;
-      this.clear(itemPart && itemPart!.endNode);
+      this.clear(itemPart && itemPart.endNode);
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/lib/parts.ts
```diff
@@ -321,7 +321,7 @@ export class NodePart implements Part {
     if (partIndex < itemParts.length) {
       // Truncate the parts array so _value reflects the current state
       itemParts.length = partIndex;
-      this.clear(itemPart && itemPart!.endNode);
+      this.clear(itemPart && itemPart.endNode);
```
ruphin: Wouldn't `this.clear(itemPart!.endNode)` also work? I'm not super familiar with TypeScript so it might be functionally different, but it seems more in line with general TypeScript style.

rictic: The `!` has no runtime effect, it's the programmer saying to the type checker "I know you can't tell, but I can, and that expression can never be null or undefined here"

rictic:
Looks like travis push is failing, but it's already failing on master with the same errors: https://travis-ci.org/Polymer/lit-html/builds/500536611

src/lib/parts.ts
ruphin: Ah, I see. I thought it functioned like a Safe Navigation Operator. Thanks :)

PR #849, opened 2019-03-03
Early exit template creation

jridgewell:
We know how many parts are in the template by the number of values that the `TemplateResult` is holding. So, when we've found that many parts, we can stop traversing the template tree and go back to rendering that DOM.

This makes the [stack](https://github.com/Polymer/lit-html/pull/828) based traversal even more appealing. 😃

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -132,7 +132,7 @@ export class TemplateInstance {
           // We've exhausted the content inside a nested template element.
           // Because we still have parts (the outer for-loop), we know:
           // - There is a template in the stack
-          // - The walker will find a nextNode outside the temlpate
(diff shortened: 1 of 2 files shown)


jridgewell:
Rebased. This should be an easy perf win.

justinfagnani:
Thanks, was just about to rebase this myself.

justinfagnani:
Re perf: in general I think we're at the point of diminishing returns on template and even instance prep. I need to do another dive into profiles, and @aomarks is doing a ton with benchmarking right now, but I've seen cost dominated by DOM operations and prep operations taking miniscule timeslices. We need to re-verify this and document it in the repo to help guide further perf-oriented work.

PR #850, opened 2019-03-03
Shorten lastAttributeNameRegex a bit

jridgewell:
The Unicode control characters include all space characters except " ". So the attribute name character class doesn't need to include them, they're already part of the `\0-\x1F` range.


ruphin:
LGTM

Issue #851, opened 2019-03-03
Failing master build


stramel:
It appears that @jridgewell's changes in #831 might have broke some tests?

It is coming from this line https://github.com/Polymer/lit-html/blob/9bdf622e1aaaf7f82ccf9d19989a3bd7c13febba/src/lib/template-instance.ts#L120 

https://travis-ci.org/Polymer/lit-html/builds/500903769

stramel:
Sorry, looks like it started failing on [build #1759](https://travis-ci.org/Polymer/lit-html/builds/496823775) ~9 days ago

PR #852, opened 2019-03-03
Run all browsers when testing on travis

jridgewell:
So that #851 doesn't happen again.


jsilvermist:
Isn't this not running at the moment due to sauce credentials?

jridgewell:
Looks like it, which is unfortunate.

justinfagnani:
Yeah, we can't do this, though we would love to.

PR #853, opened 2019-03-03
Fix tests in Chrome 41

### LINKED ISSUE #851: Failing master build

jridgewell:
- Use getter to override `HTMLTemplateElement.content`
  - It doesn't like to override the `content` if it's a data descriptor. A getter works fine, strangely.
- Use appendChild to avoid an infinite loop in ShadyDOM

Fixes #851.

### DIFF
--- a/src/polyfills/template_polyfill.ts
+++ b/src/polyfills/template_polyfill.ts
@@ -43,8 +43,9 @@ export const initTemplatePolyfill = (forced = false) => {
     Object.defineProperties(template, {
       content: {
         ...descriptor,
-        writable: false,
-        value: content,
+        get() {
+          return content;
+        },
       },
       innerHTML: {
         ...descriptor,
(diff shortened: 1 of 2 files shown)


stramel:
Thanks @jridgewell 👍

jridgewell:
The ShadyDOM infinite loop is actually a `@webcomponents/custom-elements` infinite loop: https://github.com/webcomponents/custom-elements/blob/67f5a9620aeb06c0b0d09816e4dcea3124b08810/src/Patch/Element.js#L114-L115.

I'm doing `this.innerHTML = '<div>'` in `MutateInConstructor`'s constructor, and that code does `document.createElement('mutates-in-constructor')`!

PR #854, opened 2019-03-03
Add test for rendering attribute-like parts

ruphin:
This is currently broken: 

```js
render(html`<div>foo bar=${ "baz" }</div>`, document.body);

// Produces
 > "foo bar$lit$=baz"
```

### DIFF
--- a/src/test/lib/render_test.ts
+++ b/src/test/lib/render_test.ts
@@ -120,6 +120,13 @@ suite('render()', () => {
           stripExpressionMarkers(container.innerHTML), '<div>foo </div>');
     });
 
+    test('renders parts that look like attributes', () => {
+      render(html`<div>foo bar=${'baz'}</div>`, container);
+      assert.equal(
+          stripExpressionMarkers(container.innerHTML),
+          '<div>foo bar=baz</div>');
+    });
+
     test('renders multiple parts per element, preserving whitespace', () => {
       render(html`<div>${'foo'} ${'bar'}</div>`, container);
       assert.equal(


jridgewell:
This is due to #640. In the current code, it's specifically the `boundAttributeSuffix` in https://github.com/Polymer/lit-html/blob/9bdf622e1aaaf7f82ccf9d19989a3bd7c13febba/src/lib/template-result.ts#L60-L66

When we're doing the walk in `Template` constructor, we can correct it in the [`Text` node processing](https://github.com/Polymer/lit-html/blob/master/src/lib/template.ts#L128-L131). It'll be expensive, though, requiring a `lastAttributeNameRegex` check and and replace.

ruphin:
I don't like the current implementation. I don't understand why we start off with a well-formed StringsArray, only form a blob of HTML where we lose a lot of information, and then proceed to parse all the TextNodes and attributes to get that information back.

I prefer the old implementation, which was similar to this: https://github.com/ruphin/lite-html/blob/master/src/lib/template-parser.js#L49-L90

I think this is an opportunity to create a similar parser that works well with #847, by inserting the correct commentNodes and passing on the relevant information that is required by #847, so we can skip most of the work in the Template constructor as well.

jridgewell:
That would require correctly parsing in `getHTML`. Even the in the old system, we'd still have to parse elements, comments, and text nodes (eg, `<style>${'foo'}</style>` is gonna be a text node even though `getHTML` would inject a `<!--{{lit}-->`).

I think getting a compliant parser is going to cost more than keeping the current system.

ruphin:
The parser doesn't have to be fully compliant. It just needs to pass along all information it has access to so we don't have to collect this information again.

I'll see if I can brew up a prototype

PR #855, opened 2019-03-04
Fix attribute-like text rendering "$lit$"

jridgewell:
Also gives us a chance to use a faster check when looking for bound attributes, since we're already paying for the bytes.

Fixes #854, and includes the test from #854.


ruphin:
I grant permission.

PR #857, opened 2019-03-04
set this.value conditionally

stramel:
Move setting this.value into the value changed branch for boolean attributes


stramel:
Ping @justinfagnani

PR #859, opened 2019-03-05
Mangle private properties

jridgewell:
This enables mangling of private properties by giving them the `__foo` naming convention. Really, that's just to distinguish them from the `protected _bar` pattern that was introduced with the Part APIs.

Just this small change saves us 63b after compressing.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -55,7 +55,7 @@ export class TemplateInstance {
     }
   }
 
-  clone(): DocumentFragment {
+  _clone(): DocumentFragment {
     // There are a number of steps in the lifecycle of a template instance's
     // DOM fragment:
     //  1. Clone - create the instance fragment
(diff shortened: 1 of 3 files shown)


justinfagnani:
I like this, but feel a little strange that we require a much more specific Terser config. It would be nice if Terser supported mangling properties annotated with `@private`.

src/lib/template-instance.ts
```diff
@@ -41,20 +41,20 @@ export class TemplateInstance {
...
-    for (const part of this._parts) {
+    for (const part of this.__parts) {
       if (part !== undefined) {
         part.commit();
       }
     }
   }
 
-  _clone(): DocumentFragment {
+  clone(): DocumentFragment {
```
justinfagnani: I don't think we should rename this until 2.0.

jridgewell: Undone.

jridgewell:
Done.

PR #860, opened 2019-03-05
Add test for comment bindings

ruphin:
This adds a test that is currently failing due to incorrect processing of bindings in comments:

```
html`<p>${'foo'}<!-- ${'baz'} -->${'bar'}</p>`;

// renders: <p>foo -->bar</p>
```

Reproduction:
https://codepen.io/ruphin/pen/jJVrXd?editors=1010


mihaisavezi:
so it's a bug

jridgewell:
Closed by https://github.com/Polymer/lit-html/issues/882

PR #864, opened 2019-03-12
Prevent empty styles from causing problems in `shady-render`

### LINKED ISSUE #760: Exception when using an empty style tag
# Description
If you add an empty style tag to your template, rendering fails

# Steps to Reproduce

```
  render() {
    return html`
    <style></style>
     Hello
    `;
  }
}
```
## Expected Results
"Hello" is rendered

## Actual Results
```
Uncaught (in promise) TypeError: Cannot read property 'cloneNode' of null
    at prepareTemplateStyles (shady-render.ts:169)
    at Function.exports.render (shady-render.ts:281)
    at HTMLElement.LitElement.update (lit-element.ts:53)
    at HTMLElement.UpdatingElement.performUpdate (updating-element.ts:216)
    at HTMLElement.eval (updating-element.ts:687)
```

https://3xvfkz.stackblitz.io
## Browsers Affected

- [X] Chrome
- [X] Firefox
- [X] Edge
- [X] Safari 11
- [X] Safari 10
- [X] IE 11

sorvell:
Fixes #760.

ShadyCSS eliminates empty styles. This caused 2 problems in `shady-render` when using native Shadow DOM:
1. When re-adding the style to the rendered content, the code assumed a style would exist after ShadyCSS handled the content. It now checks.
2. When modifying the lit template used for subsequent renders, the style could be eliminated. This would break node ordering. A dummy style is now inserted in this case.

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -162,8 +162,18 @@ const prepareTemplateStyles =
       if (window.ShadyCSS!.nativeShadow) {
         // When in native Shadow DOM, re-add styling to rendered content using
         // the style ShadyCSS produced.
-        const style = template.element.content.querySelector('style')!;
-        renderedDOM.insertBefore(style.cloneNode(true), renderedDOM.firstChild);
+        const style = template.element.content.querySelector('style');
+        // Note, an empty style will be eliminated so don't try to re-insert it.
+        if (style !== null) {
+          renderedDOM.insertBefore(
+              style.cloneNode(true), renderedDOM.firstChild);
+          // However, we do need to ensure the template has a style in it if
+          // ShadyCSS eliminated it.
+        } else {
+          template.element.content.insertBefore(
+              document.createElement('style'),
+              template.element.content.firstChild);
+        }
       } else {
         // When not in native Shadow DOM, at this point ShadyCSS will have
         // removed the style from the lit template and parts will be broken as a
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -160,26 +160,19 @@ const prepareTemplateStyles =
       // will actually render so that it can update the style inside when
       // needed (e.g. @apply native Shadow DOM case).
       window.ShadyCSS!.prepareTemplateStyles(template.element, scopeName);
-      if (window.ShadyCSS!.nativeShadow) {
+      const style = content.querySelector('style');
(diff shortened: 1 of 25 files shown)


src/lib/shady-render.ts
```diff
@@ -162,8 +162,18 @@ const prepareTemplateStyles =
       if (window.ShadyCSS!.nativeShadow) {
         // When in native Shadow DOM, re-add styling to rendered content using
         // the style ShadyCSS produced.
-        const style = template.element.content.querySelector('style')!;
-        renderedDOM.insertBefore(style.cloneNode(true), renderedDOM.firstChild);
+        const style = template.element.content.querySelector('style');
+        // Note, an empty style will be eliminated so don't try to re-insert it.
+        if (style !== null) {
```
justinfagnani: nit: I do prefer positive equality checks first, reverse the if blocks?

src/lib/shady-render.ts
```diff
@@ -162,8 +162,18 @@ const prepareTemplateStyles =
...
+          renderedDOM.insertBefore(
+              style.cloneNode(true), renderedDOM.firstChild);
+          // However, we do need to ensure the template has a style in it if
+          // ShadyCSS eliminated it.
+        } else {
+          template.element.content.insertBefore(
```
justinfagnani: This seems like the right spot for the comment about Shady CSS and empty style nodes

src/lib/shady-render.ts
```diff
@@ -162,8 +162,18 @@ const prepareTemplateStyles =
       if (window.ShadyCSS!.nativeShadow) {
         // When in native Shadow DOM, re-add styling to rendered content using
         // the style ShadyCSS produced.
-        const style = template.element.content.querySelector('style')!;
-        renderedDOM.insertBefore(style.cloneNode(true), renderedDOM.firstChild);
+        const style = template.element.content.querySelector('style');
```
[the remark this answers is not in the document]
sorvell: Done.

src/lib/shady-render.ts
sorvell: Done

src/lib/shady-render.ts
sorvell: Updated

sorvell:
@justinfagnani The cleanup logic simplified slightly, PTAL.

scottcorgan:
When will this empty style fix get released? (also, thanks for fixing it)

PR #865, opened 2019-03-12
Typo: fix common misspelling

rictic:
Yet another minor lint warning that keeps bugging me


arthurevans:
I do not know why the build would fail. But I can't see as how this would cause such a problem...

PR #875, opened 2019-03-22
Call customElements.upgrade() only when customElements exists.

justinfagnani:
Should make Safari 10 and others pass.

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -114,10 +114,14 @@ export class TemplateInstance {
       partIndex++;
     }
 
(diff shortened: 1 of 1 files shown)


justinfagnani:
Tests weren't passing for me with a local WCT run though... Travis kept showing an exception inside the guarded code, like it wasn't picking up my changes. Seeing if the PR will use the right change.

justinfagnani:
Oh wow, I didn't realize that Safari didn't have `customElements.upgrade`. We'll have to revert changes to TemplateInstance

Issue #861, opened 2019-03-06
CE v0 elements inside lit-html templates not upgrading


rictic:
This change: https://github.com/Polymer/lit-html/commit/5840b3a5a569b729a53cbd1f6ea7526d3407ceff#diff-287797daa6e077e3d776f3d8d71dd0b3L64

Appears to have broken the use case of creating Polymer v1 (custom elements v0) elements inside of a lit-html template. The elements are rendered, but they are not upgraded.

Reverting the change to `fragment`'s initializer resolves the issue.

Issue occurs in Chrome with native CE v0, haven't yet reprod with polyfilled CE v0. External version of b/127388674

jridgewell:
The code is using `customElements.upgrade`, so if that's not throwing you've got to have either a polyfill layer or you're running CEv1. But, Chrome 71 should be running CEv1, right?

rictic:
Chrome has CEv1, but the app that encountered this is written with CEv0 (it's migrating to CEv1, but not finished)

jridgewell:
I see now:

```js
document.registerElement('x-el', {
  prototype: Object.create(HTMLElement.prototype, {
    createdCallback: {
      value() {
        console.log("created");
      },
    },
  }),
});

const template = document.createElement('template');
template.innerHTML = `<x-el></x-el>`;

// This won't fire DRE upgrade.
const frag = template.content.cloneNode(true);
document.adoptNode(frag);
// Note this "upgrades" CEv1,
// but that doesn't help our DRE element.
customElements.upgrade(frag);

// Even inserting won't upgrade the DRE element
document.body.appendChild(frag);
```

rictic:
This has been fixed at master, probably by #876 as jridgewell says

Woot!

PR #876, opened 2019-03-22
Revert the change to template instance clone/upgrade ordering to fix Safari

justinfagnani:
Safari doesn't yet implement `CustomElementRegistry#upgrade`

### DIFF
--- a/src/lib/template-instance.ts
+++ b/src/lib/template-instance.ts
@@ -16,6 +16,7 @@
  * @module lit-html
  */
 
+import {isCEPolyfill} from './dom.js';
 import {Part} from './part.js';
 import {RenderOptions} from './render-options.js';
 import {TemplateProcessor} from './template-processor.js';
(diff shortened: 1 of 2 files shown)


src/lib/template-instance.ts
```diff
@@ -55,12 +56,47 @@ export class TemplateInstance {
...
+    //
+    // The resulting order is on native is: Clone, Adopt, Upgrade, Process,
+    // Update, Connect. document.importNode() performs Clone, Adopt, and Upgrade
+    // in one step.
+    //
+    // The Custom Elements v1 polyfill supports upgrade(), so the order when
+    // polyfilled is the more ideal: Clone, Process, Adopt, Upgrade, Update,
+    // Connect.
+
+    const fragment = isCEPolyfill ?
```
kevinpschaaf: Why isn't this just switched on existence of `customElements.upgrade` (and fall back to `importNode` when no upgrade exists)?

justinfagnani: I considered that but that means we'd have slightly different behavior across browsers with native custom elements, rather than different between native and polyfill. Since this is also the released ordering, I thought it was safer.

src/test/lib/render_test.ts
```diff
@@ -1046,15 +1046,18 @@ suite('render()', () => {
           assert.isTrue(instance.calledSetter);
         });
 
-    test('does not upgrade elements until after parts are established', () => {
-      render(
-          html`
+    // We need Safari to implement customElements.upgrade before we can enable
+    // this.
+    test.skip(
```
jridgewell: If we land https://github.com/Polymer/lit-html/pull/847, we can re-enable regardless of Safari.

justinfagnani:
@jridgewell Does this really fix #861? #861 is a problem with the release version, which uses this ordering.

Polymer 1 uses CEv0 so we'd have to detect that as a polyfill too and use `cloneNode()` (native should be fine).

src/test/lib/render_test.ts
justinfagnani: I'll take a look at that ASAP.

src/lib/template-instance.ts
kevinpschaaf: Ok so after discussion, since browsers on CE polyfill also likely need to work with ShadyDOM, keeping CE polyfill using `upgrade` offers better compatibility with elements that create polyfilled "ShadyRoots" in the constructor (which is only possible on CE polyfill because light DOM children created sync to constructor by Shady DOM would be prohibited under native CE).

This isn't an issue with native SD+CE because creating real ShadowRoots is legal in the constructor, and also hidden from the lit-html treewalk, hence we're biasing the native case to `importNode` to avoid Safari's lack of upgrade, incur fewer treewalks, and as a more conservative change from the latest released state for native browsers.

jridgewell:
Per https://github.com/Polymer/lit-html/issues/861#issuecomment-469945779, it's not an ordering issue. It's that custom elements registered via `document.registerElement` is not upgradeable when calling `customElements.upgrade`

PR #863, opened 2019-03-12
[WIP] Refactor template parser

ruphin:
This is a refactor of the template parsing system. It is a work in process, making this PR solely for testing purposes.

PR #882, opened 2019-04-01
Fix bindings in comments

### LINKED ISSUE #881: Bindings in comments broken
# Description
Bindings in comments cause the rest of the comment text, including the `-->`, to be rendered to HTML.

See #860 

# Steps to Reproduce

1. Write this code

```ts
import {html, render} from 'lit-html';

render(html`<!-- ${'foo'} bar -->`, document.body);
```

2. See this output...

bar -->

## Live Reproduction Link
https://codepen.io/ruphin/pen/jJVrXd?editors=1010

## Expected Results
Comments remain comments

## Actual Results
See above

## Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

justinfagnani:
Fixes #881 

This is a more incremental fix than #847 or #863. Those may be better in the long term, but we want to measure the perf and size impacts of those first.

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -45,28 +45,51 @@ export class TemplateResult {
   getHTML(): string {
     const l = this.strings.length - 1;
     let html = '';
-    let isTextBinding = true;
+    let isCommentBinding = false;
+
     for (let i = 0; i < l; i++) {
       const s = this.strings[i];
-      const match = lastAttributeNameRegex.exec(s);
-      const close = s.lastIndexOf('>');
-      // We're in a text position if the previous string closed its last tag, an
-      // attribute position if the string opened an unclosed tag, and unchanged
-      // if the string had no brackets at all:
+      // For each binding we want to determine the kind of marker to insert
+      // into the template source before it's parsed by the browser's HTML
+      // parser. The marker type is based on whether the expression is in an
+      // attribute, text, or comment poisition.
+      //   * For node-position bindings we insert a comment with the marker
+      //     sentinel as its text content, like <!--{{lit-guid}}-->.
+      //   * For attribute bindings we insert just the marker sentinel for the
+      //     first binding, so that we support unquoted attribute bindings.
+      //     Subsequent bindings can use a comment marker because multi-binding
+      //     attributes must be quoted.
+      //   * For comment bindings we insert just the marker sentinel so we don't
+      //     close the comment.
       //
-      // "...>...": text position. open === -1, close > -1
-      // "...<...": attribute position. open > -1
-      // "...": no change. open === -1, close === -1
-      isTextBinding = match === null && (close > -1 || isTextBinding) &&
-          s.indexOf('<', close + 1) === -1;
-
-      if (isTextBinding) {
-        html += s + nodeMarker;
-      } else if (match) {
-        html += s.substr(0, match.index) + match[1] + match[2] +
-            boundAttributeSuffix + match[3] + marker;
+      // The following code scans the template source, but is *not* an HTML
+      // parser. We don't need to track the tree structure of the HTML, only
+      // whether a binding is inside a comment, and if not, if it appears to be
+      // the first binding in an attribute.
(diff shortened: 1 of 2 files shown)


src/lib/template-result.ts
```diff
@@ -43,33 +43,34 @@ export class TemplateResult {
    * Returns a string of HTML used to create a `<template>` element.
    */
   getHTML(): string {
-    const endIndex = this.strings.length - 1;
+    const l = this.strings.length - 1;
     let html = '';
-    for (let i = 0; i < endIndex; i++) {
+    let isTextBinding = true;
```
jridgewell: We could change this to look for comment bindings instead?

```js
let isCommentBinding = false;
for (let i = 0; i < l; i++) {
  const s = this.strings[i];
  const open = s.lastIndexOf('<!--');
  isCommentBinding = (open > -1 || isCommentBinding) &&
    s.indexOf('-->', open + 1) === -1;

  if (isCommentBinding) {
    html += s + marker;
    continue;
  }
  const match = lastAttributeNameRegex.exec(s);
  if (match) {
    html += s.substr(0, match.index) + match[1] + match[2] +
      boundAttributeSuffix + match[3] + marker;
  } else {
    html += s + nodeMarker;
  }
}
```

justinfagnani:
@jridgewell it is in the weeds a bit, but I think it's pretty basic to support commenting out HTML inside a template during development. It's quite difficult to do this with JS comments as you have to break up the template literal. Also some IDEs add an HTML comment when running the comment-line command inside an `html`-tagged template literal.

src/test/lib/render_test.ts
```diff
@@ -269,6 +269,12 @@ suite('render()', () => {
       assert.equal(container.querySelector('p')!.textContent, 'baz');
     });
 
+    test('does not inject text from comments with bindings', () => {
+      const t = html`<p>${'foo'}<!-- ${'baz'} ${'baz'} -->${'bar'}</p>`;
```
jridgewell: We should add tests with a commented-out element with attr bindings, and a child node binding. I don't think the current solution is enough.

justinfagnani: Good point

src/test/lib/render_test.ts
```diff
@@ -269,12 +269,28 @@ suite('render()', () => {
...
+    test('handles elements in comments with attribute bindings', () => {
+      const t = html`A<!-- <div foo=${'bar'}>B</div> -->C`;
+      render(t, container);
+      // Use innerText instead of textContent because of a crazy bug in
+      // Chrome 41 where textContent would include the textContent of comments!
+      assert.equal(container.innerText, 'AC');
+    });
+
+    test('handles elements in comments with node bindings', () => {
+      const t = html`A<!-- <div>${'B'}</div> -->C`;
```
jridgewell: Ditto.

src/test/lib/render_test.ts
```diff
@@ -269,12 +269,28 @@ suite('render()', () => {
...
+      const t = html`<p>A<!-- ${'B'} ${'C'} -->D</p>`;
       render(t, container);
       // Use innerText instead of textContent because of a crazy bug in
       // Chrome 41 where textContent would include the textContent of comments!
-      assert.equal(container.querySelector('p')!.innerText, 'foobar');
+      assert.equal(container.querySelector('p')!.innerText, 'AD');
+    });
+
+    test('handles elements in comments with attribute bindings', () => {
+      const t = html`A<!-- <div foo=${'bar'}>B</div> -->C`;
```
jridgewell: I think you'll need one more binding after the `</div>` to hit the failure case.

justinfagnani:
So far this adds 51 bytes to the bundle size. Not great, but gives us a more realistic baseline to better compare the more invasive template prep rewrite PRs.

src/lib/template-result.ts
justinfagnani: Thanks! Added this with a couple of changes

PR #887, opened 2019-04-18
Run tachometer on Travis and report as a GitHub Check

aomarks:
Test and report benchmark results for commits and PRs.

- Runs https://github.com/PolymerLabs/tachometer on Travis.
- For branch pushes or local PRs, the results will show up as a GitHub Check, next to where the Travis build status shows.
- For PRs from forks, we don't get access to Travis secure variables, so we can't report a GitHub Check. The results will show up in the Travis logs, though, and once the commit is merged, will show up as a Check on that commit.
- We round-robin between three versions of lit-html: the commit being tested, its first parent (e.g. for a pull request to master, the tip of master), and the latest version of lit-html published to NPM.
- The report panel looks like this:

  tach
- I'll be working on making this more readable soon, by reporting the differences in the opposite direction, and with markdown instead of ASCII.


blikblum:
When i [added a benchmark](https://github.com/Swizz/snabbdom-pragma/pull/29) to a library, i considered running it at Travis but got a doubt about the reliability of the results. Travis, and other cloud services, provision an amount of server (cpu/memory) resources for each build that may vary depending of several factors. So this difference between runs could not influence the benchmark results?

justinfagnani:
@blikblum: @aomarks has done a lot of work to ensure we can reliably use the results even on Travis.

aomarks:
> When i [added a benchmark](https://github.com/Swizz/snabbdom-pragma/pull/29) to a library, i considered running it at Travis but got a doubt about the reliability of the results. Travis, and other cloud services, provision an amount of server (cpu/memory) resources for each build that may vary depending of several factors. So this difference between runs could not influence the benchmark results?

Yes, definitely that's a very valid concern! Because of that, the way we decided to integrate these benchmarks is that for each PR/commit, we round-robin between 50+ samples each of the PR/commit, the previous commit, and the latest version published to NPM. So this way the differences we report for a PR/commit all come from measurements made on the same Travis instance at the same time, instead of being based on historical data which may have been measured on a system with different performance characteristics.

PR #891, opened 2019-04-26
Bump tachometer version

aomarks:
Results are now more useful, and use markdown:

https://github.com/Polymer/lit-html/runs/111364626

The most interesting thing to look at is the "this" row, which shows that this commit is indistinguishable from master ("parent"), and is a little faster than NPM published. You can also see now see that master is already a little faster than NPM (which is why this commit is too). Could probably still do with some improvements to readability, but at least the useful numbers are now shown if you know where to look.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -783,9 +783,9 @@
       "dev": true
     },
     "@types/cookies": {
-      "version": "0.7.1",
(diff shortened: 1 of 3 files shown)


rictic:
Ignore the robot, I've got multiple browser sessions open :)

PR #895, opened 2019-05-01
Experiment: Use a Range to remove nodes

justinfagnani:
It used to be that Range was slower than manual removal. This may not be the case anymore. Let's see how the benchmark looks, though we'll need local testing on more browsers.

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -44,15 +44,25 @@ export const reparentNodes =
       }
     };
 
+
+const range = document.createRange();
+
 /**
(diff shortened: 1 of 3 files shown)


src/lib/dom.ts
```diff
@@ -44,15 +44,25 @@ export const reparentNodes =
...
+        return;
+      }
+      range.setStartBefore(start);
+      if (end === null) {
+        range.setEndAfter(container.lastChild!);
+      } else {
+        range.setEndBefore(end);
       }
+      range.deleteContents();
+      range.collapse();
```
jridgewell: Careful, `range` still holds onto the container element after this.

PR #896, opened 2019-05-02
Fix test in Edge

straversi:
Edge renders attributes in differing order.


straversi:
Given there were already two orderings being tested against, I figure this issue has come up before.

If a different ordering comes up in the future, we may want to just add all six permutations of the attribute order to this test.

PR #897, opened 2019-05-03
Set NodePart initial value to nothing, only clear if current value is nothing

justinfagnani:
This guards against accessing `Node.parentNode` and `Node.nextSibling` in some very common cases including initial render. The binding cost is high enough that avoiding the accesses should have a noticeable, is small, effect on performance.

This is a subset of #895 which should be a win in all cases, where #895's use of Range isn't fully vetted yet.


justinfagnani:
lit-html 2.0 handles this a bit better, but we need a test for `noChange`: https://github.com/Polymer/lit-html/issues/1753

PR #904, opened 2019-05-07
Allow `render` to accept any type.

bicknellr:
`render` uses `NodePart` to render the given value and `NodePart#setValue` accepts `unknown`. This change propagates the `unknown` to `render`, which only accepts `TemplateResult` currently but doesn't make use of the specificity.

This also fixes a bug where a `NodePart` that is currently rendered as a single text node will throw if  asked to render a symbol next.

---
Part of changes for Polymer/lit-element#618.

### DIFF
--- a/src/lib/render.ts
+++ b/src/lib/render.ts
@@ -24,13 +24,13 @@ import {templateFactory} from './template-factory.js';
 export const parts = new WeakMap<Node, NodePart>();
 
 /**
- * Renders a template to a container.
+ * Renders a template or value to a container.
  *
  * To update a container with new values, reevaluate the template literal and
  * call `render` with the new result.
  *
- * @param result a TemplateResult created by evaluating a template tag like
- *     `html` or `svg`.
+ * @param result Any value renderable by NodePart - typically a TemplateResult
+ *     created by evaluating a template tag like `html` or `svg`.
  * @param container A DOM parent to render to. The entire contents are either
  *     replaced, or efficiently updated if the same result type was previous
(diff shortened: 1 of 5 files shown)


src/lib/render.ts
```diff
@@ -20,18 +20,17 @@ import {removeNodes} from './dom.js';
 import {NodePart} from './parts.js';
 import {RenderOptions} from './render-options.js';
 import {templateFactory} from './template-factory.js';
-import {TemplateResult} from './template-result.js';
 
 export const parts = new WeakMap<Node, NodePart>();
 
 /**
- * Renders a template to a container.
+ * Renders a template or value to a container.
```
justinfagnani:
```suggestion
 * Renders a template result or other value to a container.
```

src/test/lib/parts_test.ts
```diff
@@ -124,6 +124,53 @@ suite('Parts', () => {
         assert.equal(stripExpressionMarkers(container.innerHTML), '');
       });
 
+      test('accepts a symbol', () => {
+        part.setValue(Symbol());
+        part.commit();
+        assert.equal(stripExpressionMarkers(container.innerHTML), 'Symbol()');
```
justinfagnani: To fix these tests with the Symbol polyfill, I think you want to toString the actual Symbol passed it:

```js
const symbol = Symbol();
part.setValue(symbol);
part.commit();
assert.equal(stripExpressionMarkers(container.innerHTML), String(symbol));
```

PR #909, opened 2019-05-13
Use ES2015 `String#endsWith` instead of a local implementation.

bicknellr:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

### DIFF
--- a/src/lib/template.ts
+++ b/src/lib/template.ts
@@ -85,7 +85,7 @@ export class Template {
           // assume a correspondence between part index and attribute index.
           let count = 0;
           for (let i = 0; i < length; i++) {
-            if (endsWith(attributes[i].name, boundAttributeSuffix)) {
+            if (attributes[i].name.endsWith(boundAttributeSuffix)) {
               count++;
             }
           }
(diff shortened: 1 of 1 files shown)


jridgewell:
Note, this implementation came from profiling: https://jsbench.github.io/#2845ec98f02d336a31ae52139dfd8056. I'm fine with whichever.

bicknellr:
Huh, that's pretty surprising that native `endsWith` loses to slice. Maybe I should just add a comment about this instead?

jridgewell:
I think it's worth having @justinfagnani decide. There's definitely a speed vs file-size trade off.

bicknellr:
Also, it looks like whatever compiler we're using doesn't insert `endsWith`.

justinfagnani:
`endsWith` isn't supported by IE11: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Browser_compatibility

And we're not using a compiler that injects polyfills. We use TypeScript, but only to compile out types.

PR #910, opened 2019-05-13
Scope styles even if the rendered value is not a TemplateResult.

bicknellr:
This should have been part of #904. The ShadyDOM/CSS-aware renderer will only take the actions to scope styles if the value being rendered is a TemplateResult. Before #904, the rendered value was always assumed to be a TemplateResult and these checks were redundant. However, to implement Polymer/lit-element#618, the style-scoping steps need to happen even if the value isn't a TemplateResult because some user (e.g. lit-element) might have added adopted style sheets for the render container.


src/lib/shady-render.ts
```diff
@@ -272,12 +273,11 @@ export const render =
...
         parts.delete(renderContainer);
-        if (part.value instanceof TemplateInstance) {
-          prepareTemplateStyles(
-              renderContainer as DocumentFragment,
-              part.value.template,
-              scopeName);
-        }
+        const template = part.value instanceof TemplateInstance ?
+            part.value.template :
+            new Template(html``, document.createElement('template'));
```
justinfagnani: What's this part for? Add a comment.

bicknellr: Added, let me know what you think.

justinfagnani: Is this really going to work? `prepareTemplateStyles` might insert styles into this template, but for what purpose if the template is never rendered anywhere?

I get the feeling that this isn't going to work with the original use-case of rendering `undefined`, then rendering a `TemplateResult`. Can we add a test specifically for that, using shady-render?

sorvell: I think the gist is that the styling we are about in this case was sent to ShadyCSS via `prepareAdoptedCssText`. Styling in the non-template result would just not be supported.

So it seems like all we need to do in this case is ensure that (1) styles are removed from any lit template for the scope (would there be any if no TemplateResult is being rendered?, (2) ensure that `prepareTemplate` is called (because this is necessary to apply any styles defined with `prepareAdoptedCssText`).  Why not change `prepareTemplateStyles` to optionally take a Template rather than creating a dummy one here?

bicknellr: > Is this really going to work? `prepareTemplateStyles` might insert styles into this template, but for what purpose if the template is never rendered anywhere?
>
> I get the feeling that this isn't going to work with the original use-case of rendering `undefined`, then rendering a `TemplateResult`. Can we add a test specifically for that, using shady-render?

Ok, I added another test specifically for this case.

> I think the gist is that the styling we are about in this case was sent to ShadyCSS via `prepareAdoptedCssText`. Styling in the non-template result would just not be supported.

AFAICT, the current setup seems to be that [all styles in the first render are scoped](https://github.com/Polymer/lit-html/blob/f37bf689fb836a009f1a6a413bc9e7d7262964ae/src/lib/shady-render.ts#L131) and [any styles that appear later aren't](https://github.com/Polymer/lit-html/blob/f37bf689fb836a009f1a6a413bc9e7d7262964ae/src/lib/shady-render.ts#L121-L125). So I think that even styles inserted into NodeParts of a top-level TemplateInstance were collected / scoped / removed from their rendered position on the first render.

Am I correct in thinking that (local) `prepareTemplateStyles`' `renderedDOM` contains the complete rendered output of the first render by the time it runs?

> So it seems like all we need to do in this case is ensure that (1) styles are removed from any lit template for the scope (would there be any if no TemplateResult is being rendered?, (2) ensure that `prepareTemplate` is called (because this is necessary to apply any styles defined with `prepareAdoptedCssText`). Why not change `prepareTemplateStyles` to optionally take a Template rather than creating a dummy one here?

I was avoiding this because it seemed difficult to pry apart the Template editing parts of (local) `prepareTemplateStyles` from the parts that were dealing with changing the output of the first render but it seems necessary at this point to match the earlier behavior - I'll update.

bicknellr: I made `prepareTemplateStyles`' `template` parameter optional and updated the logic in there to work with a new, empty template if it isn't passed. Also, I added a couple extra tests that ShadyCSS mixins and scoping are applied to styles that are part of an initial render for a scope of a non-TemplateInstance value.

justinfagnani:
The PR now has unrelated changes in it from master. Can you rebase?

justinfagnani:
I guess that's the merge of #904

src/lib/shady-render.ts
```diff
@@ -153,18 +158,22 @@ const prepareTemplateStyles =
       removeStylesFromLitTemplates(scopeName);
       // And then put the condensed style into the "root" template passed in as
       // `template`.
-      const content = template.element.content;
-      insertNodeIntoTemplate(template, condensedStyle, content.firstChild);
+      const content = templateElement.content;
+      if (template) {
+        insertNodeIntoTemplate(template, condensedStyle, content.firstChild);
+      } else {
+        content.insertBefore(condensedStyle, content.firstChild);
```
sorvell: Because of the comment below this, I think this case breaks support for `@apply` when using native Shadow DOM. That *might* be ok, but at the least we need to document it.

src/lib/shady-render.ts
```diff
@@ -275,12 +284,18 @@ export const render =
...
-              part.value.template,
-              scopeName);
-        }
+        // ShadyCSS might have style sheets (e.g. from `prepareAdoptedCssText`)
+        // that should apply to `renderContainer` even if the rendered value is
+        // not a TemplateInstance. However, it will only insert scoped styles
+        // into the document if `prepareTemplateStyles` has already been called
+        // for the given scope name. `prepareTemplateStyles` requires a
+        // template element (and the local wrapper requires a Template), so we
+        // create an empty one here to satisfy it.
```
sorvell: Update this comment since we're not creating an empty template here.

src/test/lib/render_test.ts
```diff
@@ -1409,4 +1409,43 @@ suite('render()', () => {
       assert(container.innerHTML, '<div></div>');
     });
   });
+
+  // `render` directly passes the given value to `NodePart#setValue`, so these
+  // tests are really just a sanity check that they are accepted by `render`.
+  // Tests about rendering behavior for specific values should generally be
+  // grouped with those of `NodePart#setValue` and `#commit`.
+  suite('accepts types other than TemplateResult', () => {
```
sorvell: Let's add a test for accepting a Node since that seems like it might not be uncommon.

src/lib/shady-render.ts
sorvell: We should be able to add a test that demonstrates this.

src/lib/parts.ts
```diff
@@ -245,15 +245,16 @@ export class NodePart implements Part {
   private __commitText(value: unknown): void {
     const node = this.startNode.nextSibling!;
     value = value == null ? '' : value;
+    const valueAsString: string =
```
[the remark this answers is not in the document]
bicknellr: Added the comment.

src/lib/shady-render.ts
bicknellr: Updated.

src/test/lib/render_test.ts
bicknellr: Added at the bottom.

src/lib/shady-render.ts
bicknellr: I added the comment we were working on in offline discussion.

bicknellr:
> It's unclear why styling is incorrect after you eventually render a TemplateResult (without this modification). Perhaps the underlying issue is that the user expects `:host` styles to render initially (when rendering something like `undefined`)?

Yeah, the `:host` styles should render when the first result is `undefined`. (Trying to remember our offline chat about this...) Particularly, this is a problem when Shadow DOM is supported but `adoptedStyleSheets` is not. Assuming [this check for TemplateResult in lit-element](https://github.com/Polymer/lit-element/blob/ef7e4367fd500b02533bccb9d0b55f92e9a2aaa3/src/lit-element.ts#L215) was removed, then lit-html's checks [for TemplateResult](https://github.com/Polymer/lit-html/blob/1b404333f5f92f0e7153563e46ef6c8952112396/src/lib/shady-render.ts#L254) and [for TemplateInstance](https://github.com/Polymer/lit-html/blob/1b404333f5f92f0e7153563e46ef6c8952112396/src/lib/shady-render.ts#L278) will prevent it from running ShadyCSS scoping on the first actual render if the rendered result was `undefined`, but lit-element still [inserts the styles it created to simulate `adoptedStyleSheets`](https://github.com/Polymer/lit-element/blob/ef7e4367fd500b02533bccb9d0b55f92e9a2aaa3/src/lit-element.ts#L227-L231). Then, the first time that a TemplateResult is produced by render, it will run the scoping but also [remove all of the content of `renderContainer`](https://github.com/Polymer/lit-html/blob/1b404333f5f92f0e7153563e46ef6c8952112396/src/lib/shady-render.ts#L284), which contained the styles that were inserted by lit-element.

src/lib/parts.ts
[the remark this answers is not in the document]
justinfagnani: It's not that it's on odd value, it's that an exception was thrown and it was a pretty bad dev experience.

src/lib/shady-render.ts
```diff
@@ -125,8 +125,13 @@ const shadyRenderSet = new Set<string>();
...
  */
 const prepareTemplateStyles =
-    (renderedDOM: DocumentFragment, template: Template, scopeName: string) => {
+    (scopeName: string, renderedDOM: DocumentFragment, template?: Template) => {
       shadyRenderSet.add(scopeName);
+      // If `renderedDOM` is stamped from a Template, then we need to edit that
+      // Template's underlying template element. Otherwise, we create one here
+      // to give to ShadyCSS, which still requires one while scoping.
+      const templateElement =
+          template ? template.element : document.createElement('template');
```
justinfagnani: Always prefer explicit comparisons over implicity ToBoolean conversions, here and in two `if` statements below.

bicknellr: Ok, these are explicit now.

PR #917, opened 2019-05-17
Use global TreeWalkers

justinfagnani:
WIP

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -56,3 +56,32 @@ export const removeNodes =
         start = n;
       }
     };
+
+/**
+ * A shared TreeWalker that iterates Elements, Comments, and Texts.
+ * We share the TreeWalker to avoid its slow construction.
+ * Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
+ */
+const walker = document.createTreeWalker(
+    document, 133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */, null, false);
+
+/**
+ * Returns a "clean" tree walker. We guarantee that it is clean by requiring it
+ * to be reset after each use. This gives avoids any issues with the TreeWalker
+ * holding onto Nodes that could otherwise be GC'd. It also ensures we don't
+ * hit any reentrancy bugs.
+ */
+export const elementCommentTextWalker = (root: Node): TreeWalker => {
+  if (walker.currentNode !== document) {
+    throw new Error('walker was not reset');
+  }
+  walker.currentNode = root;
+  return walker;
+};
+
+/**
+ * Resets the TreeWalker so that it doesn't hold onto any GC'able Nodes.
+ */
+export const resetElementCommentTextWalker = () => {
+  walker.currentNode = document;
+};
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -16,8 +16,8 @@
  * @module shady-render
  */
 
-import {isTemplatePartActive, Template, TemplatePart} from './template.js';
 import {elementCommentTextWalker, resetElementCommentTextWalker} from './dom.js';
+import {isTemplatePartActive, Template, TemplatePart} from './template.js';
 
 /**
  * Removes the list of nodes from a Template safely. In addition to removing
--- a/src/lib/modify-template.ts
+++ b/src/lib/modify-template.ts
@@ -17,8 +17,7 @@
  */
 
 import {isTemplatePartActive, Template, TemplatePart} from './template.js';
-
-const walkerNodeFilter = 133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */;
+import {elementCommentTextWalker, resetElementCommentTextWalker} from './dom.js';
 
 /**
  * Removes the list of nodes from a Template safely. In addition to removing
@@ -39,8 +38,7 @@ const walkerNodeFilter = 133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */;
 export function removeNodesFromTemplate(
     template: Template, nodesToRemove: Set<Node>) {
   const {element: {content}, parts} = template;
-  const walker =
-      document.createTreeWalker(content, walkerNodeFilter, null, false);
+  const walker = elementCommentTextWalker(content);
   let partIndex = nextActiveIndexInTemplateParts(parts);
   let part = parts[partIndex];
   let nodeIndex = -1;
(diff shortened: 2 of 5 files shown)


justinfagnani:
@jridgewell the PR looks like it's slower on a template-instance-heavier benchmark: https://github.com/PolymerLabs/benchmarks/blob/render/lit-html/render/repeat.js

```
┌─────────────┬─────────────────┐
│   Benchmark │ lit             │
├─────────────┼─────────────────┤
│      Params │ <none>          │
├─────────────┼─────────────────┤
│     Browser │ chrome-headless │
│             │ 75.0.3770.38    │
├─────────────┼─────────────────┤
│ Sample size │ 50              │
└─────────────┴─────────────────┘

┌─────────┬───────────┬─────────────────┬─────────────────┬─────────────────┐
│ Version │ Bytes     │        Avg time │      vs walkers │       vs master │
├─────────┼───────────┼─────────────────┼─────────────────┼─────────────────┤
│ walkers │ 52.58 KiB │ 4.80ms - 5.11ms │                 │          slower │
│         │           │                 │        -        │         2% - 9% │
│         │           │                 │                 │ 0.07ms - 0.42ms │
├─────────┼───────────┼─────────────────┼─────────────────┼─────────────────┤
│ master  │ 51.76 KiB │ 4.62ms - 4.78ms │          faster │                 │
│         │           │                 │         2% - 8% │        -        │
│         │           │                 │ 0.07ms - 0.42ms │                 │
└─────────┴───────────┴─────────────────┴─────────────────┴─────────────────┘
```

jridgewell:
See https://jsbench.github.io/#d0363ded1753dec3e060e61e1794896b. It's possible the different is just GC pressure due to allocating the walker, which we probably don't hit with so few runs in tach.

stramel:
@justinfagnani Do you think it would be beneficial to have a wiki with the pertinent jsbench links for easy reference?

justinfagnani:
@jridgewell thanks for the link. The benchmarks in the render branch of the benchmarks repo are important for internal customers who made those to represent their workloads. I'm not sure we can take a 2-9% regression on repeat, though maybe we can scale it up on the grounds that it's more realistic. We've also only been running these benchmarks on Chrome so far and need to address that.

I do wonder if we really need to share only one global walker, or if one per use site gets us most of any gains, possibly without needing the reset operation for the functions as long as they remain non-reentrant.

justinfagnani:
@stramel I think the right place to reference jsbench microbenchmarks is from the implementation of a tachometer benchmark that exercises lit-html. The comment can mention the other benchmark as the reason why we think the tachometer benchmark is important.

jridgewell:
> I'm not sure we can take a 2-9% regression on repeat, though maybe we can scale it up on the grounds that it's more realistic.

What I'm saying is I don't trust the tach tests, not that the underlying tests aren't important. I'm 100% certain that creating a `TreeWalker` is slow (on the order of creating a DOM node, but with very high variability due to GC pauses that a DOM node doesn't cause). And that changing the `currentNode` of an already created walker is much faster and more consistent. https://jsbench.github.io/#26a49a6033a8a5de00d00e05bf2cb275

> I do wonder if we really need to share only one global walker, or if one per use site gets us most of any gains, possibly without needing the reset operation for the functions as long as they remain non-reentrant.

That's another option, but would require each individual walker would still need to be reset to prevent reentrancy bugs. Though, this is entirely a test concern, we could strip it out in real builds. It was just to find the bugs, like the one you already fixed!

justinfagnani:
@jridgewell I rebased this brach to start running benchmarks against master again. The shack benchmark is still showing this is 0-12% slower:

```
┌─────────────┬──────────────┐
│   Benchmark │ shack        │
├─────────────┼──────────────┤
│     Browser │ chrome       │
│             │ 79.0.3945.79 │
├─────────────┼──────────────┤
│ Sample size │ 130          │
└─────────────┴──────────────┘

┌─────────┬────────────┬─────────────────┬─────────────────┬──────────────┐
│ Version │ Bytes      │        Avg time │       vs master │   vs walkers │
├─────────┼────────────┼─────────────────┼─────────────────┼──────────────┤
│ master  │ 127.84 KiB │ 8.31ms - 8.99ms │                 │       faster │
│         │            │                 │        -        │     0% - 11% │
│         │            │                 │                 │ 0.03ms - 1ms │
├─────────┼────────────┼─────────────────┼─────────────────┼──────────────┤
│ walkers │ 129.55 KiB │ 8.82ms - 9.51ms │          slower │              │
│         │            │                 │        0% - 12% │     -        │
│         │            │                 │ 0.03ms - 1.00ms │              │
└─────────┴────────────┴─────────────────┴─────────────────┴──────────────┘
```

We also have benchmarks from the js-framework-benchmark. Non-keyed create rows from that is inconclusive:

```
┌─────────────┬─────────────────────────────────────────────┐
│   Benchmark │ js-framework-benchmark-non-keyed/index.html │
├─────────────┼─────────────────────────────────────────────┤
│     Browser │ chrome                                      │
│             │ 79.0.3945.79                                │
├─────────────┼─────────────────────────────────────────────┤
│ Sample size │ 160                                         │
└─────────────┴─────────────────────────────────────────────┘

┌─────────┬────────────┬───────────────────┬───────────────────┬───────────────────┐
│ Version │ Bytes      │          Avg time │         vs master │        vs walkers │
├─────────┼────────────┼───────────────────┼───────────────────┼───────────────────┤
│ master  │ 317.28 KiB │ 61.40ms - 63.12ms │                   │            unsure │
│         │            │                   │          -        │         -1% - +3% │
│         │            │                   │                   │ -0.37ms - +1.82ms │
├─────────┼────────────┼───────────────────┼───────────────────┼───────────────────┤
│ walkers │ 318.99 KiB │ 60.87ms - 62.21ms │            unsure │                   │
│         │            │                   │         -3% - +1% │          -        │
│         │            │                   │ -1.82ms - +0.37ms │                   │
└─────────┴────────────┴───────────────────┴───────────────────┴───────────────────┘
```

I'd like to understand why the jsbench results are so different and showing the opposite effect in tachometer benchmarks. I'll try to set up a more automated script to run and compare each of the js-framework-benchmarks individually, and probably add some benchmarks so we're not relying just on shack and those.

If you're interesting in how I set this up, I have two copies of lit-html checked out and I'm running this command:

```bash
npx tachometer shack --package-version=master=lit-html@file:/Users/justinfagnani/Projects/Lit/lit-html --package-version=walkers=lit-html@file:/Users/justinfagnani/Projects/Lit/lit-html-2 
```

From `polymer-benchmarks/lit-html`

justinfagnani:
We're using a single walker in lit-html 2.0

Issue #912, opened 2019-05-15
ES6 modules not working with try/catch in module on MS Edge


vertical-web-720:
# Description
As of Microsoft Edge 41 or earlier and the current release v1.0.0, lit-html is not working at all with Edge. This is because Edge does not support having a try/catch block into a top level of a module.

# Steps to Reproduce

1. Write this code

```ts
import {render} from 'lit-html';

console.log('This never executes in Edge');
```

2. Check on Chrome that the log is printed into its dev tool as expected

3. Check on MS Edge that the log is never printed

4. Open lit-html's lib/parts.js file and remove the top level try/catch block in it:

```ts
// try {
    const options = {
        get capture() {
            eventOptionsSupported = true;
            return false;
        }
    };
    // tslint:disable-next-line:no-any
    window.addEventListener('test', options, options);
    // tslint:disable-next-line:no-any
    window.removeEventListener('test', options, options);
/* } catch (_e) {

}*/
```

5. Check on Chrome that the log is printed into its dev tool as expected

6. Check on MS Edge that the log is printed into its dev tool as expected

## Live Reproduction Link
Generic PoC of try/catch bug on MS Edge:
https://rawgit.com/herrernst/es6-module-try-catch/master/index.html

Edge's bug related discussion with potential solutions:
https://github.com/Microsoft/ChakraCore/issues/4038

## Expected Results
The log message prints into Edge's dev tool

## Actual Results
Edge never reached the console.log() line because it is stuck by the lit-html/lib/parts.js' top level try/catch block

## Browsers Affected
- [ ] Chrome
- [ ] Firefox
- [x] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

jridgewell:
Interesting. Is there an issue if we wrap the try catch in an IIFE? If not, would you like to make the PR?

vertical-web-720:
I can confirm that wrapping it in an IIFE works. I'll be glad to make the PR (I'll open it through my main account @astrinxit66)

PR #918, opened 2019-05-18
Fix support of try/catch block in Edge

astrinxit66:
As discussed in #912, this is necessary for lit-html to run in Edge <= v41.

Fixes https://github.com/Polymer/lit-html/issues/912.


justinfagnani:
It'd be really great to have failing tests so we don't hit this again. We test [Edge 15 and 17 on Sauce](https://github.com/Polymer/lit-html/blob/master/.travis.yml#L18), which I guess are the EdgeHTML versions and possibly correspond to Edge 40.15063 and 42.17134?

If we could get a Sauce version string that fails before the PR I'll add it.

astrinxit66:
I agree for having a test but to me this is far behind my skills unfortunately, I can't find solutions for those questions: 
- how to target a failure for a specific browser?
- how to catch the failure case as when it fails in Edge it just stuck there? 
- if we assume that Edge can reach the assertion point of our failure test case what should we assert because Edge does not produce any error of any kind (no exception, nothing) for this bug?

I would be very interested to see what we can do!

justinfagnani:
@astrinxit66 I think the only thing we really need is to know a real version of Edge that this fails on. The Sauce Labs `microsoftedge@15` and `microsoftedge@17` browser tags do not hit this error, even though it seems like `microsoftedge@15` _should_ be Edge <= 41 for the consumer version.

astrinxit66:
Ok, I have the bug on `Edge 41.16299.1004.0` with `EdgeHTML 16.16299`.

astrinxit66:
Ok, I have the bug on `Edge 41.16299.1004.0` with `EdgeHTML 16.16299`.

jridgewell:
Ping @justinfagnani

PR #919, opened 2019-05-18
Use a single tree walker, ensure it's reset

jridgewell:
Note, this is a PR into your [`walkers`](https://github.com/Polymer/lit-html/tree/walkers) branch, not master. We'll still need to merge https://github.com/Polymer/lit-html/pull/917 after this.

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -56,3 +56,32 @@ export const removeNodes =
         start = n;
       }
     };
+
+/**
+ * A shared TreeWalker that iterates Elements, Comments, and Texts.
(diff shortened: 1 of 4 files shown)


justinfagnani:
Thanks. I'll merge this into my PR and continue from there. The benchmarks results aren't conclusive that this is a big win, apparently TreeWalker creation isn't that slow. I'm adding some new benchmarks that should be more sensitive to template instance creation time to be sure.

jridgewell:
I think https://github.com/Polymer/lit-html/issues/914 would be a good test case for the perf impact.

PR #922, opened 2019-05-21
Fix: html tag will now always throw on invalid escape sequences

### LINKED ISSUE #348: Unicode characters inside style of a html template are not recognized
Flloing js file is returning "undefined" for the policy. 

```
import {html} from '@polymer/lit-element/lit-element.js';

export const policy = html`
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">

<head>
    <meta content="text/html; charset=UTF-8" http-equiv="content-type"></meta>
    <style type="text/css">
        .list>li:before {
            content: "\0025a0  "
        }
    </style>
    <title>Test unicode in lit-html </title>
</head>

<body>
    <ul class="list">
        <li> list item 1</li>
        <li> list item 2</li>
    </ul>
</body>

</html>`
```

RunDevelopment:
This fixes that TemplateResult's `getHTML` returned `undefined` if the template strings array contains `undefined`. The new behavior is that `getHTML` always throw a reference error in that case.

`undefined` in the template strings array usually happens if the template literal contains invalid escape sequences. This doesn't affect raw strings but `html` doesn't use these and changing this would be a breaking change.

I intentionally choose to make the error a standard reference error. A more detailed error message will help developers find the cause (it's not obvious that (and which) escape sequences are at fault) but this will be at the cost of file size.
I will be happy to add a better error message if the maintainers of lit-html think it's worth it.

---

This fixes #348.

### DIFF
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -52,7 +52,7 @@ export class TemplateResult {
       // For each binding we want to determine the kind of marker to insert
       // into the template source before it's parsed by the browser's HTML
       // parser. The marker type is based on whether the expression is in an
-      // attribute, text, or comment poisition.
+      // attribute, text, or comment position.
       //   * For node-position bindings we insert a comment with the marker
       //     sentinel as its text content, like <!--{{lit-guid}}-->.
       //   * For attribute bindings we insert just the marker sentinel for the
@@ -92,7 +92,7 @@ export class TemplateResult {
             marker;
       }
     }
-    html += this.strings[l];
+    html += this.strings[l].toString();
     return html;
   }
 
--- a/src/lib/template-result.ts
+++ b/src/lib/template-result.ts
@@ -47,52 +47,60 @@ export class TemplateResult {
     let html = '';
     let isCommentBinding = false;
 
-    for (let i = 0; i < l; i++) {
+    for (let i = 0; i <= l; i++) {
       const s = this.strings[i];
-      // For each binding we want to determine the kind of marker to insert
-      // into the template source before it's parsed by the browser's HTML
-      // parser. The marker type is based on whether the expression is in an
-      // attribute, text, or comment position.
-      //   * For node-position bindings we insert a comment with the marker
-      //     sentinel as its text content, like <!--{{lit-guid}}-->.
-      //   * For attribute bindings we insert just the marker sentinel for the
-      //     first binding, so that we support unquoted attribute bindings.
-      //     Subsequent bindings can use a comment marker because multi-binding
-      //     attributes must be quoted.
-      //   * For comment bindings we insert just the marker sentinel so we don't
-      //     close the comment.
-      //
-      // The following code scans the template source, but is *not* an HTML
-      // parser. We don't need to track the tree structure of the HTML, only
-      // whether a binding is inside a comment, and if not, if it appears to be
-      // the first binding in an attribute.
-      const commentOpen = s.lastIndexOf('<!--');
-      // We're in comment position if we have a comment open with no following
-      // comment close. Because <-- can appear in an attribute value there can
-      // be false positives.
-      isCommentBinding = (commentOpen > -1 || isCommentBinding) &&
-          s.indexOf('-->', commentOpen + 1) === -1;
-      // Check to see if we have an attribute-like sequence preceeding the
-      // expression. This can match "name=value" like structures in text,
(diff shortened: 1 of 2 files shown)


RunDevelopment:
I signed it!

RunDevelopment:
May I ask what I can do about the failed test?

I don't know what the error message is supposed to communicate except that the test setup had a hiccup.

<details>
<summary>Error</summary>

```
firefox 67               Tests failed: {"message":"[get(\"http://localhost:8081/components/lit-html/generated-index.html?cli_browser_id=1\")] Error response status: 13, UnknownError - An unknown server-side error occurred while processing the command. Selenium error: Failed to decode response from marionette\nBuild info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'\nSystem info: host: 'travis-job-5da673d9-b286-4acb-a455-226c854ccc23', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-...
```

</details>

jridgewell:
I restarted the test. Looks like travis flaked.

src/lib/template-result.ts
```diff
@@ -92,7 +92,7 @@ export class TemplateResult {
             marker;
       }
     }
-    html += this.strings[l];
+    html += this.strings[l].toString();
```
jridgewell: There's the `s` variable above that will also need a `toString` call.

src/test/lib/template-result_test.ts
```diff
@@ -32,4 +33,15 @@ suite('TemplateResult', () => {
...
   });
+
+  test('invalid escape sequences always throw an error', () => {
+    function throws(template: TemplateResult) {
+      assert.throw(() => template.getHTML());
+    }
+
+    throws(html`<style> .foo:before { content: "\0025a0"; } </style>`);
+    throws(html`\0025a0${'bar'}\0025a0`);
+    throws(html`\0025a0${'bar'}foo`);
```
jridgewell: I'm surprised this one throws an error due to my above comment.

src/lib/template-result.ts
RunDevelopment: That's not necessary because [the `.lastIndexOf` on next line](https://github.com/Polymer/lit-html/blob/c9cb653b0a005d5e53e899ab4911e9ac208cc23b/src/lib/template-result.ts#L69) will do the type check for us.

jridgewell: Ahh, ok.

RunDevelopment: I didn't add the `toString` for `s` to save the precious 11 characters of the call.
But if you think that this makes the code clearer, I'll gladly add it.

src/test/lib/template-result_test.ts
RunDevelopment: Explanation [here](https://github.com/Polymer/lit-html/pull/922#discussion_r287429628).

RunDevelopment:
Seems like restarting the test didn't work. Same error again.

jridgewell:
Might be some issue with wct on travis? I'm not really sure what to do there.

src/lib/template-result.ts
justinfagnani: I know it's more verbose, but I'd consider being explicit here and just checking for undefined. The error otherwise will be confusing.

```ts
if (this.strings[l] === undefined) {
  throw new Error(`Invalid escape sequence in ${strings.raw[i]}`);
}
html += this.strings[l]
```

justinfagnani:
Firefox 67 is broken under webdriver. If you rebase to master you'll pick up the travis config change that pins to 66.

RunDevelopment:
Thanks @justinfagnani! Tests pass now.

jridgewell:
Ping @RunDevelopment, can you update with the suggestion from https://github.com/Polymer/lit-html/pull/922#discussion_r287436614?

RunDevelopment:
Thanks for pinging! Will do.

RunDevelopment:
@jridgewell Is this ready to merge now?

jridgewell:
LGTM, but it requires a review from @justinfagnani before I can merge.

Issue #923, opened 2019-05-21
Add syntax for spreading attributes, properties, and event listeners


straversi:
It would be nice to be able to spread bindings in lit-html. One would be able to spread properties, attributes, and event listeners.

Syntax could look something like this:
```javascript
const spreadable = { '.property': 1, 'attribute': 'foo', '@click': handler }
html`<div ...=${spreadable}>How convenient</div>`
```

This feature request was discussed at length in this pull request: https://github.com/Polymer/lit-html/pull/213

We're investigating the tradeoff between convenience and bloat associated with this feature. This issue is for feature tracking.

straversi:
Here's an implementation of this feature: https://github.com/Polymer/lit-html/pull/925

It added **0.18 KB** to the minified and gzipped package size, for a new size of **3.46 KB**.

Questions:
* How much do people want this feature?
* Is it worth the size cost?
* Is there another more cost-effective way to implement this? (as a directive...?)

daKmoR:
uuhhh nice 🤗  - I am probably biased but I would say it's worth the cost 🤗  especially as I would consider it sort of a core features the same as `?` or `@` or `.`...

ruphin:
The updating mechanism for this is fundamentally different from how current property/attribute setters work. How will caching/updating things work? What if you spread one less property than in the previous render, does it remove the property from the node? Does it remember the previous value for every property ever spread indefinitely, so it can do updates only when the values changed?

straversi:
> How will caching/updating things work?

At least in #925, Parts are still created for each key-value pair in a spread object. The intention is for each attribute/property/event in a spread object to keep its `setValue` and `commit` behavior, including dirty checking.

> What if you spread one less property than in the previous render, does it remove the property from the node?

Our thinking is that if a key-value pair is removed from the spread object, that attribute/property/event would also be removed from the element.

Westbrook:
I am very excited for how this spreads attributes. The possibility of simplifying passing and template writing is real here. Hope some important questions can find answers/documentation to move this forward.

However, I'm a little wary that the work to form an object to `{'.value': bar, '@click': baz, '?disabled', bot}` is just about the same as:
```
<input
  .value=${bar}
  @click=${baz}
  ?disabled=${bot}
/>
```
With the various part types applied via the same `...` attribute I'm not sure the extension of the syntax really buys much. I'd also say this would be better as _only_ an attribute or _only_ a property spreader, but I feel there will be as much confusion on either side of the barrier.

[...]

PR #925, opened 2019-05-21
Add spread syntax

straversi:
Implementation of spread syntax (#923) to help answer the question "How much will this feature cost?"

Size comparison:
```
master
---
Bundle Size:  9.02 KB
Minified and Gzipped Size:  3.28 KB
Brothli size: 2.97 KB

this branch
---
Bundle Size:  9.59 KB (+0.57 KB)
Minified and Gzipped Size:  3.46 KB (+ 0.18 KB)
Brothli size: 3.13 KB (+ 0.16 KB)
```

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -42,52 +42,51 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
  */
 export class SpreadPart implements Part {
   value: unknown = undefined;
-  subparts: Map<string, ReadonlyArray<Part>> = new Map();
+  subparts: Map<string, Part> = new Map();
   element: Element;
-  handleAttribute: (name: string) =>  ReadonlyArray<Part>;
+  handleAttribute: (name: string) => Part;
   // Names of attributes that were spread in the last commit,
   // but are not anymore.
   private __removedNames: Array<string> = [];
 
-  constructor(element: Element, handleAttribute: (name: string) =>  ReadonlyArray<Part>) {
+  constructor(element: Element, handleAttribute: (name: string) => Part) {
     this.element = element;
     this.handleAttribute = handleAttribute;
   }
 
   setValue(value: unknown): void {
-    let spreadable = value as {[key: string]: any};
-    for (var name in spreadable) {
-      this.subparts.set(name, this.handleAttribute(name))
+    const spreadable = value as {[key: string]: unknown};
+    for (const name in spreadable) {
+      this.subparts.set(name, this.handleAttribute(name));
     }
-    for (var [name, parts] of this.subparts.entries()) {
-      if (parts) {
-        // Only one part per key is allowed via Spread
-        parts[0].setValue(spreadable[name]);
+    for (const [name, part] of this.subparts.entries()) {
+      part.setValue(spreadable[name]);
+      if (!(name in spreadable)) {
         // Prepare to remove unused attributes
-        if (!(name in spreadable)) {
-          this.__removedNames.push(name)
-        }
+        this.__removedNames.push(name);
       }
     }
     this.value = value;
   }
 
   commit() {
-    for (var parts of this.subparts.values()) {
-      if (parts) {
-        // Only one part per key is allowed via Spread
-        parts[0].commit()
-      }
+    for (const part of this.subparts.values()) {
+      part.commit();
     }
     // Remove attributes that used to be spread, but aren't anymore.
     // Removal of properties and events are already handled by committing
     // the undefined value.
-    for (var name of this.__removedNames) {
-      let parts = this.subparts.get(name) || [];
-      if (parts && parts[0] instanceof AttributePart) {
-        this.element.removeAttribute(name)
+    for (const name of this.__removedNames) {
+      const part = this.subparts.get(name);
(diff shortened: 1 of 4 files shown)


src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+
+  constructor(element: Element, handleAttribute: (name: string) =>  ReadonlyArray<Part>) {
+    this.element = element;
+    this.handleAttribute = handleAttribute;
+  }
+
+  setValue(value: unknown): void {
+    let spreadable = value as {[key: string]: any};
+    for (var name in spreadable) {
+      this.subparts.set(name, this.handleAttribute(name))
```
justinfagnani: Maybe pull out the part here?

```suggestion
      this.subparts.set(name, this.handleAttribute(name)[0])
```

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+    }
+    for (var [name, parts] of this.subparts.entries()) {
```
justinfagnani: no `var` ever :) Use `const` in for/of loops, `let` in plain for loops. Did you get a linter warning?

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+      if (parts) {
```
justinfagnani: We try to be very explicit with conditionals, partially for clarity and partially to avoid ToBoolean conversions.

In this case I want to know what you're checking for, that `parts` is not `undefined, or that it's not empty. Either way are we sure the check is necessary?

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+        // Only one part per key is allowed via Spread
+        parts[0].setValue(spreadable[name]);
```
justinfagnani: should this be in a conditional too?

```js
if (name in spreadable) {
  parts[0].setValue(spreadable[name]);
} else {
  this.__removedNames.push(name);
}
```

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+        // Prepare to remove unused attributes
+        if (!(name in spreadable)) {
+          this.__removedNames.push(name)
```
justinfagnani: semicolon

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+        }
+      }
+    }
+    this.value = value;
+  }
+
+  commit() {
+    for (var parts of this.subparts.values()) {
+      if (parts) {
```
justinfagnani: ditto about the conditional check

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+        // Only one part per key is allowed via Spread
+        parts[0].commit()
+      }
+    }
+    // Remove attributes that used to be spread, but aren't anymore.
+    // Removal of properties and events are already handled by committing
+    // the undefined value.
+    for (var name of this.__removedNames) {
```
justinfagnani: `var` → `const`

src/lib/parts.ts
```diff
@@ -37,6 +37,61 @@ export const isIterable = (value: unknown): value is Iterable<unknown> => {
...
+      let parts = this.subparts.get(name) || [];
+      if (parts && parts[0] instanceof AttributePart) {
```
justinfagnani: what about PropertyPart, BooleanAttributePart and EventPart?

src/lib/parts.ts
straversi: Good idea, thanks. I tried it in the default template processor because SpreadPart shouldn't have to know.

src/lib/parts.ts
straversi: Forgot to lint 🌝

src/lib/parts.ts
straversi: You're right that this check isn't necessary.

src/lib/parts.ts
straversi: I purposefully didn't include the `name in spreadable` condition here cause sometimes we want to commit `undefined` to clean up a part. For example, if someone removes an event, committing `undefined` for the EventPart will remove the event listener.

src/lib/parts.ts
straversi: At first I didn't have PropertyPart here because we'd commit `undefined` while removing a property from a spread, but I just realized we should remove the property key from the element as well. 👍

straversi: BooleanAttributePart and EventPart both handle their own clean-up when their value gets set to `undefined` by the SpreadPart. I don't love that these cleanups happen in different locations

justinfagnani:
Closing as we need to redo this for lit-html 2.0.

PR #932, opened 2019-05-23
Change classMap to set classes correctly for SVGs

### LINKED ISSUE #930: classMap directive does not work with SVG
# Description
When using the `classMap` directive in an SVG template, I get the following error:
```
Unhandled promise rejection TypeError: "setting getter-only property "className""
```

The error comes from [this line in `class-map.ts`](https://github.com/Polymer/lit-html/blob/c9cb653b0a005d5e53e899ab4911e9ac208cc23b/src/directives/class-map.ts#L51).

According to [the SVG specification](https://www.w3.org/TR/SVG11/types.html#InterfaceSVGStylable), `className` is a readonly property. This differs from [`className` on `Element`](https://developer.mozilla.org/en-US/docs/Web/API/Element/className), which can be set.

# Steps to Reproduce

1. Write this code

```ts
import {svg, render} from 'lit-html';
import {classMap} from 'lit-html/directives/class-map';

render(svg`<circle class=${classMap({active: true})} cx="23.401" cy="29.716" r="4.301"/>`, document.body);
```

2. See this output...

## Live Reproduction Link

<https://stackblitz.com/edit/lit-html-svg-classname-bug-repro>

## Expected Results

The class `active` should have been applied to `<circle>`.

## Actual Results

The following error is thrown:
```
Unhandled promise rejection TypeError: "setting getter-only property "className""
```

## Browsers Affected

Should affect all browsers, but I only tested the ones checked below.
- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

AdamVig:
The assignment to `element.className` in `classMap` was causing the error `setting getter-only property "className"`.

According to [the SVG specification](https://www.w3.org/TR/SVG11/types.html#InterfaceSVGStylable), `className` is a read-only property. This differs from [`className` on Element](https://developer.mozilla.org/en-US/docs/Web/API/Element/className), which can be set.

Fixes #930.

### DIFF
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -48,7 +48,7 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
 
   // handle static classes
   if (!classMapCache.has(part)) {
-    element.setAttribute('class', committer.strings.join(' '));
+    element.classList.value = committer.strings.join(' ');
   }
 
   const {classList} = element;
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -48,7 +48,7 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
 
   // handle static classes
   if (!classMapCache.has(part)) {
-    element.className = committer.strings.join(' ');
+    element.setAttribute('class', committer.strings.join(' '));
   }
 
   const {classList} = element;
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -46,12 +46,12 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
   const {committer} = part;
   const {element} = committer;
 
+  const {classList} = element;
+
   // handle static classes
   if (!classMapCache.has(part)) {
-    element.classList.value = committer.strings.join(' ');
+    classList.value = committer.strings.join(' ');
   }
-
-  const {classList} = element;
 
   // remove old classes that no longer apply
   const oldInfo = classMapCache.get(part);
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -46,12 +46,12 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
   const {committer} = part;
   const {element} = committer;
 
-  const {classList} = element;
-
   // handle static classes
   if (!classMapCache.has(part)) {
-    classList.value = committer.strings.join(' ');
+    element.setAttribute('class', committer.strings.join(' '));
   }
+
+  const {classList} = element;
 
   // remove old classes that no longer apply
   const oldInfo = classMapCache.get(part);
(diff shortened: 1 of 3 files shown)


src/directives/class-map.ts
```diff
@@ -48,7 +48,7 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
 
   // handle static classes
   if (!classMapCache.has(part)) {
-    element.className = committer.strings.join(' ');
+    element.setAttribute('class', committer.strings.join(' '));
```
jridgewell: Oh, we could actually save some file size here. `classList` has a getter/setter `value` property which acts the just like `element.className` but will work for both SVG and HTML Elements.

AdamVig: Good catch! Changed.

src/directives/class-map.ts
```diff
@@ -48,7 +48,7 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
 
   // handle static classes
   if (!classMapCache.has(part)) {
-    element.className = committer.strings.join(' ');
+    element.classList.value = committer.strings.join(' ');
```
jridgewell: Sorry, should have been more clear. We already have a `classList` variable just below that we can reusue here if we move it up.

AdamVig: 😅 My mistake. I should have looked a bit more at the context.

AdamVig: Fixed.

src/directives/class-map.ts
blikblum: > classList has a getter/setter value
AFAIK IE 11 does not supports it

I'm all for dropping IE11 support but...

jridgewell: Ugh, you're right.

AdamVig:
Fixed!

For posterity, here is [the Can I Use page for `classList`](https://caniuse.com/#feat=classlist). For IE 10 and 11, it says:
> Does not have support for classList on SVG or MathML elements.
Which matters here because this pull request's specific intent is to _fix_ support for SVG elements.

jridgewell:
Ping @justinfagnani, this is good to merge.

justinfagnani:
Thanks @AdamVig and @jridgewell!

@AdamVig could you update the changelog for this? Otherwise, looks good.

AdamVig:
@justinfagnani Sure, done.

justinfagnani:
@AdamVig unfortunately, IE11 doesn't support `classList` on SVG, so CI failed on master after the merge.

Looks like Stencil ran into the same problem here: https://github.com/ionic-team/stencil/issues/1079
We can do something similar use setAttribute for SVG.

blikblum:
> IE11 doesn't support classList on SVG

Only test uses classList. The actual code already uses setAttribute, so is needed to update only the tests

justinfagnani:
No, the problem is in the `classMap` directive itself: https://github.com/Polymer/lit-html/blob/master/src/directives/class-map.ts#L60

What's happening is that this PR fixes the first error with classMap on SVG, and so exposes the second, which is IE-only.

PR #933, opened 2019-05-24
Fix typos, add notes on attribute prefixes.

arthurevans:
I believe both instances of "event" here were typos... Also added some details on "attribute name" after looking at the default template processor implementation.

Even with this change, I do not understand the description of the "strings" argument:

   * @param strings The string literals. There are always at least two strings,
       event for fully-controlled bindings with a single expression.

- "the string literals"--which string literals. Where are they coming from? Why does the default template processor mostly throw them away?
- "there are always at least two strings"--why?
- "fully-controlled bindings"--we've never defined this term, and I don't know what it means.

I think most of these questions would be clarified by an example. "For example, with a simple binding like foo=${bar}, the literals would be ..."

I know this is kind of an internal class, and maybe doesn't need the highest level of doc, but this one seemed extra cryptic.


justinfagnani:
> @param strings The string literals. There are always at least two strings,
event for fully-controlled bindings with a single expression.

> "the string literals"--which string literals. Where are they coming from? Why does the default template processor mostly throw them away?

The string literals are the string parts of a template literal expression. In

```js
html`a${b}c`
```
The string literals are `['a', 'c']`

> "there are always at least two strings"--why?

In

```js
html`${b}`
```
The string literals are `['', '']`. There are always num_expressions + 1 string literals. Since we always have at least one expression by the time we call this method, there are always two string literals.

> "fully-controlled bindings"--we've never defined this term, and I don't know what it means.

This means a binding that takes up the whole attribute or property value, like `a=${foo}` or `a="${foo}"`. In this case we're not going to do a string interpolation, but just pass the value to `setAttribute()` or assign to the property. We detect this case from the string literals being `['','']`.

Hope that helps!

arthurevans:
Updated, @justinfagnani PTAL.

PR #938, opened 2019-05-29
Conditional tests should use the value of `test` when their containing suite is called.

bicknellr:
Some tests conditionally disable groups of tests by capturing either `test` or `test.skip` from the top-level scope before the tests are run and later calling the captured function instead of `test`:

https://github.com/Polymer/lit-html/blob/1d0549426df5fc420285812f634970d92a027905/src/test/lib/render_test.ts#L22-L34

However, this can potentially prevent the test runner from associating them with the correct suite, which might be a problem if they are meant to be part of a suite with (e.g.) a `setup` function.


bicknellr:
Yeah, it's pretty strange. I found this while trying to do https://github.com/Polymer/lit-html/pull/937, where the render tests were being run twice: the tests that were using a captured `test` would run outside of all suites and those that depended on setup / teardown hooks would break.

PR #939, opened 2019-05-31
Test that shady-render works without shadow DOM or custom elements.

### LINKED ISSUE #742: shady-render requires existence of `ShadowRoot`
This effectively requires either native Shadow DOM or the polyfill to be loaded.

The issue is at https://github.com/Polymer/lit-html/blob/master/src/lib/shady-render.ts#L247 and could be easily avoided with a `window.ShadowRoot` check.

bicknellr:
This PR also adds an explicitly thrown error in shady-render if `options` with a `scopeName` property is not provided. Before, it would throw when reading `options.scopeName` if `options` wasn't provided, which appears broken even though the option is required.

Fixes #742.

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -244,6 +244,9 @@ export const render =
     (result: TemplateResult,
      container: Element|DocumentFragment|ShadowRoot,
      options: ShadyRenderOptions) => {
+      if (!options || !options.scopeName) {
+        throw new Error('The `scopeName` option is required.');
+      }
       const scopeName = options.scopeName;
       const hasRendered = parts.has(container);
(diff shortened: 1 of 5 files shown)


src/lib/shady-render.ts
```diff
@@ -244,6 +244,9 @@ export const render =
     (result: TemplateResult,
      container: Element|DocumentFragment|ShadowRoot,
      options: ShadyRenderOptions) => {
+      if (!options || !options.scopeName) {
+        throw new Error('The `scopeName` option is required.');
```
justinfagnani: Can you add a test for this exception?

bicknellr: Added; I also fixed this conditional which would still throw if `options` wasn't an object.

PR #949, opened 2019-06-15
WIP: Hydration experiment

justinfagnani:
This prototypes a hydration method where we:
 - differentiate opening and closing marker nodes so we can build a tree of markers from DOM
 - calculate a template hash from template strings
 - record the template hash in NodePart opening markers when rendering nested templates
 - add a hydrate() analog to render() that doesn’t clear the container

Seems to work for an extremely basic example. We would need to add attribute part support, likely by encoding attribute parts in comments that are first-children of the element that has the part. The extra markers are a cost, so we’d want to emit them only when pre-rendering.


src/lib/render.ts
```diff
@@ -50,6 +50,73 @@ export const render =
...
+        part = new NodePart({
+          templateFactory,
+          ...options,
+          prerenderedParts: rootPart!.children
+        });
+        part.startNode = rootPart!.startNode;
+        part.endNode = rootPart!.endNode;
+        parts.set(container, part);
+      } else {
+        part = new NodePart({
```
jridgewell: Just call `render`?

src/lib/render.ts
```diff
@@ -50,6 +50,73 @@ export const render =
...
+      let part = parts.get(container);
+
+      if (part === undefined) {
+        let rootPart: PartInfo|undefined = undefined;
+        const partStack: PartInfo[] = [];
+        const walker = document.createTreeWalker(
+            container, NodeFilter.SHOW_COMMENT, null, false);
+        let node: Comment|null;
+        while ((node = walker.nextNode() as Comment | null) !== null) {
+          if (node.nodeType === Node.COMMENT_NODE) {
```
jridgewell: No need for this check

src/lib/render.ts
```diff
@@ -50,6 +50,73 @@ export const render =
...
+                const parentInfo = partStack[partStack.length - 1];
+                if (parentInfo.children === undefined) {
+                  parentInfo.children = [];
+                }
+                parentInfo.children.push(partInfo);
+              } else {
+                console.assert(
+                    rootPart === undefined,
+                    'there should be exactly one root part in a render container');
+                rootPart = partInfo;
```
jridgewell: This'll mean you can't render something like a `repeat` inside the container. Why not just initialize a root part to begin with?

src/lib/parts.ts
```diff
@@ -267,15 +268,36 @@ export class NodePart implements Part {
...
+
+      // This is the marker content that would have been generated by this
+      // TemplateResult.
+      const markerWithHash = `lit-part ${value.digest}`;
+      if (this.startNode.textContent === markerWithHash) {
+        // We got a match, so we're rehydrating DOM pre-rendered with this
+        // TemplateResult's Template.
+
+        // we need to give the instance the parts
+        if (this.options.prerenderedParts !== undefined) {
```
jridgewell: If this is `undefined`, does that imply there are no dynamic parts inside the instance? If not, I'm not sure calling `instance.update` below will actually do anything.

justinfagnani:
We now have experimental hydration in lit-html 2.0.

PR #954, opened 2019-06-28
fix(NodePart): factory method created to help us when creating NodePart introduced due to scope hoisting problem with ParcelJS

Stradivario:
## [Original Issue ParcelJS](https://github.com/parcel-bundler/parcel/issues/3172)

# Description

The problem here is that when compiled this part of code remains the same without `export` referencing him to the `module` `scope`

Typescript
```typescript
itemPart = new NodePart(this.options);
```

Javascript
```typescript
itemPart = new NodePart(this.options);
```

Instead it should be 

```typescript
itemPart = new exports.NodePart(this.options);
```

The case here is that i am getting the following error

```typescript
Uncaught ReferenceError: NodePart is not defined
```

## Type of change

- [x] Bug fix
- [x] Not introducing any existing change with existing API

# Checklist:

- [x] Implement factory method creating new NodeParts

Example repository of the problem:

https://github.com/Stradivario/parcel-scope-hoisting

Regards! :)

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -135,6 +135,15 @@ export class AttributePart implements Part {
 }
 
 /**
+ * Creates a new NodePart instance
+ * Introduced because of scope hoisting problem with parcel js
+ * https://github.com/parcel-bundler/parcel/issues/3172
+ */
+function nodePartFactory(options: RenderOptions) {
+  return new NodePart(options);
+}
+
+/**
  * A Part that controls a location within a Node tree. Like a Range, NodePart
  * has start and end locations and can set and update the Nodes between those
  * locations.
@@ -308,7 +317,7 @@ export class NodePart implements Part {
 
       // If no existing part, create a new one
       if (itemPart === undefined) {
-        itemPart = new NodePart(this.options);
+        itemPart = nodePartFactory(this.options);
         itemParts.push(itemPart);
         if (partIndex === 0) {
           itemPart.appendIntoPart(this);


Stradivario:
I signed it!

Stradivario:
I think this issue is related with how typescript compile this part of the code.
If it is working with `exports.NodePart` manually added why TS decide to not catch this reference ?

justinfagnani:
This isn't a bug with lit-html, and adds unnecessary lines of code. We should wait for the bugfix in Parcel.

Stradivario:
> This isn't a bug with lit-html, and adds unnecessary lines of code. We should wait for the bugfix in Parcel.

I understand!

What i cannot understand is why typescript doesn't put `exports.NodePart` in front since it does it in some other cases ? Is this a bug with Typescript ? 

Regards!

Stradivario:
Other strange thing is this

When typescript compile `es6` `commonjs`  class for example:

```typescript
class NodePart {}
```

Will be compiled to:
```typescript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
class NodePart {
}
exports.NodePart = NodePart;
//# sourceMappingURL=index.js.map
```

Do you know why they decided this approach ?

```typescript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.NodePart = class NodePart {
}
//# sourceMappingURL=index.js.map
```
Since even with that simple change everything work as expected ?

Stradivario:
@justinfagnani closing since Parcel Team confirmed that it is a bug in the renaming when tree shaking the whole bundle.

https://github.com/parcel-bundler/parcel/issues/3172#issuecomment-506838854

Thank you for your fast answer and correct resolution of the problem!

Regards!

PR #956, opened 2019-07-08
Make IDE plugins links, change recommendation

rictic:
I think we should recommend runem.lit-plugin because it supports type checking of template bindings, and it seems to be robust (I ran it across all of google3 and there were only a small number of issues, which I'm sending out PRs for)


jsilvermist:
I've been using it for a while, though I disabled some warnings a while ago as it had far too many incorrect warnings.

rictic:
Yeah, in my prototyping of using it at google I'm turning off unknown attributes, unimported elements (but not unknown elements), and css parse errors. Does that match your experience?

justinfagnani:
I've recently seen false-positive errors with bindings in CSS, like:

```js
html`
  <style>
    :host {
      transform-origin: ${x}% ${y}%;
    }
  </style>
`;
```

There's an error at the semi-colon:

```
term expectedlit-plugin(2322)
```

I'll file a bug upstream, but cc @runem

rictic:
@justinfagnani related: https://github.com/runem/ts-lit-plugin/pull/23

You can also write css snippets. One pattern I've seen googler code do is:

```typescript
const size = css`100px`;
const style = css`:host { width: ${size}; height: ${size} }`;
```

Even if it had better support for templates with interpolation like the second one, the first isn't a valid css document even with no interpolation.

jsilvermist:
> Yeah, in my prototyping of using it at google I'm turning off unknown attributes, unimported elements (but not unknown elements), and css parse errors. Does that match your experience?

The things I've experienced are that Unknown Attributes and Unknown Tags call way too many things unknown. Also, in Polymer, Type Checking calls `hidden` and other things incorrect due to syntax, not that it's unexpected.

runem:
@jsilvermist I've heard the problems of too many unknowns a couple of times before. I think this is partly because of missing support for mixins & constructor property assignments. However, I have implemented support for both these features, and they will come with the next update. Another reason for many unknowns is using custom attributes to style nodes from the outside (eg using an attribute like 'selected' on a <button>).

Nevertheless, I'm going to make the default configuration less strict by disabling "unknown props", "unknown attrs" and "missing imports" in the default config. I'm also going to introduce a "strict: true" rule to the config. Hopefully this will improve the initial experience with this plugin :blush: 

@justinfagnani Great catch! I'll look into the bug when I get time :+1: 

@rictic I will also make sure to support CSS snippets.

docs/_guide/08-tools.md
```diff
@@ -52,11 +52,11 @@ There are a number of IDE plugins that may be useful when developing with lit-ht
 
 VSCode plugin
 
-* https://marketplace.visualstudio.com/items?itemName=bierner.lit-html
+* [https://marketplace.visualstudio.com/items?itemName=runem.lit-plugin](https://marketplace.visualstudio.com/items?itemName=runem.lit-plugin)
```
justinfagnani: Can you add a short description of what features this has?

Also, I think it should be "VS Code" above, with a space.

docs/_guide/08-tools.md
```diff
@@ -52,11 +52,11 @@ There are a number of IDE plugins that may be useful when developing with lit-ht
...
 
 TypeScript plugin (works with Sublime and Atom)
 
-* https://github.com/Microsoft/typescript-lit-html-plugin
+* [https://github.com/Microsoft/typescript-lit-html-plugin](https://github.com/Microsoft/typescript-lit-html-plugin)
```
justinfagnani: Replace with Rune's plugin?

arthurevans:
Sorry, I don't know why I didn't see this  PR when you filed it, @rictic. Do we want to make any notes about configuration (turning down some false positives) before we merge this? Or have those changes already been made to the plugin?

runem:
@arthurevans I expect the changes to the default configuration to go live within the next 2 weeks.

justinfagnani:
Making some notes and revisiting this when lit-plugin is updated makes sense to me. Thanks @runem!

docs/guide/08-tools.md
```diff
@@ -44,25 +44,29 @@ During the development phase, you might want the following tools:
...
-ESLint is recommended for linting lit-html code. The following ESLint plugin can be added to check for some common issues in lit-html templates:
+ESLint is recommended for linting lit-html code.  The following ESLint plugin can be added to check for some common issues in lit-html templates:
 
 * [https://github.com/43081j/eslint-plugin-lit](https://github.com/43081j/eslint-plugin-lit)
 
+Another alternative is to use the `lit-analyzer` CLI alongside ESLint to detect issues in your lit-html templates:
+
+* [https://github.com/runem/lit-analyzer/tree/master/packages/lit-analyzer](https://github.com/runem/lit-analyzer/tree/master/packages/lit-analyzer)
+
+`lit-analyzer` uses the same backend as the VS Code and TypeScript plugins listed in [IDE plugins](#ide-plugins).
```
justinfagnani, line 68: I might add that beyond what `eslint-plugin-lit` does, it type-checks templates.

PR #970, opened 2019-07-29
Add trusted types support to lit html

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG, with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## Adding Trusted Types to Polymer
[Polymer](https://www.polymer-project.org/) is one of the most popular frameworks for building frontend applications. It uses [Lit Element](https://lit-element.polymer-project.org/) and [Lit HTML](https://lit-html.polymer-project.org/) for rendering components.

**Lit Element** - manipulates dangerous DOM sink only in `updating-element.ts` [_(setAttribute, line 543)_](https://github.com/Polymer/lit-element/blob/2a1e58554985e88c0f29b866ce0887d6f0dec592/src/lib/updating-element.ts#L543). This place is however under control of user. This means that if user wants to set value for an attribute which can potentially cause XSS and he has enabled Trusted Types enforcement he can use Trusted Value instead of string. That means, that there are no changes required for this module.

**Lit HTML** -  Is a templating library, which converts the template to DOM. There are a few places which need to be changed to enable Trusted Types.
1) **Attribute values interpolated in the template are stringified during rendering** which forces the Trusted Type value to be stringified _(and later be assigned to DOM sink as string)_. This can be prevented, because we can check whether the value in the template is trusted _(without any concatenated values at the beginning or end)_. 
2) **When assigning the result of interpolation (string) to `innerHTML` of template element** which will cause an error if Trusted Types are enforced. There is no way user can convert this value to Trusted Type. Instead, we will create a policy for Lit HTML called `lit-html` and convert this value to Trusted Type internally. Users using polymer only need to add `lit-html` to the CSP header of Trusted Types.

There is also an option to use [`unsafeHTML`](https://github.com/Polymer/lit-html/blob/17be9a2f7a60ea2d87a54b77cd303f80bf59f507/src/directives/unsafe-html.ts#L37). This however expects users which have Trusted Types enforcement enabled to provide a Trusted Value. 
Lastly, [place which toggles an boolean attribute](https://github.com/Polymer/lit-html/blob/17be9a2f7a60ea2d87a54b77cd303f80bf59f507/src/lib/parts.ts#L377) which is safe, as boolean attributes can't cause xss and will never require Trusted Values.


src/lib/parts.ts
```diff
@@ -69,25 +70,32 @@ export class AttributeCommitter {
...
-          text += typeof v === 'string' ? v : String(v);
-        } else {
-          for (const t of v) {
-            text += typeof t === 'string' ? t : String(t);
+
+    if (l === 1 && strings[0] === '' && strings[1] === '' &&
+        isTrustedValue(this.parts[0].value)) {
+      // this means that attribute value is an interpolated trusted type and we
+      // don't want to stringify it
+      return this.parts[0].value;
```
rictic: Given that this returns, the rest of the method body doesn't need to be put into an else.

src/lib/parts.ts
```diff
@@ -69,25 +70,32 @@ export class AttributeCommitter {
...
-      const part = this.parts[i];
-      if (part !== undefined) {
-        const v = part.value;
-        if (isPrimitive(v) || !isIterable(v)) {
-          text += typeof v === 'string' ? v : String(v);
-        } else {
-          for (const t of v) {
-            text += typeof t === 'string' ? t : String(t);
+
+    if (l === 1 && strings[0] === '' && strings[1] === '' &&
```
rictic: This is pretty similar to https://github.com/Polymer/lit-html/pull/750/files#diff-63a9df26ecde05415a902b5b5dc5e266R136

I think this part won't be necessary once that lands.

src/lib/trusted-types.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Turns the value to trusted HTML. If the application uses Trusted Types the
+ * value is transformed into TrustedHTML, which can be assigned to execution
+ * sink. If the application doesn't use Trusted Types, the return value is the
+ * same as the argument.
+ */
+export function dangerouslyTurnToTrustedHTML(value: string): string {
+  const TrustedTypes = getTrustedTypes();
+  if (!policy && TrustedTypes !== undefined) {
+    policy = TrustedTypes.createPolicy('lit-html', rules);
```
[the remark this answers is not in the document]
Siegrift: It's a little win for tests, though. If I initialize it in the module (where it will rely on Trusted Types being available in window) I may have problems with mocking TT in tests.

src/lib/template-result.ts
```diff
@@ -19,6 +19,7 @@
 import {reparentNodes} from './dom.js';
 import {TemplateProcessor} from './template-processor.js';
 import {boundAttributeSuffix, lastAttributeNameRegex, marker, nodeMarker} from './template.js';
+import {dangerouslyTurnToTrustedHTML} from './trusted-types';
```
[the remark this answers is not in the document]
Siegrift: I can do that, but I don't see that much value added by it. You can find all the usages just by searching for `dangerouslyTurnToTrustedHTML` in codebase. On the other hand, if it is in a separate module you can easily use this function in other files... Also I think it's good to have the modules separated and let `template-result.ts` be only responsible for creating the markup.

src/lib/trusted-types.ts
```diff
@@ -0,0 +1,43 @@
+function getTrustedTypes() {
+  // tslint:disable-next-line
+  return (window as any).TrustedTypes;
```
koto: Nit: Can we have window.trustedTypes || window.TrustedTypes here? We're renaming the main entry point for the API soon (https://bugs.chromium.org/p/chromium/issues/detail?id=991944)

src/lib/template-result.ts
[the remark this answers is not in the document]
koto: +1 to rictic@, we should avoid exporting risky functions where possible. TT allows you to lock down the badness, we should use this opportunity to minimize the need for grepping to be secure.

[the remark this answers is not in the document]
Siegrift: I removed the trusted-types file and inlined the code instead.

src/lib/trusted-types.ts
Siegrift: Yeah.

src/lib/trusted-types.ts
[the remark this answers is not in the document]
Siegrift: Actually, I don't need this for polymer tests, because it uses a different stack than other frameworks where testing this was a bit harder. However it's good if we decide we want to make the policy name configurable (which for not isn't)

Siegrift:
@rictic Super excited now :) 
I have aligned the changes, is there anything else that I need to update?

src/lib/template-result.ts
```diff
@@ -20,6 +20,27 @@ import {reparentNodes} from './dom.js';
...
+
+/**
+ * Turns the value to trusted HTML. If the application uses Trusted Types the
+ * value is transformed into TrustedHTML, which can be assigned to execution
+ * sink. If the application doesn't use Trusted Types, the return value is the
+ * same as the argument.
+ */
+function dangerouslyTurnToTrustedHTML(value: string): string|TrustedHTML {
+  // tslint:disable-next-line
+  const TrustedTypes = (window as any).TrustedTypes as TrustedTypePolicyFactory;
```
rictic: Check trustedTypes, then TrustedTypes

src/lib/template-result.ts
```diff
@@ -20,6 +20,27 @@ import {reparentNodes} from './dom.js';
...
 
+let policy: Pick<TrustedTypePolicy, 'createHTML'>|undefined;
+
+/**
+ * Turns the value to trusted HTML. If the application uses Trusted Types the
+ * value is transformed into TrustedHTML, which can be assigned to execution
+ * sink. If the application doesn't use Trusted Types, the return value is the
+ * same as the argument.
+ */
+function dangerouslyTurnToTrustedHTML(value: string): string|TrustedHTML {
```
rictic: This function doesn't need to have such a scary name now that it's module-local.

Maybe something like`convertConstantTemplateStringToTrustedHTML` instead?

src/lib/template-result.ts
```diff
@@ -100,7 +121,7 @@ export class TemplateResult {
 
   getTemplateElement(): HTMLTemplateElement {
     const template = document.createElement('template');
-    template.innerHTML = this.getHTML();
+    template.innerHTML = dangerouslyTurnToTrustedHTML(this.getHTML()) as string;
```
rictic: The reason this is secure is because `this.strings` is a TemplateStringsArray. Add a TODO to use `Array.isTemplateObject` to validate `this.strings` if and when it is implemented, and maybe link to https://github.com/tc39/proposal-array-is-template-object

src/lib/template-result.ts
```diff
@@ -20,6 +20,28 @@ import {reparentNodes} from './dom.js';
...
+ * Turns the value to trusted HTML. If the application uses Trusted Types the
+ * value is transformed into TrustedHTML, which can be assigned to execution
+ * sink. If the application doesn't use Trusted Types, the return value is the
+ * same as the argument.
+ */
+function convertConstantTemplateStringToTrustedHTML(value: string): string|TrustedHTML {
+  // tslint:disable-next-line
+  const w = window as any
+  const TrustedTypes = (w.trustedTypes || w.TrustedTypes) as TrustedTypePolicyFactory;
+  if (TrustedTypes && !policy) {
```
justinfagnani: Can you move this outside the function too?

src/lib/template-result.ts
```diff
@@ -20,6 +20,28 @@ import {reparentNodes} from './dom.js';
...
+/**
+ * Turns the value to trusted HTML. If the application uses Trusted Types the
+ * value is transformed into TrustedHTML, which can be assigned to execution
+ * sink. If the application doesn't use Trusted Types, the return value is the
+ * same as the argument.
+ */
+function convertConstantTemplateStringToTrustedHTML(value: string): string|TrustedHTML {
+  // tslint:disable-next-line
+  const w = window as any
+  const TrustedTypes = (w.trustedTypes || w.TrustedTypes) as TrustedTypePolicyFactory;
```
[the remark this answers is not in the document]
rictic: `trustedTypes` is the new version of the spec, `TrustedTypes` is the old (current) version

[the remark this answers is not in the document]
rictic: Discussed in thread above, that change makes testing more difficult. Though if we're moving to lit-html testing to karma I think I've got something almost ready that will let us test with native trusted types

src/lib/template-result.ts
rictic: Same as above

src/lib/template-result.ts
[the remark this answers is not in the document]
Siegrift: Yeah, I the comment will be useful.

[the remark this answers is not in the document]
justinfagnani: Testing can be done by settings `window.trustedTypes` in a script before loading lit-html. We do this in other places.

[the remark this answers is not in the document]
Siegrift: @justinfagnani Thanks, I had troubles using the polyfill in typescript modules so I mocked the TT behaviour. I forgot about possibility to add a polyfill script before loading the tests. However, this also introduces undesired behavior that tests are no longer deterministic and dependant on the polyfill. If it changes or has a bug, unrelated trusted types tests will fail.

I've tried to implement this using the polyfill, however It [didn't pass the tests on CI in firefox](https://travis-ci.org/Polymer/lit-html/jobs/581632334). There seems to be some wierd interaction with Trusted Types and other scripts that produces Trusted Types violation for untrusted URLs, which I couldn't fix (even with default policy!). However trusted URLs are going to be removed from the polyfill soon ([they are already removed from spec](https://github.com/WICG/trusted-types/commit/d14d694440659cb071306e6ffc27eb7a99a3018f)). I think we can revisit this then. I have saved the work needed to use the polyfill on [other branch](https://github.com/Siegrift/lit-html/commit/2876898ccdcbdf1debd62fea4c9c5d6c95797c2c), so it should be a quick change.

In the future we would like to create an npm package which I think is the best approach. However, it's not yet ready :/

Siegrift:
@rictic @justinfagnani Can you please have a look? I think I resolved all of the comments :)

rictic:
Great work!

src/lib/template-result.ts
```diff
@@ -20,6 +20,28 @@ import {reparentNodes} from './dom.js';
...
+ * same as the argument.
+ */
+function convertConstantTemplateStringToTrustedHTML(value: string): string|
+    TrustedHTML {
+  // tslint:disable-next-line
+  const w = window as any
+  // TrustedTypes have been renamed to trustedTypes
+  // (https://github.com/WICG/trusted-types/issues/177)
+  const trustedTypes =
+      (w.trustedTypes || w.trustedTypes) as TrustedTypePolicyFactory;
```
TimvdLippe: I think one of these should have been `w.TrustedTypes`?

rictic: Yeah, fixed in #1003

PR #969, opened 2019-07-24
Replace TSLint with ESLint

abdonrd:
Fixes #967.

Ideally we should remove these current disabled rules:
```json
  "rules": {
    "@typescript-eslint/no-non-null-assertion": "off",
    "@typescript-eslint/no-use-before-define": "off",
    "@typescript-eslint/triple-slash-reference": "off"
  }
```

Some may be disabled in favor of a formatter ([Prettier](https://github.com/prettier/prettier)?):
```json
  "rules": {
    "no-unexpected-multiline": "off",
    "@typescript-eslint/indent": "off"
  }
```

PR #971, opened 2019-07-31
Add some missed types

abdonrd:
Warnings found by ESLint (#969).


justinfagnani:
Hmm... While I'm against implicit `any`, I'm generally ok with other forms of inference, including return type inference and _especially_ return type inference for callbacks like in:

```ts
value.every((v, i) => v === previousValue[i])
```

Having to write:

```ts
value.every((v, i): boolean => v === previousValue[i])
```

seems like unproductive ceremony to me.

abdonrd:
@justinfagnani I'm ok with that! Updated!

abdonrd:
Closed in favor of #969

PR #977, opened 2019-08-09
Hydrate pre-rendered `repeat` parts

straversi:
Build NodeParts from the contents of pre-rendered `repeat` parts.

Things to think about:
How can we hydrate the item _keys_? Is this possible? Without hydrating item keys, we lose out on efficiency while hydrating. For example, if the pre-rendered items equal the hydrated items, we still have to create new parts because the repeat directive uses the item keys to compare nodes.


justinfagnani:
After in-person discussion: We want to add an option to `hydrate()` that signals that we're using the same data that was used on the server so that we can assume items are in the correct order, and simply re-populate the stored keys from the key function.

Other options to explore are requiring serializable keys and including them in the NodePart marker comment.

We also need to emit the `repeat()` item markers correctly  on the server.

jridgewell:
Can't this be solved with https://github.com/Polymer/lit-html/pull/949?

straversi:
@jridgewell This PR would be merged into the #949 branch.

src/directives/async-append.ts
```diff
@@ -84,7 +84,7 @@ export const asyncAppend = directive(
         // Check to see if we have a previous item and Part
         if (itemPart !== undefined) {
           // Create a new node to separate the previous and next Parts
-          itemStartNode = createMarker();
+          itemStartNode = createStartMarker();
```
jridgewell: Don't we need to insert end markers, too?

src/directives/repeat.ts
```diff
@@ -104,6 +104,21 @@ export const repeat =
...
+                if (containerPart.options.prerenderedParts !== undefined) {
+                  partListCache.set(containerPart, []);
+                  for (const partInfo of containerPart.options.prerenderedParts!) {
+                    // Remove prerenderedParts from options, as they only apply to
+                    // the container part.
+                    const {prerenderedParts, ...options} = containerPart.options;
+                    const nodePart = new NodePart(options);
+                    nodePart.startNode = partInfo.startNode;
+                    nodePart.endNode = partInfo.endNode;
+                    partListCache.get(containerPart)!.push(nodePart);
```
jridgewell: Nit: map lookup is slow, we can cache this before inserting above.

src/directives/async-append.ts
straversi: True. This change was just to get the branch building; we should actually update this in a separate PR, or on the `hydration` branch.

src/directives/repeat.ts
```diff
@@ -104,6 +104,34 @@ export const repeat =
...
+                  partListCache.set(containerPart, nodeParts);
+
+                  // Hydrate the template keys only if the data is unchanged since
+                  // pre-rendering.
+                  if (!containerPart.options.dataChanged) {
+                    const keys = [];
+                    let index = 0;
+                    for (const item of items) {
+                      keys.push(keyFn ? keyFn(item, index) : index);
+                      index++;
```
straversi: Maybe combine these loops? We skip items.length checks of `!containerPart.options.dataChanged` by keeping them separate.

justinfagnani:
Closing since we have hydrate in lit-html 2.0

PR #980, opened 2019-08-14
Update to Rollup v1

abdonrd:
Update to Rollup v1


rollup.config.js
```diff
@@ -24,7 +24,6 @@ export default {
   plugins: [
     terser({
       warnings: true,
-      module: true,
```
jridgewell: Why this change?

abdonrd: Because it's the default now: [rollup-plugin-terser v4.0.0](https://github.com/TrySound/rollup-plugin-terser/releases/tag/v4.0.0)

>automatically set module: true for the esm format ([Tiger Oakes](https://github.com/NotWoods) in [#22](https://github.com/TrySound/rollup-plugin-terser/pull/22))

PR #992, opened 2019-08-27
Use loose boolean checks

jridgewell:
Saves some 43b. These are all guaranteed by types, any checks that could coerce 2+ values to false were left alone.

### DIFF
--- a/src/directives/async-append.ts
+++ b/src/directives/async-append.ts
@@ -58,7 +58,7 @@ export const asyncAppend = directive(
 
         // When we get the first value, clear the part. This lets the
         // previous value display until we can replace it.
-        if (i === 0) {
+        if (!i) {
           part.clear();
         }
 
@@ -66,7 +66,7 @@ export const asyncAppend = directive(
         // transforms of iterables and async iterables requires a library,
         // we accept a mapper function. This is especially convenient for
         // rendering a template for each item.
-        if (mapper !== undefined) {
+        if (mapper) {
           // This is safe because T must otherwise be treated as unknown by
           // the rest of the system.
           v = mapper(v, i) as T;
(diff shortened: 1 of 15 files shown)


AStaroverov:
I think you lose the readability of the code and do not get any gain

justinfagnani:
Doing strict comparisons and avoiding ToBoolean conversions has been part of our team style for a long time. Our understanding from the V8 team is that avoiding implicit ToBoolean conversions is always better, and can show up in traces of hot loops. While I haven't done traces that show an impact here, it's an easy style to adopt to just avoid any problems there wholesale.

I also find the explicit style more readable, but that may just be habit. Was the 43b post-gzip? I'll check the post-gzip size again, but I'd guess that the repetitive nature of the `=== undefined` checks compresses well.

Basically, while we have some benchmarks in place now, I know they don't coverer enough scenarios, so I'm a little wary to make changes that slightly educated instinct says may be slower, even if in this case I also know the effect is small in absolute terms (ToBoolean vs a DOM operation, say).

jridgewell:
> Our understanding from the V8 team is that avoiding implicit ToBoolean conversions is always better, and can show up in traces of hot loops.

That might have been true with the old Crankshaft JIT, but I've never noticed any real perf hits.

> Was the 43b post-gzip?

Yes, I was using the [`checksize`](https://github.com/Polymer/lit-html/blob/b71489705cec75e7e20639302d571a1a167475ea/package.json#L27) command.

justinfagnani:
Closing out a few lit-html 1.x PRs.

This one is interesting because we have more benchmarks now, and run on CI, so we could see a little better what the impact is (caveat, the resolution isn't _that_ great on small changes we still see a lot of "unsure" results). I have done more digging into bytecode though and am still wary of introducing more ToBoolean conversions, but we can experiment. Closing this particular PR for now though.

PR #996, opened 2019-08-30
Fix repeat object with classmap and stylemap

### LINKED ISSUE #972: classMap doesnt support mutation
# Description
classMap doesnt update class attributes when same object reference is used to render after first time.
# Steps to Reproduce

```js
const classes = { disabled : false, checked: true };
const template = () => html`<div class=${classMap(classes)}></div>`;

render(template(),document.body);
//outputHTML = <div class="checked"></div>

classes.disabled = true;

render(template(),document.body);
//outputHTML = <div class="checked"></div>

```
## Expected Output
```<div class="checked disabled"></div>```

ruphin:
Clone the info objects before storing them in the cache, so we can detect changes.

Fixes #972

### DIFF
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -71,5 +71,6 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
       classList[method](name);
     }
   }
-  classMapCache.set(part, classInfo);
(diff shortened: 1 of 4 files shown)


justinfagnani:
Thanks!

I see some other things in the directives I'd like to fix, but I'll do it on top of this.

PR #997, opened 2019-08-30
Class style map

justinfagnani:
Follow-on to #996

### DIFF
--- a/src/directives/style-map.ts
+++ b/src/directives/style-map.ts
@@ -22,7 +22,7 @@ export interface StyleInfo {
  * Stores the StyleInfo object applied to a given AttributePart.
  * Used to unset existing values when a new StyleInfo object is applied.
  */
-const styleMapCache = new WeakMap<AttributePart, StyleInfo>();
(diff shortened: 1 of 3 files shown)


src/directives/style-map.ts
```diff
@@ -52,15 +52,18 @@ export const styleMap = directive((styleInfo: StyleInfo) => (part: Part) => {
...
     style.cssText = committer.strings.join(' ');
+    previousStylePropertyCache.set(part, previousStyleProperties = new Set());
   }
 
   // Remove old properties that no longer exist in styleInfo
-  const oldInfo = styleMapCache.get(part);
-  for (const name in oldInfo) {
+  for (const name of previousStyleProperties) {
     if (!(name in styleInfo)) {
+      previousStyleProperties.delete(name);
```
jridgewell: I wonder if this will cause deopts in the iteration.

justinfagnani: `previousStyleProperties` is a Set, so no. Sets and Maps don't have shapes and are effectively always in "dictionary" mode.

PR #1000, opened 2019-09-03
Add unsafeSVG for rendering SVG snippets inside an SVG

### LINKED ISSUE #304: Insert another raw string to svg
I have this code, but it is not work
```js
import { svg } from 'lit-html'
import sealImage from '!!raw-loader!static/assets/signes/seal.svg'

const certificateSealTemplate = svg`<g id="certificateSeal" transform="matrix(0.24731808,0,0,0.24787261,50,30)">
  ${sealImage}
</g>`
```
I can't insert another svg file to svg as string
In the DOM tree I see this

image

How I can fix it?

Artur-:
Fixes #304


jridgewell:
It seems like we could change `unsafeHTML` to be aware of the enclosing namespace.

justinfagnani:
> It seems like we could change unsafeHTML to be aware of the enclosing namespace.

We could, but as DX I think this has a nice symmetry with the `svg` template tag.

src/directives/unsafe-svg.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const previousValue = previousValues.get(part);
+
+  if (previousValue !== undefined && isPrimitive(value) &&
+      value === previousValue.value && part.value === previousValue.fragment) {
+    return;
+  }
+
+  const template = document.createElement('template');
+  template.innerHTML = `<svg>${value as string}</svg>`;  // innerHTML casts to string internally
```
justinfagnani: The template literal is a string, so the comment is unnecessary. I think you can leave out the cast too.

src/directives/unsafe-svg.ts
```diff
@@ -0,0 +1,59 @@
...
+  const svgElement = template.content.firstElementChild!;
+  template.content.removeChild(svgElement);
```
justinfagnani: Instead of a custom loop, use `reparentNodes`. This is from https://github.com/Polymer/lit-html/blob/master/src/lib/template-result.ts#L122

```ts
    const content = template.content;
    const svgElement = content.firstChild!;
    content.removeChild(svgElement);
    reparentNodes(content, svgElement.firstChild);
```

src/test/directives/unsafe-svg_test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const produced = stripExpressionMarkers(container.innerHTML);
+    assert.equal(produced, '<svg><g>inner</g></svg>');
+    const svgElement = container.firstElementChild!;
+    assert.equal(svgElement.tagName, 'svg');
+    assert.equal(svgElement.namespaceURI, 'http://www.w3.org/2000/svg');
+    const gElement = svgElement.firstElementChild!;
+    assert.equal(gElement.tagName, 'g');
+    assert.equal(gElement.namespaceURI, 'http://www.w3.org/2000/svg');
+});
+test('renders SVG with multiple children', () => {
```
justinfagnani: nit: add newline above

Artur-: Done

src/directives/unsafe-svg.ts
Artur-: Done

src/directives/unsafe-svg.ts
Artur-: Done

src/directives/unsafe-svg.ts
```diff
@@ -0,0 +1,59 @@
...
+const previousValues = new WeakMap<NodePart, PreviousValue>();
+
+/**
+ * Renders the result as SVG, rather than text.
+ *
+ * Note, this is unsafe to use with any user-provided input that hasn't been
+ * sanitized or escaped, as it may lead to cross-site-scripting
+ * vulnerabilities.
+ */
+export const unsafeSVG = directive((value: unknown) => (part: Part): void => {
```
blikblum: Should not `value` be of type `string`?

Artur-:
Added a changelog entry

src/directives/unsafe-svg.ts
jridgewell: `unsafeHTML` uses `unknown` as well. As long as it coerces to a string, I think using `unknown` is fine.

AmoebeLabs:
Is this PR still waiting for a changelog entry, or something else??

petterek:
Any progress on this?  
I need this feature 😄 , solved it for now by copying the file locally, and for me it looks like it does what it should

kevinmpowell:
I'm also interested seeing this released. I've pointed straight to master in my test app and this is working exactly where I need it to. @justinfagnani any ideas when a 1.2.0 release might happen?

PR #1001, opened 2019-09-05
Avoid closures when using directive

jridgewell:
This avoids the closure produced when using a directive. In a basic example, directives were:

```typescript
(value: unknown) => ((part: Part) => void)
```

That inner closure to wrap over the value by the caller is **expensive**. Now, we generate a simple tuple to store the value and directive's inner function (🚲🏡 the naming here), which is much easier to generate and GC.

**But the [perf improvements](https://jsbench.github.io/#759b99dbded0441f69b3cd7722e34f3b)!** All those closures are killing us with GC time. Allocating a short tuple is a lot less expensive. If we realllly wanted, we could even do away with the `WeakMap` here and [really reap some benefits](https://jsbench.github.io/#4f3d62eb97bd1c992a5257d34fcebea0).

### DIFF
--- a/src/directives/async-append.ts
+++ b/src/directives/async-append.ts
@@ -31,71 +31,75 @@ import {createMarker, directive, NodePart, Part} from '../lit-html.js';
  * @param mapper An optional function that maps from (value, index) to another
  *     value. Useful for generating templates for each item in the iterable.
  */
-export const asyncAppend = directive(
+export const asyncAppend =
+    directive(
+        <T>(value: AsyncIterable<T>,
+            mapper?: (v: T, index?: number) => unknown) => ({value, mapper}),
+        async (part: Part, {value, mapper}) => {
+          if (!(part instanceof NodePart)) {
+            throw new Error('asyncAppend can only be used in text bindings');
+          }
+          // If we've already set up this particular iterable, we don't need
+          // to do anything.
+          if (value === part.value) {
+            return;
+          }
+          part.value = value;
+
+          // We keep track of item Parts across iterations, so that we can
+          // share marker nodes between consecutive Parts.
+          let itemPart;
+          let i = 0;
+
+          for await (let v of value) {
+            // Check to make sure that value is the still the current value of
+            // the part, and if not bail because a new value owns this part
+            if (part.value !== value) {
+              break;
+            }
+
+            // When we get the first value, clear the part. This lets the
+            // previous value display until we can replace it.
+            if (i === 0) {
+              part.clear();
+            }
+
+            // As a convenience, because functional-programming-style
+            // transforms of iterables and async iterables requires a library,
+            // we accept a mapper function. This is especially convenient for
+            // rendering a template for each item.
+            if (mapper !== undefined) {
+              // This is safe because T must otherwise be treated as unknown by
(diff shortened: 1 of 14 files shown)


LarsDenBakker:
Impressive improvements :)

justinfagnani:
Thanks Justin! A few things before a detailed review.

First, this is a breaking change. You've updated the built-in directives, but third-party directives will need to be updated. I'm not sure we're ready for a 2.0 yet.

We could vend a `fastDirective` function in parallel with `directive`. Or skip the factory function and go to an object/class-based approach. I've been wanting to do this for a few reasons, including better TemplateResult introspection (I'd like the state captured by a directive invocation, and a key for the directive, to be available to tools and not locked in a closure), and for SSR (a directive object could have an optional initial render callback that returns a TemplateResult rather than imperatively mutates DOM).

Second, I'd like to see the perf benefits show up in our benchmarks. Either we should skip the WeakMap in this PR, and/or we need better benchmarks. Well, we definitely need better benchmarks, since the Shack benchmark doesn't use directives at all.

I'd love to start with a directive-using micro-benchmark, where we create and use a nearly no-op directive (it just calls `part.setValue()` with its parameter) to measure directive overhead relative to some simple rendering. From the second micro-benchmark you have it looks like the overhead is 1µs with closures vs 1ns with objects on my 2015 MBP, and 1µs vs. I'm not sure that difference will show up even with .05µs on my Pixel 3. It's a significant improvement in relative terms, but in absolute terms on real pages I'm not sure how much this will show up.

It's possible that the micro-benchmark isn't showing GC effects, which could show up in a real template, but I'm not sure why the VM would have a harder time collecting closures than objects. Do you have some insight there?

justinfagnani:
I changed the benchmark to use a closure in both variants, but set a property on the function instead of using a WeakMap. That still shows large improvements, and may be worthwhile without a breaking change: https://jsbench.github.io/#dc1a50def64994d4da5247386708b42c

jridgewell:
>  I'm not sure we're ready for a 2.0 yet.

I've got more ideas that might make a v2 sooner than later. 😉

> From the second micro-benchmark you have it looks like the overhead is 1µs with closures vs 1ns with objects on my 2015 MBP, and 1µs vs. I'm not sure that difference will show up even with .05µs on my Pixel 3.

I don't think GC times are accurately captured in the benchmarks, though, so the original benchmark isn't perfectly captured in the numbers. Looking at the updates of the "ops/sec", it feels much snappier without the closures. I think pause times alone are easily dropping frames.

> That still shows large improvements, and may be worthwhile without a breaking change:

Made https://jsbench.github.io/#18fc6d1eb87a9e3edbd7295edf9b8f35 to compare the 3 approaches. We can definitely get rid of the `WeakMap` as an improvement in v1, and wait for v2 before making the more drastic changes.

justinfagnani:
I have a similar comparison here: https://jsbench.github.io/#dc1a50def64994d4da5247386708b42c

I added an accumulator to make sure the adder wasn't being optimized away.

justinfagnani:
lit-html 2.0 introduced DirectiveResult, which is a simple object that captures values and a pointer to the directive class, eliminating the closure allocations.

PR #1002, opened 2019-09-07
Use TemplateStringsArray.raw strings to render

jridgewell:
This is a breaking change.

Now, any JS escape chars will be printed as if they were literal text characters. Eg, `\n` won't print a newline, it'll print "\\" and "n". If you want to print escape chars, use HTML's percent escapes or pass a string inside a dynamic Part.

Fixes https://github.com/Polymer/lit-html/issues/348.

### DIFF
--- a/src/lib/shady-render.ts
+++ b/src/lib/shady-render.ts
@@ -61,7 +61,7 @@ const shadyTemplateFactory = (scopeName: string) =>
       let templateCache = templateCaches.get(cacheKey);
       if (templateCache === undefined) {
         templateCache = {
-          stringsArray: new WeakMap<TemplateStringsArray, Template>(),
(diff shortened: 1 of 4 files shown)


frank-dspeed:
i am for handling that via a additional export that warps the original html it is as easy as

```js
const returnRaw = (strArr, ...valArr) => [strArr.raw,...valArr]
html(...returnRaw)
```
this is how i handle raw and literal string arrays in the same function and choose what to pass
i implamented some other helpers like curried api and such flavors but that is the basic implamentation

justinfagnani:
We definitely can't afford a breaking change like this right now, and from user feedback on Polymer 3's `html` tag, we actually moved from raw strings to cooked. I could see another tag being exported for raw strings like @frank-dspeed mentions.

PR #1003, opened 2019-09-10
Fix typo in trusted types

rictic:
We want to check for either trustedTypes or TrustedTypes here, until the capitalized TrustedTypes implementation is gone


rictic:
Looks like something is up with travis launching Chrome, but I think unrelated to this change

PR #1016, opened 2019-10-09
Use Typescript `is` to remove casting in `Template`

tjenkinson:


justinfagnani:
Thanks for the PR. Unfortunately, a cast is free, but custom type guards have an implementation that gets shipped to the client. The decision to limit helpers in this case was deliberate, and without an optimizing compiler in our toolchain I wouldn't want to take on the bytes here.

justinfagnani:
> Would be neat if typescript could get smart enough to inline simple cases like this in the future.

Definitely

PR #1018, opened 2019-10-19
Draft: Add animation directives

LarsDenBakker:
As discussed in https://github.com/Polymer/lit-html/issues/1017, these are three directives I've been working on.

I've focused on doing the animations in CSS, with the directives just there to manage/communicate the state. This can be expanded on further using web animations from javascript. I'm not sure how well supported web animations are, I don't hear much about it.

**trackAnimation / trackTransition**
The `trackAnimation` and `trackTransition` directives are pretty straightforward. You apply them on a boolean attribute and provide a list of animations/transitions to track. The attribute it set/removed based on whether there is an active animation or transition:

```js
html`
  <div ?animating=${trackAnimation('show-foo', 'hide-foo')}></div>
`;
```

This state can then be used in CSS/JS.

**animate**
Animating to/from display: none, or elements that are added/removed to/from the DOM is quite complex and impossible to do from CSS alone. The `animate` directive is intended to make this easier.

The `animate` directive should be used in a NodePart and renders whatever value is passed to it. When the value is a `TemplateResult`, it will apply a class to the first rendered child element and waits for an animation on that element to be finished.

It maintains two child parts, alternating between them as values passed to the part changes. This way it can animate the content of the previous part out while the content of the new part is being animated in.

For example to animate a float actioning button as it enters/exits:
```css
@keyframes fab-entry {
  from { transform: scale(0) }
}

@keyframes fab-exit {
  to { transform: scale(0) }
}

.fab.fab-entry {
  animation: fab-entry 300ms;
}

.fab.fab-exit {
  animation: fab-exit 300ms;
}
```

```js
html`${animate(showFab ? html`<div class="fab"></div>` : nothing, { entry: 'fab-entry', exit: 'fab-exit' })}`
```

When the element is added, it adds a `fab-entry` class until all animations on the element finish. When the element needs to be removed, it adds a `fab-exit` class and removes the element after the animation finishes.

The directive can work with any amount of templates, for example you can use it in an application router to animate page transitions:

```js
html`${animate(this._renderPage(), { entry: 'page-entry', exit: 'page-exit' })}`
```

I didn't write any tests yet, as I'd like to iron out the API first. I did add a demo to demonstrate how it works.

### DIFF
--- a/demo/animate/index.html
+++ b/demo/animate/index.html
@@ -118,10 +118,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
           <button @click=${() => renderDemo(!showFab, currentPage)}>Toggle fab</button>
 
-          ${animate(showFab ? html`<div class="fab"></div>` : nothing, { prefix: 'fab' })}
+          ${animate(showFab ? html`<div class="fab"></div>` : nothing, { entry: 'fab-entry', exit: 'fab-exit' })}
         </div>
 
-        ${animate(currentPage, { prefix: 'page' })}
+        ${animate(currentPage, { entry: 'page-entry', exit: 'page-exit' })}
 
       `, document.getElementById('demo'));
     }
--- a/demo/animate/index.html
+++ b/demo/animate/index.html
@@ -1,0 +1,133 @@
+<!--
+@license
+Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<!doctype html>
+<html>
+
+<head>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
+
+  <style>
+    body {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      margin: 0;
+    }
+
+    * {
+      box-sizing: border-box;
+    }
+
+    #demo {
+      width: 100%;
+      max-width: 800px;
+    }
+
+    .buttons {
+      padding: 36px 0;
+    }
+
+    .page {
+      padding: 16px;
+      background-color: white;
+      box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
+      border-radius: 4px;
+      max-width: 800px;
+    }
+
+    h2 {
+      margin-top: 0;
+    }
+
+    @keyframes page-entry {
+      from {
+        transform: translateX(100vw)
+      }
+    }
+
+    @keyframes page-exit {
+      to {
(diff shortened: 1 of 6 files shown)


LarsDenBakker:
After testing, I ran into some subtle race condition and some issues on browsers on the shady dom polyfill. I relied on passing along an animation name to the directive, but in shady dom animation names are rewritten in order to be able to globally scope them.

I've reworked the approach now to no longer rely on a hardcoded animation name, instead dynamically tracking any events which are started. It's a bit more code, but I'm actually quite happy with it. It makes the API a lot cleaner and easier to manage for a developer.

I've updated the code samples too.

motss:
This looks exciting. I'm thinking if it is possible to add some kind of hooks to the end of an animation or transition? So that it allows devs can schedule moar tasks when an animation or a transition finishes which I think that this is a pretty useful feature.

LarsDenBakker:
@motss how would you envision the API for that to look?

blikblum:
> `html`${animate(this._renderPage(), { entry: 'page-entry', exit: 'page-exit' })}`

How about using enter/leave instead of entry/exit?

LarsDenBakker:
> > `html`${animate(this._renderPage(), { entry: 'page-entry', exit: 'page-exit' })}`
> 
> How about using enter/leave instead of entry/exit?

I went with the terminology of https://github.com/PolymerElements/neon-animation, completely open to changing the name though.

stramel:
Having been working in React for a while now, I have seen enter/leave, start/end, transition/transitionEnd. Just wanted to add options I have seen from outside the webcomponent realm.

LarsDenBakker:
@stramel thanks for the links. Those libraries do a lot of CSS-in-JS which isn't a goal for this PR, and probably shouldn't land in the main lit-html library. 

This is just helpers to manage animation states, where the animations themselves are defined in CSS.

blikblum:
For the record vue and angularjs uses enter/leave

https://vuejs.org/v2/guide/transitions.html
https://docs.angularjs.org/api/ngAnimate

BTW: the vue page has some simple and nice demos. Would be nice this abstraction could support those use cases (they are based on CSS)

LarsDenBakker:
Thanks, I will take a look at those

jsilvermist:
> For the record vue and angularjs uses enter/leave

Also matches with native events such as `mouseenter` and `mouseleave`.

stramel:
@LarsDenBakker Thanks for taking a look at my comment. 

> Those libraries do a lot of CSS-in-JS which isn't a goal for this PR, and probably shouldn't land in the main lit-html library.
> This is just helpers to manage animation states, where the animations themselves are defined in CSS.

I understand a lot of it is CSS-in-JS and not really what we're going for. I find it useful when defining an API to look for inspiration in other libraries/frameworks. Just wanted to chime in with stuff I had seen in React world. 😄

justinfagnani:
I've been doing some animations with lit-html and CSS transitions recently, so I think have a little better perspective to check this out with now. I still haven't used CSS animations much though. @LarsDenBakker would you be able to just on a chat to walk me through a few things?

frank-dspeed:
Hi i am author of a lit-html-framework i don't want to call the name out here

## Solution
Don't use directives for that!!! use customElement api for that <my-animated-element prop="val"></my-animated-element> put all logic inside that element.
- The Template will work in SSR Scenarios and webworkers

justinfagnani:
We're adding a `flip()` directive to @lit/labs soon!

PR #1019, opened 2019-10-22
umd

taoqf:
https://github.com/Polymer/lit-element/issues/140


taoqf:
@googlebot I signed it!
Actually, I could not signed because I am chinese.

justinfagnani:
Closing for same reasons as: https://github.com/Polymer/lit-element/pull/840#issuecomment-548985884

PR #1020, opened 2019-10-27
Add null check for isCEPolyfill

### LINKED ISSUE #963: Importing from webextension contentscript results in error in chrome
window.customElements = null in webextension contentscripts context.
This results in an error due to the narrow check in /src/lib/dom.ts:26 checking only for `window.customElements !== undefined`, missing the null case.

# Steps to reproduce
1. Ensure at least one webextension with contentscripts is installed in chrome
1. Open a chrome tab to example.org
2. Open DevTools on the console page(Use ctrl+shift+i to open)
3. Select the webextension from the dropdown list to change context
4. Run `import("https://unpkg.com/lit-html@1.1.1/lit-html.js?module")`

Note: The behavior is identical if used directly in contentScript or using a userscript webextension.
Note: This behavior cannot be worked around by attempting to set the customElements value directly or using Object.defineProperty.

# Output
```
dom.js?module:18 Uncaught (in promise) TypeError: Cannot read property 'polyfillWrapFlushCallback' of null
    at dom.js?module:18
```

# Affected browsers
- [X] Chrome (relevant background: https://bugs.chromium.org/p/chromium/issues/detail?id=390807)
- [ ] firefox (has the proper interface for window.customElements in place)
No other browsers were tested.

Note: Filed https://bugs.chromium.org/p/chromium/issues/detail?id=985373 to see if this can be fixed browser side.

andynaguyen:
Fixes #963

Checks if window.customElements is null as part of isCEPolyfill function.


andynaguyen:
@googlebot I signed it!

src/lib/dom.ts
```diff
@@ -24,6 +24,7 @@ interface MaybePolyfilledCe extends CustomElementRegistry {
  * True if the custom elements polyfill is in use.
  */
 export const isCEPolyfill = window.customElements !== undefined &&
+    window.customElements !== null &&
```
jridgewell: We can change this to a single check:

```js
export const isCEPolyfill = window.customElements != null &&
```

This'll check for both `null` and `undefined`, since they are loosely equal to each other.

andynaguyen: True, I can make that change.

ngryman:
👋I'm facing the same issue and and it's a big blocker for me. Do you could merge it soon? Thanks!

andynaguyen:
@justinfagnani @jridgewell can one of you merge in the PR? I do not have write access

PR #1028, opened 2019-11-11
fixed #1027, styleMap inside vue and angular

### LINKED ISSUE #1027: styleMap variables doesn't work inside angular / vue applications
Vue / Angular are throwing errors when you want to use `lit-html` inside.

To reproduce just insert `lit-html` component with `styleMap` object that contains some variables like 

```javascript
styleMap( { '--variable': '10px' } )
```

Those frameworks prevents  setting style indexed properties with `element.style['--variable']` 

Error message:
`Uncaught (in promise) TypeError: Failed to set an indexed property on 'CSSStyleDeclaration': Index property setter is not supported.`

Related code: https://github.com/Polymer/lit-html/blob/master/src/directives/style-map.ts#L83

neuronetio:


neuronetio:
it was my bug, sorry

PR #1029, opened 2019-11-14
[performance] clone node instead of each time creating new one

neuronetio:
`element.cloneNode` is faster than `document.createElement` 
and can be implemented without any donwsides

https://www.measurethat.net/Benchmarks/Show/265/0/createelement-vs-clonenode
https://github.com/sophiebits/innerhtml-vs-createelement-vs-clonenode


neuronetio:
@googlebot I signed it!

stramel:
@jridgewell I'm curious to see your opinion on this potential enhancement. Unfortunately, we can't see the tachometer results of this change.

stramel:
Ping @justinfagnani

justinfagnani:
Thanks for this old PR. Sorry it wasn't reviewed years ago.

On latest Chrome, when directly measuring `<template>` elements, `createElement()` is actually faster.

https://www.measurethat.net/Benchmarks/Show/33556/0/createelement-vs-clonenode-for-template

Even if `cloneNode()` were a tad faster, this code is only run once per template definition, and creating the initial template DOM is going to dwarf the cost here.

PR #1032, opened 2019-11-21
Fix the build

rictic:
First thing I'm trying, don't run trusted type tests in ancient browsers.


justinfagnani:
I'm going to try some things on this branch too

PR #1041, opened 2019-11-27
Add link to LitElement

### LINKED ISSUE #1040: Add link to LitElement
Sometimes people end up on the lit-html site when they're looking for LitElement, and vice versa. Would be nice to add a disambiguation note early on and link to the LitElement site. Something like:

**lit-html and LitElement.** lit-html isn't tied to any particular component model. If you want to build web components, check out [LitElement](https://lit-element.polymer-project.org), a library for building web components using lit-html templating.

arthurevans:
Fixes #1040.


docs/guide.md
```diff
@@ -14,6 +14,12 @@ lit-html is a simple, modern, safe, small and fast HTML templating library for J
 
 lit-html lets you write HTML templates in JavaScript using [template literals] with embedded JavaScript expressions. Behind the scenes lit-html creates HTML `<template>` elements from your JavaScript templates and processes them so that it knows exactly where to insert and update the values from expressions.
 
+<div class="alert alert-info">
+  
+**lit-html and LitElement.** lit-html is not tied to any component model. If you want to build web components, check out [LitElement](https://lit-element.polymer-project.org/), a library for building web components using lit-html templates.
```
justinfagnani: How about expanding the first sentence a little, like:

lit-html is not tied to any component model, it focuses only on creating and updating DOM (or HTML?).

I'd also consider changing the second sentence to start "If you want to build components, ..."

arthurevans: Sounds good. I think maybe DOM is perhaps clearer. ("Updating HTML" sounds a little strange.)

Looking at this page, I realize the first paragraph is dipping into implementation detail that's probably unhelpful: 

"Behind the scenes lit-html creates HTML `<template>` elements from your JavaScript templates and processes them so that it knows exactly where to insert and update the values from expressions."

Should probably focus on end-user value here, like:

"lit-html identifies the static and dynamic parts of your templates so it can efficiently update the changed portions."

Or something.

justinfagnani: Yeah, that's a good point. Far too early to get into implementation details.

arthurevans: Cool. Updated. Staged here:

https://lit-element-note-dot-polymer-lit-html.appspot.com/guide

justinfagnani: The text reads great, but the alert styling seems a bit jarring to me. wdyt?

arthurevans: Like this? 
no-note-formatting

justinfagnani: I think that's better than the more intense alert styling. I do see that this is kind of an aside, and we might want to call that out some how, but for now I prefer this. You?

arthurevans: Yep. The blue "info" box is the mellowest alert/aside style we have. So I guess we'll go with this for now. 

This is kind of an aside, but as edited here, it follows the previous paragraph logically. So I'm OK with it like this.

PR #1043, opened 2019-12-02
Make aside less alarming

arthurevans:
Follow-on to #1041, because I somehow missed a commit.


arthurevans:
Merging based on prior approval for what was supposed to be in #1041 ... 😂

PR #1050, opened 2019-12-17
Load in nodejs for packaging treeshake able

frank-dspeed:
I made some adjustments to make the lib loadable via import inside nodejs current as a module so we can use it and inhire from it.

- refactored checks to be conditional so the module is still loadable in nodejs for processing in advanced packaging scenarios 

## Example
This example can be build via rollup and you will get the right html version 

```
//lit-html-loader.mjs
import { html as litHtml } from 'lit-html'
import { html as strHtml } from 'str-html'
import { isNode, isBrowser, isJsdom } from './check-env-helper.mjs'
export const html = isNode ? strHtml : litHtml;

// index.mjs
import { html } from './lit-html-loader.mjs'
```

this closes: #1051

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -23,7 +23,8 @@ interface MaybePolyfilledCe extends CustomElementRegistry {
 /**
  * True if the custom elements polyfill is in use.
  */
-export const isCEPolyfill = typeof window !== 'undefined' && window.customElements != null &&
+export const isCEPolyfill = typeof window !== 'undefined' &&
+    window.customElements != null &&
     (window.customElements as MaybePolyfilledCe).polyfillWrapFlushCallback !==
         undefined;
 
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -23,7 +23,7 @@ interface MaybePolyfilledCe extends CustomElementRegistry {
 /**
  * True if the custom elements polyfill is in use.
  */
-export const isCEPolyfill = window.customElements != null &&
+export const isCEPolyfill = typeof window !== 'undefined' && window.customElements != null &&
     (window.customElements as MaybePolyfilledCe).polyfillWrapFlushCallback !==
         undefined;
 
--- a/src/lit-html.ts
+++ b/src/lit-html.ts
@@ -56,7 +56,7 @@ declare global {
 // IMPORTANT: do not change the property name or the assignment expression.
 // This line will be used in regexes to search for lit-html usage.
 // TODO(justinfagnani): inject version number at build time
-(window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2');
+if (typeof window !== 'undefined') { (window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2'); }
 
 /**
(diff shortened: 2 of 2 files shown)


frank-dspeed:
This is now fully tested and in production in my fork and it passes all tests please review @justinfagnani

src/lit-html.ts
```diff
@@ -56,7 +56,11 @@ declare global {
 // IMPORTANT: do not change the property name or the assignment expression.
 // This line will be used in regexes to search for lit-html usage.
 // TODO(justinfagnani): inject version number at build time
-(window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2');
+const isBrowser = typeof window !== 'undefined'
+if (isBrowser) {
+  // If we run in the browser set version
+  (window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2');
```
justinfagnani: Rather than the `isBrowser` variable, let's just inline the check:

```suggestion
window ?? (window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2');
```

src/lib/dom.ts
```diff
@@ -23,9 +23,11 @@ interface MaybePolyfilledCe extends CustomElementRegistry {
 /**
  * True if the custom elements polyfill is in use.
  */
-export const isCEPolyfill = window.customElements != null &&
-    (window.customElements as MaybePolyfilledCe).polyfillWrapFlushCallback !==
-        undefined;
+export const isCEPolyfill = typeof window !== 'undefined' ?
```
justinfagnani: Since the false condition returns `false`, just use an and condition:

```suggestion
export const isCEPolyfill = typeof window !== 'undefined' &&
```

src/lit-html.ts
frank-dspeed: typeof window !== 'undefined' is importent window ?? will not work as we then try to access window needs verification!

frank-dspeed: (node:291371) ExperimentalWarning: The ESM module loader is experimental.
file:///home/frank/docker/apps/stealify/es/lit-html/lit-html-browser.mjs:49
window ?? (window['litHtmlVersions'] || (window['litHtmlVersions'] = [])).push('1.1.2');
        ^

SyntaxError: Unexpected token '?'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:84:18)
    at async link (internal/modules/esm/module_job.js:36:21)

src/lib/dom.ts
frank-dspeed: lgtm

src/lit-html.ts
justinfagnani: Right, we do need the `typeof`. `??` should be compiled out by TypeScript with the `es2017` target, so I'm not sure why you would be getting a Node error.

justinfagnani: I'd still rather not use a ternary when the false branch returns `false`. It's redundant.

justinfagnani: I see, you did that. Thanks.

justinfagnani:
This is fail the format check. Please run `npm run format`

justinfagnani:
Thanks!

PR #1055, opened 2020-01-10
Fix tests

justinfagnani:
An update to ShadyCSS made it unhappy with deleting window.ShadowRoot


justinfagnani:
Since this is fixing a broken master, and the ultimate fix was quite simple, I'm going to merge TBR.

PR #1056, opened 2020-01-11
Add `unsafeSVG` directive

### LINKED ISSUE #304: Insert another raw string to svg

justinfagnani:
Fixes #304

### DIFF
--- a/src/directives/unsafe-svg.ts
+++ b/src/directives/unsafe-svg.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * Copyright (c) 2020 The Polymer Project Authors. All rights reserved.
  * This code may only be used under the BSD style license found at
  * http://polymer.github.io/LICENSE.txt
  * The complete set of authors may be found at
@@ -28,7 +28,7 @@ interface PreviousValue {
 const previousValues = new WeakMap<NodePart, PreviousValue>();
 
 /**
- * Renders the result as HTML, rather than text.
+ * Renders the result as SVG, rather than text.
  *
  * Note, this is unsafe to use with any user-provided input that hasn't been
  * sanitized or escaped, as it may lead to cross-site-scripting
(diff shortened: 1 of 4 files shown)


justinfagnani:
@sorvell this is ready for review. The test failures are just HTML expectation differences. I think I fixed them, though it'd be nice to have a more robust matcher. I tried `@open-wc`'s semantic-dom-diff library, but it didn't support SVG.

src/directives/unsafe-svg.ts
```diff
@@ -0,0 +1,58 @@
...
+import {directive, NodePart, Part, reparentNodes} from '../lit-html.js';
+
+interface PreviousValue {
+  readonly value: unknown;
+  readonly fragment: DocumentFragment;
+}
+
+// For each part, remember the value that was last rendered to the part by the
+// unsafeSVG directive, and the DocumentFragment that was last set as a value.
+// The DocumentFragment is used as a unique key to check if the last value
```
sorvell: It's not clear why the document fragment is used. In what case does is it useful?

src/test/directives/unsafe-svg_test.ts
```diff
@@ -0,0 +1,123 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
```
sorvell: 2020

src/directives/unsafe-svg.ts
justinfagnani: See the 'renders after other values' test. If you render an `unsafeSVG()`, then any other value, then `unsafeSVG()` again with the same parameter, we need to re-render to the part, even though the parameter is the same. Checking that the current value is the right DocumentFragment as part of the dirty-check allows this.

justinfagnani:
Oops, I forgot about #1000. Good news, the implementation is exactly the same. Closing in favor of that.

PR #1057, opened 2020-01-11
Add live() directive

### LINKED ISSUE #877: Values changed outside of lit-html rendering are not updated from lit-html
Since lit-html stores previously rendered values for dirty-checking, any value changed from outside lit-html will not be overwritten in another render if the value passed to lit-html is the same.

See this live example: https://stackblitz.com/edit/lit-html-rt4n6e

This is actually working as intended, as it's intended that lit-html maintains control over bound values. But this issue has come up a few times in properties that are affected by user actions:

- `HTMLInputElement.value`
- `Element.scrollTop`

Currently the recommended solution would be to add event listeners which update the bound value so that they're in sync. That's non-obvious though, and we should handle this automatically if we can.

One option is to stop storing the previous value in `AttributePart`, `PropertyPart`, etc., and instead use the current bound attribute/property value instead as the source of truth.

There are two potential downsides:
- It might be slower, especially for native DOM properties. Crossing DOM bindings just to see if a value should be updated might slow things down.
- Values may be modified by the setter. For instance, `e.className = ['a', 'b']` results in a className of `a,b`. This would defeat the dirty check and cause it to be set every render.

Another solution is to vend a directive that defers to the underlying property specifically for these cases:

```js
import {live} from 'lit-html/directives/live.js';
// ...
html`<input .value=${live(value)}`>`;
// or
html`<div .scrollTop=${live(top)}`></div>`;
```

justinfagnani:
Fixes #877

This directive dirty-checks new values against the live DOM value rather than the previously rendered value stored by the lit-html Part. `live()` only works with attribute, property, and boolean attribute bindings.

Event bindings use addEventListener and don't have a single "live" value, so the directive doesn't make sense there. It's possible that a lit-html bound event listener is removed externally, but that seems like a very, very corner case.

I tried to make this work with NodeParts in case the text value is changed externally, like with content editable, but in those cases the part markers would likely be disturbed, making that case difficult to support. That might work with a binding to `.textContent`.

cc @LarsDenBakker who wrote a `live()` directive for open-wc.

### DIFF
--- a/src/directives/live.ts
+++ b/src/directives/live.ts
@@ -1,0 +1,73 @@
+/**
+ * @license
+ * Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {AttributePart, BooleanAttributePart, directive, EventPart, NodePart, PropertyPart} from '../lit-html.js';
+
+/**
+ * Checks binding values against live DOM values, instead of previously bound
+ * values, when determining whether to update the value.
+ *
+ * This is useful for cases where the DOM value may change from outside of
+ * lit-html, such as with a binding to an `<input>` element's `value` property,
+ * a content editable elements text, or to a custom element that changes it's
+ * own properties or attributes.
+ *
+ * In these cases if the DOM value changes, but the value set through lit-html
+ * bindings hasn't, lit-html won't know to update the DOM value and will leave
+ * it alone. If this is not what you want - if you want to overwrite the DOM
+ * value with the bound value no matter what, then use the `live()` directive:
+ *
+ *     html`<input .value=${live(x)}>`
+ *
+ * `live()` performs a strict equality check agains the live DOM value, and if
+ * the new value is equal to the live value, does nothing. This means that
+ * `live()` should not be used when the binding will cause a type conversion. If
+ * you use `live()` with an attribute binding, make sure that only strings are
+ * passed in, or the binding will update every render.
+ */
+export const live = directive(
+    (value: unknown) => (part: NodePart|AttributePart|PropertyPart|
+                         BooleanAttributePart) => {
+      let previousValue: unknown;
+      if (part instanceof EventPart || part instanceof NodePart) {
+        throw new Error(
+            'The `live` directive is not allowed on text or event bindings');
+      }
+      if (part instanceof BooleanAttributePart) {
+        checkStrings(part.strings);
+        previousValue = part.element.hasAttribute(part.name);
+        // This is a hack needed because BooleanAttributePart doesn't have a
+        // committer and does its own dirty checking after directives
+        part.value = previousValue;
+      } else {
+        const {element, name, strings} = part.committer;
+        checkStrings(strings);
+        if (part instanceof PropertyPart) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          previousValue = (element as any)[name];
+        } else if (part instanceof AttributePart) {
+          previousValue = element.getAttribute(name);
+        }
+        if (previousValue === value) {
+          return;
+        }
+      }
+      part.setValue(value);
+    });
+
+const checkStrings = (strings: readonly string[]) => {
+  if (strings.length !== 2 || strings[0] !== '' || strings[1] !== '') {
+    throw new Error('`live` bindings can only contain a single expression');
+  }
+};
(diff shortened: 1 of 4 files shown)


src/directives/live.ts
```diff
@@ -0,0 +1,73 @@
...
+      if (part instanceof EventPart || part instanceof NodePart) {
+        throw new Error(
+            'The `live` directive is not allowed on text or event bindings');
+      }
+      if (part instanceof BooleanAttributePart) {
+        checkStrings(part.strings);
+        previousValue = part.element.hasAttribute(part.name);
+        // This is a hack needed because BooleanAttributePart doesn't have a
+        // committer and does its own dirty checking after directives
+        part.value = previousValue;
```
jridgewell: What would happen if this were left out?

src/directives/live.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @license
+ * Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
```
sorvell: 2020

src/directives/live.ts
```diff
@@ -0,0 +1,73 @@
...
+ *     html`<input .value=${live(x)}>`
+ *
+ * `live()` performs a strict equality check agains the live DOM value, and if
+ * the new value is equal to the live value, does nothing. This means that
+ * `live()` should not be used when the binding will cause a type conversion. If
+ * you use `live()` with an attribute binding, make sure that only strings are
+ * passed in, or the binding will update every render.
+ */
+export const live = directive(
+    (value: unknown) => (part: NodePart|AttributePart|PropertyPart|
```
sorvell: Disallow `NodePart` here?

src/directives/live.ts
```diff
@@ -0,0 +1,73 @@
...
+      } else {
+        const {element, name, strings} = part.committer;
+        checkStrings(strings);
+        if (part instanceof PropertyPart) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          previousValue = (element as any)[name];
+        } else if (part instanceof AttributePart) {
+          previousValue = element.getAttribute(name);
+        }
+        if (previousValue === value) {
```
sorvell: Consider coercing `value` to string here?

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
+/**
+ * @license
+ * Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
```
sorvell: 2020

src/directives/live.ts
justinfagnani: The binding wouldn't update if the attribute were changed outside lit-html. I could probably implement the other parts this way, but what's here is working an small, so 🤷‍♂️

justinfagnani:
Tests should be fixed now

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+suite('live', () => {
+  let container: HTMLDivElement;
+
+  setup(() => {
+    container = document.createElement('div');
+  });
+
+  suite('properties', () => {
+    test('updates an externally set property', () => {
```
sorvell: Might be nice to validate what happens without `live`, at the least to aid readability.

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
...
+      const go = (x: string) =>
+          render(html`<div x="${live(x)}"></div>}`, container);
+      go('a');
+      const el = container.firstElementChild as LiveTester;
+      assert.equal(el.getAttribute('x'), 'a');
+
+      observer.observe(el, {attributes: true});
+
+      go('b');
+      await Promise.resolve();
```
sorvell: await a Promise that resolves at `setTimeout` or `requestAnimationFrame` to ensure you're beyond the microtask.

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
...
+      assert.equal(el.getAttribute('x'), 'b');
+      assert.equal(mutationCount, 1);
+
+      go('b');
+      await Promise.resolve();
```
sorvell: Await a task here too.

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
...
+          render(html`<div ?hidden="${live(x)}"></div>}`, container);
+
+      go(true);
+      const el = container.firstElementChild as LiveTester;
+      assert.equal(el.getAttribute('hidden'), '');
+
+      observer.observe(el, {attributes: true});
+
+      go(false);
+      await Promise.resolve();
```
sorvell: Await a task here too

src/test/directives/live_test.ts
```diff
@@ -0,0 +1,152 @@
...
+      assert.equal(el.getAttribute('hidden'), null);
+      assert.equal(mutationCount, 1);
+
+      go(false);
+      await Promise.resolve();
```
sorvell: Await a task here too

src/directives/live.ts
justinfagnani: Why? The value and previous value aren't necessarily strings.

sorvell: In the attribute case the `previousValue` is always a string. Casting `value` to string in this case avoids the foot gun that the user forgot to pass a string value.

LarsDenBakker:
Thanks for the cc. Looks similar to what we have, so we can defer to to the official implementation when it's released.

src/test/directives/live_test.ts
justinfagnani: done

PR #1061, opened 2020-01-15
Replace unsafeSVG tests

justinfagnani:
Trying the tests I wrote to see if they pass...


justinfagnani:
This fixes the tests simply because Microsoft browsers don't handle invalid SVG very well, and because it's more flexible in expectations.

PR #1064, opened 2020-01-16
ifDefined: Only set attributes when values change

### LINKED ISSUE #890: Using a directive in an attribute causes it to be marked dirty even if the final value is unchanged
# Description

The mechanism by which directives are "resolved" in AttributePart results in it erroneously marking a value as dirty as long as there is at least one level of directive "between" a template and the final output. This means that any attribute that is set to a directive is effectively always marked dirty and re-applied on every render.

# Steps to Reproduce

1. Write this code

```ts
import {html, render} from 'lit-html';
import {ifDefined} from 'lit-html/directives/if-defined';

function maybeWithMaxlength(haveMaxlength) {
  return html`<input id="my-text"
      type="text"
      maxlength="${ifDefined(haveMaxlength ? 20 : undefined)}" />`;
}

render(maybeWithMaxlength(true), document.body);

const input = document.querySelector('#my-text');
const observer = new MutationObserver(mutations => {
  alert('maxlength mutated!');
});
observer.observe(input, { attributes: true, attributeFilter: ['maxlength'] });

render(maybeWithMaxlength(true), document.body);
```

## Live Reproduction Link
https://stackblitz.com/edit/lit-html-thtgvg?file=src/index.js

## Expected Results
No alert would be shown (maxlength would not be mutated the second time, as its underlying value hasn't actually been changed)

## Actual Results
Alert is shown.

## Browsers Affected

- [X] Chrome
- [X] Firefox
- [X] Edge
- [X] Safari 11
- [X] Safari 10
- [X] IE 11

## Additional Notes
This looks to me to be a problem with `AttributePart#setValue`, here: https://github.com/Polymer/lit-html/blob/master/src/lib/parts.ts#L113

When a directive is set as the value, the template that contains the attribute calls `AttributePart.setValue` with the directive (thus changing `this.value` from the previous value of `20` to a reference to the directive function). When the directive function is later resolved as part of the commit process, `setValue` is called again with the result of the directive (still `20`), but since `this.value` is currently the directive and `20` is a primitive that isn't equal to the directive, the AttributePart gets marked as dirty.

justinfagnani:
Fixes #890

### DIFF
--- a/src/directives/if-defined.ts
+++ b/src/directives/if-defined.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {AttributePart, directive, Part} from '../lit-html.js';
+import {AttributePart, directive, noChange, Part} from '../lit-html.js';
 
 /**
  * For AttributeParts, sets the attribute if the value is defined and removes
@@ -27,6 +27,12 @@ export const ifDefined = directive((value: unknown) => (part: Part) => {
       part.committer.element.removeAttribute(name);
     }
   } else {
+    if (part instanceof AttributePart &&
+        part.committer.element.getAttribute(part.committer.name) === value) {
+      // If the value hasn't changed, we need to tell the part since parts
+      // treat any directive having dirtied the part.
+      value = noChange;
+    }
     part.setValue(value);
   }
 });
--- a/src/directives/if-defined.ts
+++ b/src/directives/if-defined.ts
@@ -28,7 +28,7 @@ export const ifDefined = directive((value: unknown) => (part: Part) => {
   if (value === undefined && part instanceof AttributePart) {
     // If the value is undefined, remove the attribute, but only if the value
     // was previously defined.
-    if (value !== previousValue || !previousValues.has(part)) {
+    if (previousValue !== undefined || !previousValues.has(part)) {
       const name = part.committer.name;
       part.committer.element.removeAttribute(name);
     }
--- a/src/directives/if-defined.ts
+++ b/src/directives/if-defined.ts
@@ -12,7 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {AttributePart, directive, noChange, Part} from '../lit-html.js';
+import {AttributePart, directive, Part} from '../lit-html.js';
+
+const previousValues = new WeakMap<Part, unknown>();
 
 /**
  * For AttributeParts, sets the attribute if the value is defined and removes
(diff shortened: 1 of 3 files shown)


src/directives/if-defined.ts
```diff
@@ -27,6 +27,12 @@ export const ifDefined = directive((value: unknown) => (part: Part) => {
       part.committer.element.removeAttribute(name);
     }
   } else {
+    if (part instanceof AttributePart &&
```
jridgewell: Wouldn't property parts need protection, too?

justinfagnani: Right... though this directive doesn't do anything interesting on any part other than attribute parts. They all need a fix, and I maybe regret making this work in those bindings. Also, the line below is wrong and I need a better test. This is definitely a flaw in the directive hook implementation though :( Need to think about this some more.

justinfagnani: I added a test for node parts which will also apply to other part types.

justinfagnani:
I realized #1066 was an issue too and added a skipped test for it in the PR, to be fixed later.

src/directives/if-defined.ts
```diff
@@ -14,19 +14,27 @@
...
  * For other part types, this directive is a no-op.
  */
 export const ifDefined = directive((value: unknown) => (part: Part) => {
+  const previousValue = previousValues.get(part);
+
   if (value === undefined && part instanceof AttributePart) {
-    if (value !== part.value) {
+    // If the value is undefined, remove the attribute, but only if the value
+    // was previously defined.
+    if (value !== previousValue || !previousValues.has(part)) {
```
jridgewell: Nit: we already know `value === undefined`, so I think it's clearer if we just test `previousValue` against `undefined`. And when defining it that way, we can reason about the `!previousValues.has(part)`. If `previousValues` doesn't contain `part`, that implies `previousValue === undefined`. So our check is actually:

```js
if (previousValue !== undefined || previousValue === undefined) {}
```

Which is always true. So let's just delete the if statement.

justinfagnani: `!previousValues.has(part)` is different than `previousValue !== undefined` though. A map will return `undefined` for both the case where the value wasn't present, and when it was but the value actually was `undefined`. Here we're checking if this is the first time we've seen this part, and if so removing the attribute.

I added a test that fails without this check.

jridgewell: `!previousValues.has(part)` is equivalent to `previousValue === undefined` (note no `!`).

The `previousValue !== undefined` (note `!`) is equivalent to `previousValue !== value`, because we know `value === undefined`.

justinfagnani: `!previousValues.has(part)` implies `previousValue === undefined`

`previousValue === undefined` does not imply `!previousValues.has(part)`

So we can't replace `!previousValues.has(part)` with `previousValue === undefined` and the expression doesn't reduce.

It's not that big of a deal though, at most it would lead to one extra call to `removeAttribute`.

jridgewell: Ohh, I see what you're saying now. If we `previousValues.set(part, undefined)`, we don't want to re-remove the attribute.

PR #1070, opened 2020-01-20
Change classMap to set classes correctly for SVGs

jridgewell:
This fixes the `classMap` directive inside SVGs by avoiding `className`, and avoiding `classList` which isn't supported in IE11 SVGs.

This is another attempt at https://github.com/Polymer/lit-html/pull/932. I've reused @AdamVig's commit to give him the commit stats.
Fixes https://github.com/Polymer/lit-html/issues/930.

### DIFF
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -14,7 +14,6 @@
 
 import {AttributePart, directive, Part, PropertyPart} from '../lit-html.js';
 
-
 export interface ClassInfo {
   readonly [name: string]: string|boolean|number;
 }
@@ -23,20 +22,21 @@ export interface ClassInfo {
  * Stores the ClassInfo object applied to a given AttributePart.
  * Used to unset existing values when a new ClassInfo object is applied.
  */
-const previousClassesCache = new WeakMap<Part, Set<string>>();
+const previousClassesCache = new WeakMap<Part, Map<string, unknown>>();
+
+const staticValue = {};
 
 /**
  * A directive that applies CSS classes. This must be used in the `class`
  * attribute and must be the only part used in the attribute. It takes each
  * property in the `classInfo` argument and adds the property name to the
- * element's `classList` if the property value is truthy; if the property value
- * is falsey, the property name is removed from the element's `classList`. For
- * example
+ * element's `class` if the property value is truthy; if the property value is
+ * falsey, the property name is removed from the element's `class`. For example
  * `{foo: bar}` applies the class `foo` if the value of `bar` is truthy.
  * @param classInfo {ClassInfo}
  */
 export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
-  if (!(part instanceof AttributePart) || (part instanceof PropertyPart) ||
+  if (!(part instanceof AttributePart) || part instanceof PropertyPart ||
       part.committer.name !== 'class' || part.committer.parts.length > 1) {
     throw new Error(
         'The `classMap` directive must be used in the `class` attribute ' +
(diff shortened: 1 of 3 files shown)


jridgewell:
Hi @AdamVig, can you comment "@googlebot I consent" so we can reuse your commit?

AdamVig:
@googlebot I consent

AdamVig:
@jridgewell Thanks for revisiting this, it has been on my list since my original pull request.

src/directives/class-map.ts
```diff
@@ -23,20 +22,21 @@ export interface ClassInfo {
  * Stores the ClassInfo object applied to a given AttributePart.
  * Used to unset existing values when a new ClassInfo object is applied.
  */
-const previousClassesCache = new WeakMap<Part, Set<string>>();
+const previousClassesCache = new WeakMap<Part, Map<string, unknown>>();
```
justinfagnani: I think this deserves a comment explaining what's happening, since the key in the Map is now either a class name or a static string...

src/directives/class-map.ts
```diff
@@ -71,13 +73,18 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
...
         previousClasses.delete(name);
       }
     }
   }
-});
+
+  if (changed) {
+    const classes: string[] = [];
+    previousClasses.forEach((_: unknown, key: string) => classes.push(key));
+    element.setAttribute('class', classes.join(' '));
```
justinfagnani: Since we're not using `.classList` anymore, this directive is essentially just doing custom dirty checking, and joining the property names of the `classInfo` object. I think there's a slightly simpler approach by using `part.setValue`.

Take the version at master, rip out the use of `classList` and setting of `className`, keep a `changed` variable as you do here, but then do:

```ts
if (changed) {
  part.setValue(Array.from(previousClasses).join(' ');
}
```

The AttributePart logic will take care of concatenating that with the static strings, so you won't need to keep them in the `previousClasses` Map.

src/directives/class-map.ts
jridgewell: Done.

src/directives/class-map.ts
```diff
@@ -21,10 +21,27 @@ export interface ClassInfo {
...
+ * the ClassInfo.
  */
-const previousClassesCache = new WeakMap<Part, Map<string, unknown>>();
+const previousClassesCache = new WeakMap<Part, Map<string, boolean>>();
 
-const staticValue = {};
+/**
+ * Classes are parsed as a series of tokens separated by ASCII whitespace. This
+ * token splitter allows us to extract the tokens from the static classes given
+ * in th template literal.
```
AdamVig: Typo, should be: `* in the template literal.`

src/directives/class-map.ts
jridgewell: > The AttributePart logic will take care of concatenating that with the static strings, so you won't need to keep them in the previousClasses Map.

That would be a breaking change, since you can currently override static classes with the `ClassInfo`. Otherwise, yes, it would be simpler.

src/directives/class-map.ts
jridgewell: Thanks!

src/test/directives/class-map_test.ts
```diff
@@ -95,6 +105,15 @@ suite('classMap', () => {
     assert.isFalse(el.classList.contains('foo'));
   });
 
+  test('adds classes on SVG elements', () => {
+    const cssInfo = {foo: 0, bar: true, zonk: true};
+    render(svg`<circle class="${classMap(cssInfo)}"></circle>`, container);
+    const el = container.firstElementChild!;
+    assert.isFalse(el.classList.contains('foo'));
```
justinfagnani: `classList` needs to be replaced with `className.split(' ')...` for SVG elements on IE

jridgewell: Done

src/test/directives/class-map_test.ts
```diff
@@ -81,6 +81,16 @@ suite('classMap', () => {
     assert.isFalse(el.classList.contains('foo'));
   });
 
+  test('can override static classes', () => {
+    renderClassMapStatic({aa: false, bb: true});
+    const el = container.firstElementChild!;
+    assert.isFalse(el.classList.contains('aa'));
+    assert.isTrue(el.classList.contains('bb'));
+    renderClassMapStatic({});
+    assert.isFalse(el.classList.contains('aa'));
```
justinfagnani: This seems bad to me, but I guess it's current behavior. I'd much rather have the behavior that the output is the same for the same input, regardless of order of calls (with the exception of out-of-band classList mutations).

jridgewell: Agreed. Wanna cut a v2 anytime soon?

justinfagnani: Not particularly :)

I'd want to move further on SSR and add a whole lot of benchmarks before that at least.

PR #1072, opened 2020-01-21
Support namespaces attributes

### LINKED ISSUE #423: Support namespaced attributes
Best I can tell, SVG cannot load external content with lit-html. The minimal repro is to use a LitElement based custom element with the template of 

```
<svg>
    <use xlink:href="external.svg#icon-id" />
</svg>
```

with a properly set up external.svg file but I have a full reproduction at https://github.com/silenceisgolden/lit-html-svg-test.

Thanks!

justinfagnani:
Fixes #423


src/lib/parts.ts
```diff
@@ -121,7 +122,14 @@ export class AttributeCommitter {
...
       kind: 'property'|'attribute' = 'attribute') {
     this.element = element;
-    this.name = name;
+    const colonIndex = name.indexOf(':');
+    if (colonIndex === -1) {
+      this.name = name;
+      this.namespace = null;
+    } else {
+      this.namespace = name.substring(0, colonIndex);
+      this.name = name.substring(colonIndex + 1);
```
rictic: Prefer slice over substring, as it behaves the same for both arrays and strings, and so is easier to remember and standardize on

src/lib/parts.ts
```diff
@@ -503,8 +512,15 @@ export class BooleanAttributePart implements Part {
...
     this.element = element;
-    this.name = name;
     this.strings = strings;
+    const colonIndex = name.indexOf(':');
+    if (colonIndex === -1) {
+      this.name = name;
+      this.namespace = null;
+    } else {
+      this.namespace = name.substring(0, colonIndex);
+      this.name = name.substring(colonIndex + 1);
```
rictic: ditto

src/lib/parts.ts
```diff
@@ -121,7 +122,14 @@ export class AttributeCommitter {
       templatePart?: AttributeTemplatePart,
       kind: 'property'|'attribute' = 'attribute') {
     this.element = element;
-    this.name = name;
+    const colonIndex = name.indexOf(':');
```
rictic: Is the colon syntax the standard/normal way to write this in HTML?

src/lib/parts.ts
rictic: Should we pass in this.name to the sanitizerFactory, or name with the colon?

jridgewell: And saves bytes!

src/lib/parts.ts
```diff
@@ -503,8 +512,15 @@ export class BooleanAttributePart implements Part {
...
     }
     this.element = element;
-    this.name = name;
     this.strings = strings;
+    const colonIndex = name.indexOf(':');
+    if (colonIndex === -1) {
+      this.name = name;
+      this.namespace = null;
+    } else {
+      this.namespace = name.substring(0, colonIndex);
```
jridgewell: Nit: helper method might save byes between the two different `Part` subclasses?

src/lib/parts.ts
```diff
@@ -121,7 +122,14 @@ export class AttributeCommitter {
       templatePart?: AttributeTemplatePart,
       kind: 'property'|'attribute' = 'attribute') {
     this.element = element;
-    this.name = name;
+    const colonIndex = name.indexOf(':');
+    if (colonIndex === -1) {
+      this.name = name;
+      this.namespace = null;
+    } else {
+      this.namespace = name.substring(0, colonIndex);
```
jridgewell: Oh, and this is incorrect. The NS is a URI, not the `xlink` substring.

src/lib/parts.ts
```diff
@@ -199,7 +207,7 @@ export class AttributeCommitter {
         // Native Symbols throw if they're coerced to string.
         value = String(value);
       }
-      this.element.setAttribute(this.name, value as string);
+      this.element.setAttributeNS(this.namespace, this.name, value as string);
```
sorvell: Let's be on the lookout for a perf regression here since `setAttributeNS` may not be as optimized as `setAttribute`.

src/lib/parts.ts
rictic: Can you expand on that? It's incorrect in what sense?

justinfagnani: https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttributeNS#Example

Is there a way to get the namespace URI from a prefix, right from the DOM? Otherwise this is going to be very tricky.

jridgewell: The `namespace` used in `getAttributeNS` and friends is `http://www.w3.org/1999/xlink`. But each NS can be totally unique URIs in SVG, you can just register them. See the console logs in https://output.jsbin.com/mexedos/quiet

justinfagnani:
I updated the tests to actually fail when the namespace isn't correct. I'll look for a way to get the real namespace of the attribute...

src/lib/parts.ts
jridgewell: HTML itself doesn’t have any namespaces (that do anything). SVG has `xml` and `xlink`, and any self registered namespaces don’t do anything. There’s also MathML, but it doesn’t matter without browser support.

I think you can get away with just hardcoding: https://github.com/google/incremental-dom/blob/master/src/attributes.ts#L28

justinfagnani:
I think in order to do this, we'd have to stash away the namespaces for attributes in the AttributeTemplatePart. It should be possible with something like: 

```ts
const attribute = (node as Element).attributes.getNamedItem(attributeLookupName)!;
const namespace = attribute.namespaceURI;
```

But this isn't working in the tests, I think because of the bound attribute suffix we place on all bound attributes in order to work around IE and Edge altering attribute values for special attributes like `style` and SVG attributes.

Overall, I don't think the complexity of the fix is worth it. `<use xlink:href="">` is the only real use case I'm aware of, and that's deprecated and Safari added support for the plain `href` attribute in 12.1. I think the recommendation should be to use `href` or a directive if you need to target older versions of Safari.

jridgewell:
@justinfagnani SVG2 is still super new, so requiring that to use SVG features isn't great.

I just pushed a few commits that should get us basic support for everything in SVG. I'm choosing not to support custom namespaces, because they're honestly useless in a browser.

src/lib/parts.ts
```diff
@@ -121,16 +121,11 @@ export class AttributeCommitter {
       // Next breaking change, consider making this param required.
       templatePart?: AttributeTemplatePart,
       kind: 'property'|'attribute' = 'attribute') {
+    const fqn = getAttrName(name);
```
rictic: Prefer immediate destructuring in this sort of case:

```typescript
const [namespace, name] = getAttrName(name);
```

src/lib/parts.ts
```diff
@@ -549,6 +539,21 @@ export class BooleanAttributePart implements Part {
...
+  'xml': 'http://www.w3.org/XML/1998/namespace',
+} as const;
+
+const getAttrName = (name: string): [string|null, string] => {
+    const colonIndex = name.indexOf(':');
+    if (colonIndex === -1) {
+      return [null, name];
+    }
+    const shorthand = name.slice(0, colonIndex);
+    const namespace = namespaces[shorthand] || null;
```
rictic: So if the namespace is unknown we just silently ignore it?

I'm not sure of a better option, but that's a bit unfortunate

justinfagnani:
@jridgewell MDN says that `<use href=>` is supported by all of our supported browsers: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use

Regardless of whether it's SVG2, what developer are we targeting that needs can't use plain `href`?

src/lib/parts.ts
jridgewell: This transpiles down to really icky code, so I usually avoid it whenever possible. I don't think there's any other use of iterator destructuring in the codebase yet. Maybe we should return an object instead, and object destructure that?

rictic: Ah, fair enough, I forget it's iterator destructuring and not array destructuring. This is fine as is, but please move this const down right next to lines where it's used.

jridgewell: Done.

jridgewell:
@googlebot I consent. (used my personal git email sig, :sigh:)

src/lib/parts.ts
justinfagnani: Not sure what else we can do here. I tried a few different ways to get the actual namespace URI from the DOM, but they all fail. Hardcoding the common prefixes or punting completely are the only options left. Since we've had multiple reports on this, I guess we should do it.

PR #1073, opened 2020-01-23
Rework TemplateInstance cloning

jridgewell:
Instead of traversing over every single `Element`, `Comment`, and `Text` node, this looks only for comment nodes that mark part positions. This likely allows us to skip **tons** of nodes in the walker (I'm assuming comments are the least used nodes in a page).

Additionally, I'm using a comment node to mark the spot after a template element, so that we know to go back and traverse the template.

By paying about 120 bytes, we can anywhere between a 2-10x speedup (for every browser) with this: https://jsbench.github.io/#d26f5d42b64a4c21dda947adcaaa1570

Additionally, this allows me to delete a significant amount of code from the shady-render module. We're no longer tied to walking nodes by `nodeIndex`, so we don't need to update every later part's `index` when we remove a single node, so removal is pretty trivial. We also don't need to pay for an expensive insertion function anymore.

(This replaces #847, which was on my personal fork and couldn't run Tach)

### DIFF
--- a/src/directives/async-append.ts
+++ b/src/directives/async-append.ts
@@ -84,7 +84,7 @@ export const asyncAppend = directive(
         // Check to see if we have a previous item and Part
         if (itemPart !== undefined) {
           // Create a new node to separate the previous and next Parts
-          itemStartNode = createMarker();
+          itemStartNode = createMarker('');
           // itemPart is currently the Part for the previous item. Set
           // it's endNode to the node we'll use for the next Part's
           // startNode.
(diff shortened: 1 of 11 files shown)


justinfagnani:
Tachometer is showing this as slower. Need to look into that.

justinfagnani:
This is obsolete with 2.0 now. We now do skip text nodes, but still visit every element. We could still try to move to only traversing comment nodes, but that would border on a breaking change with our current SSR and compiled template result support. It might be worth trying in a set of experiments for the next major version.

PR #1076, opened 2020-01-30
Port docs site to eleventy

justinfagnani:
This is mostly pixel perfect from what I can tell, except that I changed the syntax highlighting to a dark theme and a language plugin that supports lit-html templates :)

I changed how the guide collection and table of contents are generated a bit, but the output is the same.

I gave up on changing anything with the API docs, and couldn't even upgrade typedoc without weird changes. I also didn't change (or test!) the deployment environment, though I think we should move to Cloud Run with a standard Node server. Then we'll have exactly the same setup for dev.


justinfagnani:
Note to self and @arthurevans I still need to integrate analytics. There isn't a plugin from what I can tell, so I ust have to make sure all templates have the GA snippet.

arthurevans:
Looking pretty good. Some quibbles ATM. The TOC is kind of mangled--it's only including the top-level headings, and renders as a numbered list instead of a bullet list.

Also, the annotations we use to assign stable anchors to sections are getting included as part of the heading title (this also breaks the associated links).

eleventy-toc

The dark red color in the code samples fails WCAG AAA, but all the other colors that I tested pass.

arthurevans:
On p-p.org and the polymer library docs, we're using `markdown-it-attrs` to process the `{#anchorid}` annotations and friends (also used for classes, `target` attributes in a couple of places).

https://github.com/Polymer/polymer-library-docs/blob/master/gulpfile.js#L37

We were also using a module called `toc` for formatting the table of contents. Maybe we could filch the TOC formatting from polymer-library, which is much better than ours...

justinfagnani:
@arthurevans took care of everything except the syntax highlight color. We might choose another theme, though Okaidia looks the best to me so far.

arthurevans:
TOC looks good. A couple of more notes:

- markdown-it-attrs should probably be in the docs/package.json instead of the main package.json.
- the move to eleventy breaks links between documents. Extra-fun, whether the links works depends on where you came from!

Details on the links: the eleventy site defaults to adding a slash at the end of the path:

/guide/styling-templates/

In the old site, that's:

/guide/styling-templates

So where we have a relative link to a sibling page, like `[classMap](template-syntax#classmap)` the path gets transformed into: `/guide/styling-templates/template-syntax#classmap`.

To make things more complicated, if you follow a link from the interwebz, it'll take you to the old URL (`/guide/styling-templates` with no slash) which also works fine with eleventy... You'll be on the same page, except the relative links will work as expected.

I'm not sure if we can configure eleventy to just always use the old-style URLs? Otherwise we're going to need to make sure all of our links use absolute paths, which is a minor pain, but not the end of the world.

justinfagnani:
@arthurevans URLs and links are back to matching the current site. I switched the dev server to the `serve` npm package, which I think I have configured to match the behavior of the deployed site. I manually checked internal relative links in both the guide and API docs.

arthurevans:
I'm not seeing any change to the URLs. They still seem to end in `/`. Maybe some change didn't get checked in?

For example, if I click on "Styling templates" it sends me to `http://127.0.0.1:5000/guide/styling-templates/` and if I follow the link on that page to `classMap` it takes me to `http://127.0.0.1:5000/guide/styling-templates/template-reference#classmap`, which is a 404 (should be `/guide/template-reference#classmap`.

Likewise, if I click on a TOC link, I get the extra slash embedded: `http://127.0.0.1:5000/guide/styling-templates/#bindings-in-style-sheets` instead of `http://127.0.0.1:5000/guide/styling-templates#bindings-in-style-sheets`.

justinfagnani:
@arthurevans I'm not seeing that. When I click on "Styling Templates" I get brought to `http://localhost:5000/guide/styling-templates`, though I just noticed that's because the link is to `http://localhost:5000/guide/styling-templates.html` and is redirected. I'll work on that out-bound link.

justinfagnani:
Ok, got that sorted with a custom filter to remove the extensions. I looked around to try to make sure the guide TOC was the only place that needed it.

PTAL @arthurevans

Issue #1046, opened 2019-12-05
Incorrect rendering of text values in-between slots on the web component polyfill


LarsDenBakker:
I'm running into quite a strange bug. I have a web component which, after minification, has two slots and a text binding in-between. On modern browsers this works fine, but on browsers which need the web component polyfill the first rendered value is never removed.

It's best observed in this jsbin: https://jsbin.com/caradirigo/2/edit?html,output there are two renders 'foo', and 'bar'. 'foo' should be removed when 'bar' is rendered.

I forced the polyfills, so it's observable on chrome as well.

When there is a whitespace character, it does work correctly: https://jsbin.com/govovaqicu/1/edit?html,output

frank-dspeed:
To avoid a unneeded whitespace you can do 
```
<slot name="before"></slot>${this.title}<!-- --><slot name="after"></slot>
```

i added a comment node

LarsDenBakker:
Absolutely, there are many ways to work around it. This is still a bug, though.

justinfagnani:
This is likely because the polyfill is removing `<slot>` elements.

@sorvell and @dfreedm is there any way to leave the elements in, or replace them with a comment?

frank-dspeed:
if is pollyfill wrap slot on parsing with makeComment ?

justinfagnani:
This is being fixed in Shady DOM, but I think we should have a test here that it actually fixes this problem. cc @sorvell

sorvell:
Closing based on #1077

PR #1077, opened 2020-02-03
Adds rendering test composing parts and slots

sorvell:
Tests if #1046 has been addressed in the webcomponents polyfills.

Note, this test will pass after @webcomponents/webcomponentsjs is released including https://github.com/webcomponents/polyfills/pull/254.


justinfagnani:
@sorvell LGTM, but could you just skip the test for now? We'll unskip and bump the dependency when it's released.

justinfagnani:
@sorvell can you also rebase on master so the tests pass?

sorvell:
@justinfagnani PTAL Rebased and updated the polyfill version dev dependency.

PR #1078, opened 2020-02-07
Fix Chrome 80 test timeouts.

### LINKED ISSUE #1080: Tests timing out on Chrome 80
Chrome 80 has apparently increased the background timer throttling, which is slowing down tests drastically causing them to timeout.

I open a Chrome bug open here: https://bugs.chromium.org/p/chromium/issues/detail?id=1050221 and added a potential reproduction based on awaiting `setTimeout` calls.

A comment by @shaseley says that adding the `--disable-background-timer-throttling` flag stopped the slowdown. I added this to our wct.conf.json file in #1078 and this did fix the tests on my machine when, but the tests are still timing out on Travis. Travis isn't very fast, so it's possible that other Chrome scheduler changes have push tests there over the timeout threshold.

cc @dfreedm @rictic

justinfagnani:
Fixes #1080

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
-sudo: required
-dist: trusty
+sudo: false
+dist: xenial
 node_js: '10'
 addons:
   firefox: "71.0"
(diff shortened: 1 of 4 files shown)


shaseley:
Try adding --disable-renderer-backgrounding, which I believe affects process priority. There are [some comments](https://cs.chromium.org/search/?q=disable-renderer-backgrounding&type=cs) that indicate this was a problem for other folks doing automated testing in the background.

AFAIK the throttling code has been around for a while and it's not clear to me what changed in M80 that's causing issues now, but hopefully adding this flag will help.

shaseley:
One more thing we could try for scheduling that I know of: we could try adding --disable-features=BlinkSchedulerVeryHighPriorityForCompositingAfterDelay.

This is an experiment running on M80 (started in 79) that decreases the latency of some frames by prioritizing rendering if it's queued for too long. I've run the tests locally on my Linux machine with the feature both enabled and disabled and don't notice a difference, visually anyways — both chrome and firefox finish at about the same time in both cases. But maybe because of less resources on travis, rendering is somehow starving the tests and causing timeouts?

justinfagnani:
Thanks for the help @shaseley. Still no luck. It's interesting that `disable-background-timer-throttling` helped locally and not on Travis. It seems like multiple things must have slowed down our tests in M80 and any one of them is enough to cause a timeout on Travis. We've been talking about moving to GitHub actions because they're a lot faster. I think it's time. cc @aomarks

shaseley:
Ah sorry those flags didn’t help :(. And yeah I’m wondering if it’s something other than timer throttling that’s causing the problem. AFAIK the throttling has been around for quite a while and I don’t know of any recent changes to the logic. I was able to see similar behavior in M79 with your repro case.

I’m also really curious about what the difference might be on Travis. When I ran these locally, they finished around the same time. Based on the runs after you increase the timeout, Chrome looks like it’s either reaaaally slow or “stuck”. Is there a way to get debug output, like the --verbose in wct? I was going to try to do this when I was hacking around on this but didn’t know how to get that enabled on Travis.

Also, I was looking at the raw log from your last run and saw it's using chromedriver v79. Not sure if there could be something weird going on there with interaction between versions.

justinfagnani:
The confusing thing is that `--disable-background-timer-throttling` helped a _huge_ amount on my machine. Part of me was wondering if the flags aren't making it to Chrome on Travis, but now I'm only running Chrome tests, and turned off headless, so it shouldn't even be backgrounded on Travis. I also increased the WCT timeout by 10x, and now Travis itself is timing out after 10m. Something else seems up.

I'd really like to repro the hang locally, but I'm failing at that too.

justinfagnani:
I added `--verbose` and I'm watching the current run not even start tests... I wonder if you're right with the chromedriver version.

justinfagnani:
@shaseley looks like it was Travis config / Linux version problems. Got it going now. Thanks for your help!

shaseley:
Awesome, glad this is working now!

FYI on those flags, the --disable-features='BlinkScheduler...' one will eventually stop working as we're planning to ship that feature and remove the feature path, so it might be good to try removing that.

PR #1079, opened 2020-02-07
named part

### LINKED ISSUE #1081: direct access to NodePart, AttributePart, BooleanAttributePart, EventPart
we are rendering 3k+ records, happy with lit-html rendering/update speed.

but records are editable, during edit mode there is input element which overlap on record html row.
when user finish his editing and click "save" button, we want to hide input elements immediately,  but now we dont want to build/render complete rows again to see newly entered values into html rows, if we do so, build/render takes some time and  first we see new values in input, then input becomes invisible, then we see old values in html rows then after few seconds we see new values, which seems crazy. 

if we directly update html nodes for new values,  then "cancel changes" dont show old values.
did not tried "live" directive, but i think it will effect speed of rendering 

so i think there should be a way to directly access required "part" of template so that we can update html nodes using `part.setValue("new value"); part.commit(); `

surinder83singh:
Fixes #1081 

```js
let tpl = html`<div id="btn" class=${partRef('cls1',  'class')}>${partRef('body contents',  'body')}</div>`;
render(tpl, document.body);

let parts = partRefs.get(document.getElementById('btn'))

parts['class'].setValue("new class")
parts['class'].commit();
parts['body'].setValue("new contents")
parts['body'].commit();
```

### DIFF
--- a/src/directives/part-ref.ts
+++ b/src/directives/part-ref.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {directive, Part, AttributePart, BooleanAttributePart, NodePart, EventPart} from '../lit-html.js';
+import {AttributePart, BooleanAttributePart, directive, EventPart, NodePart, Part} from '../lit-html.js';
 
 /**
  * For AttributeParts, NodeParts, sets the reference of parts on partRefs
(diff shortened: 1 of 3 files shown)


justinfagnani:
@surinder83singh please see the contributing guide section on PRs: https://github.com/Polymer/lit-html/blob/master/CONTRIBUTING.md#pull-requests

We really want to have discussion on new features before PRs to add them and make sure there's consensus and that any extra cost or complexity is worth it. I'd rather not have those discussions in the PR itself.

PR #1090, opened 2020-02-16
Fix a typo in the template reference guide

rpatterson:


rpatterson:
@googlebot I signed it!

Issue #1068, opened 2020-01-16
[docs] Document unsafeSVG() directive


justinfagnani:
Added in #1000

justinfagnani:
Fixed by #1091

Issue #1067, opened 2020-01-16
[docs] Document templateContent() directive


justinfagnani:
Added in #1058

justinfagnani:
Fixed by #1091

PR #1091, opened 2020-02-18
Document live, templateContent, unsafeSVG.

### LINKED ISSUE #1063: [docs] Document live() directive
It's used to check against the actual DOM value (attribute, property, or boolean attribute only) instead of the previous values set though lit-html. This is useful in cases where the underlying value might change underneath lit-html, but the template author wants the lit-html provided version to overwrite the DOM value. The classic case is:

```js
html`<input .value=${x}>`
```

Added in #1057

arthurevans:
Fixes #1063, #1067, #1068.


docs/_guide/06-template-reference.md
```diff
@@ -420,6 +424,38 @@ const template = html`
...
+bound value, when determining whether to update the value.
+
+This is useful for cases where the DOM value may change from outside of
+lit-html. For example, when binding to an `<input>` element's `value` property,
+a content editable element's text, or to a custom element that changes its
+own properties or attributes.
+
+In these cases if the DOM value changes, but the value set through lit-html
+bindings hasn't, lit-html won't know to update the DOM value and will leave
+it alone. If this is not what you want—if you want to overwrite the DOM
```
justinfagnani: Add spaces here: "want-if", and should it be balanced with another "-" at the end of the clause?

docs/_guide/06-template-reference.md
```diff
@@ -481,6 +517,33 @@ be combined with static values:
...
+import {templateContent} from 'lit-html/directives/template-content';
+
+const templateEl = document.querySelector('template#myContent');
+
+const template = html`
+  Here's some content from a template element:
+
+  ${templateContent(templateEl)}`;
+
+render(template, document.body);
```
justinfagnani: Maybe leave out the render call?

docs/_guide/06-template-reference.md
arthurevans: Our house style is to use em dashes with no spaces. These do look weird in monospace fonts because the em dash is the same size as a hyphen, but it looks good when it's set in a normal font.

https://developers.google.com/style/dashes
Example in the current docs: https://lit-html.polymer-project.org/guide/template-reference#template-structure

docs/_guide/06-template-reference.md
arthurevans: SGTM

docs/_guide/06-template-reference.md
arthurevans: Added the closing dash, updated the API docs to match.

arthurevans:
Should be ready to merge @justinfagnani ... Addressed comments & made minor tweaks to the API docs to match (we don't actually render the API docs for directives ATM, so this is kind of a no-op).

justinfagnani:
Thanks @arthurevans !

PR #1094, opened 2020-02-20
Update terser to drop all comments

justinfagnani:
Terser started preserving license comments by default, so our bundle size checks over time aren't comparable. This reverts the behavior to what we had at the 1.1.2 release. Because we don't publish the bundle, and only use it for size tracking, it's ok to remove the license comments.


rictic:
The JS source is the expensive part to parse and init, so this makes sense as the thing to track (though it will diverge from bytes on the network)

PR #1098, opened 2020-02-25
Revert sanitization and trusted type commits

justinfagnani:
This reverts commits 24e2ab3, 3a5b34b, 2e8e797, 2f67860, cd657bd, 2a719e6, ceed9ed, d5cf13c, 2fcf37a, and b640da9.

For each revert I've kept the repo formatted, linting and building clean, and I ran local tests. We can un-revert this either as one squashed commit, or separate commits if we wish.


src/lib/parts.ts
```diff
@@ -131,34 +68,12 @@ export class AttributeCommitter {
...
-      const v = parts[0].value;
-      if (!isIterable(v)) {
-        return v;
-      }
-    }
     let text = '';
 
     for (let i = 0; i < l; i++) {
       text += strings[i];
-      const part = parts[i];
```
rictic: I think we should keep these changes, as they let us pass non-string values straight into the DOM, which is desirable for trusted types (even if we revert the trusted types integration where we .innerHTML)

justinfagnani:
I think I got all the changes to default-template-processor.ts and template.ts

src/lib/parts.ts
justinfagnani: I did revert the trusted types integration too. My feeling is we can easily bring this back, and as long as we're reverting basically everything, we might as well do everything. Keeping this would still be a few extra bytes for no observable difference yet.

Another way of looking at this is that we shouldn't split the cost of a feature across too many changes as that'll obscure the cost.

PR #1104, opened 2020-02-28
Add 1.2.0 release notes

justinfagnani:
Fixes: #1086

### DIFF
--- a/docs/guide/release-notes/1.2.0.md
+++ b/docs/guide/release-notes/1.2.0.md
@@ -1,7 +1,7 @@
 ---
 title: lit-html 1.2.0
 slug: lit-html-1-2-0
-tag: release-notes
+tags: release
 ---
 
 See the [changelog](https://github.com/Polymer/lit-html/blob/master/CHANGELOG.md) for a complete list of changes.
@@ -30,7 +30,7 @@ const onInput = (e) => {
 };
 
 const go = () => {
-  html`<input .value=${text} @input=${onInput}>`
+  render(html`<input .value=${text} @input=${onInput}>`, document.body);
 }
 go();
 ```
--- a/docs/guide/release-notes/1.2.0.md
+++ b/docs/guide/release-notes/1.2.0.md
@@ -1,0 +1,75 @@
+---
+title: lit-html 1.2.0
+slug: lit-html-1-2-0
+tag: release-notes
+---
+
+See the [changelog](https://github.com/Polymer/lit-html/blob/master/CHANGELOG.md) for a complete list of changes.
+
+## `live` directive
+
+[Documentation](../template-reference#live)
+
+The `live` directive solves the problem of DOM state changing from underneath lit-html, for example with an `<input>` that a user can type into:
+
+```js
+html`<input .value=${live(x)}>`
+```
+
+In some cases you will want `x` to overwrite what the user has typed in. lit-html tracks the previous value of a binding and only updates the DOM is the value has changed. Without the live directive, lit-html might not think that the `value` property of the input has changed and won't update the DOM.
+
+The live directive ignores the previous value of the binding and always checks the current value in the DOM, causing the binding to update the input.
+
+We still reccomend that you handle user input so that the state and the DOM agree:
+
+```ts
+let text = '';
+const onInput = (e) => {
+  text = e.target.value;
+  go();
+};
+
+const go = () => {
+  html`<input .value=${text} @input=${onInput}>`
+}
+go();
+```
+
+## `templateContent` directive
+
+[Documentation](../template-reference#templatecontent)
+
+The `templateContent` directive lets you stamp out HTML templates into lit-html templates. This is useful in a number of cases where HTML `<template>` elements are provided by users of elements or parts of a build system.
+
+HTML:
+```html
+<template id="example">
+  <p>HTML Template</p>
+</template>
+```
+
+JavaScript:
+```js
+const template = document.querySelector('#example');
+html`
+  <h1>Example</h1>
+  ${templateContent(template)}
+`;
+```
+
+## `unsafeSVG` directive
+
+[Documentation](../template-reference#unsafesvg)
+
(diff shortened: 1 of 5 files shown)


docs/guide/release-notes/1.2.0.md
```diff
@@ -0,0 +1,75 @@
...
+html`<input .value=${live(x)}>`
+```
+
+In some cases you will want `x` to overwrite what the user has typed in. lit-html tracks the previous value of a binding and only updates the DOM is the value has changed. Without the live directive, lit-html might not think that the `value` property of the input has changed and won't update the DOM.
+
+The live directive ignores the previous value of the binding and always checks the current value in the DOM, causing the binding to update the input.
+
+We still reccomend that you handle user input so that the state and the DOM agree:
+
+```ts
```
arthurevans: This example needs some explanation... I can't follow this at all. onInput returns a TemplateResult? how does that get rendered?

docs/guide/release-notes/1.2.0.md
```diff
@@ -0,0 +1,75 @@
...
+
+```js
+html`<input .value=${live(x)}>`
+```
+
+In some cases you will want `x` to overwrite what the user has typed in. lit-html tracks the previous value of a binding and only updates the DOM is the value has changed. Without the live directive, lit-html might not think that the `value` property of the input has changed and won't update the DOM.
+
+The live directive ignores the previous value of the binding and always checks the current value in the DOM, causing the binding to update the input.
+
+We still reccomend that you handle user input so that the state and the DOM agree:
```
arthurevans:
```suggestion
We still recommend that you handle user input so that the state and the DOM agree:
```

docs/guide/release-notes/1.2.0.md
```diff
@@ -0,0 +1,75 @@
...
+
+[Documentation](../template-reference#live)
+
+The `live` directive solves the problem of DOM state changing from underneath lit-html, for example with an `<input>` that a user can type into:
+
+```js
+html`<input .value=${live(x)}>`
+```
+
+In some cases you will want `x` to overwrite what the user has typed in. lit-html tracks the previous value of a binding and only updates the DOM is the value has changed. Without the live directive, lit-html might not think that the `value` property of the input has changed and won't update the DOM.
```
arthurevans:
```suggestion
In some cases you will want `x` to overwrite what the user has typed in. lit-html tracks the previous value of a binding and only updates the DOM if the value has changed. Without the live directive, lit-html will not detect that the `value` property of the input has changed and won't update the DOM.
```

docs/guide/release-notes/1.2.0.md
```diff
@@ -0,0 +1,75 @@
...
+`;
+```
+
+## `unsafeSVG` directive
+
+[Documentation](../template-reference#unsafesvg)
+
+`unsaveSVG` is the missing partner of [`unsafeHTML`](../template-reference#unsafehtml). It lets you render a frangment of SVG text as SVG elements rather than text. As with `unsafeHTML` this directive not safe to use with user-provided input, since if the input has `<script>` tags in it, there may be ways to get them to execute, etc.
+
+`unsafeSVG` creates elements in the SVG namespace, so it's for use inside of `<svg>` tags or inside of lit-html ``` svg`` ``` templates. If the input contains an `<svg>` tag itself, continute to use `unsafeHTML`.
```
arthurevans: We haven't used this format (``` html`` ```) in discussing HTML templates, and it looks a little strange here. 

Maybe just use `svg` and link it to the API doc page so it's clear what we're talking about? 
``` [`svg`](/api/modules/lit_html.html#svg) ```

arthurevans:
I think it might be good to keep the release notes in a single doc for now. If we imaging a future version with five or ten point releases with one or two new features in each, folks are going to need to click through a lot of docs to see all the relevant changes from 1.2.0 to 1.5.3 or whatever.

A single file with most-recent release on top is a little more imposing, but easier to search. For instance, for the polymer library, we had one release notes file per major release:

https://polymer-library.polymer-project.org/2.0/docs/release-notes

docs/guide/release-notes/1.2.0.md
justinfagnani: Done, but that link destination leaves a lot to be desired. It doesn't describe what `svg` does at all, and I just noticed we don't mention it anywhere in the docs. I'll open a new issue for add it to "Writing Templates".

justinfagnani:
I like having a page per release because I think it will encourage the notes to be more conversational and more like a blog post. Then we don't have to decide whether to do a blog post most of the time, we just do the release notes, and we can send that out if we want.

TypeScript actually does both. They have an overview page with everything, and a page per release. We could pretty easily do this with eleventy collections.

docs/guide/release-notes/1.2.0.md
arthurevans: 👍

docs/guide/release-notes/1.2.0.md
arthurevans: Where does it re-render? 

I know this seems really simple to you, but I don't understand how this would fit into running code.  Doesn't it need a call to render somewhere?

I tried to put this into a jsbin and got nothing. https://jsbin.com/jesixadelo/edit?html,output

justinfagnani: Sorry, I forgot the render call inside go(): https://jsbin.com/qijedenulu/edit?html,output

arthurevans: OK, I think there's a missing render around the template in `go()`?

arthurevans:
TypeScript has hundreds of lines release notes on each of their releases, so I think they're solving a problem we don't have here. 

If you want to do the separate pages and add a collection page that gathers them all, it's 👍  by me as long as it doesn't create more work per release.

docs/guide/release-notes/1.2.0.md
arthurevans: It all makes sense now 😂

PR #1112, opened 2020-03-12
Don't disturb imperatively added classes in classMap

### LINKED ISSUE #1111: classMap doesn't preserve imperatively added classes
# Description

While imperatively changing classes that are managed by classMap _shouldn't_ be a supported feature, the previous version of classMap supported it to an extent, and the new version removes imperatively added classes in more situations. This has caused internal test failures at Google.

# Steps to Reproduce

1. Write this code

```ts
const go = (classes) => render(html`<div id="a" class=${classMap(classes)}></div>`, document.body);
go({foo: true});
const el = document.querySelector('#a');
el.classList.add('bar');

go({foo: false});
el.classList.contains('bar'); // false 😢
```

## Browsers Affected
All

justinfagnani:
Fixes #1111

Revert classMap to previous implementation, but don't use classList or className.

### DIFF
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -90,6 +90,8 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
   }
 
   if (changed) {
-    element.setAttribute('class', Array.from(classes).join(' '));
+    let classString = '';
+    classes.forEach((cls) => classString += cls + ' ');
+    element.setAttribute('class', classString);
   }
 });
--- a/src/directives/class-map.ts
+++ b/src/directives/class-map.ts
@@ -14,6 +14,7 @@
 
 import {AttributePart, directive, Part, PropertyPart} from '../lit-html.js';
 
+
 export interface ClassInfo {
   readonly [name: string]: string|boolean|number;
 }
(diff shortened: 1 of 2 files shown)


src/test/directives/class-map_test.ts
```diff
@@ -81,14 +81,24 @@ suite('classMap', () => {
     assert.isFalse(el.classList.contains('foo'));
   });
 
-  test('can override static classes', () => {
```
justinfagnani: This behavior was never released so it's safe to change. With the implementation in this PR, static classes are always present, even if they're apparently overridden in the classMap. I think this behavior is fine - it should be very rare that a class is in both static and dynamic sets, but if someone was doing this then changing it would have been a breaking change.

src/directives/class-map.ts
```diff
@@ -66,55 +48,44 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
...
-    }
-    changed = true;
+    // Write static classes once
+    // Use setAttribute() because className isn't a string on SVG elements
+    element.setAttribute('class', committer.strings.join(' '));
+    previousClassesCache.set(part, previousClasses = new Set());
   }
 
+  // IE11 doesn't support classList on SVG elements, so we emulate it with a Set
+  const classes = new Set((element.getAttribute('class') || '').split(/\s+/));
```
jridgewell: We had a bug before about `Set` not supporting iterable in the constructor.

src/directives/class-map.ts
```diff
@@ -66,55 +48,44 @@ export const classMap = directive((classInfo: ClassInfo) => (part: Part) => {
...
+      changed = true;
     }
   });
 
-  // Add or remove classes based on their ClassInfo value
+  // Add or remove classes based on their classMap value
   for (const name in classInfo) {
     const value = classInfo[name];
-    const cached = previousClasses.get(name);
     // We explicitly want a loose truthy check of `value` because it seems more
```
jridgewell: This comment now applies to the `if (value)` statement inside this if statement.

src/test/directives/class-map_test.ts
```diff
@@ -81,14 +81,24 @@ suite('classMap', () => {
...
+    renderClassMap({foo: true});
+    const el = container.firstElementChild!;
+    assert.isTrue(el.classList.contains('foo'));
+
+    el.classList.add('bar');
+    assert.isTrue(el.classList.contains('bar'));
+
+    renderClassMap({foo: false});
+    assert.isFalse(el.classList.contains('foo'));
+    assert.isTrue(el.classList.contains('bar'));
```
jridgewell: Is this a behavior we actually want to support? I think lit takes complete ownership of every other attribute.

src/directives/class-map.ts
justinfagnani: I remember that too, but checked MDN and tried it... maybe that's the Chrome 41 error too.

src/test/directives/class-map_test.ts
justinfagnani: I know, but this change caused several failures in Google tests. Fixing them requires architectural change that I can't do right now.

justinfagnani:
I added the ponyfill. Running through tests now

PR #1129, opened 2020-03-19
Use downlevel-dts, test compile in TS 3.4

rictic:
Related to https://github.com/Polymer/lit-element/issues/935

Haven't heard anything about this being a problem for lit-html, but better to be ahead of any issues.


justinfagnani:
We had a breakage reported here: https://github.com/Polymer/lit-element/issues/935#issuecomment-601313642

Issue #1134, opened 2020-03-28
Export the shadyTemplateFactory function


manolakis:
## Initiative / goal
One to two sentences that describe:
-Who wants the functionality

Open-wc for scoped elements (https://github.com/open-wc/open-wc/pull/1464).

-What it is they want

We need the [`shadyTemplateFactory`](https://github.com/Polymer/lit-html/blob/master/src/lib/shady-render.ts#L58) exported to call it after preprocessing the `TemplateResult` scoping the elements.

-Why they want it

We don't want to loose the shady render functionality because the use of scoped elements and right now we have to copy the shadyTemplateFactory which is not ideal.

### Scope
Description of task/subtask

Just export the `shadyTemplateFactory`.

## Acceptance criteria and must have scope

`shadyTemplateFactory` should be available outside the `shady-render.ts` file.

justinfagnani:
Ok, we can do this.

PR #1135, opened 2020-03-28
Export shadyTemplateFactory

manolakis:
Fixes: #1134


manolakis:
@googlebot I signed it!

justinfagnani:
Merging this for now, though since our strategy for polyfill support might change drastically in lit-html 2.0, we should talk about alternate ways to integrate with this.

PR #1141, opened 2020-04-10
Lint all the project, not just src

abdonrd:
Also move `eslint-disable` comments to top as best practice.

And now in sync with the LitElement PR: https://github.com/Polymer/lit-element/pull/861 🎉


justinfagnani:
What does this get us?

abdonrd:
Lint all the project instead of just the src folder.

justinfagnani:
I gather that from the description, but what do we get out of doing that? What code outside of src is worth linting?

PR #1144, opened 2020-04-22
Create parts directly in hydrate()

justinfagnani:
This updates `hydrate()` to put the TemplateInstance tree into a usable state before `render()` is ever called. This removes hydration code from Parts, especially NodePart's handling of TemplateResults.

As part of this I changed how attribute bindings are marked. Now we just emit a marker comment after an element with attribute bindings that contains the node index in the template. This lets us continue to walk only the comment nodes during hydration, and associate attribute bindings by index.

I've reverted all changes to part.ts, all SSR-related options from RenderOptions, and a few new types that were used to pass "prerendered" part descriptions.

With this attribute expressions and nested templates are working, at least minimally. We're now priming the current part values, so I was able to add a test check that there are no DOM mutations for hydration + initial re-render with the same data.

As part of removing the SSR-specific code from parts, we don't have the code to emulate SSR'ed DOM in lit-html anymore. This is fine because that was diverging from lit-ssr. I have ported over the client-side only hydration tests from lit-html to the lit-ssr integration tests to retain coverage.

This PR breaks `repeat()`, so the `repeat()` tests ported to lit-ssr are skipped there. I think we need a bit of a rethink on directive handling to make this pay for play. I'm mulling over a way that directives can emit the serializable state they need for `hydrate()` to work.

The corresponding PR to emit the new `<!--lit-bindings {x}-->` marker is coming in soon.


src/lib/hydrate.ts
```diff
@@ -12,79 +12,257 @@
...
+ *
+ *   html`<div class=${x}>${y}</div>`
+ *
+ * The SSR DOM is:
+ *
+ *   <!--lit-part AEmR7W+R0Ak=-->  # Start marker for the root Node Part created
+ *                                 # by render(). Includes the digest of the
+ *                                 # template
+ *   <div class="TEST_X">
+ *     <!--lit-bindings 0--> # Indicates there are attribute bindings here
```
sorvell, line 78:
```suggestion
 *     <!--lit-bindings 0--> # Indicates there is an attribute binding here. Note, there is one comment per attribute binding.
```

src/lib/hydrate.ts
```diff
@@ -12,79 +12,266 @@
...
+            });
+            value = rootValue;
+          } else {
+            const state = stack[stack.length - 1];
+
+            if (state.container) {
+              // We have a current template instace, so use its template
+              // processor to create parts
+              currentNodePart =
+                  state.instance.processor.handleTextExpression(options);
```
kevinpschaaf, line 149: Just noting, the naming of the template processor API is pretty confusing from this point of view but I guess it's too late to change since that's overridable.  A better name might have been `createAttributeParts` and `createNodePart` maybe, and then here, `createNodePart` would read better.

kevinpschaaf: Also worth pointing out, the only reason to use the `processor` is to allow customizable `NodePart`/`AttributePart`s, right?  But the current design sorta has knowledge of those baked into this module.

Maybe `hydrate` needs to take a `hydrationProcessor` or something that goes along with any processor overrides and implements the per-part-type value initialization on L243-257?  Maybe premature, because the SSR-side also has baked-in knowledge of what's in the template processor too... Something to think about.

kevinpschaaf:
Looks like CI is complaining on clang-format

src/lib/hydrate.ts
```diff
@@ -12,79 +12,266 @@
...
+
+/**
+ * Information needed to rehydrate a single TemplateResult.
+ *
+ * TODO: this doesn't let us handle nested NodeParts that aren't the result of
+ * nested templates, as repeat() creates. We need the stack to contain all
+ * NodeParts for that.
+ */
+type NodePartState = {
+  container: false;
```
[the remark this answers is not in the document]
kevinpschaaf: Oh, I see in #1155 changed this to a `type` field for leaf/iterable/template-instance.

src/lib/hydrate.ts
justinfagnani: Using the template processor should be good for now, as that's the main customization point. The places where we create parts directly correspond to the same in lit-html, and the value priming maybe should be something we eventually bake into the Part API.

PR #1147, opened 2020-04-27
Doc usability improvements. Fixes #1133.

### LINKED ISSUE #1133: Early documentation makes assumptions about directive imports
I am stepping through the lit-html docs page by page to learn how it all works. I suspect many others are doing the same and appreciate the uncluttered approach of lit-html. I found a few spots that left me asking, "Why does this example not work."

## The problem: Location of directives are unclear in early documentation pages

In the [Writing Templates](https://lit-html.polymer-project.org/guide/writing-templates) page there are references made to _directives_ like repeat and cache. The same references are made on Styling Templates to _directives_ like styleMap and classMap. It's only through following the 'more on style map' link that I discovered where the directives are located and how to include them. Eliminating these points of friction will help people progress through the documentation more effortlessly. 

## Proposed solution: Update code block examples to include import statements for directives

Add the import statements to the code block examples on [Writing Templates](https://lit-html.polymer-project.org/guide/writing-templates) and [Styling Templates](https://lit-html.polymer-project.org/guide/styling-templates) so that they are more clear, consistent with entries on the [Template Syntax Reference: classMap](https://lit-html.polymer-project.org/guide/template-reference#classmap), etc

## Other recommendations (or for discussion)

1. [Writing Templates: Bind to Properties](https://lit-html.polymer-project.org/guide/writing-templates) should use standard html elements instead of custom elements to avoid confusion. Readers at this stage are still getting acclimated to lit-html. something like <input .value=${data.value} > is more clear. Perhaps add the native component example as a preceding example to the custom element example that exists for <my-list>

2. [Writing Templates: Nothing and the slot fallback content](https://lit-html.polymer-project.org/guide/writing-templates) makes many assumptions about reader knowledge early on. At this point in the docs there is no introduction about what a slot is, how html`<slot>...</slot>` relates to the <shadow-element> or how that fallback is pulled in when nothing is used in the example. 

3. Place "styleMap" and "classMap" at the beginning of the [Styling templates](https://lit-html.polymer-project.org/guide/styling-templates) page because they are the most simple to implement and require the least knowledge. Starting with things that look more like authoring stock html is easier for on-boarding. Complexity can build over time. Since lit-html is about authoring html (not web components) I believe it should start with concerns about simple html first, and then move to its integrations and implications with web components, shadow DOM, polyfills, etc.

arthurevans:


docs/guide/02-writing-templates.md
```diff
@@ -267,26 +274,47 @@ Which repeat is more efficient depends on your use case: if updating the DOM nod
...
-`nothing` is a sentinel value provided by lit-html. It really does render nothing, because
-it clears the Part.
 
-Why is this important? Imagine you have a shadow DOM enabled custom element, `shadow-element`, that uses a slot. 
-The template looks like this:
+One specific use case where an empty text node causes issues is when you're using a `<slot>` element inside a shadow root. 
+
+<div class="alert alert-info">
+
+**Not using shadow DOM?** Understanding this use case requires some knoweldge of shadow DOM and slots. On the bright side, if you're not using lit-html to render into shadow DOM, you will probably never need to use `nothing`, and you can skip this section.
```
justinfagnani, line 304: Is this note needed? I feel like this could be a note about any feature... not using it? Don't worry about it! :)

Maybe instead of a note, include links to Shadow DOM docs on MDN like https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

docs/guide/02-writing-templates.md
```diff
@@ -267,26 +274,47 @@ Which repeat is more efficient depends on your use case: if updating the DOM nod
...
+
+</div>
+
+Imagine you have a shadow DOM enabled custom element, `shadow-element`, that uses a slot. The template for the shadow DOM looks like this:
```
justinfagnani, line 308: I wonder if we should change the name of `<shadow-element>` to make it event more clearly an example custom element, rather than something that sounds like it itself might be related to shadow DOM.

docs/guide/02-writing-templates.md
```diff
@@ -297,8 +325,10 @@ html`
...
 `;
 ``` 
-If the user is logged in, DELETE button is rendered. If the user is not logged in, nothing is rendered inside of `shadow-element`.
-therefore the slot is empty and its fallback content "Sorry, no content available. I am just fallback content" will be rendered.
+
+If the user is logged in, the Delete button is rendered. If the user is not logged in, nothing is rendered inside of `shadow-element`. This means the slot is empty and its fallback content "Sorry, no content available. I am just fallback content" is rendered.
+
+Replacing `nothing` in this example with the empty string causes an empty text node to be rendered inside `shadow-element`, suppressing the fallback content.
 
 **Whitespace creates text nodes.** For the example to work, the text binding inside `<shadow-element>` must be the **entire** contents of `<shadow-element>`. Any whitespace outside of the binding delimiters adds static text nodes to the template, suppressing the fallback content. However, whitespace _inside_ the binding delimiters is fine.
```
justinfagnani, line 333: Can we add an example showing what the bad structure is?

docs/guide/03-styling-templates.md
```diff
@@ -15,6 +15,53 @@ To help with dynamic styling, lit-html provides two directives for manipulating
...
+## Inline styles with styleMap {#stylemap}
+
+You can use the `styleMap` directive to set inline styles on an element in the template.
+
+```js
+import {html} from 'lit-html';
+import {styleMap} from 'lit-html/directives/style-map.js';
+
+const normalStyles = {};
+const highlightStyles = { color: 'white', backgroundColor: 'red'};
```
justinfagnani, line 27: I think the canonical example for this `styleMap` is a single literal with dynamic values. That would remove the ternary below and simplify the template.

docs/guide/03-styling-templates.md
```diff
@@ -15,6 +15,53 @@ To help with dynamic styling, lit-html provides two directives for manipulating
...
+
+// Define a base set of classes for all menu items
+const baseClasses = { 
+  'menu-item': true,
+  // ...
+};
+
+const itemTemplate = (item) => {
+  // Merge in dynamically-generated classes
+  const mergedClasses = Object.assign({active: item.active}, baseClasses);
```
justinfagnani, line 57: I think the merging might be obscuring the basic functionality here. Maybe just add one class, like `selected`, and remove the `Object.assign`?

docs/guide/02-writing-templates.md
arthurevans: I think the key point here is that this is in the middle of the lit-html docs, which mostly don't talk about shadow DOM at all; this one section gets deep into the weeds of shadow DOM and slots, which confused the original reporter.

(If this was in the LitElement docs, I'd agree, since LitEl uses shadow DOM by default, and we talk about it in many places.)

docs/guide/03-styling-templates.md
arthurevans: Sure... Not sure exactly what you mean, though. 

Are you thinking of a literal inside the `myTemplate` function here, like:

```
const styles =  { color: currentColor; background-color: currentBackground};
```

docs/guide/03-styling-templates.md
arthurevans: OK. Like:

```
const classes = {menu-item: true, selected: item.selected}
```

Might be clearer if we expanded it to multi-line:

```
const classes = { 
  // static class
  'menu-item': true,
  // dynamic class
  selected: item.selected,
}
```

... But maybe that's overkill.

docs/guide/02-writing-templates.md
arthurevans: Sounds good.

docs/guide/03-styling-templates.md
justinfagnani: Yeah, though I might do something like move the ternary into one of the expressions to drive home that it's dynamic:

```js
const styles =  {
  color: currentColor;
  background-color: highlight ? 'yellow' : currentBackground;
};
```

docs/guide/03-styling-templates.md
justinfagnani: With `classMap` the static classes don't need to be in the object, they can be part of the HTML, and it's a bit more efficient that way. So I'm not sure about including static classes at all, lest the reader think they have to be put there.

```js
  return html`<div class="menu-item ${classMap(classes)}>Classy text</div>`
```

arthurevans: Oh. Huh. Currently our docs say that you _can't_ do that--that classMap, like styleMap, needs to be the entire attribute value. So we need to fix that, too.

https://lit-html.polymer-project.org/guide/template-reference#classmap

docs/guide/02-writing-templates.md
arthurevans: Came back to this after a couple of days, I think I agree with you. I also didn't really like the tone. So I'm rewriting this a bit.

docs/guide/03-styling-templates.md
justinfagnani: We say that it must be the only binding, but can be combined with static classes

arthurevans: Never mind, reread, says it needs to be the only _binding_, but it can include static content, exactly like this example.

docs/guide/03-styling-templates.md
```diff
@@ -15,6 +15,48 @@ To help with dynamic styling, lit-html provides two directives for manipulating
...
+
+Like `styleMap`, the `classMap` directive lets you set a group of classes based on an object.
+
+```js
+import {html} from 'lit-html';
+import {classMap} from 'lit-html/directives/class-map.js';
+
+const itemTemplate = (item) => {
+  const classes = {selected: item.selected};
+  return html`<div class="menu-item ${classMap(classes}">Classy text</div>`
```
justinfagnani, line 28: Missing `)` after `classes`, and `;` at end of line

arthurevans: D'oh. Saw this after merging. Sending in a followup PR.

PR #1151, opened 2020-04-28
WIP: Add a generic for directive factories to type check against

justinfagnani:
I'm hoping that something like this is sufficient to type-check directives generically.

The idea is directive factories would mark which parameters might be assigned to the property with the generic, and lit-analyzer would check assignability against the property.

@rictic @runem wdyt?

### DIFF
--- a/src/directives/live.ts
+++ b/src/directives/live.ts
@@ -37,7 +37,7 @@ import {AttributePart, BooleanAttributePart, directive, EventPart, NodePart, Pro
  * passed in, or the binding will update every render.
  */
 export const live = directive(
-    (value: unknown) => (part: AttributePart|PropertyPart|
+    <T>(value: T) => (part: AttributePart|PropertyPart|
                          BooleanAttributePart) => {
       let previousValue: unknown;
       if (part instanceof EventPart || part instanceof NodePart) {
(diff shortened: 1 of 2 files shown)


justinfagnani:
Turns out I had to add a type parameter to `DirectiveFn` and a cast in the `live` implementation to get `live('foo')` to correctly infer as `DirectiveFn<string>`.

I _think_ that should be enough to type check.

rictic:
Really interesting!

runem:
Indeed, this looks really interesting, - it would be so nice to be able to type check custom directives!

I just tested this PR against `lit-analyzer` and added support for type checking the generic `DirectiveFn<T>` in [this commit](https://github.com/runem/lit-analyzer/pull/112/commits/d7c8a9e1f2b068deff30bceeaf9c70fa671d1750) :-)

A small thing to consider is that the inferred type is widened, so `myDirective("foo")` becomes the type `DirectiveFn<string>`. This makes type checking against for example "string literal union types" fail unless the user of the directives use `as const` like this: `myDirective("foo" as const)`.

justinfagnani:
Closing in favor of the new lit-html 2.0 directive API?

PR #1153, opened 2020-04-30
Restore trusted types

### LINKED ISSUE #1145: Unrevert Trusted Types Support
We had support for Trusted Types, but it was reverted prior to our most recent release along with some other security hooks that are likely to only be used internally at google.

The other hooks were too obscure, and their cost in bytes and runtime was so high that the benefit to the small number of people that might use it was outweighed by the cost. It still lives on in the security branch, which we will maintain.

However Trusted Types is now releasing in Chrome shortly, and has minimal size and runtime cost, while covering the most critical surface area.

rictic:
Re-adds trusted type support now that it's soon to land in Chrome stable

Size at master:
```
┌──────────────────────────────────────┐
│                                      │
│   Destination: lit-html.bundled.js   │
│   Bundle Size:  9.13 KB              │
│   Minified Size:  9.11 KB            │
│   Gzipped Size:  3.33 KB             │
│   Brotli size: 3.01 KB               │
│                                      │
└──────────────────────────────────────┘
created lit-html.bundled.js in 601ms
    3416
```

Size at c5199ab0f1f85fc93ce3bc2d48a409b63c9360a7:
```
lit-html.js → lit-html.bundled.js...
┌──────────────────────────────────────┐
│                                      │
│   Destination: lit-html.bundled.js   │
│   Bundle Size:  9.37 KB              │
│   Minified Size:  9.35 KB            │
│   Gzipped Size:  3.43 KB             │
│   Brotli size: 3.1 KB                │
│                                      │
└──────────────────────────────────────┘
created lit-html.bundled.js in 598ms
    3518
```

Fixes #1145

### DIFF
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -69,11 +69,33 @@ export class AttributeCommitter {
   protected _getValue(): unknown {
     const strings = this.strings;
     const l = strings.length - 1;
+    const parts = this.parts;
+
+    // If we're assigning an attribute via syntax like:
+    //    attr="${foo}"  or  attr=${foo}
+    // but not
+    //    attr="${foo} ${bar}" or attr="${foo} baz"
+    // then we don't want to coerce the attribute value into one long
+    // string. Instead we want to just return the value itself directly,
+    // so that sanitizeDOMValue can get the actual value rather than
+    // String(value)
+    // The exception is if v is an array, in which case we do want to smash
+    // it together into a string without calling String() on the array.
+    //
+    // This also allows trusted values (when using TrustedTypes) being
+    // assigned to DOM sinks without being stringified in the process.
+    if (l === 1 && strings[0] === '' && strings[1] === '' &&
+        parts[0] !== undefined) {
+      const v = parts[0].value;
+      if (!isIterable(v)) {
+        return v;
+      }
+    }
     let text = '';
 
     for (let i = 0; i < l; i++) {
       text += strings[i];
-      const part = this.parts[i];
+      const part = parts[i];
       if (part !== undefined) {
         const v = part.value;
         if (isPrimitive(v) || !isIterable(v)) {
@@ -93,7 +115,12 @@ export class AttributeCommitter {
   commit(): void {
     if (this.dirty) {
       this.dirty = false;
-      this.element.setAttribute(this.name, this._getValue() as string);
+      let value = this._getValue() as string;
+      if (typeof value === 'symbol') {
+        // Native Symbols throw if they're coerced to string.
+        value = String(value);
+      }
+      this.element.setAttribute(this.name, value);
     }
   }
 }
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -84,10 +84,12 @@ export class AttributeCommitter {
     //
     // This also allows trusted values (when using TrustedTypes) being
     // assigned to DOM sinks without being stringified in the process.
-    if (l === 1 && strings[0] === '' && strings[1] === '' &&
-        parts[0] !== undefined) {
+    if (l === 1 && strings[0] === '' && strings[1] === '') {
       const v = parts[0].value;
-      if (!isIterable(v)) {
+      if (typeof v === 'symbol') {
+        return String(v);
+      }
+      if (typeof v === 'string' || !isIterable(v)) {
         return v;
       }
     }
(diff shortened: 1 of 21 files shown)


src/lib/parts.ts
```diff
@@ -93,7 +115,12 @@ export class AttributeCommitter {
   commit(): void {
     if (this.dirty) {
       this.dirty = false;
-      this.element.setAttribute(this.name, this._getValue() as string);
+      let value = this._getValue() as string;
+      if (typeof value === 'symbol') {
```
justinfagnani, line 119: moving the check for symbol into _getValue() might save a couple of bytes

src/lib/parts.ts
```diff
@@ -69,11 +69,33 @@ export class AttributeCommitter {
...
+    // String(value)
+    // The exception is if v is an array, in which case we do want to smash
+    // it together into a string without calling String() on the array.
+    //
+    // This also allows trusted values (when using TrustedTypes) being
+    // assigned to DOM sinks without being stringified in the process.
+    if (l === 1 && strings[0] === '' && strings[1] === '' &&
+        parts[0] !== undefined) {
+      const v = parts[0].value;
+      if (!isIterable(v)) {
```
justinfagnani, line 90: strings are iterable, so maybe check that too to avoid the loop below?

src/lib/parts.ts
```diff
@@ -69,11 +69,33 @@ export class AttributeCommitter {
...
+    // string. Instead we want to just return the value itself directly,
+    // so that sanitizeDOMValue can get the actual value rather than
+    // String(value)
+    // The exception is if v is an array, in which case we do want to smash
+    // it together into a string without calling String() on the array.
+    //
+    // This also allows trusted values (when using TrustedTypes) being
+    // assigned to DOM sinks without being stringified in the process.
+    if (l === 1 && strings[0] === '' && strings[1] === '' &&
+        parts[0] !== undefined) {
```
justinfagnani, line 88: Every part in the parts array should be defined

src/lib/template-result.ts
```diff
@@ -100,7 +110,17 @@ export class TemplateResult {
 
   getTemplateElement(): HTMLTemplateElement {
     const template = document.createElement('template');
-    template.innerHTML = this.getHTML();
+    let value = this.getHTML();
+    const trustedTypes = window.trustedTypes;
+    if (trustedTypes && !policy) {
+      policy = trustedTypes.createPolicy('lit-html', {createHTML: (s) => s});
```
justinfagnani, line 116: Is creating a policy expensive at all? Why not move this into the `policy` initializer and assume it'll be used very shortly. A few less bytes and one less conditional in a perf-sensitive path.

src/lib/parts.ts
rictic: Removed check

src/lib/parts.ts
rictic: Added strings check

src/lib/parts.ts
rictic: Done

src/lib/template-result.ts
rictic: Moved up and made policy a const, as a hint to the VM that the check for policy being defined can be removed, it's either always defined or always undefined.

src/lib/template-result.ts
```diff
@@ -20,6 +20,21 @@ import {reparentNodes} from './dom.js';
...
+ * That HTML is a developer-authored constant, and is parsed with innerHTML
+ * before any untrusted expressions have been mixed in. Therefor it is
+ * considered safe by construction.
+ */
+const policy = ((): Pick<TrustedTypePolicy, 'createHTML'>|void => {
+  const tt = window.trustedTypes;
+  if (tt !== undefined) {
+    return tt.createPolicy('lit-html', {createHTML: (s) => s});
+  }
+})();
```
justinfagnani, line 36: Consider:

```ts
const policy = trustedTypes && trustedTypes.createPolicy('lit-html', {createHTML: (s) => s});
```

rictic: I like it. Saves another 14 bytes after gzip

rictic: Er, 13 bytes.

Adding `window.` adds only one byte to the gzipped output! It's a common enough substring, but that's still impressive!

justinfagnani: nice :)

rictic:
Brendan released wct-mocha 1.1.0! So now we can run tests with trusted types enabled and disabled

rictic:
Ignore the failing travis run, it's trying to pull from the other PR (now merged with this one) directly, which fails because the other PR is now closed

PR #1154, opened 2020-04-30
Trusted types tests

rictic:
Builds on top of https://github.com/Polymer/lit-html/pull/1153

Needs https://github.com/Polymer/tools/pull/3503 to be released before it will work


rictic:
Ah, I guess we're not testing in Chrome canary currently

PR #1162, opened 2020-05-08
[hydration] Add noChange handling & align w/NodePart:commit

kevinpschaaf:
Integration tests added in https://github.com/PolymerLabs/lit-ssr/pull/39.

### DIFF
--- a/src/lib/hydrate.ts
+++ b/src/lib/hydrate.ts
@@ -135,112 +135,151 @@ export const hydrate =
 
       const walker = document.createTreeWalker(
           container, NodeFilter.SHOW_COMMENT, null, false);
-      let node: Comment|null;
+      let marker: Comment|null;
 
       // Walk the DOM looking for part marker comments
-      while ((node = walker.nextNode() as Comment | null) !== null) {
-        const data = node.data;
-
-        if (data.startsWith('lit-part')) {
-          // This is an NodePart opening marker
-          let value: unknown;
-
-          if (stack.length === 0) {
-            // We don't have a parent part, so this must be the root part
-            if (rootPart !== undefined) {
-              throw new Error('there must be only one root part per container');
-            }
-            rootPart = currentNodePart = new NodePart({
-              templateFactory,
-              ...options,
-            });
-            value = rootValue;
-          } else {
-            const state = stack[stack.length - 1];
-
-            if (state.type === 'template-instance') {
-              // We have a current template instace, so use its template
-              // processor to create parts
-              currentNodePart =
-                  state.instance.processor.handleTextExpression(options);
-              state.instance.__parts.push(currentNodePart);
-              value = state.result.values[state.instancePartIndex++];
-              state.templatePartIndex++;
-            } else if (state.type === 'iterable') {
-              const result = state.iterator.next();
-              if (result.done) {
-                value = undefined;
-                state.done = true;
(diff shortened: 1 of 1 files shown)


src/lib/hydrate.ts
```diff
@@ -209,25 +220,20 @@ export const hydrate =
...
-                type: 'iterable',
-                value,
-                iterator: value[Symbol.iterator](),
-                done: false,
-              });
-              currentNodePart.value = [];
-            } else {
-              stack.push({part: currentNodePart, type: 'leaf'});
-            }
+            // Objects, Nodes, Functions, etc...
```
sorvell, line 234: In `NodePart.commit`, after iterable, there's the `value === nothing` case. Does that need to be handled here?

src/lib/hydrate.ts
```diff
@@ -186,8 +187,18 @@ export const hydrate =
           // the matching marker.
           currentNodePart.startNode = node;
 
-          // Check for a template result digest
-          if (value instanceof TemplateResult) {
+          // The following logic closely follows the NodePart commit() cascade
```
sorvell, line 190: This function is crazy long now. Seems like it could benefit from some factoring. Perhaps we could start with this commit cascade?

On that function, I'd love to see a comment that calls out the ordering:
1. directive
2. noChange
3. primitive
4. TemplateResult
5. Node
6. Iterable
7. nothing
8. Everything else

src/lib/hydrate.ts
```diff
@@ -209,25 +220,20 @@ export const hydrate =
               // need to stop hydrating this subtree? Clear it? Add tests.
               throw new Error('unimplemented');
             }
+          } else if (isIterable(value)) {
```
sorvell, line 223: Nodes are ordered before iterables in `NodePart.commit`. Does that matter here?

src/lib/hydrate.ts
kevinpschaaf: In hydration we're only initializing the part's `value` (not doing anything with the actual rendered content), so no, `nothing` is treated like all the other fallback values, in that its `part.value` is just the value (`nothing`). It would be extra code to separate that out into another clause to match the NodePart.commit code, so I'm just noting that nothing is handled in the comment.

I did go back and look and notice that the fallback does initialize `part.value` to empty string for `null` and `undefined`; I don't think it really matters since fallback values aren't dirty checked anyway, but went ahead and updated the hydrate code to match.

src/lib/hydrate.ts
kevinpschaaf: We don't currently support creating `Node`s on the server (no full DOM shim), so there's no way for a user to pass them as values, so currently not supporting.  But if we did they would just be lumped in the final else fallback because again, for all those types, all we're doing is just setting `part.value = value`.  Only `noChange`, `TemplateResult`, and `Iterable` are special for this part of the code.

src/lib/hydrate.ts
kevinpschaaf: I factored `hydrate` into:
* `openNodePart()`
* `createAttributeParts()`
* `closeNodePart()`

which are called from a much tighter while loop now in `hydrate`'s treewalk over the comments, which I think is a nice improvement.  

I tried several ways to separate the commit cascade out from the other work of `openNodePart`, but it was pretty unsatisfying.

I added the order comment for review, although I don't personally think this helps

PR #1172, opened 2020-06-03
Add BLM banner

arthurevans:
Staged: https://20200603t165446-dot-polymer-lit-html.uk.r.appspot.com

### DIFF
--- a/docs/css/main.css
+++ b/docs/css/main.css
@@ -212,6 +212,31 @@ main {
   display: none;
 }
 
+.hero.blm {
+  background: #696969;
+  color: #fff;
+}
+
+.hero .blm-banner {
+  position: relative;
+  background: #000;
+  height: 60px;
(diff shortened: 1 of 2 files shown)


daKmoR:
was the lit-html page always this "red"? if so I never noticed 😅
Screenshot 2020-06-04 at 10 27 44

arthurevans:
@daKmoR Thanks for checking. :D I had to double-check to be sure, but yes--it did. Here's the version from before yesterday's updates:

https://20200527t140555-dot-polymer-lit-html.uk.r.appspot.com/

If you look at the lit-element landing page, you'll see it's very blue. It's just the headline color, but it's very noticeable on the landing page because of the way it's set up with blocks of text in `<h3>`.

On the rest of the pages, you'll see that red color for headings and links--but not big blocks of text.

PR #1178, opened 2020-06-26
Fix api doc

### LINKED ISSUE #1177: lit-html API docs are missing information
There's been a regression in API doc generation, comments are no longer showing up.

arthurevans:
Update to the same (beta) version of Typedoc we use for LitElement. 
Fixes #1177. Removes a weird hack we were using to avoid including the license comments in the 
API docs.

Also restores working source code links in the API docs.

Staged at https://fix-api-doc-dot-polymer-lit-html.uk.r.appspot.com/api/index.html


arthurevans:
The `_lit_html_` changes are a result of going from using a plugin to identify individual "modules," which required each sub-module of, say, lit-html to have a manual `@module lit-html` tag added to it.

The beta version of Typedoc's "library mode" takes a set of entry points--in this case, lit-html and shady-render--and treats each as a "library"--coalescing all of the sub-modules that it imports and re-exports symbols from. However, for whatever reason it treats the module names as "lit-html" (as far as I can tell) and then escapes a bunch of characters to end up with _lit_html_. This does result in an unfortunate change to our URLs, but I haven't figured a way around it.

Hmm... Just realized I should actually remove the @module tags from the other files, too, since we don't need them anymore.

arthurevans:
Did some minor cleanup, output unchanged.

arthurevans:
Added redirects to the python server. Also updated staging

 AFAICT, paths on the live lit-html site, like:

https://lit-html.polymer-project.org/api/classes/lit_html.booleanattributepart.html

Should be correctly re-directed if you copy the path portion to the staged version. For example, if you click on this:

https://fix-api-doc-dot-polymer-lit-html.uk.r.appspot.com/api/classes/lit_html.booleanattributepart.html

It should redirect you to the correct page--only now with actual doc comments included.

@justinfagnani PTAL.

PR #1179, opened 2020-06-30
Adds SSR support for unsafeHTML/SVG

kevinpschaaf:
Updates for integration with lit-ssr-hacker-news app:
* Adds SSR support for unsafeHTML/SVG

Associated PRs:
* https://github.com/PolymerLabs/lit-ssr-hacker-news/pull/1
* https://github.com/PolymerLabs/lit-ssr/pull/64

### DIFF
--- a/src/directives/unsafe-html.ts
+++ b/src/directives/unsafe-html.ts
@@ -12,10 +12,10 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {defaultTemplateProcessor} from '../lib/default-template-processor.js';
 import {isPrimitive} from '../lib/parts.js';
 import {directive, NodePart, Part} from '../lit-html.js';
 import {TemplateResult} from '../lit-html.js';
-import { defaultTemplateProcessor } from '../lib/default-template-processor.js';
 
 interface PreviousValue {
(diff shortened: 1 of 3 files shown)


src/directives/unsafe-html.ts
```diff
@@ -39,23 +41,19 @@ export const unsafeHTML = directive((value: unknown) => (part: Part): void => {
...
 
+  let result;
   if (previousValue !== undefined && isPrimitive(value) &&
-      value === previousValue.value && part.value === previousValue.fragment) {
-    return;
+      value === previousValue.value) {
+    result = previousValue.result;
+  } else {
+    // DO NOT COPY THIS CODE. IT IS UNSAFE TO CONSTRUCT TEMPLATE RESULTS THIS
+    // WAY (HENCE THE NAME).
```
justinfagnani, line 52: 😬

We need a big todo/issue to look into how to preserve some of the security properties of client-side rendering. Looks like this implementation will pick that up since it's passing the fake TemplateResult through the normal Part API.

kevinpschaaf: Added issue: https://github.com/PolymerLabs/lit-ssr/issues/67

PR #1186, opened 2020-07-21
Add the badge for bugs count to readme

JosefJezek:
We need minimize bugs count. ;-)


justinfagnani:
Different projects deal with bugs in different ways, and bug report volume depends on usage - so bug report numbers aren't comparable between projects and in many cases 0 open issues does not mean a project is less buggy or better maintained.

I don't think we need this, and even if we did it should be discussed in an issue before making a PR, as requested in our contributing guidelines.

If there are specific bugs you'd like to see addressed, or contribute fixes for, discussing them individually is probably a more effective way to get results than adding a badge.

PR #1193, opened 2020-07-22
Fix trusted types tests.

rictic:
Also fix support for trusted types in the unsafeSVG directive.

Also run trusted types tests by default in `npm test`


src/directives/unsafe-svg.ts
```diff
@@ -48,9 +48,11 @@ export const unsafeSVG = directive((value: unknown) => (part: Part): void => {
   }
 
   const template = document.createElement('template');
-  template.innerHTML = `<svg>${value}</svg>`;
+  const svgElement =
+      document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+  template.content.appendChild(svgElement);
+  svgElement.innerHTML = value as string;
```
justinfagnani, line 54: I'd still use `template.innerHTML` so that the element doesn't change documents:

```suggestion
  template.innerHTML = `<svg></svg>`;
  template.content.firstChild.innerHTML = value as string;
```

rictic: We can't assign a bare string to template.innerHTML, it must go through a policy.

We'd either have to export the policy from template-result (which is a security issue) or create a new policy for unsafeSVG

justinfagnani: ack. Maybe consider `t.content.ownerDocument`? Not a big deal though.

rictic: +1, done

PR #1201, opened 2020-08-13
Update build docs

arthurevans:
- Point to sample project.
- Point to new LitElement build guide.
- Fix ref to open-wc webpack info.
- Assorted other improvements.


docs/guide/08-tools.md
```diff
@@ -125,79 +125,74 @@ Build tools take your code and make it production-ready. Among the things you ma
...
 * lit-html template minification.
+* Polyfills.
+
+For more details on these considerations, see the LitElement [Build for production](https://lit-element.polymer-project.org/guide/build) guide. lit-html has the same requirements as LitElement, except that lit-html requires only the [Template polyfill](#template-polyfill), not the full Web Components polyfills.
 
-#### Transpilation and module transform
+#### Compilation and module transform {#transpilation-and-module-transform}
 
-You build tools need to transpile ES6 features to ES5 for legacy browsers. 
+You build tools need to compile ES6 features to ES5 for legacy browsers.
```
justinfagnani, line 163: You -> Your, but also, I'm having a little difficulty parsing the intent of this sentence. Is it saying that my tools definitely need to compile my code, or that they may need to if I'm targeting legacy browsers?

docs/guide/08-tools.md
```diff
@@ -125,79 +125,74 @@ Build tools take your code and make it production-ready. Among the things you ma
...
-#### Transpilation and module transform
+#### Compilation and module transform {#transpilation-and-module-transform}
 
-You build tools need to transpile ES6 features to ES5 for legacy browsers. 
+You build tools need to compile ES6 features to ES5 for legacy browsers. 
 
 If you're working in TypeScript, the TypeScript compiler can generate different output for different browsers.
 
 * In general, ES6 is faster than the ES5 equivalent, so try to serve ES6 to browsers that support it.
 * TypeScript has slightly buggy template literal support when compiling to ES5, which can hurt performance.
```
justinfagnani, line 168: Can we update this too? There seems to be no actionable advice to the TypeScript section. What we really recommend is targeting ES2017/modules from tsc and doing any other transforms in Rollup/Babel.

TypeScript's issue is non-spec-compliance when compiling to ES5 and a module setting of something other than  `esnext`, `es2015`, etc. And it's not just slightly buggy, it completely breaks lit-html's ability to detect a re-render, so all DOM is recreated all the time. So we should be very adamant here that users not compile their code to ES5/non-modules with TypeScript.

docs/guide/08-tools.md
arthurevans: Mmm. Yep, that should be flipped around. "To support legacy browsers, your build tools need to compile ..."

docs/guide/08-tools.md
arthurevans: Oh, super helpful. Yeah, I'm not sure we captured that in the main LitElement build docs, either. Will update here then check.

docs/guide/08-tools.md
```diff
@@ -122,82 +122,74 @@ Alternately, you can also use the Karma test runner. The Open Web Components rec
...
+* lit-html template minification.
+* Polyfills.
 
-For more on building with Polymer CLI, see [Build for production](https://polymer-library.polymer-project.org/3.0/docs/apps/build-for-production) in the Polymer library docs.
+For more details on these considerations, see the LitElement [Build for production](https://lit-element.polymer-project.org/guide/build) guide. lit-html has the same requirements as LitElement, except that lit-html requires only the [Template polyfill](#template-polyfill), not the full Web Components polyfills.
 
-### Build your project with webpack
+#### Compilation and module transform {#transpilation-and-module-transform}
 
+You build tools need to compile ES6 features to ES5 for legacy browsers. n general, ES6 is faster than the ES5 equivalent, so try to serve ES6 to browsers that support it.
```
justinfagnani, line 163: Did you flip this around yet? "You" still needs to be "your" and there's a missing "I" for "In".

arthurevans: Gah. Fixed a different instance, apparently maimed this one. Sigh.

docs/guide/08-tools.md
```diff
@@ -122,82 +122,74 @@ Alternately, you can also use the Karma test runner. The Open Web Components rec
...
 
-See the Open Web Components default webpack configuration provides a great starting point for building projects that use lit-html. See their [webpack page](https://open-wc.org/building/building-webpack.html#default-configuration) for instructions on getting started. 
+Your build tools need to accept JavaScript modules (also called ES modules) and transform them to another module format, such as SystemJS, if necessary. If you use node-style module specifiers, your build will also need to transform them to browser-ready modules specifiers. 
 
-### Build considerations for other tools {#build-considerations}
+If you're working in TypeScript, the TypeScript compiler, `tsc`, can generate different output for different browsers. However, there are known issues with the compiled output for older browsers. **We recommend compiling TypeScript using the ES2017 target and ES modules (the default module system for the ES2017 target).** Use Babel to transform the compiled output for older browsers.
```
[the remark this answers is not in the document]
arthurevans: Huh. The docs say that ES modules is the default for target==ES6 or newer (https://www.typescriptlang.org/tsconfig#module). But maybe the init manually adds a `module` field, which would override that... I'll just say that you should use ES2017 and ES modules, and include a config fragment.

arthurevans:
Updated and hopefully clean now, PTAL.

docs/guide/08-tools.md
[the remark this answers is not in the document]
justinfagnani: Huh x2. I didn't know that. I'll check it out

PR #1208, opened 2020-08-20
[lit-next] Add Rollup, Terser, and checksize script

justinfagnani:
This sets up a build just to check the bundled size.

Eventually we'll change the setup to publish the bundle.

### DIFF
--- a/packages/lit-html/rollup.config.js
+++ b/packages/lit-html/rollup.config.js
@@ -16,7 +16,7 @@ import filesize from 'rollup-plugin-filesize';
 import { terser } from 'rollup-plugin-terser';
 
 export default {
-  input: 'lit-html.js',
+  input: 'lib/lit-html.js',
   output: {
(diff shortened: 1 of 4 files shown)


packages/lit-html/rollup.config.js
```diff
@@ -30,6 +30,9 @@ export default {
       },
       output: {
         comments: false,
```
arthurevans, line 32: Just an FYI that I've been using the default value for `comments` in my sample Rollup configs; the default tries to retain license comments. This does leave us with a good number of duplicate license comments, and hence a larger code size. But perhaps we should make the two match one way or the other.

justinfagnani: When we publish the bundle I think we'll have to include a license comment, but we'll ensure there's only one, so this shouldn't be a problem.

For 1.x we should have a script that makes sure all our comments are dedupable.

PR #1215, opened 2020-08-29
[lit-next] Add EventPart

justinfagnani:


packages/lit-html/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2020",
+    "target": "es2019",
```
justinfagnani, line 3: This is because rollup-plugin-filesize doesn't support ES2020. It increases the bundle size by a few 10s of bytes, but I think by release we'll be able to change back to ES2020 or ES2021 even (Safari 14 beta is the last major to add logical assignment).

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -750,6 +751,80 @@ export class BooleanAttributePart extends AttributePart {
...
+      );
+    }
+    this.__value = newListener;
+  }
+
+  handleEvent(event: Event) {
+    if (typeof this.__value === 'function') {
+      // TODO (justinfagnani): do we need to default to this.__element?
+      // It'll always be the same as `e.currentTarget`.
+      this.__value.call(this.__eventContext ?? this.__element, event);
```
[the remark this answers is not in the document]
justinfagnani: Just for saving bytes, considering it seems to be redundant to `e.currentTarget`. Since this is only for events added via `@` syntax, we can have our own rules to an extent.

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -563,9 +565,8 @@ export class NodePart {
 
   private __commitText(value: unknown): void {
     const node = this.__startNode.nextSibling;
-    // If `value` isn't already a string, we explicitly convert it here in case
-    // it can't be implicitly converted - i.e. it's a symbol.
-    value = value == null ? '' : value;
+    // If `value` isn't already a string, we explicitly convert it here
+    value = value ??= '';
```
jridgewell, line 569: This should just be 

```suggestion
    value ??= '';
````

File: packages/lit-html/src/lib/lit-html.ts [hunk 4]
```diff
@@ -651,7 +652,7 @@ export class AttributePart {
   __resolveValue(value: unknown, _i: number) {
     // TODO (justinfagnani): invoke directives here, which will need
     // _i to revive directive state
-    return value == null ? '' : value;
+    return (value ??= '');
```
jridgewell, line 655:
```suggestion
    return value ?? ''
```

PR #1216, opened 2020-08-29
[lit-html] Add directives

justinfagnani:
This adds the class-based, stateful directive design.

Currently the `update()` method accepts a part, and we haven't determined what the public API of parts are yet.

Also, nested directives are not support in either attribute or node parts. To support that we'll need to slightly change the way directive state is kept and when values are resolved in order to run directives. I'm holding off on that so we discuss alternatives, such as some kinds of direct composition support.

### DIFF
--- a/packages/lit-html/src/lib/lit-html.ts
+++ b/packages/lit-html/src/lib/lit-html.ts
@@ -176,6 +176,36 @@ export const nothing = {};
  */
 const templateCache = new Map<TemplateStringsArray, Template>();
 
+export type DirectiveClass = {new (part: Part): Directive};
+
+/**
+ * This utility type extracts the signature of a directive class's render()
+ * method so we can use it for the type of the generated directive function.
+ */
+type DirectiveProps<C extends DirectiveClass> = Parameters<
+  InstanceType<C>['render']
+>;
+
+/**
+ * A generated directive function doesn't evaluate the directive, but just
+ * returns a DirectiveResult object that captures the arguments.
+ */
+type DirectiveResult<C extends DirectiveClass = DirectiveClass> = {
+  _$litDirective$: C;
+  values: DirectiveProps<C>;
+};
+
+/**
+ * Creates a user-facing directive function from a Directive class. This
+ * function has the same parameters as the directive's render() method.
+ */
+export const directive = <C extends DirectiveClass>(c: C) => (
+  ...values: DirectiveProps<C>
+): DirectiveResult<C> => ({
+  _$litDirective$: c,
+  values,
+});
+
 export interface RenderOptions {
   /**
    * An object to use as the `this` value for event listeners. It's often
@@ -212,6 +242,13 @@ const walker = d.createTreeWalker(d);
 // Keeping variable declarations and classes together improves minification.
 // Interfaces and type aliases can be interleaved freely.
 //
+
+export abstract class Directive {
+  abstract render(...props: Array<unknown>): unknown;
+  update(_part: Part, props: Array<unknown>): unknown {
+    return this.render(...props);
+  }
+}
 
 class Template {
   private __strings: TemplateStringsArray;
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/lib/lit-html.ts
```diff
@@ -213,6 +243,13 @@ const walker = d.createTreeWalker(d);
 // Interfaces and type aliases can be interleaved freely.
 //
 
+export abstract class Directive {
+  abstract render(...props: Array<unknown>): unknown;
+  update(_part: Part, props: Array<unknown>): unknown {
+    return this.render(...props);
```
kevinpschaaf, line 249: It's a little odd that `update` returns a value; it means that when doing custom update logic you might need to `return super.update(part, props)` (as opposed to just calling `super.update(part, props)`, which isn't super intuitive given the name.

All things equal, it would seem better to keep semantics similar to LitElement by doing this:
```js
update(part, props) {
  part.__setValue(this.render(...props));
}
```
And having `__commitDirective` just call `this.__directive.update(this, value.values))`

kevinpschaaf: I see the pattern used in tests is for update to call/return render directly rather than super'ing.

Oh, and I guess the current factoring of attribute parts doesn't lend well to this... Let's discuss.

justinfagnani: Chatted offline. I think side-effectful update is a bit tricky due to the inconsistent __setValue() API.

kevinpschaaf (approved):
Ok for now, with the understanding that we'll review/refine the Part & Directive API as we evaluate more directives.

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -552,15 +593,25 @@ export class NodePart {
     this.__startNode.parentNode!.insertBefore(node, this.__endNode);
   }
 
+  private __commitDirective(value: DirectiveResult) {
+    const directive = value._$litDirective$;
+    if (this.__directive?.constructor !== directive) {
```
jridgewell, line 598: This `.constructor` access is a bit funky. Is it possible to store the constructor (which is just `directive` here) as `this.__directive`, and store the new instance as `this.__value`? Then we'd know both the class constructor and instance value at the same time.

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -649,10 +701,19 @@ export class AttributePart {
    * @param value the raw input value to normalize
    * @param _i the index in the values array this value was read from
    */
-  __resolveValue(value: unknown, _i: number) {
-    // TODO (justinfagnani): invoke directives here, which will need
-    // _i to revive directive state
-    return (value ??= '');
+  __resolveValue(value: unknown, i: number) {
+    const directiveCtor = (value as DirectiveResult)?._$litDirective$;
```
jridgewell, line 705: Shouldn't this be an `any` cast?

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -649,10 +701,19 @@ export class AttributePart {
...
+    if (directiveCtor !== undefined) {
+      let directive: Directive = (this.__directives ??= [])[i];
```
jridgewell, line 707: This essentially guarantees that the `__directives` array will go into slow mode because of the OOB access the first time. Can we initialize to a correct length to begin with?

packages/lit-html/src/lib/lit-html.ts
justinfagnani: What's funky about the `.constructor` access?

We want to store the value returned by the directive so we can do change detection on it and not require that directives handle that themselves.

packages/lit-html/src/lib/lit-html.ts
justinfagnani: We can initialize to `new Array(this.__strings.length - 1)`. It'd be good to skip the array when `this.__strings === undefined`

justinfagnani: One thing that might complicate this soon is that in order to support nested directives we might have a longer array than the values array, and won't know the length until we see all the directives. I'll leave a TODO for now.

PR #1217, opened 2020-08-30
[lit-html] Add cache directive

justinfagnani:
Also adds NodePart.detach() and NodePart.restore()

### DIFF
--- a/packages/lit-html/src/lib/directives/cache.ts
+++ b/packages/lit-html/src/lib/directives/cache.ts
@@ -1,0 +1,73 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {
+  directive,
+  TemplateResult,
+  NodePart,
+  NodePartState,
+  Directive,
+} from '../lit-html.js';
+
+/**
+ * Enables fast switching between multiple templates by caching the DOM nodes
+ * and TemplateInstances produced by the templates.
+ *
+ * Example:
+ *
+ * ```
+ * let checked = false;
+ *
+ * html`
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/lib/lit-html.ts
```diff
@@ -569,6 +582,28 @@ export class NodePart {
     public options: RenderOptions | undefined
   ) {}
 
+  detach(): NodePartState {
```
sorvell, line 585: Moving this out of the core would be ideal since it's only needed for some directives. 

Maybe the directive base class could have these methods and maybe that could be a separate import.

packages/lit-html/src/lib/directives/cache.ts
```diff
@@ -0,0 +1,73 @@
...
+ * `
+ * ```
+ */
+export const cache = directive(
+  class extends Directive {
+    templateCache = new WeakMap<TemplateStringsArray, NodePartState>();
+    value?: TemplateResult;
+
+    render(v: unknown) {
+      if ((v as TemplateResult)._$litType$ !== undefined) {
```
jridgewell, line 43: `v` can be nullish here, right?
```suggestion
      if ((v as TemplateResult | null)?._$litType$ !== undefined) {
```

packages/lit-html/src/lib/directives/cache.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+      return v;
+    }
+
+    update(part: NodePart, [v]: Parameters<this['render']>) {
+      // If the new value is not a TemplateResult from the same Template as the
+      // previous value, move the nodes from the DOM into the cache.
+      if (
+        this.value !== undefined &&
+        this.value.strings !== (v as TemplateResult).strings
```
jridgewell, line 56: ditto.

packages/lit-html/src/lib/lit-html.ts
justinfagnani: Moving these to a separate module

packages/lit-html/src/lib/directives/cache.ts
```diff
@@ -0,0 +1,73 @@
...
+ * ```
+ */
+export const cache = directive(
+  class extends Directive {
+    templateCache = new WeakMap<TemplateStringsArray, NodePartState>();
+    value?: TemplateResult;
+
+    render(v: unknown) {
+      if ((v as TemplateResult)._$litType$ !== undefined) {
+        this.value = v as TemplateResult;
```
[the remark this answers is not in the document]
justinfagnani: render() is here for SSR. We'll discuss the directive lifecycle more...

justinfagnani:
I'm going to close this and #1220 in favor of a PR that puts part mutation in a helper module.

PR #1219, opened 2020-08-30
[lit-html] Add ifDefined directive

justinfagnani:


packages/lit-html/src/lib/directives/if-defined.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import {nothing} from '../lit-html.js';
+
+/**
+ * For AttributeParts, sets the attribute if the value is defined and removes
+ * the attribute if the value is undefined.
+ *
+ * For other part types, this directive is a no-op.
+ */
+export const ifDefined = (value: unknown) => value ?? nothing;
```
sorvell, line 23: This is not a directive. Do we care?

sorvell (approved):
LGTM. So much better. This works because of https://github.com/Polymer/lit-html/blob/lit-next-ifdefined/packages/lit-html/src/lib/lit-html.ts#L792-L797

packages/lit-html/src/lib/directives/if-defined.ts
justinfagnani: It's here mainly for back compat

PR #1220, opened 2020-08-30
[lit-html] Add iterable support to NodePart

justinfagnani:
This includes the new `insert()` and `remove` special cases and a test for swapping list items with `NodePart.detach()/restore()`


packages/lit-html/src/lib/lit-html.ts
```diff
@@ -690,6 +715,94 @@ export class NodePart {
     }
   }
 
+  private __commitIterable(value: Iterable<unknown>): void {
```
sorvell, line 718: There's a lot of code here and it makes me want to step back and ask questions that might be breaking but be more satisfying:
* should we just do the dumbest thing possible (clear and re-render if the value is actually different) here and rely on `repeat` to do anything better?
* should we not support iterables other than with repeat?
* or should we suck it up and just put the full repeat here?

justinfagnani:
Closing in favor of #1250

PR #1228, opened 2020-09-01
[lit-next] Run tests via playwright on chromium, webkit, and firefox

justinfagnani:


packages/lit-html/package.json
```diff
@@ -36,9 +36,11 @@
     "@esm-bundle/chai": "^4.1.5",
     "@types/mocha": "^8.0.3",
     "@web/test-runner": "^0.7.11",
+    "@web/test-runner-playwright": "^0.5.4",
+    "@web/test-runner-saucelabs": "0.0.1",
```
kevinpschaaf, line 40: Need `saucelabs` yet?

justinfagnani: oops!

PR #1230, opened 2020-09-01
[lit-element] Changes `requestUpdate` to no longer return a promise

sorvell:
...and removes `requestUpdateInternal`.  Previously users customizing the property setter would call `requestUpdateInternal` and now they should call `requestUpdate` instead. It is identical to the old method.

### DIFF
--- a/packages/lit-element/src/lib/updating-element.ts
+++ b/packages/lit-element/src/lib/updating-element.ts
@@ -690,7 +690,7 @@ export abstract class UpdatingElement extends HTMLElement {
   /**
    * Requests an update which is processed asynchronously. This should
    * be called when an element should update based on some state not triggered
-   * by setting a property. In this case, pass no arguments. It should also be
+   * by setting a reactive property. In this case, pass no arguments. It should also be
    * called when manually implementing a property setter. In this case, pass the
    * property `name` and `oldValue` to ensure that any configured property
    * options are honored. Returns the `updateComplete` Promise which is resolved
--- a/packages/lit-element/src/lib/updating-element.ts
+++ b/packages/lit-element/src/lib/updating-element.ts
@@ -380,7 +380,7 @@ export abstract class UpdatingElement extends HTMLElement {
           name as string
         ];
         ((this as {}) as {[key: string]: unknown})[key as string] = value;
-        ((this as unknown) as UpdatingElement).requestUpdateInternal(
+        ((this as unknown) as UpdatingElement).requestUpdate(
           name,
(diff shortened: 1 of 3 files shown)


packages/lit-element/src/lib/updating-element.ts
```diff
@@ -688,11 +688,19 @@ export abstract class UpdatingElement extends HTMLElement {
   }
 
   /**
-   * This protected version of `requestUpdate` does not access or return the
-   * `updateComplete` promise. This promise can be overridden and is therefore
-   * not free to access.
+   * Requests an update which is processed asynchronously. This should
+   * be called when an element should update based on some state not triggered
+   * by setting a property. In this case, pass no arguments. It should also be
```
justinfagnani, line 693:
```suggestion
   * by setting a reactive property. In this case, pass no arguments. It should also be
```

I've started to standardize on calling our special properties "reactive properties". We had been calling them "declared properties", but that doesn't seem to distinguish them correctly to me. @arthurevans has been considering standardizing across docs I think.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -735,24 +743,6 @@ export abstract class UpdatingElement extends HTMLElement {
...
-   * options are honored. Returns the `updateComplete` Promise which is resolved
-   * when the update completes.
-   *
-   * @param name {PropertyKey} (optional) name of requesting property
-   * @param oldValue {any} (optional) old value of requesting property
-   * @returns {Promise} A Promise that is resolved when the update completes.
-   */
-  requestUpdate(name?: PropertyKey, oldValue?: unknown) {
-    this.requestUpdateInternal(name, oldValue);
-    return this.updateComplete;
```
kevinpschaaf, line 753: Just a thought, but since this is a pretty silent breaking change (i.e. awaiting requestUpdate won't error), we _could_ add a devmode-only thing where we add a return like e.g.
```js
if (DEVMODE)
  return { 
    then() { 
      throw new Error(`In LitElement v3.0+, requestUpdate should not be awaited; instead await updateComplete`);
    }
  }
}
```

Although I guess you do get a TS warning: `'await' has no effect on the type of this expression`.

Anyway, maybe worth a TODO? Or maybe we should start an issue listing DEVMODE warnings?

justinfagnani: This would be good for non-TS users, and for non-await users.

I wonder if throwing in then() or returning a rejected Promise is better?

PR #1232, opened 2020-09-02
[lit-html] Adds element bindings and ref() directive

### LINKED ISSUE #923: Add syntax for spreading attributes, properties, and event listeners

justinfagnani:
This doesn't fully fix #923, but a spread directive can be built on top of this.


packages/lit-html/src/lib/directives/ref.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  render(_ref: Ref) {
+    return nothing;
+  }
+
+  update(part: SpreadPart, [ref]: Parameters<this['render']>) {
+    if (ref.value !== part.__element) {
+      ref.set(part.__element);
+    }
+    return nothing;
```
kevinpschaaf, line 37: Pattern feedback: It's odd to me to have a `render` method but never actually call it from `update`.

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -888,6 +917,58 @@ export class EventPart extends AttributePart {
...
+    const directive = (value as DirectiveResult)._$litDirective$;
+    if (directive !== undefined) {
+      if (this.__directive?.constructor !== directive) {
+        this.__directive = new directive(this);
+        // TODO (justinfagnani): clear?
+      }
+      value = this.__directive.update(this, (value as DirectiveResult).values);
+    }
+    if (value === nothing) {
+      // ?
```
kevinpschaaf, line 936: Should maybe remove the previous value's bound attributes (if value was of type `AttributesTemplate`)? Should probably do that regardless, if the type of the previous value didn't match new value. This is the tricky part.

kevinpschaaf (commented):
So, we need to decide what we're going to do overallwith SpreadPart/AttributeTemplate. To me, the only sane thing is to factor NodePart/Template(/Instance) so that spread can reuse the code related to parsing, creating, and committing AttributePart/EventPart/PropertyParts.

I don't think we want a dichotomy of syntax support or duplication of code here. Are there other approaches you're considering?

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -888,6 +917,58 @@ export class EventPart extends AttributePart {
...
 
+export class SpreadPart {
+  __value: unknown;
+  __directive?: Directive;
+
+  constructor(public __element: Element) {}
+
+  __setValue(value: unknown): void {
+    const directive = (value as DirectiveResult)._$litDirective$;
+    if (directive !== undefined) {
```
sorvell, line 928: It's unfortunate that each part type has to deal with directives. Can this be centralized?

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -888,6 +917,58 @@ export class EventPart extends AttributePart {
...
+        let value = strings[0];
+        const size = strings.length - 1;
+        for (let i = 0; i < size; i++) {
+          const v = values[from++];
+          value +=
+            (typeof v === 'string' ? v : v == null ? '' : String(v)) +
+            strings[i + 1];
+        }
+        // TODO: dirty-check
+        element.setAttribute(name, value as string);
```
sorvell, line 966: What about property and event bindings and values that are not strings? This feature doesn't seem worth it unless you can bind what you normally can in attribute position.

sorvell (changes requested):

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Sets the value of a Ref object to the element it's bound to.
+ * 
+ * A Ref acts as a container for a reference to an element. The ref
+ * directive sets the value of the Ref obejct during rendering.
+ * 
+ * @example
+ * 
+ *     const inputRef = new Ref();
+ *     render(html`<input ${inputRef}>`, container);
+ *     input.value.focus();
```
arthurevans, line 51:
```suggestion
 *     inputRef.value.focus();
```

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Sets the value of a Ref object to the element it's bound to.
+ * 
+ * A Ref acts as a container for a reference to an element. The ref
+ * directive sets the value of the Ref obejct during rendering.
+ * 
+ * @example
+ * 
+ *     const inputRef = new Ref();
+ *     render(html`<input ${inputRef}>`, container);
```
arthurevans, line 50:
```suggestion
 *     render(html`<input ${ref(inputRef)}>`, container);
```

packages/lit-html/src/lib/directives/ref.ts
justinfagnani: Right now `render()` is needed to generate the directive function signature. We can try to update the types to use `update()` if `render()` isn't defined.

packages/lit-html/src/lib/lit-html.ts
justinfagnani: N/A since removing attr templates

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -451,6 +459,24 @@ class Template {
   }
 }
 
+class AttributesTemplate {
+  __attrs: Array<{name: string; strings: Array<string>}> = [];
+  constructor(protected __strings: TemplateStringsArray) {
+    const el = d.createElement('template');
+    el.innerHTML = `<div ${__strings.join(marker)}></div>`;
+    console.log(el.innerHTML);
+    for (const {name, value} of el.content.firstElementChild!.attributes) {
```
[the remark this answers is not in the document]
justinfagnani: N/A

packages/lit-html/src/lib/lit-html.ts
justinfagnani: Kevin has part of this in #1426

packages/lit-html/src/lib/lit-html.ts
justinfagnani: N/A

packages/lit-html/src/lib/directives/ref.ts
```diff
@@ -0,0 +1,53 @@
...
+
+class RefDirective extends Directive {
+
+  render(_ref: Ref) {
+    return nothing;
+  }
+
+  update(part: SpreadPart, [ref]: Parameters<this['render']>) {
+    if (ref.value !== part.__element) {
+      ref.set(part.__element);
```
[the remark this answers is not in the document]
justinfagnani: I think that makes common usage a little more cumbersome. You'll always need a way to read a ref, so you need some container.

For a common case of linking two elements:

```ts
const fooRef = new Ref();
render(html`
  <x-foo ${ref(fooRef)}></x-foo>
  <x-bar .foo=${fooRef.value}></x-bar>
`, container);
```

you'd have to do:
```ts
const foo;
const fooRef (e) => fooRefValue = e;
render(html`
  <x-foo ${ref(fooRef)}></x-foo>
  <x-bar .foo=${foo}></x-bar>
`, container);
```

Not hugely different, but enough I think. I'm also looking forward towards possibly having a RefController that can trigger a re-render if the element changes. This would have the same `set()`/`value` API. React can take both Ref objects and callbacks, btw. We can discuss.

packages/lit-html/src/directives/ref.ts
justinfagnani: done

packages/lit-html/src/directives/ref.ts
justinfagnani: done

sorvell (changes requested):

sorvell:
Notes from offline chat:

- Right now the template cache is only keyed on the static strings in the template and this means the following type of usage is not supported:

  ```
  inputTemplate(type) => html`<input type=${unsafeStatic(type)}>` // always returns first call result
  ```
  We do want to support this use case. To do so, the static values must be included in the template key.

- To avoid confusion between `ref` and `Ref`, we'll change `new Ref()` to `createRef()`.
- To avoid leaking,`ref` should become a disconnectable directive and clear its value when disconnected. We *might* want to add an option to avoid disposing of the ref.
- We want to support easily binding a ref node into a property of another element in the template. This works today only if the accepting element is after the node where the ref is established in the template. We'd like to support this for elements before the ref in the template. We'll probably tackle this in a subsequent PR but it might be something like `forwardRef(fooRef)` and `forwardRef` can know when `fooRef` is set and either set its value or trigger an update on an enclosing LitElement (via controller integration).

packages/lit-html/src/test/directives/ref_test.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  test('sets a ref', () => {
+    const divRef = new Ref();
+    render(html`<div ${ref(divRef)}></div>`, container);
+    const div = container.firstElementChild;
+    assert.equal(divRef.value, div);
+  });
+
+  test.only('two refs', () => {
```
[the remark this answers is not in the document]
justinfagnani: done

packages/lit-html/src/lit-html.ts
```diff
@@ -1321,6 +1328,22 @@ export class EventPart extends AttributePart {
   }
 }
 
+export class ElementPart {
+  readonly type = ELEMENT_PART;
+  private _directive?: Directive;
+
+  constructor(public element: Element) {}
+
+  _setValue(value: unknown): void {
```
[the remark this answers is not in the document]
justinfagnani: Possibly with the factoring out of resolveDirective, though this is more streamlined since we only accept directives here.

packages/lit-html/src/lit-html.ts
```diff
@@ -541,22 +541,26 @@ const getTemplateHtml = (
       );
     }
 
-    // If we're in text position, and not in a raw text element
-    // (regex === textEndRegex), we insert a comment marker. Otherwise, we
-    // insert a plain maker. If we have a attrNameEndIndex, it means we need
-    // to rewrite the attribute name to add a bound attribute suffix.
+    // We have four cases:
```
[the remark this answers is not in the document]
justinfagnani: (3) had a sub-case. I separated it out

packages/lit-html/src/lit-html.ts
```diff
@@ -541,22 +541,26 @@ const getTemplateHtml = (
...
+    //  1. We're in text position, and not in a raw text element
+    //     (regex === textEndRegex): insert a comment marker.
+    //  2. We have a non-negtive attrNameEndIndex which means we need to
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+class RefDirective extends Directive {
+  render(_ref: Ref) {
+    return nothing;
+  }
+
+  update(part: ElementPart, [ref]: Parameters<this['render']>) {
+    if (ref.value !== part.element) {
```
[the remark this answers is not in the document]
justinfagnani: it's not... this isn't observable yet, but when we add `forwardingRef`, we'll need something like this.

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+import {Directive, directive, nothing, ElementPart} from '../lit-html.js';
+
+/**
+ * A container for a reference to an element.
+ */
+export class Ref<T = Element> {
+  value?: T;
+
+  set(v: T): void {
```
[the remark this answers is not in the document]
justinfagnani: When the set becomes observable with `forwardingRef` (or whatever it's called) I want the method to indicate that this can cause work.

packages/lit-html/src/test/directives/ref_test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const divRef1 = new Ref();
+    const divRef2 = new Ref();
+    render(html`<div ${ref(divRef1)} ${ref(divRef2)}></div>`, container);
+    const div = container.firstElementChild;
+    assert.equal(divRef1.value, div);
+    assert.equal(divRef2.value, div);
+  });
+});
```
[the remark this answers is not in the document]
justinfagnani: Added a conditional rendering test which uses a ref on more than one element, but only one at a time.

We need to add other tests when we get disconnectable directives

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,70 @@
...
+    return nothing;
+  }
+}
+
+/**
+ * Sets the value of a Ref object or ref callback to the element it's bound
+ * to.
+ *
+ * A Ref object acts as a container for a reference to an element. A ref
+ * callbacks is a function that takes an element as its only argument.
```
sorvell, line 55:
```suggestion
 * callback is a function that takes an element as its only argument.
```

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,70 @@
...
+  render(_ref: RefOrCallback) {
+    return nothing;
+  }
+
+  update(part: ElementPart, [ref]: Parameters<this['render']>) {
+    if (typeof ref === 'function') {
+      if (this._element !== part.element) {
+        ref((this._element = part.element));
+      }
+    } else if (ref.value !== part.element) {
```
sorvell, line 43: If you pass a different ref object, it'll always be filled, but if you pass a different ref callback it won't. While a corner case, that's unexpected. Seems like we should either always call the ref function or store and dirty check it.

packages/lit-html/src/lit-html.ts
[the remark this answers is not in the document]
sorvell: ok

packages/lit-html/src/test/directives/ref_test.ts
```diff
@@ -0,0 +1,86 @@
...
+    go(false);
+    queriedEl = container.firstElementChild;
+    assert.equal(queriedEl?.tagName, 'SPAN');
+    assert.equal(el, queriedEl);
+    assert.equal(callCount, 2);
+
+    // TODO (justinfagnani): when ref() is disconnectable, should the
+    // callback be called in this case? Or only called with the new element?
+  });
+
```
sorvell, line 77: Add tests for changing the ref object/callback passed to `ref`, or let's specifically decide not to support that.

packages/lit-html/src/directives/ref.ts
sorvell: Suggest we decide whether or not to support changing what you pass to `ref` and handling the callback/object cases the same.

sorvell (changes requested):

kevinpschaaf (commented):
Code looks good with exception of handling Ref callback changing correctly as Steve noted.

Let's also either hold this PR or add issues for integrating with these pending changes:
* Use `resolveDirective` helper in `ElementPart._setValue` (once #1426 lands)
* Make `ref` a `DisconnectableDirective` to handle clearing the box (once #1432 lands)

packages/lit-html/src/directives/ref.ts
justinfagnani: Updated to track the identity of the Ref or callback

packages/lit-html/src/directives/ref.ts
justinfagnani: fixed

packages/lit-html/src/test/directives/ref_test.ts
justinfagnani: done

sorvell:
Do we need to specifically add a forwardRef or is it just adding tests that bind the ref down? And if it's the latter, let's just add those tests?

sorvell (changes requested):
If forwarding refs just needs tests, let's add those?

Let's also merge the other directive stuff before merging this.

packages/lit-html/src/test/directives/ref_test.ts
```diff
@@ -154,4 +149,47 @@ suite('ref', () => {
     assert.equal(divRef1.value, div);
     assert.equal(divRef2.value, div);
   });
+
+  test('two ref callbacks alternating', () => {
```
sorvell, line 153: Add a test to verify the refs are always set in tree order:

```
html`<div id="first" ${ref(myRef)}></div><div id="next" ${ref(myRef)}>${html`<span id="last" ${ref(myRef)}`}</div>`

File: packages/lit-html/src/directives/ref.ts (hunk 13 above)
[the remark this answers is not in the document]
Comment (sorvell) [reply packages/lit-html/src/directives/ref.ts]: We elected not to add `fowardingRef` since this can be covered by using a callback.

File: packages/lit-html/src/directives/ref.ts [hunk 20]
```diff
@@ -0,0 +1,137 @@
...
+   */
+  value?: T;
+
+  /**
+   * Updates the element value of this Ref. Users should generally not need to
+   * call this function; it will be called automatically when passed to the
+   * `ref()` directive.
+   * @param element
+   */
+  set(element: T | undefined): void {
```
sorvell, line 35: Not sure why we have this if we're not doing `fowardingRef`.

packages/lit-html/src/lit-html.ts
```diff
@@ -1439,6 +1454,33 @@ export class EventPart extends AttributePart {
...
+  /** @internal */
+  _$disconnetableChildren?: Set<Disconnectable> = undefined;
+  /** @internal */
+  _setDirectiveConnected?: (
+    directive: Directive | undefined,
+    isConnected: boolean,
+    removeFromParent?: boolean
+  ) => void = undefined;
+
+  constructor(public element: Element, parent: Disconnectable) {
```
sorvell, line 1475: Can the actual code here be in the `Part` base class to be more DRY?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
kevinpschaaf: I have a stash where I tried making `ElementPart` a base class to `AttributePart` (since it's basically a superset of it), but it still needs a separate class, `_$setValue` implementation, and `AttributePart` then needs to super to it, so the checksize didn't change meaningfully (to the number of digits it shows).

sorvell: ok

packages/lit-html/src/directives/ref.ts
justinfagnani: My thought was we could eventually make Ref observable to trigger re-rendering when you forwarded it. We can do that later though, and with a setter.

packages/lit-html/src/test/directives/ref_test.ts
kevinpschaaf: Good call. This is a weird usage of ref because it's last-in-wins (aka there's no point in 'first' and 'next' refs since they'll always be overwritten), but we weren't calling callbacks correctly in this setup (if `myRef` was a callback, it should call all 3 in order each render I think...?). Added tests for that.

packages/lit-html/src/directives/ref.ts
kevinpschaaf: Removed for now and made `readonly`.

kevinpschaaf:
@justinfagnani @sorvell Updated based on feedback, PTAL

sorvell (approved):
LGTM

packages/lit-html/src/directives/ref.ts
```diff
@@ -0,0 +1,133 @@
...
+      this._ref = ref;
+      this._updateRefValue((this._element = part.element));
+    }
+    return nothing;
+  }
+
+  private _updateRefValue(element: Element | undefined) {
+    if (typeof this._ref === 'function') {
+      // If the current ref was called with a previous value,
+      // call with `undefined`
```
justinfagnani, line 71: I'd leave a little comment why we do this. There are cases where we can't _not_ call with undefined when the ref changes and this is to normalize all the cases, yeah?

justinfagnani (commented):
LGTM

packages/lit-html/src/directives/ref.ts
kevinpschaaf: Exactly. I'll clarify.

PR #1236, opened 2020-09-03
[lit-html] Add unsafeHTML and unsafeSVG directives

justinfagnani:


packages/lit-html/src/lib/directives/unsafe-html.ts
```diff
@@ -0,0 +1,73 @@
...
+        strings,
+        values: [],
+      });
+    }
+  }
+);
+
+// https://tc39.github.io/ecma262/#sec-typeof-operator
+type Primitive = null | undefined | boolean | number | string | symbol | bigint;
+const isPrimitive = (value: unknown): value is Primitive =>
```
jridgewell, line 72: Nit: Can this not be imported from `../lit-html.js`?

packages/lit-html/src/lib/directives/unsafe-html.ts
```diff
@@ -0,0 +1,73 @@
...
+        return value;
+      }
+      if (isPrimitive(value) && value === this.value) {
+        return this.templateResult;
+      }
+      this.value = value;
+      const strings = ([String(value)] as unknown) as TemplateStringsArray;
+      (strings as any).raw = strings;
+      return (this.templateResult = {
+        _$litType$: 1,
```
jridgewell, line 62: Can we import `HTML_RESULT` from `../lit-html.js`?  Or create a local constant so this is self descriptive?

jridgewell (approved):

packages/lit-html/src/lib/directives/unsafe-html.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    constructor(part: Part) {
+      super();
+      // TODO (justinfagnani): We probably don't want to use instanceof
+      if (!(part instanceof NodePart)) {
+        throw new Error('unsafeHTML can only be used in text bindings');
+      }
+    }
+
+    render(value: unknown) {
```
sorvell, line 46: What happens if the value is a TemplateResult? Can we add a test for that?

sorvell: Should we throw? Or throw if values are passed?

sorvell (changes requested):

packages/lit-html/src/lib/directives/unsafe-html.ts
kevinpschaaf: Seems like we should only allow `string`, frankly. I don't see a reason for passing `nothing` or `noChange` (or any other value type, even non-string primitives) as an argument to the directive; the directive turns strings into HTML; if you want something else, you should do that using conditionals on the outside of the directive. Less code and very hard to imagine this would break someone.

kevinpschaaf (changes requested):

packages/lit-html/src/lib/directives/unsafe-html.ts
justinfagnani: Defined locally. Definitely don't want to encourage this.

packages/lit-html/src/lib/directives/unsafe-html.ts
justinfagnani: Removed in favor of only allowing strings. I wouldn't really want to make this part of lit-html's public API either.

packages/lit-html/src/lib/directives/unsafe-html.ts
justinfagnani: Updated to throw on non-strings on the theory that this is usually only used for strings already, so it isn't much of a breaking change. We also don't even import this directive internally.

justinfagnani:
PTAL

jridgewell (approved):

sorvell (approved):

packages/lit-html/src/directives/unsafe-html.ts
```diff
@@ -0,0 +1,86 @@
...
+        `${
+          (this.constructor as typeof UnsafeHTML).directiveName
+        }() called with a non-string value`
+      );
+    }
+    if (value === this.value) {
+      return this.templateResult;
+    }
+    this.value = value;
+    const strings = ([String(value)] as unknown) as TemplateStringsArray;
```
kevinpschaaf, line 65: We're coercing the value to a String already; this seems fine, so I'd just remove the value type check and error. I guess do one or the other...

kevinpschaaf (approved):
LGTM with small nit to consider.

packages/lit-html/src/directives/unsafe-html.ts
justinfagnani: done

PR #1237, opened 2020-09-03
[lit-element] Refactor LitElement

sorvell:
* removes all polyfill code and places in `lit-element-polyfill` which is an in place patch and must be loaded separately (preserves support for styling shadowRoot without adoptedStylesheets).
* Includes caching of CSSResult (adopted from https://github.com/Polymer/lit-element/pull/952)
* `getStyles` is now the main entry point for gathering element styles and can be overridden to customize. It is called once per class in `finalize`.
* `adoptStyles` is now always called, previously was only called if renderRoot was a shadowRoot.


packages/lit-element/src/lib/css-tag.ts
```diff
@@ -31,24 +29,22 @@ export class CSSResult {
...
+    // Note, if `supportsAdoptingStyleSheets` is true then we assume
+    // CSSStyleSheet is constructable.
+    if (supportsAdoptingStyleSheets) {
+      sheet = new CSSStyleSheet();
+      sheet.replaceSync(this.cssText);
+    } else {
+      sheet = null;
     }
-    return this._styleSheet;
+    return sheet;
```
justinfagnani, line 47: If the return type changes to `undefined`, then you can just handle the `supportsAdoptingStyleSheets` case and save some bytes

```ts
if (supportsAdoptingStyleSheets) {
  const sheet = new CSSStyleSheet();
  sheet.replaceSync(this.cssText);
  return sheet;
}
```

packages/lit-element/src/lib/css-tag.ts
```diff
@@ -80,6 +76,8 @@ const textFromCSSResult = (value: CSSResult | number) => {
   }
 };
 
+const cssResultCache = new Map<string, CSSResult>();
```
justinfagnani, line 79: WeakMap?

packages/lit-element/src/lib/css-tag.ts
```diff
@@ -94,5 +92,12 @@ export const css = (
     (acc, v, idx) => acc + textFromCSSResult(v) + strings[idx + 1],
     strings[0]
   );
-  return new CSSResult(cssText, constructionToken);
+  let result = cssResultCache.get(cssText);
+  if (!result) {
```
justinfagnani, line 96:
```suggestion
  if (result === undefined) {
```

to avoid ToBoolean conversions

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -126,58 +125,42 @@ export class LitElement extends UpdatingElement {
    * using the [[`css`]] tag function or via constructible stylesheets.
    */
   static styles?: CSSResultOrNative | CSSResultArray;
-
-  private static _styles: Array<CSSResultOrNative | CSSResult> | undefined;
+  private static elementStyles?: CSSResultArray;
```
justinfagnani, line 128: Name private variables with an underscore? Or double underscore so we can minify it?

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -126,58 +125,42 @@ export class LitElement extends UpdatingElement {
    * using the [[`css`]] tag function or via constructible stylesheets.
    */
   static styles?: CSSResultOrNative | CSSResultArray;
```
justinfagnani, line 127: One problem with the current typings is that is messes up subclassing.

```ts
class A extends LitElement {
  static styles = css`...`;
}

class B extends A {
  static styles = [A.styles, css`...`]; // Error CSSResult[] not assignable to CSSResult
}
```

We could fix this by making `css`'s return type be the same as `static styles`.

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -126,58 +125,42 @@ export class LitElement extends UpdatingElement {
...
       const addStyles = (
-        styles: CSSResultArray,
+        stylesList: CSSResultArray,
         set: Set<CSSResultOrNative>
       ): Set<CSSResultOrNative> =>
-        styles.reduceRight(
+        stylesList.reduceRight(
           (set: Set<CSSResultOrNative>, s) =>
             // Note: On IE set.add() does not return the set
             Array.isArray(s) ? addStyles(s, set) : (set.add(s), set),
```
justinfagnani, line 155: Remove the IE workaround? We can patch `Set.add`

justinfagnani, line 155: There are couple of `Array.isArray` accesses. Maybe tear-off the function so it's minifiable and so there's no property lookup on `Array` for it.

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -126,58 +125,42 @@ export class LitElement extends UpdatingElement {
...
+  ): CSSResultArray {
+    const elementStyles = [];
+    if (Array.isArray(styles)) {
       const addStyles = (
-        styles: CSSResultArray,
+        stylesList: CSSResultArray,
         set: Set<CSSResultOrNative>
       ): Set<CSSResultOrNative> =>
-        styles.reduceRight(
+        stylesList.reduceRight(
```
justinfagnani, line 152: Can you use `new Set(stylesList.flat(Infinity))`

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -240,54 +222,39 @@ export class LitElement extends UpdatingElement {
...
-    if (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow) {
-      window.ShadyCSS.ScopingShim!.prepareAdoptedCssText(
-        styles.map((s) => (s as CSSResult).cssText),
-        this.localName
-      );
-    } else if (supportsAdoptingStyleSheets) {
+    if (supportsAdoptingStyleSheets) {
       (this.renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) =>
-        s instanceof CSSStyleSheet ? s : s.styleSheet!
+        s instanceof CSSStyleSheet ? s : (s as CSSResult).styleSheet!
```
justinfagnani, line 242: Is the `CSSResult.styleSheet` only called when guarded by a `if (supportsAdoptingStyleSheets)` check? If so, you can remove the check in the `styleSheet` getter.

justinfagnani, line 242: I could consider a for loop over `.map()`

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -422,13 +422,17 @@ export abstract class UpdatingElement extends HTMLElement {
...
+  }
+
   /**
    * Creates property accessors for registered properties and ensures
    * any superclasses are also finalized.
    * @nocollapse
    */
   protected static finalize() {
-    if (this.hasOwnProperty(finalized)) {
+    if (this.hasFinalized) {
```
justinfagnani, line 435: +1

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -422,13 +422,17 @@ export abstract class UpdatingElement extends HTMLElement {
     );
   }
 
+  protected static get hasFinalized() {
+    return this.hasOwnProperty(finalized);
```
justinfagnani, line 426: How about:

```ts
this[finalized] = false;
```

The we shouldn't need hasOwnProperty checks?

justinfagnani:
How about removing the check for `getOwnPropertySymbols`?

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -216,16 +204,10 @@ export class LitElement extends UpdatingElement {
    */
   protected initialize() {
```
LarsDenBakker, line 205: I've seen people run into issues because they implemented a method called `initialize`. Would you consider using a different name for this and the static `finalize`? It's a shame for lit-element to take up such common names.

sorvell: Good suggestion. Will handle separately via https://github.com/Polymer/lit-html/issues/1239.

packages/lit-element/src/lib/css-tag.ts
sorvell: Done.

packages/lit-element/src/lib/css-tag.ts
sorvell: The key is a string.

packages/lit-element/src/lib/css-tag.ts
sorvell: Done (more bytes).

packages/lit-element/src/lib/lit-element.ts
sorvell: Done.

packages/lit-element/src/lib/lit-element.ts
sorvell: It's a trade off. If we do that users of `CSSResult` can't really use its API without casting. I guess that's mostly internal and tests.

Instead, we could require an array? It's a breaking change but very mechanical to fix. Wdyt?

packages/lit-element/src/lib/lit-element.ts
sorvell: Moot after simplification.

sorvell: Moot after simplification.

packages/lit-element/src/lib/lit-element.ts
sorvell: Sheesh, yes I believe so. We need to reverse the flattened array so that the set preserves the last instance. I believe we didn't do this initially because IE doesn't have Array.flat.

packages/lit-element/src/lib/lit-element.ts
sorvell: Done.

sorvell: Why? We have to allocate a new Array anyway.

packages/lit-element/src/lib/updating-element.ts
sorvell: Finalize has to be done on each class in the prototype chain. This is why there's an own property check.

packages/lit-element/src/lib/lit-element.ts
justinfagnani: Yeah, I'm not sure all evergreen browsers had it then and this was smaller than a polyfill. But now I think we can go for it.

packages/lit-element/src/lib/lit-element.ts
justinfagnani: I don't like requiring an array. We should try to drive down as much boilerplate as possible, symbol-by-symbol.

packages/lit-element/src/lib/css-tag.ts
justinfagnani: I know it's a few more bytes, but done repeatedly it compresses well, and it generates fewer byte codes. That won't matter here, but maybe in in hot paths like hasChanged. I like to be consistent.

packages/lit-element/src/lib/lit-element.ts
sorvell: Let's handle this in a separate PR.

packages/lit-element/src/lib/css-tag.ts
```diff
@@ -31,24 +29,20 @@ export class CSSResult {
...
-      // Note, if `supportsAdoptingStyleSheets` is true then we assume
-      // CSSStyleSheet is constructable.
-      if (supportsAdoptingStyleSheets) {
-        this._styleSheet = new CSSStyleSheet();
-        this._styleSheet.replaceSync(this.cssText);
-      } else {
-        this._styleSheet = null;
-      }
+  get styleSheet(): CSSStyleSheet | undefined {
+    let sheet: CSSStyleSheet | undefined;
```
kevinpschaaf, line 38: Why remove the memoization? The cache to reuse the same `CSSResult` is only really useful if you can reuse the stylesheet as well, right?

packages/lit-element/src/lib/css-tag.ts
```diff
@@ -94,5 +90,12 @@ export const css = (
     (acc, v, idx) => acc + textFromCSSResult(v) + strings[idx + 1],
     strings[0]
   );
-  return new CSSResult(cssText, constructionToken);
+  let result = cssResultCache.get(cssText);
```
kevinpschaaf, line 93: Strongly keying the cache off of the strings means every `CSSResult` will be retained for the life of the program. We need to be sure the benefits (reusing the constructible stylesheet) is worth it. Since these are generally going to be used in CE ctors (which are also retained for the life of the program), probably ok?

kevinpschaaf (changes requested):

packages/lit-element/src/lib/css-tag.ts
sorvell: That was a mistake. Putting it back.

packages/lit-element/src/lib/css-tag.ts
sorvell: This seems worth it given the known browser optimizations when sharing styleSheets.

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -126,83 +140,47 @@ export class LitElement extends UpdatingElement {
...
-  static getStyles(): CSSResultOrNative | CSSResultArray | undefined {
-    return this.styles;
-  }
-
-  /** @nocollapse */
-  private static _getUniqueStyles() {
-    // Only gather styles once per class
-    if (this.hasOwnProperty(JSCompiler_renameProperty('_styles', this))) {
-      return;
+  protected static getStyles(
```
kevinpschaaf, line 158: Since this is a braking change to the API anyway, I think a name change would be good.  This is is now a transformer, not a getter, in that it works only on the argument passed in, not on any state.  `flattenStyles`? `transformStyles`?

kevinpschaaf (changes requested):

packages/lit-element/src/lib/lit-element.ts
sorvell: Renamed to `finalizeStyles`.

packages/lit-element/src/lib/css-tag.ts
justinfagnani: I suspect it's going to be really rare that program code is GC'ed

packages/lit-element/CHANGELOG.md
```diff
@@ -20,9 +20,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [3.0.0] - Unreleased
 
 ### Changed
+* For efficiency, the `css` function now maintains a cache and will used a cached value if available when the same style text is requested.
+* [Breaking] LitElement's `adoptStyles` method is now always called. Previously it was called only if the element's `renderRoot` was a shadowRoot. The default implementation only applies styling to the element's shadowRoot and does nothing if the `renderRoot` is not a shadowRoot. This method may be overridden to customize this behavior.
+* [Breaking] LitElement's `static getStyles` method now takes a list of styles the user provided and returns the styles which should be used in the element. If this method is overridden to integrate into a style management system, typically the `super` implementation should be called.
```
justinfagnani, line 25: mention the rename fo finalizeStyles

packages/lit-element/src/lib/updating-element.ts
justinfagnani, line 426: I'd consider making this a helper function rather than a method. It'll be more likely to get inlined by terser.

justinfagnani (approved):

kevinpschaaf (approved):

packages/lit-element/CHANGELOG.md
sorvell: Done.

packages/lit-element/src/lib/updating-element.ts
sorvell: Removed.

PR #1238, opened 2020-09-03
[lit-html] Add live directive

justinfagnani:


packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  update(part: AttributePart, [value]: Parameters<this['render']>) {
+    const element = part.__element;
+    const name = part.name;
+
+    // TODO (justinfagnani): This is essentially implementing a getLiveValue()
+    // method for each part type. Should that be moved into the AttributePart
+    // interface?
+    // TODO (justinfagnani): find alternative to instanceof
+    if (part instanceof PropertyPart) {
```
jridgewell, line 76: Nit: `instanceof` is the slowest operator in JS. Can we do this check during `constructor`?

jridgewell (approved):

packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,97 @@
...
+
+class LiveDirective extends Directive {
+  constructor(part: Part) {
+    super();
+    if (part instanceof EventPart || part instanceof NodePart) {
+      throw new Error(
+        'The `live` directive is not allowed on text or event bindings'
+      );
+    }
+    if ((part as AttributePart).__strings !== undefined) {
```
sorvell, line 59: Using private API

packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,97 @@
...
+ * you use `live()` with an attribute binding, make sure that only strings are
+ * passed in, or the binding will update every render.
+ */
+
+class LiveDirective extends Directive {
+  constructor(part: Part) {
+    super();
+    if (part instanceof EventPart || part instanceof NodePart) {
+      throw new Error(
+        'The `live` directive is not allowed on text or event bindings'
```
sorvell, line 56: Why not allow this on text bindings?

sorvell (commented):
Could we consider making this the default behavior for property parts?

packages/lit-html/src/lib/directives/live.ts
justinfagnani: How would that work, and for what use case? We could maybe special-case strings and Nodes (but not DocumentFragment), but it'd only be useful in case someone was using contenteditable or otherwise messing with the DOM which would be pretty likely to mess up the markers. Whereas with properties, we have a relatively common use case in `<input .value>`

packages/lit-html/src/lib/directives/live.ts
justinfagnani: Yeah

justinfagnani:
> Could we consider making this the default behavior for property parts?

It's not fast enough. I measured this in Chrome and for DOM properties it's a very large speed up to dirty-check on the JS side. Safari is within reason.

packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,96 @@
...
+        return noChange;
+      }
+    } else if (part.type === ATTRIBUTE_PART) {
+      if (element.getAttribute(name) === String(value)) {
+        return noChange;
+      }
+    }
+    // Setting the part's value to nothing causes its dirty-check to fail so
+    // that it always sets the value
+    part.__value = nothing;
```
justinfagnani, line 69: @sorvell @kevinpschaaf I haven't figured out a better way to do this.

What if we made it public, but named it `_value` and documented it as not for general use?

packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,96 @@
...
+    // TODO (justinfagnani): This is essentially implementing a getLiveValue()
+    // method for each part type. Should that be moved into the AttributePart
+    // interface?
+    if (part.type === PROPERTY_PART) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (value === (element as any)[name]) {
+        return noChange;
+      }
+    } else if (part.type === BOOLEAN_ATTRIBUTE_PART) {
+      if (value == element.hasAttribute(name)) {
```
kevinpschaaf, line 59: Should coerce `value === nothing` to `false` before this check?

kevinpschaaf (changes requested):

packages/lit-html/src/lib/directives/class-map.ts
```diff
@@ -0,0 +1,147 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {
```
sorvell, line 15: Why is `classMap` included here?

packages/lit-html/src/lib/directives/live.ts
```diff
@@ -0,0 +1,96 @@
...
+    return value;
+  }
+
+  update(part: AttributePart, [value]: Parameters<this['render']>) {
+    const element = part.element;
+    const name = part.name;
+
+    // TODO (justinfagnani): This is essentially implementing a getLiveValue()
+    // method for each part type. Should that be moved into the AttributePart
+    // interface?
```
sorvell, line 52: Consider moving this to the `parts.ts` helpers?

packages/lit-html/src/lib/directives/live.ts
sorvell: `unset`?

sorvell: The dirty check in attribute part is confusing. Can we add a comment there?

`!((isPrimitive(v) || v === nothing) && v === this.__value) &&
        v !== noChange`

Is `(isPrimitive(v) || v === nothing)` to prevent non-primitive types from setting to attributes? Why bother?

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -597,7 +626,7 @@ export class NodePart {
     const directive = value._$litDirective$;
     if (this.__directive?.constructor !== directive) {
       this.__clear();
-      this.__directive = new directive(this);
+      this.__directive = new directive(this as NodePartInfo);
```
sorvell, line 629: Is this enough safety? You still get the real part.

sorvell (changes requested):

packages/lit-html/src/lib/directives/live.ts
justinfagnani: I changed this to set a sentinel owned by live() that will always fail the dirty check.

justinfagnani: `(isPrimitive(v) || v === nothing)` is to limit the dirty check (`v === this.__value`) to primitives and `nothing`. Otherwise just always set the value.

justinfagnani: Added comment. I tried distributing the `!` but I though it was harder to understand.

packages/lit-html/src/lib/directives/live.ts
justinfagnani: I don't think this is necessarily generally useful. The comment is more me thinking about polymorphism...

packages/lit-html/src/lib/lit-html.ts
justinfagnani: It's some help. We could generate an info object instead...

packages/lit-html/src/lib/directives/live.ts
justinfagnani: done, and coerced `value` to boolean alwasy

justinfagnani:
PTAL

packages/lit-html/src/directives/live.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  render(value: unknown) {
+    return value;
+  }
+
+  update(part: AttributePart, [value]: Parameters<this['render']>) {
+    const element = part.element;
+    const name = part.name;
+
```
jridgewell, line 53: Should we be checking for `noChange`?

sorvell (approved):

kevinpschaaf (approved):
LGTM

packages/lit-html/src/directives/class-map.ts
```diff
@@ -21,13 +21,52 @@ import {
   ATTRIBUTE_PART,
 } from '../lit-html.js';
 
-/**
- * A key-value set of class names to truthy values.
- */
+// IE11 doesn't support classList on SVG elements, so we emulate it with a Set
+// TODO (justinfagnani): Can we move this to the IE compat patch?
+class ClassList {
```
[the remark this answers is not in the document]
justinfagnani: Weird. Something to do with the merges. I rebased to get rid of it.

packages/lit-html/src/directives/live.ts
justinfagnani: Yeah. Added tests.

PR #1240, opened 2020-09-04
[lit-element] Refactor UpdatingElement

sorvell:
* Streamlines some internal private API
* Reduces code size by setting additional instance properties. This may use more memory but trades smaller code size and more readable/usable code.
* renamed `_getUpdateComplete` to `getUpdateComplete`
* when a reflecting property is undefined, its attribute is removed (previously it was unchanged)
* dirty check in `attributeChangedCallback` removed (requested by user)
* Fixes https://github.com/Polymer/lit-element/issues/965


packages/lit-element/src/lib/updating-element.ts
```diff
@@ -669,37 +560,41 @@ export abstract class UpdatingElement extends HTMLElement {
       // in `update` this should not be possible (or an extreme corner case
       // that we'd like to discover).
       // mark state reflecting
-      this._updateState = this._updateState | STATE_IS_REFLECTING_TO_ATTRIBUTE;
+      this._reflectingProperty = name;
```
kevinpschaaf, line 563: I get that using name vs. boolean was added to help with re-entrancy of setting a different reflecting property in a user `attributeChangeCallback` or custom setter, but I was thrown off around whether there is a correctness issue given that an inner reflecting property will have nulled out `_reflectingProperty` by the time the outer/initiating property's logic runs.

But I guess this is erring on the side of ensuring that setting a different reflecting property definitely causes reflection to happen for that property, at the expense of causing extra reflection for the initiating property? We don't expect that to cause an issue?

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -482,66 +443,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
-  private _updateState!: UpdateState;
-  private _instanceProperties?: PropertyValues;
+  private _instanceProperties?: PropertyValues = new Map();
   // Initialize to an unresolved Promise so we can make sure the element has
   // connected before first update.
   private _updatePromise!: Promise<unknown>;
-  private _enableUpdatingResolver: (() => void) | undefined;
+
+  protected isUpdatePending = false;
+  protected hasUpdated = false;
```
kevinpschaaf, line 452: In the previous version, I believe there was a value placed on saving instance memory (i.e. state bitmap, not initializing rarely used instance fields like `_instanceProperties`).

Are the changes here turning the dial from optimizing memory to optimizing code? If so, probably worth stating in the PR goals and document that as an explicit decision?

packages/lit-element/src/lib/updating-element.ts
sorvell: I suggest we not worry about this at the moment because it may become moot if the [follow on PR](https://github.com/Polymer/lit-html/pull/1245) that changes reflection rules lands. If we don't decide to do that, we can consider if there's a corner case here.

packages/lit-element/src/lib/updating-element.ts
sorvell: Yes, good point, will update PR.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -593,40 +503,20 @@ export abstract class UpdatingElement extends HTMLElement {
...
     this.enableUpdating();
   }
 
-  protected enableUpdating() {
-    if (this._enableUpdatingResolver !== undefined) {
-      this._enableUpdatingResolver();
-      this._enableUpdatingResolver = undefined;
-    }
-  }
+  protected enableUpdating() {}
```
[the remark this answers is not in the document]
sorvell: Done.

justinfagnani:
> Reduces code size by setting additional instance properties. This may use more memory but trades smaller code size and more readable/usable code.

How much does moving from a bitmask save? Terser is inlining these constants in lit-html, I'd check the minified output to see that the bitshifts aren't preventing inlining.

sorvell:
It's a minimal code savings of about 0.05KB gzipped, but we are now saving only 1 instance field by using the bit mask after serialization is tracked with a string. It also adds cognitive load to the code.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -640,27 +535,28 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    const attr = (this
+      .constructor as typeof UpdatingElement).attributeNameForProperty(
+      name,
+      options
+    );
+    if (attr !== undefined && options.reflect) {
+      const converter = options.converter;
+      const toAttribute =
+        (converter && (converter as ComplexAttributeConverter).toAttribute) ||
+        defaultConverter.toAttribute;
```
justinfagnani, line 558: Use optional chaining and nullish coalescing here:

```suggestion
      const toAttribute =
        ((options.converter as ComplexAttributeConverter)?.toAttribute) ??
        defaultConverter.toAttribute;
```

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -669,37 +565,41 @@ export abstract class UpdatingElement extends HTMLElement {
...
       // mark state reflecting
-      this._updateState = this._updateState | STATE_IS_REFLECTING_TO_PROPERTY;
+      this._reflectingProperty = propName;
+      const converter = options.converter;
+      const fromAttribute =
+        (converter &&
+          (typeof converter === 'function'
+            ? converter
+            : converter.fromAttribute)) ||
+        defaultConverter.fromAttribute;
```
justinfagnani, line 597: Similar suggestion to the one in `_propertyToAttribute`, but also why does this version have a typeof check and the other doesn't?

If you can leave off that check, it can be:

```suggestion
      const fromAttribute = options.converter?.fromAttribute ?? defaultConverter.fromAttribute;
```

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -669,37 +565,41 @@ export abstract class UpdatingElement extends HTMLElement {
...
     // tslint:disable-next-line:no-unnecessary-type-assertion
     const propName = (ctor._attributeToPropertyMap as AttributeMap).get(name);
-    if (propName !== undefined) {
+    // Use tracking info to avoid reflecting a property value to an attribute
+    // if it was just set because the attribute changed.
+    if (propName !== undefined && this._reflectingProperty !== propName) {
       const options = ctor.getPropertyOptions(propName);
       // mark state reflecting
-      this._updateState = this._updateState | STATE_IS_REFLECTING_TO_PROPERTY;
+      this._reflectingProperty = propName;
```
justinfagnani, line 590: Try to make variable declarations contiguous so that they can be minified into one declaration. Can you move this to below the declaration of `fromAttribute`?

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -877,13 +767,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
-   *     async _getUpdateComplete() {
-   *       await super._getUpdateComplete();
+   *     async getUpdateComplete() {
+   *       await super.getUpdateComplete();
    *       await this._myChild.updateComplete;
    *     }
    *   }
    */
-  protected _getUpdateComplete() {
+  protected getUpdateComplete() {
```
justinfagnani, line 776: Are we sure we don't want to convert `updateComplete()` into a method?

We could use @kevinpschaaf's thenable hack:

```ts
if (DEV_MODE) {
  UpdatingElement.prototype.updateComplete.then = () => {
    throw new Error('updateComplete is a method');
  };
}
```

justinfagnani (changes requested):

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -416,10 +381,7 @@ export abstract class UpdatingElement extends HTMLElement {
...
    */
   protected static getPropertyOptions(name: PropertyKey) {
-    return (
-      (this._classProperties && this._classProperties.get(name)) ||
-      defaultPropertyDeclaration
-    );
+    return this._classProperties!.get(name) || defaultPropertyDeclaration;
   }
 
   protected static get hasFinalized() {
```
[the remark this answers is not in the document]
sorvell: I used it in LitElement but I'm gonna take this out and just have the super return true if it actually finalized.

packages/lit-element/src/lib/updating-element.ts
sorvell: If the converter is a function, it's `fromAttribute`. You have to specify `toAttribute` by name.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -468,7 +429,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * Returns the property name for the given attribute `name`.
    * @nocollapse
    */
-  private static _attributeNameForProperty(
+  protected static attributeNameForProperty(
```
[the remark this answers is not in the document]
sorvell: I added it based on [this](https://github.com/Polymer/lit-element/issues/985), but the subclass could also override the attribute property option, so we can probably remove it.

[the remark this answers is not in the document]
sorvell: OK, responded to that issue and making this private again.

packages/lit-element/src/lib/updating-element.ts
sorvell: https://github.com/Polymer/lit-html/issues/1253

packages/lit-element/src/lib/updating-element.ts
sorvell: Done.

packages/lit-element/src/lib/updating-element.ts
sorvell: Done.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -640,27 +535,28 @@ export abstract class UpdatingElement extends HTMLElement {
...
-      // an undefined value does not change the attribute.
-      if (attrValue === undefined) {
-        return;
-      }
+    const attr = (this
+      .constructor as typeof UpdatingElement).attributeNameForProperty(
+      name,
+      options
+    );
+    if (attr !== undefined && options.reflect) {
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -343,13 +309,12 @@ export abstract class UpdatingElement extends HTMLElement {
...
     // user-defined accessors. Note that if the super has an accessor we will
     // still overwrite it
-    if (options.noAccessor || this.prototype.hasOwnProperty(name)) {
-      return;
-    }
-    const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-    const descriptor = this.getPropertyDescriptor(name, key, options);
-    if (descriptor !== undefined) {
-      Object.defineProperty(this.prototype, name, descriptor);
+    if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
```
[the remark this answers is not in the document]
sorvell: Ended up being very slightly larger.

packages/lit-element/src/lib/updating-element.ts
justinfagnani: ah

justinfagnani (approved):

kevinpschaaf (approved):

PR #1241, opened 2020-09-04
[lit-element] Changes the type of `css` to the same as `LitElement.styles`

sorvell:
Adds the type `CSSResultGroup` and types both `css` and `LitElement.styles`
this way. This prevents the need for a subclass to cast `styles` to `any` when setting an Array when the superclass set a single value (or visa versa).


justinfagnani:
Looks good to me, and by good I mean kind of ugly but a pragmatic usability win for developers. I'd like to maybe get @rictic or @aomarks opinion on this though.

packages/lit-element/src/lib/css-tag.ts
```diff
@@ -18,6 +18,13 @@ export const supportsAdoptingStyleSheets =
   'adoptedStyleSheets' in Document.prototype &&
   'replace' in CSSStyleSheet.prototype;
 
+export type CSSResultOrNative = CSSResult | CSSStyleSheet;
```
csvn, line 21: > Looks good to me, and by good I mean kind of ugly but a pragmatic usability win for developers.

If minimum Typescript version is 3.7 allowed, [recursive type aliases](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#more-recursive-type-aliases) can be used to avoid needing to use `CSSResultOrNative` and `CSSResultArray`, and simplify the code a bit:

```ts
export type CSSResultGroup = CSSResult | CSSStyleSheet | CSSResultGroup[];
```
> [Plaground link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDCBlDAlYBnAVwBsYBxKCAsOAXnS10JLgB96skjgMALYYeG0w58xMhSoBtALoBuAFCKAxkQCGePO0Zi4AbwC+8oA)

Sorry if I'm just barging in on the review 🙂

justinfagnani (approved):

packages/lit-element/src/lib/css-tag.ts
rictic: We try to produce types that are back compatible with earlier versions of TypeScript due to people using tooling that constrains their TypeScript version. I think the main example last time was TS 3.4, which I think Angular was bound to?

Unless this gives a benefit to users, probably better to have slightly awkward code for us and better compatibility.

justinfagnani: Since this is a major version change, we can bump our minimum TypeScript version. We should do another survey of where tools that integrate tsc are at - Angular, Storybook, Deno, etc.

rictic: Formally, I don't think we can consider type changes to be breaking changes, because of the typescript compiler's complete lack of semver.

That said, we've gotta keep in mind that breaking changes always have a cost, and they don't get cheaper just because they're bundled with other breaking changes. Every incompatibility that we add will mean cost in time and effort to projects and people downstream, so the benefit has to outweigh the cost.

PR #1242, opened 2020-09-04
[lit-html] Add disconnectedCallback for directives

### LINKED ISSUE #283: Disconnected handler for directives
I've been playing around with combining lit-html and observables, and it feels really nice!
The only issue I've run into is that, since observables are a push-based API, they must be explicitly unsubscribed when the source outlives the destination to avoid leaking events, memory etc....

## The problem

Lets say I'm using a very simple directive:

```js
const observeReplace = value$ => directive(part => {
  value$.subscribe(value => part.setValue(value));
});
```

And rendering it as such:

```js
const second$ = Rx.Observable.interval(1000).do(console.log);

render(html`Seconds running: ${observeReplace(second$)}`, document.body);
```

Everything is fine. However if you later replace the template:

```js
setTimeout(() => {
  render(html`Should be stopped.`, document.body);
}, 5000);
```

The interval subscription is still running, as the `.do(console.log)` shows, and lit-html errors:

```
Subscriber.js:214 Uncaught TypeError: Cannot read property 'nodeType' of null
    at NodePart._setText (lit-html.js:483)
    at NodePart.setValue (lit-html.js:449)
    at SafeSubscriber.value$.subscribe.value [as _next] (index.js:13)
    at SafeSubscriber.next (Subscriber.js:173)
    at Subscriber._next (Subscriber.js:118)
    at Subscriber.next (Subscriber.js:82)
    at TapSubscriber._next (tap.js:98)
    at TapSubscriber.next (Subscriber.js:82)
    at AsyncAction.dispatch [as work] (interval.js:56)
    at AsyncAction._execute (AsyncAction.js:105)
```

To fix this, a directive needs to provide a way to know when new values should stop being pushed to the part. So far the only way I know of to do this with the current api is to somehow use [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect `part.instance.template.element` being removed, but that requires that you observe `childList` mutations on the *parent*, which isn't present when the directive callback runs.

## Possible solutions

The ideal for me would be to simply accept observables 😉 (which are basically equivalent to directives) but a possibly smaller change would be to just steal a trick from observable subscriber functions and have directive callbacks optionally return a cleanup function:

```js
const observeReplace = value$ => directive(part => {
  const sub = value$.subscribe(value => part.setValue(value));
  return () => sub.unsubscribe();
});

// not observable specific:
const now = directive(part => {
  const id = setInterval(() => part.setValue(new Date().toString()), 1000);
  return () => clearInterval(id);
});
```

justinfagnani:
This is the simplest approach to directive disconnection, and possibly the best performance tradeoff. It requires a value-tree walk when a template instance is removed, but it doesn't have any first-render impact.

In writing the tests I noticed that we don't really handle dynamically switching a directive to a non-directive and back. The correct behavior is probably to clear the directive when switching away. One could argue that we shouldn't support dynamic directives.

Fixes #283


packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1469,4 +1469,81 @@ suite('lit-html', () => {
...
+    }
+
+    disconnectedCallback() {
+      this.log.push('disconnected');
+    }
+  }
+  const disconnectingDirective = directive(DisconnectingDirective);
+
+  // TODO (justinfagnani): dynamic directives are not handled correctly in
+  // general. We may not even want to support them. TBD.
```
kevinpschaaf, line 1488: I didn't follow, what isn't handled correctly here? We [dirty-check the constructor value](https://github.com/Polymer/lit-html/pull/1216/files#diff-073dbf13ebed6ac6ef2cce7a849a36a3R599) to know whether to clear the node part (which should disconnect the previous directive); committing any other value type (like `nothing` here) should also cause a clear. What am I missing?

kevinpschaaf (commented):
Impl looks fine; just need perf test. :P

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -248,6 +246,17 @@ export abstract class Directive {
...
+  /**
+   * An optional lifecycle method that is called when a directive, or the
+   * template it's attached to, is disconnected from the render tree.
+   *
+   * Note: this is different from being connected to or disconnected from
+   * the document. disconnectedCallback is called even when a directive is
+   * disconnected from a render tree that itself is not connected to the
+   * document.
+   */
+  disconnectedCallback?(): void;
```
sorvell, line 259: Consider if we want to name this `disconnected` because as noted in the comments it's not strictly the same as an element's `disconnectedCallback`

packages/lit-html/src/lib/lit-html.ts
```diff
@@ -463,6 +472,12 @@ class TemplateInstance {
     this.__template = template;
   }
 
+  __disconnect() {
```
sorvell, line 475: Let's add a comment about this being a part tree walk and a TODO for perf testing.

sorvell (changes requested):
Can a directive be `connected` as well? For example, when using `live` or `repeat`?. If so, won't that be an issue? If cleanup work has occurred in disconnected and it gets re-used, it won't be in a good state.

packages/lit-html/src/test/lit-html_test.ts
justinfagnani: I hit this in `templateContent` and filed an issue: https://github.com/Polymer/lit-html/issues/1286

packages/lit-html/src/lit-html.ts
```diff
@@ -755,6 +767,12 @@ class TemplateInstance {
     this._template = template;
   }
 
+  _disconnect() {
```
sorvell, line 770: Let's add a comment describing what this does and make especially clear that it recurses down the tree of parts that have disconnect

packages/lit-html/src/lit-html.ts
```diff
@@ -794,20 +812,24 @@ class TemplateInstance {
     return fragment;
   }
 
-  _update(values: Array<unknown>) {
+  _update(values: Array<unknown>): boolean {
```
sorvell, line 815: Since most of these objects end up having state flags anyway (`_hasDisconnect`), could we just read that flag rather than carefully thread a seemingly arbitrary return value up through these functions? I think that's likely less code overall.

packages/lit-html/src/lit-html.ts
```diff
@@ -755,6 +767,12 @@ class TemplateInstance {
     this._template = template;
   }
 
+  _disconnect() {
+    for (const part of this._parts) {
```
sorvell, line 771: Couldn't `TemplateInstance` have a `_hasDisconnect` flag and avoid this work if it's not set?

packages/lit-html/src/lit-html.ts
```diff
@@ -1044,9 +1092,13 @@ export class NodePart {
       // itemParts always have end nodes
       this._clear(itemPart?._endNode!.nextSibling);
     }
+    return hasDisconnect;
   }
 
   private _clear(start: ChildNode | null = this._startNode.nextSibling) {
+    if (this._hasDisconnect) {
+      this._disconnectValue();
```
sorvell, line 1100: Why is this `_disconnectValue` rather than `_disconnect`? Maybe add a comment to clarify?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
kevinpschaaf: It compiles to a hair more code to read state flags rather than return a value. 🤷‍♂️

On top of that, in NodePart setValue, we still have to return a boolean from `_commitTemplateResult` and `_commitIterable`, since we can't reset the flag to false and let those set the flag to true because those commit functions need to clear based on the previous state of the flag. I'm inclined to leave it as-is for now, for these reasons, plus it's easier to grok what's going on imo.

If we need to pass more info up the tree, we have options to change it.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

kevinpschaaf:
Closing, as this feature was implemented via an alternate implementation in https://github.com/Polymer/lit-html/pull/1432.

PR #1244, opened 2020-09-04
[lit-html] Add static.js module and unsafeStatic() - static bindings

### LINKED ISSUE #78: Allow injection of static content into templates

justinfagnani:
This adds a new entrypoint, `lit-html/static.js`, that exports new `html` and `css` template tags and a `unsafeStatic()` wrapper function, to provide unsafe static value support.

Static values are interpreted as part of the template strings and parsed along with the template strings. They can be used to emulate dynamic portions of a template that otherwise can't have bindings, like tag names and attribute names or complete attribute name/value pairs. Static values are unsafe to use with user input, since the content is interpreted as HTML, and not text content, which could lead to XSS vulnerabilities.

Static values should usually not change between renders. Rendering with a different static value essentially creates an entirely new template, and the entire template instance being rendered will be replaced.

Fixes #78


rictic:
What's the use case here?

rictic:
I think the trusted types check we want is:

```typescript
const isTrustedHTML: (s: unknown)=> s is TrustedHTML = 
    typeof TrustedHTML !== 'undefined' ? 
        (s) => s instanceof TrustedHTML : 
        (s) => typeof s === 'string';
```

kevinpschaaf (commented):
No feedback on impl, just need to decide if we really need a general static feature, or if e.g. `TagPart` and `StyleTextPart` might be a better way to avoid (micro) runtime perf and (larger) XSS concerns, especially now that we have scanner support for those.

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1370,6 +1371,41 @@ suite('lit-html', () => {
...
+      // the public API
+    });
+
+    test('static tag binding', () => {
+      const tagName = unsafeStatic('div');
+      render(html`<${tagName}></${tagName}>`, container);
+      assert.equal(stripExpressionComments(container.innerHTML), '<div></div>');
+    });
+
+    test('dynamic binding after static text binding', () => {
```
justinfagnani, line 1400: TODO: We need a test with both static values and nulls.

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,77 @@
...
+      // Collect any unsafeStatic values, and their following template strings
+      // so that we treat a run of template strings and unsafe static values as
+      // a single template string.
+      while (i < l && (values[i] as StaticValue)?._$litStatic$ !== undefined) {
+        s += (values[i] as StaticValue)?._$litStatic$ + strings[++i];
+      }
+      newStrings.push(s);
+      // If by consuming unsafe static values we use the last template string,
+      // then this template has no dynamic bindings.
+      if (i++ > l) {
```
sorvell, line 45: Do you need this?

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Generates a template literal tag function that returns a TemplateResult with
+ * the given result type.
+ */
+const tag = (coreTag: typeof coreHtml) => (
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): TemplateResult => {
+  let processedStrings = stringsCache.get(strings);
```
sorvell, line 27: If the static values change, we won't get a different `processedStrings`. Didn't we decide to include the static values as part of the template key?

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+      newStrings.push(s);
+      // If by consuming unsafe static values we use the last template string,
+      // then this template has no dynamic bindings.
+      if (i++ > l) {
+        break;
+      }
+    }
+    if (i <= l) {
+      newStrings.push(strings[l]);
```
sorvell, line 50: What case is this? `strings[l]` is outside the array range isn't it?

sorvell (changes requested):
Needs tests.

sorvell:
Note from offline chat:

Right now the template cache is only keyed on the static strings in the template and this means the following type of usage is not supported:

```
inputTemplate(type) => html`<input type=${unsafeStatic(type)}>` // always returns first call result
```
We do want to support this use case. To do so, the static values must be included in the template key.

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,102 @@
...
+      // Collect any unsafeStatic values, and their following template strings
+      // so that we treat a run of template strings and unsafe static values as
+      // a single template string.
+      while (i < l && (values[i] as StaticValue)?._$litStatic$ !== undefined) {
+        s += (values[i] as StaticValue)._$litStatic$ + strings[++i];
+      }
+      newStrings.push(s);
+      // If by consuming unsafe static values we use the last template string,
+      // then this template has no dynamic bindings.
+      if (i++ > l) {
```
justinfagnani, line 70: Oops. This was from when it was in core.

packages/lit-html/src/static.ts
justinfagnani: This is for any case where the last value isn't static and so the last string wasn't consumed. `strings[l]` is the last element of the array because `l = strings.length - 1`. This works out because the strings array is one longer than the values array. I updated the condition to use `===` and added a comment.

packages/lit-html/CHANGELOG.md
```diff
@@ -34,6 +34,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `renderBefore` to render options. If specified, content is rendered before the node given via render options, e.g. `{renderBefore: node}`.
 - Added development mode, which can be enabled by setting the `development` Node exports condition. See `README.md` for more details.
 
+* Added `unsafeStatic()`, which allows template authors to add strings to the
+  static structure of the template, before it's parsed as HTML.
+
+* Added `renderBefore` to render options. If specified, content is rendered before the node given via render options, e.g. `{renderBefore: node}`.
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,77 @@
...
+    let i = 0;
+
+    while (i < l) {
+      let s = strings[i];
+
+      // Collect any unsafeStatic values, and their following template strings
+      // so that we treat a run of template strings and unsafe static values as
+      // a single template string.
+      while (i < l && (values[i] as StaticValue)?._$litStatic$ !== undefined) {
+        s += (values[i] as StaticValue)?._$litStatic$ + strings[++i];
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/lit-html/src/static.ts
justinfagnani: Fixed

justinfagnani:
PTAL

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,109 @@
...
+const tag = (coreTag: typeof coreHtml) => (
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): TemplateResult => {
+  let key = strings[0];
+  for (let i = 0; i < strings.length; i++) {
+    if ((values[i] as StaticValue)?._$litStatic$ !== undefined) {
+      key += (values[i] as StaticValue)._$litStatic$;
+    } else {
+      key += '${lit-html-joiner}';
```
sorvell, line 48: Maybe make this shorter? Any reason it looks like a template expression?

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    key += strings[i + 1];
```
sorvell, line 50: This should iterate the values length, right? Won't the `key` include `undefined` at the end here?

packages/lit-html/src/static.ts
sorvell: Maybe add a comment about why this is included in the key? It's so that the values length is part of the key and helps with uniqueness, yeah?

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Generates a template literal tag function that returns a TemplateResult with
+ * the given result type.
+ */
+const tag = (coreTag: typeof coreHtml) => (
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): TemplateResult => {
+  let key = strings[0];
```
sorvell, line 43: Is it worth having the `stringsCache` at all? We create the key string for every call which iterates and joins everything into a string. The cache is used to avoid iterating and making an array with combined static strings. Those processes seem pretty similar and therefore it doesn't seem worth avoiding the array creation. Plus, there's extra work to fill the cache. Maybe benchmark it?

sorvell (changes requested):

packages/lit-html/src/static.ts
kevinpschaaf: You need a cache of the strings array, since you need to return _the same_ strings array for every call with the same strings+statics (key).

kevinpschaaf: It does seem like this can all be done with one loop plus a join as below, though. Are you creating the key first via concat (rather than join) to bias the cache hit to be faster, assuming that concat is faster than creating the strings array and joining it?

Example input:
```js
html`string1${unsafeStatic('static1')}string2${value1}string3${unsafeStatic('static2')}${value2}string4`
```
Do a single loop over `strings`/`values` and create the collapsed array of strings + non-static values, resulting in:
```js
strings = ['string1static1string2', 'string3static2', 'string4'];
values = [value1, value2];
```

Join strings to get key, and use the strings we just made if it doesn't exist yet:
```js
const key = strings.join('$lit$part$');
let cachedStrings = stringsCache.get(key);
if (!cachedStrings) {
  stringsCache.set(key, cachedStrings = strings
}
return coreTag(cachedStrings, values);
```

kevinpschaaf: Either way, the non-static values could be accumulated in the first loop creating the key, rather than running a separate filter at the end.

kevinpschaaf (changes requested):

packages/lit-html/src/static.ts
justinfagnani: I was first inspired by @aomarks using `${` as a joiner in @lit/localize on the idea that it can't be used in a template string without escaping, but then realized that the non-raw string value would still contain `${` even if escaped. It can be anything unlikely to appear in the template.

packages/lit-html/src/static.ts
justinfagnani: fixed

packages/lit-html/src/static.ts
justinfagnani: String concat is a bit faster, but it might not be worth the code

justinfagnani: Refactored to one loop

packages/lit-html/src/test/static_test.ts
```diff
@@ -0,0 +1,135 @@
...
+    assert.strictEqual(div2, div);
+
+    render(t('span', 'abc'), container);
+    // Rendering with a new static value should work, though it re-renders
+    // since we have a new template.
+    assert.equal(
+      stripExpressionComments(container.innerHTML),
+      '<span>abc</span>'
+    );
+    const span = container.querySelector('div');
```
sorvell, line 115: This should query for `span`.

sorvell: And you should assert that it's not null.

packages/lit-html/src/test/static_test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    render(t('span', 'def'), container);
+    assert.equal(
+      stripExpressionComments(container.innerHTML),
+      '<span>def</span>'
+    );
+    const span2 = container.querySelector('div');
```
sorvell, line 122: Should query for `span`.

sorvell (changes requested):

packages/lit-html/src/static.ts
```diff
@@ -0,0 +1,99 @@
...
+
+type StaticValue = ReturnType<typeof unsafeStatic>;
+
+const stringsCache = new Map<string, TemplateStringsArray>();
+
+/**
+ * Generates a template literal tag function that returns a TemplateResult with
+ * the given result type.
+ */
+const tag = (coreTag: typeof coreHtml) => (
```
kevinpschaaf, line 39: Can we export this with a good name, i.e. `enhanceWithStatic` or `addStaticSupport` to support tag composition?

kevinpschaaf (approved):
LGTM, with the suggestion about exporting the tag factory (and the test bug Steve noted)

packages/lit-html/src/test/static_test.ts
justinfagnani: fixed

packages/lit-html/src/test/static_test.ts
justinfagnani: fixed

packages/lit-html/src/static.ts
justinfagnani: exported as `withStatic`, updated comments here and on the tags too

sorvell (approved):

PR #1245, opened 2020-09-07
[lit-element] Changes attribute reflection rules

sorvell:
This is being done for simplicity and consistency.
* old: set attribute => UpdatingElement sets reflect: true property => STOP
* new: set attribute => UpdatingElement sets reflect: true property => UpdatingElement sets attribute

Note, reflection in the other direction is still prevented so that values that don't uniquely deserialize do not reset properties:
* set a reflect: true property => UpdatingElement sets attribute => STOP


packages/lit-element/src/lib/updating-element.ts
```diff
@@ -789,17 +772,18 @@ export abstract class UpdatingElement extends HTMLElement {
...
-        this._propertyToAttribute(k, this[k as keyof this], v)
-      );
-      this._reflectingProperties = undefined;
-    }
+   * @param changedProperties Map of changed properties with old values
+   */
+  protected update(changedProperties: PropertyValues) {
+    // Use forEach so this works even if for/of loops are compiled to for
+    // loops expecting arrays
+    changedProperties.forEach((_v, k) =>
```
kevinpschaaf, line 780: Why do we do the `options.reflect` test in `_propertyToAttribute` rather than before calling `_propertyToAttribute`?  This was surprising to me, seems like `_propertyToAttribute` should do the thing, given its name?  It also wastes an [`attributeNameForProperty`](https://github.com/Polymer/lit-html/blob/28be4b442d27b2a1cd3693113a1a5432555000c2/packages/lit-element/src/lib/updating-element.ts#L540) lookup for every changed property, regardless of whether it is reflecting now that we're looping over all changed properties, not just the reflecting ones.

kevinpschaaf (changes requested):
One performance optimization suggestion below.

packages/lit-element/src/lib/updating-element.ts
sorvell: Done.

justinfagnani:
Does this mean setting the attribute of a reflecting property _always_ results in two attribute sets? What's the perf implication and what's the byte-size savings?

sorvell:
I think the current code results in unexpected behavior. Consider, for example, the un-reflect attribute pattern can be done just with a `toAttribute = () => null, reflect: true` with this change, but before, it confusingly didn't work.

In addition, previously setting the property and attribute of a reflecting property in the same update cycle would cause a reflect. This is typical setup with a default value + an initial attribute... also just weird behvaior.

It'd like to have no reflect prevention, but I think we need it to stop reflecting objects and arrays from always being reset to new property values. That's arguable though.

* Code difference: 0.2KB minified, 0.05KB gzipped.
* Will run a benchmark.

PR #1246, opened 2020-09-08
[lit-element] Lit next updating controller

sorvell:


packages/lit-element/src/lib/updating-controller.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (host) {
+      host.connectedCallbacks.delete(controller._onConnected);
+      host.disconnectedCallbacks.delete(controller._onDisconnected);
+      host.updateCallbacks.delete(controller._onUpdate);
+      host.updatedCallbacks.delete(controller._onUpdated);
+    }
+  }
+
+  requestUpdate() {
+    if (this.isActive) {
```
kevinpschaaf, line 89: This is for squelching async updates after removal?

packages/lit-element/demo/controllers.html
```diff
@@ -0,0 +1,433 @@
...
+        }
+
+        set value(value) {
+          this._value = value;
+          this.requestUpdate();
+        }
+
+      }
+
+    const createContext = (initialValue, key, ProviderBase = Provider, ConsumerBase = Consumer) => {
```
kevinpschaaf, line 207: There's an alternative API for context where we use directives, which I think we should at least explore.  Because context is tree based, directives seem more semantic and more closely match React, but doesn't lend itself as well to some patterns.

Roughly:

```html
<my-app ${provide(ThemeContext, 'dark')}>
  ...
  <my-button>
    #shadow-root
      <button class=${consume(ThemeContext, value => value)}>
```

The above strawman would require bringing back async commit ability to directives, and getting the initial value for consumption requires being connected to the tree. But API wise, this seems nicer to me.

arthurevans (commented):
As I was reading through to understand this concept, I noticed a couple of typos.
These edits are probably premature, so feel free to dismiss this review. 

Also was not entirely clear on why you'd use another UpdatingController as the host for an UpdatingController, and unless I missed it there wasn't an example of that.

packages/lit-element/src/lib/updating-controller.ts
sorvell: Right.

packages/lit-element/demo/controllers.html
sorvell: Context is intended to basically be similar to DI. I don't see how the desire "I want something provided to me from somewhere up the tree" makes sense as a directive?!?

T-Knott-Mesh:
Hey WC Friends :) 

Love to see the draft here - if anyone would like to chat about this idea/concept Runtime.dev has developed tremendously along these lines. 

Would love to have a conversation about reasonings/validations/objectives with this (to further validate our approach) but to also understand how a component-controller model could potentially be cross lib. Would be great to put a Meet on the calendar and catch up.

Nevertheless - I love to see the work here. LMK how I can help/provide situations that we ran into. 

Hope all has been well and everyone is healthy and safe. 

-Tsavo

PR #1247, opened 2020-09-08
[lit-element] Separates decorators into individual modules

sorvell:
* Also separates decorators tests.
* Moved tests around so decorators are only used in tests for those decorators.

No code has really changed here, just moving around the pieces.


kevinpschaaf (commented):
While we're at it, WDYT about separating decorators into `typescript` and `babel-old` (or whatever it is called??) folders?  Since these are clearly not standard, seems good for people to choose the flavor of decorators they want and not get the cruft? And if we ever get standard decorators from TC39, we won't have to carry around this code, which would all become cruft?

sorvell:
Based on offline discussion it's possible to mix and match TS and Babel decorators in practice so separating them seems like it would lead to bad DX.

packages/lit-element/src/lib/decorators/customElement.ts
```diff
@@ -0,0 +1,71 @@
...
+ * class MyElement {
+ *   render() {
+ *     return html``;
+ *   }
+ * }
+ * ```
+ * @category Decorator
+ * @param tagName The name of the custom element to define.
+ */
+export const customElement = (tagName: string) => (
```
justinfagnani, line 66: btw, I'd love to bring back the type here of `tagName: keyof HTMLElementTagNameMap`. There was a problem with that type being expanded to a union at our compile time, but I think there are workarounds, like maybe using a `function` instead of an arrow function.

packages/lit-element/src/lib/decorators/decorators.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/*
+ * IMPORTANT: For compatibility with tsickle and the Closure JS compiler, all
+ * property decorators (but not class decorators) in this file that have
+ * an @ExportDecoratedItems annotation must be defined as a regular function,
+ * not an arrow function.
+ */
+
+export * from './customElement.js';
```
justinfagnani, line 22: why don't we keep this file at `lib/decorators.ts`?

justinfagnani (approved):
LGTM

I think you can rebase this right into `lit-next` if you want to merge right away.

packages/lit-element/src/lib/decorators/decorators.ts
sorvell: Done.

kevinpschaaf (approved):

PR #1249, opened 2020-09-10
[lit-next] Set up stub tachometer benchmarks

aomarks:
- Creates packages/benchmarks
- Two trivial throwaway benchmarks, one for lit-html, one for lit-element
- Uses https://github.com/andrewiggins/tachometer-reporter-action GitHub Action to run and comment on PRs
- Depends on https://github.com/Polymer/tachometer/pull/191 (using pre-release for now)

Fixes https://github.com/Polymer/internal/issues/1064

cc @andrewiggins

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -10,11 +10,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
 
-      - name: Initialize tachometer comment
-        uses: andrewiggins/tachometer-reporter-action@v1
-        with:
-          initialize: true
-
       - name: NPM install
         run: npm ci
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -36,8 +36,8 @@ jobs:
         with:
           report-id: lit-html-stub1
           path: packages/benchmarks/lit-html/stub1/results.json
-          pr-bench-name: local
-          base-bench-name: lit-next
+          pr-bench-name: this-change
+          base-bench-name: tip-of-tree
 
       - name: Benchmark lit-element/stub1
         run: |
@@ -51,5 +51,5 @@ jobs:
         with:
           report-id: lit-element-stub1
           path: packages/benchmarks/lit-element/stub1/results.json
-          pr-bench-name: local
-          base-bench-name: lit-next
+          pr-bench-name: this-change
+          base-bench-name: tip-of-tree
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,55 @@
+name: Benchmarks
+
+on: [pull_request]
+
+jobs:
+  benchmarks:
+    name: benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+
+      - name: Initialize tachometer comment
+        uses: andrewiggins/tachometer-reporter-action@v1
+        with:
+          initialize: true
+
+      - name: NPM install
+        run: npm ci
+
+      - name: Lerna bootstrap
+        run: npm run bootstrap
+
+      - name: Build
+        run: npm run build
+
+      - name: Benchmark lit-html/stub1
+        run: |
+          cd packages/benchmarks
+          npx tachometer \
+            --config lit-html/stub1/tachometer.json \
+            --json-file lit-html/stub1/results.json
+
+      - name: Report lit-html-stub1
+        uses: andrewiggins/tachometer-reporter-action@v1
+        with:
+          report-id: lit-html-stub1
+          path: packages/benchmarks/lit-html/stub1/results.json
+          pr-bench-name: local
+          base-bench-name: lit-next
+
+      - name: Benchmark lit-element/stub1
+        run: |
+          cd packages/benchmarks
+          npx tachometer \
+            --config lit-element/stub1/tachometer.json \
+            --json-file lit-element/stub1/results.json
+
+      - name: Report lit-element-stub1
+        uses: andrewiggins/tachometer-reporter-action@v1
+        with:
+          report-id: lit-element-stub1
+          path: packages/benchmarks/lit-element/stub1/results.json
+          pr-bench-name: local
+          base-bench-name: lit-next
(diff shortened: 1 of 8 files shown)


packages/benchmarks/lit-html/stub1/stub1.html
```diff
@@ -0,0 +1,14 @@
...
+<script type="module">
+  import { html, render } from "lit-html";
+
+  (async () => {
+    performance.mark("start");
+    for (let i = 0; i < 1000; i++) {
+      render(html`<p>Hello World</p>`, document.body);
+    }
+    // Wait for LitElement render microtasks to complete.
+    await Promise.resolve(null);
```
kevinpschaaf, line 11: Hmm, I think you meant to put this in the lit-element test? (`render` is sync in lit-html).  Not a huge deal since these are stubs.

packages/benchmarks/lit-element/stub1/tachometer.json
```diff
@@ -0,0 +1,63 @@
...
+        "name": "chrome",
+        "headless": true
+      },
+      "measurement": {
+        "mode": "performance",
+        "entryName": "render"
+      },
+      "expand": [
+        {
+          "name": "local"
```
kevinpschaaf, line 18: For names in the table, maybe something like this would be clearer?
* ~local~ → This change
* ~lit-next~ → Tip-of-tree
* ~npm~ → Previous release

?

kevinpschaaf (commented):
This is awesome!

Just a couple minor nits.

packages/benchmarks/lit-element/stub1/tachometer.json
aomarks: Done.

packages/benchmarks/lit-html/stub1/stub1.html
aomarks: Oops, yeah I updated the wrong file when adding that await. Fixed.

kevinpschaaf (approved):

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,55 @@
...
+
+jobs:
+  benchmarks:
+    name: benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+
+      - name: Initialize tachometer comment
```
justinfagnani, line 13: What does "Initialize tachometer comment" mean? Is "comment" supposed to be there?

packages/benchmarks/lit-element/stub1/tachometer.json
```diff
@@ -0,0 +1,63 @@
...
+              }
+            }
+          }
+        },
+        {
+          "name": "previous-release",
+          "packageVersions": {
+            "label": "previous-release",
+            "dependencies": {
+              "lit-html": "^1.0.0",
```
justinfagnani, line 55: Do you need this, or will it come along with lit-element?

packages/benchmarks/lit-html/stub1/tachometer.json
```diff
@@ -0,0 +1,63 @@
...
+                "repo": "https://github.com/Polymer/lit-html.git",
+                "ref": "lit-next",
+                "subdir": "packages/lit-html",
+                "setupCommands": [
+                  "npm ci",
+                  "npm run bootstrap",
+                  "npm run build"
+                ]
+              },
+              "lit-element": {
```
justinfagnani, line 36: don't need this, right?

packages/benchmarks/lit-html/stub1/tachometer.json
```diff
@@ -0,0 +1,63 @@
...
+            }
+          }
+        },
+        {
+          "name": "previous-release",
+          "packageVersions": {
+            "label": "previous-release",
+            "dependencies": {
+              "lit-html": "^1.0.0",
+              "lit-element": "^2.0.0"
```
justinfagnani, line 56: don't need this either?

justinfagnani (approved):
Awesome!

packages/benchmarks/lit-html/stub1/tachometer.json
kevinpschaaf: I think we need those at least temporarily right now because the package.json for `packages/benchmarks` currently includes an unreleased lit-element version 3.0.0-pre (which when installed during lerna bootstrap is fine), but when re-installed via `packageVersions` by tachometer would error if it weren't replaced with a good version (even if it is unused by the benchmark).  I hit this in working on top of this PR last night.

packages/benchmarks/lit-element/stub1/tachometer.json
aomarks: Since `lit-html` is declared in the `package.json` at `^2.0.0-pre.0`, we'd get an error since that's not published, plus it would install two versions even if it was published.

.github/workflows/benchmarks.yml
aomarks: Got it from here: https://github.com/andrewiggins/tachometer-reporter-action#multiple-benchmark-jobs and https://github.com/andrewiggins/tachometer-reporter-action#initialize

But reading it again, looks like it's only required if we're doing multiple jobs in parallel (which is something we could do, but I was starting with this simpler config). Removed for now.

packages/benchmarks/lit-html/stub1/tachometer.json
aomarks: Exactly! Kind of annoying. I was thinking tachometer could check for symlinks in `node_modules`, and convert those into a local file path version for NPM whenever doing a dependency swap.

packages/benchmarks/lit-html/stub1/tachometer.json
aomarks: See above.

.github/workflows/benchmarks.yml
andrewiggins: Yup, that's right. If you start running benchmarks in different jobs then you'll want an initial job to create the PR comment the other jobs will then fill in (there is a way to do it without needing an initial job but it is susceptible to race conditions leading to multiple result comments so I just recommend the "initialize" job approach. )

andrewiggins:
Let me know if you have any questions/feedback/thoughts for tachometer-reporter-action!

aomarks:
> Let me know if you have any questions/feedback/thoughts for tachometer-reporter-action!

It worked perfectly right away, docs are really good so far, will let you know if we hit any issues. Thanks for making it!

PR #1250, opened 2020-09-11
[lit-html] Add cache(), repeat(), and part helpers

justinfagnani:


packages/lit-html/src/lib/lit-html.ts
```diff
@@ -655,6 +661,54 @@ export class NodePart {
...
+            this.__insert(createMarker()),
+            this.__insert(createMarker()),
+            this.options
+          ))
+        );
+      } else {
+        // Reuse an existing part
+        itemPart = itemParts[partIndex];
+      }
+      itemPart.__setValue(item);
```
sorvell, line 700: Let's make sure to make an issue about resolving the visibility of the part API.

packages/lit-html/src/lib/parts.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export const restoreNodePart = (part: NodePart, state: NodePartState) => {
+  // TODO (justinfagnani): make an interal-only interface
+  (part as any).__commitNode((state as NodePartStateInternal).__fragment);
+  part.__value = (state as NodePartStateInternal).__value;
+};
+
+const createMarker = () => document.createComment('');
+
+export const createAndInsertPart = (
```
sorvell, line 41: Why not have a createPart + insertPartBefore. Is the idea that that would do extra work v. this?

packages/lit-html/src/lib/parts.ts
```diff
@@ -0,0 +1,137 @@
...
+ * `container`.
+ */
+export const removeNodes = (
+  container: Node,
+  start: Node | null,
+  end: Node | null = null
+): void => {
+  while (start !== end) {
+    const n = start!.nextSibling;
+    container.removeChild(start!);
```
sorvell, line 134: Did we want to use `.remove()`?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -509,6 +510,312 @@ suite('lit-html', () => {
...
+
+      children = document.createTextNode('foo');
+      render(t(), container);
+      assert.equal(
+        stripExpressionComments(container.innerHTML),
+        '<div>foo</div>'
+      );
+    });
+
+    // test('removes the first element from an array', () => {
```
sorvell, line 630: Why are these still commented?

packages/lit-html/src/test/parts_test.ts
```diff
@@ -0,0 +1,80 @@
...
+import {detachNodePart, restoreNodePart} from '../lib/parts.js';
+
+suite('lit-html', () => {
+  let container: HTMLDivElement;
+
+  setup(() => {
+    container = document.createElement('div');
+  });
+
+  test('detach / restore', () => {
```
sorvell, line 33: What this is testing is not super obvious. Maybe add some more comments?

packages/lit-html/src/lib/directives/repeat.ts
```diff
@@ -0,0 +1,445 @@
...
+    let index = 0;
+    for (const item of items) {
+      this.itemKeys[index] = keyFn ? keyFn(item, index) : index;
+      values[index] = template!(item, index);
+      index++;
+    }
+    return values;
+  }
+
+  update<T>(
```
sorvell, line 89: Is this the same implementation as 1.x?

packages/lit-html/src/test/directives/repeat_test.ts
```diff
@@ -0,0 +1,698 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+// import {guard} from '../../directives/guard.js';
+import {repeat} from '../../lib/directives/repeat.js';
```
sorvell, line 16: Are these tests the same as 1.x?

sorvell (changes requested):

packages/lit-html/src/lib/parts.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+type NodePartStateInternal = {
+  __value: unknown;
+  __fragment: DocumentFragment;
+};
+
+export const detachNodePart = (part: NodePart): NodePartState => {
+  const fragment = document.createDocumentFragment();
+  const state: NodePartStateInternal = {
+    __value: part.__value,
```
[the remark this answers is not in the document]
justinfagnani: Private only applies to class fields, but it's ok because we never export this interface. We export the `NodePartState` interface which doesn't have this property.

packages/lit-html/src/lib/parts.ts
justinfagnani: I just copied over the utility functions from `repeat()`, we can definitely refine these.

packages/lit-html/src/test/lit-html_test.ts
justinfagnani: I need to remove them

packages/lit-html/src/lib/directives/repeat.ts
justinfagnani: Basically, except for the call to `this.render()`

packages/lit-html/src/test/directives/repeat_test.ts
justinfagnani: Exactly the same

packages/lit-html/src/lib/lit-html.ts
justinfagnani: https://github.com/Polymer/lit-html/issues/1261

packages/lit-html/src/lib/parts.ts
justinfagnani: done

packages/lit-html/src/test/lit-html_test.ts
justinfagnani: done

packages/lit-html/src/test/parts_test.ts
justinfagnani: Added comment

packages/lit-html/src/parts.ts
```diff
@@ -0,0 +1,117 @@
...
+): NodePart => {
+  const container = containerPart.__startNode.parentNode as Node;
+
+  const endNode =
+    refPart === undefined ? containerPart.__endNode : refPart.__startNode;
+
+  const startNode = container.insertBefore(createMarker(), endNode);
+
+  container.insertBefore(createMarker(), endNode);
+  const newPart = new NodePart(startNode, endNode, containerPart.options);
```
kevinpschaaf, line 53: Shouldn't `endNode` passed to `NodePart` be the marker created on the line above (not the `ref` node, which may be the next part's start marker)?

The `endNode` on L47 is named wrong, it should be the `refNode` used in this function for insertion, and `endNode` for this part will be the node created on L52.

packages/lit-html/src/parts.ts
```diff
@@ -0,0 +1,117 @@
...
+  return newPart;
+};
+
+export const updatePart = (part: NodePart, value: unknown) => {
```
kevinpschaaf, line 57: Just noting that from an API point of view, this method is pretty marginal; it made a little more sense in 1.x when it was set+commit (and was useful to return the part so repeat could do update+assign as one-liners, but that's really just a convenience for repeat).

I guess its real value is giving a public API for `__setValue` _just_ for `NodePart` (which we have valid use cases for); and the idea is we would not need an analog for `AttributePart` (whose `setValue` API is more complex) because you shouldn't be constructing those yourself, and thus so returning the value from `render` is sufficient?

packages/lit-html/src/parts.ts
```diff
@@ -0,0 +1,117 @@
...
+export const removeNodes = (
+  start: ChildNode | null,
+  end: ChildNode | null = null
+): void => {
+  while (start !== end) {
+    const n = start!.nextSibling;
+    start!.remove();
+    start = n;
+  }
+};
```
kevinpschaaf, line 117: Can we keep `reparentNodes` and `removeNodes` non-exported for now if we don't have a use for them outside the part helpers? Especially since this module is named `parts.ts`, the APIs should be parts-related anyway.

And should this module be named `node-part-helpers` or something, since it's all NodePart-specific?

kevinpschaaf (changes requested):

packages/lit-html/src/parts.ts
kevinpschaaf: But... we already differentiate the `__setValue` call between Attribute & NodeParts... so seems like NodePart's setValue could just be public?  Something to consider.

packages/lit-html/parts.d.ts
```diff
@@ -0,0 +1,23 @@
+import { NodePart } from './lit-html.js';
```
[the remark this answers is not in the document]
justinfagnani: I updated .gitignore, but I need to remove the files from the old commits.

packages/lit-html/src/directives/repeat.ts
```diff
@@ -0,0 +1,445 @@
...
+    // New value list is eagerly generated from items along with a
+    // parallel array indicating its key.
+    const newValues: unknown[] = [];
+    const newKeys: unknown[] = [];
+    let index = 0;
+    for (const item of items) {
+      newKeys[index] = keyFn ? keyFn(item, index) : index;
+      newValues[index] = template!(item, index);
+      index++;
+    }
```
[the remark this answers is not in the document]
justinfagnani: Done. I made a side-effect-free helper method though.

packages/lit-html/src/directives/repeat.ts
```diff
@@ -0,0 +1,445 @@
...
+        newParts[newHead] = updatePart(oldParts[oldHead]!, newValues[newHead]);
+        oldHead++;
+        newHead++;
+      } else if (oldKeys[oldTail] === newKeys[newTail]) {
+        // Old tail matches new tail; update in place
+        newParts[newTail] = updatePart(oldParts[oldTail]!, newValues[newTail]);
+        oldTail--;
+        newTail--;
+      } else if (oldKeys[oldHead] === newKeys[newTail]) {
+        newParts[newTail] = updatePart(oldParts[oldHead]!, newValues[newTail]);
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/lit-html/src/directives/cache.ts
```diff
@@ -0,0 +1,68 @@
...
+export const cache = directive(
+  class extends Directive {
+    templateCache = new WeakMap<TemplateStringsArray, NodePartState>();
+    value?: TemplateResult;
+
+    render(v: unknown) {
+      if ((v as TemplateResult)._$litType$ !== undefined) {
+        this.value = v as TemplateResult;
+      } else {
+        this.value = undefined;
```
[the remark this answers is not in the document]
justinfagnani: done

packages/lit-html/src/directives/repeat.ts
```diff
@@ -0,0 +1,445 @@
...
+    // Remove any remaining unused old parts
+    while (oldHead <= oldTail) {
+      const oldPart = oldParts[oldHead++];
+      if (oldPart !== null) {
+        removePart(oldPart);
+      }
+    }
+    // Save order of new parts for next round
+    // partListCache.set(containerPart, newParts);
+    // keyListCache.set(containerPart, newKeys);
```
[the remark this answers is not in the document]
justinfagnani: removed

packages/lit-html/src/directives/repeat.ts
```diff
@@ -0,0 +1,445 @@
...
+    containerPart: NodePart,
+    [items, keyFnOrTemplate, template]: [
+      Iterable<T>,
+      KeyFn<T> | ItemTemplate<T>,
+      ItemTemplate<T>
+    ]
+  ) {
+    // Old part & key lists are retrieved from the last update
+    // TODO: deal with directive being swapped out?
+    let oldParts = containerPart.__value as Array<NodePart | null>;
```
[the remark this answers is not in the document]
justinfagnani: Added an issue earlier, and renamed these to use a single underscore to bypass property renaming.

packages/lit-html/src/lit-html.ts
```diff
@@ -661,6 +667,54 @@ export class NodePart {
...
+    let partIndex = 0;
+    let itemPart: NodePart | undefined;
+
+    for (const item of value) {
+      if (partIndex === itemParts.length) {
+        // If no existing part, create a new one
+        itemParts.push(
+          (itemPart = new NodePart(
+            this.__insert(createMarker()),
+            this.__insert(createMarker()),
```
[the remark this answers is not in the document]
justinfagnani: done

packages/lit-html/src/parts.ts
justinfagnani: Fixed

packages/lit-html/src/parts.ts
justinfagnani: Yeah, this would let us still minify the Part API. Let's consider as part of the Part API overall.

packages/lit-html/src/parts.ts
justinfagnani: Removed exports. Not sure about the file name... I could see AttributePart helpers in the future.

justinfagnani:
PTAL

packages/lit-html/src/parts.ts
kevinpschaaf: Sorta curious how this didn't break any tests.  A `clear` would have removed the marker... seems like that should have broken something in `repeat` tests. Would be nice to make sure we have a regression test for this.

packages/lit-html/src/lib/parts.ts
kevinpschaaf: Wait, if we're using `append` in core, we might as well use `remove`.  Both will need to be in the DOM shim we talked about adding to wcjs platform.

packages/lit-html/src/directives/repeat.ts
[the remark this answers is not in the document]
kevinpschaaf: Ugh, ok, but now `render` is not used on the client at all.  I think that's also a pretty weird pattern to implement a `render` that's _never_ used. My mental model is that `render` should be used for at least first render on both client & server.

packages/lit-html/src/parts.ts
justinfagnani: Me too. I'll try to add a test. Maybe rendering a repeat() then a non-repeat value.

packages/lit-html/src/directives/repeat.ts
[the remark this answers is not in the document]
justinfagnani: Because of the key array I think there's a choice here between it being side-effectful and always called, and being side-effect-free and not called on the client. I was thinking side-effect free would be better for cross-compilation. It'd be nice to be able to just drop `@soyCompatible` on here and have it work, though admittedly this isn't soy compatible because of the local var mutation, and we need to special-case repeat() anyway to be idom compatible...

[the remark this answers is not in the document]
kevinpschaaf: Agree, although there's a 3rd option which neither of us probably like, which is to do the loop twice, once in update to get/store the keys, and another in render to get the values. That would be the pure/semantic way to do it, at the cost of two loops.

One other downside to having directives where update never calls render is that we're not actually going to test the render implementation, at least in the client-side tests (note my related [comment on styleMap](https://github.com/Polymer/lit-html/pull/1218#discussion_r488317095) where this led to a bug). Yes, once we get the SSR tests in place, _we'll_ have good test coverage, but again this is evidence to me that it's pretty easy to accidentally write non-SSR compatible/correct code, where I was hoping we'd get a much stronger guarantee.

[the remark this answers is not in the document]
justinfagnani: I don't disagree, I just don't have a better option in my head yet. I don't want to do require a side-effectful `render()` or `super.update()` (that calls `render()`) to require the call to `render()`.

In LitElement if you don't call `super.update()` your element basically won't work, which is what effectively forces the call to `render()`. Of course an element could return nothing from `render()` and do all its rendering in `update()` or `updated()` and have a bad render in SSR. So there's no guarantee, even if LitElement pushes harder to the right thing. The equivalent would be something like `Directive.update()` setting some flag that it was called. I don't really want to require that directives extend from Directive though, even if it's what we usually document.

packages/lit-html/src/parts.ts
justinfagnani: Added a test that fails on the bad impl

packages/lit-html/src/directives/repeat.ts
```diff
@@ -0,0 +1,458 @@
...
+      const oldPart = oldParts[oldHead++];
+      if (oldPart !== null) {
+        removePart(oldPart);
+      }
+    }
+
+    // Save order of new parts for next round
+    this.itemKeys = newKeys;
+    // Directly set part value, bypassing it's dirty-checking
+    containerPart._value = newParts;
```
kevinpschaaf, line 447: Add a TODO linked to #1261

kevinpschaaf: Also note that this doesn't strictly need to use `_value` to store the `newParts`. The only actual requirement is that first render try `getPartValue` to get hydrated parts; after that, it can prefer an instance variable on the part to save/retrieve the previously rendered parts.

kevinpschaaf (approved):
Ok to merge I think, making sure we have TODOs for any instances of private API (I think it's just down to the one noted below?)

packages/lit-html/src/directives/repeat.ts
justinfagnani: That's true, but might as well reuse that storage, and this makes it work that way for hydration.

PR #1251, opened 2020-09-11
[benchmarks] Replace lit-html stub benchmark with kitchen-sink

kevinpschaaf:
### Kitchen-sink benchmark for lit-html

Features exercised:
- NodePart
  - value: string
  - value: Node
  - value: Array<TemplateResult>
  - value: directive: repeat
  - value: nothing
- AttributePart
  - value: string (single)
  - value: string (multiple)
  - value: directive: classMap
  - value: directive: styleMap
- PropertyPart
  - value: object
- EventPart
  - value: function

The base benchmark file can be run with these query params:
- `width`: number of items in each map/repeat per item
- `depth`: number of levels of item recursion
- `updateCount`: number of times to update after initial render
- `updateData`: boolean, false to reuse same data each update, true to regenerate

Width & depth of 4 seems to be a good sweet spot of DOM created.

This sets up 3 tachometer configs/`tachometer-reporter-action` s reported:
- `kitchen-sink-render` initial render time (`width=4`, `depth=4`, `updateCount=0`)
- `kitchen-sink-update` update time with re-generated data (`width=4`, `depth=4`, `updateCount=10`, `updateData=true`)
- `kitchen-sink-update-nop` update time with same data (`width=4`, `depth=4`, `updateCount=10`, `updateData=true`)

**Notes**
* The tip-of-tree results will be wildly fast because most of the meat of the benchmark comes from a recursive template call via Array mapping/repeat, which isn't implemented on tip-of-tree right now (it just renders `[object Object]` for arrays and doesn't recurse).  As soon as `lit-next-directive-helpers` (#1250) lands, the benchmark results should read a little better.
* The benchmark currently uses stub implementations of `repeat`, `classMap`, and `styleMap`, which should be switched out to their real imports when those features are landed.


aomarks (approved):

packages/benchmarks/lit-html/kitchen-sink/index.ts
```diff
@@ -0,0 +1,205 @@
...
+      <div class="foo bar baz"></div>
+      <div><span></span></div>
+      <div class="foo bar baz"></div>
+      <div><span></span></div>
+      <div class="foo bar baz"></div>
+    </div>
+    <div class="foo bar baz"></div>
+    ${
+      data.childData
+        ? html`<div mode="${data.useRepeat ? 'repeat' : 'map'}">
```
justinfagnani, line 167: nice mode switch

packages/benchmarks/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
```
justinfagnani, line 3: We're using ES2019 in the other packages

packages/benchmarks/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "module": "es2015",
+    "lib": ["es2017", "esnext", "DOM", "DOM.Iterable"],
```
justinfagnani, line 5: You should need both es2017 and esnext

justinfagnani (approved):
LGTM, but try to get multi-measure in rather than running 3x

packages/benchmarks/lit-html/kitchen-sink/index.ts
```diff
@@ -0,0 +1,194 @@
...
+ * - depth: number of levels of item recursion
+ * - updateCount: number of times to update after initial render
+ * - updateData: boolean, false to reuse same data each update, true to regenerate
+ */
+
+import {html, render, nothing} from 'lit-html';
+
+// TODO(kschaaf) use real repeat once landed
+// import { repeat } from "lit-html/lib/directives/repeat.js";
+const repeat = (items: any[], _kfn: (i: any) => any, tfn: (i: any) => any) =>
```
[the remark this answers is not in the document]
kevinpschaaf: Ah cool, haven't looked since I rebased. Lemme try using that.

packages/benchmarks/tsconfig.json
kevinpschaaf: Done, copied over lit-html's config

packages/benchmarks/tsconfig.json
kevinpschaaf: Done, copied over lit-html's config

packages/benchmarks/lit-html/kitchen-sink/index.ts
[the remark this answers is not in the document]
kevinpschaaf: Well, I put it in but had to take it back out, because then the benchmark fails when run on tip-of-tree (which doesn't yet have it and so the import fails).

It raises a tricky issue with adding new features + benchmarks at the same time. As long as the features we're adding are all getting back to "feature parity", we can just add benchmarks in a following PR after the feature lands in tot.  But any actual new features can't really be put in the kitchen-sink benchmark, they'd need their own, so that we can configure it to not test against previous/tot branches.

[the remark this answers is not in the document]
justinfagnani: That makes sense and seems fine.

I don't think it's work worrying about now, but you could use dynamic import to get repeat if it's there. Maybe we use that if we need to benchmark a new feature in the future.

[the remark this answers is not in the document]
kevinpschaaf: Right I thought about that, but then the benchmark timing won't be comparable to when the feature doesn't exist.  Which is why kitchen-sink needs to be stable features.

kevinpschaaf:
@andrewiggins Any idea why the `tachometer-reporter-action` doesn't seem to be updating the comment here? It was working with v1, and then I updated to v2 and refactored the kitchen-sink benchmark to take advantage of multiple measurements, and somewhere in there it stopped working.  I thought maybe because I'd changed versions on the same PR (and tried deleting the comment) something got messed up, so I tried making another PR here: https://github.com/Polymer/lit-html/pull/1256, but that seems to have the same problem.

sorvell (approved):
Can be a follow on, but let's add the comparison against LitElement master. Even with the trivial element you have in the stub, it'd still be worthwhile.

andrewiggins:
@kevinpschaaf Ack. Will take a look

andrewiggins:
Okay, I think I have fixed the reporter. Check out [my fork of lit-html](https://github.com/andrewiggins/lit-html/pull/1) using a [branch of tachometer-reporter-action with the bug fix](https://github.com/andrewiggins/tachometer-reporter-action/pull/25) to see it fixed.

I'll publish this fix on the v2 line by tomorrow morning. Re-running the benchmark job after that publish should fix up this PR's comment

kevinpschaaf:
Thanks @andrewiggins!! 🙌

andrewiggins:
Fix is published! Try re-running the jobs and things should be fixed.

justinfagnani:
Re-ran and the comment updated. Multi-mark support is great @aomarks! Thanks for the quick fix @andrewiggins!

PR #1252, opened 2020-09-11
[lit-html] Prod and dev builds for lit-html

aomarks:
- Updated file layout to support prod vs dev builds:
  - TypeScript source is now emitted to `development/` instead of the top-level, because the top-level will now be used for the minified modules, and users can instead import from `development/` when they want the dev build (we expect using some kind of `lit-html` import statement rewriting build step, to be researched and documented).
  - Removed the `lib/` sub-directory from `src/`, and the `index.js` file (which re-exported `* from lit-html.js`). This simplifies things, since we get a flatter layout, and it matches the existing v1 file layout (e.g. we want to import directives from `lit-html/directives/foo`, not `lit-html/lib/directives/foo`).

- Updates to minified output:
  - Output to `lit-html.js` instead of `lit-html.bundled.js`
  - Add directives as entrypoints (just the one we have for now).
  - Add back the copyright header (though it is still excluded for the `checksize` script).
  - Copy `.d.ts` files so they are adjacent.
  - Emit sourcemaps, composed with the existing tsc sourcemap so that they show the original TypeScript source.

- Add `DEV_MODE` variable, which is `true` in `development/` but `false` for the minified build, and hence blocks that are conditioned on that are dead-code removed by terser. Turns out we couldn't use terser's `global_defs`, so I'm using a rollup plugin that just does a search-and-replace (more details in a comment in the code). This will be compatible with Closure `goog.define` with a trivial transform.

- Add production testing mode, which runs the same test files, but against the minified version of the lit-html libraries, by using [@rollup/plugin-alias](https://github.com/rollup/plugins/tree/master/packages/alias) to rewrite imports (btw, it's really cool that `wtr` supports rollup plugins!).

- Adds `npm run dev` script, which uses [concurrently](https://github.com/kimmobrunfeldt/concurrently#readme) to run TypeScript, Rollup, and both test configurations continuously in watch mode, from one command.

Fixes https://github.com/Polymer/lit-html/issues/1223


packages/lit-html/.gitignore
```diff
@@ -0,0 +1,2 @@
+/lit-html.*
+/directives/
```
justinfagnani, line 2: I didn't know you could do this! 👍

packages/lit-html/package.json
```diff
@@ -6,41 +6,54 @@
...
     "/directives/",
     "/src/",
-    "!/src/test/"
+    "!/src/test/",
+    "/development/",
+    "!/development/test/"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "npm run clean && tsc && rollup -c",
```
justinfagnani, line 26: I'd like to keep a quick-build script around. Maybe as `build:ts`?

packages/lit-html/rollup.config.js
```diff
@@ -13,36 +13,108 @@
...
+    //   const DEV_MODE = goog.define('lit-html.DEV_MODE', false);
+    //
+    // We can't use terser's compress.global_defs option, because it won't
+    // replace the value of a variable that is already defined in scope (see
+    // https://github.com/terser/terser#conditional-compilation). It seems to be
+    // designed assuming that you are _always_ using terser to set the def one
+    // way or another, so it's difficult to define a default in the source code
+    // itself.
+    replace({
+      'const DEV_MODE = true': 'const DEV_MODE = false',
```
justinfagnani, line 75: Does this take a regex? Maybe anchor it to the beginning of a line to speed it up?

packages/lit-html/rollup.config.js
```diff
@@ -13,36 +13,108 @@
...
+    // This plugin automatically composes the existing TypeScript -> raw JS
+    // sourcemap with the raw JS -> minified JS one that we're generating here.
+    sourcemaps(),
     terser({
       warnings: true,
       ecma: 2017,
       compress: {
         unsafe: true,
+        // An extra pass can squeeze out an extra byte or two.
+        passes: 2,
```
justinfagnani, line 86: neat

packages/lit-html/rollup.config.js
```diff
@@ -13,36 +13,108 @@
...
-        // TODO (justinfagnani): benchmark
-        // wrap_func_args: false,
       },
       mangle: {
         properties: {
           regex: /^__/,
+          // TODO A test in lit-html_test.ts calls
+          // `part.__element.dispatchEvent`, but when testing the production
+          // output, this will fail since `__element` is renamed. Is there
+          // another way to write this test, or should this property be public?
```
justinfagnani, line 99: We're going to make these public. You can do it in this PR

packages/lit-html/web-test-runner.config.js
```diff
@@ -1,15 +1,36 @@
...
+const plugins = [];
+if (process.env.TEST_PROD_BUILD) {
+  // Our test files use relative imports for the locally built lit-html modules.
+  // By default that means the ones in the development/ directory. Our
+  // production build has the same layout, just one directory up. Assume that
+  // "lit-html.js" and "directives/" module specifiers must refer to these
+  // modules, and re-write them to reach up a directory.
+  console.log('Using production build');
+  plugins.push(
+    fromRollup(alias)({
```
justinfagnani, line 14: this is awesome!

justinfagnani (approved):
Really nice!

packages/lit-html/rollup.config.js
aomarks: No, unfortunately does not seem to support regexes.

packages/lit-html/rollup.config.js
aomarks: Done (`__element` -> `element`)

packages/lit-html/package.json
aomarks: Ok, I added `build:ts` and also renamed `build:watch` to `build:ts:watch` to make it more clear which step it is.

PR #1259, opened 2020-09-14
[lit-html] Add guard() directive

justinfagnani:
Tests are failing until #1250 is merged due to lack of array support.

### DIFF
--- a/packages/lit-html/src/directives/guard.ts
+++ b/packages/lit-html/src/directives/guard.ts
@@ -12,36 +12,30 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {
-  directive,
-  Directive,
-  noChange,
-  nothing,
-  Part
-} from '../lit-html.js';
+import {directive, Directive, noChange, Part} from '../lit-html.js';
+
+// A sentinal that indicates guard() hasn't rendered anything yet
+const initialValue = {};
 
 class Guard extends Directive {
-  previousValue: unknown = nothing;
+  previousValue: unknown = initialValue;
 
   render(_value: unknown, f: () => unknown) {
     return f();
   }
 
   update(_part: Part, [value, f]: Parameters<this['render']>) {
-    console.log('update', value);
     if (Array.isArray(value)) {
       // Dirty-check arrays by item
-      if (Array.isArray(this.previousValue) &&
-          this.previousValue.length === value.length &&
-          value.every((v, i) => v === (this.previousValue as Array<unknown>)[i])) {
-        console.log('A');
+      if (
+        Array.isArray(this.previousValue) &&
+        this.previousValue.length === value.length &&
+        value.every((v, i) => v === (this.previousValue as Array<unknown>)[i])
+      ) {
         return noChange;
       }
-    } else if (
-        this.previousValue === value &&
-        (value !== undefined || this.previousValue !== nothing)) {
+    } else if (this.previousValue === value) {
       // Dirty-check non-arrays by identity
-      console.log('B');
       return noChange;
     }
 
@@ -49,7 +43,6 @@ class Guard extends Directive {
     // what the previous values were.
     this.previousValue = Array.isArray(value) ? Array.from(value) : value;
     const r = this.render(value, f);
-    console.log('R', ((r as any).values as any)[0]);
     return r;
   }
 }
@@ -57,6 +50,12 @@ class Guard extends Directive {
 /**
  * Prevents re-render of a template function until a single value or an array of
  * values changes.
+ *
+ * Values are checked against previous values with strict equality (`===`), and
+ * so the check won't detect nested property changes inside objects or arrays.
+ * Arrays values have each item checked against the previous value at the same
+ * index with strict equality. Nested arrays are also checked only by strict
+ * equality.
  *
  * Example:
  *
@@ -67,7 +66,7 @@ class Guard extends Directive {
  *   </div>
  * ```
  *
- * In this case, the template only renders if either `user.id` or `company.id`
+ * In this case, the template only rerenders if either `user.id` or `company.id`
  * changes.
  *
  * guard() is useful with immutable data patterns, by preventing expensive work
--- a/packages/lit-html/src/directives/guard.ts
+++ b/packages/lit-html/src/directives/guard.ts
@@ -1,0 +1,90 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {
+  directive,
+  Directive,
+  noChange,
+  nothing,
+  Part
+} from '../lit-html.js';
+
+class Guard extends Directive {
+  previousValue: unknown = nothing;
+
+  render(_value: unknown, f: () => unknown) {
+    return f();
+  }
+
+  update(_part: Part, [value, f]: Parameters<this['render']>) {
+    console.log('update', value);
+    if (Array.isArray(value)) {
+      // Dirty-check arrays by item
+      if (Array.isArray(this.previousValue) &&
+          this.previousValue.length === value.length &&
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+      // Dirty-check arrays by item
+      if (Array.isArray(this.previousValue) &&
+          this.previousValue.length === value.length &&
+          value.every((v, i) => v === (this.previousValue as Array<unknown>)[i])) {
+        console.log('A');
+        return noChange;
+      }
+    } else if (
+        this.previousValue === value &&
+        (value !== undefined || this.previousValue !== nothing)) {
```
jridgewell, line 42: Can you comment why this is necessary?

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+ *     ${guard([immutableItems], () => immutableItems.map(i => html`${i}`))}
+ *   </div>
+ * ```
+ *
+ * In this case, items are mapped over only when the array reference changes.
+ *
+ * @param value the value to check before re-rendering
+ * @param f the template function
+ */
+export const guard = directive(Guard);
```
jridgewell, line 90: Maybe we should call this `memoize`?

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Example:
+ *
+ * ```js
+ * html`
+ *   <div>
+ *     ${guard([user.id, company.id], () => html`...`)}
+ *   </div>
+ * ```
+ *
+ * In this case, the template only renders if either `user.id` or `company.id`
```
jridgewell, line 70:
```suggestion
 * In this case, the template only rerenders if either `user.id` or `company.id`
```

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+} from '../lit-html.js';
+
+class Guard extends Directive {
+  previousValue: unknown = nothing;
+
+  render(_value: unknown, f: () => unknown) {
+    return f();
+  }
+
+  update(_part: Part, [value, f]: Parameters<this['render']>) {
```
jridgewell, line 30: If we make the API variadic, we can simplify the implementation a little. We'd no longer need to check for array or non-array, it's just always an array of values.

kevinpschaaf: The API & name changes would probably be more of a breaking change than we're comfortable with since the public API is just following the decisions made in 1.x.

I do agree `memoize` is a more obvious naming of what the directive does.

Re: the parameters to memoize against, I recall discussing this before the 1.x release and believe there was pushback on using flat vargs due to the template function being a different type of arg, and an aversion to using vargs where e.g. "first/last arg is special".

packages/lit-html/src/directives/guard.ts
justinfagnani: That'd be more of a breaking change than we want to make: 2.x should be a drop-in replacement for most users. We went with guard initially because it's not technically memoization, and "memoization" itself is a very technical term, so it seemed bad to use it to invoke a specific CS technique, and then not have it use that technique.

kevinpschaaf: I almost agreed, but then remembered the nuance between what this actually does and the notion of memoize:
* This actually returns `noChange` when the arguments match, telling the underlying layer to do nothing
* The typical contract of a memoizer would be that it _return_ the same value for the same arguments (and possibly build up a cache of arguments -> result); if we did it this way, it would be relying on the underlying layer to do something efficient when it gets the same value

Both would effectively bypass the work of the template function on subsequent renders, but `guard` also bypasses the underlying value dirty-check.

kevinpschaaf (approved):
LGTM, just remove the console.logs

packages/lit-html/src/directives/guard.ts
sorvell: I also think passing the values as var args makes sense. It's pretty convenient with a rest arg as well. Perhaps that can be addressed in a separate PR that refines the directive API.

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+    } else if (
+        this.previousValue === value &&
+        (value !== undefined || this.previousValue !== nothing)) {
+      // Dirty-check non-arrays by identity
+      console.log('B');
+      return noChange;
+    }
+
+    // Copy the value if it's an array so that if it's mutated we don't forget
```
sorvell, line 48: Any Array/Object values are, however, only checked by identity. Are the docs clear on that?

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Example:
+ *
+ * ```js
+ * html`
+ *   <div>
+ *     ${guard([immutableItems], () => immutableItems.map(i => html`${i}`))}
+ *   </div>
+ * ```
+ *
+ * In this case, items are mapped over only when the array reference changes.
```
sorvell, line 85: Maybe add something like: "If an array of values is given, any of its values are checked by identity not deeply for equality."

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+      // Dirty-check non-arrays by identity
+      console.log('B');
+      return noChange;
+    }
+
+    // Copy the value if it's an array so that if it's mutated we don't forget
+    // what the previous values were.
+    this.previousValue = Array.isArray(value) ? Array.from(value) : value;
+    const r = this.render(value, f);
+    console.log('R', ((r as any).values as any)[0]);
```
sorvell, line 52: More `console.log`'s to remove...

packages/lit-html/src/directives/guard.ts
sorvell: Yeah, the checks other than the equality check seem unnecessary?

sorvell: What should happen if you explicitly give a value of `noChange`?

sorvell (changes requested):

packages/lit-html/src/directives/guard.ts
justinfagnani: This was a vestige from the old Directive API where we didn't have a constructor. I removed it.

packages/lit-html/src/directives/guard.ts
justinfagnani: done

packages/lit-html/src/directives/guard.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  update(_part: Part, [value, f]: Parameters<this['render']>) {
+    console.log('update', value);
+    if (Array.isArray(value)) {
+      // Dirty-check arrays by item
+      if (Array.isArray(this.previousValue) &&
+          this.previousValue.length === value.length &&
+          value.every((v, i) => v === (this.previousValue as Array<unknown>)[i])) {
+        console.log('A');
```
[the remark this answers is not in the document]
justinfagnani: done

packages/lit-html/src/directives/guard.ts
justinfagnani: I think we're talking about two difference things here. One is whether `Directive.update()` is variadic, and the other is whether `guard()` is.

I think @jridgewell is saying that if we made the `guard()` API _non-variadic_ - required the values to always be an array - that we could eliminate the `Array.isArray()` check. I agree on that. If I remember, we added arrays of values only later, so we preserved the single-value form for backwards compatibility. The only question for me is breaking change budget and the subtleness of the potential break when giving iterables as the single-value. If a string is passed as the value, we could easily iterate over each character on each render. We'd probably still want the `Array.isArray` check as an assertion.

For the `update()` API, I worry that making it variadic would lock us into these and only these arguments. For instance, we couldn't safely add an options argument. We'd have to pipe in any additional information via the part or another call like the constructor. Of course `render()` is variadic, with the idea that we want that call to have less information, and it makes it much easier for the author to automatically declare the signature of the directive function.

I can open separate issues for both things.

packages/lit-html/src/directives/guard.ts
justinfagnani: Updated the docs a bit

packages/lit-html/src/directives/guard.ts
justinfagnani: done

packages/lit-html/src/directives/guard.ts
justinfagnani: done

sorvell (approved):

PR #1265, opened 2020-09-14
[lit-html, lit-element] Adds `refNode` to `RenderOptions`

sorvell:
* `render` now takes an optional `refNode` in render options. If specified, content is rendered before the given node.
* LitElement now renders content before any content existing in its shadowRoot. This fixes an issue with shimming `adoptedStyleSheets` on browsers without native support. The adopted styles will now (again, since it was like this in 2.x) be rendered last.


packages/lit-element/package.json
```diff
@@ -39,30 +39,31 @@
   },
   "author": "The Polymer Authors",
   "devDependencies": {
-    "@babel/cli": "^7.2.3",
```
kevinpschaaf, line 42: Any reason why `package.json` changed?

kevinpschaaf (approved):
LGTM, just maybe revert the package.json change since seems unintended?

packages/lit-element/package.json
kevinpschaaf: Oh, this also looks like why CI is failing (`package.json` & `package-lock.json` disagreement)

packages/lit-html/src/lit-html.ts
```diff
@@ -233,9 +237,10 @@ export const render = (
 ) => {
   let part: NodePart = (container as any).$lit$;
   if (part === undefined) {
-    const marker = createMarker();
-    container.append(marker);
-    (container as any).$lit$ = part = new NodePart(marker, null, options);
+    const start = createMarker();
+    const end = options?.refNode || null;
+    container.insertBefore(start, end);
+    (container as any).$lit$ = part = new NodePart(start, end, options);
```
justinfagnani, line 243: Since Terser won't know that insertBefore returns the new node, we can manually inline it and save a couple of bytes when it (hopefully) inlines the `start` variable because it's only used once:

```ts
(container as any).$lit$ = part = new NodePart(container.insertBefore(start, end), end, options);
```

packages/lit-html/src/lit-html.ts
```diff
@@ -233,9 +237,10 @@ export const render = (
 ) => {
   let part: NodePart = (container as any).$lit$;
```
justinfagnani, line 238: When given a `refNode` we should store the part on it instead of the container. That way we could render multiple times to the same container. I think it'd be confusing if a second call to render with a different `refNode` only updated the previous call's DOM.

justinfagnani (changes requested):

packages/lit-element/package.json
sorvell: Fixed.

packages/lit-element/src/lib/lit-element.ts
```diff
@@ -258,7 +258,7 @@ export class LitElement extends UpdatingElement {
       (this.constructor as typeof LitElement).render(
         templateResult,
         this.renderRoot,
-        {eventContext: this}
+        {eventContext: this, refNode: this.renderRoot.firstChild}
```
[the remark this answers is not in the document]
sorvell: Suggest we consider reconciling that when we port over the SSR work.

packages/lit-html/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Multiple bindings in an attribute value don't need to be in quotes
 
 ### Added
+* Added `refNode` to render options. If specified, content is rendered before the node given via render options, e.g. `{refNode: node}`.
```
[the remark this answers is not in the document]
sorvell: I don't think so. I think normally it is logical to specify a container and when you want to append, that's all you could specify. The `refNode` is for when you want more control over where the content goes.

packages/lit-element/package.json
justinfagnani: I'm still seeing the package.json changes. Why are those in these PR?

packages/lit-html/src/lit-html.ts
justinfagnani: I think this should be addresses. This will give unexpected results, IMO:

```ts
render(templ1(), container, {refNode: a});
render(templ2(), container, {refNode: b});
```

justinfagnani (changes requested):

packages/lit-element/package.json
sorvell: I sync'd the package.json with lit-html so that we can test multiple browsers.

packages/lit-html/src/lit-html.ts
```diff
@@ -218,6 +218,10 @@ export interface RenderOptions {
    * useful to set this to the host component rendering a template.
    */
   readonly eventContext?: EventTarget;
+  /**
+   * A DOM node before which to render content in the container.
+   */
+  readonly refNode?: ChildNode|null;
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-html/src/lit-html.ts
sorvell: Done, test added.

packages/lit-html/src/lit-html.ts

packages/lit-html/src/lit-html.ts
```diff
@@ -262,11 +266,14 @@ export const render = (
   container: HTMLElement | DocumentFragment,
   options?: RenderOptions
 ) => {
-  let part: NodePart = (container as any).$lit$;
+  let part: NodePart = (options?.renderBefore ?? (container as any)).$lit$;
```
justinfagnani, line 269: why does `container` need the cast, but not `options?.renderBefore`?

packages/lit-html/src/lit-html.ts
```diff
@@ -262,11 +266,14 @@ export const render = (
...
   if (part === undefined) {
-    const marker = createMarker();
-    container.append(marker);
-    (container as any).$lit$ = part = new NodePart(marker, null, options);
+    const end = options?.renderBefore ?? null;
+    (container as any).$lit$ = part = new NodePart(
```
justinfagnani, line 272: Doesn't this need to change too?

I think we want something like:

```ts
const partOwnerNode = options?.renderBefore ?? container;
let part = (partOwnerNode as any).$lit$;
if (part === undefined) {
  const endNode = options?.renderBefore ?? null;
  (partOwnerNode as any).$lit$ = ...
}
```

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -392,6 +397,33 @@ suite('lit-html', () => {
...
+    });
+
+    test('renders before `renderBefore`, if specified', () => {
+      container.appendChild(document.createElement('div'));
+      assertRender(html`<span></span>`, '<span></span><div></div>', {
+        renderBefore: container.firstChild,
+      });
+    });
+
+    test('renders same template before different `renderBefore` nodes', () => {
```
justinfagnani, line 413: I think this test needs an update to catch the problem with storing the new part. The current code will create a new part every render.

justinfagnani (changes requested):

packages/lit-html/src/test/lit-html_test.ts
sorvell: Done.

packages/lit-html/src/lit-html.ts
sorvell: Oops, thanks.

packages/lit-html/src/lit-html.ts
sorvell: Seems like TS just not noticing that. Fixing.

justinfagnani (approved):

PR #1267, opened 2020-09-15
[lit-element] Prod and dev builds for lit-element

aomarks:
Extension of lit-html's https://github.com/Polymer/lit-html/pull/1252 for lit-element.

- Adds `common/` package for misc monorepo-local utilities.

- Replaces `@rollup/plugin-alias` with a new custom rollup plugin at `common/rollup-resolve-remap.js`. `@rollup/plugin-alias` is not really very good at remapping *path* based module specifiers, because it just supports search/replace on the raw module specifier. That means relative paths at different levels in the tree are difficult to handle. This new plugin _first_ does a standard module resolution, and _then_ applies substitutions, so that we are dealing with normalized file paths, and can easily manipulate the directory structure.

- Also adds `common/wtr-config.js` with some common module-remapping configuration for use in all our packages.

- Updated `lit-element` file layout to support prod vs dev builds:
  - TypeScript source is now emitted to `development/` instead of the top-level, because the top-level will now be used for the minified modules, and users can instead import from `development/` when they want the dev build (we expect using some kind of import statement rewriting build step, to be researched and documented).
  - Removed `index.js` file (which re-exported `* from lit-element.js`) and didn't seem to serve much purpose (it doesn't exist in v2).

- Updates to `lit-element` minified output:
  - Output to `lit-element.js` instead of `lit-element.bundled.js`
  - Add `updating-element.js`, `css-tag.js`, `decorators.js`, and each individual decorator module as their own entrypoints. 
  - Add back the copyright header (though it is still excluded for the `checksize` script).
  - Copy `.d.ts` files so they are adjacent.
  - Emit sourcemaps, composed with the existing tsc sourcemap so that they show the original TypeScript source.

- Add `DEV_MODE` variable, which is `true` in `development/` but `false` for the minified build, and hence blocks that are conditioned on that are dead-code removed by terser. This will be compatible with Closure `goog.define` with a trivial transform.

- Add production testing mode, which runs the same test files, but against the minified version of the `lit-element` and `lit-html` libraries to rewrite imports.

- Adds `npm run dev` script, which uses [concurrently](https://github.com/kimmobrunfeldt/concurrently#readme) to run TypeScript, Rollup, and both test configurations continuously in watch mode, from one command.

Fixes https://github.com/Polymer/lit-html/issues/1254


sorvell (changes requested):
Thanks, this looks pretty good. It seems like the `src/lib` folder is not helping anything. Wdyt about removing that completely?

aomarks:
> Thanks, this looks pretty good. It seems like the `src/lib` folder is not helping anything. Wdyt about removing that completely?

I was thinking it would be another breaking change for anyone who imported directly from `lib/` in `v2`, or was that never really supposed to be done?

aomarks:
> I was thinking it would be another breaking change for anyone who imported directly from `lib/` in `v2`, or was that never really supposed to be done?

In google3, looks like there are lots of imports of the form `lit-element/lib/updating-element` and `lit-element/lib/decorators` etc., even though `lit-element` re-exports those, so I'm guessing a lot of external users are doing the same thing, and so we should probably not change the file layout.

sorvell (approved):
Looks good. I still want to revisit changing the file layout, but let's consider that separately.

packages/common/rollup-resolve-remap.js
```diff
@@ -0,0 +1,52 @@
...
+
+const pathLib = require("path");
+
+/**
+ * Rollup plugin that remaps import module specifiers.
+ *
+ * Similar to @rollup/plugin-alias, but different in that it works better with
+ * relative path module specifiers, not just bare module specifiers.
+ *
+ * Instead of applying substitutions directly to the original module specifier
```
justinfagnani, line 23: Can you file a feature request against `@rollup/plugin-alias`?

packages/common/rollup-resolve-remap.js
```diff
@@ -0,0 +1,52 @@
...
+ * always normalized and relative to a given root directory.
+ */
+module.exports = function resolveRemap({ root, remap }) {
+  return {
+    name: "resolve-remap",
+    async resolveId(importee, importer) {
+      const resolved = await this.resolve(importee, importer, {
+        skipSelf: true,
+      });
+      if (!resolved || !resolved.id) {
```
justinfagnani, line 36:
```suggestion
      if (!resolved?.id) {
```

packages/common/rollup-resolve-remap.js
```diff
@@ -0,0 +1,52 @@
...
+        return null;
+      }
+      const { id: fullResolvedPath } = resolved;
```
justinfagnani, line 39: Nit: For one variable with a rename, I think it's more readable w/o the destructuring:

```ts
const fullResolvedPath = resolved.id;
```

packages/lit-element/src/lit-element.ts
```diff
@@ -54,13 +54,18 @@
...
 
-export * from './updating-element.js';
+export * from './lib/updating-element.js';
 export {html, svg, TemplateResult} from 'lit-html';
-export * from './css-tag.js';
+export * from './lib/css-tag.js';
+
+const DEV_MODE = true;
+if (DEV_MODE) {
+  console.log('lit-element is in dev mode. Not recommended for production!')
```
justinfagnani, line 67: One thing I noticed with this is that it prints during tests for every test file we have, which is just a bit of an annoyance. I wonder if we could skip this message during tests?

justinfagnani (approved):

packages/common/rollup-resolve-remap.js
aomarks: Done.

packages/common/rollup-resolve-remap.js
aomarks: Done.

packages/common/rollup-resolve-remap.js
aomarks: Will do.

packages/lit-element/src/lit-element.ts
aomarks: We could add a rollup plugin that suppresses the log statements I guess? Punting for now.

PR #1283, opened 2020-09-17
[lit-next] Unify lit-html and lit-element test runners

aomarks:
Moves test configs and dependencies (but not the test code itself) from `lit-html` and `lit-element` into a unified `tests` package.

This should speed up testing since we'll be launching fewer browsers. This will be much more significant once we add Sauce browsers.

It's still possible to run just specific package tests by passing the file names, and this is what `npm test` will continue to do when you run them from the specific packages. The top-level repo `npm test` script runs the combined tests, as does CI.

Also:

- Removed some unused dependencies.

- Removed `--concurrency=1` from `lerna bootstrap` command, since we now only depend on `playwright` in one sub-package, so we won't hit https://github.com/microsoft/playwright/issues/3912 anymore.

- Added a `wtr` plugin that strips out the `console.warn('lit-html is in dev mode. Not recommended for production!');` statements (kinda hacky, but works until https://github.com/modernweb-dev/web/issues/595 is implemented).

- Converted the common test utilities to modules.

- Some weird symlink hack in the `tests/` directory for the moment, due to https://github.com/modernweb-dev/web/issues/593.


packages/tests/web-test-runner.config.js
```diff
@@ -0,0 +1,48 @@
+import { playwrightLauncher } from "@web/test-runner-playwright";
+import { fromRollup } from "@web/dev-server-rollup";
+import { resolveRemap } from "./rollup-resolve-remap.js";
+import { prodResolveRemapConfig, devResolveRemapConfig } from "./wtr-config.js";
+
+const removeDevModeLogging = {
```
justinfagnani, line 6: thanks!

justinfagnani (approved):

LarsDenBakker:
@aomarks we do have an option to filter based on log level, but a filter function would be useful for a case like this.

aomarks:
> @aomarks we do have an option to filter based on log level, but a filter function would be useful for a case like this.

Sweet, thanks for the feature request. Added comment.

PR #1287, opened 2020-09-18
Add templateContent() directive

justinfagnani:


packages/lit-html/src/directives/template-content.ts
```diff
@@ -0,0 +1,43 @@
...
+    super();
+    if (part.type !== NODE_PART) {
+      throw new Error('templateContent can only be used in text bindings');
+    }
+  }
+
+  render(template: HTMLTemplateElement) {
+    if (this.__previousTemplate === template) {
+      return noChange;
+    }
```
jridgewell, line 30: Should we be supporting `noChange` and `nothing` values, too?

jridgewell (approved):

sorvell (approved):

packages/lit-html/src/directives/template-content.ts
justinfagnani: Let me check in with the users who requested this. I suspect the more valuable value would be `null` since you might try to get a template out of the DOM and if it's not there the ref will be null.

We can add more value types at any time though, and this matches 1.x, so I'm ok with just this for now.

PR #1290, opened 2020-09-19
Fix benchmark measurements

justinfagnani:
This PR makes sure there's only one performance entry with the name specified in the Tachometer config. It seems like there's a Tachometer issue in here, where it should disambiguate between marks and measures.

### DIFF
--- a/packages/benchmarks/lit-element/stub1/stub1.html
+++ b/packages/benchmarks/lit-element/stub1/stub1.html
@@ -10,13 +10,13 @@
   customElements.define('x-greeter', XGreeter);
 
   (async () => {
-    performance.mark('start');
+    performance.mark('render-start');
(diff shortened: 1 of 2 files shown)


aomarks:
> This PR makes sure there's only one performance entry with the name specified in the Tachometer config. It seems like there's a Tachometer issue in here, where it should disambiguate between marks and measures.

I think marks and measures are both valid things to retrieve (see https://github.com/Polymer/tachometer#performance-api), the issue is that we grab the first mark or measure with a given name (https://github.com/Polymer/tachometer/blob/master/src/measure.ts#L69). We could add a syntax to distinguish between measurement types to disambiguate, but I think it might be better to just print a warning if we find more than one mark/measure for a given name.

aomarks (approved):

PR #1292, opened 2020-09-20
[lit-next] Unify Prettier dependency to the root

abdonrd:
Unify Prettier dependency to the root.
Also format all files in all packages.

I split the exec command (`npm run format`) in a commit in case we want to delete it and run it at another moment.


package.json
```diff
@@ -2,10 +2,18 @@
...
   "dependencies": {
     "lerna": "^3.22.1"
+  },
+  "devDependencies": {
+    "prettier": "^2.1.2"
+  },
+  "prettier": {
+    "singleQuote": true,
+    "bracketSpacing": false
   }
```
abdonrd, line 18: Now just one config! 🎉

package.json
```diff
@@ -2,10 +2,18 @@
...
     "build": "lerna run build",
+    "format": "lerna run format",
     "test": "lerna run test --scope=tests",
     "benchmarks": "cd packages/benchmarks && npm run benchmarks"
   },
   "dependencies": {
     "lerna": "^3.22.1"
+  },
+  "devDependencies": {
+    "prettier": "^2.1.2"
```
abdonrd, line 13: Now just one Prettier dependency! 🎉

justinfagnani (changes requested):

package.json
abdonrd: What do you mean? Move the config to a `.prettierrc.json` file?

abdonrd:
Config moved to a `.prettierrc.json` file!

abdonrd:
@aomarks merged with `master` and ready to review!

aomarks (approved):

PR #1297, opened 2020-09-20
[lit-next] Unify ESLint dependency to the root

abdonrd:
- [x] Move the ESLint config and dependencies to the root package.
- [x] Remove all the TSLint references.
- [x] Remove unneded disabled global rules.


package.json
```diff
@@ -2,10 +2,74 @@
...
+          "argsIgnorePattern": "^_"
+        }
+      ],
+      "@typescript-eslint/no-use-before-define": "off",
+      "@typescript-eslint/triple-slash-reference": "off",
+      "no-empty": "off",
+      "no-extra-boolean-cast": "off",
+      "no-prototype-builtins": "off",
+      "no-unexpected-multiline": "off",
+      "no-var": "off"
```
abdonrd, line 53: In some moment we should try to fix errors and re-enable the disabled rules. 🙂

abdonrd:
@aomarks ready for review! 🎉

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -577,14 +576,14 @@ export abstract class UpdatingElement extends HTMLElement {
     const ctor = this.constructor as typeof UpdatingElement;
     // Note, hint this as an `AttributeMap` so closure clearly understands
     // the type; it has issues with tracking types through statics
-    // tslint:disable-next-line:no-unnecessary-type-assertion
```
aomarks, line 580: I think we could keep the `no-unnecessary-type-assertion` rule, and remove the cast here.

packages/benchmarks/lit-html/kitchen-sink/index.ts
```diff
@@ -114,6 +114,7 @@ const generateData = (
       'border-left-width': '0px',
       'border-right-width': '0px',
     },
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
```
aomarks, line 117: Seems like we need a lot of exceptions to `@typescript-eslint/no-empty-function`, and I think they're all valid. I think we should probably disable this rule.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -577,14 +576,14 @@ export abstract class UpdatingElement extends HTMLElement {
...
     const propName = (ctor._attributeToPropertyMap as AttributeMap).get(name);
     // Use tracking info to avoid reflecting a property value to an attribute
     // if it was just set because the attribute changed.
     if (propName !== undefined && this._reflectingProperty !== propName) {
       const options = ctor.getPropertyOptions(propName);
       const converter = options.converter;
       const fromAttribute =
+        // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
         (converter as ComplexAttributeConverter)?.fromAttribute! ??
```
aomarks, line 587: Looks like we can remove the `!` here.

aomarks (changes requested):
Thanks for this!

packages/lit-element/src/env.d.ts
```diff
@@ -25,6 +25,7 @@ interface ShadowRoot {
   adoptedStyleSheets: CSSStyleSheet[];
 }
 
+// eslint-disable-next-line no-var
```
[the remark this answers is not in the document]
abdonrd: Ok! I will update it!

packages/lit-element/src/lib/decorators/base.ts
```diff
@@ -13,7 +13,6 @@
  */
 
 export type Constructor<T> = {
-  // tslint:disable-next-line:no-any
```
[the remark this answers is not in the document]
abdonrd: `no-any` was already set as warning in ESLint config.
But the `lit-element` package was using TSLint.

packages/lit-element/src/lib/updating-element.ts
abdonrd: Yeah! I will update it!

packages/lit-element/src/lib/decorators/base.ts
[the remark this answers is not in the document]
aomarks: Ah I see. I don't think we've been consistently running lint anyway, since it's not in our CI yet, so this may not be very intentional. I would suggest keeping it an error, and restoring the exceptions for now. If we do turn on lint enforcement, I think introducing new `any`s should be an error.

@justinfagnani any thoughts on this?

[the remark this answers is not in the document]
abdonrd: And I imagine that the ideal would be to try to eliminate all the anys at some point (maybe better in other PRs), but as you prefer. 🙂

packages/lit-element/src/lib/updating-element.ts
abdonrd: It is not necessary with ESLint.

abdonrd:
@aomarks comments addressed!

aomarks (approved):

PR #1298, opened 2020-09-20
[lit-next] Unify the both check-version-tracker scripts

abdonrd:
Extracted from https://github.com/Polymer/lit-html/pull/1297

### DIFF
--- a/packages/lit-html/check-version-tracker.cjs
+++ b/packages/lit-html/check-version-tracker.cjs
@@ -12,14 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as fs from 'fs';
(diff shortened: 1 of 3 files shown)


packages/lit-html/check-version-tracker.cjs
```diff
@@ -12,14 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as fs from 'fs';
-import * as path from 'path';
+const fs = require('fs');
```
justinfagnani, line 15: I want to keep our utility scripts as modules

justinfagnani:
Thanks for the PR. I had to do this same thing real quick to publish the pre-release, so it's already in.

PR #1299, opened 2020-09-21
[lit-next] Lit next benchmark lit element

sorvell:
* Adds a list benchmark for LitElement
* Adds a similar list benchmark for UpdatingElement

The benchmark has 3 levels of element composition, and flows data through to create/update ~1750 elements total.


packages/benchmarks/lit-element/list/index.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+  return data;
+}
+
+const data = generateData(250);
+const otherData = generateData(500).slice(250);
+
+const propertyOptions: PropertyDeclaration = {};
+
+let updates: Promise<unknown>[] = [];
```
justinfagnani, line 41: We should figure out a more general pattern that works without this registering of promises. In the Shack benchmark @aomarks had just use a rAF to measure the end of rendering: https://github.com/Polymer/benchmarks/blob/master/lit-element/shack/index-2.js#L21

because the rAF callback goes before style and layout, it was a pretty good measure. If the benchmark takes longer than 16ms, the the callback seemed like it was firing as soon as possible.

.github/workflows/benchmarks.yml
```diff
@@ -57,17 +57,32 @@ jobs:
           pr-bench-name: this-change
           base-bench-name: tip-of-tree
 
-      - name: Benchmark lit-element/stub1
+      - name: Benchmark lit-element/list
```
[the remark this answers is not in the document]
sorvell: Let's get a sense of how long it takes to run em then decide.

packages/benchmarks/lit-element/list/index.ts
```diff
@@ -0,0 +1,262 @@
+import {
```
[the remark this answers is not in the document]
sorvell: Done.

packages/benchmarks/lit-element/list/index.ts
```diff
@@ -0,0 +1,262 @@
...
+    container.innerHTML = '';
+  };
+
+  const benchmark = params.benchmark;
+  const start = 'start';
+
+  // Initial Render
+  let test = 'render';
+  if (benchmark === test || !benchmark) {
+    performance.mark(`${test}${start}`);
```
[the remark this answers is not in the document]
sorvell: Done.

packages/benchmarks/lit-element/list/index.ts
sorvell: Done.

packages/benchmarks/lit-element/list/index.ts
```diff
@@ -214,49 +178,68 @@ export class XApp extends MonitorUpdate {
...
+  const update = async () => {
+    const updateCount = 6;
+    const test = 'update';
+    if (benchmark === test || !benchmark) {
+      el = create();
+      const start = getTestStartName(test);
+      performance.mark(start);
+      for (let i = 0; i < updateCount; i++) {
+        el.items = i % 2 ? otherData : data;
+        await updateComplete();
```
justinfagnani, line 213: By awaiting rAFs sequentially, this measure _will_ include style/layout/paint. That's ok for some benchmarks, but assuming we're doing the minimal DOM updates, we may want to isolate the LitElement+lit-html work. I think that would require marking the start time at the beginning of the loop body, then adding up the individual times manually.

kevinpschaaf: Unfortunately, tachometer doesn't currently support this. I hit this on the repeat benchmark and opened https://github.com/Polymer/tachometer/issues/196.

Seems like we need to decide if the LitElement benchmarks should care about style/layout/paint or just user JS time.  Just measuring JS time seems likely more reliable, so it's a question of what we might miss. Given a certain tree size & choice of styling, do we expect implementation choices to affect style/layout/paint?  If we stay with rAF, we should make sure the time under test is significantly >16ms.

packages/benchmarks/lit-element/list/index.ts
```diff
@@ -0,0 +1,245 @@
...
+function generateData(count: number) {
+  let data = [];
+  for (let i = 0; i < count; i++) {
+    data.push(makeItem(i));
+  }
+  return data;
+}
+
+const data = generateData(250);
+const otherData = generateData(500).slice(250);
```
kevinpschaaf, line 48: Might be nice to pull all the constants used in the benchmark up to the top, just so it's easier to grok/tinker with. There's `updateCount` below also, maybe others?

kevinpschaaf (approved):
Approved, with minor nits for consideration.

packages/benchmarks/lit-element/list/index.ts
```diff
@@ -0,0 +1,245 @@
...
+  .slice(1)
+  .split('&')
+  .map((p) => p.split('='))
+  .reduce((p: {[key: string]: any}, [k, v]) => ((p[k] = v || true), p), {});
+
+type SimpleItem = {[index: string]: string};
+
+function makeItem(prefix: number) {
+  let o: SimpleItem = {};
+  for (let i = 0; i < 99; i++) {
```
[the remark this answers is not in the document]
sorvell: Yeah

packages/benchmarks/lit-element/list/index.ts
sorvell: Done.

PR #1302, opened 2020-09-21
[lit-html] Fix scanning of unquotted attributes

justinfagnani:


jridgewell (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -400,18 +392,13 @@ class Template {
...
-        attrNames.push(attrName!);
-        html +=
-          s.slice(0, attrNameEnd) + '$lit$' + s.slice(attrNameEnd) + marker;
-      } else {
-        html += regex === textRegex ? s + nodeMarker : s + marker;
-      }
+      html +=
+        regex === textRegex
+          ? s + nodeMarker
+          : (attrNameEnd !== -1
```
jridgewell, line 398: I think we'd need to reset the `attrNameEnd`, or inline it in the for-loop? Else a `foo="${part1} ${part2} ${part3}"` might get weird. I think `attrNames` will contain `['foo', ' ', ' ']`, and the `<template>` will have `$lit$`s where it shouldn't.

justinfagnani: `attrNameEnd` is a loop-local variable, we had mistaken code in here assuming it wasn't 🤫

jridgewell: Ohh, for some reason it appeared above the for-loop for me. This should be fine, then.

justinfagnani: Yeah, we had the same confusion for a second

sorvell: Let's add a comment explaining this change. We discussed refining this more and the comment will probably help us do that.

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -324,10 +325,20 @@ suite('lit-html', () => {
     //   );
     // });
 
-    // test('inside start tag after unquoted attribute', () => {
-    //   // prettier-ignore
-    //   assertRender(html`<div a=b ${attr`c="d"`}></div>`, '<div a="b" c="d"></div>');
-    // });
+    test('inside start tag after quoted attribute', () => {
+      // prettier-ignore
```
sorvell, line 329: Maybe ignore this whole file?

packages/lit-html/src/lit-html.ts
```diff
@@ -399,14 +395,13 @@ class Template {
...
-      // console.assert(!(attrNameEnd !== -1 && regex === textRegex));
-      if (attrNameEnd !== -1) {
-        attrNames.push(attrName!);
-        html +=
-          s.slice(0, attrNameEnd) + '$lit$' + s.slice(attrNameEnd) + marker;
-      } else {
-        html += regex === textRegex ? s + nodeMarker : s + marker;
-      }
+      html +=
+        regex === textRegex
```
kevinpschaaf, line 399: As discussed, you might have a hanging `attrNameEnd` with other states (comment or raw text element), need to not emit the attribute in those states too.  So need to add tests for those, and either make sure `attrNameEnd` is always cleared when exiting attribute/tag position, or change the test here.

kevinpschaaf (commented):
Capturing what we discussed irl

packages/lit-html/src/lit-html.ts
justinfagnani: That needed to be cleaned up to handle raw text elements. Added tests.

packages/lit-html/src/lit-html.ts
```diff
@@ -104,7 +104,7 @@ const comment2Regex = />/g;
  *    * (') then any non-(')
  */
 const tagRegex = new RegExp(
-  `>|${SPACE_CHAR}(${NAME_CHAR}+)(${SPACE_CHAR}*=${SPACE_CHAR}*(?:${ATTR_VALUE_CHAR}|("|')|$))`,
+  `>|${SPACE_CHAR}(${NAME_CHAR}+)(${SPACE_CHAR}*=${SPACE_CHAR}*(?:${ATTR_VALUE_CHAR}|("|')|))`,
```
sorvell, line 107: The comments for these regex's should very clearly describe the capturing groups since they are critical to the code below,

sorvell: Is it true that `ATTR_VALUE_CHAR` can include `/`?
Is it true that ` NAME_CHAR` can include `<`?
Are there tests for these?

packages/lit-html/src/lit-html.ts
```diff
@@ -366,17 +371,21 @@ class Template {
             regex = comment2Regex;
           } else if (match[2] !== undefined) {
             if (rawTextElement.test(match[2])) {
-              regex = new RegExp(`<\/${match[2]}`, 'g');
-            } else {
-              regex = tagRegex;
+              // Record if we encounter a raw-text element. We'll switch to
+              // this regex at the end of the tag
+              rawTextTag = new RegExp(`<\/${match[2]}`, 'g');
```
sorvell, line 376: This change seems to imply that you could see the end of the tag without actually being at the end. What case is that? Maybe put an example in the comment?

packages/lit-html/src/lit-html.ts
```diff
@@ -366,17 +371,21 @@ class Template {
             regex = comment2Regex;
           } else if (match[2] !== undefined) {
             if (rawTextElement.test(match[2])) {
-              regex = new RegExp(`<\/${match[2]}`, 'g');
-            } else {
-              regex = tagRegex;
+              // Record if we encounter a raw-text element. We'll switch to
```
sorvell, line 374: I understand that the matches are not named to save code, but I suggest having comments above each section clearly labelling the meaning of each match.

packages/lit-html/src/lit-html.ts
sorvell: Suggest adding a comment explaining what's happening here with some examples.

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
sorvell: Looks like it: `<input foo<bar=a/b>` is valid. Can add tests in a follow on PR.

packages/lit-html/src/lit-html.ts
sorvell: OK, `rawTextTag` indicates we are inside a script|style|textarea and moving right to this would skip parsing the attributes inside the tag itself.

packages/lit-html/src/lit-html.ts
```diff
@@ -366,17 +371,21 @@ class Template {
...
             // dynamic tag name
             regex = tagRegex;
           }
         } else if (regex === tagRegex) {
           if (match[0] === '>') {
-            regex = textRegex;
+            // End of a tag. If we had started a raw-text element, use that
+            // regex
+            regex = rawTextTag ?? textRegex;
+            attrNameEnd = -1;
```
sorvell, line 388: Suggest adding a comment about why it's important to reset this here exactly.

packages/lit-html/src/lit-html.ts
sorvell: If changing the ordering of the conditions is important, call that out in the comment.

packages/lit-html/src/lit-html.ts
```diff
@@ -350,63 +367,96 @@ class Template {
...
         match = regex.exec(s);
         if (match === null) {
-          // TODO (justinfagnani): add test coverage from spread parts
-          // If we're not in a quoted attribute value, make sure we clear
-          // the attrNameEnd marker
-          if (regex !== singleQuoteAttr && regex !== doubleQuoteAttr) {
-            attrNameEnd = -1;
+          // If the current regex doesn't match we've come to a binding inside
+          // that state and must break and insert a marker
+          if (regex === tagEndRegex) {
```
justinfagnani, line 372: This is a better check. The only time we could have a positive `attrNameEndIndex` is when regex was `tagEndRegex`, `doubleQuoteAttrEndRegex`, or `singleQuoteAttrEndRegex`. The assertion after the while loop checks that.

packages/lit-html/src/lit-html.ts
justinfagnani: Added consts which should be inlined by Terser. We'll need to keep an eye on the output though.

packages/lit-html/src/lit-html.ts
justinfagnani: These parts should cover that:
```
 *    * Any character except space, ('), ("), "<", ">", "=", (`), or
 *    * (") then any non-("), or	 *    * (") then any non-("), or
 *    * (') then any non-(')	 *    * (') then any non-(')
```

They're pretty wide-open in terms of what's allowed.

I'll add that test.

packages/lit-html/src/lit-html.ts
justinfagnani: Added comment

packages/lit-html/src/lit-html.ts
```diff
@@ -328,17 +337,25 @@ class Template {
     let node: Node | null;
     let nodeIndex = 0;
     let bindingIndex = 0;
+
+    // If we're in attribute value position, this will be the index of the end
+    // of the attribute name.
```
sorvell, line 342: Comment confuses this with `attrNameEndIndex`

packages/lit-html/src/lit-html.ts
sorvell: This name doesn't seem to fit since this can match without ending the tag, right?

`tagOrAttrEndRegex`?

packages/lit-html/src/lit-html.ts
sorvell: ok, makes sense.

packages/lit-html/src/lit-html.ts
```diff
@@ -350,63 +367,96 @@ class Template {
...
-      if (attrNameEnd !== -1) {
-        attrNames.push(attrName!);
-        html +=
-          s.slice(0, attrNameEnd) + '$lit$' + s.slice(attrNameEnd) + marker;
-      } else {
-        html += regex === textRegex ? s + nodeMarker : s + marker;
+      if (DEV_MODE) {
+        // If we have a attrNameEndIndex, which indicates that we should
+        // rewrite the attribute name, assert that we're in a valid attribute
+        // position - either in a tag, or a quoated attribute value.
```
sorvell, line 437:
```suggestion
        // position - either in a tag, or a quoted attribute value.
```

packages/lit-html/src/lit-html.ts
sorvell: This is the index into the `attrNames` array, right?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
justinfagnani: I guess it depends on the definition of the state. What I mean here is that this regex detects the end of the "tag state", which can be the tag end, or an attribute.

packages/lit-html/src/lit-html.ts
justinfagnani: Oops, I added this to the wrong thing

packages/lit-html/src/lit-html.ts
justinfagnani: fixed

justinfagnani (commented):
PTAL

packages/lit-html/src/lit-html.ts
sorvell: Still think this name is weird, but it's hard to name.

sorvell (approved):

PR #1305, opened 2020-09-22
[lit-next] Set up Sauce with Firefox ESR

aomarks:
- Set up the rigging for Sauce with `@web/test-runner`.

- Adds Firefox 68 (ESR) on Windows to our CI.

- For now, running the sauce browsers in a separate job, because I think it's likely we'll see some failures, and it will be less disruptive if we can quickly see whether its local vs sauce that's failing.

- The other browsers still don't work, since the plugin is very new. Firefox 68 is the only one I've gotten to work from the preliminary set for lit-next. https://github.com/modernweb-dev/web/issues/472 is tracking the problems, so we'll just keep an eye on it and uncomment browsers as they start working.

- Adds a `BROWSER` environment variable, read by our wtr config. Makes it easy to pick the browsers from the commandline. You can specify Sauce ones like `SAUCE_USERNAME=polymer-devs SAUCE_ACCESS_KEY=<redacted> BROWSERS="sauce:Windows 10/firefox@68" npm test`, Playwright ones like `BROWSERS=chromium,firefox npm test`, or the same set we run in CI with `BROWSERS=preset:sauce npm test`.

Fixes https://github.com/Polymer/lit-html/issues/1222

### DIFF
--- a/packages/tests/web-test-runner.config.js
+++ b/packages/tests/web-test-runner.config.js
@@ -1,11 +1,13 @@
 import { playwrightLauncher } from "@web/test-runner-playwright";
 import { fromRollup } from "@web/dev-server-rollup";
+import { createSauceLabsLauncher } from "@web/test-runner-saucelabs";
+import { legacyPlugin } from "@web/dev-server-legacy";
 import { resolveRemap } from "./rollup-resolve-remap.js";
 import { prodResolveRemapConfig, devResolveRemapConfig } from "./wtr-config.js";
 
 // TODO Replace this with log filter feature when/if added to wtr
 // https://github.com/modernweb-dev/web/issues/595
-const removeDevModeLogging = {
+const removeDevModeLoggingPlugin = {
   name: "remove-dev-mode-logging",
   transform(context) {
     if (context.response.is("js")) {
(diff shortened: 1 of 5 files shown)


packages/tests/web-test-runner.config.js
```diff
@@ -16,33 +18,149 @@ const removeDevModeLogging = {
...
+  } else {
+    console.log("Using development builds");
+    resolveRemapConfig = devResolveRemapConfig;
+  }
+  return [
+    fromRollup(resolveRemap)(resolveRemapConfig),
+    removeDevModeLoggingPlugin,
+    // Detect browsers without modules (e.g. IE11) and transform to SystemJS
+    // (https://modern-web.dev/docs/dev-server/plugins/legacy/).
+    legacyPlugin,
```
justinfagnani, line 35: Is this being exercised yet? Firefox ESR has modules. Maybe add it later?

packages/tests/web-test-runner.config.js
```diff
@@ -16,33 +18,149 @@ const removeDevModeLogging = {
...
+    "sauce:Windows 10/firefox@68", // Current ESR
+    //"sauce:Windows 10/chrome@latest-3", // Fails with no error message
+    //"sauce:macOS 10.15/safari@latest", // Timeout on "elements with custom properties can move between elements"
+    //"sauce:Windows 10/MicrosoftEdge@18", // Browser start timeout
+    //"sauce:Windows 7/internet explorer@11", // Browser start timeout
+  ],
+};
+
+function getBrowsers() {
+  return (process.env.BROWSERS || "preset:default")
```
justinfagnani, line 61: nit: use `??` instead of `||`

packages/tests/web-test-runner.config.js
```diff
@@ -16,33 +18,149 @@ const removeDevModeLogging = {
...
+    return [];
+  }
+
+  if (browser.startsWith("preset:")) {
+    const preset = browser.substring("preset:".length);
+    const entries = browserPresets[preset];
+    if (!entries) {
+      throw new Error(
+        `Unknown preset "${preset}", please pick one of: ` +
+          Object.keys(browserPresets).join(", ")
```
justinfagnani, line 110: FYI: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat

Not saying to use it, just popped into my head. It's an option though.

packages/tests/web-test-runner.config.js
```diff
@@ -16,33 +18,149 @@ const removeDevModeLogging = {
...
+    return entries.map(parseBrowser).flat();
+  }
+
+  if (browser.startsWith("sauce:")) {
+    // Note this is the syntax used by WCT. Might as well use the same one.
+    const match = browser.match(/^sauce:(.+)\/(.+)@(.+)$/);
+    if (!match) {
+      throw new Error(
+        `Invalid Sauce browser string, ` +
+          `expected format "sauce:os/browser@version", ` +
```
justinfagnani, line 122: maybe add common examples too, since I always get it wrong and have to dig for whether there are spaces, capital letters, etc. At least `sauce:Windows 7/internet explorer@11`

justinfagnani (approved):

packages/tests/web-test-runner.config.js
aomarks: We know we'll need it, might as well be running it? Would rather notice problems it's introducing sooner, and it shouldn't add any latency. It should only take effect automatically on browsers that don't support modules.

packages/tests/web-test-runner.config.js
aomarks: I want `||` here because I want empty string to count as default.

packages/tests/web-test-runner.config.js
aomarks: Huh, didn't know about that. But seems kinda overkill.

packages/tests/web-test-runner.config.js
aomarks: Done.

aomarks:
Since I saw some occasional failures even with Firefox 68, I moved the Sauce browsers to a separate job for now, so that it will be easy to see when it's only Sauce that is failing. Plus we'll see the local tests really fast even while the Sauce ones are still running.

PR #1307, opened 2020-09-22
[lit-next] Document manual testing

aomarks:
Also, makes it so the Sauce job won't fail a commit (since it's still flaky), and disables the Sauce job on forked PRs (since they can't access secrets).

Fixes https://github.com/Polymer/lit-html/issues/1306


justinfagnani (approved):
Thanks!

PR #1311, opened 2020-09-22
[lit-next] Unify the both check-version-tracker scripts

abdonrd:
Now with modules as requested here: https://github.com/Polymer/lit-html/pull/1298#discussion_r493071519


aomarks (approved):

aomarks:
Thanks!

PR #1314, opened 2020-09-23
[lit-next] Inject version number at build time

abdonrd:
I use the `__DEV_VERSION_NUMBER__` string. Let me know if you prefer another.

### DIFF
--- a/packages/lit-element/rollup.config.js
+++ b/packages/lit-element/rollup.config.js
@@ -19,6 +19,8 @@ import * as pathLib from 'path';
 import sourcemaps from 'rollup-plugin-sourcemaps';
 import replace from '@rollup/plugin-replace';
 
+import * as packageJson from './package.json';
+
 // In CHECKSIZE mode we:
 // 1) Don't emit any files.
 // 2) Don't include copyright header comments.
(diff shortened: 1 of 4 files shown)


aomarks (commented):
Nice work!

However, I think we probably want the development build to have a version number too, since it could be useful to be able to check the version at runtime from a development server. Maybe with a `-dev` suffix or similar? As currently written, it would be literally `"__DEV_VERSION_NUMBER__"`.

@justinfagnani WDYT?

That will be tricky to automate, because the `development/` directory is currently the pure output of `tsc`, and that's a nice attribute to preserve if possible. A simple in-place search/replace after tsc runs would break sourcemaps.

So unfortunately maybe just continuing to enforce that the TypeScript source contains the right version number before publishing, plus maybe stripping out the `-dev` suffix for the minified output if we want that, is the simplest solution for now.

At some point maybe we'll automate the publishing step, which could include updating the TypeScript source automatically.

justinfagnani:
I think @aomarks is right here, for the moment. We want `/development/` to contain the right version for when users are in development mode.

I hadn't yet thought about wanting that version to have a `-dev` suffix, which is a great idea.

The biggest problem we had recently was that the version check wasn't running on prepublishOnly in the lit-next branch. Now that that's fixed, the automation, while it would be nice, doesn't really buy all that much.

abdonrd:
Got it! Closing then!

PR #1315, opened 2020-09-23
[lit-next] shady render

sorvell:
Adds `platform-support` module. This module has a couple of breaking changes:

1. Bindings in style elements are no longer supported. Previously these could not change and in the future they may be supported via static bindings.
1. `ShadyCSS.styleElement` is no longer called automatically (in lit-html, it still is in LitElement). This must be called whenever dynamic changes that affect styling are made that involve css custom property shimming (older browsers) or changes to custom properties used via the deprecated `@apply` feature. It was previously called only on first render, and it is now up to the user to decide when this should be called.

**Review Notes**
* The most important main code changes are in [lit-html](https://github.com/Polymer/lit-html/pull/1315/files#diff-43727d8a017c2bf3455b019c3966e0c8R299) and the added [platform-support](https://github.com/Polymer/lit-html/pull/1315/files#diff-f8f8f94b7b5f632bfb5a2fcdc2eaf489R1) module.
* The other changes are mostly related to testing setup. No new tests were added, but all of the LitElement and lit-html tests for shady-render were moved into the `lit-element/test/platform-support` folder.


packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+          HTMLStyleElement
+        >;
+        scopeCss.push(
+          ...Array.from(styles).map((style) => {
+            style.parentNode?.removeChild(style);
+            return style.textContent!;
+          })
+        );
+        // Apply styles if this is the result rendering into shadowRoot.
+        if (currentResult.__renderIntoShadowRoot) {
```
kevinpschaaf, line 297: If we're not rendering into shadowRoot, then what? Shouldn't we not touch anything if so?

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+    }
+    let scopeCss = scopeCssStore.get(name);
+    if (scopeCss === undefined) {
+      scopeCssStore.set(name, (scopeCss = []));
+    }
+    return scopeCss;
+  };
+
+  const ensureStyles = (name: string, template: HTMLTemplateElement) => {
+    if (isScopeStyled(name)) {
```
kevinpschaaf, line 249: The logic/relationship between `ensureStyles` and `styledScopes` and `isScopeStyled` is pretty circular and hard to follow. I might need an explanation.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+      } else {
+        const previousScope = currentScope;
+        currentScope = this.options.scope ?? '';
+
+        const renderingTemplateResult = !!(value as ShadyTemplateResult)
+          ?._$litType$;
+        // If result is a TemplateResult decorate it so that when the template
+        // is created, styling can be shimmed.
+        if (renderingTemplateResult) {
+          (value as ShadyTemplateResult).__renderIntoShadowRoot = renderingIntoShadowRoot;
```
kevinpschaaf, line 331: If we're going to set this (always true), can be the scope instead and avoid one of the side-channel variables (`currentScope`)?

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+          // Temporarily swizzle tracked nodes so the `__insert` is correct.
+          this._startNode = this._endNode = renderContainer.firstChild!;
+        }
+        this.__baseSetValue(value);
+
+        // Ensure styles are scoped if a TemplateResult is *not* being rendered.
+        // Note, if a TemplateResult *is* being rendered, scoping is done
+        // in `Template._createElement`, before Template parts are created.
+        if (renderingIntoShadowRoot && !renderingTemplateResult) {
+          ensureStyles(currentScope, document.createElement('template'));
```
kevinpschaaf, line 354: This is so `prepareAdoptedCssText` styles get applied? Comment?

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+          ?._$litType$;
+        // If result is a TemplateResult decorate it so that when the template
+        // is created, styling can be shimmed.
+        if (renderingTemplateResult) {
+          (value as ShadyTemplateResult).__renderIntoShadowRoot = renderingIntoShadowRoot;
+        }
+
+        // Note, in first scope render, render into a fragment and move to container
+        // to preserve outer => inner ordering important for @apply production
+        // before consumption.
```
kevinpschaaf, line 336: I'm not grokking exactly how this is different from what the normal flow does. Maybe expand on that?

packages/lit-html/src/lit-html.ts
```diff
@@ -505,14 +521,16 @@ class Template {
...
-              (node as Element).removeAttribute(name);
-              i--;
+              attrsToRemove.push(name);
               this._parts.push({
                 _type: ELEMENT_PART,
                 _index: nodeIndex,
               });
             }
           }
+          for (const name of attrsToRemove) {
```
kevinpschaaf, line 531: Is mutating attributes while iterating them only bad on IE? Add comment to that effect?

packages/lit-html/src/lit-html.ts
```diff
@@ -558,6 +576,12 @@ class Template {
       nodeIndex++;
     }
   }
+
+  _createElement(html: string) {
```
kevinpschaaf, line 580: Note this is an override point for ShadyDOM?

packages/lit-html/src/lit-html.ts
```diff
@@ -769,6 +790,14 @@ export class NodePart {
     }
   }
 
+  private _getTemplate(strings: TemplateStringsArray, result: TemplateResult) {
```
kevinpschaaf, line 793: Note this is an override point for ShadyDOM?

packages/lit-element/src/lib/platform-support.ts
kevinpschaaf: I see, we _accumulate_ if we have a scope but are not the root template going into the SR.

packages/lit-element/src/lib/platform-support.ts
kevinpschaaf: I misunderstood, this is marking the _top_ template, but nested templates still need the `currentScope`.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+ */
+
+/**
+ * LitElement/lit-html patch to support browsers without native web components.
+ *
+ * @packageDocumentation
+ */
+
+// TODO(sorvell): Remove these once ShadyDOM/webcomponentsjs supports them.
+// Source: https://github.com/jserz/js_piece/blob/master/DOM/ParentNode/append()/append().md
```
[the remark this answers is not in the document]
sorvell: Appears to have been deleted, updated to reference MDN.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+  // console.log(
+  //   '%c Making LitElement compatible with ShadyDOM/CSS.',
+  //   'color: lightgreen; font-style: italic'
+  // );
+
+  /**
+   * Patch `render` to include scope.
+   */
+  Object.assign(LitElement, {
+    __render: ((LitElement as unknown) as LitElementConstructorStandIn).render,
```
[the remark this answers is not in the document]
sorvell: So that they are called in the right context without having to use `.call`

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  // console.log(
+  //   '%c Making LitElement compatible with ShadyDOM/CSS.',
+  //   'color: lightgreen; font-style: italic'
+  // );
+
+  /**
+   * Patch `render` to include scope.
+   */
+  Object.assign(LitElement, {
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+  /**
+   * Patch to apply adoptedStyleSheets via ShadyCSS
+   */
+  Object.assign(LitElement.prototype, {
+    __baseAdoptStyles: LitElement.prototype.adoptStyles,
+    adoptStyles(this: LitElementStandIn, styles: CSSResults) {
+      if (!this.constructor.hasOwnProperty(SCOPE_KEY)) {
+        const name = (this.constructor[SCOPE_KEY] = this.localName);
+        if (window.ShadyCSS!.nativeShadow) {
+          this.__baseAdoptStyles(styles);
```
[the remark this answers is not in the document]
sorvell: Added.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+);
+
+interface RenderOptions {
+  readonly renderBefore?: ChildNode | null;
+  scope?: string;
+}
+
+const SCOPE_KEY = '__localName';
+
+interface LitElementConstructorStandIn {
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
sorvell: Added.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Patch connectedCallback to apply ShadyCSS custom properties shimming.
+     */
+    __baseConnectedCallback: LitElement.prototype.connectedCallback,
+    connectedCallback(this: LitElementStandIn) {
+      this.__baseConnectedCallback();
+      // Note, must do first update separately so that we're ensured
+      // that rendering has completed before calling this.
+      if (this.hasUpdated) {
+        window.ShadyCSS!.styleElement(this);
```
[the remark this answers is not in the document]
sorvell: It's called by LitElement, not by lit-html. That's what the PR is noting... will clarify.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+    },
+  });
+
+  /**
+   * Patch to apply adoptedStyleSheets via ShadyCSS
+   */
+  Object.assign(LitElement.prototype, {
+    __baseAdoptStyles: LitElement.prototype.adoptStyles,
+    adoptStyles(this: LitElementStandIn, styles: CSSResults) {
+      if (!this.constructor.hasOwnProperty(SCOPE_KEY)) {
```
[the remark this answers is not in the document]
sorvell: Added comment. We'd still need to override `adoptStyles` to ensure it doesn't do the wrong thing under ShadyCSS. Given that, the other changes don't seem worth it.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+    return;
+  }
+
+  // console.log(
+  //   '%c Making lit-html compatible with ShadyDOM/CSS.',
+  //   'color: lightgreen; font-style: italic'
+  // );
+
+  const styledScopes: Set<string> = new Set();
+  const scopeCssStore: Map<string, string[]> = new Map();
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+  let currentScope: string;
+
+  const isScopeStyled = (name: string | undefined) =>
+    needsPlatformSupport && name !== undefined ? styledScopes.has(name) : true;
```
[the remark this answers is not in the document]
sorvell: Removed.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+  };
+
+  const ensureStyles = (name: string, template: HTMLTemplateElement) => {
+    if (isScopeStyled(name)) {
+      return;
+    }
+    // Mark this scope as styled.
+    styledScopes.add(name);
+    // Remove stored data
+    scopeCssStore.delete(name);
```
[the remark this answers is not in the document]
sorvell: There is none. This is here just so we don't hold on to the css data.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,388 @@
...
+    Map<TemplateStringsArray, TemplateStandIn>
+  >();
+
+  // Note, it's ok to subclass Template since it's only used via NodePart.
+  let currentResult: ShadyTemplateResult;
+  class ShadyTemplate extends Template {
+    constructor(result: ShadyTemplateResult) {
+      // TODO(sorvell): Decide if Template should store the result to avoid this.
+      // We need the `result` in `_createElement` which is called directly
+      // from the constructor. Since we can't access this before super, uce
```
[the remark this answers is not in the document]
sorvell: Fixed.

packages/lit-element/src/lib/platform-support.ts
sorvell: Changed: `ensureStyles` to `prepareStyles` and added comments.

packages/lit-element/src/lib/platform-support.ts
sorvell: Refined comment.

packages/lit-html/src/lit-html.ts
sorvell: Added.

packages/lit-html/src/lit-html.ts
sorvell: Added.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -302,87 +318,105 @@ interface TemplateStandIn {
...
-        scopedTemplateCache.set(currentScope, (templateCache = new Map()));
+      // Note, @apply requires outer => inner scope rendering on initial
+      // scope renders to apply property values correctly. Style preparation
+      // is tied to rendering into `shadowRoot`'s and this is typically done by
+      // custom elements. If this is done in `connectedCallback`, as is typical,
+      // the code below ensures the right order since content is rendered
+      // into a fragment first so the hosting element can prepare styles first.
+      // If rendering is done in the constructor, this won't work, but that's
+      // not supported in ShadyDOM anyway.
+      let renderContainer: DocumentFragment | undefined = undefined;
```
kevinpschaaf, line 363: Ok, so the _key_ thing is "prevent custom elements in the templates I've just rendered from upgrading until I have had a chance to prepare my styles, by not allowing them to connect yet", right?

If so, why doesn't the fact that any elements that might be connected sync to rendering wait a microtask to do _their_ rendering accomplish the same thing?

kevinpschaaf: Ah, it's compatibility with legacy Polymer elements that render sync to connected, is that it?

packages/lit-html/src/lit-html.ts
kevinpschaaf: Justin covered in his PR

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -367,48 +348,52 @@ let currentScope: string;
       if (needsStyleScoping && renderingIntoShadowRoot) {
```
kevinpschaaf, line 348: `renderingIntoShadowRoot` is `true` here due to test on L330 right?

kevinpschaaf (changes requested):

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -136,63 +137,77 @@ interface LitElementStandIn extends HTMLElement {
...
-        window.ShadyCSS!.styleElement(this);
+  LitElement.prototype.__baseAdoptStyles = LitElement.prototype.adoptStyles;
+  LitElement.prototype.adoptStyles = function (
+    this: PatchableLitElement,
+    styles: CSSResults
+  ) {
+    // Note, `SCOPE_KEY` is used both to determine if the element has already
+    // been styled by ShadyCSS and also to pass the scope name to the static
+    // side of the class.
+    if (!this.constructor.hasOwnProperty(SCOPE_KEY)) {
```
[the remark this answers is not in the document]
sorvell: Good catch, will add a test for this.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -219,6 +219,7 @@ interface PatchableNodePart {
 interface PatchableTemplate {
   new (...args: any[]): PatchableTemplate;
   _createElement(html: string): HTMLTemplateElement;
+  _element: HTMLTemplateElement;
```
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,413 @@
...
+      }
+      // Note, any nested template results render here and their styles will
+      // be extracted and collected.
+      this.__baseSetValue(value);
+
+      let template: HTMLTemplateElement | undefined = undefined;
+      if (needsStyleScoping && renderingIntoShadowRoot) {
+        // Get the template for this result or create a dummy one if a result
+        // is not being rendered.
+        template = !!(value as ShadyTemplateResult)?._$litType$
```
kevinpschaaf, line 363: To get the template, you could also do
```js
 template = this._value._template?._element || document.createElement('template')
```
after `__baseSetValue` since `__commitTemplateResult` sets the `part._value` to a `TemplateInstance` instance

packages/lit-element/src/lib/platform-support.ts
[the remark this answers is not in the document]
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
sorvell: Fixed.

packages/lit-element/src/lib/platform-support.ts
sorvell: Done.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,380 @@
...
+      // When using native Shadow DOM, include prepared style in shadowRoot.
+      if (window.ShadyCSS?.nativeShadow) {
+        const style = template.content.querySelector('style');
+        if (style !== null) {
+          renderContainer.appendChild(style.cloneNode(true));
+        }
+      }
+      container.insertBefore(
+        renderContainer,
+        this.options?.renderBefore || null
```
kevinpschaaf, line 348:
```suggestion
        endNode
```

packages/lit-element/src/lit-element.ts
```diff
@@ -208,6 +186,9 @@ export class LitElement extends UpdatingElement {
    */
   protected initialize() {
     super.initialize();
+    (this as {
+      _renderOptions: RenderOptions;
+    })._renderOptions = {eventContext: this};
```
kevinpschaaf, line 191: Why isn't this just a field with an initializer?
```ts
private _renderOptions = {eventContext: this};
```

kevinpschaaf: can't because adopt styles needs it

kevinpschaaf (approved):
LGTM

packages/lit-element/CHANGELOG.md
```diff
@@ -20,224 +20,285 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
-* [Breaking] Removed build support for TypeScript 3.4.
-* [Breaking] Decorators are no longer exported from the `lit-element` module. Instead, import any decorators you use from `lit-element/decorators/*`.
-* [Breaking] `lit-html` has been updated to 2.x. Note, shady-render support has been removed. Import the `lit-element-polyfill` module to support Shady DOM.
-* [Breaking] For simplicity, `requestUpdate` no longer returns a Promise. Instead await the `updateComplete` Promise.
-* [Breaking] The type of the `css` function has been changed to `CSSResultGroup` and is now the same as `LitElement.styles`. This avoids the need to cast the `styles` property to `any` when a subclass sets `styles` to an Array and its super class set a single value (or visa versa).
-* For efficiency, the `css` function now maintains a cache and will use a cached value if available when the same style text is requested.
+
+- [Breaking] For simplicity, static `render` has been removed.
```
justinfagnani, line 24: btw, I don't think we need the reasons for things like this

packages/lit-element/CHANGELOG.md
```diff
@@ -20,224 +20,285 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
-* [Breaking] Removed build support for TypeScript 3.4.
-* [Breaking] Decorators are no longer exported from the `lit-element` module. Instead, import any decorators you use from `lit-element/decorators/*`.
-* [Breaking] `lit-html` has been updated to 2.x. Note, shady-render support has been removed. Import the `lit-element-polyfill` module to support Shady DOM.
-* [Breaking] For simplicity, `requestUpdate` no longer returns a Promise. Instead await the `updateComplete` Promise.
-* [Breaking] The type of the `css` function has been changed to `CSSResultGroup` and is now the same as `LitElement.styles`. This avoids the need to cast the `styles` property to `any` when a subclass sets `styles` to an Array and its super class set a single value (or visa versa).
-* For efficiency, the `css` function now maintains a cache and will use a cached value if available when the same style text is requested.
+
```
justinfagnani, line 23: Is there any change to mention here around support for `<style>` in render() or `@apply`?

packages/lit-element/src/env.d.ts
```diff
@@ -1,12 +1,20 @@
 interface ShadyCSS {
+  nativeCss: boolean;
```
justinfagnani, line 2: Did @aomarks add an interface for this to ShadyCSS?

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,377 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+/**
+ * LitElement/lit-html patch to support browsers without native web components.
+ *
+ * @packageDocumentation
```
justinfagnani, line 17: I _think_ this should be `@file`. Not 100% sure though.

packages/lit-element/src/lib/platform-support.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  // console.log(
+  //   '%c Making LitElement compatible with ShadyDOM/CSS.',
+  //   'color: lightgreen; font-style: italic'
+  // );
+
+  /**
+   * Patch to apply adoptedStyleSheets via ShadyCSS
+   */
+  LitElement.prototype.__baseAdoptStyles = LitElement.prototype.adoptStyles;
```
justinfagnani, line 113: bit: I would look up the prototype once

packages/lit-html/src/lit-html.ts
```diff
@@ -470,7 +474,8 @@ class Template {
 
     // Note, we don't add '</svg>' for SVG result types because the parser
     // will close the <svg> tag for us.
-    this._element.innerHTML = html + this._strings[l];
+    this._element = this._createElement(html + this._strings[l]);
```
justinfagnani, line 477: I'm interested in exploring removing classes for Template and TemplateInstance, since they're never really polymorphic. Even here, this is a global patch, so it's not polymorphic. I guess we can have a settable module variable, and we could possibly have a build config to replace it with a const if we want to make builds w/o IE support.

packages/lit-html/src/lit-html.ts
```diff
@@ -1095,6 +1113,11 @@ export class EventPart extends AttributePart {
   }
 }
 
+// Apply polyfills if available
+if ((globalThis as any)['litHtmlPlatformSupport'] !== undefined) {
```
justinfagnani, line 1117: Consider writing this as:

```ts
(globalThis as any)['litHtmlPlatformSupport']?.({NodePart, Template});
```

and it'll be compiled out depending on our tsconfig

packages/lit-html/src/test/test-utils/security.ts
```diff
@@ -0,0 +1,40 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+export function trustedTypesIsEnforced(): boolean {
```
justinfagnani, line 15: Do we need this now?

justinfagnani (approved):
Did a quick review. I really like the patching approach!

packages/lit-element/CHANGELOG.md
sorvell: Sure, removed

packages/lit-element/CHANGELOG.md
sorvell: Added.

packages/lit-element/src/lib/platform-support.ts
sorvell: Let's investigate separately and change in all locations if so.

packages/lit-element/src/lib/platform-support.ts
sorvell: Done.

packages/lit-html/src/lit-html.ts
sorvell: Yup, seems possible.

packages/lit-html/src/lit-html.ts
sorvell: Done.

packages/lit-html/src/test/test-utils/security.ts
sorvell: Removed module since it wasn't fully implemented and is being done separately.

PR #1316, opened 2020-09-24
[lit-html] IE11 support

justinfagnani:
Adds IE11 support to lit-html.

Currently, WTR is doing a good job of transpiling our JS for IE and injecting needed JS polyfills. ChildNode#remove and ParentNode#append polyfills are added to a tests-only polyfills.js file, in anticipation of them being added to the webcomponents polyfills.

With that and a couple of changes to tests with `<script>` tags (add a type so they're not parsed as JS) and a change to attribute handling, we have 100% passing on lit-html_test.js.


.gitignore
```diff
@@ -2,8 +2,11 @@
 
 /packages/*/node_modules/
 /packages/*/lib/
-/packages/*/test/
-/packages/*/development/
+/packages/*/test
+/packages/*/development/**/*.js
```
aomarks, line 6: FYI you can also do this (but I don't have a preference):

```
/packages/*/development/
!/packages/*/development/**/*.html
```

packages/lit-html/package.json
```diff
@@ -33,7 +33,7 @@
     "build:ts:watch": "tsc --watch",
     "dev": "scripts/dev.sh",
     "test": "npm run test:dev && npm run test:prod",
-    "test:dev": "cd ../tests && npx wtr '../lit-html/development/**/*_test.js'",
+    "test:dev": "cd ../tests && npx wtr '../lit-html/development/test/**/lit-html_test.js'",
```
aomarks, line 36: Did you mean to include this change?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -33,6 +33,47 @@ import {
 const ua = window.navigator.userAgent;
 const isIe = ua.indexOf('Trident/') > 0;
 
+(function (arr) {
```
aomarks, line 36: Put these polyfills in their own file?

aomarks (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -505,14 +518,16 @@ class Template {
...
-              (node as Element).removeAttribute(name);
-              i--;
+              attrsToRemove.push(name);
               this._parts.push({
                 _type: ELEMENT_PART,
                 _index: nodeIndex,
               });
             }
           }
+          for (const name of attrsToRemove) {
```
kevinpschaaf, line 528: Is mutating attributes while iterating them only bad on IE? Add comment to that effect?

packages/lit-html/src/lit-html.ts
```diff
@@ -744,7 +759,7 @@ export class NodePart {
       // set its value, rather than replacing it.
       (node as Text).data = value as string;
     } else {
-      this._commitNode(new Text(value as string));
+      this._commitNode(document.createTextNode(value as string));
```
kevinpschaaf, line 762: Is it worth including this as a dom shim?

packages/lit-html/src/test/polyfills.ts
```diff
@@ -0,0 +1,41 @@
...
+      enumerable: true,
+      writable: true,
+      value: function remove() {
+        this.parentNode.removeChild(this);
+      }
+    });
+  });
+})([Element.prototype, CharacterData.prototype, DocumentType.prototype]);
+
+// Source: https://github.com/jserz/js_piece/blob/master/DOM/ParentNode/append()/append().md
```
kevinpschaaf, line 18: Link 404'ed; just wanted to double-check license.

kevinpschaaf (changes requested):

packages/lit-html/src/lit-html.ts
justinfagnani: done

packages/lit-html/src/lit-html.ts
justinfagnani: I don't think so... it's such a small change in only one place.

packages/lit-html/src/test/polyfills.ts
justinfagnani: That link was copied with the code from MDN. I'll link to MDN instead.

packages/lit-html/src/test/lit-html_test.ts
justinfagnani: done

kevinpschaaf (approved):
LGTM

PR #1321, opened 2020-09-25
[lit-next] Mangle more lit-html properties by fixing Terser name mangling problem

aomarks:
The Terser shared name cache allows us to mangle the names of properties consistently across modules, so that e.g. `parts.js` can safely access internal details of `lit-html.js`.

However, we still have to account for the problem of mangled names getting _re-used for different properties across files_, because Terser does not consult the `nameCache` to decide whether a mangled name is available or not.

For example:

```
file1:
  obj.foo -> A
  obj.bar -> B

file2:
  obj.bar -> B (Correctly chosen from nameCache.)
  obj.baz -> A (Oops, foo and baz are different properties on the same
                object, but now they both have the same mangled name,
                which could result in very unpredictable behavior).
```

To trick Terser into doing what we need here, we first create a giant bundle of all our code in a single file, tell Terser to minify that, and then throw it away. This seeds the name cache in a way that guarantees every property
gets a unique mangled name.

Before:
image

After:
image

### Still to do
- We can't yet mangle `_value`, because a test file accesses it, so that will fail when testing in minified mode. One solution is to generate a proxy file just for our tests that has something like `export const partValue(part) => part.__value;`.
- Haven't updated `lit-element` yet. I'm planning on unifying the two rollup configs, since they are getting kind of large, and are identical except for the list of entrypoints.

https://github.com/Polymer/lit-html/issues/1272

### DIFF
--- a/packages/lit-html/rollup.config.js
+++ b/packages/lit-html/rollup.config.js
@@ -18,6 +18,7 @@ import copy from 'rollup-plugin-copy';
 import * as pathLib from 'path';
 import sourcemaps from 'rollup-plugin-sourcemaps';
 import replace from '@rollup/plugin-replace';
+import virtual from '@rollup/plugin-virtual';
 
 // In CHECKSIZE mode we:
 // 1) Don't emit any files.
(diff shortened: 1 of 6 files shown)


packages/lit-html/rollup.config.js
```diff
@@ -53,82 +54,129 @@ const entryPoints = [
...
+//                 which could result in very unpredictable behavior).
+//
+// To trick Terser into doing what we need here, we first create a giant bundle
+// of all our code in a single file, tell Terser to minify that, and then throw
+// it away. This seeds the name cache in a way that guarantees every property
+// gets a unique mangled name.
 const nameCache = {};
+const nameCacheSeederInfile = 'name-cache-seeder-virtual-input.js';
+const nameCacheSeederOutfile = 'name-cache-seeder-throwaway-output.js';
+const nameCacheSeederContents = entryPoints
```
justinfagnani, line 84: The original idea was that we would make our directives use the public API and only share names between lit-html.ts and parts.ts. Accessing private APIs from directives probably means that our public API is deficient.

We can do that here by just including those two files, right?

packages/lit-html/src/lit-html.ts
```diff
@@ -671,8 +671,8 @@ export class NodePart {
   protected __directive?: Directive;
 
   constructor(
-    public _startNode: ChildNode,
-    public _endNode: ChildNode | null,
+    public __startNode: ChildNode,
```
justinfagnani, line 674: It would be great to make these private again, but define another interface that makes them public and cast to that where needed. I had some difficulty with getting TS to allow this last I tried.

justinfagnani (changes requested):

packages/lit-html/rollup.config.js
aomarks: We could do that (though not just by changing this array -- we'd need to make separate batches of rollup runs, because this array is only related to the "don't mangle two different properties in two different files to the same name" concern, not the "do mangle the same property in two different files to the same name" concern).

Currently, `directives/live.ts` and `directives/repeat.ts` both directly write to `NodePart._value` to bypass dirty checking, so before we did this, we would need to resolve what that public interface should be. A `setPartValueDirectly` or something function? Looks like https://github.com/Polymer/lit-html/issues/1261 is the place for this discussion.

packages/lit-html/src/lit-html.ts
aomarks: Done.

packages/lit-html/src/parts.ts
```diff
@@ -13,15 +13,23 @@ type NodePartStateInternal = {
   __fragment: DocumentFragment;
 };
 
+/**
+ * Package private members of NodePart.
+ */
+type NodePartInternal = {
+  __startNode: NodePart['__startNode'];
```
justinfagnani, line 20: nice!

justinfagnani (approved):

PR #1322, opened 2020-09-25
Make tests pass in saucelabs

LarsDenBakker:
I'm investigating different issues in the saucelabs launcher with different projects, and making improvements. Making a PR of the changes I made so far, feel free to merge or close it.

The test failures on chrome and safari are all related to styling, so it likely has to do with the way we run tests concurrently. Setting a concurrency of 1 makes them pass at least, though it's slower. I'm looking into what we can do here to either fix the issue or make it faster.

This also updates the test runner dependencies to a version which handles logging in tests run with selenium, so that errors in saucelabs are easier to debug. For IE11 and Edge there are actionable errors.

### DIFF
--- a/packages/tests/package-lock.json
+++ b/packages/tests/package-lock.json
@@ -1,6158 +1,6080 @@
 {
-	"name": "tests",
-	"requires": true,
-	"lockfileVersion": 1,
-	"dependencies": {
-		"@babel/code-frame": {
-			"version": "7.10.4",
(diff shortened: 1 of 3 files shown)


aomarks (approved):
Confirmed these are passing in https://github.com/Polymer/lit-html/runs/1168402634 -- awesome!

Thank you @LarsDenBakker! Much appreciated.

PR #1326, opened 2020-09-29
[lit-next] Add NPM exports configs for lit-html and lit-element

aomarks:
Adds `exports` configurations to the lit-html and lit-element `package.json` files, to enable users to switch between production and development mode using a Node export condition.

## Background

lit-html and lit-element next will be distributed with two builds:

1. A default minified build suitable for serving to users.
2. An optional development build, which will log additional correctness, debugging, performance etc. information to the console, suitable only during development.

The development builds are configured using a Node export condition (see https://nodejs.org/api/esm.html#esm_conditional_exports) named `"development"`.

<strike>Note: I picked `"lit-development"` over `"development"` so that it's less likely to collide with other packages that might also have a similar export condition, because export conditions are global.</strike>

A number of tools are on the verge of supporting export conditions:

### @web/dev-server

Confirmed working with local copy of https://github.com/rollup/plugins/pull/540. Will work once it gets merged and published.

```js
// web-dev-server.config.mjs

export default {
  nodeResolve: {
    exportConditions: [ "development" ]
  }
}
```

### es-dev-server

Uses an older version of `@rollup/node-plugin-resolve`, which is not compatible with https://github.com/rollup/plugins/pull/540, so an update to es-dev-server would be needed (probably really small). Maybe we just tell people to upgrade to `@web/dev-server` (which replaces `es-dev-server`).

```js
// es-dev-server.config.js

exports.default = {
  nodeResolve: {
    exportConditions: [ "development" ]
  }
}
```

### Rollup

Confirmed working with local copy of https://github.com/rollup/plugins/pull/540. Will work once it gets merged and published.

```js
// rollup.config.js

export default {
  plugins: [
    nodeResolve({
      exportConditions: [ "development" ]
    })
  ]
};
```

### Webpack

Confirmed working with webpack 5 (currently in beta). See https://gist.github.com/sokra/e032a0f17c1721c71cfced6f14516c62 for docs.

```bash
npm install --save-dev webpack@next
```

```js
// webpack.config.js

module.exports = {
  entry: './index.js',
  mode: 'development',
  resolve: {
    conditionNames: [ "development" ]
  }
};
```

cc @LarsDenBakker Awesome work on https://github.com/rollup/plugins/pull/540!

Fixes https://github.com/Polymer/lit-html/issues/1255

### DIFF
--- a/packages/lit-element/package.json
+++ b/packages/lit-element/package.json
@@ -9,6 +9,16 @@
   "main": "lit-element.js",
   "module": "lit-element.js",
   "typings": "lit-element.d.ts",
+  "exports": {
+    ".": {
+      "lit-development": "./development/lit-element.js",
+      "default": "./lit-element.js"
+    },
+    "./lib/": {
+      "lit-development": "./development/lib/",
+      "default": "./lib/"
+    }
+  },
   "directories": {
     "test": "test"
   },
--- a/packages/lit-element/package.json
+++ b/packages/lit-element/package.json
@@ -11,11 +11,11 @@
   "typings": "lit-element.d.ts",
   "exports": {
     ".": {
-      "lit-development": "./development/lit-element.js",
+      "development": "./development/lit-element.js",
       "default": "./lit-element.js"
     },
     "./lib/": {
-      "lit-development": "./development/lib/",
+      "development": "./development/lib/",
       "default": "./lib/"
     }
   },
--- a/packages/lit-element/CHANGELOG.md
+++ b/packages/lit-element/CHANGELOG.md
@@ -32,6 +32,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Breaking] The type of the `css` function has been changed to `CSSResultGroup` and is now the same as `LitElement.styles`. This avoids the need to cast the `styles` property to `any` when a subclass sets `styles` to an Array and its super class set a single value (or visa versa).
 * For efficiency, the `css` function now maintains a cache and will use a cached value if available when the same style text is requested.
 
+### Added
+* Adds development mode, which can be enabled by setting the `development` Node exports condition. See `README.md` for more details.
+
 ### Removed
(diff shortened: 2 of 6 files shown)


aomarks:
Once you have `exports` defined, paths that are not in `exports` cannot be imported. So as currently written in this PR, it's no longer possible to `import 'lit-html/lit-html.js'`, you can _only_ `import 'lit-html'`. This might be nice to encourage more consistency, but we may want to add a `lit-html.js` export for backwards compatibility. Same goes for `lit-element.js`.

@justinfagnani thoughts?

packages/lit-element/package.json
```diff
@@ -9,6 +9,16 @@
   "main": "lit-element.js",
   "module": "lit-element.js",
   "typings": "lit-element.d.ts",
+  "exports": {
+    ".": {
+      "lit-development": "./development/lit-element.js",
```
justinfagnani, line 14: I think we might want to have this condition be just `development` so that the ecosystem can converge on that and dev servers, build tools, and such can have an easy to use `--dev` flag.

Node docs give an example of the `development` condition: https://nodejs.org/api/esm.html#esm_conditional_exports

@LarsDenBakker I'm curious what you think there?

justinfagnani (commented):
Awesome!!

Aside from the question on the name of the condition, the only other thing I would ask to include is a small section in the readmes and changelog entries.

justinfagnani:
cc @arthurevans dev mode and the flags to trigger it in various tools are going to be new things to doc. We might want to mention dev mode early on in the docs, maybe even in getting started, as a nice DX improvement (super-minified prod mode will be the default) with details on enabling it in tools?

packages/lit-element/package.json
aomarks: Switched to `development`

aomarks:
> Awesome!!
> 
> Aside from the question on the name of the condition, the only other thing I would ask to include is a small section in the readmes and changelog entries.

Updated both changelogs and readmes, PTAL

justinfagnani (approved):

PR #1327, opened 2020-09-29
[lit-next] Upgrade to latest wtr, use new built-in log filtering, other small tweaks

aomarks:


justinfagnani (approved):

sorvell (approved):

packages/tests/web-test-runner.config.js
```diff
@@ -160,17 +153,28 @@ export default {
     "../lit-element/development/**/*_test.js",
   ],
   nodeResolve: true,
-  concurrency: !process.env.BROWSERS || process.env.BROWSERS === "preset:local" ? 10 : 1,
+  concurrency: sauceLauncher ? 1 : 10,
```
LarsDenBakker, line 156: I noticed the failure on this PR so I was looking if it was caused by the latest WTR - but saw that `sauceLauncher` is not defined until `getBrowsers` called setting the concurrency to 10.

aomarks: 🤦 good catch thanks!

PR #1334, opened 2020-09-29
[lit-next] Minify private lit-element properties, and rename lit-html private props to one underscore

aomarks:
- Mangle all lit-element properties starting with `_`
- Change private lit-html property prefixes from `__` to `_` for consistency with lit-element
  - Includes exclusions for `_value`, `_setValue`, `_$litType$`, and `_$litDirective$`
- Pre-seed the terser name cache with random names for `_createElement`, `_endNode`, and `_startNode` since those will be needed by Steve in #1315. We can use this pattern to ensure that any private properties we need to access *across packages* will be consistent, including across versions.

Also some small typing improvements I noticed:
- Add a cast to `UpdatingElement` prototype in lit-element
- Replace an `any` cast in lit-html with a `NodePartInternal` cast

Before:
image

After:
image

Fixes https://github.com/Polymer/lit-html/issues/1330

### DIFF
--- a/packages/lit-element/src/lib/updating-element.ts
+++ b/packages/lit-element/src/lib/updating-element.ts
@@ -30,7 +30,7 @@ window.JSCompiler_renameProperty = <P extends PropertyKey>(
 ): P => prop;
 
 declare global {
-  var JSCompiler_renameProperty: <P extends PropertyKey>(
+  const JSCompiler_renameProperty: <P extends PropertyKey>(
     prop: P,
     _obj: unknown
   ) => P;
--- a/packages/lit-element/src/lib/updating-element.ts
+++ b/packages/lit-element/src/lib/updating-element.ts
@@ -396,9 +396,9 @@ export abstract class UpdatingElement extends HTMLElement {
     }
     this[finalized] = true;
     // finalize any superclasses
-    const superProto = Object.getPrototypeOf(this);
+    const superProto = Object.getPrototypeOf(this) as typeof UpdatingElement;
     superProto.finalize();
-    this._classProperties = new Map(superProto._classProperties);
+    this._classProperties = new Map(superProto._classProperties!);
     // initialize Map populated in observedAttributes
     this._attributeToPropertyMap = new Map();
     // make any properties
--- a/packages/lit-html/CHANGELOG.md
+++ b/packages/lit-html/CHANGELOG.md
@@ -26,7 +26,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Breaking] Arrays and other iterables passed to attribute bindings are not specially handled. Arrays will be rendered with their default toString representation. This means that ```html`<div class=${['a', 'b']}>``` will render `<div class="a,b">` instead of `<div class="a b">`. To get the old behavior, use `array.join(' ')`.
 * Multiple bindings in a single attribute value don't require the attribute value is quoted, as long as there is no whitespace or other attribute-ending character in the attribute value. ```html`<div id=${a}-${b}>```
 * [Breaking] The directive and part APIs are significantly different. See the [README](README.md) for mroe details.
-* Private names are now named with a `_` prefix instead of `__`.
 
 ### Added
 * Added `renderBefore` to render options. If specified, content is rendered before the node given via render options, e.g. `{renderBefore: node}`.
--- a/packages/lit-html/CHANGELOG.md
+++ b/packages/lit-html/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Breaking] Arrays and other iterables passed to attribute bindings are not specially handled. Arrays will be rendered with their default toString representation. This means that ```html`<div class=${['a', 'b']}>``` will render `<div class="a,b">` instead of `<div class="a b">`. To get the old behavior, use `array.join(' ')`.
 * Multiple bindings in a single attribute value don't require the attribute value is quoted, as long as there is no whitespace or other attribute-ending character in the attribute value. ```html`<div id=${a}-${b}>```
 * [Breaking] The directive and part APIs are significantly different. See the [README](README.md) for mroe details.
+* Private names are now named with a `_` prefix instead of `__`.
 
 ### Added
 * Added `renderBefore` to render options. If specified, content is rendered before the node given via render options, e.g. `{renderBefore: node}`.
--- a/packages/lit-html/rollup.config.js
+++ b/packages/lit-html/rollup.config.js
@@ -29,12 +29,4 @@ export default litRollupConfig({
     'lit-html',
     'parts',
   ],
-  reservedProperties: [
-    '_$litType$',
-    '_$litDirective$',
-    // TODO Decide on public API
-    // https://github.com/Polymer/lit-html/issues/1261
-    '_value',
-    '_setValue',
-  ],
 });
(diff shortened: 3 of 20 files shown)


packages/lit-html/src/directives/repeat.ts
```diff
@@ -71,7 +71,7 @@ class RepeatDirective extends Directive {
     }
   }
 
-  __getValuesAndKeys<T>(
+  _getValuesAndKeys<T>(
```
sorvell, line 74: `private` ?

packages/lit-html/src/lit-html.ts
```diff
@@ -318,20 +318,20 @@ export abstract class Directive {
 }
 
 class Template {
-  private __strings: TemplateStringsArray;
-  __element: HTMLTemplateElement;
-  __parts: Array<TemplatePart> = [];
+  private _strings: TemplateStringsArray;
+  _element: HTMLTemplateElement;
```
sorvell, line 322: `private` ?

packages/lit-html/src/lit-html.ts
```diff
@@ -318,20 +318,20 @@ export abstract class Directive {
...
+  _parts: Array<TemplatePart> = [];
```
sorvell, line 323: `private` ?

sorvell (changes requested):
#1315 actually requires access to:
NodePart
* _value
* _startNode
*  _endNode
* options
* _setValue

Template
*  _createElement

packages/lit-html/src/directives/repeat.ts
aomarks: Done.

packages/lit-html/src/lit-html.ts
aomarks: Same logic as https://github.com/Polymer/lit-html/pull/1334/files/a420729a180f98727e27ebedc1ad46e02189b3ba#r497248149

packages/lit-html/src/lit-html.ts
aomarks: Since `Template` is not an exported class, and `_element` is accessed within this module but outside this class, this is really "package private" and it seems simpler to just keep it public, otherwise we'll need to cast it.

I guess if a `Template` instance was accessible by some other public typed route, even though the class itself is not exported, then it would make sense for it to be `private` and cast, but it doesn't look like that's the case.

I did find a few other misc properties that could safely become `private`.

aomarks (commented):
> #1315 actually requires access to:
> NodePart
> 
> * _value
> * _startNode
> * _endNode
> * options
> * _setValue
> 
> Template
> 
> * _createElement

Ok, added those to the config in `rollup-common.js`. I didn't include `options`, since that's public.

packages/lit-html/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [Breaking] Arrays and other iterables passed to attribute bindings are not specially handled. Arrays will be rendered with their default toString representation. This means that ```html`<div class=${['a', 'b']}>``` will render `<div class="a,b">` instead of `<div class="a b">`. To get the old behavior, use `array.join(' ')`.
 * Multiple bindings in a single attribute value don't require the attribute value is quoted, as long as there is no whitespace or other attribute-ending character in the attribute value. ```html`<div id=${a}-${b}>```
 * [Breaking] The directive and part APIs are significantly different. See the [README](README.md) for mroe details.
+* Private names are now named with a `_` prefix instead of `__`.
```
justinfagnani, line 29: I'm not sure this needs to go in the changelog since it's not publicly visible.

aomarks:
> > #1315 actually requires access to:
> > NodePart
> > 
> > * _value
> > * _startNode
> > * _endNode
> > * options
> > * _setValue
> > 
> > Template
> > 
> > * _createElement
> 
> Ok, added those to the config in `rollup-common.js`. I didn't include `options`, since that's public.

Actually just realized `_value` and `_setValue` are already reserved from any mangling at all in lit-html until we decide whether this is a public API or not (see https://github.com/Polymer/lit-html/issues/1261) -- and that made me realize we should share the reserved property config between lit-html and lit-element or else it would be renamed in lit-element but not lit-html.

packages/lit-element/src/lib/updating-element.ts
```diff
@@ -396,9 +396,9 @@ export abstract class UpdatingElement extends HTMLElement {
     }
     this[finalized] = true;
     // finalize any superclasses
-    const superProto = Object.getPrototypeOf(this);
+    const superProto = Object.getPrototypeOf(this) as typeof UpdatingElement;
```
justinfagnani, line 399: what made this cast necessary?

packages/lit-html/rollup.config.js
```diff
@@ -29,4 +29,12 @@ export default litRollupConfig({
     'lit-html',
     'parts',
   ],
+  reservedProperties: [
+    '_$litType$',
```
justinfagnani: Could we do a pattern instead? This would also enforce that these special properties are named similarly.

rollup-common.js
```diff
@@ -39,7 +39,23 @@ const skipBundleOutput = {
   },
 };
 
-export function litRollupConfig({ entryPoints, external = [] } = options) {
+// Any private properties which we share between different _packages_ are
+// hard-coded here because they must never change between versions. Mangled
+// names are randomly chosen uppercase letters, in case we ever might want to
+// use lowercase letters for short, public APIs.
+const crossPackagePropertyMangles = {
+  _createElement: "Y",
```
justinfagnani, line 56: We should consider prefixing these, like with `$`

rollup-common.js
```diff
@@ -64,12 +80,28 @@ export function litRollupConfig({ entryPoints, external = [] } = options) {
   // of all our code in a single file, tell Terser to minify that, and then throw
   // it away. This seeds the name cache in a way that guarantees every property
   // gets a unique mangled name.
-  const nameCache = {};
+  const nameCache = {
+    props: {
+      // Note all properties in the terser name cache are prefixed with '$'
+      // (presumably to avoid collisions with built-ins).
```
justinfagnani, line 89: Maps are useful :)

justinfagnani (approved):
LGTM, with some small non-blocking comments.

rollup-common.js
aomarks: When you're using the CLI this is written to a JSON file, so I guess that's why they do it this way.

packages/lit-element/src/lib/updating-element.ts
justinfagnani: Huh. I guess https://github.com/Microsoft/TypeScript/issues/22875 is the latest on this. Doesn't seem like TS will track prototype type information...

rollup-common.js
justinfagnani: It's not a slam-dunk idea, but it'd be to clearly mark that a property is special in case someone's looking at these objects in devtools.

aomarks: They aren't _that_ special though. They're just APIs that we happened to need to share across the internal package boundaries we invented?

justinfagnani: They're special compared to public properties. We don't want other code to use these properties.

aomarks: Right, but they're not too special compared to other minified properties, which is what a `$` would distinguish them from?

aomarks: Going to merge to move forward, but we can revisit this in a follow up if needed.

PR #1336, opened 2020-09-30
Pass eventContext to PropertyCommitter

### LINKED ISSUE #1335: Pass eventContext to PropertyCommitter
# Description
Right now, `options` are passed to NodeParts https://github.com/Polymer/lit-html/blob/master/src/lib/default-template-processor.ts#L55 and the `eventContext` part of `options` is passed to `EventPart`: https://github.com/Polymer/lit-html/blob/master/src/lib/default-template-processor.ts#L42

Is there a reason that `options` or `options.eventContext` are not passed to `PropertyCommitter`?

Use case: 
Declaratively define a renderer for a vaadin-grid-column, something like
```
<vaadin-grid-column .renderer=${renderer((data:MyData) => html`<span @click="${this.nameClick}">{$data.name}</span>`)}>
```

This works fine except that when rendering the inner template you have no access to `eventContext` so `this` points to something else.

The fix would be quite trivial, something like
```
-      const committer = new PropertyCommitter(element, name.slice(1), strings);
+      const committer = new PropertyCommitter(element, name.slice(1), strings, options.eventContext);
```

```
-
-  constructor(element: Element, name: string, strings: ReadonlyArray<string>) {
+  readonly eventContext?: EventTarget;
+  
+  constructor(element: Element, name: string, strings: ReadonlyArray<string>, eventContext?: EventTarget) {
     super(element, name, strings);
+    this.eventContext = eventContext;
```

Artur-:
Fixes #1335

### DIFF
--- a/src/lib/default-template-processor.ts
+++ b/src/lib/default-template-processor.ts
@@ -35,7 +35,8 @@ export class DefaultTemplateProcessor implements TemplateProcessor {
       options: RenderOptions): ReadonlyArray<Part> {
     const prefix = name[0];
     if (prefix === '.') {
-      const committer = new PropertyCommitter(element, name.slice(1), strings);
+      const committer = new PropertyCommitter(
+          element, name.slice(1), strings, options.eventContext);
       return committer.parts;
     }
     if (prefix === '@') {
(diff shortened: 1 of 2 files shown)


justinfagnani:
Closing this since 2.0 makes render options available to directives via parts.

PR #1338, opened 2020-09-30
[lit-next] Bump to latest wtr and increase browser concurrency

aomarks:
Enables the new iframe mode (see https://github.com/modernweb-dev/web/issues/472#issuecomment-701630841) and increases browser concurrency on CI. This drops the sauce test time from ~7m30s to ~3m40s!
Also adds `npm run nuke` command to bump all dependencies in all packages.


packages/tests/web-test-runner.config.js
```diff
@@ -143,7 +143,8 @@ export default {
     "../lit-element/development/**/*_test.js",
   ],
   nodeResolve: true,
-  concurrency: sauceLauncher ? 1 : 10,
+  concurrency: sauceLauncher ? 1 : 10, // default cores / 2
```
LarsDenBakker, line 146: we *should* be able up the test file concurrency now as well in saucelabs, 6 seemed to work for me

aomarks: Hm, got this failure twice in a row with the production mode:

```
Using production builds
[Saucelabs] Setting up Sauce Connect proxy...

Error while running tests:
Error: 30 Sep 21:58:39 - Sauce Connect could not establish a connection.
30 Sep 21:58:39 - Please check your firewall and proxy settings.
30 Sep 21:58:39 - Goodbye.

    at Socket.<anonymous> (/home/runner/work/lit-html/lit-html/packages/tests/node_modules/saucelabs/build/index.js:316:25)
    at Socket.emit (events.js:314:20)
    at addChunk (_stream_readable.js:307:12)
    at readableAddChunk (_stream_readable.js:282:9)
    at Socket.Readable.push (_stream_readable.js:221:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:188:23)
```

But it could be totally unrelated, will retry again. Maybe some retry logic is all that's needed for this kind of failure?

LarsDenBakker: It sounds unrelated... potentially the dev + prod run follow each other too fast now? Or could it be that tunnels are still staying open on the saucelabs side? I've still been seeing this issue, even though I think we're closing the tunnel properly on our side.

LarsDenBakker: I tested the lit-next branch a lot with the iframe mode, with concurrentBrowsers set to 2 (since our open source plan doesn't allow more) and concurrency 6.

aomarks: That worked in both jobs, so yeah maybe some subtle timing thing. New times were 3m30s and 3m50s. Very nice!

.github/workflows/tests.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
```
justinfagnani, line 52: Why do we have Sauce env vars in the test-local job?

.github/workflows/tests.yml
```diff
@@ -109,4 +110,5 @@ jobs:
...
           BROWSERS: preset:sauce
+          CONCURRENT_BROWSERS: 2
```
justinfagnani, line 113: What's the reason for the different values? Can you add a comment for future readers?

justinfagnani (approved):

.github/workflows/tests.yml
aomarks: Oops, removed, just forgot to remove that when I first split apart the jobs.

.github/workflows/tests.yml
aomarks: Added comment.

PR #1339, opened 2020-09-30
[lit-html] Add CompiledTemplateResult

justinfagnani:
This add a second type of template result named `CompiledTemplateResult` that has a reference to a Template rather than template strings. When committing template results we check the type of result and if it's a `CompiledTemplateResult` we skip the template cache and prep steps.

This will let us build a lit-html template compiler, as well as other ways to author a template that don't use the `html` tag, such as templates in HTML.


packages/benchmarks/lit-html/kitchen-sink/index-compiled.ts
```diff
@@ -0,0 +1,246 @@
...
+    }),
+  };
+};
+
+// These are "hand compiled" templates for the benchmark. This allows lit-html
+// to skip the template parsing/prep steps for faster bootup.
+const _$lit_template_1$ = {
+  _element: document.createElement('template'),
+  _parts: [
+    {_type:1,_index:1,_name:"class",_strings:[""," static"], _constructor: AttributePart},
```
kevinpschaaf, line 151: Since we already have the `PartType` enum, we could use those numbers here to indirect the constructor, but then we'd need to add code in the core to have that switch.

I guess this could also compile to `get _constructor: window.litGetPartCtor(1)` where that is installed somehow.

kevinpschaaf: That brings up a separate point: all of these names will be mangled right? Seems like we need a gambit for that. Make them all `stableProperties` and the compiler compiles to that?

justinfagnani: Stable properties sounds right, since the compiled templates should work with dev and prod

justinfagnani: Do you see anything particularly bad with using private-ssr-support for this? We could rename that...

justinfagnani: I added relevant properties to stable properties list. Still using private-ssr-support to provide the stable name for TemplateImpl and part constructors. We could make stable properties for them too though, with little code-size impact.

justinfagnani: We might want to export Template and parts like https://github.com/Polymer/lit-html/pull/1692 did perviously, to avoid bringing in ssr support into compiled modules.

justinfagnani:
@kevinpschaaf PTAL. Tests are passing now, just need to decide what to do about names and using ssr-support.

packages/lit-html/src/lit-html.ts
```diff
@@ -519,19 +527,17 @@ const getTemplateHtml = (
 };
 
 export type Template = Interface<TemplateImpl>;
+export type {TemplateImpl};
```
sorvell, line 530: Call this `TemplateType`?

Seems slightly unfortunate to export a thing called `Impl`.

packages/lit-html/src/lit-html.ts
```diff
@@ -1046,13 +1053,22 @@ class ChildPartImpl {
...
 
-  private _commitTemplateResult(result: TemplateResult): void {
-    const {values, strings} = result;
-    const template = this._$getTemplate(strings, result);
+  private _commitTemplateResult(
+    result: TemplateResult | CompiledTemplateResult
+  ): void {
+    const {values} = result;
+    let template: Template | undefined;
+    if (typeof result._$litType$ === 'number') {
```
sorvell, line 1061: This isn't a "type" anymore; maybe just `_$lit$`?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1468,6 +1477,82 @@ suite('lit-html', () => {
     );
   });
 
+  suite('compiled', () => {
```
sorvell, line 1480: Nit. Maybe break these out into a separate test module just for readability?

packages/lit-html/src/lit-html.ts
```diff
@@ -226,6 +226,14 @@ export type TemplateResult<T extends ResultType = ResultType> = {
 };
 
 export type SVGTemplateResult = TemplateResult<typeof SVG_RESULT>;
+export type TemplateFactory = (options?: RenderOptions) => Template;
```
sorvell, line 229: It's sort of unfortunate that the factory means you have to call a function and get a memoized result for every TemplateResult. We also can't change this since it's technically public. Maybe consider not making this a factory and require the compiler to handle shady scoping?

packages/lit-starter-js/docs/index.html
```diff
@@ -1,75 +1,96 @@
-
-<!doctype html>
+<!DOCTYPE html>
```
sorvell, line 1: Unfortunate that there are so many formatting changes here.

rollup-common.js
```diff
@@ -54,65 +54,72 @@ const reservedProperties = ['_$litType$', '_$litDirective$', '_$litPart$'];
...
+
+for (const [prop, mangle] of Object.entries(stableProperties)) {
   if (!prop.startsWith('_$')) {
     throw new Error(
       `stableProperties should start with prefix '_$' ` +
         `(property '${prop}' violates the convention)`
     );
   }
-  if (mangle.charCodeAt(0) !== 'A'.charCodeAt(0) + i) {
+  if (mangle.startsWith('Σ')) {
```
sorvell, line 110: What case is this?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
justinfagnani: We have to do something to get options from the ChildPart into preparing the template. I can try to push this into core, but there will still be some check that the template was prepared.

packages/lit-starter-js/docs/index.html
justinfagnani: I'll revert these

packages/lit-html/src/lit-html.ts
justinfagnani: I got rid of the factory with a touch more logic in _commitTemplateResult. This makes compiled templates smaller.

rollup-common.js
justinfagnani: It was for a case when I wanted public, short, renamed properties like `Σe` for `element`. Removed.

packages/lit-html/src/lit-html.ts
justinfagnani: It's kind of back to a type with it not being a function. The "type" is the template it references.

packages/lit-html/src/lit-html.ts
justinfagnani: done

packages/lit-html/src/hydrate.ts
```diff
@@ -255,22 +255,17 @@ const openChildPart = (
...
-    const markerWithDigest = `lit-part ${digestForTemplateResult(
-      value as TemplateResult
-    )}`;
+    const markerWithDigest = `lit-part ${digestForTemplateResult(value)}`;
     if (marker.data === markerWithDigest) {
-      const template = ChildPart.prototype._$getTemplate(
-        (value as TemplateResult).strings,
-        value as TemplateResult
-      );
+      const template = ChildPart.prototype._$getTemplate(value);
```
justinfagnani, line 260: btw, this isn't doing much on the prototype. I think we should move it to static like with `Template.createElement`.

rollup-common.js
```diff
@@ -54,65 +54,67 @@ const reservedProperties = ['_$litType$', '_$litDirective$', '_$litPart$'];
...
 };
 
+const alpha = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
+const validMangledNames = [...alpha, ...alpha.map((c) => `A${c}`)];
+
 // Validate stableProperties list, just to be safe; catches dupes and
 // out-of-order mangled names
-Object.entries(stableProperties).forEach(([prop, mangle], i) => {
+let mangledNameCount = 0;
+const reservedPropertySet = new Set();
```
justinfagnani, line 101: removed

packages/lit-html/src/lit-html.ts
```diff
@@ -653,13 +665,14 @@ class TemplateImpl {
       }
       nodeIndex++;
     }
+    // console.log('Template', this);
```
justinfagnani, line 668: removed

sorvell (approved):
LGTM

kevinpschaaf (approved):

PR #1341, opened 2020-10-01
[lit-next] Disable benchmarks action on forks

aomarks:
Unfortunately we can't currently run benchmarks on PRs from forked repos, because the tachometer action reports results by posting a comment, which requires a github token that isn't available.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,6 +5,12 @@ on: [pull_request]
 jobs:
   benchmarks:
     name: benchmarks
+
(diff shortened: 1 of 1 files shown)


justinfagnani (approved):

aomarks:
cc @andrewiggins do you know if it's possible for a github action to report as a *check*? Checks can have markdown in them. I prefer the comment because it's very visible, but it does seem to have this downside that it won't work for forks.

andrewiggins:
@aomarks Ack. I've opened andrewiggins/tachometer-reporter-action#29 & andrewiggins/tachometer-reporter-action#30 to address this. Let me know your thoughts!

PR #1342, opened 2020-10-01
[lit-next] Add prettier-plugin-package

abdonrd:
Use Prettier plugin to auto-format the `package.json` files.


abdonrd:
Moved to #1351.

PR #1343, opened 2020-10-01
[lit-next] Add missed ignore-path

abdonrd:
To ignore the builded files.


abdonrd:
Friendly ping to @aomarks.

abdonrd:
Moved to #1351.

PR #1348, opened 2020-10-07
[lit-next] Adds `static shadowRootOptions`

sorvell:
This makes it a bit easier to customize options for the shadowRoot. Rather than overriding `createRenderRoot`, this property can be set.

Micro-improvement: this also avoids creating an options object per instance.


justinfagnani (approved):
🎉

Issue #12, opened 2017-07-28
Long term vision


LarsDenBakker:
As a long time user of Polymer, this project expresses exactly what I feel is missing from our toolbox when building apps in Polymer.

What's the long term vision for this, and it's relation to Polymer? Is it something that will be officially supported? I'd hate to see this end up gathering dust.

justinfagnani:
At this point we're still experimenting and benchmarking.

Regarding Polymer, Polymer 2.0 is organized in a way that lets you pick and choose the features to use, and only "pay" for those features, not the others, via mixins. It's possible to use Polymer's sugar and property change management with alternate, or even no, templating systems. Monica gave a good talk that touches on this at this year's I/O: https://www.youtube.com/watch?v=zfQoleQEa4w

justinfagnani:
I think this is answered as well as can be for the moment. We'll know more as the project matures!

LarsDenBakker:
Indeed, thanks. I will keep keep track of the project:)

PR #1351, opened 2020-10-08
[lit-next] Improve the ESLint and Prettier setup

abdonrd:
In this PR I address a few related things:

---

- Move the `lint` command to the root.
  - Now there is one entry point to lint all the project.
  - Having lerna in between, made us lose the colors and links in the terminal that ESLint provides. Now we have them again; a nice improvement to the development experience!

---

- Move the `format` command to the root.
  - Now there is one entry point to lint all the project.
  - In addition to Prettier, now we also auto-format with the ESLint `--fix` option.

---

- Now we also use the `.gitignore` file in ESLint and Prettier.

---

- Now we have the `.gitignore` files unified.

---

- ~~Now use Prettier plugin ([`prettier-plugin-package`](https://github.com/shellscape/prettier-plugin-package)) to auto-format the `package.json` files.~~
  - Temporally removed because it also sorts the `files` field: [prettier-plugin-package #12](https://github.com/shellscape/prettier-plugin-package/issues/12).
With my proposal PR: [prettier-plugin-package #17](https://github.com/shellscape/prettier-plugin-package/pull/17)
PR merged! Waiting the next release...

---

- Now we also lint and format the root files (like the `rollup-common.js`, etc).

---

- Now all the project is linted and formatted.
  - In new PRs we can try to remove temporally ignored rules, inlined ignored comments, etc.

---

- Now we have `husky` (as pre-commit hook) & `lint-staged` (to run linter & formatter only on the staged files).
This will avoid commits that don't comply the rules of the linter and the formatter. 🎉🎉

---


abdonrd:
Friendly ping to @aomarks!

aomarks (approved):
Wow, great work!

I like the idea of trying out husky and lint-staged. It looks like it's going to automatically format before commit? Will that just happen silently, or does it fail the commit command and just stage the fixed changes?

I'm going to hold off on merging so that @justinfagnani can review, since there are a lot of changes here.

abdonrd:
Thanks Alexander!

Yes, it makes changes automatically and silently.
But if the linter of formatter fails, it doesn't allow to commit. For example, if an ESLint rule is not met that cannot be automatically with the `--fix` flag.

abdonrd:
Updated now with the latest changes from `lit-next` branch!

abdonrd:
Updated now with the latest changes from `lit-next` branch!

aomarks:
@abdonrd thanks for merging again, I think there's just a conflict with the package-lock being out of sync, just need to delete it and re-install I think.

abdonrd:
Yay, thanks @aomarks! 🎉

Remember to update the dependencies locally so that everything works correctly!

PR #1355, opened 2020-10-08
[lit-next] Creates `updating element` package

### LINKED ISSUE #1352: [lit-next] Create separate package for `updating-element`
## Description
Users have requested that `updating-element` be a separate package. With the lit mono-repo, this should be fairly straightforward. One issue is where to put lit-element's decorators. 

## Acceptance criteria
PR submitted.

sorvell:
Fixes #1352 

* Creates an updating-element package
* Moves updating-element out of lit-element
* Removes the /lib folder from lit-element
* Moves all decorators to the updating-element package but re-exports them from LitElement
* UpdatingElement now creates a renderRoot/shadowRoot via `createRenderRoot` and has support for `static styles`.
* Moves relevant tests from LitElement to UpdatingElement.
* Minor changes to UpdatingElement:
  * removes the `initialize` method in favor of the `constructor`
  * creates `renderRoot` before first update rather than in the `constructor`
  * removes the `adoptStyles` method and does that work in `createRenderRoot` instead

WIP: platform-support module not yet moved from LitElement. The non-lit portion should be moved to UpdatingElement.


bicknellr:
(continuing my naming bikeshedding from the meeting)
Maybe "`UpdatingComponent`"? That might reduce some of the perceived "this shouldn't use shadow roots" implication of the name and history of `UpdatingElement`.

justinfagnani:
@bicknellr I think we should keep the naming convention of subclasses of `Element` ending in `Element`.

packages/lit-element/src/lit-element.ts
```diff
@@ -54,7 +54,7 @@
  *
  * @packageDocumentation
  */
-import {PropertyValues, UpdatingElement} from './lib/updating-element.js';
+import {PropertyValues, UpdatingElement} from 'updating-element';
```
justinfagnani, line 57: Let's make sure we get consensus on the package name

packages/updating-element/README.md
```diff
@@ -0,0 +1,174 @@
+# LitElement 3.0 Pre-release
```
justinfagnani, line 1: Update this? It can be extremely minimal for now.

packages/updating-element/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "test:prod:watch": "MODE=prod npm run test:dev -- --watch",
+    "test:watch": "npm run test:dev -- --watch",
+    "checksize": "rollup -c --environment=CHECKSIZE",
+    "format": "prettier \"**/*{cjs,html,js,json,md,ts}\" --write",
+    "lint": "eslint \"**/*.{js,ts}\" --ignore-path .gitignore",
+    "check-version": "node scripts/check-version-tracker.js",
+    "prepublishOnly": "npm run check-version && npm run lint && npm test",
+    "regen-package-lock": "rm -rf node_modules package-lock.json; npm install",
+    "publish-dev": "npm test && VERSION=${npm_package_version%-*}-dev.`git rev-parse --short HEAD` && npm version --no-git-tag-version $VERSION && npm publish --tag dev",
+    "release": "np --any-branch --yolo"
```
justinfagnani, line 54: remove the release script

packages/updating-element/scripts/check-version-tracker.js
```diff
@@ -0,0 +1,30 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+
+const packageFile = fs.readFileSync(path.resolve('package.json'));
+const {version} = JSON.parse(packageFile);
+const ts = fs.readFileSync(path.resolve('src/lit-element.ts'));
+
+if (!ts.includes(`'${version}'`)) {
+  console.log(
+    `\nExpected lit-element.ts to contain current version "${version}"`
```
justinfagnani, line 24: update filename

packages/updating-element/scripts/check-version-tracker.js
```diff
@@ -0,0 +1,30 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+const packageFile = fs.readFileSync(path.resolve('package.json'));
+const {version} = JSON.parse(packageFile);
+const ts = fs.readFileSync(path.resolve('src/lit-element.ts'));
```
justinfagnani, line 20: update filename

packages/updating-element/typedoc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "readme": "docs/_api/api-readme.md",
```
justinfagnani, line 2: I think we can remove this for now

packages/updating-element/src/env.d.ts
```diff
@@ -0,0 +1,44 @@
+interface ShadyCSS {
+  nativeCss: boolean;
```
justinfagnani, line 2: is this required in updating element yet?

justinfagnani (approved):

packages/updating-element/README.md
sorvell: Done.

packages/updating-element/package.json
sorvell: Done.

packages/updating-element/scripts/check-version-tracker.js
sorvell: Done.

packages/updating-element/scripts/check-version-tracker.js
sorvell: Done.

packages/updating-element/typedoc.json
sorvell: Done.

packages/updating-element/src/env.d.ts
sorvell: It is now, yeah.

packages/lit-element/src/lit-element.ts
```diff
@@ -127,127 +105,16 @@ export class LitElement extends UpdatingElement {
...
-        }).renderBefore ??= style;
-      });
+  protected createRenderRoot() {
+    const renderRoot = super.createRenderRoot();
+    if (!supportsAdoptingStyleSheets) {
+      (this._renderOptions as {
+        renderBefore: ChildNode;
+      }).renderBefore ??= renderRoot!.lastChild as ChildNode;
     }
+    return renderRoot;
```
justinfagnani, line 117: Could this be simplified to:

```suggestion
  protected createRenderRoot() {
    const renderRoot = super.createRenderRoot();
    (this._renderOptions as {
        renderBefore: ChildNode;
      }).renderBefore ??= renderRoot!.firstChild;
    }
    return renderRoot;
```

If we always render to the first thing in the container, we'll go before emulated adoptedStyleSheets.

Also, is there no way around that cast? Maybe `this._renderOptions.renderBefore! ?? =`

packages/updating-element/src/css-tag.ts
```diff
@@ -105,3 +107,38 @@ export const css = (
   }
   return result;
 };
+
+/**
+ * Applies the given styles to a `shadowRoot`. When Shadow DOM is
+ * available but `adoptedStyleSheets` is not, styles are appended to the the
+ * `shadowRoot` to [mimic spec behavior](https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets).
```
justinfagnani, line 114: I think we should mention that rendered DOM should go _before_ emulated adopted styles to match the spec for ordering between `<style>` tags and adopted styles.

packages/lit-element/src/lit-element.ts
```diff
@@ -123,131 +101,20 @@ export class LitElement extends UpdatingElement {
...
-    return wasFinalized;
-  }
-
-  /**
-   * Node or ShadowRoot into which element DOM should be rendered. Defaults
-   * to an open shadowRoot.
-   */
-  readonly renderRoot!: HTMLElement | DocumentFragment;
-  readonly _renderOptions!: RenderOptions;
+  readonly _renderOptions: RenderOptions = {eventContext: this};
```
kevinpschaaf, line 104: Why readonly if we need to modify it? It's private so it's for us.

rollup-common.js
```diff
@@ -63,7 +64,11 @@ const crossPackagePropertyMangles = {
   _element: "E",
   _options: "O",
   _template: "t",
-  _renderOptions: "R"
+  _renderOptions: "R",
+  _baseConnectedCallback: "bC",
+  _baseUpdate: "bU",
```
kevinpschaaf, line 68: Nit: the naming here (also, `_template: t`) doesn't match the description above:
> Mangled names are ***randomly chosen uppercase letters***, in case we ever might want to use lowercase letters for short, public APIs.

rollup-common.js
```diff
@@ -63,7 +64,11 @@ const crossPackagePropertyMangles = {
   _element: "E",
   _options: "O",
   _template: "t",
-  _renderOptions: "R"
+  _renderOptions: "R",
```
kevinpschaaf, line 66: Can we add a comment after each field here to name the class/object they are associate with, just for bookkeeping?

packages/updating-element/README.md
```diff
@@ -0,0 +1,149 @@
...
+```html
+<my-element mood="awesome"></my-element>
+```
+
+Note, this example uses decorators to create properties. Decorators are a proposed
+standard currently available in [TypeScript](https://www.typescriptlang.org/) or [Babel](https://babeljs.io/docs/en/babel-plugin-proposal-decorators). UpdatingElement also supports a [vanilla JavaScript method](https://lit-element.polymer-project.org/guide/properties#declare) of declaring reactive properties.
+
+## Examples
+
+- Runs in all [supported](#supported-browsers) browsers: [Glitch](https://glitch.com/edit/#!/hello-lit-element?path=index.html)
```
kevinpschaaf, line 85: These are the lit-element links... :(

packages/lit-element/src/lit-element.ts
sorvell: Fixed. Changed from a readonly property to allow stuff like this.

packages/updating-element/src/css-tag.ts
sorvell: Done.

packages/lit-element/src/lit-element.ts
sorvell: We don't modify `_renderOptions` but we do modify the options themselves. Changed in lit-html.

packages/updating-element/README.md
sorvell: Tracked in https://github.com/Polymer/lit-html/issues/1359.

rollup-common.js
sorvell: Done.

kevinpschaaf: Nice, ty!

kevinpschaaf (approved):
LGTM, after fixing benchmarks issue

sorvell:
Merging with benchmarks in a questionable state. The thought is they should work when merged to `lit-next` and if not, will fix with a subsequent PR.

PR #1362, opened 2020-10-14
[lit-next] Refactors platform-support

sorvell:
* `platform-support` added to `lit-html`
* `platform-support` added to `updating-element`
* `platform-support` added to `lit-element`
  * The `lit-element` version pulls in `lit-html` and `updating-element` `platform-support`
  * Build adjusted to support bundling `platform-support` so that in `lit-element` only one file must be loaded.
* Moves `platform-support` tests to the appropriate package.
* Also adds a `lit.min` bundled build (built via separate command) which is `umd` for easy consumption.


packages/lit-element/README.md
```diff
@@ -1,7 +1,6 @@
 # LitElement 3.0 Pre-release
 
-[![Build Status](https://github.com/polymer/lit-html/workflows/Tests/badge.svg?branch=lit-next
-)](https://github.com/Polymer/lit-html/actions?query=workflow%3ATests)
+[![Build Status](https://github.com/polymer/lit-html/workflows/Tests/badge.svg?branch=lit-next)](https://github.com/Polymer/lit-html/actions?query=workflow%3ATests)
```
sorvell, line 3: This file should was just reformatted.

packages/lit-html/README.md
```diff
@@ -1,8 +1,8 @@
 # lit-html 2.0 Pre-release
+
```
sorvell, line 2: This file was just reformatted.

packages/tests/web-test-runner.config.js
```diff
@@ -158,7 +147,7 @@ export default {
   rootDir: '../',
   // Note this file list can be overridden by wtr command-line arguments.
   files: [
-    '../lit-html/development/**/*_test.js',
+    '../lit-html/development/**/*_test.(js|html)',
```
sorvell, line 150: Other than this change, this file was just re-formatted.

kevinpschaaf (approved):
LGTM

packages/updating-element/src/platform-support.ts
```diff
@@ -0,0 +1,143 @@
...
+  new (...args: any[]): PatchableUpdatingElement;
+  constructor: PatchableUpdatingElementConstructor;
+  connectedCallback(): void;
+  hasUpdated: boolean;
+  _afterUpdate(changedProperties: unknown): void;
+  createRenderRoot(): Element | ShadowRoot;
+  _renderOptions: RenderOptions;
+}
+
+(globalThis as any)['updatingElementPlatformSupport'] ??= ({
```
justinfagnani, line 51: I wonder if we need to key the platform support by version. In the case where a newer version of the file is loaded after an older version, we'd want the newer to overwrite so if it contains additional patches they'll take effect.

packages/lit-element/rollup.bundle.config.js
```diff
@@ -0,0 +1,21 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {litBundle} from '../../rollup-common.js';
+
+export default litBundle({
```
aomarks, line 17: You can export an array of configs in a rollup config -- maybe it would be simpler to just have the one `lit-element/rollup.config.js` that looks like this?

```js
export default [
  ...litProdConfig(...),
  ...litMonoBundleConfig(...),
]
```

rollup-common.js
```diff
@@ -205,30 +210,71 @@ export function litRollupConfig({
...
+        name,
+        terserOptions: terserOptions,
+      })
+    ),
   ];
-}
\ No newline at end of file
+}
+
+export const litBundle = ({
```
aomarks, line 256: `litBundle` and `litRollupConfig` are kinda undifferentiated now. How about something like:

- `litRollupConfig` -> `litProdConfig`
- `litBundle` -> `litMonoBundleConfig`

(Probably this whole rollup config organization needs a re-think now, like inverting the control somewhat by exporting some smaller pieces instead of these big functions, but we can think about that another time. Maybe I'll take a pass at it someday).

aomarks (approved):

packages/lit-element/rollup.bundle.config.js
sorvell: I decided to keep it separate to make sure it didn't get published to npm accidentally.

packages/updating-element/src/platform-support.ts
sorvell: The idea is that platform-support should be loaded once by the application. Loading multiple versions would generally be bad. It's also first one wins, so even if some bad component includes an old platform-support, this is easy to address by loading a new one right when you load your polyfills.

Since this is a corner case, suggest we consider any issues with this as a follow on.

rollup-common.js
sorvell: Done. And happy to have you clean this up as a follow up.

PR #1370, opened 2020-10-16
Port sanitization hooks to lit-next. 

rictic:
See #750 for details of this API.

This is a basic implementation. More of it can be probably be removed from the release build with some more terser cleverness.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -13,10 +13,87 @@
  */
 
 const DEV_MODE = true;
+const ENABLE_EXTRA_SECURITY_HOOKS = true;
 
 if (DEV_MODE) {
   console.warn('lit-html is in dev mode. Not recommended for production!');
 }
+
+/**
+ * Used to sanitize any value before it is written into the DOM. This can be
+ * used to implement a security policy of allowed and disallowed values in
+ * order to prevent XSS attacks.
+ *
+ * One way of using this callback would be to check attributes and properties
+ * against a list of high risk fields, and require that values written to such
+ * fields be instances of a class which is safe by construction. Closure's Safe
+ * HTML Types is one implementation of this technique (
+ * https://github.com/google/safe-html-types/blob/master/doc/safehtml-types.md).
+ * The TrustedTypes polyfill in API-only mode could also be used as a basis
+ * for this technique (https://github.com/WICG/trusted-types).
+ *
+ * @param node The HTML node (usually either a #text node or an Element) that
+ *   is being written to. Note that this is just an exemplar node, the write
+ *   may take place against another instance of the same class of node.
+ * @param name The name of an attribute or property (for example, 'href').
+ * @param type Indicates whether the write that's about to be performed will
+ *   be to a property or a node.
+ * @returns A function that will sanitize this class of writes.
+ */
+export type SanitizerFactory = (
+  node: Node,
+  name: string,
+  type: 'property' | 'attribute'
+) => ValueSanitizer;
+
+/**
+ * A function which can sanitize values that will be written to a specific kind
+ * of DOM sink.
+ *
+ * See SanitizerFactory.
+ *
+ * @param value The value to sanitize. Will be the actual value passed into
+ *   the lit-html template literal, so this could be of any type.
+ * @returns The value to write to the DOM. Usually the same as the input value,
+ *   unless sanitization is needed.
+ */
+export type ValueSanitizer = (value: unknown) => unknown;
+
+const identityFunction: ValueSanitizer = (value: unknown) => value;
+const noopSanitizer: SanitizerFactory = (
+  _node: Node,
+  _name: string,
+  _type: 'property' | 'attribute'
+) => identityFunction;
+
+let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
+
+/** Sets the global sanitizer factory. */
+export const setSanitizer = (newSanitizer: SanitizerFactory) => {
+  if (!ENABLE_EXTRA_SECURITY_HOOKS) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


rictic:
Looks like the playwright github action is failing to install/initialize?

packages/lit-html/src/lit-html.ts
```diff
@@ -757,6 +840,20 @@ export class NodePart {
...
     if (this._value !== value) {
       this._clear();
+      const parent = this._startNode.parentNode;
+      if (
+        ENABLE_EXTRA_SECURITY_HOOKS &&
+        sanitizerFactoryInternal !== noopSanitizer &&
+        (parent?.nodeName === 'STYLE' || parent?.nodeName === 'SCRIPT')
+      ) {
+        this._insert(
+          new Text(
```
justinfagnani, line 850: I vaguely remember discussing allowing TemplateResults, since they're presumed safe, just not other values.

And how does this work with safe value types? Should those be allowed?

packages/lit-html/src/lit-html.ts
```diff
@@ -774,11 +871,29 @@ export class NodePart {
...
+        // When setting text content, for security purposes it matters a lot
+        // what the parent is. For example, <style> and <script> need to be
+        // handled with care, while <span> does not. So first we need to put a
+        // text node into the document, then we can sanitize its contentx.
+        if (this._textSanitizer === undefined) {
+          this._textSanitizer = sanitizerFactory(textNode, 'data', 'property');
+        }
+        value = this._textSanitizer(value);
+      }
+      textNode.data = value as string;
```
justinfagnani, line 896: In the case of `<style>` and `<script>` this might be a detached Text because of the check in `_commitNode`, yes? Can you add a comment for that?

packages/lit-html/src/lit-html.ts
```diff
@@ -13,11 +13,88 @@
...
+export type ValueSanitizer = (value: unknown) => unknown;
+
+const identityFunction: ValueSanitizer = (value: unknown) => value;
+const noopSanitizer: SanitizerFactory = (
+  _node: Node,
+  _name: string,
+  _type: 'property' | 'attribute'
+) => identityFunction;
+
+let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
```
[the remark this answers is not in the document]
rictic: Done

packages/lit-html/src/lit-html.ts
```diff
@@ -13,11 +13,88 @@
...
+        ` setSanitizeDOMValueFactory should be called at most once.`
+    );
+  }
+  sanitizerFactoryInternal = newSanitizer;
+};
+
+/**
+ * Only used in internal tests, not a part of the public API.
+ */
+export const __testOnlyClearSanitizerFactoryDoNotCallOrElse = () => {
```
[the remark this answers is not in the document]
rictic: Done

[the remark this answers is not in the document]
rictic: Nice! Done

packages/lit-html/src/lit-html.ts
```diff
@@ -13,11 +13,88 @@
...
+  if (DEV_MODE && ENABLE_EXTRA_SECURITY_HOOKS) {
+    sanitizerFactoryInternal = noopSanitizer;
+  }
+};
+
+export const sanitizerFactory: SanitizerFactory = (node, name, type) => {
```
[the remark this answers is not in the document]
rictic: +1

This wasn't released open source, and it's only used in a few places internally, not expensive to change the name.

packages/lit-html/src/lit-html.ts
rictic: TemplateResults aren't presumed safe in all contexts. For example, this is a template result, but isn't safe to interpolate inside a style or script tag:

```typescript
html`${somethingFromUserInput}`
```

I believe you can commit SafeScript text as the text content of a `<script>` tag, and SafeStyle as the text content of a `<style>` tag, but in order to keep complexity down we're just banning rendering anything more complex inside a `<script>` or `<style>`

packages/lit-html/src/lit-html.ts
rictic: No, the check in `_commitNode` calls `_insert` which writes immediately to `this.startNode`.

packages/lit-html/src/lit-html.ts
justinfagnani: Won't that template result have any bindings within it checked by another call to `_commitNode`?

rictic: Hm, maybe? I think this option is safe and simple though, and we haven't had demand for the ability to write into scripts and style tags. Expanding options is back-compatible, contracting them isn't. Let's start with this.

justinfagnani: sounds good

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -34,6 +34,9 @@ import './polyfills.js';
 
 const ua = window.navigator.userAgent;
 const isIe = ua.indexOf('Trident/') > 0;
+// We don't have direct access to DEV_MODE, but this is a good enough
+// proxy.
+const DEV_MODE = render.setSanitizer != null;
```
justinfagnani, line 39: with the replacement rules we have, shouldn't we be able to do the same as in the main source: `const DEV_MODE = true;`?

rictic: Do we build our tests? It looks like no, making that change causes the prod tests to fail

justinfagnani: Ah, ok

justinfagnani (approved):

PR #1371, opened 2020-10-16
[lit-next] Add lint step to the CI

abdonrd:
Now that everything works and is up to date, I think we can add a lint step to the GitHub Actions.

Just 24 seconds:

Screenshot 2020-10-16 at 11 33 42


abdonrd:
Friendly ping to @aomarks

aomarks (approved):

PR #1372, opened 2020-10-16
[lit-next] Re-fires errors in update cycle asynchronously

sorvell:
Errors that occur during the update cycle are squelched to allow subsequent updates to proceed normally. However, they are now refired so developers have a chance to observe them. This can be done by implementing `window.onunhandledrejection`. Fixes https://github.com/Polymer/lit-element/issues/1081

### DIFF
--- a/packages/updating-element/src/updating-element.ts
+++ b/packages/updating-element/src/updating-element.ts
@@ -744,9 +744,7 @@ export abstract class UpdatingElement extends HTMLElement {
       // cycle. Errors are refired so developers have a chance to observe
       // them, and this can be done by implementing
       // `window.onunhandledrejection`.
-      Promise.resolve().then(() => {
-        throw e;
-      });
+      Promise.reject(e);
     }
     const result = this.performUpdate();
     // If `performUpdate` returns a Promise, we await it. This is done to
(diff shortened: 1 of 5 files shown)


packages/updating-element/src/updating-element.ts
```diff
@@ -740,8 +740,13 @@ export abstract class UpdatingElement extends HTMLElement {
       // This `await` also ensures that property changes are batched.
       await this._updatePromise;
     } catch (e) {
-      // Ignore any previous errors. We only care that the previous cycle is
-      // done. Any error should have been handled in the previous update.
+      // Refire any previous errors async so they do not disrupt the update
+      // cycle. Errors are refired so developers have a chance to observe
+      // them, and this can be done by implementing
+      // `window.onunhandledrejection`.
+      Promise.resolve().then(() => {
```
justinfagnani, line 747: Can you just do `Promise.reject(e)`?

Both `Promise.resolve().then(...)` and `Promise.reject(e)` seem to have the same behavior in devtools, though neither is triggering `onunhandledrejection` or `onerror` for me. Might just be devtools weirdness.

sorvell: Good call, done.

kevinpschaaf (approved):
LGTM, with minor docs fix

justinfagnani (approved):

PR #1373, opened 2020-10-16
[lit-next] Fixes queryAssignedNodes when querying for nodes on browsers that do …

sorvell:
…not support Element.matches

Fixes https://github.com/Polymer/lit-element/issues/1088

### DIFF
--- a/packages/updating-element/src/decorators/queryAssignedNodes.ts
+++ b/packages/updating-element/src/decorators/queryAssignedNodes.ts
@@ -80,6 +80,7 @@ export function queryAssignedNodes(
           nodes = nodes.filter(
             (node) =>
               node.nodeType === Node.ELEMENT_NODE &&
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
               ((node as any).matches
                 ? (node as Element).matches(selector)
                 : legacyMatches.call(node as Element, selector))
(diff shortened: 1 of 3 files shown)


packages/updating-element/src/decorators/queryAssignedNodes.ts
```diff
@@ -80,6 +80,7 @@ export function queryAssignedNodes(
           nodes = nodes.filter(
             (node) =>
               node.nodeType === Node.ELEMENT_NODE &&
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
justinfagnani, line 83: Free bytes: remove the whole check!

https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Browser_compatibility

justinfagnani (changes requested):

packages/updating-element/src/decorators/queryAssignedNodes.ts
sorvell: IE11 uses the prefix.

justinfagnani: argh... can we push this to the polyfills?

sorvell: Definitely: https://github.com/webcomponents/polyfills/issues/397

kevinpschaaf (approved):
LGTM

justinfagnani (approved):

Issue #1367, opened 2020-10-15
[lit-next] Add dev mode warnings


sorvell:
### Updating/LitElement
1. shadowed reactive properties
2. ShadyDOM use without platform-support
3. (optional?) changing properties after update
4. use of removed APIs like:
   * `static render` 
   * `static getStyles`
   * `adoptStyles`
   * `_getUpdateComplete`
   * `initialize`
   * `requestUpdateInternal`
   * undefined behavior for attributes?
   * warning when requestUpdate is used and noting that it no longer returns a promise (ideally this would happen only if you await or `then` it but that may not be possible?)

### lit-html
1. Warn on using directive in unsupported part type
2. Warn on bad HTML (e.g. scanner ends in unexpected state). This is already included but needs a test.

## Acceptance criteria
PR landed.

sorvell:
Fixed via https://github.com/Polymer/lit-html/pull/1375.

PR #1375, opened 2020-10-19
[lit-next] Adds dev mode warnings to UpdatingElement

sorvell:
Addresses part of #1367.

Adds warnings to UpdatingElement for:
* implementing removed API: `static render`, `static getStyles`, `adoptStyles`, `_getUpdateComplete`, `initialize`, `requestUpdateInternal`
* changed behavior for `toAttribute` returning `undefined`
* awaiting `requestUpdate`
* reactive properties shadowed by class fields
* ShadyDOM used without platform-support


packages/updating-element/src/updating-element.ts
```diff
@@ -436,13 +449,52 @@ export abstract class UpdatingElement extends HTMLElement {
...
         // index types
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         this.createProperty(p, (props as any)[p]);
       }
     }
     this.elementStyles = this.finalizeStyles(this.styles);
+    // DEV mode warnings
+    if (DEV_MODE) {
+      const warnRemoved = (obj: any, base: any, name: string) => {
+        if (obj[name] !== base[name]) {
```
kevinpschaaf, line 462: It was a little confusing that this isn't just a `obj[name] !== undefined` check, given that the warning is that "the API has been removed". 

This checks that there's not an _override_ on the user class, not that it doesn't exist at all. Is there a case where it's implemented on the base class (not removed altogether) and you're just not supposed to override it?

packages/updating-element/src/updating-element.ts
```diff
@@ -637,6 +689,14 @@ export abstract class UpdatingElement extends HTMLElement {
         (options.converter as ComplexAttributeConverter)?.toAttribute ??
         defaultConverter.toAttribute;
       const attrValue = toAttribute!(value, options.type);
+      if (DEV_MODE && attrValue === undefined) {
+        console.warn(
+          `The attribute value for the ` +
+            `${name as string} property is undefined. The attribute will be ` +
```
kevinpschaaf, line 695: Hmm, this doesn't indicate an error state, so this is more of a migration helper warning. Seems like you should be able to write code to the new rule (set properties to `undefined` and have them removed) without seeing a warning every time. The other warnings are always errors.  Maybe we need another debug warning level which is like "help me migrate".

kevinpschaaf (changes requested):

packages/updating-element/src/updating-element.ts
sorvell: Done

packages/updating-element/src/updating-element.ts
sorvell: Done

kevinpschaaf (approved):
LGTM, just a couple of warning prose suggestions, from the eye of a new user who isn't as versed with the lingo/details.

packages/updating-element/src/updating-element.ts
```diff
@@ -27,6 +27,13 @@ import {
 } from './css-tag.js';
 export * from './css-tag.js';
 
+const DEV_MODE = true;
+if (DEV_MODE) {
+  console.warn(
+    `updating-element is in dev mode. Not recommended for production!`
```
justinfagnani, line 33: Maybe strong wording than "Not recommended"? Like "Do not use in production!"

It'd also be great to include a link to docs on how to fix this. TODO?

packages/updating-element/src/updating-element.ts
```diff
@@ -436,13 +449,54 @@ export abstract class UpdatingElement extends HTMLElement {
...
     }
     this.elementStyles = this.finalizeStyles(this.styles);
+    // DEV mode warnings
+    if (DEV_MODE) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const warnRemoved = (obj: any, name: string) => {
+        if (obj[name] !== undefined) {
+          console.warn(
+            `\`${name}\` is implemented. It ` +
+              `has been removed from this version of UpdatingElement.`
```
justinfagnani, line 466: A link to changelogs would be good.

packages/updating-element/src/updating-element.ts
```diff
@@ -436,13 +449,54 @@ export abstract class UpdatingElement extends HTMLElement {
...
+          );
+        }
+      };
+      [`render`, `getStyles`].forEach((name: string) =>
```
justinfagnani, line 470: `render` was removed from LitElement, not UpdatingElement, yes?

packages/updating-element/src/updating-element.ts
```diff
@@ -436,13 +449,54 @@ export abstract class UpdatingElement extends HTMLElement {
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        warnRemoved(this as any, name)
+      );
+      [
+        `adoptStyles`,
```
justinfagnani, line 475: `adoptStyles` was also removed from LitElement

packages/updating-element/src/updating-element.ts
```diff
@@ -436,13 +449,54 @@ export abstract class UpdatingElement extends HTMLElement {
...
+        `requestUpdateInternal`,
+        `_getUpdateComplete`,
+      ].forEach((name: string) =>
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        warnRemoved(this.prototype as any, name)
+      );
+
+      // Issue platform support warning once only.
+      if (
+        window.ShadyDOM?.inUse &&
```
justinfagnani, line 486: I think it'd be nice to move this to the top level of the module. You won't need the `_issuedPlatformSupportWarning` field.

packages/updating-element/src/updating-element.ts
```diff
@@ -637,6 +691,20 @@ export abstract class UpdatingElement extends HTMLElement {
         (options.converter as ComplexAttributeConverter)?.toAttribute ??
         defaultConverter.toAttribute;
       const attrValue = toAttribute!(value, options.type);
+      if (
+        DEV_MODE &&
+        (this.constructor as typeof UpdatingElement).warnings!.has(
```
justinfagnani, line 696: I don't quite like the name `warnings` here. That sounds like some field I'd _read_ as a user to see what warnings were triggered. `showWarnings`?

packages/updating-element/src/updating-element.ts
```diff
@@ -779,6 +847,28 @@ export abstract class UpdatingElement extends HTMLElement {
...
     }
+    // create renderRoot before first update.
+    if (!this.hasUpdated) {
+      // Produce warning if any class properties are shadowed by class fields
+      if (DEV_MODE) {
+        (this.constructor as typeof UpdatingElement).elementProperties!.forEach(
+          (_v, p) => {
+            if (this.hasOwnProperty(p) && !this._instanceProperties?.has(p)) {
+              console.warn(
+                `Reactive properties cannot be initialized using class ` +
```
justinfagnani, line 858: I'd change this warning text since in source text you _can_ initialize class fields with initializers depending on decorators and compiler settings. You mention decorators, but then those sometimes don't work depending on the compiler. I would definitely link to docs from this message. Want to chat to brainstorm the message?

packages/updating-element/src/updating-element.ts
```diff
@@ -926,3 +1024,50 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    options?: PropertyDeclaration
+  ) {
+    requestUpdate.call(this, name, oldValue, options);
+    return requestUpdateThenable;
+  };
+}
+
+if (DEV_MODE) {
+  // Default warning set.
+  UpdatingElement.warnings = new Set(['change-in-update']);
```
justinfagnani, line 1059: I'd probably make this field an array, just for the simpler syntax if someone wants to write `warnings = ['migrate']`. It's not going to be faster to search a set than a small list, and duplicates aren't a problem.

justinfagnani (changes requested):

packages/updating-element/src/updating-element.ts
sorvell: Changed.

packages/updating-element/src/updating-element.ts
sorvell: Updated message.

packages/updating-element/src/updating-element.ts
sorvell: Changed to `enabledWarnings`.

packages/updating-element/src/updating-element.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
sorvell: Moved

packages/updating-element/src/updating-element.ts
sorvell: Moved

packages/updating-element/src/updating-element.ts
sorvell: Added TODO.

packages/updating-element/src/updating-element.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
```diff
@@ -926,3 +1021,55 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    requestUpdate.call(this, name, oldValue, options);
+    return requestUpdateThenable;
+  };
+}
+
+if (DEV_MODE) {
+  // Default warning set.
+  UpdatingElement.enabledWarnings = ['change-in-update'];
+  const ensureOwnWarnings = function (ctor: typeof UpdatingElement) {
+    if (!ctor.hasOwnProperty('enabledWarnings')) {
```
justinfagnani, line 1058: I don't think we need the rest of this code. Users can just write:

```js
UpdatingElement.enabledWarnings?.push('my-warning')
```

This will work in prod mode, as the push() is skipped. And duplicates will be fine, as the code that checks enabled warnings will get a hit on the first duplicate. There probably won't be many duplicates since this is debug code, and perf won't be a problem with small lists.

But also my comment about using an array was assuming that users could just assign the whole array:

```js
UpdatingElement.enabledWarnings = ['my-warning']
```

If that's not viable because you don't want them to override the default of `['change-in-update']`, then the Set is just fine too, since `UpdatingElement.enabledWarnings?.push('my-warning')` isn't so different from `UpdatingElement.enabledWarnings?.add('my-warning')`.

packages/updating-element/src/updating-element.ts
```diff
@@ -926,3 +1021,55 @@ export abstract class UpdatingElement extends HTMLElement {
...
+          `Use \`updateComplete\` instead.`
+      );
+      if (onfulfilled !== undefined) {
+        onfulfilled(false);
+      }
+    },
+  };
+
+  const requestUpdate = UpdatingElement.prototype.requestUpdate;
+  UpdatingElement.prototype.requestUpdate = function (
```
justinfagnani, line 1043: Rather than patching `requestUpdate`, can you put the `if (DEV_MODE) { return ...` in `requestUpdate`?

justinfagnani (changes requested):

packages/updating-element/src/updating-element.ts
sorvell: The array is more ergonomic for setting the warnings as a block (which seems reasonable to support), but the point of the API `enableWarning/disableWarning` is to be able to customize warnings per class, copying down your superclass' settings. The uniqueness check in the array is so that you can disable a warning (can't have duplicates for that).

This may be fancier than we need, but since we're going to the trouble of having dev mode and it's basically "free code," it seemed worth it.

packages/updating-element/src/updating-element.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
```diff
@@ -236,13 +277,19 @@ export interface Controller {
...
  * Base element class which manages element properties and attributes. When
  * properties change, the `update` method is asynchronously called. This method
  * should be supplied by subclassers to render updates as desired.
  * @noInheritDoc
  */
 export abstract class UpdatingElement extends HTMLElement {
+  // Note, these are patched in only in DEV_MODE.
+  static enabledWarnings?: Warnings[];
+  static enableWarning: (type: Warnings) => void;
+  static disableWarning: (type: Warnings) => void;
```
justinfagnani, line 292: I still think it'd be better if these APIs didn't exist at all in prod. We're paying for two fields in prod for here. Will continue discussion below.

packages/updating-element/src/updating-element.ts
justinfagnani: As noted above, we're always carrying the `enableWarning` and `disableWarning` APIs around, even in prod.  I don't think we need a fancy API here, especially for disabling warnings. With an optional array devs can either add to the list, replace the list, remove the whole list, or do fancy things if they care, but I doubt it'll be necessary.

The use cases I see:

- Add a warning:

```ts
UpdatingElement.enabledWarnings?.push('my-warning');
```

- Replace all warnings:

```ts
UpdatingElement.enabledWarnings = ['my-warning'];
```

- Remove all warnings:

```ts
UpdatingElement.enabledWarnings = undefined;
```


- Remove one warning (will this ever be needed?):

```ts
UpdatingElement.enabledWarnings = UpdatingElement.enabledWarnings.filter((w) => w !== 'my-warning');
```

justinfagnani (changes requested):

packages/updating-element/src/updating-element.ts
sorvell: They do not exist in prod.

packages/updating-element/src/updating-element.ts
kevinpschaaf: Nit: This change means we leave a a `return void 0` in the production minified code that wasn't strictly needed before right?

kevinpschaaf (approved):

justinfagnani (approved):

PR #1379, opened 2020-10-20
[lit-next] Adds controller support to UpdatingElement

sorvell:
This provides the foundation onto which element controllers that hook into element lifecycle can be built.

Note, this adds ~0.1K gzipped.


kevinpschaaf:
Seems really unfortunate to add infinity Sets unconditionally for every element to support an add-on feature like Controllers.

What about a `hooks.js`  module which exports APIs for adding callbacks, which patch in the code to call them on first use, to shift both the runtime and payload cost to pay-for-play? Along the lines of `parts.js` in lit-html (in terms off add-on APIs) and `*-support` (in terms of patching in functionality)?

packages/labs/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "dev": "scripts/dev.sh",
+    "build:ts": "tsc",
+    "build:ts:watch": "tsc --watch",
+    "test": "npm run test:dev && npm run test:prod",
+    "test:dev": "cd ../tests && npx wtr '../labs/development/**/*_test.js'",
+    "test:prod": "TEST_PROD_BUILD=true npm run test:dev",
+    "test:watch": "npm run test:dev -- --watch",
+    "checksize": "rollup -c --environment=CHECKSIZE",
+    "format": "prettier src/* --write",
+    "lint": "tslint --project ./",
```
abdonrd, line 41:
```suggestion
```

Now the format and lint scripts are on the root `package.json`.

sorvell: Done.

justinfagnani:
@sorvell can you set the base branch to the labs branch?

packages/updating-element/src/updating-element.ts
```diff
@@ -534,6 +537,26 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   private _reflectingProperty: PropertyKey | null = null;
 
+  /**
+   * Set of callbacks called in connectedCallback.
+   */
+  connectedCallbacks: Set<connectCallback> = new Set();
```
justinfagnani: I'm not sure sets really buy us anything here. If the pattern is that a controller accepts a host in its constructor, we're very unlikely to to get duplicate callbacks. If we do, that's on the controller.

packages/updating-element/src/updating-element.ts
```diff
@@ -534,6 +537,26 @@ export abstract class UpdatingElement extends HTMLElement {
...
+
+  /**
+   * Set of callbacks called in disconnectedCallback.
+   */
+  disconnectedCallbacks: Set<connectCallback> = new Set();
```
justinfagnani: Did you explore lazily creating the collections? Then at callback time you could do `this.connectedCallbacks?.forEach(...)`

packages/updating-element/src/updating-element.ts
```diff
@@ -534,6 +537,26 @@ export abstract class UpdatingElement extends HTMLElement {
...
+
+  /**
+   * Set of callbacks called before update.
+   */
+  updateCallbacks: Set<updateCallback> = new Set();
+
+  /**
+   * Set of callbacks called after updated.
+   */
+  updatedCallbacks: Set<updateCallback> = new Set();
```
justinfagnani: Rather than four sets of callbacks I'd be interested in seeing a single `addController()` API that tears off the callbacks at that time, similar to `customElements.define()`

sorvell: Using a single set now.

packages/updating-element/src/updating-element.ts
sorvell: The set is lazily created now.

packages/updating-element/src/updating-element.ts
sorvell: The set buys efficient removal and less code for doing so.

packages/updating-element/src/updating-element.ts
```diff
@@ -544,6 +556,17 @@ export abstract class UpdatingElement extends HTMLElement {
     this.requestUpdate();
   }
 
+  addCallbacks(callbacks: LifecycleCallbacks) {
+    if (this._callbacks === undefined) {
```
justinfagnani, line 560: One liner for lazy init and add:

```js
(this._callbacks ??= new Set()).add(callbacks);
```

packages/updating-element/src/updating-element.ts
```diff
@@ -595,6 +618,9 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   connectedCallback() {
     this.enableUpdating();
+    if (this._callbacks !== undefined) {
+      this._callbacks.forEach((c) => c.onConnected());
+    }
```
justinfagnani, line 623:
```suggestion
    this._callbacks?.forEach((c) => c.onConnected());
```

packages/updating-element/src/updating-element.ts
```diff
@@ -544,6 +556,17 @@ export abstract class UpdatingElement extends HTMLElement {
...
 
+  addCallbacks(callbacks: LifecycleCallbacks) {
+    if (this._callbacks === undefined) {
+      this._callbacks = new Set();
+    }
+    this._callbacks.add(callbacks);
+  }
+
+  removeCallbacks(callbacks: LifecycleCallbacks) {
+    this._callbacks!.delete(callbacks);
```
justinfagnani, line 567: This should use nullish coalescing in case the collection hasn't been created yet:

```suggestion
    this._callbacks?.delete(callbacks);
```

If this were an array, it'd be a bit longer, but not too much:

```suggestion
    this._callbacks = this._callbacks?.filter((c) => c !== callbacks);
```

Removal should be _exceedingly_ rare, so we should optimize for rendering then adding then removing. The code size is probably the only significant factor on removal. Iteration cost _might_ matter - I bet the common case is just one or two controllers.

The nice thing about the `addCallbacks()` approach is that the storage is abstracted over, so we can change this over time.

packages/updating-element/src/updating-element.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
sorvell: Done. Now that the Set is an implementation detail, it should be easy to revisit this.

justinfagnani (approved):

packages/updating-element/src/updating-element.ts
sorvell: Changed to array.

packages/updating-element/src/updating-element.ts
```diff
@@ -544,6 +558,10 @@ export abstract class UpdatingElement extends HTMLElement {
     this.requestUpdate();
   }
 
+  addController(controller: Controller) {
```
kevinpschaaf, line 561: I still like `addCallbacks` but won't insist.

kevinpschaaf (approved):
Code LGTM.

Noting the maybe very small regression on initial render (although lit-element goes the other way). Have you run this at high sample count? Either way, seems like we're open to a small regression on that order to add the feature?

Only nits/naming otherwise.

packages/updating-element/src/updating-element.ts
```diff
@@ -788,17 +815,14 @@ export abstract class UpdatingElement extends HTMLElement {
...
-        renderRoot: Element | DocumentFragment;
-      }).renderRoot = this.createRenderRoot();
-    }
     let shouldUpdate = false;
     const changedProperties = this._changedProperties;
     try {
       shouldUpdate = this.shouldUpdate(changedProperties);
       if (shouldUpdate) {
+        this._controllers?.forEach((c) => c.willUpdate?.(changedProperties));
+        this.willUpdate(changedProperties);
```
kevinpschaaf, line 824: We need to add calling `willUpdate` during SSR, if that's the intention. Either add to this PR, or file an issue?

packages/updating-element/src/updating-element.ts
sorvell: Going with `addController` as that's consensus.

packages/updating-element/src/updating-element.ts
sorvell: Tracked via https://github.com/Polymer/lit-html/issues/1417

packages/updating-element/src/test/updating-element-controllers_test.ts
```diff
@@ -0,0 +1,165 @@
...
+    assert.equal(el.controller.connectedCount, 1);
+    assert.equal(el.controller.disconnectedCount, 1);
+    container.appendChild(el);
+    assert.equal(el.connectedCount, 2);
+    assert.equal(el.disconnectedCount, 1);
+    assert.equal(el.controller.connectedCount, 2);
+    assert.equal(el.controller.disconnectedCount, 1);
+  });
+
+  test('controllers can implement willUpdate/update/updated', async () => {
```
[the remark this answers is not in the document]
sorvell: Done.

sorvell:
Re: perf, testing locally, I consistently get unsure even with a high sample rate.

PR #1388, opened 2020-10-23
[lit-ssr] Add defer-hydration protocol for coordinating host/child hydration

kevinpschaaf:
Goal: Ensure an SSR'ed LitElement does not enable and hydrate itself before its initial data has been supplied to it from its host (by virtue of its host being hydrated). Since in SSR, element upgrade occurs as a result of CE registration being loaded, and since module loading order naturally occurs leaves-up (by virtue of parents depending on their children), a child will typically run its `connectedCallback` (which in turn schedules its update) before its host when the application bundle is loaded.

Approach:
* During SSR opcode generation, inject a `<!--lit-node-->` marker to denote each custom element. Note, the previous `<!--lit-bindings-->` marker was repurposed (and renamed) to now indicate any node that either has attribute bindings _or_ is a custom element.
* During SSR rendering in the `custom-element-attributes` opcode, if the custom element _host_ stack is >1 deep (meaning the CE is within another CE), then a `defer-hydration` attribute is emitted on the CE. A new `customElementHostStack` (distinct from `customElementInstanceStack`) is introduced to track shadow-dom hosts only (the instance stack tracks every open CE, including light-dom parents)
* During hydration, when `<!--lit-node-->` is encountered, in addition to looking for bound attributes, we also look for the `defer-hydration` attribute after hydrating attribute/property/event/element parts, and if present we simply remove it
* The `hydration-support` module patches `observedAttributes`, `attributeChangedCallback`, and `connectedCallback`, such that `connectedCallback` does not run the `LitElement.prototype.connectedCallback` if it has a `defer-hydration` attribute at connected time; in that case, `attributeChangedCallback` detects when `defer-hydration` is removed and only then calls the base `connectedCallback`, causing the element to enable updating and ultimately hydrate itself.

This PR also fixes a latent bug with hydrating attribute parts on void elements: we need to check previousSibling of the `<!--lit-node>` marker for void elements first, and then use parentElement otherwise (it's guaranteed to be the first node of a parent otherwise).


packages/labs/ssr/src/lib/render-lit-html.ts
```diff
@@ -184,6 +184,20 @@ type CustomElementClosedOp = {
   type: 'custom-element-close';
 };
 
+/**
+ * Operation to possibly emit the `<!--lit-node-->` marker; the operation
+ * always emits if there were attribtue parts, and may emit if the node
+ * was a custom element and it needed `defer-hydration` because it was
+ * rendered in the shadow root of another custom element host; we don't
```
sorvell, line 191: Why don't we know this? It's unclear given that `customElementHostStack` is maintained.

packages/lit-element/src/hydrate-support.ts
```diff
@@ -30,31 +31,75 @@ interface PatchableLitElement extends HTMLElement {
 }: {
   LitElement: PatchableLitElement;
 }) => {
-  // Capture whether we need hydration or not
-  const createRenderRoot = LitElement.prototype.createRenderRoot;
-  LitElement.prototype.createRenderRoot = function (this: PatchableLitElement) {
+  const observedAttributes = Object.getOwnPropertyDescriptor(
```
sorvell, line 34: I'd like to briefly consider putting this code into `ReactiveElement` with perhaps a more generic attribute name.

If you boot up with a `shadowRoot`, we could set a `hasDeclarativeShadowRoot` property.

Perhaps we could just use `disabled` or `inert` (or prefix this with `lit-`) to do what `defer-hydration` is doing? It adds very little code and could be more generally useful. It could also be dynamic by using `shouldUpdate`?

In addition to the code, this would add a `hasAttribute` check to initialization.

sorvell: Via offline chat we decided this wasn't worth doing.

packages/lit-html/src/hydrate.ts
```diff
@@ -95,7 +95,7 @@ type ChildPartState =
  *                                 # by render(). Includes the digest of the
  *                                 # template
  *   <div class="TEST_X">
- *     <!--lit-bindings 0--> # Indicates there are attribute bindings here
+ *     <!--lit-node 0--> # Indicates there are attribute bindings here
```
sorvell, line 98: Let's rename this file `experimental-hydrate.js`

packages/labs/ssr/src/lib/render-lit-html.ts
sorvell: Answer: because this is template based and not instance based.

sorvell (dismissed):

sorvell (changes requested):
Basically seems ok, but let's rename the `hydrate` file to `experimental-hydrate`.

packages/lit-element/src/hydrate-support.ts
kevinpschaaf: Decided against, due to multi-actor issue; `defer-hydration` can only be used by ssr/hydration code.

packages/lit-element/src/hydrate-support.ts
```diff
@@ -30,31 +31,75 @@ interface PatchableLitElement extends HTMLElement {
...
+
+  // Override `connectedCallback` to capture whether we need hydration, and
+  // defer `enableUpdate()` if the 'defer-hydration' attribute is set
+  const connectedCallback = LitElement.prototype.connectedCallback;
+  LitElement.prototype.connectedCallback = function (
+    this: PatchableLitElement
+  ) {
     if (this.shadowRoot) {
       this._$needsHydration = true;
+    }
```
sorvell, line 69: Why does `_$needsHydration` need to be set here?

packages/lit-element/src/hydrate-support.ts
```diff
@@ -30,31 +31,75 @@ interface PatchableLitElement extends HTMLElement {
...
+    if (!this.hasAttribute('defer-hydration')) {
+      connectedCallback.call(this);
+    }
+  };
+
+  // If we've been server-side rendered, just return `this.shadowRoot`, don't
+  // call the base implementation, which would also adopt styles (for now)
+  const createRenderRoot = LitElement.prototype.createRenderRoot;
+  LitElement.prototype.createRenderRoot = function (this: PatchableLitElement) {
+    if (this._$needsHydration) {
```
sorvell, line 81: Should set `_$needsHydration` here.

packages/lit-html/src/hydrate.ts
kevinpschaaf: Done in https://github.com/Polymer/lit-html/pull/1743

packages/lit-element/src/hydrate-support.ts
kevinpschaaf: Removed

packages/lit-element/src/hydrate-support.ts
kevinpschaaf: Done, updated

sorvell (approved):

packages/lit-element/src/hydrate-support.ts
```diff
@@ -30,7 +31,49 @@ interface PatchableLitElement extends HTMLElement {
...
+    LitElement.prototype.attributeChangedCallback;
+  LitElement.prototype.attributeChangedCallback = function (
+    name: string,
+    old: string | null,
+    value: string | null
+  ) {
+    if (name === 'defer-hydration' && value === null) {
+      connectedCallback.call(this);
+    } else {
+      attributeChangedCallback.call(this, name, old, value);
```
justinfagnani, line 58: I think you should always call into the original `attributeChangedCallback`. What if an element wanted to do something special on hydration?

packages/lit-element/src/hydrate-support.ts
```diff
@@ -30,7 +31,49 @@ interface PatchableLitElement extends HTMLElement {
...
+    get() {
+      return [...observedAttributes.call(this), 'defer-hydration'];
+    },
+  });
+
+  // Enable element when 'defer-hydration' attribute is removed by calling the
+  // super.connectedCallback()
+  const attributeChangedCallback =
+    LitElement.prototype.attributeChangedCallback;
+  LitElement.prototype.attributeChangedCallback = function (
```
justinfagnani, line 50: I think it might be good to put as much of this logic onto ReactiveElement as possible for other subclasses to use...

justinfagnani (approved):

packages/lit-element/src/hydrate-support.ts
kevinpschaaf: Yeah good call. Updated.

packages/lit-element/src/hydrate-support.ts
kevinpschaaf: Let's do that in a later PR, since it involves adding a hydrate-support for reactive-element, which doesn't yet exist.

PR #1389, opened 2020-10-24
[lit-html] Add the `until` directive.

### LINKED ISSUE #1356: [lit-html] Port until() to lit-html 2.0
We're going to need until() to upgrade Google. The question here is the API around setting a part value for attribute parts. The directive will need an index/offset to represent the expression it's targeting in multi-value attribute bindings.

We'll need a helper like:

```
setPartValue(part, value, this.index)
```

Where index is passed to the directive's constructor.

bicknellr:
Fixes #1356. ~A few of the old tests are broken currently: some seem related to timing and others seem related to `EventPart`s.~

This PR also changes lit-html itself in a few ways:
- Directives that are associated with an `AttributePart` are now passed their index in the `AttributePart`'s values array.
- `setPartValue` now takes an `index` when setting an `AttributePart`'s value.


sorvell:
There's an unrelated test error which I think is due to flakiness caused by polyfilling promises on IE11. Suggest up'ing the timeout to something like 10ms [here](https://github.com/Polymer/lit-html/blob/lit-next/packages/updating-element/src/test/updating-element_test.ts#L2414).

packages/lit-html/src/directives/until.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  constructor(_part: PartInfo, index?: number) {
+    super();
+
+    this._index = index;
+    this._latestUpdateId = 0;
+  }
+
+  render(...values: Array<unknown>) {
+    return values.find(isPromiseLike) ?? nothing;
```
bicknellr, line 41:
```suggestion
    return values.find((x) => !isPromiseLike(x)) ?? nothing;
```

bicknellr:
I'm actually kind of confused that the other tests are passing, they don't pass locally for me right now. :/

bicknellr:
Oh, they're passing remotely because of the way I named the file where I put the old tests. _facepalm_

bicknellr:
Ok, that should be enough to break things. :)

bicknellr:
I merged `lit-next-ssr` and made it the base branch of this PR, but `lit-next-ssr` hasn't merged `lit-next` recently, so this PR's diff against it currently includes #1370 and possibly other changes.

kevinpschaaf (approved):
LGTM

kevinpschaaf:
Oh one more tiny thing, please update the README to remove `until` from the list of unimplemented directives.

bicknellr:
Ok, updated. I also found that it was listed as removed in the CHANGELOG so I removed that. I'll merge once the tests pass.

PR #1402, opened 2020-10-31
Remove a `test.only` in the lit-html tests.

bicknellr:
I haven't looked into the cause for the failures this exposes, but I believe they're all legitimately failing tests so we shouldn't (necessarily) wait to make them pass in this PR before merging.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1181,7 +1181,7 @@ export class AttributePart {
           (this._value as Array<unknown>)[i] = v;
         }
         attributeValue +=
-          (typeof v === 'string' ? v : String(v)) + strings[i + 1];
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
Sorry about this; I'm working on fixes for this, I think it's better to fix them than to have master red for some amount of time.

kevinpschaaf:
Ok, that was a simple merging mistake when reverting one of the code factorings late in the `lit-ssr` PR: https://github.com/Polymer/lit-html/commit/961e95c95ef3a2dd36fbe00d660ab927012d923e#diff-bead45bb43cda474d916704550a1b26322667aa42b814c48300b05be79b9a7ebR1181

Will push a fix to this branch.

kevinpschaaf (approved):
LGTM, with added fix

PR #1412, opened 2020-11-03
Release @lit/localize v0.5.0

aomarks:


packages/localize/package.json
```diff
@@ -16,7 +16,7 @@
   "bin": {
     "lit-localize": "bin/lit-localize.js"
   },
-  "main": "index.js",
+  "main": "lit-localize.js",
```
justinfagnani, line 19: Can we add `"type": "module"` to this? Might also be good to add exports, but that's a bit more involved and could wait.

aomarks: No not easily, it will cause the CLI to execute as a node ES module, which will require a bunch of refactoring.

I intend to split out all of the CLI/Node code into a separate package sometime soon anyway, because they shouldn't be bundled together.

(Side note: I've been trying to write Node programs as ES modules lately, but I always fail, because you can only do default exports from CommonJS modules, and TypeScript typings seem to always be wrong.)

justinfagnani: We really should be indicating to tools that the exported files are modules.

You can work around the TypeScript issues in a few ways. More recent Node versions allow for named CJS imports. you can also cast the default export as the type of the entire module object:

Something like:
```ts
import lib, * as libTypes from 'lib';
const {exportA, exportB} = (lib as libTypes);
```

And you can also put tools scripts in a folder with a package.json with just `"type": "commonjs"`.

aomarks: Done. I just did the nested `package.json` for now.

aomarks: Also switched from Karma to @web/test-runner, since Karma did not survive the module conversion (I got the config file to launch, but then it just hung forever and never did anything -- good riddance!).

justinfagnani: Awesome. Thank you!

justinfagnani (approved):

PR #1415, opened 2020-11-04
Generate separate Prettier and ESLint ignore files

abdonrd:
Another proposal about #1414


abdonrd:
After https://github.com/foray1010/ignore-sync/pull/301, now we can split the files!

abdonrd:
@aomarks do we need something else to merge this?

aomarks (approved):

aomarks:
> @aomarks what do you think of this proposal? Do we need something else to merge it?

I think we should fix https://github.com/foray1010/ignore-sync/issues/302, because the semantics are currently wrong. But we should be fine in practice for now, so we can merge.

PR #1420, opened 2020-11-10
[lit-next] Adds react wrapper

sorvell:
Adds `createComponent(react, tagName, options)` for wrapping an individual custom element in a web component.


packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+  // Use cached value if available.
+  let ComponentClass = componentMap.get(tagName);
+  if (ComponentClass !== undefined) {
+    return ComponentClass;
+  }
+
+  class ReactComponent extends Component<Props> {
+    element!: T;
+
+    ref = (element: T) => {
```
justinfagnani, line 102: should this be public?

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Renders creates the custom element with a `ref` prop which allows this
+     * component to reference the custom element.
+     *
+     * Note, element properties are updated in DidMount/DidUpdate.
+     *
+     */
+    render() {
+      return createElement(tagName, {ref: this.ref}, this.props.children);
```
justinfagnani, line 166: why not let the attributes flow through to React's diffing? - just filter out the properties and events?

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+> = new Map();
+
+/**
+ *  Creates a React component from a CustomElement.
+ */
+// TODO(sorvell): Was unable to remove type E here.
+export const createComponent = <T extends HTMLElement, E>(
+  React: typeof ReactModule,
+  tagName: string,
+  events?: E
```
justinfagnani, line 84: Events is very loosely typed this way - effectively `unknown`, so the generic doesn't buy you much. I'd type as at least `object` or `extends object`, or better an indexed type with a string value.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+  type Props = React.PropsWithoutRef<ElementProps> & ForwardedRef;
+
+  // Use cached value if available.
+  let ComponentClass = componentMap.get(tagName);
+  if (ComponentClass !== undefined) {
+    return ComponentClass;
+  }
+
+  class ReactComponent extends Component<Props> {
+    element!: T;
```
justinfagnani, line 100: should this be public?

packages/labs/src/frameworks/react/create-component.ts
justinfagnani, line 166: Why do we need to implement `componentDidMount`, `componentDidUpdate`? Can't we do everything in `render()` and `this.ref()`?

packages/labs/src/tests/frameworks/react/create-component_test.ts
```diff
@@ -0,0 +1,214 @@
...
+    document.body.appendChild(container);
+  });
+
+  teardown(() => {
+    if (container && container.parentNode) {
+      container.parentNode.removeChild(container);
+    }
+  });
+
+  const BasicElementComponent = createComponent<
```
justinfagnani, line 73: Can you rewrite this as:

```ts
  const basicEvents = {
    onFoo: 'foo',
    onBar: 'bar',
  };
  const BasicElementComponent = createComponent<BasicElement, typeof basicEvents>(
    window.React, elementName, basicEvents);
```

just to show in tests how you can avoid duplicating the events type?

justinfagnani (changes requested):

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+  node: E,
+  name: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value: any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  old: any,
+  events?: T
+) => {
+  // Dirty check and prevent setting reserved properties.
+  if (Object.is(value, old) || reservedReactProperties.has(name as string)) {
```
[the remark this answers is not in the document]
sorvell: As opposed to `===`? probably not, can change.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+      //     setProperty(this.element, prop, undefined, oldProps[prop], eventNames);
+      //   }
+      // }
+    }
+
+    /**
+     * Updates element properties correctly setting attributes v. properties
+     * on mount.
+     */
+    componentDidMount() {
```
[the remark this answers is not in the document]
sorvell: They are exclusive, that's correct (https://reactjs.org/docs/react-component.html#componentdidupdate).

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+    if (old !== undefined) {
+      node.removeEventListener(event, old);
+    }
+    node.addEventListener(event, value);
+  } else {
+    // For properties, use an `in` check
+    if (name in node) {
+      node[name as keyof E] = value;
+      // And for everything else, set the attribute.
+    } else if (value == null) {
```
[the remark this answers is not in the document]
sorvell: Yeah, will add comment.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+      node.removeAttribute(name);
+    } else {
+      node.setAttribute(name, value);
+    }
+  }
+};
+
+type ForwardedRef = {
```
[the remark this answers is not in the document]
sorvell: It's our own type, will add a comment.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+  events?: T
+) => {
+  // Dirty check and prevent setting reserved properties.
+  if (Object.is(value, old) || reservedReactProperties.has(name as string)) {
+    return;
+  }
+  // For events, use an explicit list.
+  const event = (events?.[
+    name as keyof T
+  ] as unknown) as keyof HTMLElementEventMap;
```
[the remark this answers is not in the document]
sorvell: Changed to `string`

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+          (forwardedRef as (e: T) => void)(element);
+        } else {
+          // This has to be `any` because `current` is readonly.
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (forwardedRef as any).current = element;
+        }
+      }
+    };
+
+    updateElement(oldProps?: Props) {
```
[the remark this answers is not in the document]
sorvell: Made private.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Made private.

packages/labs/src/frameworks/react/create-component.ts
sorvell: We can't filter out the properties here with a simple `in` check here, since we do not have the element, we'd have to use the element class prototype or lit specific `classProperities`.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Is there a way of typing `E` as requiring its members to be of type `string` without using an `index` signature? The idea is that we only want the fields passed via `E` to be acceptable `as props`.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Made private.

packages/labs/src/frameworks/react/create-component.ts
sorvell: `ref` is only fulfilled once, so at the least we'd also need `componentDidUpdate`.

packages/labs/src/tests/frameworks/react/create-component_test.ts
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
justinfagnani: The thing that's rough with this approach is that in SSR, these elements will have no attributes at all, because `componentDidMount` doesn't run.

We might need to take a config of properties and event names, and then if we want good ergonomics we might want a separate wrapper for when you have the component definition vs when you don't. ie, make a lazy wrapper that takes a config and an eager wrapper that takes the class. The eager wrapper, because it needs to import the definition, will just work. The lazy wrapper will need to take a loader function that performs a dynamic import as an alternative to `React.lazy`.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (Object.is(value, old) || reservedReactProperties.has(name as string)) {
+    return;
+  }
+  // For events, use an explicit list.
+  const event = (events?.[
+    name as keyof T
+  ] as unknown) as keyof HTMLElementEventMap;
+  if (event !== undefined) {
+    if (old !== undefined) {
+      node.removeEventListener(event, old);
```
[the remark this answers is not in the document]
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Now computes a list of properties and events that are handled specially and separates the user's `props` into a list that's passed to React (so that React's rules for attribute setting are used) and a list of properties and events that's handled specially by the wrapper.

packages/labs/src/frameworks/react/create-component.ts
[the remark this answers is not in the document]
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+    events.delete(event);
+    node.removeEventListener(event, handler);
+  }
+};
+
+const setProperty = <E extends Element, T>(
+  node: E,
+  name: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value: any,
```
justinfagnani, line 61: `any` -> `unknown`

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  old: any,
```
justinfagnani, line 63: `any` -> `unknown`

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+  events?: T
+) => {
+  // Dirty check and prevent setting reserved properties.
+  if (value === old || reservedReactProperties.has(name as string)) {
+    return;
+  }
+  // For events, use an explicit list.
+  const event = (events?.[name as keyof T] as unknown) as string;
```
justinfagnani, line 71: You shouldn't need the `as unknown` cast here. Events should be typed as `{[method: string]: string}`.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+  node: E,
+  name: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value: any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  old: any,
+  events?: T
+) => {
+  // Dirty check and prevent setting reserved properties.
+  if (value === old || reservedReactProperties.has(name as string)) {
```
justinfagnani, line 67: `name` is already a string, remove cast.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+    return;
+  }
+  // For events, use an explicit list.
+  const event = (events?.[name as keyof T] as unknown) as string;
+  if (event !== undefined) {
+    setEvent(node, event, value);
+  } else {
+    node[name as keyof E] = value;
```
justinfagnani, line 75: You can type `name` as `keyof E` in the function signature and push the cast out to the caller.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+type Events<S> = {
+  [P in keyof S]?: (e: Event) => unknown;
+};
+
+type Constructor<T> = {new (): T};
+
+/**
+ *  Creates a React component from a CustomElement.
+ */
+export const createComponent = <I extends HTMLElement, E>(
```
justinfagnani, line 88: `E extends {[prop: string]: string]}`

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+     * Standard attributes are passed to React and element properties and events
+     * are updated in componentDidMount/componentDidUpdate.
+     *
+     */
+    render() {
+      // Filters class properties out and passes the remaining
+      // attributes to React. This allows attributes to use framework rules
+      // for setting attributes and render correctly under SSR.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const props: any = {ref: this._ref};
```
justinfagnani, line 193: See if `props: object` or `props: {}` lets you remove the `any`.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+      // Note, save element props while iterating to avoid the need to
+      // iterate again when setting properties.
+      this._elementProps = {};
+      for (const p in this.props) {
```
justinfagnani, line 197: You might consider

```ts
for (const [k, v] of Object.entries(props)) {
```

justinfagnani (changes requested):

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+          this._element,
+          prop,
+          this.props[prop as keyof ComponentProps],
+          oldProps ? oldProps[prop as keyof ComponentProps] : undefined,
+          events
+        );
+      }
+      // Note, the spirit of React might be to "unset" any old values that
+      // are no longer included; however, there's no reasonable value to set
+      // them to so we just leave the previous state as is.
```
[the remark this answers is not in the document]
sorvell: I don't think so. I believe users write LitElement's with `@property({type: String}) foo = ''` and expect that value to never be `undefined`. They shouldn't but I think this is not uncommon.

packages/labs/src/frameworks/react/create-component.ts
kevinpschaaf: Isn't it `Partial<ComponentProps>`? And then can get rid of the `p as keyof ComponentProps` below.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+
+type Events<S> = {
+  [P in keyof S]?: (e: Event) => unknown;
+};
+
+type Constructor<T> = {new (): T};
+
+/**
+ *  Creates a React component from a CustomElement.
+ */
```
kevinpschaaf, line 87: Can you add docs for the arguments?

packages/labs/src/frameworks/react/create-component.ts
kevinpschaaf, line 67: Why the dirty check? Shouldn't that defer to the element's accessor dirty check (and potential `hasChanged` logic?)

kevinpschaaf (changes requested):

packages/labs/src/frameworks/react/create-component.ts
justinfagnani: I'm not sure I follow. Prop keys can be strings, so that's the right type, no?

packages/labs/src/frameworks/react/create-component.ts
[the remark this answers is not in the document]
justinfagnani: This behavior needs to be documented then. I'm still not sure setting `undefined` isn't the right behavior though. If a user of an elements sets and then "unsets" a property, isn't that user at least expecting something to happen on unset?

And to be clear, this can only happen with spread, right? If a prop is explicitly set to undefined it shows up in the props object?

packages/labs/src/frameworks/react/create-component.ts
justinfagnani: looks like `old` got changed to `unknown`, but not `value`

packages/labs/src/frameworks/react/create-component.ts
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Fixed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
sorvell: The `any` helps deal with the finicky React type. Using `ComponentProps` is problematic with the `ref` type. Using object or an index type doesn't work either.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Using an index type breaks the React type for Props. It allows anything to be passed as a prop.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
sorvell: That results in `Type 'unknown' is not assignable to type 'E[keyof E]'.`

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+      const forwardedRef = this.props?.__forwardedRef;
+      if (forwardedRef) {
+        // Note, there are 2 kinds of refs and we manually fulfill them here.
+        // There is no built in React API for this.
+        if (typeof forwardedRef === 'function') {
+          (forwardedRef as (e: I | null) => void)(element);
+        } else {
+          // This has to be `any` because `current` is readonly.
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (forwardedRef as any).current = element;
```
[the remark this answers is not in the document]
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+  // it's both needed in this component to get access to the rendered element
+  // and must fulfill any ref passed by the user.
+  type ComponentProps = UserProps & {
+    __forwardedRef?: React.Ref<unknown>;
+  };
+
+  // List of properties/events which should be specially handled by the wrapper
+  // and not handled directly by React.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const elementPropsMap: {[index: string]: any} = {...(events ?? {})};
```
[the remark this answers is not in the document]
sorvell: Changed to unknown. Using an object helps type the corresponding `_elementProps` property since it can be `typeof elementPropsMap`. Is that possible using a Set?

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    private _ref = (element: I | null) => {
+      this._element = element;
+      // Fulfill any ref we receive in our props.
+      // The `ref` prop is special and cannot be accessed without an error
+      // so use `__forwardedRef`.
+      const forwardedRef = this.props?.__forwardedRef;
+      if (forwardedRef) {
+        // Note, there are 2 kinds of refs and we manually fulfill them here.
+        // There is no built in React API for this.
```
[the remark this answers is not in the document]
sorvell: Good catch. Fixed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Changed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Using an index type breaks the React type for Props. It allows anything to be passed as a prop.

packages/labs/src/frameworks/react/create-component.ts
justinfagnani: Right, but that's a better case than an any and a linter exception since it spells out the intention.

You could also try to pass the merged props type here and make `<..., K extends keyof P>(..., name: K, value: P[K]` but that will likely push a cast to the caller. This is all very dynamic and internal at this point in the calls, so I don't know if it buys much, but in general it's at least better to avoid `any` by using `unknown` and a specific cast.

sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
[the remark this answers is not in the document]
sorvell: Fixed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Fixed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Fixed on the consuming property.

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,237 @@
...
+  'style',
+  'className',
+]);
+
+const listenedEvents: WeakMap<
+  Element,
+  Map<string, EventListenerObject>
+> = new WeakMap();
+
+const setEvent = (
```
justinfagnani, line 31: Add a comment on the purpose of this function

justinfagnani, line 31: Add a comment on the purpose of this function

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,237 @@
...
+
+const setProperty = <E extends Element, T>(
+  node: E,
+  name: string,
+  value: unknown,
+  _old: unknown,
+  events?: StringValued<T>
+) => {
+  if (reservedReactProperties.has(name)) {
+    return;
```
justinfagnani, line 65: Shouldn't reserved properties never reach this point? The need to be passed on to the underlying `createElement` call, so they shouldn't be filtered.

Also, should we warn at `createComponent()` time if the element class contains reserved properties?

packages/labs/src/frameworks/react/create-component.ts
```diff
@@ -0,0 +1,237 @@
...
+  // and must fulfill any ref passed by the user.
+  type ComponentProps = UserProps & {
+    __forwardedRef?: React.Ref<unknown>;
+  };
+
+  // Set of properties/events which should be specially handled by the wrapper
+  // and not handled directly by React.
+  const elementPropsMap = new Set(Object.keys(events ?? {}));
+  for (const p in elementClass.prototype) {
+    if (!(p in HTMLElement.prototype)) {
```
justinfagnani, line 141: check against reserved properties here

justinfagnani (changes requested):

packages/labs/src/frameworks/react/create-component.ts
sorvell: Done.

sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Fixed.

packages/labs/src/frameworks/react/create-component.ts
sorvell: Done.

packages/labs/src/frameworks/react/create-component.ts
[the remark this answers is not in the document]
justinfagnani: This is my remaining concern on this PR. Make a note of it in the docs on `createComponent`? I think it's only an issue with spread.

justinfagnani (approved):

packages/labs/src/tests/frameworks/react/create-component_test.ts
```diff
@@ -89,7 +89,7 @@ suite('React createComponent', () => {
     props?: ReactModule.ComponentProps<typeof BasicElementComponent>
   ) => {
     window.ReactDOM.render(
-      window.React.createElement(BasicElementComponent, props),
+      window.React.createElement(BasicElementComponent, props, props?.children),
```
kevinpschaaf, line 92: Is there any harm in duplicating `children` in `props` and in the 3rd arg?

kevinpschaaf (approved):

packages/labs/src/tests/frameworks/react/create-component_test.ts
sorvell: It looks like you're supposed to pass children separate from props but it looks harmless to include children in props, see this [example](https://babeljs.io/en/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oFMAeAXZA7AJjASsmMOgHQDC8AtgA7xbbrQDeAUNNAE7Y7KcAUAShbsOXZOgCunLNAA8OAJYA3AHzN0AC0URSNTvBq7g2kDm5YAvnID0StQG5Rl1i9ahIMABLIQIJGiYuAREJBTUdAxYTGwcFrwCwrFi3FIy8ghIJopmFgC8Gtq6-obGpubYlupaOnoGRqTZuZUKKqq29u02mapOHC5uhMToACIA8gCypPF8_KJyPn5Z4FAAcmBUyHkARABmiNAARmCc2-2d2h1ttov-qgA0ojjwwJKb0aQA5hIAoiDI73QACEAJ4ASRw_AA5AZ4OgoYJWIIHEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.12.10&externalPlugins=).

sorvell: It looks fine to remove so will do so.

PR #1424, opened 2020-11-10
Remove the inline append / remove polyfills from the platform-support module.

bicknellr:
webcomponentsjs includes these as of 2.5.0, which the lit-next branch is now using.

### DIFF
--- a/packages/lit-html/src/platform-support.ts
+++ b/packages/lit-html/src/platform-support.ts
@@ -32,52 +32,6 @@
  * @packageDocumentation
  */
 
-// TODO(sorvell): Remove these once ShadyDOM/webcomponentsjs supports them.
-// Source: https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append
-(function (arr) {
-  arr.forEach(function (item) {
-    if (item.hasOwnProperty('append') && !window.ShadyDOM?.inUse) {
-      return;
-    }
-    Object.defineProperty(item, 'append', {
-      configurable: true,
-      enumerable: true,
-      writable: true,
-      value: function append() {
-        // eslint-disable-next-line prefer-rest-params
-        const argArr = Array.prototype.slice.call(arguments),
-          docFrag = document.createDocumentFragment();
-
-        argArr.forEach(function (argItem) {
-          const isNode = argItem instanceof Node;
-          docFrag.appendChild(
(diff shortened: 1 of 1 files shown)


justinfagnani:
Does this actually bump wcjs to 2.5.0?

I tried to update package locks in #1419, and I think wcjs 2.5.0 is causing IE11 platform-support tests to not load.

bicknellr:
> Does this actually bump wcjs to 2.5.0?

No, that happened a couple weeks ago in #1381. I just happened to look at this file and noticed these inlined polyfills earlier today.

> I tried to update package locks in #1419, and I think wcjs 2.5.0 is causing IE11 platform-support tests to not load.

Hmm, yeah that sounds like it might be related. I've noticed recently that installing ends up with changes to the package locks - is that what #1419 is intending to fix?

justinfagnani:
Yeah, #1419 is trying to get things back to a locally reproducible build. So #1381 already bumped wcjs to 2.5.0 in `packages/tests`, but `lit-html` is still on 2.2.3.

bicknellr:
I've opened #1425 to bump those to ^2.5.0, but I'm not really sure if those packages should even depend on webcomponentsjs. IIRC, web-test-runner automatically injects it into all of the tests. Either way, that PR will tell us whether or not this is related to bumping those versions specifically or something else.

bicknellr:
#1425 passes, so that issue probably isn't related specifically to the bump to 2.5.0 . However, the platform-support tests are trying to pass some settings to the polyfills and I'm not sure how that interacts with web-test-runner's polyfill injection. I think they're injected before anything in the page, so I'm not sure if the settings in the platform-support tests will actually do anything because they seem like they're too late and the manually loaded polyfills bundle is probably a noop since it's effectively a second copy.

bicknellr:
Oh, the platform-support test files are HTML but the link from the root page when running the tests locally and manually attempts to import them like JS. Is that meant to work?

bicknellr:
> Is our test setup supposed to be able to run HTML test files?

Yes, according to https://modern-web.dev/docs/test-runner/writing-tests/html-tests/ , so I'm probably not loading them correctly on my machine.

bicknellr:
FWIW, #1427 (which I thought was completely unrelated) fails in the same way as #1419.

justinfagnani (approved):

PR #1426, opened 2020-11-10
[lit-html] Resolve/cleanup directive as first step in setValue

### LINKED ISSUE #1286: [lit-html] Swapping a directive for a non-directive doesn't clean up the directive instance
This caused a failing test when porting the `templateContent()` directive.

We may want to make a restriction that directive are not considered "structural" and cannot dynamically change that way.

kevinpschaaf:
Moves directive resolution to the first step in setValue, used by all part types, and also cleans up the directive if a non-directive was rendered in its place.

Fixes #1286


packages/lit-html/src/lit-html.ts
```diff
@@ -708,6 +706,41 @@ class Template {
...
+  value: unknown,
+  i: number
+): unknown;
+function resolveDirective(
+  part: NodePart | AttributePart,
+  value: unknown,
+  i?: number
+) {
+  let directive =
+    i != null
```
sorvell, line 721: `i !== undefined`

packages/lit-html/src/lit-html.ts
```diff
@@ -708,6 +706,41 @@ class Template {
...
+      ? (part as AttributePart)._directives?.[i]
+      : (part as NodePart)._directive;
+  const directiveClass = isPrimitive(value)
+    ? undefined
+    : (value as DirectiveResult)?._$litDirective$;
+  if (directive?.constructor !== directiveClass) {
```
sorvell, line 727: `directive` is the current one but `directiveClass` is the new one and that's confusing. Maybe rename `directive` to `currentDirective` and `directiveClass` to `nextDirectiveConstructor`?

packages/lit-html/src/lit-html.ts
```diff
@@ -708,6 +706,41 @@ class Template {
...
+    directive =
+      directiveClass !== undefined
+        ? new directiveClass(part as PartInfo)
+        : undefined;
+    if (i != null) {
```
sorvell, line 732: `i !== undefined`

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
```diff
@@ -711,6 +715,48 @@ class Template {
...
+    ? undefined
+    : (value as DirectiveResult)?._$litDirective$;
+  if (currentDirective?.constructor !== nextDirectiveConstructor) {
+    if (nextDirectiveConstructor !== undefined) {
+      currentDirective = new nextDirectiveConstructor(
+        part as PartInfo,
+        attributeIndex
+      );
+      currentDirective._part = part;
+      currentDirective._attributeIndex = attributeIndex;
```
kevinpschaaf, line 738: This is just a choice. We can do this in the constructor since we're passing them already, but it requires the user to super them through, which is sorta annoying (directives may need to set up their own state in the ctor, but normally they only need these arguments in limited cases).

packages/lit-html/src/parts.ts
```diff
@@ -78,13 +100,17 @@ export const setPartValue = <T extends Part>(
...
-    (part as AttributePart)._setValue(newValues, 0);
+    (part as AttributePart)._setValue(newValues, directiveParent, 0);
   } else {
-    part._setValue(value);
+    part._setValue(value, directiveParent);
   }
   return part;
 };
 
+export const updateDirectiveValue = (directive: Directive, value: unknown) => {
```
justinfagnani, line 110: Seems like this might as well be an instance method now. I'd consider moving the whole Directive class here now.

packages/lit-html/src/lit-html.ts
kevinpschaaf: I'm fine with that, will go ahead and update ll the directives to do this.

packages/lit-html/src/lit-html.ts
```diff
@@ -406,9 +407,12 @@ let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
  * change in future pre-releases.
  */
 export abstract class Directive {
+  _part!: NodePart | AttributePart;
+  _attributeIndex!: number | undefined;
+  _directive?: Directive;
   /** @internal */
-  _resolve(part: Part, props: Array<unknown>): unknown {
-    return this.update(part, props);
+  _resolve(part: Part, props: Array<unknown>, index?: number): unknown {
```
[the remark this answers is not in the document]
sorvell: or index and `this._attributeIndex`?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
```diff
@@ -711,6 +715,48 @@ class Template {
...
+  directiveParent: DirectiveParent = part,
+  attributeIndex?: number
+): unknown {
+  let currentDirective =
+    attributeIndex !== undefined
+      ? (directiveParent as AttributePart)._directives?.[attributeIndex]
+      : (directiveParent as NodePart | Directive)._directive;
+  const nextDirectiveConstructor = isPrimitive(value)
+    ? undefined
+    : (value as DirectiveResult)?._$litDirective$;
```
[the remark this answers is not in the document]
kevinpschaaf: `null` treated as a primitive in the `isPrimitive` check, so it's still not safe to dereference even when it passes. I added `isPrimitive` because it seemed to shave some time off the benchmarks, but I should double-check that.

packages/lit-html/src/lit-html.ts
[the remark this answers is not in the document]
kevinpschaaf: Done (arguments for part & index removed, pulled off of directive instance).

packages/lit-html/src/parts.ts
kevinpschaaf: Let's handle in a separate PR. Added to https://github.com/Polymer/lit-html/issues/1261.

kevinpschaaf:
@sorvell @justinfagnani PTAL

sorvell (approved):
LGTM

kevinpschaaf:
I could not reproduce the benchmark regressions locally, either on Mac or linux. Going to go ahead and merge this as-is.

packages/lit-html/src/lit-html.ts
```diff
@@ -406,9 +407,25 @@ let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
...
+  //@internal
+  _directive?: Directive;
+  constructor(part: PartInfo, attributeIndex: number | undefined) {
+    this._part = part as Part;
+    this._attributeIndex = attributeIndex;
+  }
   /** @internal */
-  _resolve(part: Part, props: Array<unknown>): unknown {
-    return this.update(part, props);
+  _resolve(props: Array<unknown>): unknown {
```
kevinpschaaf, line 421: Remove `_resolve`, can just overwrite `update` on server because we don't need the user `update`.

kevinpschaaf: Ah, I remember now: We rely on the base `AttributePart:_setValue` (with `noCommit` argument set to `true`) to resolve the individual directives and return the result on the server. Thus we don't have direct access to the individual directive instances to monkey patch `update` to `render`, and we can't change `Directive.prototype.update` because users override it.

kevinpschaaf: I did fix nested directives in SSR and added test for that.

packages/lit-html/src/parts.ts
```diff
@@ -78,15 +138,52 @@ export const setPartValue = <T extends Part>(
...
+    (part as AttributePart)._setValue(newValues, directiveParent, 0);
   } else {
-    part._setValue(value);
+    part._setValue(value, directiveParent);
   }
   return part;
 };
 
+/**
+ * Sets the value of a directive asynchronously, outside the normal
```
justinfagnani, line 149: When I read this I wonder "what's a directive value?" Does a directive have a single value? I know this means the value typically returned by the directive's `update()` method, but the value phrasing reads odd to me. I wonder if we should talk about the directive "result" instead?

justinfagnani (commented):
One question around the naming of `setDirectiveValue`

packages/lit-html/src/parts.ts
kevinpschaaf: Maybe the API should be `updateDirectiveResult`?

packages/lit-html/src/lit-html.ts
```diff
@@ -851,25 +901,21 @@ export class NodePart {
     return this._startNode.parentNode!;
   }
 
-  _setValue(value: unknown): void {
-    // TODO (justinfagnani): when setting a non-directive over a directive,
-    // we don't yet clear this._directive.
-    // See https://github.com/Polymer/lit-html/issues/1286
+  _setValue(value: unknown, directiveParent: DirectiveParent = this): void {
+    value = resolveDirective(this, value, directiveParent);
```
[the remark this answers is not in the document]
kevinpschaaf: The value before resolving the directive and the value after resolving the directive may be different, so in the worst case. you still need to do the `isPrimitive` check twice. If the pre-resolving `isPrimitive === false`, then it will stay false, but if it's true, it may become false as a result of the directive returning a primitive.

So it would be:
```ts
const valueIsPrimitive = isPrimitive(value);
value = resolveDirective(this, value, directiveParent, valueIsPrimitive);
if (valueIsPrimitive || isPrimitive(value)) {
  // ...
```
and 
```ts
const nextDirectiveConstructor = valueIsPrimitive || isPrimitive(value)
  ? undefined
  : ...
```
Worth it?

packages/lit-html/src/parts.ts
kevinpschaaf: In https://github.com/Polymer/lit-html/pull/1432 I'm moving this to an instance method on `DisconnectableDirective`, so maybe just `setPartValue` there would be good?

PR #1427, opened 2020-11-10
Import modules in tests using node-style module specifiers.

bicknellr:


bicknellr:
Huh, this is failing in the same way as #1419. Maybe a clue?

PR #1432, opened 2020-11-13
[lit-html] Add DisconnectableDirective implementation

### LINKED ISSUE #283: Disconnected handler for directives

kevinpschaaf:
This is an alternate implementation to https://github.com/Polymer/lit-html/pull/1242 for adding `disconnectCallback` to directives, which is more pay-for-play, both in runtime and payload costs, especially for the baseline case when no disconnectable directives are used.

User-facing changes:

- Adds `DisconnectableDirective`, which is a `Directive` subclass whose `disconnectedCallback` will be called when the part containing the directive is cleared as a result of re-rendering, or when the user calls `part.setDirectiveConnection(false)` on a part that was previously rendered containing the directive. If `part.setDirectiveConnection(true)` is subsequently called on container, the directive's `reconnectedCallback` will be called prior to its next `update`/`render` callbacks. When implementing `disconnectedCallback`, `reconnectedCallback` should also be implemented to be compatible with reconnection. Note that `LitElement` will disconnect directives upon element disconnection, and re-connect directives upon element re-connection.
- `render` now returns the `NodePart` that was created/updated by `render`.
- Update and render callbacks will only be called in `UpdatingElement`/`LitElement` when the element is connected to the document. If an element is disconnected while an update is pending, or if an update is requested while the element is disconnected, update callbacks will be called if/when the element is re-connected.
  - This is important for this feature, as re-renders will automatically re-connect directives, however it is also likely a performance optimization in some cases, and aligns with the behavior of the element prior to connection.

Implementation details:
- Parent reference is passed down the tree of `Part`s / `TemplateInstance`s (this is in the core)
- `DisconnectableDirective` climbs up that tree in its constructor and builds a sparse tree of branches with a need for disconnection by placing child `Part`/`TemplateInstance`s in a set keyed to parent (this is in an optional module)
- When a part is cleared, the sparse tree is then walked to disconnect directives on the way back down (this code is installed by the optional module to reduce core code impact)
- When a part is cleared, it removes itself from its parent in the tree (however descendants don't bother removing themselves from their parent by virtue of a `isClearingValue` flag being false as an optimization, since this would be wasteful).
- A user-facing `part.setConnected(bool)` API is provided to manually disconnect/reconnect directives (used by `LitElement` connected/disconnectedCallbacks). It uses the same sparse tree created by `DisconnectableDirective` as when clearing.

Open TODO's/questions:
- [x] Add `reconnectedCallback` tests to lit-html (the end-to-end for disconnect/reconnect is already tested in `lit-element` tests)
  - Done
- [x] Do we like `part.setDirectiveConnection(bool)` vs. `setDirectiveConnection(bool, container, options: {refNode})`? I started with the latter, but it was ~0.2kb more code and realized that e.g. directives could possibly want to do this on the part, so better as a part API than a container API?
  - [x] Done - Settled on `part.setConnected(bool)`
- [x] The current implementation installs `_setValueConnected` and `_setDirectiveConnected` onto `NodePart` and `AttributePart` instances during the parent climb, which is both a code and runtime optimization (instance fields for these methods are reserved using `undefined`). Are we ok with this? The entirety of the `disconnected-directive` module currently sits at 638b gzip. If so, need to give these reserved/version-safe names in the rollup config
  - [x] Done - Settled on `NodePart:_$setNodePartConnected` and `DisconnectableDirective:_$setDirectiveConnected`
- [x] The current implementation has no debouncing in `LitElement` for disconnect/connect thrashing. Should we add that? Maybe in a separate PR since this is getting big?
  - [x] Added issue https://github.com/Polymer/lit-html/issues/1457
- [x] ~The current implementation uses a `WeakMap` keyed to the parents to store the sparse tree of children needing disconnection to avoid extra instance fields; the `WeakMap` will have multiple-version issues, so we can either:~
  - a). ~Try using a `window[Symbol.for('$litDisconnectableChildrenSets')] ??= new WeakMap();` pattern for sharing these between lit-html instances, or~
  - b.) Move this back to an instance property on each parent (still created lazily)
    - [x] Done - Settled on `_$disconnectableChildren`
- [x] (noticed while writing this up) It looks `_setDirectiveConnected` should also take a `removeFromParent` argument so that when a single directive goes away, it potentially removes its part from the parent tree (currently they are only removed upon `NodePart` clear)
  - [x] Done - added tests

Fixes #283


packages/lit-html/src/lit-html.ts
```diff
@@ -557,9 +559,12 @@ const getTemplateHtml = (
...
   return [
-    // We don't technically need to close the SVG tag since the parser
-    // will handle it for us, but the SSR parser doesn't like that
-    html + strings[l] + (type === SVG_RESULT ? '</svg>' : ''),
+    // We don't technically need to close the SVG tag since the parser will
+    // handle it for us, but the SSR parser doesn't like that.
+    // Note that the html must end with a node after the final expression to
+    // ensure the last NodePart has an end node, hence adding a comment if the
+    // last string was empty.
+    html + (strings[l] || '<?>') + (type === SVG_RESULT ? '</svg>' : ''),
```
justinfagnani, line 567: Use `nodeMarker` instead of `'<?>'` to save a couple of bytes

packages/lit-html/src/lit-html.ts
```diff
@@ -706,6 +711,8 @@ class Template {
   }
 }
 
+type HasParent = NodePart | AttributePart | TemplateInstance;
```
justinfagnani, line 714: Add comment. Because it's not obvious, even with TypeScript's help, I'd highlight the common interface that's used.

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,218 @@
...
+ */
+const setConnected = (
+  child: DisconnectableChild,
+  isConnected: boolean,
+  shouldRemoveFromParent = false
+) => {
+  const children = disconnectableChildrenForParent.get(child);
+  if (children !== undefined) {
+    for (const child of children) {
+      if (child instanceof DisconnectableDirective) {
```
justinfagnani, line 53: The `instanceof` checks worry me here. If we somehow get multiple instances of this module, then this function could encounter `DisconnectableDirective` from the other instance and the check will fail. Same for the `instanceof NodePart`, etc checks.

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * TODO(kschaaf): Patches disconnection API onto the parent; we could also just
+ * install this on the prototype once, or bite the bullet and put it in core;
+ * the former would add a small runtime tax, and the latter would add a large
+ * code size tax.
+ */
+const installDisconnectAPI = (child: DisconnectableChild) => {
+  if (child instanceof AttributePart) {
+    child._setDirectiveConnected = _setDirectiveConnected;
```
justinfagnani, line 157: What happens if we get multiple copies of lit-html? I think installing of the prototype would be safer, because each prototype (for each copy of say, `NodePart`) would get the method installed once. Then if we avoid `instanceof` this should work across copies.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Heh that's what I started with, but `nodeMarker` will be interpreted as a `NodePart`, which is not what we want! We just need a static node, but not one with meaning.

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Yup, there's a few instances of multiple-version safety that I still need to work out; we can do the branding thing here, but  I wanted to decide on overall strategy here first before fixing them all; if we want to put all this code in the core, it's moot.

packages/lit-html/static.js
```diff
@@ -0,0 +1,15 @@
+import{html as t,svg as o}from"./lit-html.js";
```
sorvell, line 1: Should this be removed?

packages/lit-html/static.js.map
```diff
@@ -0,0 +1 @@
...tatic.js`.\n */\nexport const html = withStatic(coreHtml);\n\n/**\n * Interprets a template literal as an SVG template that can efficiently\n * render to and update a container.\n *\n * Includes static value support from `lit-html/static.js`.\n */\nexport const svg = withStatic(coreSvg);\n"],"names":["unsafeStatic","value","ct","stringsCache","Map","withStatic","coreTag","strings","values","l","length","staticValue","dynamicValue","staticStrings","dynamicValues","s","i","hasStatics","undefined","_$litStatic$","push","key","join","get","set","html","coreHtml","svg","coreSvg"],"mappings":";;;;;;;;;;;;;SA0BaA,EAAgBC,KAC3BC,GAAcD,IAKVE,EAAe,IAAIC,IAKZC,EAAcC,GAA6B,CACtDC,KACGC,WAEH,MAAMC,EAAID,EAAOE,OACjB,IAAIC,EACAC,EACJ,MAAMC,EAA+B,GAC/BC,EAAgC,GACtC,IAEIC,EAFAC,EAAI,EACJC,GAAa,EAGjB,KAAOD,EAAIP,GAAG,CAKZ,IAJAM,EAAIR,EAAQS,GAKVA,EAAIP,QAE6DS,KAD/DN,EAAeJ,EAAOQ,GACvBL,YAAeC,wBAA8BO,KAE9CJ,GAAKJ,EAAcJ,IAAUS,GAC7BC,GAAa,EAEfH,EAAcM,KAAKR,GACnBC,EAAcO,KAAKL,GACnBC,IAQF,GAJIA,IAAMP,GACRI,EAAcO,KAAKb,EAAQE,IAGzBQ,EAAY,CACd,MAAMI,EAAMR,EAAcS,KAAK,gBAEfJ,KADhBX,EAAUJ,EAAaoB,IAAIF,KAEzBlB,EAAaqB,IACXH,EACCd,EAAWM,GAGhBL,EAASM,EAEX,OAAOR,EAAQC,KAAYC,IAShBiB,EAAOpB,EAAWqB,GAQlBC,EAAMtB,EAAWuB"}
```
sorvell, line 1: Should this be removed?

sorvell (changes requested):

packages/lit-element/src/lit-element.ts
```diff
@@ -120,7 +122,17 @@ export class LitElement extends UpdatingElement {
...
+    this._nodePart = render(value, this.renderRoot, this._renderOptions);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this._nodePart?.setDirectiveConnection(true);
+  }
+
+  disconnectedCallback() {
+    super.connectedCallback();
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/static.js
kevinpschaaf: Done

packages/lit-html/static.js.map
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,234 @@
...
+        // Part or Directive case
+        setChildrenConnected(
+          value as DisconnectableParent,
+          isConnected,
+          valueOnly
+        );
+      }
+      removeFromParentIfEmpty(this);
+    } else {
+      setChildrenConnected(this, isConnected);
```
sorvell, line 123: What if the value is an array here? Does that not make sense?

packages/lit-html/src/lit-html.ts
```diff
@@ -1103,10 +1200,12 @@ export class AttributePart {
     element: HTMLElement,
     name: string,
     strings: ReadonlyArray<string>,
+    parent: DisconnectableParent | undefined,
```
sorvell, line 1203: Can this be optional?

packages/lit-ssr/src/lib/render-lit-html.ts
```diff
@@ -514,7 +514,8 @@ export function* renderTemplateResult(
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           (null as any) as HTMLElement,
           op.name,
-          statics
+          statics,
+          undefined
```
sorvell, line 518: If this argument is optional, can it be removed?

packages/updating-element/src/updating-element.ts
```diff
@@ -850,6 +862,10 @@ export abstract class UpdatingElement extends HTMLElement {
       // Ensure any previous update has resolved before updating.
       // This `await` also ensures that property changes are batched.
       await this._updatePromise;
+      // If we were disconnected, wait until re-connected to flush an update
+      while (this._pendingConnectionPromise) {
```
sorvell, line 866: Is there good test coverage for this while needing to be there?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -509,6 +511,15 @@ suite('lit-html', () => {
       );
       assert.equal(container.querySelector('span'), beforeRenderedNode);
     });
+
+    test('back-to-back expressions', () => {
```
sorvell, line 515: How is this test relevant? (probably just missed it)

sorvell (changes requested):

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,234 @@
...
+  Directive,
+  PartInfo,
+  NodePart,
+  NODE_PART,
+  DisconnectableParent,
+} from './lit-html.js';
+export {directive} from './lit-html.js';
+
+/**
+ * Recursively walks down the tree of Parts/TemplateInstances/Directives to set
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,234 @@
...
+ * subtree, only the top-most part need to remove itself from the parent.
+ *
+ * Note, this method will be patched onto NodePart instances and called from the
+ * core code when parts are cleared or the connection state is changed by the
+ * user.
+ */
+function setNodePartConnected(
+  this: NodePart,
+  isConnected: boolean,
+  valueOnly = false,
```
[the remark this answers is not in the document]
sorvell: OK, the key is that it's only acting on the value. Maybe change `valueOnly` to `shouldSetConnectedOnValue` ?

packages/lit-html/src/disconnectable-directive.ts
sorvell: It's covered since this will remove all the disconnect-able children.

packages/lit-element/src/lit-element.ts
```diff
@@ -117,7 +119,17 @@ export class LitElement extends UpdatingElement {
...
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this._nodePart?.setDirectiveConnection(true);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this._nodePart?.setDirectiveConnection(false);
```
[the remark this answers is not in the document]
sorvell: Let's rename this to `setConnected`

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,350 @@
...
+ */
+const setChildrenConnected = (
+  parent: DisconnectableParent,
+  isConnected: boolean,
+  shouldRemoveFromParent = false
+) => {
+  const children = parent._$disconnetableChildren;
+  if (children !== undefined) {
+    for (const child of children) {
+      if ((child as Directive)._$setDirectiveConnected) {
```
sorvell, line 148: Can this be `_$setConnected`?

packages/lit-html/src/test/lit-html_test.ts
kevinpschaaf: It tests a small fix added to `getTemplateHtml`

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Added comment, it's treated as a brand and disambiguates it from NodePart's `_$setNodePartConnected`

packages/lit-element/src/lit-element.ts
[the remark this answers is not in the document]
kevinpschaaf: Done: https://github.com/Polymer/lit-html/issues/1457
Done: renamed to `setConnected`

packages/lit-element/src/test/lit-element_test.ts
```diff
@@ -314,4 +319,177 @@ import {assert} from '@esm-bundle/chai';
...
+      };
+      attr = 'default';
+      prop = 'default';
+      render() {
+        return html`<div attr=${d('child-attr')} .prop=${d('child-prop')}>
+          ${d('child-node')}
+        </div>`;
+      }
+      get child() {
+        return this.shadowRoot!.firstElementChild as Child;
```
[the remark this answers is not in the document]
kevinpschaaf: Added comment, it's so can access .prop off of the div

packages/lit-element/src/test/lit-element_test.ts
```diff
@@ -314,4 +319,177 @@ import {assert} from '@esm-bundle/chai';
...
+      container.appendChild(host);
+      await nextFrame();
+      assertRendering(host);
+      container.removeChild(host);
+      log.length = 0;
+      host.requestUpdate();
+      host.child.requestUpdate();
+      container.appendChild(host);
+      assertRendering(host);
+      assert.deepEqual(log, [
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-element/src/test/lit-element_test.ts
```diff
@@ -314,4 +319,177 @@ import {assert} from '@esm-bundle/chai';
...
+        'render-host-attr',
+        'render-host-prop',
+        'render-host-node',
+        'render-child-attr',
+        'render-child-prop',
+        'render-child-node',
+      ]);
+    });
+
+    test('directives reconnect on reconnection', async () => {
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/CHANGELOG.md
```diff
@@ -17,6 +17,34 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+## Unreleased
+
+<!-- ### Changed -->
+
+### Added
+
+- `render` now returns the `NodePart` that was created/updated by `render`.
+
+- Added `DisconnectableDirective`, which is a `Directive` subclass whose
+  `disconnectedCallback` will be called when the part containing the directive
```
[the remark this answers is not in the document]
kevinpschaaf: Correct, clarified.

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,234 @@
...
+      }
+    }
+    if (shouldRemoveFromParent) {
+      // When a NodePart is being cleared, this method will be called with its
+      // value (a TemplateInstance or iterable of NodeParts) and
+      // shouldRemoveFromParent==true. In that case, we will delete the value
+      // (`parent` in this function) from the NodePart, and then the NodePart
+      // will call `removeFromParentIfEmpty` to determine whether to remove the
+      // NodePart from _its_ parent.
+      parent._$parent?._$disconnetableChildren!.delete(parent);
```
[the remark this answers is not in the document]
kevinpschaaf: Done, good suggestion

packages/lit-html/src/disconnectable-directive.ts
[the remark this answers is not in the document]
kevinpschaaf: Done: `isClearingValue`

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-ssr/src/lib/render-lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+ * Recursively walks down the tree of Parts/TemplateInstances/Directives to set
+ * the connected state of directives and run `disconnectedCallback`/
+ * `reconnectedCallback`s.
+ */
+const setChildrenConnected = (
+  parent: DisconnectableParent,
+  isConnected: boolean
+): boolean => {
+  const children = parent._$disconnetableChildren;
+  if (children !== undefined) {
```
justinfagnani, line 149: nit: good spot for an early return:

```ts
if (children === undefined) {
  return false;
}
```

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+
+/**
+ * Recursively walks down the tree of Parts/TemplateInstances/Directives to set
+ * the connected state of directives and run `disconnectedCallback`/
+ * `reconnectedCallback`s.
+ */
+const setChildrenConnected = (
+  parent: DisconnectableParent,
+  isConnected: boolean
+): boolean => {
```
justinfagnani, line 147: document the return value

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+  return false;
+};
+
+/**
+ * Removes the given child from its parent list of disconnectable children, and
+ * if the parent list becomes empty as a result, removes the parent from its
+ * parent, and so forth up the tree when that causes subsequent parent lists to
+ * become empty.
+ */
+const removeFromParent = (child: DisconnectableParent) => {
```
justinfagnani, line 179: Could you name this to indicate the type of removal? ie, that's it's removing from disconnectable children?

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+  let parent = child._$parent;
+  let children = parent?._$disconnetableChildren;
+  children?.delete(child);
+  while (children !== undefined && children.size === 0) {
+    child._$disconnetableChildren = undefined;
```
justinfagnani, line 184: why is this line necessary? or should this be `parent._$disconnetableChildren = undefined;`?

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+
+/**
+ * Removes the given child from its parent list of disconnectable children, and
+ * if the parent list becomes empty as a result, removes the parent from its
+ * parent, and so forth up the tree when that causes subsequent parent lists to
+ * become empty.
+ */
+const removeFromParent = (child: DisconnectableParent) => {
+  let parent = child._$parent;
+  let children = parent?._$disconnetableChildren;
```
justinfagnani, line 181: in what cases would `parent` be undefined?

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+  children?.delete(child);
+  while (children !== undefined && children.size === 0) {
+    child._$disconnetableChildren = undefined;
+    if (parent !== undefined) {
```
justinfagnani, line 185: do an early break:

```ts
if (parent === undefined) {
  break;
}
```

packages/lit-html/src/disconnectable-directive.ts
justinfagnani: If these are brands, then do they need a trailing `$` and be excluded from minification?

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+   * connection state of a tree (in which case `shouldRemoveFromParent` should
+   * be false).
+   *
+   * @param isConnected
+   * @param shouldRemoveFromParent
+   * @internal
+   */
+  _$setDirectiveConnected(isConnected: boolean, shouldRemoveFromParent = true) {
+    this._setConnected(isConnected);
+    setChildrenConnected(this, isConnected);
```
justinfagnani, line 308: If you move `setChildrenConnected(this, isConnected);` into the condition below, you can simplify `setChildrenConnected`

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+  const children = parent._$disconnetableChildren;
+  if (children !== undefined) {
+    for (const child of children) {
+      // The existence of `_$setDirectiveConnected` is used as a "brand" to
+      // disambiguate DisconnectableDirectives from other DisconnectableChildren
+      // (as opposed to using an instanceof check to know when to call it); the
+      // redundancy of "Directive" in the API name is to avoid conflicting with
+      // `_$setNodePartConnected`, which exists `NodeParts` which are also in
+      // this list
+      if ((child as Directive)._$setDirectiveConnected) {
```
justinfagnani, line 157: If `_$setDirectiveConnected` changes a bit, this can be:

```ts
    for (const child of children) {
      (child as DisconnectableDirective)?._$setDirectiveConnected(
        isConnected,
        false
      );
      setChildrenConnected(child, isConnected);
    }
```

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+  }
+};
+
+/**
+ * Sets the connected state on any directives contained within the committed
+ * value of this part (i.e. within a TemplateInstance or iterable of NodeParts)
+ * and runs their `disconnectedCallback`/`reconnectedCallback`s, as well as
+ * within any directives stored on the NodePart (when `valueOnly` is false).
+ *
+ * `isClearingValue` should be passed as `true` on a top-level part that is
```
justinfagnani, line 202: Seems like this parameter is similar to `shouldRemoveFromParent` in `_$setDirectiveConnected` in that both are true at the top-level and false otherwise. Could they be named similarly?

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,384 @@
...
+        for (let i = fromPartIndex; i < value.length; i++) {
+          setChildrenConnected(value[i], isConnected);
+          removeFromParent(value[i]);
+        }
+      } else if (value != null) {
+        // TemplateInstance case: If the value has disconnectable children (will
+        // only be in the case that it is a TemplateInstance), we disconnect it
+        // and remove it from this NodePart's disconnectable children
+        setChildrenConnected(value as DisconnectableParent, isConnected);
+        removeFromParent(value as DisconnectableParent);
```
justinfagnani, line 240: should `removeFromParent` be called when `isConnected` is true?

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: `removeFromParent` -> `removeDisconnectable`

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: It should have been `parent._$disconnetableChildren`. I just removed it for now; it's not really necessary to remove the Set once it's empty, and allows switching to `do`/`while`.

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Top-level node. This got cleaner when changing to a `do`/`while`

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done.
In `_$setNodePartConnected` it's `isClearingValue` and 
in `_$setDirectiveConnected` it's now `isClearingDirective`.

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Ah, so if `isClearingValue` is true, `isConnected` is always assumed to be false. To make this clearer, I changed it to just pass `false` for `isConnected` when inside the `isClearingValue` case.

It's basically a tri-state of valid values:
* `isClearingValue === true` && `isConnected === false` - value is clearing
* `isClearingValue === false` && `isConnected === false` - part is being manually disconnected
* `isClearingValue === false` && `isConnected === true` - part is being manually reconnected

It sort of points out that these could really just be two separate functions. But I sorta only wanted to patch one method onto `NodePart`... 🤷‍♂️ WDYT?

kevinpschaaf:
@justinfagnani @sorvell I think I got everything we discussed; PTAL.

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,382 @@
...
+const removeDisconnectableFromParent = (obj: Disconnectable) => {
+  let parent, children;
+  do {
+    if ((parent = obj._$parent) == undefined) {
+      break;
+    }
+    children = parent._$disconnetableChildren!;
+    children.delete(obj);
+    obj = parent;
+  } while (children !== undefined && children.size === 0);
```
justinfagnani, line 187:
```suggestion
  } while (children?.size === 0);
```

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Removes the given child from its parent list of disconnectable children, and
+ * if the parent list becomes empty as a result, removes the parent from its
+ * parent, and so forth up the tree when that causes subsequent parent lists to
+ * become empty.
+ */
+const removeDisconnectableFromParent = (obj: Disconnectable) => {
+  let parent, children;
+  do {
+    if ((parent = obj._$parent) == undefined) {
```
justinfagnani, line 181: use `===` here, since this can't be `null`

packages/lit-html/src/disconnectable-directive.ts
```diff
@@ -0,0 +1,382 @@
...
+    setChildrenConnected(this, isConnected);
+  }
+}
+
+/**
+ * Patches disconnection API onto NodeParts.
+ */
+const installDisconnectAPI = (obj: Disconnectable) => {
+  if ((obj as NodePart).type == NODE_PART) {
+    (obj as NodePart)._$setNodePartConnected ??= setNodePartConnected;
```
justinfagnani, line 248: modern JS is so nice :)

packages/lit-html/src/disconnectable-directive.ts
justinfagnani: sounds fine as-is

packages/updating-element/src/updating-element.ts
kevinpschaaf: Added here: https://github.com/Polymer/lit-html/pull/1432/files/a0c5122f469acf219e0125549514431c1483d603#diff-afa936faa2534e921592fa2c0672a14d38d5b9963b8e2c752e46fcb6e700effaR2847

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: The thought was that since these are all known objects (`NodePart`, `TemplateInstance`, or `Directive`), they don't need collision-proof brands.

But we need to determine if minifying user `Directive` subclasses as part of application builds could result in collisions with our own minified symbols or not. If they could, it's an issue not of branding, but of overall collision-safety.

I added this issue to track: https://github.com/Polymer/lit-html/issues/1461

sorvell (approved):
LGTM

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

packages/lit-html/src/disconnectable-directive.ts
kevinpschaaf: Done

justinfagnani (approved):
nice!!

PR #1433, opened 2020-11-13
Fix styleMap example

tikotus:
Example function has a body, hence it doesn't have an implicit return value. An explicit return value is needed for the example to make sense.


tikotus:
@googlebot I signed it!

arthurevans (commented):
LGTM. Thanks for the fix!

arthurevans (approved):

PR #1435, opened 2020-11-16
[reactive-element] Adds UpdatingController and refines UpdatingElement API

sorvell:
* Adds UpdatingController
* Refactors `UpdatingMixin` out of `UpdatingElement` and applies it to `UpdatingController` to add support for reactive properties
* Refines `UpdatingElement` lifecycle:
  * didUpdate added as lifecycle entry point which calls `firstUpdated`/`updated`
  * controller `updated` changed to `didUpdate`
  * controller `willUpdate`, `update`, and `didUpdate` called inside `UpdatingElement`'s methods facilitating element code going before or after controller code.
  * element changed properties no longer passed to controller methods, but controller changed properties are
* Adds additional tests to controllers for derived properties and subscribing to a service


packages/updating-element/src/updating-element.ts
```diff
@@ -184,19 +170,10 @@ export interface PropertyDeclarations {
   readonly [key: string]: PropertyDeclaration;
 }
 
-type PropertyDeclarationMap = Map<PropertyKey, PropertyDeclaration>;
+// type PropertyDeclarationMap = Map<PropertyKey, PropertyDeclaration>;
```
kevinpschaaf, line 173: Remove?

packages/updating-element/src/updating-element.ts
```diff
@@ -352,173 +294,38 @@ export abstract class UpdatingElement extends HTMLElement {
...
-      const propKeys = [
-        ...Object.getOwnPropertyNames(props),
-        ...Object.getOwnPropertySymbols(props),
-      ];
-      // This for/of is ok because propKeys is an array
-      for (const p of propKeys) {
-        // note, use of `any` is due to TypeScript lack of support for symbol in
-        // index types
+  static finalize() {
+    const didFinalize = super.finalize();
```
kevinpschaaf, line 304: Nit: The meaning here is `didFinalizeThisCall` (I read with a "hasFinalized" meaning at first), so maybe `shouldFinalize` would be a clearer naming.

packages/updating-element/src/updating-element.ts
```diff
@@ -930,37 +712,77 @@ export abstract class UpdatingElement extends HTMLElement {
...
   // @internal
-  _didUpdate(changedProperties: PropertyValues) {
+  didUpdate(changedProperties: PropertyValues) {
+    this._controllers?.forEach((c) => c.didUpdate?.());
     if (!this.hasUpdated) {
       this.hasUpdated = true;
       this.firstUpdated(changedProperties);
     }
-    this._controllers?.forEach((c) => c.updated?.(changedProperties));
     this.updated(changedProperties);
```
kevinpschaaf, line 786: Will we eventually deprecate this in favor of `didUpdate`? Add a DEV_MODE warning to use `willUpdate` as `updated` may go away in a future version?

packages/labs/src/controllers/updating-controller.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.host.requestUpdate();
+  }
+
+  requestUpdate(
+    name?: PropertyKey,
+    oldValue?: unknown,
+    options?: PropertyDeclaration
+  ) {
+    return name !== undefined
+      ? super.requestUpdate(name, oldValue, options)
```
kevinpschaaf, line 97: The `super.reqeustUpdate` in `UpdatingMixin` doesn't trigger anything, it just records changed properties, right? Shouldn't we still call `host.requestUpdate` to actually trigger update lifecycle in this case?

packages/updating-element/src/updating-element.ts
```diff
@@ -789,56 +589,38 @@ export abstract class UpdatingElement extends HTMLElement {
...
-          this._reflectingProperties.set(name, options);
-        }
-      } else {
-        // Abort the request if the property should not be considered changed.
-        shouldRequestUpdate = false;
-      }
-    }
-    if (!this.isUpdatePending && shouldRequestUpdate) {
+    const needsUpdate =
+      super.requestUpdate(name, oldValue, options) && !this.isUpdatePending;
```
kevinpschaaf, line 617: What is the significance of the `isUpdatePending` flag being implemented/checked here in UpdatingElement and not in UpdatingMixin?

packages/updating-element/src/updating-element.ts
```diff
@@ -789,56 +589,38 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    if (needsUpdate) {
       this._updatePromise = this._enqueueUpdate();
     }
     // Note, since this no longer returns a promise, in dev mode we return a
     // thenable which warns if it's called.
-    return DEV_MODE ? requestUpdateThenable : undefined;
+    return DEV_MODE ? requestUpdateThenable : needsUpdate;
```
kevinpschaaf, line 623: Ugh, changing `requestUpdate` to return a boolean sort of messes up the DEV_MODE warning gambit; it's no longer safe to override `requestUpdate`, since in `DEV_MODE` you'll always get a sync-truthy thing.

packages/labs/src/controllers/updating-controller.ts
kevinpschaaf: Oh, the `_propertyChange` override was pretty confusing to me. Can't you get rid of `_propertyChanged` as an override point (always collect changedProperties in the core impl) and make this just do `if (super.reqeustUpdate(...)) { this.host.requestUpdate(); }`?

packages/updating-element/src/updating-element.ts
kevinpschaaf: Hmm, that would sort of leave `firstUpdated` as the only non-aligned name. I forget what we discussed about where to land with aligning the naming + where super is required.

kevinpschaaf (changes requested):

packages/labs/src/controllers/updating-controller.ts
kevinpschaaf: I guess you still need to override `_propertyChange` in `UpdatingElement`, but it seems like `UpdatingController` could be simplified as above.

rollup-common.js
```diff
@@ -48,6 +48,8 @@ const reservedProperties = [
   '_value',
   '_setValue',
   'createTreeWalker',
+  // TODO(sorvell): The properties below are only reserved so dev and prod can
+  // work together x-package.
```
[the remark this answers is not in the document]
sorvell: Done.

packages/labs/src/controllers/updating-controller.ts
sorvell: Done.

packages/updating-element/src/updating-element.ts
sorvell: `UpdatingMixin` has the updating lifecycle interface but it doesn't implement it so it doesn't have any notion of `isUpdatePending`. So this means trigger an update if the change suggested we should and one is not already pending.

packages/updating-element/src/updating-element.ts
sorvell: Hmm, good point. I can add `_scheduleUpdate` as an override point rather than using the return value.

sorvell:
Closing due to age.

PR #1442, opened 2020-11-21
[localize] New msg signature with options object

aomarks:
Changes the lit-localize `msg` signature, the first step in adding support for automatic localize message IDs (https://github.com/Polymer/lit-html/issues/1437).

 Before:

  ```ts
  msg(id: string, template: string|TemplateResult|Function, ...args: any[])
  ```

  After:

  ```ts
  msg(template: string|TemplateResult|Function, options: {id: string: args?: any[]})
  ```

In the next PR, `id` will become optional.

Also fixes the json-schema build issue.

### DIFF
--- a/packages/localize/src/program-analysis.ts
+++ b/packages/localize/src/program-analysis.ts
@@ -14,6 +14,10 @@ import * as parse5 from 'parse5';
 import {ProgramMessage, Placeholder, Message} from './messages';
 import {createDiagnostic} from './typescript';
 
+type ResultOrError<R, E> =
+  | {result: R; error?: undefined}
+  | {result?: undefined; error: E};
+
 /**
  * Extract translation messages from all files in a TypeScript program.
  */
@@ -63,13 +67,11 @@ function extractMessagesFromNode(
     descStack = dedupeMsgDescs([...descStack, ...newDescs]);
   }
 
-  const msg = extractMsg(file, program, node, descStack);
-  if (msg !== undefined) {
-    if ('contents' in msg) {
-      messages.push(msg);
-    } else {
-      errors.push(msg);
-    }
+  const extractResult = extractMsg(file, program, node, descStack);
+  if (extractResult.error) {
+    errors.push(extractResult.error);
+  } else if (extractResult.result !== undefined) {
+    messages.push(extractResult.result);
   }
 
   ts.forEachChild(node, (node) => {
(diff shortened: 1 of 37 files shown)


packages/localize/src/program-analysis.ts
```diff
@@ -86,112 +88,250 @@ function extractMsg(
...
+> {
+  if (node === undefined) {
+    return {result: {}};
+  }
+  if (!ts.isObjectLiteralExpression(node)) {
+    return {
+      error: createDiagnostic(
+        file,
+        node,
+        `Expected second argument to msg() to be an object`
```
justinfagnani, line 155: "object" -> "object literal"

packages/localize/src/program-analysis.ts
```diff
@@ -86,112 +88,250 @@ function extractMsg(
...
+    if (ts.isIdentifier(property.name)) {
+      name = property.name.escapedText;
+    } else if (ts.isStringLiteral(property.name)) {
+      name = property.name.text;
+    } else {
+      return {
+        error: createDiagnostic(
+          file,
+          property.name,
+          `Options shorthand object assignment must be identifier or string literal`
```
justinfagnani, line 184: Is this message correct? Here would would have a computed property name I think.

packages/localize/src/program-analysis.ts
```diff
@@ -86,112 +88,250 @@ function extractMsg(
...
     };
   }
 
-  if (isLitTemplate(contentsArg)) {
-    if (ts.isNoSubstitutionTemplateLiteral(contentsArg.template)) {
-      // E.g. msg('foo', html`bar <b>baz</b>`)
+  let id: string | undefined = undefined;
+  let args: ts.NodeArray<ts.Expression> | undefined = undefined;
+
+  for (const property of node.properties) {
```
justinfagnani, line 163: Does `.properties` include object spread?

packages/localize/src/program-analysis.ts
```diff
@@ -86,112 +88,250 @@ function extractMsg(
...
+          ),
+        };
+      }
+      args = property.initializer.elements;
+    } else {
+      return {
+        error: createDiagnostic(
+          file,
+          property,
+          `Options shorthand object property must be "id" or "args"`
```
justinfagnani, line 220: Is "shorthand" right here?

packages/localize/src/program-analysis.ts
```diff
@@ -86,112 +88,250 @@ function extractMsg(
...
+}
+
+/**
+ * Analyze the template argument to a msg call.
+ */
+export function extractTemplate(
+  templateArg: ts.Node,
+  file: ts.SourceFile
+): ResultOrError<
+  Pick<ProgramMessage, 'contents' | 'params' | 'isLitTemplate'> & {
```
justinfagnani, line 236: nit: I think it'd be nice to give this type a name

justinfagnani (approved):

packages/localize/src/program-analysis.ts
aomarks: Done.

packages/localize/src/program-analysis.ts
aomarks: Yep. Here are all the forms, I added this as a comment too:

```ts
const x = {
  a: 0,     // properties > PropertyAssignment > Identifier
  'b': 0,   // properties > PropertyAssignment > StringLiteral
  ['c']: 0, // properties > PropertyAssignment > ComputedPropertyName
  d,        // properties > ShorthandPropertyAssignment
  ...e,     // properties > SpreadAssignment
};
```

https://astexplorer.net/#/gist/4c8b4f5d6e957befe180b429d58fb2ef/1661e78f3669a9fd66c7aafb35040d38a1191cb1

packages/localize/src/program-analysis.ts
aomarks: Oops yeah, fixed. I updated both errors to be more clear too.

packages/localize/src/program-analysis.ts
aomarks: Nope, copy/paste error I think. Fixed.

packages/localize/src/program-analysis.ts
aomarks: Done (`ExtractedTemplate`).

PR #1445, opened 2020-11-24
[localize] Fix and document msgdesc comments

aomarks:
- Fix an XML serialization bug that broke `// msgdesc:` comments.
- Add some test cases.
- Document them in the README.
- Fix incorrect README transform vs runtime table.

Fixes https://github.com/Polymer/lit-html/issues/1444


packages/localize/src/formatters/xliff.ts
```diff
@@ -250,9 +250,10 @@ export class XliffFormatter implements Formatter {
 
       if (descStack.length > 0) {
         // https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#note
-        const note = document.createElement('note');
+        const note = doc.createElement('note');
```
justinfagnani, line 253: ouch!

Can you remove the `"DOM"` lib from the tsconfig? Hopefully TS would catch this.

aomarks: Done. (But DOM types still get loaded because the `xmldob` package I'm loading has a reference to them)

justinfagnani: Filed an issue here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/49953

justinfagnani (approved):

PR #1450, opened 2020-12-01
[lit-next] Fix .gitignores to ensure src files/folders aren't ignored

kevinpschaaf:
Add leading slashes to ensure src folders aren't accidentally ignored.

For example, `lit-element.*` will ignore `lit-element.js`, but also `src/lit-element.ts`; the intention of the changes were to only ignore the generated files at the root, which requires a leading slash.


rictic (approved):
+1, this is definitely a good practice

aomarks:
Also make sure to run "npm run ignore-sync" to sync the prettier and eslint ignore files.

abdonrd (approved):
Great, thanks Kevin!

kevinpschaaf:
@aomarks It seems to not have affected the ignore file generation, since it looks like `ignore-sync` just appends the folder name to nested .gitignore patterns (i.e. both `lit-element.*` and `/lit-element.*` in `packages/lit-element/.gitignore` result in a top-level `packages/lit-element/lit-element.*`.

Seems like there's a loss in semantic difference with having the leading `/` there, but it worked out for us in this case.

To be correct, it seems like `ignore-sync` should turn nested file patterns not starting with a leading `/` into e.g. `packages/lit-element/**/lit-element.*`... 🤷‍♂️

aomarks:
> @aomarks It seems to not have affected the ignore file generation, since it looks like `ignore-sync` just appends the folder name to nested .gitignore patterns (i.e. both `lit-element.*` and `/lit-element.*` in `packages/lit-element/.gitignore` result in a top-level `packages/lit-element/lit-element.*`.
> 
> Seems like there's a loss in semantic difference with having the leading `/` there, but it worked out for us in this case.
> 
> To be correct, it seems like `ignore-sync` should turn nested file patterns not starting with a leading `/` into e.g. `packages/lit-element/**/lit-element.*`... 🤷‍♂️

Oh yeah, I filed an issue about that, and it was fixed in 3.0.1, so we just need to update the top-level package-lock I think. https://github.com/foray1010/ignore-sync/issues/302

PR #1452, opened 2020-12-03
[lit-next] Refactor reserved properties

kevinpschaaf:
Cleans our rollup/terser reserved property situation a bit:
* Renames `crossPackagePropertyMangles` to `stableProperties`, and adds some better descriptions and validation to the list
* Implements a convention that all `stableProperties` are prefixed with `_$`, so it's clear in the source code when a property requires version stability and needs to be treated carefully
* Adds `$private` helper API for all properties that were reserved only for use by lit-ssr; the only `reservedProperties` are now the ones used for branding `TemplateResult` and `Directive`
  * This adde a `$private` export and `private-ssr-support.ts` module to `lit-element` as well, to expose `attributeToProperty`; it's a bit marginal, but follows the same pattern as the other packages that need it
* Adds `resetPartValue` helper for use by directives that were accessing the private `_value` property


rollup-common.js
```diff
@@ -40,54 +40,75 @@ const skipBundleOutput = {
...
+// they are easily identifiable as properties requiring version stability and
+// thus special attention.
+//
+// Mangled names are uppercase letters, in case we ever might want to use
+// lowercase letters for short, public APIs. Keep this list in order by mangled
+// name to avoid accidental re-assignments. When adding a name, add to the end
+// and choose the next letter.
+//
+// ONCE A MANGLED NAME HAS BEEN ASSIGNED TO A PROPERTY, IT MUST NEVER BE USED
+// FOR A DIFFERENT PROPERTY IN SUBSEQUENT VERSIONS.
```
kevinpschaaf, line 64: I'm obv breaking this rule here once to get the list in order; after this we should follow the rule and not have gaps in the list, since it was getting pretty confusing.

packages/lit-element/src/lit-element.ts
```diff
@@ -170,3 +170,13 @@ if (DEV_MODE) {
     return true;
   };
 }
+
+export const $private = {
```
sorvell, line 174: Few questions:

* Why is this here and not in UpdatingElement since `_$attributeToProperty` is defined there? Is it because we only support SSR for LitElement?
* What does the leading `$` indicate here? 
* If this is for SSR, should it be named as such?

packages/lit-element/src/private-ssr-support.ts
```diff
@@ -0,0 +1,19 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {$private as p} from './lit-element.js';
```
sorvell, line 15: I realize you tried to explain this in the PR comments, but I think this module needs some comments or we'll forget why it's being done this circuitous way.

packages/lit-html/src/directives/live.ts
```diff
@@ -25,10 +25,7 @@ import {
   BOOLEAN_ATTRIBUTE_PART,
   ATTRIBUTE_PART,
 } from '../lit-html.js';
-
-// A sentinal value that can never appear as a part value except when set by
-// live(). Used to force a dirty-check to fail and cause a re-render.
-const RESET_VALUE = {};
+import {resetPartValue} from '../parts.js';
```
sorvell, line 28: What does this have to do with reserved properties?

packages/lit-html/src/parts.ts
```diff
@@ -76,35 +76,43 @@ export const setPartValue = <T extends Part>(
...
 };
 
-export const getPartValue = (part: NodePart) => part._value;
+export const getPartValue = (part: NodePart) => part._$value;
+
+// A sentinal value that can never appear as a part value except when set by
+// live(). Used to force a dirty-check to fail and cause a re-render.
+const RESET_VALUE = {};
+
+export const resetPartValue = (part: Part, value: unknown = RESET_VALUE) =>
```
sorvell, line 94: Why is this here? Add to PR description?

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -29,4 +35,22 @@ export const $private = {
   markerMatch: p._markerMatch,
   HTML_RESULT: p._HTML_RESULT,
   getTemplateHtml: p._getTemplateHtml,
+  patchDirectiveResolve: (
```
sorvell, line 38: Why were these added?

rollup-common.js
```diff
@@ -40,54 +40,75 @@ const skipBundleOutput = {
...
+const stableProperties = {
```
sorvell, line 65: I think we should try to add better descriptions for each property as to why it's here.

sorvell (changes requested):

packages/lit-html/src/parts.ts
kevinpschaaf: Added to PR description:
> * Adds `resetPartValue` helper for use by directives that were accessing the private `_value` property

packages/lit-html/src/directives/live.ts
kevinpschaaf: Added to PR description:
> * Adds `resetPartValue` helper for use by directives that were accessing the private `_value` property

packages/lit-element/src/private-ssr-support.ts
kevinpschaaf: Whoops, this is intended to just copy the pattern already established in `lit-html`; I added the comment from the same file in `lit-html`:
```js
/**
 * END USERS SHOULD NOT RELY ON THIS OBJECT.
 *
 * We currently do not make a mangled rollup build of the lit-ssr code. In order
 * to keep a number of (otherwise private) top-level exports  mangled in the
 * client side code, we export a $private object containing those members (or
 * helper methods for accessing private fields of those members), and then
 * re-export them for use in lit-ssr. This keeps lit-ssr agnostic to whether the
 * client-side code is being used in `dev` mode or `prod` mode.
 */
```

packages/lit-element/src/lit-element.ts
sorvell: OK, I understand this based on the pattern done in lit-html

packages/lit-element/src/private-ssr-support.ts
sorvell: Let's try to change this to `_private` instead of `$private` to match our convention for private (assuming top level exports don't get mangled).

packages/lit-html/src/private-ssr-support.ts
sorvell: Its' to avoid use of mangled properties.

packages/lit-element/src/lit-element.ts
kevinpschaaf: Background: This is following the patterns already established in `lit-html`

1. Yes, SSR only for LitElement
2. Just extra warning to users that they shouldn't rely on fields in it. Open to other suggestions.
3. We can use a different name; since it's a top-level export, the longer the name the longer

kevinpschaaf: I updated it to `_$private`

rollup-common.js
```diff
@@ -40,54 +40,75 @@ const skipBundleOutput = {
...
+  _$endNode: 'E',
+  _$getTemplate: 'F',
   // lit-html: TemplateInstance
-  _template: 'H',
+  _$template: 'G',
   // updating-element: UpdatingElement
-  _didUpdate: 'S',
-  _controllers: 'T',
+  _$didUpdate: 'H',
+  _$controllers: 'I',
```
sorvell, line 78: Let's try removing this.

sorvell (approved):
LGTM. (A couple minor nits in my review are being addressed.)

packages/lit-element/src/lit-element.ts
```diff
@@ -170,3 +170,26 @@ if (DEV_MODE) {
...
+ * Private exports for use by other Lit packages, not intended for use by
+ * external users.
+ *
+ * We currently do not make a mangled rollup build of the lit-ssr code. In order
+ * to keep a number of (otherwise private) top-level exports  mangled in the
+ * client side code, we export a _$private object containing those members (or
+ * helper methods for accessing private fields of those members), and then
+ * re-export them for use in lit-ssr. This keeps lit-ssr agnostic to whether the
+ * client-side code is being used in `dev` mode or `prod` mode.
+ */
```
justinfagnani, line 186: might as well add a `@private` jsdoc annotation here for any doc generators

packages/lit-element/src/private-ssr-support.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * END USERS SHOULD NOT RELY ON THIS OBJECT.
+ *
+ * We currently do not make a mangled rollup build of the lit-ssr code. In order
+ * to keep a number of (otherwise private) top-level exports  mangled in the
+ * client side code, we export a _$private object containing those members (or
+ * helper methods for accessing private fields of those members), and then
+ * re-export them for use in lit-ssr. This keeps lit-ssr agnostic to whether the
+ * client-side code is being used in `dev` mode or `prod` mode.
```
justinfagnani, line 25: add `@private` jsdoc annotation

rollup-common.js
```diff
@@ -40,54 +40,75 @@ const skipBundleOutput = {
...
+// Private properties which should be stable between versions but are used on
+// unambiguous objects and thus are safe to mangle. These include properties on
+// objects accessed between packages or objects used as values which may be
+// accessed between different versions of a given package.
+//
+// By convention, stable properties should be prefixed with `_$` in the code so
+// they are easily identifiable as properties requiring version stability and
+// thus special attention.
+//
+// Mangled names are uppercase letters, in case we ever might want to use
```
justinfagnani, line 58: I think we should consider prefixing with `_` for names of properties that end up on publicly accessible objects, like Parts and TemplateResults.

rollup-common.js
```diff
@@ -40,54 +40,75 @@ const skipBundleOutput = {
...
+  // lit-html: Template (used by platform-support)
+  _$createElement: 'A',
+  _$element: 'B',
+  _$options: 'C',
+  // lit-html: NodePart (used by platform-support)
+  _$startNode: 'D',
+  _$endNode: 'E',
+  _$getTemplate: 'F',
+  // lit-html: TemplateInstance (used by platform-support)
+  _$template: 'G',
```
justinfagnani, line 75: I'm surprised TemplateResult.strings and values isn't in here already. We need those to be stable across versions. Can you add them?

justinfagnani (approved):
LGTM, but we need to add TemplateResult fields.

rollup-common.js
kevinpschaaf: Done

rollup-common.js
kevinpschaaf: They're not prefixed with `_`, so they don't get mangled, and putting them in this list wouldn't do anything. Do we want them to go ahead and follow the `_$` convention also?

rollup-common.js
kevinpschaaf: This is just a list of the "special" mangled properties that need to be stable; there will be other mangled properties on e.g. Parts that will be whatever Terser picks for it. Does this suggest that all Terser mangling should use `_`?

packages/lit-element/src/private-ssr-support.ts
kevinpschaaf: Done

rollup-common.js
justinfagnani: Yeah. Removing the `_` was a temporary workaround

rollup-common.js
justinfagnani: oh right... we can't control the naming for the other properties. nvm then. If terser ever allows this we can enable it broadly. The trick would be to only incur the extra char on properties of publicly accessible objects.

packages/updating-element/src/updating-element.ts
```diff
@@ -615,7 +615,7 @@ export abstract class UpdatingElement extends HTMLElement {
   /**
    * Set of controllers.
    */
-  _controllers?: Controller[];
+  _$controllers?: Controller[];
```
sorvell, line 618: Take out `_$` from controllers

sorvell (changes requested):

sorvell (approved):

PR #1453, opened 2020-12-03
Heya

lithtml-dev:
We fixed this yesterday, we can pull it now


lithtml-dev:
merge it cunts

lithtml-dev:
also why can i see you here arthur, you should be working at home to stop the spread of covid-19, you selfish idiot

PR #1454, opened 2020-12-03
Update .eslintignore

lithtml-dev:


lithtml-dev:
we did this eeryesterday

PR #1459, opened 2020-12-04
Output ES2020

justinfagnani:
- Regen package-lock to pick up rollup-plugin-filesize 9.1.0
- Update most tsconfigs to target es2020.
- Update Firefox ESR in tests to latest version

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@ npm run build
 
 ### Test on a custom set of browsers
 
-1. `BROWSERS="chromium,sauce:Windows 10/firefox@68" npm test`
+1. `BROWSERS="chromium,sauce:Windows 10/firefox@78" npm test`
 
    Examples:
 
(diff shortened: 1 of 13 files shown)


packages/lit-html/static.d.ts
```diff
@@ -0,0 +1,47 @@
+/**
```
aomarks, line 1: Shouldn't be checked in?

aomarks:
> * Update most tsconfigs to target es2020. `localize` is left at es2018

Why not update localize too?

packages/lit-html/static.d.ts
justinfagnani: er... I need to fix the .gitignore

justinfagnani:
> Why not update localize too?

localize was the only package targeting es2018, I thought maybe for a reason. Updating the language level is a breaking change, is that ok? If so, I'll update.

aomarks:
> localize was the only package targeting es2018, I thought maybe for a reason. Updating the language level is a breaking change, is that ok? If so, I'll update.

Yep it's fine

justinfagnani:
PTAL

aomarks (approved):

PR #1460, opened 2020-12-05
[localize] Remove export/more concise hash delimiter code

aomarks:
- Remove unintentional _msg export
- Use ascii escape sequence instead of `fromCharCode`


aomarks:
Will subsume this into another PR.

PR #1463, opened 2020-12-08
[lit-next] Always passes options to all parts and makes available to directives via update

### LINKED ISSUE #1441: [lit-next] Provide access to host element via `renderOptions` to Directives
Directive authors sometimes want to hook into the LitElement within which the directive is used. While it's possible to do this by passing the element to the directive, that's a bit cumbersome `<div>${myDirective1(this)}${myDirective2(this)}</div>`. This info is already passed into to events via the `eventContext` in `renderOptions`. It would be convenient if directives had access to that.

sorvell:
Fixes #1441. Ensures `options` are consistently passed to all parts. Since directives have access to part in update, this makes options available to directives.


packages/lit-html/src/lit-html.ts
```diff
@@ -1200,8 +1200,8 @@ export class AttributePart {
     element: HTMLElement,
     name: string,
     strings: ReadonlyArray<string>,
-    parent?: Disconnectable,
-    _options?: RenderOptions
+    parent: Disconnectable | undefined,
+    public options: RenderOptions | undefined
```
justinfagnani, line 1204: I prefer not to use constructor field initializers - it makes the shape of the class less clear when reading.

packages/lit-html/src/lit-html.ts
```diff
@@ -1373,11 +1373,11 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
  */
 export class EventPart extends AttributePart {
   readonly type = EVENT_PART;
-  private _eventContext?: unknown;
+  private _host?: unknown;
 
   constructor(...args: ConstructorParameters<typeof AttributePart>) {
     super(...args);
-    this._eventContext = args[4]?.eventContext;
+    this._host = args[4]?.host;
```
justinfagnani, line 1380: Since options are stored on the part now, can we eliminate the constructor? We can use `this.options.host` in `handleEvent`

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1594,6 +1596,32 @@ suite('lit-html', () => {
       assert.isOk(event);
     });
 
+    test('directives have access to renderOptions', () => {
```
justinfagnani, line 1599: This is just testing attribute directives, so add "attribute" to the name.

Should we pass options to NodeParts too?

justinfagnani (changes requested):

packages/lit-html/src/test/lit-html_test.ts
sorvell: It also tests NodeParts

justinfagnani: oops!

packages/lit-html/src/lit-html.ts
sorvell: Changed.

packages/lit-html/src/lit-html.ts

packages/lit-html/src/lit-html.ts
```diff
@@ -1373,11 +1375,11 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
  */
 export class EventPart extends AttributePart {
   readonly type = EVENT_PART;
-  private _eventContext?: unknown;
+  private _host?: unknown;
 
   constructor(...args: ConstructorParameters<typeof AttributePart>) {
     super(...args);
-    this._eventContext = args[4]?.eventContext;
+    this._host = this.options?.host;
```
justinfagnani, line 1382: I'd remove the constructor entirely

packages/lit-html/src/lit-html.ts
```diff
@@ -1373,11 +1375,11 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
  */
 export class EventPart extends AttributePart {
   readonly type = EVENT_PART;
-  private _eventContext?: unknown;
+  private _host?: unknown;
```
justinfagnani, line 1378: remove this in favor of `this.options.host`

justinfagnani (changes requested):

packages/lit-html/src/lit-html.ts
sorvell: Done.

packages/lit-html/src/lit-html.ts
```diff
@@ -1432,7 +1434,7 @@ export class EventPart extends AttributePart {
     if (typeof this._$value === 'function') {
       // TODO (justinfagnani): do we need to default to this._$element?
       // It'll always be the same as `e.currentTarget`.
-      this._$value.call(this._eventContext ?? this.element, event);
+      this._$value.call(this._host ?? this.element, event);
```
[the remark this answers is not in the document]

packages/lit-html/src/lit-html.ts

packages/lit-html/src/lit-html.ts
```diff
@@ -343,7 +343,7 @@ export interface RenderOptions {
    * An object to use as the `this` value for event listeners. It's often
    * useful to set this to the host component rendering a template.
    */
-  eventContext?: EventTarget;
+  host?: EventTarget;
```
justinfagnani, line 346: I think we should update this to `Element` since it's more general now.

justinfagnani (approved):

PR #1465, opened 2020-12-08
[lit-html] Use Element#getAttributeNames() instead of Element#attributes

### LINKED ISSUE #1353: [lit-html] Use Element#getAttributeNames() instead of Element#attributes
It should be a bit faster.

justinfagnani:
Fixes #1353

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -647,20 +647,18 @@ class Template {
         // increment the bindingIndex, and it'll be off by 1 in the element
         // and off by two after it.
         if ((node as Element).hasAttributes()) {
-          const {attributes} = node as Element;
           // We defer removing bound attributes because on IE we might not be
           // iterating attributes in their template order, and would sometimes
           // remove an attribute that we still need to create a part for.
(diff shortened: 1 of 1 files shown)


kevinpschaaf (approved):

kevinpschaaf:
How do we turn off the annoying `'ADirective' is defined but never used` lint warnings on github that don't happen on CI?

It's a dumb warning anyway; giving a class a name even when used as an expression is not "unused" because it'll still end up showing in `this.constructor.name` and will show in devtools nicely, etc.

abdonrd:
> How do we turn off the annoying `'ADirective' is defined but never used` lint warnings on github that don't happen on CI?

It happen on CI, but only the errors (not the warnings) "break" the CI task:
https://github.com/Polymer/lit-html/pull/1465/checks?check_run_id=1525541472#step:5:10

abdonrd:
> It's a dumb warning anyway; giving a class a name even when used as an expression is not "unused" because it'll still end up showing in `this.constructor.name` and will show in devtools nicely, etc.

These are the options that the rule has: https://eslint.org/docs/rules/no-unused-vars#options

PR #1466, opened 2020-12-09
[localize] Switch to UTF-16 version of FNV-1A hash

aomarks:
This drops total `lit-localize.js` minified + brotli size from 1.09 KiB to 0.95 KiB (13.94 -> 11.25 uncompressed). The hash module by itself drops from 0.38 KiB to 0.25 KiB (4.15 -> 1.46 uncompressed).

Previously we were using a version that converted each charcode to UTF-8, but there's no reason to do that other than compatibility with other more-common UTF-8 based implementations, which isn't a requirement.

First render of 100 templates:

image

Also two additional changes that further drop `lit-localize.js` to 0.94 KiB (10.98 uncompressed):

- Omit the `a` version prefix from IDs. The `h` and `s` can effectively already serve as a version indicator, because if we did a V2, we could just switch to different characters, or add a prefix. This also removes a byte from each generated ID.
- Use `\x1e` instead of `String.fromCharCode(30)` to generate the record delimiter code.

### DIFF
--- a/packages/localize/src_client/fnv1a64.ts
+++ b/packages/localize/src_client/fnv1a64.ts
@@ -3,10 +3,11 @@ Copyright 2014 Travis Webb
 SPDX-License-Identifier: MIT */
 
 // This module is derived from the file:
-// https://github.com/tjwebb/fnv-plus/blob/1e2ce68a07cb7dd4c3c85364f3d8d96c95919474/index.js#L530
+// https://github.com/tjwebb/fnv-plus/blob/1e2ce68a07cb7dd4c3c85364f3d8d96c95919474/index.js#L309
 //
 // Changes:
-// - Only the _hash64_1a_fast_utf function is included.
+// - Only the _hash64_1a_fast function is included.
+// - Removed loop unrolling.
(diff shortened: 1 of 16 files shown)


packages/localize/src_client/fnv1a64.ts
```diff
@@ -36,100 +34,8 @@ export function fnv1a64(str: string): string {
...
-      t1 += t0 >>> 16;
-      v0 = t0 & 65535;
-      t2 += t1 >>> 16;
-      v1 = t1 & 65535;
-      v3 = (t3 + (t2 >>> 16)) & 65535;
-      v2 = t2 & 65535;
-      v0 ^= (c & 63) | 128;
-    }
+  for (let i = 0; i < str.length; i++) {
+    v0 ^= str.charCodeAt(i);
```
justinfagnani, line 38: wow, nice!

justinfagnani (approved):

aomarks:
Pushed two additional changes to further shrink lib and ID size, PTAL!

justinfagnani (approved):
LGTM+1

PR #1470, opened 2020-12-09
Change nothing and noChange to global Symbols

justinfagnani:
This makes the `nothing` and `noChange` sentinels work across multiple instances of the core library.


kevinpschaaf (approved):
Nice, I like the changes in `AttributePart:_$setValue`.

I just double-checked, and it looks like the change in semantics `nothing`/`noChange` re: `isPrimitive` didn't require changes in `lit-ssr` because we explicitly test for them, nor in `hydrate` because they were being handled in the fallback, which is almost identical to the `isPrimitive` case.

PR #1473, opened 2020-12-09
Small documentation cleanups

rictic:
Fixed a few typos, like
the the -> the
@returns -> @return

Dropped types from jsdoc, as they're now in TypeScript.

Reflowed a few comments in @param and @return for consistency.


justinfagnani (approved):
Thanks!

PR #1480, opened 2020-12-11
[lit-html] Rename NodePart to ChildPart

justinfagnani:
Also renames several methods, properties, and constants related to NodeParts.

### DIFF
--- a/packages/lit-html/src/directives/class-map.ts
+++ b/packages/lit-html/src/directives/class-map.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {AttributePart, noChange, PartInfo} from '../lit-html.js';
+import {AttributePart, DirectiveParameters, noChange, PartInfo} from '../lit-html.js';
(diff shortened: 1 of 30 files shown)


packages/lit-html/src/directives/class-map.ts
```diff
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {AttributePart, noChange, PartInfo} from '../lit-html.js';
+import {AttributePart, DirectiveParameters, noChange, PartInfo} from '../lit-html.js';
```
kevinpschaaf, line 15: Nit: `DirectiveParameters` is also exported from `directive.js`, might be nicer to see all the directive stuff imported from there. But nice that you cleaned this up.

kevinpschaaf (approved):

packages/lit-html/src/directives/class-map.ts
justinfagnani: I'm going to do another pass on where the directive related interfaces are exported from...

PR #1481, opened 2020-12-11
Move more directive related types to directive.ts

justinfagnani:


packages/lit-html/src/directives/live.ts
```diff
@@ -12,8 +12,13 @@
...
+import {AttributePart, noChange, nothing, PartInfo} from '../lit-html.js';
+import {
+  directive,
+  Directive,
+  DirectiveParameters,
+  PartType,
+} from '../directive.js';
 import {resetPartValue} from '../directive-helpers.js';
 
 class LiveDirective extends Directive {
```
kevinpschaaf, line 24: While you're in here, I realized that now that the part type checks aren't instanceof, we can actually check `partInfo.type === PartType.ATTRIBUTE` (whereas `instanceof AttributePart` would have also matched e.g. `PropertyPart`, so they had to be more blacklistey). So we can change the checks here to be what it works _with_, not the other way around, which I think is a better pattern that allows adding more types in the future. 

Wanna make that change? (and check the others?)

kevinpschaaf (approved):
These changes LFTM.

Do you want to go ahead and make the `Part`'s type-only exports here too?

packages/lit-html/src/directives/live.ts
justinfagnani: done. live was the only one that did a positive check

Issue #189, opened 2017-11-14
[lit-html] Precompiling Templates


jridgewell:
Related to https://github.com/PolymerLabs/lit-html/issues/187, I'm opening this so we can discuss how to precompile templates.

The general idea is how do we take creating a `Template` instance out of the first-render load time. Ie, make this [new](https://github.com/PolymerLabs/lit-html/blob/1eb0baf/src/lit-html.ts#L55) disappear, taking [all of this](https://github.com/PolymerLabs/lit-html/blob/1eb0baf/src/lit-html.ts#L184-L347) (163 lines of very slow code!) out of the first render.

I think we can manage to precompile these in a build step, making a file instantiate `Templates` with the correct `TemplateStringsArray` and already instantiated `TemplatePart`s  (with the correct indexes). We might need to expose the `templateCache` externally.

RayBenefield:
👍 

This would be awesome for what I'm working on as well. I've been using https://www.npmjs.com/package/es6-template-strings as an intermediary to setup a templating system with individual files as templates loaded in instead of baked into the JS files to enable designers similar to templating engines in full stack frameworks like Django/Laravel/Rails, using raw ES6. Something that takes a stringified form of an ES6 template would be huge in cutting that out of my pipeline.

jridgewell:
So what I'm imagining is:

```js
// templates/hello-world.js
import { html } from 'lit-html';

export function hello(name) {
  return html`<div>
    ${name}
  </div>`;
}
```

Would be transformed into:
```js
import { html, templateCache, Template } from 'lit-html';

const template = new Template([], false);

function precompiled(array, values) {
  template.element.innerHTML = interpolate(array, values);
  template.parts = [
    new TemplatePart('node', 2),
  ];
  templateCache.put(array, template);
};
precompiled`<div>
    ${<!--marker-->}
  </div>`;

export function hello(name) {
  return html`<div>
    ${name}
  </div>`;
}
```

Maybe even:

```js
import { Template, TemplateResult } from 'lit-html';

// Imagine if Template's constructor was split into a separate method
// and instead was simple.
const template = new Template();

const strings = [`<div>
    `, `
  </div>`];
template.element.innerHTML = interpolate(strings, ['<!--marker-->']);
template.parts = [
  new TemplatePart('node', 2),
];

export function hello(name) {
  return new TemplateResult(template, [name]);
}
```

lukechu10:
Has there been any progress on this? The last message was over 2 years ago!

[...]

PR #1484, opened 2020-12-14
[lit-html] Initial prototype of template compiler transform

justinfagnani:
Fixes https://github.com/lit/lit/issues/189


augustjk:
This got incorporated into `@lit-labs/compiler`

PR #1485, opened 2020-12-14
[lit-next] Changes UpdatingElement to ReactiveElement and @internalProperty to @state

sorvell:
**ReactiveElement**
* changes all files/folders named `updating-element` to `reactive-element`
* in all files
  * `updating-element` => `reactive-element`
  * `UpdatingElement` => `ReactiveElement`

** @state**
* changes `@internalProperty` to `@state`
   * changes all files named `internaProperty` to `state`
   * changes in files `internalProperty` to `state`
* adds to `PropertyDeclaration` `state` field which when set to `true` automatically sets `attribute` to `false`.

updates CHANGELOG


packages/reactive-element/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [Breaking] The `@internalProperty` decorator has been renamed to `@state`.
+- [Breaking] `UpdatingElement` has been renamed to `ReactiveElement` and the `updating-element` package has been renamed to `reactive-element`.
```
justinfagnani, line 25: add a "(since 3.0.0-pre1)" after the `[Breaking]`, and maybe split this into two items.

We should also consider re-exporting ReactiveElement as UpdatingElement from the lit-element package for back-compat.

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1095,24 +1110,24 @@ export abstract class UpdatingElement extends HTMLElement {
 
 // Apply polyfills if available
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-(globalThis as any)['updatingElementPlatformSupport']?.({UpdatingElement});
+(globalThis as any)['updatingElementPlatformSupport']?.({ReactiveElement});
```
kevinpschaaf, line 1113: Rename `updatingElementPlatformSupport` -> `reactiveElementPlatformSupport`?

packages/reactive-element/src/decorators/state.ts
```diff
@@ -39,6 +39,10 @@ export interface InternalPropertyDeclaration<Type = unknown> {
  * properties may be renamed by optimization tools like closure compiler.
  * @category Decorator
  */
-export function internalProperty(options?: InternalPropertyDeclaration) {
-  return property({attribute: false, hasChanged: options?.hasChanged});
+export function state(options?: InternalPropertyDeclaration) {
+  return property({
+    state: true,
+    attribute: false,
+    hasChanged: options?.hasChanged,
```
kevinpschaaf, line 46: This seems like it was implemented unsafely earlier; shouldn't we spread props into the object first, otherwise you can't add features via `getPropertyDescriptor` like you can with `property`?

kevinpschaaf (changes requested):

packages/reactive-element/src/reactive-element.ts

packages/reactive-element/src/decorators/state.ts
sorvell: Done.

kevinpschaaf (approved):
LGTM, with nit about reordering the discussion about property vs. state

PR #1487, opened 2020-12-15
Add isSingleExpression directive helper

justinfagnani:


packages/lit-html/src/directive-helpers.ts
```diff
@@ -59,6 +64,16 @@ export const isDirectiveResult = (
     ? (value as DirectiveResult)?._$litDirective$ !== undefined
     : (value as DirectiveResult)?._$litDirective$ === klass;
 
+/**
+ * Tests is a part is a single-expression part.
```
kevinpschaaf, line 68:
```suggestion
 * Tests whether a part has a single-expression and no strings to interpolate between.
```

kevinpschaaf, line 68:
```suggestion
 * Tests whether a part has a single-expression and no strings to interpolate between.
```

kevinpschaaf: This is tricky; this function returns true if there's one value AND empty strings to either side of it. So there's two conditions to meet.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -59,6 +64,16 @@ export const isDirectiveResult = (
...
+ *
+ * Only AttributePart and PropertyPart can by multi-valued, when the expression
+ * that created them is an interpolation. Multi-expression parts have a
```
kevinpschaaf, line 71: The phrasing "when the expression that created them is an interpolation" reads strange to me. I'm not sure defining "multi-valued" is helping here, because being multi-valued (if that means `values.length > 1`) isn't the only way for this to return false. It could also be `values.length===1` but the strings were also empty.

kevinpschaaf (approved):
Code LGTM, but maybe see if you can tighten up the API docs there?

sorvell (approved):

packages/lit-html/src/directive-helpers.ts
justinfagnani: done

packages/lit-html/src/directive-helpers.ts
justinfagnani: removed that clause

PR #1492, opened 2020-12-15
[lit-next] Use __ for class properties and add mangling prefix

kevinpschaaf:
We determined that our exported classes intended for user subclassing are susceptible to mangling collision if the user subclass is subsequently mangled using similar settings. This occurred in practice between our `LitElement` and `ReactiveElement` classes, but could theoretically occur with in-the-wild subclasses.

As such, this PR introduces a new convention where fields on user-subclassable base classes use a double-underscore `__` prefix. We then modify the seeded nameCache to prefix the chosen mangled name for these with a unique unicode prefix per package, which should ensure we get the benefit of minifying private class fields while minimizing the risk of collision.

### DIFF
--- a/rollup-common.js
+++ b/rollup-common.js
@@ -145,6 +145,10 @@ const prefixClassProperties = (context, nameCache, prefix) => {
       props: {props},
     } = nameCache;
     for (const p in props) {
+      // Note all properties in the terser name cache are prefixed with '$'
+      // (presumably to avoid collisions with built-ins). Checking for the
+      // prefix is just to ensure we don't double-prefix properties if
+      // `prefixClassProperties` is called twice on the same `nameCache`.
       if (p.startsWith('$__') && !props[p].startsWith(prefix)) {
         props[p] = prefix + props[p];
       }
--- a/rollup-common.js
+++ b/rollup-common.js
@@ -99,10 +99,6 @@ const stableProperties = {
   _$setChildPartConnected: 'S',
   // lit-html: ChildPart (used by directive-helpers)
   _$clear: 'T',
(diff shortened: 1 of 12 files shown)


rollup-common.js
```diff
@@ -123,7 +119,42 @@ Object.entries(stableProperties).forEach(([prop, mangle], i) => {
...
+ */
+const addedClassPrefix = new WeakSet();
+const prefixClassProperties = (context, nameCache, prefix) => {
+  // Only prefix class properties once per options context, as a perf optimization
+  if (nameCache && !addedClassPrefix.has(context)) {
+    const {
+      props: {props},
+    } = nameCache;
+    for (const p in props) {
+      if (p.startsWith('$__') && !props[p].startsWith(prefix)) {
```
sorvell, line 148: Why is it `$__` and not `__` and why do you need to check if the prop doesn't start with the prefix given the optimization?

sorvell (changes requested):

rollup-common.js
kevinpschaaf: Al had a comment in there about this:
https://github.com/Polymer/lit-html/blob/15624b3b163643a8979151fffaaee7d9ee988d14/rollup-common.js#L218-L219

Checking if the prop starts with the prefix makes it safe to run `prefixClassProperties` more than once on the same `nameCache`; I was just thinking that since the code could get refactored, we might not catch if we were double-prefixing properties.

kevinpschaaf: Added comments.

kevinpschaaf:
@sorvell PTAL

sorvell (approved):

justinfagnani (approved):

PR #1493, opened 2020-12-15
[lit-next] Upgrade to TypeScript 4.1.3

### LINKED ISSUE #1491: [lit-next] Builds fail when upgrading to latest TypeScript
After removing the package-lock and installing, builds fail for at least for ReactiveElement and LitElement. It looks like this might be TypeScript version related.

The code should work with the latest TypeScript.

sorvell:
Fixes #1491. The latest version of TypeScript is a little more strict around `unknown` types which required some minor tweaks.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -307,7 +307,7 @@ export const render = (
   if (part === undefined) {
     const endNode = options?.renderBefore ?? null;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (partOwnerNode as any).$lit$ = part = new ChildPart(
+    (partOwnerNode as any).$lit$ = part = new ChildPartImpl(
       container.insertBefore(createMarker(), endNode),
       endNode,
       undefined,
@@ -346,8 +346,8 @@ let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
 // `resolveDirective`
 export interface DirectiveParent {
   _$parent?: DirectiveParent;
-  _directive?: Directive;
-  _directives?: Array<Directive | undefined>;
+  __directive?: Directive;
+  __directives?: Array<Directive | undefined>;
 }
 
 /**
@@ -506,7 +506,8 @@ const getTemplateHtml = (
   ];
 };
 
-class Template {
+export type Template = Interface<TemplateImpl>;
+class TemplateImpl {
   /** @internal */
   _$element!: HTMLTemplateElement;
   /** @internal */
@@ -573,12 +574,12 @@ class Template {
                 _strings: statics,
                 _constructor:
                   m[1] === '.'
-                    ? PropertyPart
+                    ? PropertyPartImpl
                     : m[1] === '?'
-                    ? BooleanAttributePart
+                    ? BooleanAttributePartImpl
                     : m[1] === '@'
-                    ? EventPart
-                    : AttributePart,
+                    ? EventPartImpl
+                    : AttributePartImpl,
               });
               bindingIndex += statics.length - 1;
             } else if (name.startsWith(marker)) {
@@ -647,7 +648,7 @@ class Template {
 }
 
 export interface Disconnectable {
-  _$parent: Disconnectable | undefined;
+  _$parent?: Disconnectable;
   _$disconnetableChildren?: Set<Disconnectable>;
 }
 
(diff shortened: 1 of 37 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -124,7 +124,7 @@ const isArray = Array.isArray;
 const isIterable = (value: unknown): value is Iterable<unknown> =>
   isArray(value) ||
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  (value && typeof (value as any)[Symbol.iterator] === 'function');
+  ((value && typeof (value as any)[Symbol.iterator] === 'function') as boolean);
```
justinfagnani, line 127: I wonder if we should write this as `typeof (value as any)?.[Symbol.iterator] === 'function'`

packages/reactive-element/src/reactive-element.ts
```diff
@@ -737,7 +737,7 @@ export abstract class ReactiveElement extends HTMLElement {
    * overridden on the element instance with a function that triggers the first
    * update.
    */
-  protected enableUpdating() {}
+  protected enableUpdating(_value: boolean) {}
```
justinfagnani, line 740: why was this needed? Can you make it optional to leave out the `true` constant above?

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
sorvell: Done and this removed the need for the cast.

packages/reactive-element/src/reactive-element.ts

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -2197,13 +2197,13 @@ suite('ReactiveElement', () => {
       promiseFulfilled = false;
 
       get updateComplete() {
-        return (async () => {
+        return (async (): Promise<boolean | void> => {
```
kevinpschaaf, line 2200: Seems like you changed this to not resolve to `boolean`?

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -2294,7 +2294,7 @@ suite('ReactiveElement', () => {
   });
 
   test('can override performUpdate()', async () => {
-    let resolve: (() => void) | undefined;
+    let resolve: ((value?: boolean) => void) | undefined;
```
kevinpschaaf, line 2297: Also never resolved to a `boolean`?

kevinpschaaf (commented):
Some minor nits; it's possible I'm just misunderstanding.

abdonrd (commented):
As you are in a mono-repo, why not remove the `typescript` dependency from the packages? With the root `package.json` should be enough and more easy to mantain.

packages/reactive-element/src/test/reactive-element_test.ts
sorvell: Changed to `unknown`. It appears the promise resolver argument is now required.

packages/reactive-element/src/test/reactive-element_test.ts
sorvell: Removed. The `updateComplete` promise previously resolved to `unknown`. This was changed and now reverted back.

It normally resolves to a boolean but the first one resolves to `undefined` and it it's maybe not worth changing that?

kevinpschaaf: Hmm, if we're going to resolve to a boolean, shouldn't the initial `updateComplete` correctly reflect the `isUpdatePending` state? If that's intended to be a useful thing, seems bad if you can't rely on that to know if a further update was enqueued on the first render, only on subsequent ones.

It sort of points out that the gambit of just making `enableProperties` be the initial `__updatePromise` resolver is maybe too simplistic, since we're not actually resolving after `performUpdate()` is done; it just sort of works because a user's `await el.updateComplete` will be the second thing queued behind the `await this.__updatePromise` in `__enqueueUpdate()`. We should really be waiting to resolve the `__updatePromise` after `performUpdate` when we know the `isUpdatePending` state.

It feels like we should be able to refactor a little to use the new `__pendingConnectionPromise` now to do the initial connection holdoff as well, maybe?

Let's make an issue and not block, but seems like we should fix that or remove the boolean, one or the other.

kevinpschaaf:
@abdonrd Since we run the build scripts per-package (and I think we want to keep that way, at least for now), it'd take a little more work to run tsc, rollup, etc. from the top level. We may consider that in a subsequent PR.

abdonrd:
@kevinpschaaf got it, thanks Kevin! New PR: https://github.com/Polymer/lit-html/pull/1498

PR #1494, opened 2020-12-16
Export part interfaces instead of constructors

justinfagnani:


packages/lit-html/CHANGELOG.md
```diff
@@ -78,6 +78,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   and `PartInfo` types have been moved from `lit-html` to
   `lit-html/directive.ts`.
 
+- [Breaking] (since 2.0.0-pre.3) The part exports (`ChildPart`,
+  `AttributePart`, etc) have been change to interface-only exports. The constructors are no longer exported.
```
kevinpschaaf, line 82:
```suggestion
  `AttributePart`, etc) have been change to interface-only exports. The constructors are no longer exported. Directive authors should use helpers in `directive-helpers.js` to construct parts.
```

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -52,4 +52,10 @@ export const _$private = {
     return committedValue;
   },
   resolveDirective: p._resolveDirective,
+  // ChildPart: p._ChildPart,
+  // AttributePart: p._AttributePart,
+  // PropertyPart: p._PropertyPart,
+  // BooleanAttributePart: p._BooleanAttributePart,
+  // EventPart: p._EventPart,
```
kevinpschaaf, line 59: Remove?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -12,10 +12,11 @@
  * http://polymer.github.io/PATENTS.txt
  */
 import {
+  _$private,
```
kevinpschaaf, line 15: Remove?

packages/lit-html/src/platform-support.ts
```diff
@@ -87,11 +87,11 @@ const scopeCssStore: Map<string, string[]> = new Map();
...
 (globalThis as any)['litHtmlPlatformSupport'] ??= ({
-  ChildPart,
-  Template,
+  _$ChildPart: ChildPart,
+  _$Template: Template,
 }: {
-  ChildPart: PatchableChildPart;
-  Template: PatchableTemplate;
+  _$ChildPart: PatchableChildPart;
+  _$Template: PatchableTemplate;
```
kevinpschaaf, line 94: I would just change these to an argument list; doesn't seem worth burning more stable properties to make this an options bag...

kevinpschaaf (changes requested):

packages/lit-html/CHANGELOG.md
justinfagnani: done

packages/lit-html/src/private-ssr-support.ts
justinfagnani: done

packages/lit-html/src/platform-support.ts
justinfagnani: done

cc @sorvell to confirm. It means future changes just have to be adding to the list, but maybe we don't expect any.

packages/lit-html/src/test/lit-html_test.ts
justinfagnani: done

packages/lit-html/src/lit-html.ts
```diff
@@ -819,7 +829,10 @@ export type Part =
   | ElementPart
   | EventPart;
 
-export class ChildPart {
+type Interface<T> = {[P in keyof T]: T[P]};
+
+export type ChildPart = Interface<ChildPartImpl>;
+class ChildPartImpl implements ChildPart {
```
sorvell, line 835: Is the `implements` necessary? Isn't this just "I implement myself"?

packages/lit-html/src/lit-html.ts
```diff
@@ -1063,7 +1076,8 @@ export class ChildPart {
   }
 }
 
-export class AttributePart {
+export type AttributePart = Interface<AttributePartImpl>;
+class AttributePartImpl {
```
sorvell, line 1080: Why doesn't this have the `implements` thing?

sorvell (approved):
LGTM, just a couple nits

packages/lit-html/src/lit-html.ts
justinfagnani: oh oops, this is from when I had a hand-written interface

kevinpschaaf (approved):

PR #1495, opened 2020-12-16
[lit-html] Fix cache() directive

justinfagnani:
The cache tests has a wayward `test.only` in them, hiding some failures. The directive wasn't clearing the container part at the right times, and was setting the container part's value wrong when restoring from cache. The fix should read a bit better now.

### DIFF
--- a/packages/lit-html/src/directives/cache.ts
+++ b/packages/lit-html/src/directives/cache.ts
@@ -57,12 +57,12 @@ export const cache = directive(
     }
 
     update(containerPart: ChildPart, [v]: DirectiveParameters<this>) {
-      // If the new value is not a TemplateResult from the same Template as the
-      // previous value, move the nodes from the DOM into the cache.
+      // If the previous value is a TemplateResult and the new value is not,
+      // or is a different Template as the previous value, move the child part
+      // from into the cache.
       if (
-        this.value !== undefined &&
-        this.value.strings !== (v as TemplateResult).strings
-      ) {
+        isTemplateResult(this.value) && 
+        (!isTemplateResult(v) || this.value.strings !== v.strings)) {
         // This is always an array because we return [v] in render()
         const partValue = getComittedValue(containerPart) as Array<ChildPart>;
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/directives/cache.ts
```diff
@@ -57,38 +57,42 @@ export const cache = directive(
     }
 
     update(containerPart: ChildPart, [v]: DirectiveParameters<this>) {
-      // If the new value is not a TemplateResult from the same Template as the
-      // previous value, move the nodes from the DOM into the cache.
+      // If the previous value is a TemplateResult and the new value is not,
+      // or is a different Template as the previous value, move the child part
+      // from into the cache.
```
kevinpschaaf, line 62:
```suggestion
      // into the cache.
```

packages/lit-html/src/directives/cache.ts
```diff
@@ -57,38 +57,42 @@ export const cache = directive(
...
           const fragment = new DocumentFragment();
           cachedContainerPart = render(nothing, fragment);
-          setComittedValue(cachedContainerPart, [childPart]);
           this.templateCache.set(this.value.strings, cachedContainerPart);
         }
         // Move into cache
-        insertPart(containerPart, undefined, childPart);
+        setComittedValue(cachedContainerPart, [childPart]);
+        insertPart(cachedContainerPart, undefined, childPart);
         clearPart(containerPart);
```
kevinpschaaf, line 78: Should the clear here be necessary if we've definitely just moved the whole `childPart` which is what would have been committed there into the other part?

packages/lit-html/src/directives/cache.ts
justinfagnani: done

packages/lit-html/src/directives/cache.ts
justinfagnani: ah, good point

kevinpschaaf (approved):
LGTM!

PR #1506, opened 2020-12-18
[lit-next] Give private exports unique names

### LINKED ISSUE #1505: [lit-next] Conflicting re-exports _$private between lit-html and lit-element
# Description
`lit-element.ts` does `export * from 'lit-html'`, and since both `lit-element` and `lit-html` export a `_$private` object, they conflict. Although this doesn't cause any problems in the browser, when passing them through rollup to create a bundle, rollup will warn: 

```
(!) Conflicting re-exports
../../node_modules/lit-element/lit-element.js re-exports '_$private' from both ../../node_modules/lit-element/lit-element.js and ../../node_modules/lit-html/lit-html.js (will be ignored)
```

kevinpschaaf:
Renames the private object export `_$private` to have non-conflicting names, since `lit-element` re-exports all of `lit-html`:

* `lit-html.ts`: `_$private` → `_Σ`
* `lit-element.ts`: `_$private` → `_Φ`

Fixes #1505


sorvell (approved):

justinfagnani:
how about instead of `le` and `lh`, we use the same greek symbols we use for prefixes?

justinfagnani (approved):

PR #1507, opened 2020-12-18
[labs] Reorganizes labs content into separate packages

sorvell:
Instead of one large labs package, there are now individual packages for labs features. Moves `async-task` and `create-react-component` to their own packages.


justinfagnani:
The re-org looks good. I would consider different names. May just `react` and `task`? I could see other related utilities going in those packages.

packages/labs/react/package.json
```diff
@@ -74,9 +55,7 @@
     "@types/react-dom": "^16.9.8"
   },
   "dependencies": {
-    "@lit/reactive-element": "^1.0.0-pre.1",
-    "lit-element": "^3.0.0-pre.2",
-    "lit-html": "^2.0.0-pre.4"
+    "@lit/reactive-element": "^1.0.0-pre.1"
```
kevinpschaaf, line 58: Should be a `devDependency` I think? Not imported in `create-component.ts`

packages/labs/react/rollup.config.js
```diff
@@ -12,14 +12,10 @@
...
 export default litProdConfig({
   classPropertyPrefix: 'Ω',
-  entryPoints: [
-    'index',
-    'frameworks/react/create-component',
-    'controllers/async-task',
-  ],
-  external: ['@lit/reactive-element', 'lit-html', 'lit-element'],
+  entryPoints: ['index', 'create-component'],
+  external: ['@lit/reactive-element'],
```
kevinpschaaf, line 20: Is this required?

packages/labs/task/README.md
```diff
@@ -0,0 +1,66 @@
...
+import {AsyncTask, TaskStatus} from '@lit-labs/task';
+// ...
+
+class MyElement extends LitElement {
+
+  @state()
+  private _userId: number;
+
+  private _apiTask = new AsyncTask(
+      this,
```
kevinpschaaf, line 44: (Commenting here since the file is unchanged): Just noting that `AsyncTask` contructor's first argument is typed as `ReactiveElement` which could/should now be `ReactiveElementHost`:

https://github.com/Polymer/lit-html/blob/f547ac3f8ef8ebc252b8aa6b1bd1cb60c014df2f/packages/labs/task/src/async-task.ts#L96-L100

kevinpschaaf (changes requested):
A couple minor comments...

Also I guess to @justinfagnani re: `task` vs. `async-task`... what else would you put in this package that would be task-related and not want its own package?  `async-task` as a package name to match the controller class seems better unless there are concrete ideas for a more general-purpose package imo.

packages/labs/react/package.json
sorvell: Done.

packages/labs/react/rollup.config.js
sorvell: Removed.

packages/labs/task/README.md
sorvell: Done.

justinfagnani:
@kevinpschaaf 

> Also I guess to @justinfagnani re: task vs. async-task... what else would you put in this package that would be task-related and not want its own package? async-task as a package name to match the controller class seems better unless there are concrete ideas for a more general-purpose package imo.

I just think that "async" is kind of redundant with "task". We could rename `AsyncTask` to `Task` too, IMO.

justinfagnani (approved):

sorvell:
Changed `AsyncTask` to `Task`

kevinpschaaf (approved):

PR #1510, opened 2020-12-18
[lit-next] Reduce the ESLint config

abdonrd:
A few files no longer exist.


abdonrd:
Friendly ping to @aomarks

PR #1514, opened 2020-12-31
turn ban-types eslint rule on

### LINKED ISSUE #1509: [lit-next] Enable @typescript-eslint/ban-types rule
## Description

When we setup the ESLint config, we disabled the [`@typescript-eslint/ban-types`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.9.0/packages/eslint-plugin/docs/rules/ban-types.md) to avoid blocking merge the setup.
https://github.com/Polymer/lit-html/blob/7a625b71c72d0a8f96a855b172394cc6cd745c1b/.eslintrc.json#L18

But ideally, now we should address it to improve typing.

Right now, we have "just" 32 problems. And we have hints like this:

```
/Users/abdonrd/Work/Forks/Polymer/lit-html/packages/reactive-element/src/reactive-element.ts

  468:36  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead  @typescript-eslint/ban-types

  471:19  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead  @typescript-eslint/ban-types
```

## Acceptance criteria

Remove this line:
https://github.com/Polymer/lit-html/blob/7a625b71c72d0a8f96a855b172394cc6cd745c1b/.eslintrc.json#L18

And fix the current problems.

43081j:
fixes #1509

if nobody did it already, here you go.


packages/reactive-element/src/decorators/property.ts
```diff
@@ -71,10 +71,11 @@ const standardProperty = (
 
 const legacyProperty = (
   options: PropertyDeclaration,
-  proto: Object,
+  proto: unknown,
```
43081j, line 74: decorators were the main thing using `Object`, so im not sure this is the right solution. extra pair of eyes would be nice, would this break them for consumers? also specifically the property decorator needs an extra cast this way but the rest don't

packages/labs/ssr/src/lib/render-module.ts
```diff
@@ -32,7 +32,7 @@ export const renderModule = async (
     },
   });
   const module = await importModule(specifier, referrer, window);
-  const f = module.namespace[functionName] as Function;
+  const f = module.namespace[functionName] as (...args: unknown[]) => unknown;
```
43081j, line 35: this means the renderModule return type is now `unknown` i guess, but the only usage i found of it casts the result anyway so should be fine?

packages/lit-html/src/lit-html.ts
```diff
@@ -310,7 +310,7 @@ export interface RenderOptions {
    * An object to use as the `this` value for event listeners. It's often
    * useful to set this to the host component rendering a template.
    */
-  host?: object;
+  host?: unknown;
```
43081j, line 313: another one to double check since i guess if we have consumers of `host`, it would be a breaking change?

43081j:
@justinfagnani i rebased this onto main and fixed up any new lint errors in case its still of use.

also left some comments around the changes to exposed types

43081j:
rebased again and fixed a couple new lint errors

packages/reactive-element/src/decorators/property.ts
justinfagnani: `Object` is arguably the correct type here. I can't remember why we didn't tighten the type to `{constructor: typeof ReactiveElement}` or something. We can't do that now, but that's what I'd use in the cast below.

justinfagnani (changes requested):
Thanks!

I think this is all internal changes except for one thing in SSR. It'd be good to get one more set of eyes.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -23,15 +23,15 @@ export interface ClassElement {
...
+  initializer?: () => void;
   extras?: ClassElement[];
   finisher?: <T>(clazz: Constructor<T>) => void | Constructor<T>;
   descriptor?: PropertyDescriptor;
 }
 
 export const legacyPrototypeMethod = (
   descriptor: PropertyDescriptor,
-  proto: Object,
+  proto: unknown,
```
rictic, line 34: proto should at least be `object` yeah? i.e. it can't be null, undefined, or a primitive

packages/reactive-element/src/decorators/property.ts
rictic: Yeah, we don't want unknown here, this can't be null, undefined, or a primitive. `object` is also a good candidate for this.

packages/localize/src/init/reentrant.ts
```diff
@@ -53,6 +53,7 @@ export interface ReentrantConfiguration {
  */
 export const configureReentrantLocalization: ((
   config: ReentrantConfiguration
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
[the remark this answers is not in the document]
43081j: yup i wasn't entirely sure what the intention was here with the `{}` return type. calls of this function don't seem to use the return type as far as i can see

packages/labs/ssr/src/lib/render-module.ts
43081j: both `any` and `Function` trigger  a lint rule so we could just leave it as-is, with a TODO and an `eslint-disable` maybe?

rictic: sounds good to me

rictic (approved):

packages/labs/virtualizer/src/layouts/shared/SizeGapPaddingBaseLayout.ts
```diff
@@ -76,9 +76,20 @@ type Padding = {[key in side]: number};
 export abstract class SizeGapPaddingBaseLayout<
   C extends SizeGapPaddingBaseLayoutConfig
 > extends BaseLayout<C> {
-  protected _itemSize: Size | {} = {};
-  protected _gaps: Gaps | {} = {};
-  protected _padding: Padding | {} = {};
+  protected _itemSize: Size = {
```
43081j, line 79: @rictic what do you think of this?

i had to make it `Infinity` because `0` might trip up the equality checks later on (in that the initial equality check before this change was `undefined === someNumber`). so if we made it `0`, it could be the same as the new value initially.

alternatively you make it `undefined` (e.g. `this._gaps?.row`, `this._itemSize?.width`). but we end up with places like `get _gap` which then need to do `this._gaps?.row ?? whatAmI`

packages/labs/ssr/src/lib/render-module.ts
rictic: This still needs to be `any` I think, though we could do an immediate follow-up breaking change to the `3.0` branch

packages/labs/virtualizer/src/layouts/shared/SizeGapPaddingBaseLayout.ts
rictic: Hm, I think this is a question for @graynorton

rictic (changes requested):

PR #1520, opened 2021-01-06
[lit-next] Nested directives whose parent returns `noChange` are unaffected

### LINKED ISSUE #1519: [lit-next] Ensure nested directives handle `noChange` as expected
When a directive returns `noChange` any nested directive value it might otherwise return should be unchanged. Currently, the `resolveDirective` helper does not handle `noChange` and treats this as a value that should clear the nested directive. The nested directive gets removed and then re-created if it is subsequently returned. This prevents, for example, the `guard` directive from being able to guard a nested directive from running.

To address this, `resolveDirective` can special case `noChange` to bail early and simply not process any nested directives.

sorvell:
Fixes #1519. Now when a parent directive that previous returned another directive returns `noChange` the child directive is unaffected, meaning it does not update and is not disconnected or re-created.


packages/lit-html/src/test/directives/guard_test.ts
```diff
@@ -163,4 +164,39 @@ suite('guard', () => {
 
     assert.equal(callCount, 2);
   });
+
+  test.only('guards directive from running', () => {
```
kevinpschaaf, line 168: Remove the `only`

kevinpschaaf (approved):
LGTM, but need to remove `test.only` and make sure tests pass

packages/lit-html/src/test/directives/guard_test.ts
kevinpschaaf: Made a PR to catch these: https://github.com/Polymer/lit-html/pull/1521

justinfagnani (approved):

packages/lit-html/src/test/directives/guard_test.ts
sorvell: Fixed.

PR #1523, opened 2021-01-08
[lit-next] Add lit package

justinfagnani:
- This doesn't have any tests yet. Not 100% sure what to do there... the starter kits using this might be enough.
- I've moved the lit-min bundle to this package. I haven't yet made a more back-compat index.ts module in lit-element that exports the decorators.
- I didn't include the `/development/` folder, because there wouldn't be any differences. It's still built though, need to figure that out.
- README is very bare


packages/lit/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+export * from 'lit-element';
```
sorvell, line 15: Seems like `lit-element` should be sufficient here and it should export these others?

sorvell (approved):
Just a couple nits but basically seems good.

packages/lit/src/index.ts
justinfagnani: I did this to hoist those specifiers and shorten the loading waterfall. That way this module doesn't increase the max module depth at all and should be basically no-cost.

packages/lit/rollup.config.js
```diff
@@ -0,0 +1,57 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {litProdConfig} from '../../rollup-common.js';
+
+export default litProdConfig({
+  classPropertyPrefix: 'Φ',
+  entryPoints: [
```
kevinpschaaf, line 19: Do we need a rollup module build for `lit` at all? There should be no code here, and every re-export is to an out-of-package module. Is there any benefit? Would be nice to avoid that if possible.

If we need keep it for some reason, `classPropertyPrefix` should at least be something different than `lit-element`'s, to protect against some class source winding up in this module for some reason.

packages/lit/src/index.ts
kevinpschaaf: Did you confirm that there's not a rollup warning when building the bundle due to duplicate names being exported via the `*` (like we had with `_private`)? I'm not sure if rollup is smart enough to know that the duplicated names being exported ultimately resolve to the same symbol.

packages/lit/README.md
```diff
@@ -0,0 +1,16 @@
+# Lit 2.0
+
+Fast, lightweight web components
+
+## About this package
+
+The `lit` package contains everything needed to build Lit components with the LitElement base class, lit-html templates, and all first-party lit-html directives.
+
+Modules:
```
kevinpschaaf, line 9: You didn't mention:
* `directive-helpers.js`
* `disconnectable-directive.js`
* `hydrate-support.js`

packages/lit/README.md
```diff
@@ -0,0 +1,16 @@
...
+- `lit`: The main module exports the core pieces needed for component development: `LitElement`, `html`, `css`, and the most 
+- `lit/decorators.js`: Exports all the TypeScript/Babel decorators from one module.
+- `lit/decorators/...`: The `decorators/` folder contains a module for each decorator (`@customElement()`, `@property()`, etc.) for optimal pay-as-you-go module loading.
+- `lit/html.js`: Just the exports needed for standalone `lit-html` usage: `render()`, `html`, `svg`, etc.
+- `lit/static-html.js`: The lit-html `static.js` module
+- `lit/directives/...`: The `directives/` folder contains all of the first-party lit-html directives, like `repeat`, `classMap`, etc.
+- `lit/platform-support.js`: A module that connected Lit to the web components polyfills.
```
kevinpschaaf, line 16:
```suggestion
- `lit/platform-support.js`: A module that loads support for using Lit on older browsers in conjunction with the web components polyfills.
```

packages/lit/src/index.ts
kevinpschaaf: The plan also called for changing the `main` in `lit-element` to point to a new backward-compatible `index.js` which exports all of the stuff the old version did; in which case this line should be `export * from lit-element/lit-element.js` and not the main. Can you do that in this PR? If not, make an issue?

kevinpschaaf (changes requested):

packages/lit/README.md
justinfagnani: oh yeah, this reminds me: do we want to move `directive.ts`, etc., to live under `directives/`? That might be nicer when there's three top-level directive related files and the top level isn't lit-html specific anymore.

justinfagnani:
@kevinpschaaf 

> Did you confirm that there's not a rollup warning when building the bundle due to duplicate names being exported via the * (like we had with _private)? I'm not sure if rollup is smart enough to know that the duplicated names being exported ultimately resolve to the same symbol.

TypeScript will actually complain about this if you have a problem. It's sensitive to the order of exports (I need to look up why one order is ambiguous and the other is not) and I did change the order to match the export order of `lit-element.ts` itself and the complaint did pop up. The ambiguity is solvable by exporting `_$private` explicitly, so I added that.

packages/lit/rollup.config.js
justinfagnani: We could probably get away without most of the rollup build, but copying the standard setup we have for other projects is a really easy way to get whatever small minification we can and copy the files to their final location. I changed the classPropertyPrefix, but it's not used now. We do need some rollup step, or a copy step, to get the platform-support.js file.

I suppose we can change the TS config to not output to `/development/`. otoh, that's a transparent change too...

packages/lit/README.md
justinfagnani: updated

justinfagnani:
@kevinpschaaf @sorvell I added the new lit-element entrypoint. PTAL

packages/lit-element/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from '@lit/reactive-element';
+export * from 'lit-html';
+export * from './lit-element.js';
+
```
kevinpschaaf, line 4: Wasn't the idea to make this backward-compatible w.r.t. the exports? I guess that just means adding `export * from './decorators.js'`? That'll help with the number of g3 changes needed also.

packages/lit-element/src/index.ts
```diff
@@ -0,0 +1,10 @@
...
+// TODO: link to docs on the new site
+console.warn(
+  "The main 'lit-element' package is deprecated. " +
+    "Please import from the 'lit' package: 'lit' and 'lit/decorators.ts' " +
+    "or import from 'lit-element/lit-element.ts'."
```
kevinpschaaf, line 9:
```suggestion
  "The main 'lit-element' module entrypoint is deprecated. " +
    "Please update your imports to the 'lit' package: 'lit' and 'lit/decorators.js' ".
```
Don't want it to read like `LitElement` is deprecated.

kevinpschaaf (changes requested):

packages/lit/README.md
justinfagnani: @sorvell any thoughts?

packages/lit/README.md
```diff
@@ -0,0 +1,21 @@
...
+Modules:
+
+- `lit`: The main module exports the core pieces needed for component development: `LitElement`, `html`, `css`, and the most
+- `lit/decorators.js`: Exports all the TypeScript/Babel decorators from one module.
+- `lit/decorators/...`: The `decorators/` folder contains a module for each decorator (`@customElement()`, `@property()`, etc.) for optimal pay-as-you-go module loading.
+- `lit/html.js`: Just the exports needed for standalone `lit-html` usage: `render()`, `html`, `svg`, etc.
+- `lit/static-html.js`: The lit-html `static.js` module
+- `lit/directives.ts`: Contains the `Directive` base class for implementing directives.
+- `lit/directive-helpers.ts`: Optional helper utilities for impementing directives.
+- `lit/disconnectable-directive.ts`: A directive base class that supports disconnection and reconnection.
```
kevinpschaaf, line 18: Minor fixes: List of files should all probably be `.js`. And it's `directive.js` (not plural):
```suggestion
- `lit/directive.js`: Contains the `Directive` base class for implementing directives.
- `lit/directive-helpers.js`: Optional helper utilities for impementing directives.
- `lit/disconnectable-directive.js`: A directive base class that supports disconnection and reconnection.
```

packages/lit-element/rollup.config.js
```diff
@@ -35,10 +36,5 @@ export default litProdConfig({
     {
       file: 'platform-support',
     },
-    {
-      file: 'lit-element',
-      output: 'lit.min',
-      name: 'Lit',
```
kevinpschaaf, line 41: You moved the `lit.min.js` bundle from `lit-element` to `lit`, but the bundle tests are still in `lit-element`, which is the cause of the CI failure AFAICT?

kevinpschaaf (changes requested):

packages/lit-element/rollup.config.js
justinfagnani: fixed

packages/lit/README.md
justinfagnani: fixed

kevinpschaaf (approved):
LGTM, I concur the failing sauce issue is the Safari flake being investigated in #1550.

```
Browser tests did not start after 120000ms You can increase this timeout with the testsStartTimeout option. Check the browser logs or open the browser in debug mode for more information.  (failed on macOS Safari 13.1)
```

PR #1524, opened 2021-01-09
[lit-next] Fix failing IE tests

kevinpschaaf:
Always inject wcjs polyfill, unless <meta> opt-out is found. All .html tests use this mechanism to manually load the polyfills; avoids double polyfill loading, which was leading to errors on IE.


sorvell (approved):
LGTM, thanks for figuring this out!

justinfagnani:
Nice job fixing this!

PR #1525, opened 2021-01-11
Optionally enables support for `ShadyDOM.noPatch` setting

sorvell:
Enables support for the `ShadyDOM.noPatch` setting in ReactiveElement, LitElement, and lit-html. This support is off by default in production builds. It is on and tested only in development builds. When the `noPatch` setting is in use:

* All relevant lit-html DOM API usage is wrapped so that ShadyDOM perform the proper operation.
* The ReactiveElement prototype is made ShadyDOM `noPatch` aware so that it always uses ShadyDOM to perform DOM operations.


packages/reactive-element/src/platform-support.ts
```diff
@@ -72,6 +72,16 @@ interface PatchableReactiveElement extends HTMLElement {
 
   const elementProto = ReactiveElement.prototype;
 
+  // In noPatch mode, patch the ReactiveElement prototype so that no
+  // ReactiveElements must be wrapped.
+  if (
+    window.ShadyDOM &&
+    window.ShadyDOM.inUse &&
+    window.ShadyDOM.noPatch === true
```
kevinpschaaf, line 80: Calculate this condition once?

kevinpschaaf (commented):
Small nit, and suggest adding some directive tests that move nodes in/out of DOM (e.g. `repeat`, `cache`) in known areas where the polyfill gets involved, i.e. in/out of a host of a shadow root w/slot, in/out of the top-level of a shadow root, etc.

packages/lit-html/src/platform-support.ts
```diff
@@ -250,3 +259,8 @@ const scopeCssStore: Map<string, string[]> = new Map();
     return template;
   };
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-misused-new
+(globalThis as any)[
```
justinfagnani, line 264: Is this call eliminated in the prod build? You might need:

```ts
ENABLE_SHADYDOM_NOPATCH && (globalThis....
```

justinfagnani (approved):
Assuming the `wrap()` calls are completely eliminated in prod builds, LGTM.

packages/reactive-element/src/platform-support.ts
sorvell: That code is run 1x only.

packages/lit-html/src/platform-support.ts
sorvell: Done.

packages/reactive-element/src/platform-support.ts
kevinpschaaf: Ah, my bad.

kevinpschaaf (approved):
LGTM!

PR #1528, opened 2021-01-12
[lit-next] remove demos from Lit/ReactiveElement

sorvell:
These scenarios are covered by tests and examples are being moved to the Lit docs/site. Tests are also added for the Lit bundle which was not previously tested.


packages/lit-element/src/test/platform-support/lit-element-bundle_test.html
```diff
@@ -0,0 +1,178 @@
...
+      } from '../../../../tests/node_modules/@web/test-runner-mocha/dist/standalone.js';
+      import {assert} from '../../../node_modules/@esm-bundle/chai/esm/chai.js';
+      import {
+        getComputedStyleValue,
+        generateElementName,
+        nextFrame,
+      } from '../test-helpers.js';
+      // import { mocha, sessionFinished, sessionFailed } from '@web/test-runner-mocha';
+
+      (async () => {
```
kevinpschaaf, line 27: Doesn't seem like the async iffe is required?

kevinpschaaf (commented):
Code LGTM, just one nit. Need to figure out why the tests are failing to load the minified bundle

packages/lit-element/src/test/platform-support/lit-element-bundle_test.html
sorvell: Done.

packages/lit-element/package.json
```diff
@@ -35,7 +35,7 @@
     }
   },
   "scripts": {
-    "build": "npm run clean && tsc && rollup -c",
+    "build": "npm run clean && tsc && rollup -c && npm run build:bundle",
```
kevinpschaaf, line 38: I'm confused why there's a separate build for the bundle... it looks like you already added ability to add a list of `bundled` entrypoints in `litProdConfig` (`platform-support` uses this); is there a reason you didn't just add the build for `lit.min` there? Speed of `build:watch`?

kevinpschaaf: Actually, it looks like this may be the cause of the platform-support sauce test failures also. The standalone use of `litMonoBundleConfig` does not get the `nameCache` treatment, so `_$renderOptions` isn't being renamed to the stable property name, I believe.

sorvell: Fixed.

kevinpschaaf (approved):
LGTM

justinfagnani (approved):

PR #1530, opened 2021-01-12
[lit-next] Add TypeScript project references

justinfagnani:
This adds the `composite` flag, and a `references` section to every tsconfig (except localize, as it's not on next yet), and the `--build` flag to `tsc` in build scripts. This has a number of effects:

- A project with references will build its referenced projects first. So when you build lit-ssr, you're now sure that lit-html and lit-element are built.
- Composite projects are built incrementally, including referenced projects
- Referenced projects are watched in watch mode. `npm run build:watch` in lit-ssr will watch lit-html, etc.

So this should make development a bit nicer. It won't immediately speed up CI builds yet, but I think we could see a small speed up if we had one tsc process on CI - say by making a root tsconfig that referenced all the other projects.

There is a potential downside I want to highlight: `tsc --build` implies `noEmitOnError`. See: https://www.typescriptlang.org/docs/handbook/project-references.html#caveats This has bitten me a couple of times with `--watch` and auto-reloading with WDS where even a warning, like an unused import or variable, prevent emit and I didn't see the change I made take effect. To me the benefits of watching referenced projects is worth it, but we could offer another script if that's a problem.


kevinpschaaf (approved):
LGTM. This'll be nice to just run one command to keep the whole monorepo built.

Would be really nice if there was a way to trigger the associated rollup build each time a ts watcher ran... Mostly for lit-ssr, which imports the prod builds, so I always have to start all the rollup watchers as well. Although I need to experiment to see if the export map conditions for switching to `dev` in `node` will just solve that. We use custom resolution in `importModule` I think, which should ideally use the same condition `node` is using; I guess that means just needs to use newer version of the node resolve library that supports conditions and pass it to that or something...? Will experiment next time I'm working in SSR.

PR #1532, opened 2021-01-13
[labs/react] Add useController() hook

### LINKED ISSUE #1531: [labs] Add reactive controller -> hooks adapter to labs
Some subset of reactive controllers will depend only on the generic ReactiveControllerHost interface and not specifics of that host being a LitElement. These types of controllers can probably be wrapped into React hooks, making them a lower-level primitive for code sharing across frameworks and web components.

To make a hook wrapper we need to create and store a controller instance with `useState`, trigger the controller lifecycle via the built-in React hooks, and possibly allow mapping any OOP (property/method) APIs on the controller to hook function arguments and return values for more idiomatic hooks.

justinfagnani:
Fixes #1531


sorvell:
@justinfagnani Can you rebase this against the current lit-next? That'll make it easier to review.

justinfagnani:
@sorvell this is all rebased and cleaned up now!

packages/labs/react/src/use-controller.ts
```diff
@@ -0,0 +1,160 @@
...
+  React: React,
+  createController: (host: ReactiveControllerHost) => C
+): C => {
+  const {useState, useLayoutEffect} = React;
+
+  // State to force updates of the React component
+  const [kickCount, kick] = useState(0);
+
+  // Create and store the controller instance. We use useState() instead of
+  // useMemo() because React does not gaurentee that it will preserve values
```
sorvell, line 129:
```suggestion
  // useMemo() because React does not guarantee that it will preserve values
```

packages/labs/react/src/use-controller.ts
```diff
@@ -0,0 +1,160 @@
...
+    const host = new ReactControllerHost<C>(kickCount, kick);
+    const controller = createController(host);
+    host._primaryController = controller;
+    host._connected();
+    return host;
+  });
+
+  host._updatePending = true;
+
+  // We use useLayoutEffect because we need update() called synchronously
```
sorvell, line 147:
```suggestion
  // We use useLayoutEffect because we need updated() called synchronously
```

sorvell (changes requested):

sorvell (approved):

PR #1536, opened 2021-01-14
[lit-html] Add support for re-parenting Disconnectables. Fixes #1535

### LINKED ISSUE #1535: [lit-next] `cache` fails to remove DOM if it's the last node in a container
If the cached part is the last node in a container 

```
html`<div>${cache(v)}</div>`
```

then if the part DOM should be removed, it is not. This is because we don't correctly detect a "move to a new parent." To fix this, we could add a parent check in the move code. However, this would fail because when the part is moved, the parent is not updated. 

To address this, if the part is moved (1) the parent chain needs to be updated and (2) the implications for the "connection state" should be separate from that. This suggests that we should not have `connected`/`reconnected` but instead `activate`/`deactivate` and the user can just decide what that state is. So, for example, in `cache`, the container cache part could be set to be inactive.

kevinpschaaf:
Factors `addDisconnectableToParent` out of the `DisconnectableDirective` constructor, and adds a new `_$reparentDisconnectables` methods patched onto `ChildPart` which is called when moving a part to a new container part via `insertPart` to unlink the part from its previous parent and link it into its new parent.

Fixes #1535


packages/lit-html/src/directive-helpers.ts
```diff
@@ -121,7 +121,12 @@ export const insertPart = (
     );
   } else {
     const endNode = wrap(part._$endNode!).nextSibling;
-    if (endNode !== refNode) {
+    const parentChanged = part._$parent !== containerPart;
+    if (parentChanged) {
+      part._$reparentDisconnectables?.(containerPart);
```
sorvell, line 126: Any reason this is not the more straightforward `_$setParent` that sets the part's parent too?

packages/lit-html/src/directives/cache.ts
```diff
@@ -94,6 +95,7 @@ export const cache = directive(
           // Move cached part back into DOM
           clearPart(containerPart);
           insertPart(containerPart, undefined, cachedPart);
+          cachedPart.setConnected(true);
```
sorvell, line 98: Does it matter if this is after or before the next line putting the `cachedPart` into the `committedValue` of the `containerPart`?

justinfagnani (approved):

packages/lit-html/src/directive-helpers.ts
```diff
@@ -121,7 +121,12 @@ export const insertPart = (
...
+      part._$parent = containerPart;
```
sorvell, line 127: Setting `_$parent` appears to be unnecessary as it's done in `_$reparentDisconnectables`?

sorvell (changes requested):

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: `_$reparentDisconnectables` only exists if there is a disconnectable somewhere in the part, as it's a patched-on API for pay-for-play code economy. As such, we can't rely on that being there to update `_$parent`, so the line below needs to do that unconditionally.

packages/lit-html/src/directives/cache.ts
kevinpschaaf: Hmm, yeah it might be better to do it after, since it runs user code...? Will make that change.

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: Mentioned above, because `_$reparentDisconnectables` is patched in, we can't rely on that to set `_$parent`.

`_$reparentDisconnectables` does set `_$parent` in between unlinking/re-linking because the API s were set to work off the current state of the part; I tried refactoring `removeDisconnectableFromParent` and `addDisconnectableToParent` but it felt weird to ignore the initial part's `_$parent` state and use an argument, but use subsequent part's `_$parent` as we climb up the tree.

I'll add a comment for now; lmk if you think we should refine that.

kevinpschaaf:
The failing IE tests is a result of https://github.com/Polymer/lit-html/issues/1537 and should be resolved by https://github.com/Polymer/lit-html/pull/1542.

packages/lit-html/src/directives/cache.ts
kevinpschaaf: Done.

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: Done.

sorvell (approved):
LGTM

kevinpschaaf:
Confirmed Sauce Safari timeouts are related to https://github.com/Polymer/lit-html/issues/1550; merging ahead with those outstanding.

PR #1542, opened 2021-01-15
Fix ElementPart when IE reorders attrs. Fixes #1537

### LINKED ISSUE #1537: [lit-next] Element bindings can fail on IE11
This construction fails:

```
html`<div slot=${slot} ${ref(myRef)}></div>`
```

but this is ok:

```
html`<div ${ref(myRef)} slot=${slot}></div>`
```

kevinpschaaf:
Fix `ElementPart` when IE reorders attributes. The `attrNames` array now holds `undefined` for `ElementPart`s, and uses the same logic as `AttributePart` to ensure parts are created in the original order, despite bound attributes possibly being re-ordered.

Also adds SSR support for `ElementPart`, and adds belated SSR adds tests for `ElementPart`, `DisconnectableDirective`, and nested directives also.

Uncompressed code size when up a tad but gzip/brotli went down:

```
Name                            Size       Minified   Gzipped    Brotli    
---------------------------------------------------------------------------
lit-html.js                     6.086 KB   6.078 KB   2.759 KB   2.501 KB
lit-html.js                     6.101 KB   6.093 KB   2.758 KB   2.488 KB  
```

Fixes #1537

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -365,7 +365,7 @@ export interface DirectiveParent {
 const getTemplateHtml = (
   strings: TemplateStringsArray,
   type: ResultType
-): [string, string[]] => {
+): [string, Array<string|undefined>] => {
   // Insert makers into the template HTML to represent the position of
   // bindings. The following code scans the template strings to determine the
   // syntactic position of the bindings. They can be in text position, where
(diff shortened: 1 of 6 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -491,7 +494,7 @@ const getTemplateHtml = (
           s.slice(0, attrNameEndIndex) +
             boundAttributeSuffix +
             s.slice(attrNameEndIndex)) + marker
-        : s + marker + (attrNameEndIndex === -2 ? `:${i}` : '');
```
kevinpschaaf, line 494: I removed the `:` here to economize on a couple bytes since I knew I was adding a bit... Don't think it's required.

justinfagnani (approved):

PR #1543, opened 2021-01-15
WIP: feat: add compatibility with scoped registries

manolakis:
When a TemplateInstance is cloned, we must import the template content not only from the document but the ShadowRoot of the components, so in this way the new nodes are going to be created with the correct scope and registry.

I didn't add automatic test because I'm not sure how I could add it using wct and the scoped registries polyfill. However I did a manual test with a pet project using open-wc scoped-elements, updating the mixin to use the scoped registries, and it worked fine 👍

PokeApp using scoped elements before applying the scoped registries polyfill
https://user-images.githubusercontent.com/329771/104722801-aa2b2b00-572e-11eb-9f1d-5fb1519c67be.mp4

PokeApp using scoped elements after applying the scoped registries polyfill
https://user-images.githubusercontent.com/329771/104723197-30477180-572f-11eb-8678-2917a5c2feef.mp4

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -25,6 +25,14 @@ export const isCEPolyfill = typeof window !== 'undefined' &&
         undefined;
 
 /**
+ * True if scoped registries are supported
+ */
+export const scopedRegistriesSupported = typeof ShadowRoot !== 'undefined' &&
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
+    ShadowRoot.prototype.importNode !== undefined;
+
+/**
  * Reparents nodes, starting from `start` (inclusive) to `end` (exclusive),
  * into another container (could be the same container), before `before`. If
  * `before` is null, it appends the nodes to the container.
--- a/src/lib/parts.ts
+++ b/src/lib/parts.ts
@@ -13,7 +13,7 @@
  */
 
 import {isDirective} from './directive.js';
-import {removeNodes} from './dom.js';
+import {removeNodes, scopedRegistriesSupported} from './dom.js';
 import {noChange, nothing, Part} from './part.js';
 import {RenderOptions} from './render-options.js';
 import {TemplateInstance} from './template-instance.js';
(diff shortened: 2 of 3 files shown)


justinfagnani:
Hey @manolakis thanks for prototyping this!

We're working on a new major version of lit-html in the lit-next branch. We're trying to release that somewhat soon, so I'd rather see this done against that branch. Also, we're very, very sensitive to code size, and since the scoped element registry proposal isn't quite finalized yet, I'd probably ask for this to be behind a compile-time flag like we use for a few other features in lit-next.

I also want to make sure that this feature integrates with how the polyfills and noPatch support work in lit-next, so cc'ing @sorvell and @kevinpschaaf. I'm not sure if the scoped CE polyfill that layers on native CEs works with the CE polyfill or SD w/ noPatch.

kevinpschaaf:
@manolakis Awesome pushing this forward!

Just had a chat with @justinfagnani, and given that the feature is still speculative and subject to change, I think we can unblock experimenting with scoped registries in the following way for `lit-next`, while incurring the least runtime/payload cost and avoiding the risk of breaking changes to the core lib if the platform feature evolves:

1. Add a `cloneRoot` (naming?) field to `RenderOptions`, which is used in `_clone`, with fallback to document; i.e. 
https://github.com/Polymer/lit-html/blob/9xa8715fbed75c5ea7432bdc66ff63d6c95a6e012/packages/lit-html/src/lit-html.ts#L725 
would change to
   ```ts
    const fragment = (options.cloneRoot ?? d).importNode(content, true); 
    ```
2. Makes `_$renderOptions` in `LitElement` public `renderOptions` (I guess we hadn't had a reason to make it public thus far)
3. Add a `UseScopedRegistry` mixin for `LitElement` to [labs](https://github.com/Polymer/lit-html/tree/lit-next/packages/labs/src) that does a couple things:
   * Creates a scoped registry (maybe from a `static scopedElements` field?) and puts it in `shadowRootOptions`
   * Overrides `createRenderRoot` to set the `this.renderOptions.cloneRoot = this.shadowRoot`

Are you interested in moving that forward in a PR?

kevinpschaaf:
@justinfagnani The scoped CE polyfill right now doesn't integrate well with the standard CE polyfill or ShadyCSS; those are todo's on that polyfill. So another reason for this to be as much of a `labs` feature as we can make it.

manolakis:
@kevinpschaaf makes totally sense! In the app sample I used an updated version of the `ScopedElementsMixin` which creates the registry in this way https://github.com/open-wc/open-wc/pull/2049/files#diff-b8ca10750399bc6264c9b8232857a280ecf4d7e35423121d5d84f313777b2e98R10, so I think is really aligned with your suggestion. I can create the `UseScopedRegistry` for labs also.

I'll send the PR.

Thanks!

PR #1545, opened 2021-01-16
[lit-next] Adds `removeController`

### LINKED ISSUE #1539: [reactive-element] Add a `removeController` to `ReactiveElement`
## Description
This is needed for directives that are controllers. Ideally, they should remove themselves when disconnected.

## Acceptance criteria
PR landed.

sorvell:
Fixes #1539. Note, please review with `w=1` option to ignore whitespace changes.

* added to `ReactiveController` interface
* implemented in `ReactiveElement`
* also adds a test to `LitElement` for a controller directive

### DIFF
--- a/packages/lit-element/src/test/lit-element_test.ts
+++ b/packages/lit-element/src/test/lit-element_test.ts
@@ -17,6 +17,8 @@ import {
   LitElement,
   UpdatingElement,
   ReactiveElement,
+  Part,
+  nothing,
 } from '../lit-element.js';
 import {
   directive,
@@ -31,6 +33,7 @@ import {
 import {assert} from '@esm-bundle/chai';
 
 import {createRef, ref} from 'lit-html/directives/ref.js';
+import {ReactiveController} from '@lit/reactive-element';
 
 (canTestLitElement ? suite : suite.skip)('LitElement', () => {
   let container: HTMLElement;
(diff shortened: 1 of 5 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -668,6 +668,12 @@ export abstract class ReactiveElement
     (this.__controllers ??= []).push(controller);
   }
 
+  removeController(controller: ReactiveController) {
+    if (this.__controllers !== undefined) {
+      this.__controllers = this.__controllers?.filter((c) => c !== controller);
+    }
```
justinfagnani, line 674: This kind of makes me want to switch back to a Set 🤔

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts
justinfagnani, line 674: Check this out:

```suggestion
    this.__controllers?.splice(this.__controllers?.indexOf(c) >>> 0, 1);
```

Usually we should expect that add/remove calls are balanced, so `indexOf(c)` will have the correct index to remove. If for some reason `c` isn't in the list, `indexOf(c)` will return `-1`, and `splice()` would remove the last item, but `-1 >>> 0` strips the sign bit and returns `0xFFFFFFFF`. `splice()` doesn't remove anything if the first arg is greater than the length of the array, so that's a no op.

justinfagnani: On the current version you can also remove the optional chaining.

packages/lit-element/src/test/lit-element_test.ts
```diff
@@ -599,4 +604,108 @@ import {createRef, ref} from 'lit-html/directives/ref.js';
     assert.equal(host.elRef.value, host.child.trueDiv);
     assert.equal(host.count, 3);
   });
+
+  test('directive as controller can be added/removed via connect/disconnect', async () => {
+    const log: string[] = [];
+    const controllerDirective = directive(
+      class extends DisconnectableDirective {
```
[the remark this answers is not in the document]
sorvell: https://github.com/Polymer/lit-html/issues/1546

packages/reactive-element/src/reactive-element.ts
sorvell: Iterating the array is slightly faster. Let's leave it for now.

sorvell: Nifty, changed.

kevinpschaaf (approved):
LGTM, minor nit

sorvell:
Note, merging even though Safari 13 tests are failing on Sauce Labs since they pass locally. This issue is being separately investigated.

PR #1547, opened 2021-01-17
[lit-next] Add compatibility with scoped registries

manolakis:
Add compatibility with scoped registries using a mixin.

This PR follows the conversation initiated [in PR 1543](https://github.com/Polymer/lit-html/pull/1543)

# What I did

1. Add a `shadowRoot` field to `RenderOptions`, which is used in `_clone`
2. Makes `_$renderOptions` public as `renderOptions`
3. Add a `UseScopedRegistry` mixin for `LitElement` to labs

Note: ~I just added the polyfill for testing purposes. It should be removed once it can be retrieved as an npm package.~ Polyfill retrieved from git repo temporally.

@justinfagnani @kevinpschaaf


packages/labs/polyfills/scoped-registry.js
```diff
@@ -0,0 +1,330 @@
...
+// > Another option for looking up registries is to store an element's
+// > originating registry with the element. The Chrome DOM team was concerned
+// > about the small additional memory overhead on all elements. Looking up the
+// > root avoids this.
+const scopeForElement = new WeakMap();
+
+// Constructable CE registry class, which uses the native CE registry to
+// register stand-in elements that can delegate out to CE classes registered
+// in scoped registries
+window.CustomElementRegistry = class {
```
justinfagnani, line 33: Shouldn't this be part of the web components polyfills? Is there something Lit specific about this polyfill?

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,60 @@
...
+     */
+    static get scopedElements(): ScopedElementsMap {
+      return {};
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
+      super(...args);
+
+      this.registry = new CustomElementRegistry();
```
justinfagnani, line 41: I think all instances of an element definition should share a single registry by default, especially seeing as how the `scopedElements` property is static.

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,60 @@
...
+export function UseScopedRegistry<SuperClass extends LitElementConstructor>(
+  superclass: SuperClass
+): SuperClass {
+  return class ScopedRegistryMixin extends superclass {
+    registry?: CustomElementRegistry;
+
+    /**
+     * Obtains the scoped elements definitions map
+     */
+    static get scopedElements(): ScopedElementsMap {
```
justinfagnani, line 33: I'd make this an optional property, and if it's not defined don't create a registry.

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      const {scopedElements} = this.constructor as typeof ScopedRegistryMixin;
+
+      Object.entries(scopedElements).forEach(([tagName, klass]) =>
+        this.registry?.define(tagName, klass)
+      );
+    }
+
+    createRenderRoot() {
+      this.renderOptions.shadowRoot = this.attachShadow({
```
justinfagnani, line 51: This should use `shadowRootOptions`

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,60 @@
...
+        mode: 'open',
+        // @ts-expect-error: customElements not yet in ShadowRootInit type
+        customElements: this.registry,
+      });
+
+      return this.renderOptions.shadowRoot;
```
justinfagnani, line 57: you can just put the `return` keyword in front of the assignment above

justinfagnani (commented):
Just some initial comments...

packages/labs/polyfills/scoped-registry.js
manolakis: Yes. I just added it for testing purposes as right now I can't retrieve it through npm

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: We could have lazy definitions later and we can't bind the registry later to the shadowRoot, don't you think?

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: And couldn't problems arise later with lazy definitions if we share the registry between different instances?

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Right. Thanks!

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: oh, I miss "by default". We could always declare a registry per instance if needed. Makes sense.

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Not sure about this. If it's not going to declare the registry, it doesn't make sense to use the mixin, no?

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: right, thanks

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
justinfagnani: I could see a custom base class add this mixin where not every element that uses it may use scopes.

manolakis: so in  case I want to add lazy definitions then I must override 'createRenderRoot` to add a registry per instance... makes sense. I didn't think about that. Thanks

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Done

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Done

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Done

packages/labs/src/scoped-registry/scoped-registry-mixin.ts
manolakis: Done

packages/lit-html/src/lit-html.ts
```diff
@@ -296,6 +296,11 @@ export interface RenderOptions {
    * A DOM node before which to render content in the container.
    */
   renderBefore?: ChildNode | null;
+  /**
+   * ShadowRoot which is going to clone and import the TemplateInstance in case Scoped Registries
+   * are available.
+   */
+  shadowRoot?: ShadowRoot;
```
kevinpschaaf, line 303: I'm wondering if we should make this a more general name, like `ownerRoot`, `cloneContext`, or something like that, so that it's not tied only to this use case (and also because this is a speculative API and it's possible `ShadowRoot` might never get the `importNode` API, and then the field would become a misnomer).

Another use case I'm thinking about https://github.com/Polymer/lit-element/issues/1139, which (despite how we go about fixing that issue), pointed out that it's not really possible to e.g. render nodes into an iframe such that they are owned by that iframe. If that ever came back up, this could also solve that issue, e.g. 
```js
render(template, iframe.contentWindow.document, {ownerRoot: iframe.contentWindow.document})`)
```
🤷  wdyt @justinfagnani

packages/labs/polyfills/scoped-registry.js
```diff
@@ -0,0 +1,330 @@
+/**
```
kevinpschaaf, line 1: We'll try and get a release of this (along with `template-shadowroot`, which is also needed for `lit-next` out soon, so that this doesn't have to live here long-term.

manolakis: Awesome! 🤗

packages/lit-html/src/lit-html.ts
manolakis: mmm, I was unaware about other use cases, that's why I named it shadowRoot, but makes sense to name it to something more general. If I had to choose between `ownerRoot` and `cloneContext` I prefer `ownerRoot` because IMHO `cloneContext` looks more like an action. I'll do the change.

manolakis: done

packages/labs/polyfills/scoped-registry.js
manolakis: I changed to retrieve the package from github. It can be changed later to retrieve it from npm once published

packages/labs/scoped-registry-mixin/README.md
```diff
@@ -0,0 +1,43 @@
+# @lit-labs/scoped-registry-mixin
+
+`UseScopedRegistry` mixin for `LitElement` that integrates with the speculative [Scoped CustomElementRegistry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry).
+
+🚨 **Warning: Scoped Custom Element Registries is a speculative browser API that has not shipped and may change.** This mixin is provided to evaluate the spec proposal and provide feedback. **Use this feature and the polyfill in production code at your own risk.**
```
justinfagnani, line 5: Rather than "speculative browser API" I'd say that it's a proposal.

```suggestion
🚨 **Warning: Scoped Custom Element Registries is a proposed browser API. It has not been finalized nor shipped in any browser.** This mixin is provided to evaluate the proposal and facilitate feedback. **Use this feature and the polyfill in production code at your own risk.**
```

packages/labs/scoped-registry-mixin/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Overview
+
+The [Scoped Custom Element Registries](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) WICG proposal introduces new APIs for scoping custom elements to ShadowRoots, such that the tagName-to-custom element class mapping is local per Shadow Root.
```
justinfagnani, line 9:
```suggestion
The [Scoped Custom Element Registries](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) WICG proposal introduces new APIs for scoping custom element definitions to shadow roots, such that the mapping of tag names to custom element class doesn't rely on a single global registry.

When new elements are created within a shadow root with a scoped custom element registry, the browser will use the scoped registry rather than the global registry to look up custom element definitions. When using a scoped element registry, all used custom element elements must be defined within that scope.

`UseScopedRegistry` adds the following features to LitElement:
1. Automatically creates a scoped CustomElementRegistry for the class
2. Provides sugar for defining custom elements in the scoped registry
3. Passes the registry to `attachShadow()` to enable scoping
4. Passes the shadow root to lit-html to create DOM within the scope
```

packages/labs/scoped-registry-mixin/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+A speculative polyfill exists at [Scoped CustomElementRegistry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry). This package introduces a `UseScopedRegistry` mixin that may be applied to `LitElement`, used in conjunction with this polyfill, to register local tag-class mappings for a given element.
```
justinfagnani, line 11:
```suggestion
A speculative [Scoped CustomElementRegistry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) is being developed. Because the proposal is not implemented natively, `UseScopedRegistry` requires the polyfill.
```

packages/labs/scoped-registry-mixin/README.md
```diff
@@ -0,0 +1,43 @@
...
+```js
+import {LitElement, html} from 'lit';
+import {UseScopedRegistry} from '@lit-labs/scoped-registry-mixin';
+
+import {SimpleGreeting} from './simple-greeting.js';
+
+class ScopedComponent extends UseScopedRegistry(LitElement) {
+  // Elements here will be registered against the tag names provided only
+  // in the shadow root for this element
+  static scopedElements = {
```
justinfagnani, line 28: I would consider renaming `scopedElements` to `elementDefinitions`. The word "scoped" doesn't actually appear in any of the API names for the proposal. It's just CE definitions applied to a shadow root.

packages/labs/scoped-registry-mixin/README.md
```diff
@@ -0,0 +1,43 @@
+# @lit-labs/scoped-registry-mixin
+
+`UseScopedRegistry` mixin for `LitElement` that integrates with the speculative [Scoped CustomElementRegistry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry).
```
justinfagnani, line 3: "Use" just sounds a lot like React hooks to me now. Is there another name that works?

justinfagnani (commented):
Comments on readme and names

packages/labs/scoped-registry-mixin/README.md
manolakis: what about `withScopedRegistry`?

packages/labs/scoped-registry-mixin/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "regen-package-lock": "rm -rf node_modules package-lock.json; npm install"
+  },
+  "author": "The Polymer Authors",
+  "devDependencies": {
+    "@esm-bundle/chai": "^4.1.5",
+    "@types/mocha": "^8.0.3",
+    "@types/chai": "^4.0.1",
+    "@types/trusted-types": "^1.0.1",
+    "@web/test-runner-mocha": "^0.3.5",
+    "@webcomponents/scoped-custom-element-registry": "git+https://github.com/webcomponents/polyfills.git",
```
bicknellr, line 51:
```suggestion
    "@webcomponents/scoped-custom-element-registry": "^0.0.1",
```

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
```diff
@@ -0,0 +1,113 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+// TODO(kschaaf): update to `@webcomponents/scoped-custom-element-registry` once
+// released, and update package.json to match
+import '@webcomponents/scoped-custom-element-registry/packages/scoped-custom-element-registry/scoped-custom-element-registry.js';
```
bicknellr, line 17:
```suggestion
import '@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js';
```

bicknellr (commented):
Hi, I just published `@webcomponents/scoped-custom-element-registry@0.0.1`. These changes ought to do it - please ping me if you run into any issues while switching over.

packages/labs/scoped-registry-mixin/README.md
justinfagnani: I prefer mixins to be named like classes with capitalized noun phrases, since it represents a type like a class. I'd suggest something like `ScopedRegistryHost`.

manolakis: oh, right, my mistake. I agree. I prefer capitalized names for mixins too.

`ScopedRegistryHost` looks good to me.

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
manolakis: done

packages/labs/scoped-registry-mixin/package.json
manolakis: done

packages/labs/scoped-registry-mixin/README.md
manolakis: done

packages/labs/scoped-registry-mixin/README.md
manolakis: done

packages/labs/scoped-registry-mixin/README.md
manolakis: done

packages/labs/scoped-registry-mixin/README.md
manolakis: done

packages/labs/scoped-registry-mixin/README.md
manolakis: done

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,54 @@
...
+    createRenderRoot() {
+      const constructor = this.constructor as typeof ScopedRegistryMixin;
+      // @ts-expect-error: customElements not yet in ShadowRootInit type
+      const {registry, scopedElements, shadowRootOptions} = constructor;
+
+      if (scopedElements && !registry) {
+        constructor.registry = new CustomElementRegistry();
+
+        Object.entries(scopedElements).forEach(([tagName, klass]) =>
+          constructor.registry?.define(tagName, klass)
```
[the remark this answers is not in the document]
manolakis: I couldn't make this change to work. I'm a noob on Typescript so probably I'm doing something wrong 🤔️

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,54 @@
...
+        );
+      }
+
+      // @ts-expect-error: importNode not yet in ShadowRoot
```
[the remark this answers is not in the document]
manolakis: I don't know how to fix a browser type 🤔️

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -0,0 +1,54 @@
...
+  return class ScopedRegistryMixin extends superclass {
+    /**
+     * Obtains the scoped elements definitions map
+     */
+    static scopedElements?: ScopedElementsMap;
+    static registry?: CustomElementRegistry;
+
+    createRenderRoot() {
+      const constructor = this.constructor as typeof ScopedRegistryMixin;
+      // @ts-expect-error: customElements not yet in ShadowRootInit type
```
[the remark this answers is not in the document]
manolakis: It's still complaining about the `shadowRootOptions`

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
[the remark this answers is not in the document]
justinfagnani: That's fine. Just replace the `?` with a `!` because you definitely assign to it just above.

justinfagnani (approved):

kevinpschaaf:
I went ahead and pushed a the TypeScript nits. I think we're good to merge now once CI passes. Thanks for all the work on this @manolakis!

manolakis:
It was my pleasure. Thanks to you! 🤗

kevinpschaaf:
Since the PR hasn't been running on Sauce, I was thinking we might need to exclude IE11 from the tests, since we haven't integrated the scoped registry polyfill with the custom elements polyfill. But... all the tests here passed when I ran it manually on sauce. We know things like styling, etc. won't work, but it's interesting that this much of it did. So seems ok to leave as-is.

PR #1549, opened 2021-01-18
[lit-next] Adds `willUpdate` to `lit-element-renderer`

sorvell:
The `willUpdate` method can be used to compute properties before rendering. It is required not to use DOM APIs.


packages/lit-ssr/src/lib/lit-element-renderer.ts
```diff
@@ -55,6 +55,11 @@ export class LitElementRenderer extends ElementRenderer {
       }
       yield '</style>';
     }
+    // Call LitElement's `willUpdate` method.
+    // Note, this method is required not to use DOM APIs.
+    const changedProperties = new Map();
```
kevinpschaaf, line 60: Should use the correct `changedProperties` map from the instance, as they keys identify properties that were changed. It seems likely that initialize-state-from-props patterned code will check `changeProps` for existence of a `prop` edge that might cause state to reset.

kevinpschaaf: In thinking about whether `_changedProperties` needs to be in `stableProperties`, earlier I said that it only needs to have a helper in `private-ssr-support.js` which gets compiled along with `LitElement` and exposes a non-mangled API, and so technically the property doesn't need to be stable for `lit-ssr` to access it.  However, we need to decide whether `lit-ssr` itself can rely on only ever loading one copy of the core libs, or whether the one it loads needs to be compatible with all versions. In which case any private symbols accessed in `private-ssr-support` do need to be stable (not all of `lit-html`'s are, currently).

Right now, `importModule` in `lit-ssr` does a pretty hardcore workaround of forcing all `lit-html` and `lit-element` packages to resolve to the one it depends on, but I'm not sure if we're going to be able to rely on that long-term, depending on whether we require the module sandbox for all use cases or not.

cc @justinfagnani for thoughts; this is something we need to get right or else risk non-version-compatibility down the road...

packages/lit-ssr/src/lib/lit-element-renderer.ts
```diff
@@ -55,6 +55,11 @@ export class LitElementRenderer extends ElementRenderer {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (this.element as any)?.willUpdate(changedProperties);
```
kevinpschaaf, line 62: Not totally sure it matters since `willUpdate` is not supposed to access DOM, but calling `willUpdate` here will go _after_ attributes are reflected on the server (whereas it goes before attributes are reflected on the client).

While we don't run the user `update` on the server, we do run `ReactiveElement.prototype.update` in order to reflect properties to attributes, and we do that in `connectedCallback` (see L32). It's done in `connectedCallback` and not `renderShadow` because `connectedCallback` is called before closing the tag (`>`), when we can still emit attributes, whereas `renderShadow` is called after the tag is closed when we're ready to emit child content.

TL;DR: seems like this should maybe just go in `connectedCallback` above, before `ReactiveElement.prototype.update()`?

packages/lit-ssr/src/test/lit/render-lit_test.ts
```diff
@@ -204,6 +204,16 @@ test('element with property', async (t: Test) => {
   );
 });
 
+test('element with `willUpdate`', async (t: Test) => {
```
kevinpschaaf, line 207: It would also be good to add an `integration` test to test `willUpdate` interaction with `hydration` and that the use cases work as expected. `LitElement` integration tests are here: https://github.com/Polymer/lit-html/blob/lit-next/packages/lit-ssr/src/test/integration/tests/basic.ts#L3972

For debugging, use `npm run test:integration -- --debug` or `Test: Integration (debug) (lit-ssr)` in the `Run` tab in VSCode:
image

kevinpschaaf (changes requested):

packages/lit-ssr/src/lib/lit-element-renderer.ts
sorvell: Made this a reserved property. It was previously private so it had to change.

packages/lit-ssr/src/lib/lit-element-renderer.ts
sorvell: Done.

packages/lit-ssr/src/test/lit/render-lit_test.ts
sorvell: Done.

packages/lit-ssr/src/lib/lit-element-renderer.ts
kevinpschaaf: Per https://github.com/Polymer/lit-html/issues/1554, we'll address the question of how version-safe private-ssr-support needs to be in that issue, so can revert or leave this change, either way.

kevinpschaaf (approved):
LGTM

sorvell:
Note, merging even though Safari 13 tests are failing on Sauce Labs since they pass locally. This issue is being separately investigated.

PR #1561, opened 2021-01-25
[lit-html] Renders `null`, `undefined`, and empty string the same as `nothing` in `ChildPart`

### LINKED ISSUE #1559: [lit-next] Consider making `null`, `undefined`, and empty string act the same as `nothing` in `ChildPart`
## Description
We should consider changing values that obviously mean "render nothing" to have the same behavior as explicit `nothing` for `ChildPart`. Currently, we must carefully document that `nothing` *must* be used to ensure that default `<slot>` fallback content be rendered (see [docs](https://lit-html.polymer-project.org/guide/writing-templates#nothing-and-the-slot-fallback-content)). If we change this behavior, we can remove this from the docs and prevent a potential usage footgun.

## Acceptance criteria
Decide if we should make this change and land PR if so.

sorvell:
Fixes #1559. When rendering into elements with Shadow DOM, this makes it harder to inadvertently prevent `<slot>` fallback content from rendering.


packages/lit-html/src/lit-html.ts
```diff
@@ -920,9 +920,17 @@ class ChildPartImpl {
   _$setValue(value: unknown, directiveParent: DirectiveParent = this): void {
     value = resolveDirective(this, value, directiveParent);
     if (isPrimitive(value)) {
-      if (value === nothing) {
+      // Non-rendering child values. It's important that these do not render
+      // empty text nodes to avoid issues with preventing default <slot>
+      // fallback content.
+      if (
+        value === nothing ||
```
justinfagnani, line 927: I'd do just `value === nothing || value == null`.

I'm a bit uneasy with this behavior for empty strings. If someone specifically put a string there, I'd thing we'd render the string, even if empty.

kevinpschaaf: I suppose a footgun to consider is that it's _shorter_ to just write 
```html
<ce-with-slot>${user ? user.name : ''}</ce-with-slot>
```
than any of the alternatives for the else case (optional chaining aside), and so I think that's still going to require the footnote in the docs that that wouldn't trigger the fallback content, which is the footnote Steve is trying to remove with this change.

packages/lit-html/src/lit-html.ts
```diff
@@ -920,9 +920,17 @@ class ChildPartImpl {
...
+        value === null ||
+        value === undefined ||
+        value === ''
+      ) {
         this._$clear();
-        this._$committedValue = nothing;
+        this._$committedValue = value;
```
kevinpschaaf, line 933: Sorta sucks that we don't have a dirty check for `nothing`; on a repeated render, it means we'll always do a `this._$startNode.nextSibling` DOM access for the part.

Worth adding the code? If so, leaving this as `nothing` and adding ` && this._ $committedValue !== nothing` to the conditions above would do it I think?

packages/lit-html/src/lit-html.ts
sorvell: Based on offline chat, we'll keep including empty string to preserve compatibility with SSR, but we'll add dirty checking so we don't always call clear.

packages/lit-html/src/lit-html.ts

justinfagnani (approved):

PR #1563, opened 2021-01-25
[labs] WIP: Add ResizeController to labs

justinfagnani:


justinfagnani:
Closing since we now have the labs/observers package. I think this ResizeController still has a unique feature in being able to observe individual child elements though.

PR #1565, opened 2021-01-26
[lit-html] Rename DisconnectableDirective to AsyncDirective

### LINKED ISSUE #1533: [lit-html] Rename `DisconnectableDirective` to `AsyncDirective`
## Description
`DisconnectableDirective` is long and hard to type, and there's the tying together of async and disconnect that might not be obvious to devs. `AsyncDirective` is shorter and it might be easier to explain that only `AsyncDirective`s get connected/disconnected signal, than to explain that if you do async work your directive needs to be disconnectable.

There are use cases for both extremes (async directives that don't need the disconnect callback, and directives that need disconnect that won't be async). But `AsyncDirective` seems to fit the 90% case, and it's pretty easy to explain that you get extra lifecycle with it.

## Acceptance criteria
Rename `DisconnectableDirective` to `AsyncDirective` throughout monorepo and docs.

kevinpschaaf:
Fixes #1533


sorvell (approved):
LGTM

justinfagnani (approved):

PR #1569, opened 2021-01-27
[tests] Bump testing dependencies

### LINKED ISSUE #1550: [lit-html] Sauce testing on Safari routinely flaky with `testsStartTimeout` error
# Description
Safari 13 often fails on CI (Sauce) with the following error (flaky):

```
Browser tests did not start after 120000ms You can increase this timeout with the testsStartTimeout option. Check the browser logs or open the browser in debug mode for more information.
```

Initial debugging indicates it is a failure that occurs when `startSession()` (called from `@web/test-runner-mocha/dist/standalone.js`) in an `html` test is not acknowledged by the test runner.

## Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [X] Safari 13
- [ ] Safari 10
- [ ] IE 11

kevinpschaaf:
Updates `@web/test-runner-saucelabs` (among other dependencies) to pull in https://github.com/modernweb-dev/web/pull/1232 (fix for https://github.com/modernweb-dev/web/issues/1216), which hopefully fixes flaky Safari tests noted in https://github.com/Polymer/lit-html/issues/1550 and https://github.com/modernweb-dev/web/pull/1258, which hopefully fixes IE11 flaky tests.

Fixes #1550.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -201,5 +201,4 @@ jobs:
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           BROWSERS: preset:sauce-ie11
           CONCURRENT_BROWSERS: 1
-          CONCURRENT_FRAMES: 1
         run: npm run test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -137,3 +137,69 @@ jobs:
           # shutting down on the Sauce VM from the previous run?).
           CONCURRENT_BROWSERS: 2
         run: npm run test
+
+  tests-sauce-ie11:
+    # We can't run Sauce tests on PRs from forked repos, since they don't have
+    # access to secrets.
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
+
+    # Sauce tests are still unreliable. Run them, but don't mark a commit as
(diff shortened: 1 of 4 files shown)


justinfagnani:
It looks like only the Edge tests are failing. I'd back that out, get everything green, merge, and add Edge later with fixes. We need to enable Edge 15 anyway.

.github/workflows/tests.yml
```diff
@@ -137,3 +137,68 @@ jobs:
...
+
+      - name: Build
+        run: npm run build
+
+      - name: Test
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+          BROWSERS: preset:sauce-ie11
+          CONCURRENT_BROWSERS: 1
```
justinfagnani, line 203: If we think the timeouts are websocket related and not specifically iframe related, want to bump this back up?

justinfagnani (approved):

.github/workflows/tests.yml
kevinpschaaf: That is just the IE11 job. The other job with the rest of the browsers is set to 2 (and has a note that increasing it made it flaky)

PR #1575, opened 2021-01-29
[lit-next] Fixes build issues

sorvell:
* Fixes invalid TS in localize
* Bumps react versions to avoid build issues
* Also changes the nuke script to not remove package-lock's.


justinfagnani (approved):
For regenerating package-locks we might still want a script to remove them.

PR #1579, opened 2021-01-30
[lit-next] Add tsconfig for dowleveling polyfill-support to ES5

kevinpschaaf:
Fixes https://github.com/Polymer/lit-html/issues/1564

### DIFF
--- a/packages/lit-starter-js/web-test-runner.config.js
+++ b/packages/lit-starter-js/web-test-runner.config.js
@@ -1,14 +1,76 @@
-import {playwrightLauncher} from '@web/test-runner-playwright';
 import {legacyPlugin} from '@web/dev-server-legacy';
+import {playwrightLauncher} from '@web/test-runner-playwright';
 
-let commandLineBrowsers;
-try {
-  commandLineBrowsers = process.env.BROWSERS?.split(',').map((b) =>
-    playwrightLauncher({product: b})
-  );
-} catch {
-  console.warn(`BROWSER ${process.env.BROWSERS} unknown; using defaults`);
-}
+// Uncomment for testing on Sauce Labs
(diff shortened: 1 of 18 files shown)


kevinpschaaf:
`lit`'s `polyfill-support` still results in some `const`'s that I don't expect. The other ones don't. Need to investigate.

packages/lit-starter-js/web-test-runner.config.js
```diff
@@ -33,6 +96,17 @@ export default {
     legacyPlugin({
       polyfills: {
         webcomponents: true,
+        // Inject lit's polyfill-support module into test files, which is required
+        // for interfacing with the webcomponents polyfills
+        custom: [
```
justinfagnani, line 101: wow, this is awesomely easy

justinfagnani (approved):

rollup-common.js
```diff
@@ -379,6 +379,9 @@ const litMonoBundleConfig = ({
         'const ENABLE_EXTRA_SECURITY_HOOKS = false',
       'const ENABLE_SHADYDOM_NOPATCH = true':
         'const ENABLE_SHADYDOM_NOPATCH = false',
+      // For downleveled ES5 build of polyfill-support
+      'var ENABLE_SHADYDOM_NOPATCH = true':
+        'const ENABLE_SHADYDOM_NOPATCH = false',
```
sorvell, line 384: `const` is not ES5. Is this an issue?

sorvell (changes requested):

rollup-common.js
kevinpschaaf: It wasn't since it gets DCE'ed and I had FUD about terser not DCE'ing with vars, but confirmed replacing to a var is fine, so I've updated to that.

sorvell (approved):
LGTM

PR #1592, opened 2021-02-03
chore: include package exports field

jorgecasar:
# What it does

- Define [Package entrypoints](https://nodejs.org/api/packages.html#packages_package_entry_points)

## Why

It will help bundlers to tree shaking the bundle and services like skypack.dev (@FredKSchott) can inspect all entry points to avoid serving some files as raw or helps to generate URLs correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "/lit-html.d.ts.map",
     "/lib/",
     "/directives/",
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
Closing, as we've added these on all Lit 2 packages. Thanks!

PR #1595, opened 2021-02-03
Fix for bindings before end of self-closing tags

### LINKED ISSUE #1594: [lit-html] Regression: event bindings right before self-closing tag ends are broken
Example

```ts
html`<input @click=${listener}/>`
```

We don't support this well for attribute/property bindings - they will include the `"/"` in the value - but this does work for events in 1.x. If we fix this binding position in general it would be a nice improvement.

justinfagnani:
Fixes #1594 and adds support for bindings like:

```ts
html`<input .value=${v}/>`
```

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -495,8 +495,10 @@ const getTemplateHtml = (
     //     position (attrNameEndIndex === -2), add a sequential suffix to
     //     generate a unique attribute name.
 
-    // detect a binding next to self-closing tag end
-    const end = strings[i + 1].startsWith('/>') ? ' ' : '';
+    // Detect a binding next to self-closing tag end and insert a space to
+    // separate the marker from the tag end:
+    const end =
+      regex === tagEndRegex && strings[i + 1].startsWith('/>') ? ' ' : '';
     html +=
       regex === textEndRegex
         ? s + nodeMarker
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -494,17 +494,24 @@ const getTemplateHtml = (
     //  4. We're somewhere else inside the tag. If we're in attribute name
     //     position (attrNameEndIndex === -2), add a sequential suffix to
     //     generate a unique attribute name.
+
+    // Detect a binding next to self-closing tag end and insert a space to
+    // separate the marker from the tag end:
+    const end =
+      regex === tagEndRegex && strings[i + 1].startsWith('/>') ? ' ' : '';
```
jridgewell, line 501: Nit: ? Could we simplify this and just always insert a space?

jridgewell (approved):

packages/lit-html/src/lit-html.ts
justinfagnani: We can't because the current regex is `tagEndRegex` for unquoted attributes, _and_ we also support multiple bindings in an unquoted attribute now, like `<div foo=${a}${b}></div>`. We can't insert a space after the first binding as that'll close the attribute early.

kevinpschaaf (approved):

PR #1598, opened 2021-02-04
A couple more compiler-compat fixes

rictic:


packages/lit-element/src/lit-element.ts
```diff
@@ -161,7 +161,7 @@ export class LitElement extends ReactiveElement {
 // DEV mode warnings
 if (DEV_MODE) {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  (LitElement as any).finalize = function () {
+  (LitElement as any)['finalize'] = function (this: typeof LitElement) {
```
justinfagnani, line 164: can you add a comment that this needs to stay as index access?

justinfagnani (approved):

PR #1599, opened 2021-02-04
[lit-ssr] Fix usage of non-stable properties in private-ssr-support

### LINKED ISSUE #1554: [lit-next] Decide on stable names convention for private ssr support
## Description
We need to decide which properties need stable names for private ssr support.

## Acceptance criteria
If changes are needed, land PR.

kevinpschaaf:
The main issue was that `patchDirectiveResolve` in `lit-html/private-ssr-support.ts` was patching `_resolve`, despite the fact that directives may possibly come from any version of the library.  So `_$resolve` is now in `stableProperties`.

However, it also pointed out that `lit-ssr` was directly importing `Directive` from `lit-html` and patching it, and relying on that to affect all directive instances. This is only a safe assumption if lit-ssr is only seeing Directives created with the exact same version of the library that it is importing. Although the `import-module.ts` does force module resolution for user code to use the same version, we don't necessarily want the VM-sandbox to be the _only_ environment where you can run SSR.

For this reason, this change makes lit-ssr no longer import & patch `Directive`, but instead patches `Directive` constructors in any `DirectiveResult` values that are being rendered by lit-ssr.

The PR also includes a change from using the individual `lit-html` and `lit-element` packages to using `lit`, to avoid all the warning spamming from using the deprecated `lit-element` entrypoint.

Fixes #1554

### DIFF
--- a/packages/lit-html/CHANGELOG.md
+++ b/packages/lit-html/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Changed
 
+- (Since 2.0.0-pre.4) Removes second `klass` argument from `isDirectiveResult` since it is generally not version-agnostic to test directives using `instanceof`. A new `classForDirectiveResult` helper is introduced, which allows for e.g. directive class branding checks instead.
 - (Since 2.0.0-pre.4) `DisconnectableDirective` was renamed to `AsyncDirective`, and its module name was renamed from `disconnectable-directive` to `async-directive`.
 - (Since 2.0.0-pre.4) Rendering `null`, `undefined`, or empty string in a `ChildPart` now has the same affect as rendering `nothing`: it does not produce an empty text node. When rendering into an element with Shadow DOM, this makes it harder to inadvertently prevent `<slot>` fallback content from rendering.
 - (Since 2.0.0-pre.4) `DisconnectableDirective`'s `disconnectedCallback` and `reconnectedCallback` were renamed to `disconnected` and `reconnected`.
--- a/packages/lit-html/CHANGELOG.md
+++ b/packages/lit-html/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Changed
 
-- (Since 2.0.0-pre.4) Removes second `klass` argument from `isDirectiveResult` since it is generally not version-agnostic to test directives using `instanceof`. A new `classForDirectiveResult` helper is introduced, which allows for e.g. directive class branding checks instead.
+- (Since 2.0.0-pre.4) Removes second `klass` argument from `isDirectiveResult` since it is generally not version-agnostic to test directives using `instanceof`. A new `getDirectiveClass` helper is introduced, which allows for directive class branding checks instead.
 - (Since 2.0.0-pre.4) `DisconnectableDirective` was renamed to `AsyncDirective`, and its module name was renamed from `disconnectable-directive` to `async-directive`.
 - (Since 2.0.0-pre.4) Rendering `null`, `undefined`, or empty string in a `ChildPart` now has the same affect as rendering `nothing`: it does not produce an empty text node. When rendering into an element with Shadow DOM, this makes it harder to inadvertently prevent `<slot>` fallback content from rendering.
 - (Since 2.0.0-pre.4) `DisconnectableDirective`'s `disconnectedCallback` and `reconnectedCallback` were renamed to `disconnected` and `reconnected`.
--- a/packages/lit-html/src/directive-helpers.ts
+++ b/packages/lit-html/src/directive-helpers.ts
@@ -67,13 +67,15 @@ export const isTemplateResult = (
 /**
  * Tests if a value is a DirectiveResult.
  */
-export const isDirectiveResult = (
-  value: unknown,
-  klass?: DirectiveClass
-): value is DirectiveResult =>
-  klass === undefined
-    ? (value as DirectiveResult)?._$litDirective$ !== undefined
-    : (value as DirectiveResult)?._$litDirective$ === klass;
+export const isDirectiveResult = (value: unknown): value is DirectiveResult =>
+  (value as DirectiveResult)?._$litDirective$ !== undefined;
+
+/**
+ * Retrieves the Directive class for a DirectiveResult
+ */
+export const classForDirectiveResult = (
+  value: unknown
+): DirectiveClass | undefined => (value as DirectiveResult)?._$litDirective$;
 
 /**
  * Tests whether a part has only a single-expression with no strings to
--- a/packages/lit-html/src/directive-helpers.ts
+++ b/packages/lit-html/src/directive-helpers.ts
@@ -73,9 +73,8 @@ export const isDirectiveResult = (value: unknown): value is DirectiveResult =>
 /**
  * Retrieves the Directive class for a DirectiveResult
  */
-export const classForDirectiveResult = (
-  value: unknown
-): DirectiveClass | undefined => (value as DirectiveResult)?._$litDirective$;
+export const getDirectiveClass = (value: unknown): DirectiveClass | undefined =>
+  (value as DirectiveResult)?._$litDirective$;
 
 /**
  * Tests whether a part has only a single-expression with no strings to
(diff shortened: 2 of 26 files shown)


packages/lit-ssr/package.json
```diff
@@ -58,6 +58,7 @@
     "koa": "^2.7.0",
     "koa-node-resolve": "^1.0.0-pre.5",
     "koa-static": "^5.0.0",
+    "lit": "^2.0.0-pre.1",
     "lit-element": "^3.0.0-pre.2",
     "lit-html": "^2.0.0-pre.4",
```
kevinpschaaf, line 63: Note, the only reason lit-ssr now still relies on `lit-element` and `lit-html` are to import their respective `private-ssr-support.js` packages, since we don't forward those out through `lit`. That seems ... ok?

justinfagnani: yeah

packages/lit-ssr/src/lib/render-lit-html.ts
```diff
@@ -15,11 +15,11 @@
  */
 
 // Type-only imports
-import {TemplateResult, ChildPart} from 'lit-html';
+import {TemplateResult, ChildPart} from 'lit';
```
justinfagnani, line 18: You can use `import type` now to clarify this

packages/lit-ssr/src/lib/render-lit-html.ts
```diff
@@ -53,30 +53,58 @@ import {
...
-      __attributeIndex
+/**
+ * Looks for values of type `DirectiveResult` and patches a `Directive` class's
+ * implementation of `_$resolve` to call `render` rather than `update`, per the
+ * SSR contract
+ */
+const patchIfDirective = (value: unknown) => {
+  const directiveCtor = (value as DirectiveResult)?._$litDirective$;
+  if (directiveCtor !== undefined) {
+    patchDirectiveResolve(
```
justinfagnani, line 75: It seems safer to me to subclass the directive to override resolve

packages/lit-ssr/src/lib/render-lit-html.ts
```diff
@@ -53,30 +53,58 @@ import {
...
-      this.render(...values),
-      this,
-      __attributeIndex
+/**
+ * Looks for values of type `DirectiveResult` and patches a `Directive` class's
+ * implementation of `_$resolve` to call `render` rather than `update`, per the
+ * SSR contract
+ */
+const patchIfDirective = (value: unknown) => {
+  const directiveCtor = (value as DirectiveResult)?._$litDirective$;
```
justinfagnani, line 73: do you have to check for `undefined` and `null` first? Might be better to check for an object anyway to narrow the set of types you'll do this lookup against.

packages/lit-ssr/src/lib/render-lit-html.ts
sorvell: Agree, if it's chill.

sorvell (changes requested):

packages/lit-ssr/src/lib/render-lit-html.ts
kevinpschaaf: Done

packages/lit-ssr/src/lib/render-lit-html.ts
kevinpschaaf: It's optionally-chained...

packages/lit-ssr/src/lib/render-lit-html.ts
kevinpschaaf: Done, changed the helper exported in `private-ssr-support` to be a mixin that overrides resolve to render, and then SSR uses a cache to avoid subclassing every DirectiveResult's directive anew.

I also relocated where `resolveDirective` is called on the return value of a directive (to support nested directives), since it was unsafely dereferencing possibly unstable class fields from `Part`. This has the benefit of also making the mixin override simpler.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -67,13 +67,15 @@ export const isTemplateResult = (
...
  * Tests if a value is a DirectiveResult.
  */
-export const isDirectiveResult = (
-  value: unknown,
-  klass?: DirectiveClass
-): value is DirectiveResult =>
-  klass === undefined
-    ? (value as DirectiveResult)?._$litDirective$ !== undefined
-    : (value as DirectiveResult)?._$litDirective$ === klass;
+export const isDirectiveResult = (value: unknown): value is DirectiveResult =>
```
kevinpschaaf, line 70: I'm not sure how useful this is if we add `classForDirectiveResult`... We don't use it at all now.

We could build a more formal branding mechanism in, like an optional `static directiveType = Symbol.for('RenderLight')`, and if you pass the 2nd class argument, it will compare the `directiveType`s for both? (and then get rid of `classForDirectiveResult` and go back to having just `isDirectiveResult`)?

packages/lit-ssr/src/lib/render-lit-html.ts
justinfagnani: oh... I missed that :)

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -33,12 +33,14 @@ export const _Σ = {
...
   HTML_RESULT: p._HTML_RESULT,
   getTemplateHtml: p._getTemplateHtml,
-  patchDirectiveResolve: (
-    directive: Directive,
-    fn: (this: Directive, values: unknown[]) => unknown
-  ) => {
-    directive._resolve = fn;
-  },
+  overrideDirectiveResolve: (
+    directiveClass: new (part: PartInfo) => Directive & {render(): unknown}
```
justinfagnani, line 37: we have the interface `DirectiveClass` in `directive.js`. Why do you need `render()` here?

packages/lit-ssr/src/lib/render-lit-html.ts
```diff
@@ -464,6 +490,7 @@ export function* renderValue(
   value: unknown,
   renderInfo: RenderInfo
 ): IterableIterator<string> {
+  patchIfDirective(value);
```
justinfagnani, line 493: Is this where nested directive results would be patched?

packages/lit-html/src/private-ssr-support.ts
kevinpschaaf: That was the only way I could make that mixin not complain about `directiveClass` not implementing the abstract render method.

The `Concrete<T>` thing doesn't work because it loses the "methodness" of `_$resolve`, and complains that you're assigning a function to a property in the subclass.

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: Ok, so will leave it

packages/lit-ssr/src/lib/render-lit-html.ts
kevinpschaaf: No, it's on the return value of `render` in the patch: https://github.com/Polymer/lit-html/pull/1599/commits/09d6df9a9563024d70a35ab6a9705df9f4e08dee#diff-b8cfd0fdb8410f04f051c052e6ce548e5403f407d6954d48df02b7c5bdb0e506R89

packages/lit-html/src/directive-helpers.ts
```diff
@@ -67,13 +67,15 @@ export const isTemplateResult = (
...
+  (value as DirectiveResult)?._$litDirective$ !== undefined;
+
+/**
+ * Retrieves the Directive class for a DirectiveResult
+ */
+export const classForDirectiveResult = (
```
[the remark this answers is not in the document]
kevinpschaaf: SGTM, done.

packages/lit-html/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Changed
 
+- (Since 2.0.0-pre.4) Removes second `klass` argument from `isDirectiveResult` since it is generally not version-agnostic to test directives using `instanceof`. A new `classForDirectiveResult` helper is introduced, which allows for e.g. directive class branding checks instead.
```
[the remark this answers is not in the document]
kevinpschaaf: Done

justinfagnani (approved):

sorvell (approved):
LGTM

PR #1600, opened 2021-02-04
Enable stripInternal in tsconfig

rictic:
This means that we won't emit declarations in .d.ts files for things annotated with @internal.

I believe this is desirable, to make it clearer to users which fields they can count on existing and which we're free to change.

This also aligns with the internal tsconfig.

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -216,7 +216,7 @@ export const _Φ = {
     name: string,
     value: string | null
   ) => {
-    el._$attributeToProperty(name, value);
+    (el as any)._$attributeToProperty(name, value);
   },
-  _$changedProperties: (el: LitElement) => el._$changedProperties,
+  _$changedProperties: (el: LitElement) => (el as any)._$changedProperties,
 };
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -216,7 +216,9 @@ export const _Φ = {
     name: string,
     value: string | null
   ) => {
+    // eslint-disable-next-line
     (el as any)._$attributeToProperty(name, value);
   },
+  // eslint-disable-next-line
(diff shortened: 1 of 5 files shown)


packages/lit-element/tsconfig.json
```diff
@@ -20,9 +24,19 @@
     "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
-    "importHelpers": true
+    "importHelpers": true,
+    "stripInternal": true
```
justinfagnani, line 28: TIL 🤯

packages/lit-element/tsconfig.json
```diff
@@ -3,7 +3,11 @@
     "composite": true,
     "target": "es2020",
     "module": "es2015",
-    "lib": ["es2020", "DOM", "DOM.Iterable"],
+    "lib": [
```
justinfagnani, line 6: did you run `npm run format`?

packages/lit-element/src/lit-element.ts
```diff
@@ -216,7 +216,9 @@ export const _Φ = {
     name: string,
     value: string | null
   ) => {
-    el._$attributeToProperty(name, value);
+    // eslint-disable-next-line
+    (el as any)._$attributeToProperty(name, value);
```
justinfagnani, line 220: It'd be nice to cast to an internal interface...

justinfagnani (approved):
I didn't even know about that setting. Cool!

Looks like you might have different formatter settings, btw.

packages/lit-element/tsconfig.json
rictic: I did not, looks like I've got the wrong json formatter configured, fixed.

packages/lit-element/src/lit-element.ts
rictic: Yeah, I think we'll need to export some non-`@internal` types from lit-html to get ssr compiling. Reverted the change to lit-html for now.

PR #1603, opened 2021-02-06
[lit-starter-ts/js] Fix .pretterignore syncing with lit-starter templates

kevinpschaaf:
Also changes the glob just to look in the top level of packages, so that it doesn't unnecessarily recurse into node_modules

### DIFF
--- a/.prettierignore-sync
+++ b/.prettierignore-sync
@@ -2,7 +2,10 @@
 .gitignore
 
 [relative]
-packages/**/{.gitignore,.prettierignore}
+packages/{*,labs/*}/{.gitignore,.prettierignore}
+# Starter templates specify the lint/format ignore in .eslintignore to stay DRY
+packages/{lit-starter-ts,lit-starter-js}/.eslintignore
+
 
 [inline]
 packages/localize/examples/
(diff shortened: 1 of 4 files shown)


.prettierignore-sync
```diff
@@ -2,7 +2,10 @@
 .gitignore
 
 [relative]
-packages/**/{.gitignore,.prettierignore}
+packages/{*,labs/*}/{.gitignore,.prettierignore}
+# Starter templates specify the lint/format ignore in .eslintignore to stay DRY
+packages/{lit-starter-ts,lit-starter-js}/.eslintignore
```
abdonrd, line 7: We should move this to the `.eslintignore-sync`, right?
The starters has `.eslintignore` and no `. prettierignore `.

abdonrd (changes requested):
Thanks @kevinpschaaf! 👏🏻

.prettierignore-sync
justinfagnani: The starter kits use their `.eslintignore` as the ignore file for Prettier. This just reflects that.

justinfagnani (approved):

PR #1611, opened 2021-02-12
[lit-html] Add asyncAppend and asyncReplace

### LINKED ISSUE #1610: [lit-html] Add asyncAppend and asyncReplace directives
## Description
Needed for feature parity with 1.x

## Acceptance criteria
1.x tests for asyncAppend and asyncReplace committed and passing

kevinpschaaf:
Port from 1.x, tests are wholly unchanged apart from adding test.

Also fixes `until` to be more idiomatic by using class fields for state (noticed when comparing code).

Fixes #1610

### DIFF
--- a/packages/lit-html/src/directives/async-append.ts
+++ b/packages/lit-html/src/directives/async-append.ts
@@ -40,7 +40,9 @@ class AsyncAppendDirective extends AsyncDirective {
     }
   }
 
-  render<T>(_value: AsyncIterable<T>, _mapper?: Mapper<T>) {
+  // @ts-expect-error value not used, but we want a nice parameter for docs
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  render<T>(value: AsyncIterable<T>, _mapper?: Mapper<T>) {
     return noChange;
   }
 
@@ -90,15 +92,15 @@ class AsyncAppendDirective extends AsyncDirective {
     }
   }
 
-  // Pause iteration while disconnected
   disconnected() {
+    // Pause iteration while disconnected
     this.reconnectPromise = new Promise(
       (resolve) => (this.reconnectResolver = resolve)
     );
   }
 
-  // Resume iteration when reconnected
   reconnected() {
+    // Resume iteration when reconnected
     this.reconnectPromise = undefined;
     this.reconnectResolver!();
   }
@@ -107,6 +109,7 @@ class AsyncAppendDirective extends AsyncDirective {
 /**
  * A directive that renders the items of an async iterable[1], appending new
  * values after previous values, similar to the built-in support for iterables.
+ * This directive is usable only in child expressions.
  *
  * Async iterables are objects with a [Symbol.asyncIterator] method, which
  * returns an iterator who's `next()` method returns a Promise. When a new
@@ -115,7 +118,7 @@ class AsyncAppendDirective extends AsyncDirective {
  * directive has been set on the Part, the iterable will no longer be listened
  * to and new values won't be written to the Part.
  *
- * [1]: https://github.com/tc39/proposal-async-iteration
+ * [1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of
  *
  * @param value An async iterable
  * @param mapper An optional function that maps from (value, index) to another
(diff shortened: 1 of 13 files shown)


packages/lit-html/src/directives/async-replace.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      // As a convenience, because functional-programming-style
+      // transforms of iterables and async iterables requires a library,
+      // we accept a mapper function. This is especially convenient for
+      // rendering a template for each item.
+      if (mapper !== undefined) {
+        v = mapper(v, i);
+      }
+
+      this.setValue(v);
```
kevinpschaaf, line 78: @justinfagnani The original 1.x implementation created a nested child part here, but it was sorta unclear why. I'm just using the parent part for replace. 🤷

packages/lit-html/src/directives/async-append.ts
```diff
@@ -0,0 +1,124 @@
...
+ * values after previous values, similar to the built-in support for iterables.
+ *
+ * Async iterables are objects with a [Symbol.asyncIterator] method, which
+ * returns an iterator who's `next()` method returns a Promise. When a new
+ * value is available, the Promise resolves and the value is appended to the
+ * Part controlled by the directive. If another value other than this
+ * directive has been set on the Part, the iterable will no longer be listened
+ * to and new values won't be written to the Part.
+ *
+ * [1]: https://github.com/tc39/proposal-async-iteration
```
justinfagnani, line 118: Maybe like to this now: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

Can't really find a better page though

packages/lit-html/src/directives/async-replace.ts
justinfagnani: I think this change makes asyncReplace compatible with attribute/property parts, yeah?

Maybe document/test that? It could be useful.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -0,0 +1,115 @@
...
+  reconnectPromise?: Promise<void>;
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    if (partInfo.type !== PartType.CHILD) {
+      throw new Error('asyncAppend can only be used in child expressions');
+    }
+  }
+
+  render<T>(_value: AsyncIterable<T>, _mapper?: Mapper<T>) {
```
justinfagnani, line 38: same comment here about the name `_value`

justinfagnani (approved):

packages/lit-html/src/directives/async-append.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (mapper !== undefined) {
+        v = mapper(v, i);
+      }
+      const newPart = insertPart(part);
+      setChildPartValue(newPart, v);
+      i++;
+    }
+  }
+
+  // Pause iteration while disconnected
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/src/directives/async-append.ts
```diff
@@ -0,0 +1,124 @@
...
+  reconnectPromise?: Promise<void>;
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    if (partInfo.type !== PartType.CHILD) {
+      throw new Error('asyncAppend can only be used in child expressions');
+    }
+  }
+
+  render<T>(_value: AsyncIterable<T>, _mapper?: Mapper<T>) {
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/src/directives/async-append.ts
kevinpschaaf: Done

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Yeah good point. Done.

I thought about ways it might make sense to have a switch on part type in `asyncAppend` to keep the current behavior for `ChildPart` and do something different for the other part types, but it doesn't really make too much sense I think. You could either concat as string (maybe good for attributes, but then you'd probably want a joiner), or else spread into a new array each time (maybe good for properties), but not a slam dunk and not worth going there now.

PR #1612, opened 2021-02-12
[lit-starter-*] Fix tests on macOS 11.2

justinfagnani:
Updates `@web/test-runner-playwright` to bump Playwright to a version that works on macOS 11.2.

Also updates a module specifier in lit-starter-ts that was failing to resolve.


kevinpschaaf (approved):
LGTM

PR #1616, opened 2021-02-16
[lit-html] Allow styleMap values to be nullable

matthias-ccri:
Fixes this error:
image

Here's the spec for [CSSStyleDeclaration.setProperty](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty). It appears to treat `null` and `''` (empty string) differently. Step 3 says an empty string value triggers `removeProperty`. Steps 6 says that after parsing the value, just return void if it's null (no `removeProperty`).

I'm not sure the reasoning behind making `null` a noop (if that's indeed the case). Of course it's a spec not a user guide.

In any case, the idea is that noops have no place in declarative code. To be declarative, the UI must match the declaration. So, if you declare the style as "empty" (null or undefined), then that style should become empty (unset).

This is the current behavior in plain JS, but typescript forbids `undefined`.

### DIFF
--- a/src/directives/style-map.ts
+++ b/src/directives/style-map.ts
@@ -15,7 +15,7 @@
 import {AttributePart, directive, Part, PropertyPart} from '../lit-html.js';
 
 export interface StyleInfo {
-  readonly [name: string]: string;
+  readonly [name: string]: string | undefined | null;
 }
 
 /**
--- a/src/directives/style-map.ts
+++ b/src/directives/style-map.ts
@@ -15,7 +15,7 @@
 import {AttributePart, directive, Part, PropertyPart} from '../lit-html.js';
 
 export interface StyleInfo {
-  readonly [name: string]: string | undefined | null;
+  readonly [name: string]: string|undefined|null;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


justinfagnani:
Interesting. I think this is fine and a good tightening of behavior. We need to have a test though.

src/directives/style-map.ts
```diff
@@ -82,7 +82,7 @@ export const styleMap = directive((styleInfo: StyleInfo) => (part: Part) => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       (style as any)[name] = styleInfo[name];
     } else {
-      style.setProperty(name, styleInfo[name]);
+      style.setProperty(name, styleInfo[name] || '');
```
justinfagnani, line 85: We generally avoid coercing operators. I'd use a ternary, and hoist `styleInfo[name]` since it would be used 3 times then:

```ts
    previousStyleProperties.add(name);
    const value = styleInfo[name];
    if (name.indexOf('-') === -1) {
      (style as any)[name] = value;
    } else {
      style.setProperty(name, value == null ? '' : value);
```

justinfagnani (changes requested):

matthias-ccri:
Fixed with https://github.com/lit/lit/pull/1665

PR #1624, opened 2021-02-25
[lit-html] Support for <host> wrapper

kevinpschaaf:
Initial swag at supporting an optional `<host>` root tag in the template for declarative rendering of host attributes & expressions.

Example:
```ts
@customElement('my-element')
  @property()
  mode = 'a';
  render() {
    html`
      <host tabindex="1" role=${this.mode === 'a' ? 'button' : 'tab'}>
        <div>...</div>
      </host>
    `;
  }
}
```
Renders:
```html
<my-element tabindex="1" role="button">
  #shadow-root
    <div>...</div>
</my-element>
```

Code increases from 2.80 -> 2.88 gzip

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -526,6 +526,11 @@ const getTemplateHtml = (
   ];
 };
 
+const replaceContent = (content: DocumentFragment, withContent: Node[] | NodeListOf<Node>) => {
+  content.textContent = '';
+  content.append(...withContent)
+}
+
 export type Template = Interface<TemplateImpl>;
 class TemplateImpl {
   /** @internal */
@@ -534,6 +539,8 @@ class TemplateImpl {
   _parts: Array<TemplatePart> = [];
   // Note, this is used by the `polyfill-support` module.
   _$options?: RenderOptions;
+  /** @internal */
+  _hostWrapper: Element | undefined;
 
   constructor(
     {strings, _$litType$: type}: TemplateResult,
@@ -551,12 +558,21 @@ class TemplateImpl {
     this._$element = this._$createElement(html);
     walker.currentNode = this._$element.content;
 
-    // Reparent SVG nodes into template root
+    const {content} = this._$element;
+    const wrapper = content.firstElementChild;
     if (type === SVG_RESULT) {
-      const content = this._$element.content;
-      const svgElement = content.firstChild!;
-      svgElement.remove();
-      content.append(...svgElement.childNodes);
+      // Reparent SVG nodes into template root
+      replaceContent(content, wrapper!.childNodes);
+    } else if (wrapper?.nodeName === 'HOST') {
+      this._hostWrapper = wrapper;
+      if (DEV_MODE) {
+        if (wrapper.nextElementSibling !== null) {
+          throw new Error('When using <host> in template, there must only be one top-level element');
+        }
+      }
+      // Ensure there are no stray text nodes before/after the host element
+      // so they don't mess up node counts once the content is reparented
+      replaceContent(content, [wrapper]);
     }
 
     // Walk the template to find binding markers and create TemplateParts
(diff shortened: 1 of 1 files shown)


Westbrook:
Definitely looks cool, even if it’s a touch framework-like. With full usage in an app/element library I feel like I’d easily make out the size difference. 

Questions that seem likely:
- Could this approach also allow merging CSS classes to the host in some way?
- Could this approach also manage the use and options of a shadow root in some way?
- Would this surface also allow for element expression so that an element could `${flip()}` itself, etc.?

sorvell:
This looks awesome. I do want to think about how we should use it for LitElement. For example:
*  How does it interact with reflected attributes? (The round-tripping optimization we do for reflecting attributes would not be preserved)
* Are event handlers added at a time that is good for interacting with child components?
* Do we want any declarative system for putting events on shadowRoot?
* How do we document this and when to we recommend using it and for what?

kevinpschaaf:
Capturing some thoughts from offline discussion:

> Could this approach also allow merging CSS classes to the host in some way?

The `classMap` directive will naturally allow merging classes into the element since it uses `classList` rather than `className`, as long as there are no conflicts over the source of truth for a given class. Dealing with conflicts seems hard and hopefully not worth doing something specific there, other than last-in-wins. We'll likely need to take care to match runtime semantics in the SSR implementation.

> Would this surface also allow for element expression so that an element could `${flip()}` itself, etc.?

Yeah it should, and as long as any outside inline styles don't conflict with inside styles, and the directive is using the `style` OM and not the `style` attribute. TBH, I'm not sure either cases of a host element adding classes or modifying its inline styles are great practices, since generally those should be the domain of the owner. I think the use cases that are less fraught are mostly around adding host event listeners, setting ARIA attributes, tabIndex, etc.

> How does it interact with reflected attributes?

I don't think we'd change that feature or recommend `<host prop=${this.prop}>` be used in place of `@property({reflect: true})`. Reflected attributes go along with properties, and I think that's the right layer to define and link them. The use cases above (events, ARIA, etc.) are what `<host>` would be recommended for.

> Are event handlers added at a time that is good for interacting with child components

Confirmed that on initial render, `TemplateInstance.update` is run before DOM insertion, which ensures listeners would be installed before events would be dispatched following child `connectedCallback`.

> Do we want any declarative system for putting events on shadowRoot?

There are a few possibilities: 

* Via directive (but precludes events on both host & shadow)
  ```html
  <host @click=${onShadowRoot(this.onShadowRootClick)}>
  ```
* Via lit-html syntax (using event name):
  ```html
   <host @click=${this.onHostClick} @shadow:click=${this.onShadowRootClick}>`
  ```
* Via lit-html syntax (new prefix):
  ```html
   <host @click=${this.onHostClick} $click=${this.onShadowRootClick}>`
  ```
* Via _another_ dummy wrapper element (but likely not worth the code to handle `<host>` but no `<shadow-root>`, `<shadow-root>` but no `<host>`, both `<host>` and `<shadow-root>`, especially since the only thing you could do on shadow-root is add event listeners):
  ```html
  <host @click=${onHostClick}>
    <shadow-root @click=${onShadowRootClick}>
      ...
    </shadow-root>
  </host>
  ```

packages/lit-html/src/lit-html.ts
```diff
@@ -750,11 +770,18 @@ class TemplateInstance {
...
+      _hostWrapper,
     } = this._$template;
     const fragment = d.importNode(content, true);
     walker.currentNode = fragment;
 
-    let node = walker.nextNode();
+    if (DEV_MODE) {
+      if (_hostWrapper && !(options?.host instanceof HTMLElement)) {
+        throw new Error('When using <host> in template, RenderOptions must include a `host` property set to the host element');
+      }
```
rictic, line 781: Should we also check that <host> is present at most once, and as the sole root nonempty Node in the template?

justinfagnani:
I personally think we should close this PR and not pursue this feature. I think that host bindings encourage bad practices like setting host attributes that should be owned by the host's outer scope. @sorvell's recent work on default values should have us reflecting to attributes more carefully, which probably makes it harder for this feature to follow suit since host bindings would have to check initial values against default values.

PR #1625, opened 2021-02-25
[localize] Natural expressions

aomarks:
Templates can now contain arbitrary expressions, and no longer need to be wrapped in a function.

Fixes https://github.com/Polymer/lit-html/issues/1544

#### Before

```ts
msg((name) => html`Hello <b>${name}</b>!`, {args: [getUsername()]});
```

#### After

```ts
msg(html`Hello <b>${getUsername()}</b>!`);
```

### Plain strings with expressions

Plain strings containing expressions must now be tagged with the new `str` tag. We need this so that lit-localize can access dynamic values at runtime.

```ts
import {msg, str} from 'lit-localize';
msg(str`Hello ${name}`);
```

### Implementation

The trickiest part of this PR was dealing with these issues:

1. Variable scopes are different between the source module and generated locale modules.
2. The order of expressions can change from one locale to another.

Previously, these issues were both solved by requiring that templates with expressions were wrapped in a function, and by enforcing that template expressions only referenced parameters to that function:

```ts
// some-locale.ts
export const templates = {
  abc123: (a, b) => html`foo ${a} bar ${b}`,
};

// another-locale.ts
export const templates = {
  abc123: (a, b) => html`${b} rab ${a} oof`,
};
```

Here is what we now do instead:

1. Instead of passing the `options.args` value through to the localized template function, we now instead "transplant" the `values` array from the source `TemplateResult` to the localized `TemplateResult`.

2. We encode the *ordering* of expressions into the locale modules. Since we no longer need to encode actual expression values into generated locale modules, we can overload the expression slots to encode the index into the source template values array.

So, generated locale modules now look like this:

```ts
// some-locale.ts
export const templates = {
  abc123: html`foo ${0} bar ${1}`,
};

// another-locale.ts
export const templates = {
  abc123: html`${1} rab ${0} oof`,
};
```

And in our `msg` function, we do something roughly like this:

```ts
const expressionOrder = localizedTemplate.values as number[];
localizedTemplate.values = expressionOrder.map((i) => sourceTemplate.values[i]);
```


packages/localize/src_client/lit-localize.ts
```diff
@@ -311,64 +336,92 @@ const setLocale: ((newLocale: string) => Promise<void>) & {
...
+        }).values = order.map((i) => (template as TemplateResult).values[i]);
+        return localized;
+      }
     }
   }
-  if (typeof template === 'function') {
-    return template(...(options?.args ?? []));
+  if (typeof template !== 'string' && 'strTag' in template) {
+    // E.g. str`Hello ${name}!` in original source locale.
+    return joinStringsAndValues(template.strings, template.values);
```
rictic, line 395: Why is this needed for str tagged templates but not html tagged templates??

packages/localize/src/typescript.ts
```diff
@@ -89,3 +94,34 @@ export function escapeStringToEmbedInTemplateLiteral(
...
+): ts.TemplateLiteral {
+  const file = ts.createSourceFile(
+    '__DUMMY__.ts',
+    '`' + templateLiteralBody + '`',
+    ts.ScriptTarget.ESNext,
+    false,
+    ts.ScriptKind.JS
+  );
+  if (file.statements.length !== 1) {
+    throw new KnownError('Internal error: expected 1 statement');
```
rictic, line 114: Should these be KnownErrors? Seems like a stack trace might be useful here.

rictic (approved):

packages/localize/testdata/xliff/input/xliff/es-419.xlf
```diff
@@ -1,54 +1,66 @@
...
-    <trans-unit id="h3c44aff2d5f5ef6b">
-      <source>Hello <ph id="0">&lt;b></ph>World<ph id="1">&lt;/b></ph>!</source>
-      <target>Hola <ph id="0">&lt;b></ph>Mundo<ph id="1">&lt;/b></ph>!</target>
-    </trans-unit>
-    <trans-unit id="h82ccc38d4d46eaa9">
-      <source>Hello <ph id="0">&lt;b>${name}&lt;/b></ph>!</source>
-      <target>Hola <ph id="0">&lt;b>${name}&lt;/b></ph>!</target>
-    </trans-unit>
-  </body>
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
aomarks: The previous version was generated by an older version of the XLIFF generator that omitted the header, and I never updated it. In this PR, I re-generated it from scratch.

rictic (approved):

PR #1629, opened 2021-02-25
[lit-html]Call ref callback against options.host. Fixes #1628

### LINKED ISSUE #1628: [lit-html] ref callback should be called against options.host context if present
# Description

The `ref` callback currently does not support calling callbacks bound to a `LitElement` (similar to how we do events).  

# lit-html version

2.0.0-pre.6

# Steps to Reproduce

```ts
@customElement('my-element')

  render() {
    return html`
      <input ${ref(() => this.inputChanged())}> // ok
      <input ${ref(this.inputChanged)}> // fails
    `;
  }

  inputChanged(input: HTMLInputElement) {
    input.focus();
  }
}
```

kevinpschaaf:
Fixes #1628


sorvell (approved):
LGTM

PR #1631, opened 2021-02-26
Fix a selector that determines if the polyfills are injected into tests.

bicknellr:
This doesn't actually seem to be affecting anything. I tried out calling `querySelector` with a similarly 'broken' selector and returns the element you'd expect if the `]` was there.


kevinpschaaf (approved):
Thanks!

PR #1634, opened 2021-02-26
[lit-next] Re-escape null characters after Terser minification

aomarks:
Replace literal null bytes with escaped `\0`, assuming they only appear inside strings.

With unsafe minifications on, Terser replaces `\0` and `\x00` escape sequences with actual null bytes. This can cause problems with tools, especially if the null byte appears early in the file. E.g. VSCode will not detect the file as UTF-8, and a bug caused @web/dev-server to crash.

See https://github.com/Polymer/lit-html/issues/1633

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "husky": "^4.3.0",
     "ignore-sync": "^3.1.0",
     "lint-staged": "^10.4.0",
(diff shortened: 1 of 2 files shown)


aomarks:
Did https://github.com/Polymer/lit-html/pull/1635 instead

PR #1635, opened 2021-02-26
[lit-next] Simplify lit-html attribute name regexp with \s

aomarks:
Replaces `\0-\x1F\x7F-\x9F ` with `\s` in the regexp that is used to match attribute names.

- Smaller is good
- Doesn't include `\0`, so Terser minification to a null byte is no longer an issue
- Doesn't regress any use-cases we can think of

See https://github.com/Polymer/lit-html/issues/1633 for discussion

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -176,7 +176,7 @@ const comment2EndRegex = />/g;
  *
  * So an attribute is:
  *  * The name: any character except a whitespace character, ("), ('), ">",
- *    "=", or "/"
+ *    "=", or "/". Note: this is different from the HTML spec which also excludes control characters.
  *  * Followed by zero or more space characters
  *  * Followed by "="
  *  * Followed by zero or more space characters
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -179,12 +174,9 @@ const comment2EndRegex = />/g;
...
  * https://infra.spec.whatwg.org/#ascii-whitespace
  *
- * "\0-\x1F\x7F-\x9F" are Unicode control characters, which includes every
- * space character except " ".
- *
  * So an attribute is:
- *  * The name: any character except a control character, space character, ('),
- *    ("), ">", "=", or "/"
+ *  * The name: any character except a whitespace character, ("), ('), ">",
+ *    "=", or "/"
```
justinfagnani, line 179: Since we link to the HTML spec, and then don't use one of its definitions, I think it's worth pointing out here how we diverge. Maybe:

```suggestion
 *  * The name: any character except a whitespace character, ("), ('), ">",
 *    "=", or "/". Note: this is different from the HTML spec which also excludes control characters.
```

justinfagnani (approved):

PR #1642, opened 2021-03-03
Reintroduce `startNode` and `endNode` on `ChildPart`.

### LINKED ISSUE #1641: [lit-next] Reintroduce `startNode` and `endNode` to `ChildPart`
`NodePart` from Lit 1 had `startNode` and `endNode` properties that allowed reaching into the DOM rendered by the part. `ChildPart` should also have these properties.

bicknellr:
This PR exposes the start and end marker nodes of `ChildPart` through `startNode` and `endNode` getters, like `NodePart` in Lit 1.

Fixes #1641.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -909,19 +909,39 @@ class ChildPartImpl {
     this._$setChildPartConnected?.(isConnected);
   }
 
+  /**
+   * The parent node into which the part renders its content.
+   *
+   * A ChildPart's content consists of a range of adjacent child nodes of
+   * `.parentNode`, possibly bordered by 'marker nodes' (`.startNode` and
+   * `.endNode`).
+   *
+   * - If both `.startNode` and `.endNode` are defined, then the part's content
+   * consists of all siblings between `.startNode` and `.endNode`, exclusively.
+   *
+   * - If `.startNode` is defined but `.endNode` is undefined, then the part's
+   * content consists of all siblings following `.startNode`, up to and
+   * including the last child of `.parentNode`. (If `.endNode` is defined, then
+   * `.startNode` will always be defined.)
+   *
+   * - If both `.endNode` and `.startNode` are undefined, then the part's
+   * content consists of all child nodes of `.parentNode`.
+   */
   get parentNode(): Node {
     return wrap(this._$startNode).parentNode!;
   }
 
   /**
-   * The first node within the part.
+   * The part's leading marker node, if any. See `.parentNode` for more
+   * information.
    */
   get startNode(): Node | undefined {
     return this._$startNode === this._$endNode ? undefined : this._$startNode;
   }
 
   /**
-   * The last node within the part.
+   * The part's trailing marker node, if any. See `.parentNode` for more
+   * information.
    */
   get endNode(): Node | undefined {
     return this._$startNode === this._$endNode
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -909,10 +909,46 @@ class ChildPartImpl {
...
   get parentNode(): Node {
     return wrap(this._$startNode).parentNode!;
   }
 
+  /**
+   * The part's leading marker node, if any. See `.parentNode` for more
+   * information.
+   */
+  get startNode(): Node | undefined {
+    return this._$startNode === this._$endNode ? undefined : this._$startNode;
```
kevinpschaaf, line 939: Just return startNode/endNode for now, but this protects for doing what you wrote in the future.

kevinpschaaf (changes requested):

packages/lit-html/src/lit-html.ts
bicknellr: Ok, updated.

packages/lit-html/src/lit-html.ts
```diff
@@ -911,10 +911,44 @@ class ChildPartImpl {
...
+  get startNode(): Node | undefined {
+    return this._$startNode;
+  }
+
+  /**
+   * The part's trailing marker node, if any. See `.parentNode` for more
+   * information.
+   */
+  get endNode(): Node | undefined {
+    return this._$endNode ?? undefined;
```
bicknellr, line 949: Would you prefer I change the `| undefined` in the return types to `| null` also? I guess it's DOM convention and would let us avoid `?? undefined`.

bicknellr: I guess I typically prefer using `undefined` as the 'true' null value thing, but this seems like a case where we could go the other way.

justinfagnani: It's an interesting call. In JS `undefined` is the missing value, and in the DOM `null` is. 🤬

One reason why `null` flows in here so naturally is that `null` is returned from `node.nextSibling` and a valid value for `insertBefore()`'s ref node. I'd say in this case that we don't try to force the natural type of this into our default style and type the getters as `Node | null`.

bicknellr: Ok, updated, PTAL

packages/lit-html/src/lit-html.ts
```diff
@@ -911,10 +911,44 @@ class ChildPartImpl {
...
+   * `.parentNode`, possibly bordered by 'marker nodes' (`.startNode` and
+   * `.endNode`).
+   *
+   * - If both `.startNode` and `.endNode` are defined, then the part's content
+   * consists of all siblings between `.startNode` and `.endNode`, exclusively.
+   *
+   * - If `.startNode` is defined but `.endNode` is undefined, then the part's
+   * content consists of all siblings following `.startNode`, up to and
+   * including the last child of `.parentNode`. (If `.endNode` is defined, then
+   * `.startNode` will always be defined.)
```
justinfagnani, line 927:
```suggestion
   * - If `.startNode` is defined but `.endNode` is null, then the part's
   * content consists of all siblings following `.startNode`, up to and
   * including the last child of `.parentNode`. If `.endNode` is non-null, then
   * `.startNode` will always be non-null.
```

justinfagnani (approved):

PR #1643, opened 2021-03-03
[labs/task] Better typing for Task and initialState

justinfagnani:
Adds deps and result types to Task making `.value` and `render()` a little easier to use. Also adds a `initialState` sentinel value that a task function can return to rest the task status to INITIAL.

### DIFF
--- a/packages/labs/task/src/test/task_test.ts
+++ b/packages/labs/task/src/test/task_test.ts
@@ -284,12 +284,6 @@ suite('Task', () => {
         async ([state]) => (state === 'initial' ? initialState : 'A'),
         () => [this.state]
       );
-
-      t2 = new Task(
-        this,
-        () => new Promise(() => []),
-        () => []
-      );
     }
     customElements.define(generateElementName(), TestEl);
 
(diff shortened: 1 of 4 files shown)


packages/labs/task/src/test/task_test.ts
```diff
@@ -273,4 +273,41 @@ suite('Task', () => {
...
+    }
+    customElements.define(generateElementName(), TestEl);
+
+    const el = new TestEl();
+    assert.equal(el.task.status, TaskStatus.INITIAL, 'initial');
+    container.append(el);
+
+    // After one microtask we expect the task function to have been
+    // called, but not completed
+    await Promise.resolve();
```
sorvell, line 302: Make the task resolution explicit so we're not depending on this?

sorvell (approved):
LGTM. Just one testing nit.

packages/labs/task/src/test/task_test.ts
justinfagnani: Added a `taskComplete` promise, but that doesn't address this await which is waiting for the task function to be called, but not completed. It did replace the next `await Promise.resolve()` though.

PR #1647, opened 2021-03-03
[lit-next] Use WeakMap for templateCache

### LINKED ISSUE #1646: [lit-html] templateCache should be a WeakMap to allow GC'ing unsafeHTML templates
# Description
Currently the `templateCache` is a `Map`, which is generally fine for normal `html`-generated `TemplateResult`s since the `TemplateStringsArray` used for the cache key is tied to the module, and typically won't ever get GC'ed anyway. However `unsafeHTML` returns `TemplateResult`s created with "fake" TSAs, which will be stored indefinitely in the `Map`. Although `unsafeHTML` does return the same `TemplateResult` for the same string, once the string changes, a new `TemplateResult` will be created and the previous template will be held in the cache indefinitely, despite no way to re-render it.

For this reason, we should use a `WeakMap`, so that we can GC templates that are no longer associated with a live `unsafeHTML` directive.

kevinpschaaf:
Fixes #1646

Want to watch out for any movement in perf numbers based on this change.


justinfagnani (approved):

voidus:
I think this change caused [an issue](https://github.com/lit/lit/issues/3445), mostly because the cache isn't a "proper" cache because a hit/miss actually influences the behaviour: If there's a cache miss, the existing DOM nodes won't be re-used, which causes browser state (e.g. in form inputs) to be lost.

If you have time, it would be much appreciated if you could have a look at the linked issue.

PR #1649, opened 2021-03-04
[lit-next] Fix typings to match module/main

Artur-:
Fixes https://github.com/Polymer/lit-element/issues/1164

### DIFF
--- a/packages/lit-element/package.json
+++ b/packages/lit-element/package.json
@@ -11,6 +11,7 @@
   "homepage": "https://lit-element.polymer-project.org/",
   "main": "index.js",
   "module": "index.js",
+  "types": "index.d.ts",
   "type": "module",
   "exports": {
     ".": {
@@ -93,7 +94,6 @@
     "tslib": "^2.0.3",
     "typescript": "^4.1.3"
   },
-  "typings": "index.d.ts",
   "directories": {
     "test": "test"
   }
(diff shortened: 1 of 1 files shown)


packages/lit-element/package.json
```diff
@@ -93,7 +93,7 @@
     "tslib": "^2.0.3",
     "typescript": "^4.1.3"
   },
-  "typings": "lit-element.d.ts",
+  "typings": "index.d.ts",
```
justinfagnani, line 96: Do we even need this field anymore? I feel like TypeScript looks here by default now.

Artur-: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html says

"Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so."

Artur-: It does seems to work also if this is deleted

justinfagnani: Let's keep it, but it does like like `"types"` is the newer field.

Artur-: Changed to 'types' and moved it next to other index files

Artur-:
Anything more needed for this?

kevinpschaaf (approved):

kevinpschaaf:
Thanks!

PR #1652, opened 2021-03-04
[lit-next] Add constructor types for LitElement & ReactiveElement

kevinpschaaf:
Enables easier writing of mixins.

```ts
const MyMixin = <T extends LitElementConstructor>(superClass: T) => {
  class MyMixinSubclass extends superClass {
    // ...
  }
  return MyMixinSubclass;
}
```

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -141,6 +141,20 @@ export class LitElement extends ReactiveElement {
   }
 }
 
+/**
+ * Constructor type for easier definition of mixins that apply to LitElement.
+ *
+ * ```ts
+ * const MyMixin = <T extends LitElementConstructor>(superClass: T) => {
+ *   class MyMixinSubclass extends superClass {
+ *     // ...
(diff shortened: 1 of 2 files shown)


packages/lit-element/src/lit-element.ts
```diff
@@ -141,6 +141,20 @@ export class LitElement extends ReactiveElement {
...
+ * ```ts
+ * const MyMixin = <T extends LitElementConstructor>(superClass: T) => {
+ *   class MyMixinSubclass extends superClass {
+ *     // ...
+ *   }
+ *   return MyMixinSubclass;
+ * }
+ * ```
+ */
+export type LitElementConstructor = new (...args: unknown[]) => LitElement;
```
justinfagnani, line 156: Did you try this in an actual mixin? I thought it had to be `new (...args: any[])`

kevinpschaaf:
Closing, as the docs are just going to recommend using the generic `Constructor<T>` helper since it's also needed for casting the return value of the mixin.

PR #1654, opened 2021-03-05
[lit-html] add Lit 2 directive syntax to Lit 1

e111077:
Add forwards compatibility for Lit 1 directives. This will allow users to write with the Lit 2 directive class syntax and also adds some files so that the we can migrate users to the new import targets and then switch them to Lit 2 rather than those changes being atomic.

Notable changes:
* `directive` can now take a `DirectiveClass` as an argument in `lit-html/directives`
* The API of the lit 2 `[a-zA-Z]+Part`s have added to `lit-html/directives`
* creation of root level:
  * `async-directive.ts`
  * `directive.ts`
* creation of `ChildPart`

Note: this seems to have broken lit-analyzer 1.1.9 internally, but i could not reproduce those errors externally. This seems to work just fine in lit-analyzer 1.2.0.

Additionally, usage of `ChildPart` in a directive may cause lit-analyzer to not detect that a directive is a directive, but this should not cause any failures because `NodePart` (Lit 1 equivalent) is never bound to anything that needs type checking.

Related #1653

### DIFF
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -78,6 +78,13 @@ export abstract class AsyncDirective extends Directive {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected disconnected() {
   }
+  /**
+   * User callback to restore the working state of the directive prior to the next
+   * render. This should generally re-do the work that was undone in `disconnected`.
+   *
+   * NOTE: In lit-html 1.x, the `disconnected` and `reconnected` callbacks WILL NOT BE
+   * CALLED. The interface is provided here for forward-compatible directive authoring only.
+   */
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected reconnected() {
   }
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -66,14 +66,14 @@ export abstract class AsyncDirective extends Directive {
   }
 
   /**
-   * User callbacks for implementing logic to release any
+   * User callback for implementing logic to release any
    * resources/subscriptions that may have been retained by this directive.
    * Since directives may also be re-connected, `reconnected` should also be
    * implemented to restore the working state of the directive prior to the next
    * render.
    *
-   * In the v1 version of these APIs, we don't monitor for disconnection and
-   * reconnection.
+   * NOTE: In lit-html 1.x, the `disconnected` and `reconnected` callbacks WILL NOT BE
+   * CALLED. The interface is provided here for forward-compatible directive authoring only.
    */
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected disconnected() {
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -75,8 +75,10 @@ export abstract class AsyncDirective extends Directive {
    * In the v1 version of these APIs, we don't monitor for disconnection and
    * reconnection, we only call these methods when setValue is called.
    */
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected disconnected() {
   }
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected reconnected() {
   }
 }
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -1,0 +1,81 @@
+/**
+ * @license
+ * Copyright (c) 2021 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import { Directive, Part, PartInfo } from "./directive.js";
+import * as legacyLit from "./lit-html.js";
+
+/**
+ * A superclass for directives that need to asynchronously update.
+ */
+export abstract class AsyncDirective extends Directive {
+  private readonly ddPart: legacyLit.Part;
+  private renderedYet = false;
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    this.ddPart = (partInfo as Part).legacyPart;
+  }
+
+  private ddGetNode(): Node | undefined {
+    if (this.ddPart instanceof legacyLit.NodePart) {
+      return this.ddPart.startNode;
+    } else if (this.ddPart instanceof legacyLit.EventPart) {
+      return this.ddPart.element;
+    } else if (this.ddPart instanceof legacyLit.BooleanAttributePart) {
+      return this.ddPart.element;
+    } else if (
+      this.ddPart instanceof legacyLit.PropertyPart ||
+      this.ddPart instanceof legacyLit.AttributePart
+    ) {
+      return this.ddPart.committer.element;
+    }
+    return undefined;
+  }
+
+  private shouldRender() {
+    if (!this.renderedYet) {
+      this.renderedYet = true;
+      return true;
+    }
+    const node = this.ddGetNode();
+    if (node != null && !node.isConnected) {
+      // node is disconnected, do not render
+      return false;
+    }
+    return true;
+  }
+
+  setValue(value: unknown) {
+    if (!this.shouldRender()) {
+      // node is disconnected, do nothing.
+      return;
+    }
+
+    this.ddPart.setValue(value);
+    this.ddPart.commit();
+  }
+
+  /**
+   * User callbacks for implementing logic to release any
+   * resources/subscriptions that may have been retained by this directive.
+   * Since directives may also be re-connected, `reconnected` should also be
+   * implemented to restore the working state of the directive prior to the next
+   * render.
+   *
+   * In the v1 version of these APIs, we don't monitor for disconnection and
+   * reconnection, we only call these methods when setValue is called.
+   */
+  protected disconnected() {}
+  protected reconnected() {}
+}
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -73,7 +73,7 @@ export abstract class AsyncDirective extends Directive {
    * render.
    *
    * In the v1 version of these APIs, we don't monitor for disconnection and
-   * reconnection, we only call these methods when setValue is called.
+   * reconnection.
    */
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   protected disconnected() {
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -48,11 +48,7 @@ export abstract class AsyncDirective extends Directive {
       return true;
     }
     const node = this._legacyGetNode();
-    if (node != null && !node.isConnected) {
-      // node is disconnected, do not render
-      return false;
-    }
-    return true;
+    return !!(node?.isConnected);
   }
 
   setValue(value: unknown) {
--- a/src/async-directive.ts
+++ b/src/async-directive.ts
@@ -13,41 +13,41 @@
  */
 
 import {Directive, Part, PartInfo} from './directive.js';
-import * as legacyLit from './lit-html.js';
+import {AttributePart as LegacyAttributePart, BooleanAttributePart as LegacyBooleanAttributePart, EventPart as LegacyEventPart, NodePart as LegacyNodePart, Part as LegacyPart, PropertyPart as LegacyPropertyPart,} from './lit-html.js';
 
 /**
  * A superclass for directives that need to asynchronously update.
  */
 export abstract class AsyncDirective extends Directive {
-  private readonly ddPart: legacyLit.Part;
-  private renderedYet = false;
+  private readonly _legacyPart: LegacyPart;
+  private _renderedYet = false;
 
   constructor(partInfo: PartInfo) {
     super(partInfo);
-    this.ddPart = (partInfo as Part).legacyPart;
+    this._legacyPart = (partInfo as Part).legacyPart;
   }
 
-  private ddGetNode(): Node|undefined {
-    if (this.ddPart instanceof legacyLit.NodePart) {
-      return this.ddPart.startNode;
-    } else if (this.ddPart instanceof legacyLit.EventPart) {
-      return this.ddPart.element;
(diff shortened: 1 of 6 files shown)


src/directive.ts
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Yields all toplevel nodes inside the dom range managed by the ChildPart.
+ *
+ * So if the ChildPart has rendered:
+ *     html`<label><input></label> <div></div>`
+ *
+ * This function will yield the label and the div, but not the input.
+ */
+export function getRenderedNodes(childPart: ChildPart) {
```
kevinpschaaf, line 70: Unclear why this is exported? It's not v2 api...

src/async-directive.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * User callbacks for implementing logic to release any
+   * resources/subscriptions that may have been retained by this directive.
+   * Since directives may also be re-connected, `reconnected` should also be
+   * implemented to restore the working state of the directive prior to the next
+   * render.
+   *
+   * In the v1 version of these APIs, we don't monitor for disconnection and
+   * reconnection, we only call these methods when setValue is called.
```
kevinpschaaf, line 76: Docs seem wrong here... they're never called right?

src/directive.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+  return results;
+}
+
+export type AttributePart = Interface<AttributePartImpl>;
+class AttributePartImpl {
+  readonly type: typeof PartType.ATTRIBUTE|typeof PartType.PROPERTY;
+
+  get options(): legacyLit.RenderOptions|undefined {
+    return undefined;
```
kevinpschaaf, line 87: For a very modest code size increase, we could go and put `options` on each legacy part type (the `DefaultTemplateProcessor` that creates all the parts has it) via an optional constructor argument on each Part class, so that the v2 parts can actually return something here.

It still won't guarantee it's there for people who created custom template processors, but seems like it would be an improvement (and they can't migrate that code to v2 anyway)?

Or maybe we just don't care because they couldn't have relied on options in v1, so it's not a migration concern?

@justinfagnani thoughts?

kevinpschaaf (commented):
Just a few minor questions

src/directive.ts
e111077: You're right. This was a copy paste will clean up

src/async-directive.ts
e111077: @rictic

kevinpschaaf:
We also need to add forward-compat for at least `directive-helpers.js`:`isTemplateResult` (needed for open-wc/storybook to migrate to hybrid support 1.x and 2.x).

We should also consider adding these to the `directive-helpers.js`:
* `isTemplateResult`
* `insertPart`
* `setChildPartValue`
* `setCommittedValue`
* `getCommittedValue`
* `removePart`
* `clearPart`

Seems like they would be required to migrate more advanced directives.

src/directive-helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Yields all toplevel nodes inside the dom range managed by the ChildPart.
+ *
+ * So if the ChildPart has rendered:
+ *     html`<label><input></label> <div></div>`
+ *
+ * This function will yield the label and the div, but not the input.
+ */
+export function getRenderedNodes(childPart: ChildPart) {
```
bicknellr, line 32: Just a heads up, we shouldn't need `getRenderedNodes` anymore now that #1642 is merged.

src/async-directive.ts
kevinpschaaf: They won't get called in V1 (we should just document that fact in the API docs). The idea is this is ok because a V1 directive couldn't have relied on the callback being called.

The only reason to put them here on the interface is so that users can go ahead and write code that _will_ have the callbacks run in V2.

rictic: Yeah, sorry, this doc is out of date. I think for a little while I have a halfway implementation here, but it diverged enough from the behavior of the v2 version that it seemed like it might do more harm than good.

+1 to having clear docs that these are just for interface compatibility, that lit-html v1 will not call them

src/directive.ts
rictic: IMO since this is for allowing existing directives to migrate to the new API surface, we don't need to support novel features, just translation

kevinpschaaf: Yeah, makes sense.

src/directive.ts
```diff
@@ -0,0 +1,267 @@
...
+
+export type PartType = typeof PartType[keyof typeof PartType];
+
+export interface ChildPartInfo {
+  readonly type: typeof PartType.CHILD;
+}
+
+export interface AttributePartInfo {
+  // eslint-disable-next-line @typescript-eslint/type-annotation-spacing
+  readonly type:|typeof PartType.ATTRIBUTE|
```
justinfagnani, line 35: looks like this needs formatting

src/directive.ts
justinfagnani: I agree w/ @rictic here

src/async-directive.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * User callbacks for implementing logic to release any
+   * resources/subscriptions that may have been retained by this directive.
+   * Since directives may also be re-connected, `reconnected` should also be
+   * implemented to restore the working state of the directive prior to the next
+   * render.
+   *
+   * In the v1 version of these APIs, we don't monitor for disconnection and
+   * reconnection.
+   */
```
kevinpschaaf, line 77:
```suggestion
  /**
   * User callback for implementing logic to release any
   * resources/subscriptions that may have been retained by this directive.
   * Since directives may also be re-connected, `reconnected` should also be
   * implemented to restore the working state of the directive prior to the next
   * render.
   *
   * NOTE: In lit-html 1.x, the `disconnected` and `reconnected` callbacks WILL NOT BE
   * CALLED. The interface is provided here for forward-compatible directive authoring only.
   */
```

src/directive.ts
```diff
@@ -0,0 +1,267 @@
...
+   * this is undefined.
+   */
+  get strings() {
+    return this.legacyPart.strings;
+  }
+  get tagName() {
+    return this.element.tagName;
+  }
+
+  constructor(readonly legacyPart: legacyLit.BooleanAttributePart) {
```
justinfagnani, line 134: Don't use constructor parameter properties - declare this as a regular property.

src/async-directive.ts
```diff
@@ -0,0 +1,84 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  protected disconnected() {
+  }
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  protected reconnected() {
```
kevinpschaaf, line 82:
```suggestion
  /**
   * User callback to restore the working state of the directive prior to the next
   * render. This should generally re-do the work that was undone in `disconnected`.
   *
   * NOTE: In lit-html 1.x, the `disconnected` and `reconnected` callbacks WILL NOT BE
   * CALLED. The interface is provided here for forward-compatible directive authoring only.
   */
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  protected reconnected() {
```

kevinpschaaf (dismissed):
LGTM, pending CI green(formatting issues?).

Added some suggestions for the API docs for the callbacks (I realize the lit-next ones only had one API doc for both callbacks; we'll have to fix it there also.

src/directive.ts
```diff
@@ -0,0 +1,267 @@
...
+    return this.render(...props);
+  }
+}
+
+export function directive<C extends DirectiveClass>(directiveClass: C) {
+  const partToInstance =
+      new WeakMap<legacyLit.Part, readonly[Part, InstanceType<C>]>();
+  const result = legacyDirective((...props: unknown[]) => {
+    return (part: legacyLit.Part) => {
+      const cached = partToInstance.get(part);
```
justinfagnani, line 250: I think this is going to break in some nested directive situations. We need to try things like `guard([items], repeat(items, ...))`.

justinfagnani (changes requested):

src/async-directive.ts
```diff
@@ -0,0 +1,84 @@
...
+export abstract class AsyncDirective extends Directive {
+  private readonly ddPart: legacyLit.Part;
+  private renderedYet = false;
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    this.ddPart = (partInfo as Part).legacyPart;
+  }
+
+  private ddGetNode(): Node|undefined {
```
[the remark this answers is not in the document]
e111077: renaming to legacy. Peter said it used to be called disconnectable directive

src/async-directive.ts
```diff
@@ -0,0 +1,84 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Directive, Part, PartInfo} from './directive.js';
+import * as legacyLit from './lit-html.js';
```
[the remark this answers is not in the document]
e111077: this causes import name clashing and will have to still import `Part as LegacyPart`

src/directive.ts
e111077: Google style guide

src/directive.ts
e111077: Heya had a quick chat with peter and russell and we were a bit confused on what the concern is. Is it you think it'll write to the wrong location?

Additionally, we can't test with `guard` and `repeat` because they are not implemented using this class-based `Directive` impl. Additionally, assuming the user transferred their own directive to this and use it in one of our built-in ones, it should handle just fine.

The case where we think your concern about nested directives would come into play would be if a user implemented two of their own directives and nest them. I'd argue that's quite a fringe case.

I'd be happy to put a quick warning in a jsdoc saying this may not support certain advanced directive usage, but I'd like to understand the concern a bit more

src/directive-helpers.ts
```diff
@@ -0,0 +1,21 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {TemplateResult} from './lit-html.js';
```
bicknellr, line 15:
```suggestion
import {TemplateResult} from './lit-html.js';
export {isPrimitive} from './lib/parts.js';
```

bicknellr: (Adding this in http://cl/364917016)

src/directive.ts
kevinpschaaf: To clarify, I think:
* In 1.x, each type of directive hung its state off the part via weak map and it persisted even when directives changed and then changed back. Nested directives just worked by virtue of each having their own weak map on the part.
* In 2.x, each type of directive (by class) hangs its state off the part (or in the case of nesting, the nested directive), and its state is lost when directive type changes and then changes back.
* This PR currently matches the 2.x behavior as long as we don't consider nesting, but nesting will fail I think (but only for user directives, since the built-in directives don't use this, like Elliot said). Not sure if that's better or worse than caching by class type, which would make it more backward compatible (but doesn't get the clearing behavior of 2.x)

justinfagnani: It's not just nesting. This would fail:

```ts
html`${cond ? fooDirective() : barDirective()}`
```

because Foo and Bar directive classes would be mixed up.

Do something like, but with proper undefined handling:
```ts
const partToInstance = directiveCache.get(directiveClass);
const cached = partToInstance.get(part);
```

justinfagnani (changes requested):
Once the cache issue is fixed and there are some tests this looks good.

src/directive.ts
e111077: Just checked. This does not fail because there is a weakmap per instance of `directive(...)` call rather than a persistent one. Added tests to show this

src/async-directive.ts
```diff
@@ -0,0 +1,94 @@
...
+    return undefined;
+  }
+
+  private _shouldRender() {
+    if (!this._renderedYet) {
+      this._renderedYet = true;
+      return true;
+    }
+    const node = this._legacyGetNode();
+    if (node != null && !node.isConnected) {
```
sorvell, line 51: Can this be a bit simpler, like this?

`return !!(node?.isConnected);`

src/directive.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * An AttributePart that manages an event listener via add/removeEventListener.
+ *
+ * This part works by adding itself as the event listener on an element, then
+ * delegating to the value passed to it. This reduces the number of calls to
+ * add/removeEventListener if the listener changes frequently, such as when an
+ * inline function is used as a listener.
+ *
+ * Because event options are passed when adding listeners, we must take case
```
sorvell, line 148:
```suggestion
 * Because event options are passed when adding listeners, we must take care
```

src/test/directives/migration-directives_test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    const parentName = this._targetEl!.tagName.toLocaleLowerCase();
+    return this.render(`${value} ${parentName}`);
+  }
+}
+
+const addParentLength = directive(AddParentLengthDirective);
+const addParentName = directive(AddParentNameDirective);
+
+suite('migration directives', () => {
```
sorvell, line 98: Should add a test verifying that directive state is maintained across re-renders.

sorvell (changes requested):

sorvell (approved):
LGTM

kevinpschaaf (dismissed):

e111077:
Ping @justinfagnani to update review

src/directive.ts
justinfagnani: We don't use Google internal style guide, and we don't use non-standard syntactic features of TypeScript outside of the type system. So no constructor parameter properties, namespaces or enums. Please be consistent with this codebase.

justinfagnani (changes requested):

justinfagnani (approved):

e111077:
ugh i had to format and I need an LGTM again

justinfagnani (approved):

PR #1665, opened 2021-03-11
[lit-html] Some fixes for styleMap()

justinfagnani:
* Accepts `undefined` and `null` property values, and removes the property. See #1616 for a PR on master
* Doesn't render `name: undefined;` for `{name: undefined}` on first render
* Makes `previousStyleProperties` re-namable to save some bytes


packages/lit-html/src/directives/style-map.ts
```diff
@@ -64,37 +64,41 @@ class StyleMap extends Directive {
...
     // iteration.
-    this.previousStyleProperties!.forEach((name) => {
-      if (!(name in styleInfo)) {
-        this.previousStyleProperties!.delete(name);
-        if (name.indexOf('-') === -1) {
+    this._previousStyleProperties!.forEach((name) => {
+      // If the name isn't in styleInfo or it's null/undefined
+      if (styleInfo[name] == null) {
+        this._previousStyleProperties!.delete(name);
+        if (name.includes('-')) {
```
sorvell, line 82: Are we sure we're covered for this in IE11?

packages/lit-html/src/test/directives/style-map_test.ts
```diff
@@ -77,13 +85,19 @@ suite('styleMap', () => {
...
+      marginTop: '2px',
+      'padding-bottom': '4px',
+      borderRadius: '5px',
+    });
     const el = container.firstElementChild as HTMLElement;
     assert.equal(el.style.marginTop, '2px');
     assert.equal(el.style.paddingBottom, '4px');
-    renderStyleMap({});
+    assert.equal(el.style.borderRadius, '5px');
+    renderStyleMap({borderRadius: undefined});
```
sorvell, line 97: Add an explicit `null`?

sorvell (approved):
LGTM with a couple nits

sorvell (changes requested):
Oh, need to add this change to the changelog.

packages/lit-html/src/directives/style-map.ts
justinfagnani: It's included in the core-js polyfills.

sorvell (approved):
LGTM

PR #1666, opened 2021-03-11
[lit-html] Make directive class fields private and re-namable

justinfagnani:
Takes 600B off total minified size, 80B gzipped.

styleMap is handled in #1665


kevinpschaaf (approved):
LGTM

PR #1668, opened 2021-03-11
[localize] Split localize CLI into extract and build commands

aomarks:
Previously the `lit-localize` CLI always does both extraction and project building. It's nice to separate these steps, in particular because of the next PR, which directly exposes locale transformers for use with `@rollup/plugin-typescript` (or other tools).

With this change, running `lit-localize extract` will generate XLIFF files (translation requests), and then building is done either with `lit-localize build` or from within a Rollup configuration.

As part of this, I created some new classes to more cleanly separate the high level logic from the CLI : `LitLocalizer`, `TransformLitLocalizer`, `RuntimeLitLocalizer`.

There's a bit of churn in the e2e test files, because we now test `build` separately from `extract`.

First half of https://github.com/Polymer/lit-html/issues/1667


packages/localize/README.md
```diff
@@ -516,30 +516,20 @@ In transform mode, applications of the `Localized` mixin are removed.
...
+| --------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `extract` | Extract templates from `msg()` calls across all source files included by your `tsconfig.json`, and create or update XLIFF (`.xlf`) files containing translation requests.                                                                                                                                                                      |
+| `build`   | Read translations and build the project according to the configured mode.<br><br>In _transform_ mode, compile your TypeScript project for each locale, replacing `msg` calls with localized templates.<br><br>In _runtime_ mode, generate a `<locale>.ts` file for each locale, which can be dynamically loaded by the `@lit/localize` module. |
```
rictic, line 528: This is much easier to understand and explain, a good sign for this change. +1

packages/localize/src/cli.ts
```diff
@@ -6,24 +6,40 @@
...
+Options:
+  --help      Display this help message.
+  --config    Path to JSON configuration file.
+              Default: ./lit-localize.json
+              See https://github.com/Polymer/lit-html/tree/lit-next/packages/localize#readme for details.
+`;
+
+const validCommands = ['build', 'extract'] as const;
+type Command = typeof validCommands[number];
+const isValidCommand = (str: string): str is Command =>
```
rictic, line 35: Minor & optional: consider just isCommand rather than isValidCommand

isCommand has a natural relationship to the type relationship `str is Command`

isValidCommand suggests more, like it might throw, or do validation beyond checking whether it's in the type

packages/localize/src/cli.ts
```diff
@@ -60,93 +77,88 @@ function version() {
...
-  } else {
-    throwUnreachable(
-      config.output,
-      `Internal error: unknown output mode ${
-        (config.output as typeof config.output).mode
-      }`
-    );
+function makeLocalizer(config: Config): LitLocalizer {
+  switch (config.output.mode) {
+    case 'transform':
```
rictic, line 119: Do you need to know the output mode when you're just extracting?

packages/localize/src/formatters/xlb.ts
```diff
@@ -70,30 +70,17 @@ class XlbFormatter implements Formatter {
...
-          if (err) {
-            reject(err);
-          } else {
-            resolve(files);
-          }
-        }
-      )
-    );
-    const bundles: Array<Promise<Bundle>> = [];
+  readTranslations(): Array<Bundle> {
```
rictic, line 73: Consider leaving async to preserve the option in the future? Maybe someday FS latency will be high enough and node will be optimized enough that doing it async in parallel is faster?

packages/localize/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    return this._translations;
+  }
+
+  /**
+   * Check that all translations are valid given the current set of source
+   * messages.
+   */
+  validateTranslations(): ValidateTranslationsResult {
+    const {translations} = this.readTranslations();
```
rictic, line 103: This means that we read translations twice when building. Once to validate, and then a second time as part of building.

Consider taking the translations as an argument to this function.

packages/localize/testdata/transform/goldens/xliff/es-419.xlf
```diff
@@ -1,23 +1,34 @@
...
-<trans-unit id="variables_1">
-  <source>Hello <ph id="0">${name}</ph>!</source>
-  <target>Hola <ph id="0">${name}</ph>!</target>
-</trans-unit>
-<trans-unit id="lit_variables_1">
-  <source>Hello <ph id="0">${name}</ph>, click <ph id="1">&lt;a href="${url}"></ph>here<ph id="2">&lt;/a></ph>!</source>
-  <target>Hola <ph id="0">${name}</ph>, clic <ph id="1">&lt;a href="${url}"></ph>aquí<ph id="2">&lt;/a></ph>!</target>
-</trans-unit>
-</body>
+<?xml version="1.0" encoding="UTF-8" ?>
```
rictic, line 1: Why the change in header? This is the reverse of what happened in one file in the natural msg expressions PR

aomarks: Hm, I think something weird happened. But it's fixed now. This PR wasn't based on the natural expressions one originally, but now it is.

I also got really confused about which tests were which while looking into this, so I've renamed them all to be more clear.

packages/localize/src/formatters/xlb.ts
rictic: Yeah, TS proper is pretty committed to being sync, for performance reasons, but that becomes an ecosystem integration problem. Seems like the `@rollup/typescript-plugin` should allow asynchrony.

rictic (approved):

PR #1669, opened 2021-03-12
[localize] Rollup integration

### LINKED ISSUE #1667: [localize] Rollup integration
If a project is using Rollup with `@rollup/plugin-typescript`, then there's currently no way for them to run their build with locale transformation. They'll have to run `lit-localize build` directly, and then consume JS instead of TS in their Rollup config. We should expose the transformers, and provide a helper to feed them into `@rollup/plugin-typscript` (which does support transformers).

aomarks:
Exposes the TypeScript transformer for each locale, and adds a module that integrates with `@rollup/plugin-typescript`, so that you can write configs like this:

```js
import typescript from '@rollup/plugin-typescript';
import {localeTransformers} from '@lit/localize/lib/rollup.js';
import resolve from '@rollup/plugin-node-resolve';
import {terser} from 'rollup-plugin-terser';

// Config is read from ./lit-localize.json by default.
// Pass a path to read config from another location.
const locales = localeTransformers();

export default locales.map(({locale, localeTransformer}) => ({
  input: `src/index.ts`,
  plugins: [
    typescript({
      transformers: {
        before: [localeTransformer],
      },
    }),
    resolve(),
    terser(),
  ],
  output: {
    file: `bundled/${locale}/index.js`,
    format: 'es',
  },
}));
```

Fixes #1667


packages/localize/examples/transform/src/locale-codes.ts
```diff
@@ -1,9 +1,3 @@
-/**
```
rictic, line 1: Accidental deletion?

packages/localize/src/modes/transform.ts
```diff
@@ -39,6 +39,10 @@ export interface TransformOutputConfig {
   localeCodesModule?: string;
 }
 
+type TypeScriptTransformerFactoryFactory = (
```
rictic, line 42: if a bean gets involved we've gone too far

packages/localize/src/rollup.ts
```diff
@@ -0,0 +1,61 @@
...
+ *       file: `bundled/${locale}/index.js`,
+ *       format: 'es',
+ *     },
+ *   }));
+ *
+ * [0] https://github.com/rollup/plugins/tree/master/packages/typescript/#transformers
+ */
+export const localeTransformers = (
+  configPath = './lit-localize.json'
+): Array<{
```
rictic, line 42: Recommend giving this type a name, both so this code is more readable, and so that user code can more easily name it.

rictic (approved):

packages/localize/examples/transform/src/locale-codes.ts
aomarks: Actually intentional -- this license header got caught by the script I wrote the other week when we updated all of our license headers in the whole monorepo to the new style. However, this file is a generated artifact of lit localize, it doesn't get generated with a copyright header (maybe that needs to be an option or something for the very license header pedantic but... whatever).

packages/localize/src/modes/transform.ts
aomarks: I couldn't resist this name

packages/localize/src/rollup.ts
aomarks: Done.

rictic (approved):

PR #1670, opened 2021-03-12
[lit-html] Export Part classes from directive.js

e111077:
This enables us to add a forward-compatible directive.js module in lit-html 1.x

also type => interface per request of @rictic

### DIFF
--- a/packages/lit-html/src/directive.ts
+++ b/packages/lit-html/src/directive.ts
@@ -16,6 +16,7 @@ export {
   AttributePart,
   BooleanAttributePart,
   ChildPart,
+  ElementPart,
   EventPart,
   Part,
 } from './lit-html';
--- a/packages/lit-html/src/directive.ts
+++ b/packages/lit-html/src/directive.ts
@@ -12,9 +12,17 @@ import {
   ElementPart,
 } from './lit-html';
 
-export type DirectiveClass = {
+export {
+  AttributePart,
+  BooleanAttributePart,
+  ChildPart,
+  EventPart,
+  Part,
+} from './lit-html';
+
+export interface DirectiveClass {
   new (part: PartInfo): Directive;
-};
+}
 
 /**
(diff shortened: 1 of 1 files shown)


packages/lit-html/src/directive.ts
```diff
@@ -12,9 +12,17 @@ import {
...
 } from './lit-html';
 
-export type DirectiveClass = {
+export {
+  AttributePart,
+  BooleanAttributePart,
+  ChildPart,
+  EventPart,
+  Part,
+} from './lit-html';
```
rictic, line 21: Hm, was there a reason why we didn't export the parts from directive.ts before?

I guess if they were exported from lit-html then they're not private, so I don't see any reason not to.

rictic (approved):
I don't have any objection, but

packages/lit-html/src/directive.ts
e111077: before in lit1 it was because of a circular dependency which internal build can't handle. Idk why this doesn't trigger the same, but it don't

justinfagnani: I was just defaulting to exporting from one place.

packages/lit-html/src/directive.ts
```diff
@@ -12,9 +12,17 @@ import {
   ElementPart,
 } from './lit-html';
 
-export type DirectiveClass = {
+export {
+  AttributePart,
+  BooleanAttributePart,
+  ChildPart,
+  EventPart,
+  Part,
```
justinfagnani, line 20: I think you need to add `ElementPart`

justinfagnani (approved):

packages/lit-html/src/directive.ts
e111077: I didn't export it because it doesn't exist in Lit1. But that makes sense to do here

Issue #1663, opened 2021-03-11
[reactive-element] Add static initializers list for extending class definitions from decorators


justinfagnani:
We would like to support extending component definitions with third-party decorators without always requiring a new subclass of LitElement.

Since decorators run as class definition time, they don't have access to the class instance. Usually this is solved by having a facility for the decorator to register itself or add some static metadata so that instance-level logic can implement the desired behavior. This is somewhat how `@property` works.

But it's a bit cumbersome to require this for every decorator that needs to add instance behavior. A `@listen()` decorator would need the base class to have support for iterating and adding/removing listeners registered with the decorator, probably in `connectedCallback()` and `disconnectedCallback()`. Similarly with `@observe()` decorator would need custom logic to run in `update()` or `updated()` depending on its specific semantics.

Luckily we now have Reactive Controllers as a way to add lifecycle-dependent helpers externally to the class definition. It's easy to imagine how a ListenController or ObserverController would be implemented. What's missing is a way to instantiate and add those controller to the host.

A very simple and general solution would be to add a `static addInit(this) => void)` method to the class. Decorators can use that to run code on instance creation.

Example with `@listen`:

```ts
export function listen(event: string) {
  return (proto: Object, name: PropertyKey) => {
    (proto.constructor as typeof ReactiveElement).addInit((i: ReactiveElement) => {
      const listener = (e) => proto[name].call(i, e);
      i.addController({
        hostConnected() {
          i.addEventListener(event, listener);
        },
        hostDisconnected() {
          i.removeEventListener(event, listener);
        },
    }));
  }
}
```

sorvell:
Fixed via https://github.com/Polymer/lit-html/pull/1672

PR #1672, opened 2021-03-13
[reactive-element] Adds `addInitializer`

sorvell:
Fixes: #1663.

Allows initializers to be installed which run at element creation. This provides a way for decorators to leverage controllers in order to install features.

Also provides `decorateProperty` which provides an easy way to create a compatible decorator which installs an initializer.


sorvell:
Could use some help with improving the typing.

packages/reactive-element/src/test/decorators/initializer_decorators_test.ts
```diff
@@ -0,0 +1,222 @@
...
+    el.event1 = undefined;
+    el.event2 = undefined;
+    container.removeChild(el);
+    document.body.dispatchEvent(new Event('nug', {bubbles: true}));
+    document.body.dispatchEvent(new Event('zug', {bubbles: true}));
+    assert.isUndefined(el.event1);
+    assert.isUndefined(el.event2);
+  });
+
+  test('can create controller decorator that uses hostUpdate', async () => {
```
justinfagnani, line 111: Do we need a test for every controller callback? Once we verify that an init runs, we know it can call `addController()` and that `addController()` is tested to implement the ReactiveControllerHost protocol.

justinfagnani (approved):

packages/reactive-element/src/test/decorators/initializer_decorators_test.ts
sorvell: Changed test names.

justinfagnani: Do we need the tests at all though? I think only tests for `addInitializer()` are really needed, since an initializer is a callback on the instance and we already have tests that cover the `addController()` instance method.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,30 @@ export const standardPrototypeMethod = (
   key: element.key,
   descriptor,
 });
+
+export const decorateProperty = (
```
kevinpschaaf, line 50: The idea is that this is a TS/babel normalization helper for writing your own property decorators, yeah? 

Can you add API docs for this and when it's intended to be used?

And a few questions: Do we intend end users to use this, or is it just for tests? Can you use this for defining all property decorators? Or just ones that are used in conjunction with `@property`? Is there a reason all of our property decorators couldn't/shouldn't be updated to use this helper for cleaner/smaller code?

packages/reactive-element/src/decorators/base.ts
```diff
@@ -47,8 +47,21 @@ export const standardPrototypeMethod = (
...
+ * @param finisher {(ctor: typeof ReactiveElement, property: PropertyKey) => void)} Optional finisher
+ * @param descriptor {(property: PropertyKey) => PropertyDescriptor} Optional descriptor generator
+ * @returns {ClassElement|void}
+ */
 export const decorateProperty = (
-  decorator: (ctor: typeof ReactiveElement, property: PropertyKey) => void
+  finisher?:
+    | ((ctor: typeof ReactiveElement, property: PropertyKey) => void)
+    | null,
+  descriptor?: (property: PropertyKey) => PropertyDescriptor
```
kevinpschaaf, line 64: Since these options are basically mutually exclusive in practice, API-wise seems better as a single a keyed bag `{finisher: ..., descriptor: ...}`.

kevinpschaaf (changes requested):

packages/reactive-element/src/decorators/base.ts
sorvell: Done.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,60 @@ export const standardPrototypeMethod = (
...
+) => (
+  protoOrDescriptor: ReactiveElement | ClassElement,
+  name?: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): void | any => {
+  if (name !== undefined) {
```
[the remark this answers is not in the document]
sorvell: Done.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,60 @@ export const standardPrototypeMethod = (
...
+export const decorateProperty = (
+  finisher?:
+    | ((ctor: typeof ReactiveElement, property: PropertyKey) => void)
+    | null,
+  descriptor?: (property: PropertyKey) => PropertyDescriptor
+) => (
+  protoOrDescriptor: ReactiveElement | ClassElement,
+  name?: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): void | any => {
```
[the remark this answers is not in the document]
sorvell: Required by TypeScript, added comment.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,60 @@ export const standardPrototypeMethod = (
...
+  if (name !== undefined) {
+    const ctor = (protoOrDescriptor as ReactiveElement)
+      .constructor as typeof ReactiveElement;
+    if (descriptor !== undefined) {
+      Object.defineProperty(protoOrDescriptor, name, descriptor(name));
```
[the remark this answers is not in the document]
sorvell: No, just Babel does that.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,60 @@ export const standardPrototypeMethod = (
...
+    }
+    return finisher?.(ctor, name!);
```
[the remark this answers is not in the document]
sorvell: Removed.

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,66 @@ export const standardPrototypeMethod = (
...
 });
+
+/**
+ * Helper for decorating a property that is compatible with both TypeScript
+ * and Babel decorators. The optional `finisher` can be used to perform work on
+ * the class. The optional `descriptor` should return a PropertyDescriptor
+ * to install for the given property.
+ *
+ * @param finisher {(ctor: typeof ReactiveElement, property: PropertyKey) => void)} Optional finisher
+ * @param descriptor {(property: PropertyKey) => PropertyDescriptor} Optional descriptor generator
```
kevinpschaaf, line 57: Not sure if that's the right way to write jsdoc for destructured params like that...

packages/reactive-element/src/decorators/base.ts
```diff
@@ -44,3 +46,66 @@ export const standardPrototypeMethod = (
...
+            descriptor: descriptor((protoOrDescriptor as ClassElement).key),
+          }
+        : {...(protoOrDescriptor as ClassElement), key};
+    if (finisher != undefined) {
+      info.finisher = function <ReactiveElement>(
+        ctor: Constructor<ReactiveElement>
+      ) {
+        finisher(
+          (ctor as unknown) as typeof ReactiveElement,
+          (protoOrDescriptor as ClassElement).key
```
kevinpschaaf, line 105: This doesn't need to be `key` so it gets the `originalKey ?? key` treatment?

kevinpschaaf (approved):
Check the things noted but otherwise LGTM

packages/reactive-element/src/decorators/base.ts
sorvell: Fixed.

packages/reactive-element/src/decorators/base.ts
sorvell: Edited.

PR #1673, opened 2021-03-14
[localize-tools] Convert localize-tools to JS modules and tape

aomarks:
- Convert from CommonJS to JS modules
- Convert from ava to tape (using https://github.com/vanillaes/tape-es as test runner)
- Refactor config types because `typescript-json-schema` doesn't work properly with JS modules. Also now runs much faster as a side effect.


packages/localize-tools/src/cli.ts
```diff
@@ -72,9 +75,16 @@ export async function runAndLog(argv: string[]): Promise<number> {
   return 0;
 }
 
-function version() {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  return require(path.join('..', 'package.json')).version;
+async function version() {
+  const packageJsonPath = path.resolve(
+    dirname(fileURLToPath(import.meta.url)),
+    '..',
```
justinfagnani, line 81: with resolve() you can do `'../package.json'`

packages/localize-tools/src/cli.ts
```diff
@@ -72,9 +75,16 @@ export async function runAndLog(argv: string[]): Promise<number> {
   return 0;
 }
 
-function version() {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  return require(path.join('..', 'package.json')).version;
+async function version() {
+  const packageJsonPath = path.resolve(
+    dirname(fileURLToPath(import.meta.url)),
```
justinfagnani, line 80: I don't think you need `dirname()`

packages/localize-tools/src/patches.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {Locale} from './locales';
+import {Locale} from './types/locale';
```
justinfagnani, line 7: add `.js`?

justinfagnani (approved):

packages/localize-tools/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
 - Initial release of `@lit/localize-tools` package. This new package provides
   the `lit-localize` binary, while `@lit/localize` continues to provide the
   browser library (`msg`, `LocalizedElement`, etc.).
+
+- **BREAKING** `lit-localize` now uses JS modules instead of CommonJS, so it
```
rictic, line 16: hm. internally we're still on Node 12

aomarks: Internally we'd be consuming the TypeScript directly though, so the way the NPM binary is built doesn't have any effect right?

rictic: oh, true!

packages/localize-tools/src/cli.ts
```diff
@@ -5,16 +5,19 @@
...
-import {TransformLitLocalizer, TransformOutputConfig} from './modes/transform';
-import {RuntimeLitLocalizer, RuntimeOutputConfig} from './modes/runtime';
-
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-require('source-map-support').install();
+import {TransformLitLocalizer} from './modes/transform.js';
+import {RuntimeLitLocalizer} from './modes/runtime.js';
+import {TransformOutputConfig, RuntimeOutputConfig} from './types/modes.js';
+import {dirname} from 'path';
+import {fileURLToPath} from 'url';
```
rictic, line 18: oh nice!

packages/localize-tools/src/locales.ts
```diff
@@ -4,12 +4,11 @@
...
 
-import * as fsExtra from 'fs-extra';
+import fsExtra from 'fs-extra';
 import * as pathLib from 'path';
-import {KnownError} from './error';
-import {escapeStringToEmbedInTemplateLiteral} from './typescript';
-
-export type Locale = string & {__TYPE__: 'Locale'};
+import {KnownError} from './error.js';
+import {Locale} from './types/locale';
```
rictic, line 10: Any reason to leave off the .js here?

packages/localize-tools/src/tests/analysis.unit.test.ts
```diff
@@ -64,6 +65,7 @@ function checkAnalysis(
 test('irrelevant code', (t) => {
   const src = 'const foo = "foo";';
   checkAnalysis(t, src, []);
+  t.end();
```
rictic, line 68: hm, the requirement for an explicit t.end() is awkward and accident prone compared to just awaiting the return value of the test

wrapper function?

packages/localize-tools/src/locales.ts
rictic: Oh I see, it's a .d.ts file, right. Consider `import type` for clarity?

rictic (approved):

packages/localize-tools/src/patches.ts
aomarks: Done.

PR #1675, opened 2021-03-15
Copy d.ts and d.ts.map files from development/ to root after tsc

aomarks:
This allows us to run `tsc` only builds without Rollup. Previously this would fail because `d.ts` files were only copied after a Rollup build. This is useful for e.g. generating API documentation more quickly.

Also, previously we didn't include `d.ts.map` files, and we didn't handle sub-directories inside `development/`.

Also use `--ci` option in `bootstrap` command.


rollup-common.js
```diff
@@ -312,13 +312,6 @@ export function litProdConfig({
         ...(CHECKSIZE
           ? [skipBundleOutput]
           : [
-              // Place a copy of each d.ts file adjacent to its minified module.
-              copy({
-                targets: entryPoints.map((name) => ({
-                  src: `development/${name}.d.ts`,
-                  dest: pathLib.dirname(name),
```
rictic, line 319: I think 'pathLib' might now be defined but never used

packages/internal-scripts/src/treemirror.ts
```diff
@@ -0,0 +1,78 @@
...
+import fastGlob from 'fast-glob';
+import * as fs from 'fs/promises';
+import * as pathlib from 'path';
+
+const unreachable = (x: never) => x;
+
+const main = async () => {
+  const [, , mode, root, dest, ...globs] = process.argv;
+  if (
+    !(mode === 'copy' || mode === 'symlink') ||
```
rictic, line 16: is symlink used anywhere currently?

rictic (approved):

packages/internal-scripts/src/treemirror.ts
aomarks: Removed. I was trying out both approaches, because a symlink works better for a cross-package `tsc --watch` workflow. But we must have copies for npm publish because npm ignores symlinks. We could do both depending on workflow, but it was a bit more complicated, plus was slightly worried about somehow accidentally having the wrong type of file there for publish.

rollup-common.js
aomarks: Removed.

packages/internal-scripts/src/treemirror.ts
```diff
@@ -0,0 +1,48 @@
...
+  const matches = await fastGlob(globs, {cwd: root, absolute: true});
+  const mirrors = [];
+  const dirs = new Set<string>();
+  for (const original of matches) {
+    const mirror = pathlib.resolve(dest, pathlib.relative(root, original));
+    mirrors.push({original, mirror});
+    dirs.add(pathlib.dirname(mirror));
+  }
+
+  await Promise.all([...dirs].map((dir) => fs.mkdir(dir, {recursive: true})));
```
justinfagnani, line 38: copyFile doesn't do this I guess? I think fs-extras' copy() might.

justinfagnani (approved):

packages/internal-scripts/src/treemirror.ts
aomarks: Yeah, `copyFile` doesn't. fs-extra `copy` might, though it's not documented if it does. I'm generally trying to avoid fs-extra nowadays, since `fs/promises` now has *almost* everything I usually need.

PR #1683, opened 2021-03-19
[lit-html] Fix up repeat() signature

justinfagnani:
@aomarks this should allow for better API docs by declaring the overrides explicitly. We might even want to hide the actual signature which is a not-so-useful union of the other two.


kevinpschaaf (approved):
LGTM

sorvell (approved):

PR #1686, opened 2021-03-22
[reactive-element] The `createRenderRoot` method is now called only once.

### LINKED ISSUE #1679: [reactive-element] [lit-next] Appended multiple times within the same microtask causes `createRenderRoot` get called each time
I accidentally stumbled upon this scenario just now. If appended to the DOM multiple times before `performeUpdate()` has been able to execute, `createRenderRoot()` will be called each time. This may be unexpected behavior for subclasses overriding `createRenderRoot()`. An effect of this is when adoptable stylesheets are not supported the style tags are also duplicated multiple times (otherwise the stylesheets are readopted).

I would assume this is unwanted behavior and is probably a vestigial check from a time when `createRenderRoot()` and `firstUpdated(changed)`/`_$didUpdate(changed)` may have had different tasks to perform.

https://github.com/Polymer/lit-html/blob/f0d683b071aad02c423e45ea0d3db8816259759b/packages/reactive-element/src/reactive-element.ts#L735-L739

I propose changing the condition from `!this.hasUpdated` to `!this.renderRoot`. I am open to creating a PR for this (or for any other solution that the maintainers may prefer).

sorvell:
Fixes #1679.

### DIFF
--- a/packages/reactive-element/CHANGELOG.md
+++ b/packages/reactive-element/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Removed -->
 <!-- ### Fixed -->
 
+## Unreleased
+
+### Fixed
+
+- (Since 1.0.0-pre.2) The `createRenderRoot` method is now called only once [#1679](https://github.com/Polymer/lit-html/issues/1679).
+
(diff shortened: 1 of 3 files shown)


kevinpschaaf (approved):
LGTM

justinfagnani (approved):

PR #1689, opened 2021-03-23
[labs] Add @lit-labs/ssr-client and move renderLight

kevinpschaaf:
Moves `renderLight` directive out of lit-html proper and into the labs package.

This new package can then be the home for experimenting with a few follow-on ideas:
* `serverUntil` directive that allows pausing the stream until a promise resolves
* `ServerController` interface that allows a controller to indicate to `LitElementRenderer` to pause rendering of the shadow root until a promise resolves
* etc.

Also fixes the labs packages to include the treemirror script for publishing correct declaration files


packages/labs/ssr-client/tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "target": "es2020",
```
sorvell, line 4: Should be es2019

sorvell (changes requested):
Just one nit. Are the package.json's for some of the other packages updated just for consistency? (maybe add to PR description?)

packages/labs/ssr-client/tsconfig.json
kevinpschaaf: Done

sorvell (approved):
LGTM

PR #1692, opened 2021-03-25
Export part types as type-only, instead of custom `Interface<T>`.

bicknellr:
This PR exports the part classes as type-only, instead of creating a custom `Interface<T>` type.


sorvell (commented):
Seems like a good change. Is there any substantive difference between these other than the name?

```
class Foo {}
export type {Foo}
export type FooThing = Interface<Foo>;
```

bicknellr:
```ts
type Interface<T> = {[P in keyof T]: T[P]};
```

I'm not totally sure but my guess that `Interface<T>` would end up scrubbing some non-property-like info off of the type? Relationship with parent classes, etc.

This might explain why I had to add the cast on that one `instanceof` check. Previously, `PropertyPart` on that line was referring to `export type PropertyPart = Interface<PropertyPartImpl>`, now it's referring directly to the class' type; so maybe it now understands the hierarchy of `AttributePart` well enough to complain that `instancePart` is already sufficiently narrowed by `instancePart instanceof EventPart`?

bicknellr:
Though, the assumption that `instancePart instanceof EventPart` narrows `instancePart` enough doesn't seem 100% correct given that `AttributePart` is in a property of `_Σ` - so somebody else _could_ make a new subclass. Maybe that's not static enough for TS to know about and it just assumes that it knows everything that could extend `AttributePart`, IDK.

sorvell (approved):
LGTM

packages/lit-html/src/lit-html.ts
```diff
@@ -1353,8 +1351,8 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
  * Because event options are passed when adding listeners, we must take case
  * to add and remove the part as a listener when the event options change.
  */
-export type EventPart = Interface<EventPartImpl>;
-class EventPartImpl extends AttributePartImpl {
+export type {EventPart};
```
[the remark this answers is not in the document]
justinfagnani: I bet TypeDoc handles this fine.

packages/lit-html/src/lit-html.ts
```diff
@@ -1485,17 +1483,17 @@ export const _Σ = {
...
   _resolveDirective: resolveDirective,
   // Used in tests and private-ssr-support
-  _ChildPart: ChildPartImpl,
-  _AttributePart: AttributePartImpl as AttributePartConstructor,
-  _BooleanAttributePart: BooleanAttributePartImpl as AttributePartConstructor,
-  _EventPart: EventPartImpl as AttributePartConstructor,
-  _PropertyPart: PropertyPartImpl as AttributePartConstructor,
-  _ElementPart: ElementPartImpl as ElementPartConstructor,
+  _ChildPart: ChildPart,
+  _AttributePart: AttributePart as AttributePartConstructor,
```
justinfagnani, line 1487: Can you try removing these casts? Hopefully exporting the class types makes `typeof C` be a visible type to users of this object.

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -814,35 +814,19 @@ class TemplateInstance {
...
   readonly type: typeof CHILD_PART;
   readonly index: number;
 };
-type ElementPartConstructor = {
-  new (
-    element: HTMLElement,
-    parent: Disconnectable | undefined,
-    options: RenderOptions | undefined
-  ): ElementPart;
-};
```
bicknellr, line 845: After removing the casts, I had to remove `ElementPartConstructor` because that was its only use. I removed `AttributePartConstructor` above also, because its one use seems like it can be replaced with `typeof AttributePart`. WDYT?

packages/lit-html/src/lit-html.ts
bicknellr: Ok, I removed these, left another comment about something related below.

packages/lit-html/src/lit-html.ts
bicknellr: cc @justinfagnani

justinfagnani: Seems fine. As long as everything checks, we're good.

packages/lit-html/src/hydrate.ts
```diff
@@ -382,7 +375,7 @@ const createAttributeParts = (
         // others to avoid perf impact of touching the DOM unnecessarily
         const noCommit = !(
           instancePart instanceof EventPart ||
-          instancePart instanceof PropertyPart
+          (instancePart as unknown) instanceof PropertyPart
```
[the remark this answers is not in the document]
justinfagnani: I'm surprised we're using `instanceof` here. We should be checking `instancePart.type`. @kevinpschaaf

justinfagnani (approved):

packages/lit-html/src/hydrate.ts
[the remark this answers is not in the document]
bicknellr: Updated, PTAL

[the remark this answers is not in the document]
kevinpschaaf: I think the assumption was that since hydrate is the one making the parts, it's safe to use instanceof. However, compiled template results potentially changes that (ctor might come form another version), so definitely a good change to make.

bicknellr:
Once these last tests pass (mod IE11 thing that requires manually checking results), I'll merge.

bicknellr:
All the tests pass, including IE11.

PR #1698, opened 2021-03-25
[lit-html] Skip walking text nodes

justinfagnani:
This changes the template prep and instance prep tree walks to not walk over text nodes. We're able to do this because of the new accurate HTML syntax position scanning, so we don't ever insert attribute-style markers into text content that we need to scan for - we always insert comment markers.

This is showing a small but non-trivial performance improvement on the template-heavy benchmark.

@lit/ssr was updated to the new indexing scheme, which shows that indexing is a public API contract now, especially with compiled template results.


packages/lit-html/src/lit-html.ts
```diff
@@ -637,11 +637,15 @@ class Template {
...
-              (node as Element).append(strings[i] || createMarker());
+              (node as Element).append(strings[i], createMarker());
+              // Walk past the marker node we just added
+              walker.nextNode();
               this.parts.push({type: CHILD_PART, index: ++nodeIndex});
               bindingIndex++;
             }
-            (node as Element).append(strings[lastIndex] || createMarker());
+            // Note the nodeIndex for this marker is accounted for in the
+            // nodeIndex++ expression at the end of the tree walker loop
```
kevinpschaaf, line 647:
```suggestion
            // Note because this marker is added after the walker's current
            // node, it will be walked to (and ignored), so we don't need to
            // adjust nodeIndex here
```

kevinpschaaf (approved):
LGTM, with a small clarification on a comment

sorvell (approved):

justinfagnani:
This has a legit failure on IE right now. I think IE is removing empty text nodes.

jridgewell (approved):
Woooo

PR #1702, opened 2021-03-26
[localize] Replace msgdesc comment with desc option

aomarks:
Replaces `// msgdesc` comments with a new `desc` property on the `msg` options object.

#### Why

- A `msg` option is trivial to parse, and almost impossible for the user to misuse, because of type checking. Comments can be tricky to associate with AST nodes correctly, and silently fail when in the wrong place or typo'd.
- The recursive behavior of `// msgdesc:` was unusual and not worth the complexity.
- This approach will allow future runtime access to the description which, while won't be used by the standard runtime, does open the door for a future dev mode that supports e.g. hovering to see descriptions, or even setting translations directly.

#### Cost

This does regress byte size for runtime mode applications that are currently minified with comment stripping. However, once we have a transform for runtime mode (https://github.com/Polymer/lit-html/issues/1469), we'll be able to remove comments there (in addition to hard-coding IDs so that we can skip template hashing).

#### Before

```js
// msgdesc: Home page
class HomePage {
  hello() {
    // msgdesc: Greeting to Earth
    return msg(html`Hello World`);
  }
  goodbye() {
    // msgdesc: Farewell to Earth
    return msg(html`Goodbye World`);
  }
}
```

#### After

```js
class HomePage {
  hello() {
    return msg(html`Hello World`, {
      desc: 'Home page / Greeting to Earth',
    });
  }
  goodbye() {
    return msg(html`Goodbye World`, {
      desc: 'Home page / Farewell to Earth',
    });
  }
}
```


justinfagnani (approved):

fetherston:
👏👏 you read my mind. I was preparing feedback on the message comments and this new api is exactly what i was going to suggest.

PR #1703, opened 2021-03-30
[lit-html] Ensure directives get full PartInfo

### LINKED ISSUE #1697: [lit-html] `PartInfo`s are not assigned properties from getters in their respective `Part`.
# Description
`AttributePartInfo`'s `tagName` property is always `undefined`.

`PartInfo` objects are created here, by spreading the `Part`:

https://github.com/Polymer/lit-html/blob/be81dc75c23bf40604d175ea0420487d40061ce7/packages/lit-html/src/lit-html.ts#L694-L699

This prevents properties of the `Part` that are retrieved through getters in the prototype chain from being assigned because the spread operator only copies an object's enumerable own properties.

# lit-html version

`2.0.0-pre.5`

# Steps to Reproduce

1. Write this code

```ts
import { html, render } from 'lit-html';
import { Directive, directive, PartType } from 'lit-html/directive.js';

class ExampleDirective extends Directive {
  constructor(partInfo) {
    super(partInfo);

    if (partInfo.type !== PartType.ATTRIBUTE) throw new Error();

    if (typeof partInfo.tagName !== 'string') {
      throw new Error(`partInfo.tagName: ${partInfo.tagName}`);
    }
  }
}

const example = directive(ExampleDirective);

const myTemplate = name =>
  html`
    <h1 some-attr="${example()}">Hello ${name}!</h1>
  `;

render(myTemplate('lit-html'), document.body);
```

2. See this output...

## Live Reproduction Link
https://stackblitz.com/edit/lit-html-next-8x5yle?file=src%2Findex.js

## Expected Results
`AttributePartInfo`'s `tagName` property should be a string.

## Actual Results
`AttributePartInfo`'s `tagName` is `undefined`.

kevinpschaaf:
Ensures that getters are available in PartInfo.

~Using `__proto__` like this isn't the most awesome, but it's likely the least amount of code. We don't want to test each type and copy different fields over selectively. `Object.assign(Object.create(part), {...})` is equivalent, just a bit more code.~

~For background on why we need to make a copy, we added the extra directive-specific information into `PartInfo` (particularly the _$parent) to hide that detail from users so they can just do `super(partInfo)` in the constructor and the extra fields can opaquely get to the directive instance.~

Edit: Both `__proto__` and `Object.create` approaches caused significant performance regressions on kitchen-sink benchmark. Switched implementation to add an internal `_$initialize()` method called by `resolveDirective` after the directive constructor to pass the internal fields. This turned out to be a big initial render perf improvement over the current approach of copying the `Part` to `PartInfo`.

Note when adding tests for SSR, I hit an annoying TODO re: properties not properly reflecting, so I went ahead and added the full reflected property map and crossed off the TODO.

Fixes #1697


sorvell (commented):
Based on the benchmark results, seems like we should try `Object.create` + `Object.assign`.

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1494,6 +1494,27 @@ suite('lit-html', () => {
       assertContent('<div>TEST:A</div>');
     });
 
+    test('PartInfo includes metadata for directive in ChildPart', () => {
```
sorvell, line 1497: Would this test have previously broken?

kevinpschaaf: No, just for completeness.

sorvell (approved):
LGTM.

What was the gambit for making `parent` a setter and is this still worth pursuing?

packages/lit-html/src/lit-html.ts
```diff
@@ -720,7 +716,8 @@ function resolveDirective(
  * An updateable instance of a Template. Holds references to the Parts used to
  * update the template instance.
  */
-class TemplateInstance {
+export type TemplateInstance = Interface<TemplateInstanceImpl>;
```
justinfagnani, line 719: Don't we not need to do this? Is `export type {TemplateInstance}` ok?

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -691,26 +691,22 @@ function resolveDirective(
...
-            ...part,
-            _$part: part,
-            _$parent,
-            _$attributeIndex,
-          } as PartInfo);
-    if (_$attributeIndex !== undefined) {
-      ((_$parent as AttributePart).__directives ??= [])[
-        _$attributeIndex
+        : new nextDirectiveConstructor(part as PartInfo);
+    currentDirective?._$initialize(part, parent, attributeIndex);
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Reverted

PR #1714, opened 2021-04-01
Add `literal` tag to static module.

### LINKED ISSUE #1713: Add `static` tag function to allow non-unsafe static parts of templates
## Description
The `static` module in lit-html provides an `unsafeStatic` function that allows interpolating static text into lit templates. We should also add a tag-function version that need not be labeled "unsafe" that works on template literals.

## Acceptance criteria
`static` export added to `static.ts` and tests that confirm operation

kevinpschaaf:
Unfortunately, we can't use `static` as a function name due to being a reserved keyword. After a short bikeshed, we chose `literal`.

Fixes #1713


justinfagnani (approved):

sorvell (approved):
LGTM

PR #1716, opened 2021-04-02
Lit next readme

kevinpschaaf:


README.md
```diff
@@ -1,36 +1,39 @@
-# lit-html 2.0 monorepo
+# Lit 2.0 Monorepo
```
justinfagnani, line 1: I think we should prep for `lit-next` being moved to `main` and skip the "2.0" and calling it "development monorepo" and such.

This update looks good though, we can do another pass for that.

justinfagnani (approved):

sorvell (approved):

README.md
kevinpschaaf: 👍  Went ahead and made those changes.

PR #1718, opened 2021-04-02
[lit-html] Use export type instead of Impl/Ctors

kevinpschaaf:
This mostly cleans up the code around exporting just the types for parts.

When the values and types of a class are exported separately like this, it doesn't seem like you can import both the value and the type as the same name, although you can declare a type of the same name locally (which is odd). So I think using the `type ImportedClass = InstanceType<typeof importedClass>` pattern seems better than what we had?


kevinpschaaf:
Whoops, didn't see we already had a PR for this.

PR #1722, opened 2021-04-02
[all] Fix errors reported by the internal linter.

bicknellr:
I'm working on getting set up for another import and in my CL the linter pointed out some stuff that seemed useful to upstream. Most of the internal linter complaints are that they don't want `_`-prefixed private properties, so I just ignored those. Also, there are a couple of places where we use `forEach` instead of for-of that it doesn't like, but there were comments indicating that these are intentional.

As far as the stuff it found that seems legit, there are a handful of unnecessary assertions and also a few places where arrow functions that return undefined weren't using a block body. I think they prefer that style for arrow functions because it makes it obvious that the result must be undefined - also, it turned out to improve the formatting in most places IMO.

---

Also, there was one place where a string and number were being concatenated with `+`, so I wrapped the number in `String(...)`, and one place where a class member could be made `readonly`.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -518,7 +518,9 @@ const getTemplateHtml = (
           end
         : s +
           marker +
-          (attrNameEndIndex === -2 ? (attrNames.push(undefined), i) : end);
+          (attrNameEndIndex === -2
+            ? (attrNames.push(undefined), String(i))
+            : end);
   }
 
   // Returned as an array for terseness
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -707,9 +707,9 @@ export abstract class ReactiveElement
     // ensures first update will be caught by an early access of
     // `updateComplete`
     this.requestUpdate();
-    (this.constructor as typeof ReactiveElement)._initializers?.forEach((i) =>
-      i(this)
-    );
+    (this.constructor as typeof ReactiveElement)._initializers?.forEach((i) => {
+      i(this);
+    });
   }
 
   /**
@@ -1082,7 +1082,9 @@ export abstract class ReactiveElement
   // Note, this is an override point for polyfill-support.
   // @internal
   _$didUpdate(changedProperties: PropertyValues) {
-    this.__controllers?.forEach((c) => c.hostUpdated?.());
+    this.__controllers?.forEach((c) => {
+      c.hostUpdated?.();
+    });
     if (!this.hasUpdated) {
       this.hasUpdated = true;
       this.firstUpdated(changedProperties);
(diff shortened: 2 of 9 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -518,7 +518,9 @@ const getTemplateHtml = (
           end
         : s +
           marker +
-          (attrNameEndIndex === -2 ? (attrNames.push(undefined), i) : end);
+          (attrNameEndIndex === -2
+            ? (attrNames.push(undefined), String(i))
```
justinfagnani, line 522: I'd rather not do this. The `+` operator will already do the coercion, so this is just extra bytes. How about a cast?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1082,7 +1082,9 @@ export abstract class ReactiveElement
   // Note, this is an override point for polyfill-support.
   // @internal
   _$didUpdate(changedProperties: PropertyValues) {
-    this.__controllers?.forEach((c) => c.hostUpdated?.());
+    this.__controllers?.forEach((c) => {
+      c.hostUpdated?.();
```
justinfagnani, line 1086: I'm not sure Terser will undo this change, since it can't know that hostUpdate() returns void. I'd rather keep the shorter form.

justinfagnani (commented):
For the lint checks we do want to align on, can we update our linter to catch them too?

packages/reactive-element/src/reactive-element.ts
bicknellr: What would you say to making this a for-of?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -707,9 +707,9 @@ export abstract class ReactiveElement
     // ensures first update will be caught by an early access of
     // `updateComplete`
     this.requestUpdate();
-    (this.constructor as typeof ReactiveElement)._initializers?.forEach((i) =>
-      i(this)
-    );
+    (this.constructor as typeof ReactiveElement)._initializers?.forEach((i) => {
+      i(this);
+    });
```
bicknellr, line 712: This one also seems like it could be a for-of.

packages/reactive-element/src/reactive-element.ts
justinfagnani: Can you check the size impact? Both forms work, I prefer the smaller.

justinfagnani:
ping @bicknellr

bicknellr:
Closing for now. This isn't blocking and doesn't really cover all of the internal linter errors since these were only those pointed out for new code I happened to be importing at the time. If we want to handle these at some point it would probably be better to do a fake import to a new directory so that the linter thinks the entire library is new code and points out all existing errors.

PR #1726, opened 2021-04-04
[localize-tools, ssr] move tests to using uvu

### LINKED ISSUE #1693: [lit-next] Convert node tests to uvu
We currently use tape for ssr and localize node tests. Let's uvu instead.

https://github.com/lukeed/uvu

- uvu doesn't make you awkwardly write `t.end` or `t.plan` in every test.
- uvu comes with a CLI that works with modules, unlike tape which comes with a CLI that doesn't handle modules (so the only way to run tests is by calling them one by one with a direct `node` invocation, or by using https://github.com/vanillaes/tape-es)

43081j:
Fixes #1693 

I hope you don't mind me randomly trying to tackle this. If you already did it, or did it a better way, feel free to close this. I just wanted to understand more of the repo and give uvu a try meanwhile.

I also don't fully understand how you've been making lockfiles with lerna, so its possible the ones i generated are wrong. would be worth a look as they introduced dependencies for other packages in the monorepo which weren't there before.

i also don't know if you were relying on TAP output somewhere, as that will have been lost (uvu has no tap support).

### DIFF
--- a/packages/labs/ssr/package-lock.json
+++ b/packages/labs/ssr/package-lock.json
@@ -9,15 +9,11 @@
       "version": "1.0.0-rc.1",
       "license": "BSD-3-Clause",
       "dependencies": {
-        "@lit-labs/ssr-client": "^1.0.0-rc.1",
         "@webcomponents/template-shadowroot": "^0.1.0",
         "escape-html": "^1.0.3",
         "koa": "^2.7.0",
         "koa-node-resolve": "^1.0.0-pre.5",
         "koa-static": "^5.0.0",
-        "lit": "^2.0.0-rc.1",
-        "lit-element": "^3.0.0-rc.1",
-        "lit-html": "^2.0.0-rc.1",
         "node-fetch": "^2.6.0",
         "parse5": "^5.1.0",
         "resolve": "^1.10.1"
@@ -38,18 +34,14 @@
         "@types/node-fetch": "^2.5.7",
         "@types/parse5": "^5.0.0",
         "@types/resolve": "^1.14.0",
-        "@types/tape": "^4.2.33",
-        "@types/tape-promise": "^4.0.0",
         "@web/test-runner": "^0.12.19",
         "chai": "^4.2.0",
         "command-line-args": "^5.1.1",
         "deepmerge": "^4.2.2",
         "koa-cors": "^0.0.16",
         "mocha": "^7.1.1",
-        "tap-spec": "^5.0.0",
-        "tape": "^4.11.0",
-        "tape-promise": "^4.0.0",
-        "typescript": "^4.1.3"
+        "typescript": "^4.1.3",
+        "uvu": "^0.5.1"
       },
       "engines": {
         "node": ">=13.9.0"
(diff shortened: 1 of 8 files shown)


justinfagnani:
Thanks @43081j! It's great to contribute to an open issue like that :)

This looks good at a scan, we just need to run it locally since our CI permissions won't run all the tests for a non-repo-member PR.

aomarks:
Missed this PR, this is great! Looks like we just need to resolve some merge conflicts from https://github.com/Polymer/lit-html/pull/1741

43081j:
that was a hefty conflict 😂 

@aomarks i may need your help as the two lock files got updated in this PR but one of them looks like such a huge diff, i dont trust its correct...

i basically ran `npx lerna exec --scope=@lit-labs/ssr -- npm i` to try update it. i have no clue how we normally update lock files under lerna

also the tests now failed for some strange chokidar reason im stumped by 😞

aomarks:
> i basically ran `npx lerna exec --scope=@lit-labs/ssr -- npm i` to try update it. i have no clue how we normally update lock files under lerna

I would just delete the `package-lock.json` files and run `npx lerna bootstrap` to generate fresh ones.

43081j:
when i try that, lerna doesn't re-generate them 🤔 

i could go in each one and `npm i` but that produces yet another result

aomarks:
> when i try that, lerna doesn't re-generate them

Try also removing the `node_modules` folders.

There's actually an `npm run nuke` command that's supposed to regenerate everything for this kind of situation, but it doesn't delete the `package-lock.json` files for some reason (feel free to update it in this PR so that it does)

43081j:
`npm nuke` didn't help either, i suppose they were generated by someone manually running `npm i` in there or something. some googling tells me lerna doesn't do it for you

aomarks:
> `npm nuke` didn't help either, i suppose they were generated by someone manually running `npm i` in there or something. some googling tells me lerna doesn't do it for you

That shouldn't be the case, they should be generated by lerna. Can you push your merge and I can try to reproduce?

Another thing to check is that you're on NPM 7, because we're using its new package lock format.

43081j:
i am, and i already did! :D i rebased onto `main`

aomarks:
> i am, and i already did! :D i rebased onto `main`
> 
> thanks for taking a look at this too

`rm -rf node_modules package-lock.json packages/{*,labs/*}/{node_modules,package-lock.json} && npm install && npx lerna bootstrap` works for me for regenerating everything

43081j:
That's so weird. I've always had this problem in lerna repos 

The root lock file updates but it doesn't generate any nested ones. Many lerna repos actually don't have any other than the root I noticed

aomarks:
> That's so weird. I've always had this problem in lerna repos
> 
> The root lock file updates but it doesn't generate any nested ones. Many lerna repos actually don't have any other than the root I noticed

Very weird. https://github.com/Polymer/lit-html/tree/uvu is your branch plus me running the above command to regenerate new package locks.

43081j:
@aomarks i think i found why mine wasn't regenerating them... i was doing `npm run bootstrap` originally which does `npx lerna bootstrap --ci` _not_ `npx lerna bootstrap`

presumably that flag causes it to leave things alone.

looks like i got another big package conflict, too. i guess from a version bump of something/everything (the lit release i bet :D)

trying to solve that now. weirdly though im getting lockfiles now, they're indented with tabs... 🤦‍♂️

43081j:
i've finished the rebase, tests fail. the PR is also huge now because all the lock files updated.

if you could take another look, that'd be super helpful

the tests can't find `lit-html/lit-html.js` which sounds like it may be related to the lock files...

i regenerated them the way you said, and it did make them this time

aomarks:
> if you could take another look, that'd be super helpful

Yeah it's just broken with new package locks, unrelated to your change. Will take a look and let you know when main is updated. Sorry for all the fuss!

LarsDenBakker:
Would be interesting to look into uvu for WTR as well, then we have a nice consistent modern testing set.

43081j:
@LarsDenBakker FWIW my thoughts on it so far:

its assertion library is no better or worse than any other _but_ it ships as an ES module and chai doesn't (chai ships a node-only es module, worthless).

its test library (the main library) is pretty useful, but again its main benefit is its size (which is purely because it has less features tbf) and it ships as an ES module.

the one thing it does badly is assertion exceptions. it throws its own custom exception, which is fine _within uvu_ but completely useless outside uvu. This means if you use uvu in mocha (which you can do), the errors are meaningless when tests fail.

Other than that, its a cool little library due to its size and modular structure

43081j:
@aomarks i rebased onto `main`. now that i can regen package locks we dont necessarily need the commit that does them all, although i guess it would be good to get them in sync

aomarks:
@43081j this should be ready to go for a final merge resolution if you merge with main now that https://github.com/lit/lit/pull/1892 has been merged to fix those test failures.

aomarks:
Oh and note that the `npm run nuke` command no longer updates package locks, so if you want to do a full lock regeneration after merging, run `npm run upgrade` from the root.

43081j:
awesome thanks for the update al

i've rebased onto `main`, dropped the various lockfile commits meanwhile and just regenerated the ones i touched in the end

aomarks (approved):

aomarks:
Thanks for doing this James!

PR #1727, opened 2021-04-05
[lit-next] scoped registries typings (wrong path)

sync:
Was trying to use this new package from typescript and I wasn't able to find typings for it and had a look at what was published on npm for `@lit-labs/scoped-registry-mixin` and `index.d.ts` is definitely not there. The good news is that we are publishing the `.d.ts` but under a different file name.


kevinpschaaf (approved):

kevinpschaaf:
Thank you!

PR #1729, opened 2021-04-05
[labs] When publishing we also need to include some files form the development folder

sync:
As per the exports map

see https://github.com/Polymer/lit-html/issues/1728

# Description
Exports points to a development scoped-registry-mixin.js but it's missing from published npm package.

```
"exports": {
    ".": {
      "development": "./development/scoped-registry-mixin.js",
      "default": "./scoped-registry-mixin.js"
    }
  },
````

# @lit-labs/scoped-registry-mixin
1.0.0-pre.1

# Compilation error

Could not load /Users/anthonymittaz/Projects/DesignSystems/mono/node_modules/@lit-labs/scoped-registry-mixin/development/scoped-registry-mixin.j

### DIFF
--- a/packages/labs/scoped-registry-mixin/package.json
+++ b/packages/labs/scoped-registry-mixin/package.json
@@ -25,6 +25,7 @@
   "files": [
     "/src/",
     "!/src/test/",
+    "/development/scoped-registry-mixin.{d.ts,d.ts.map,js,js.map}",
     "/scoped-registry-mixin.{d.ts,d.ts.map,js,js.map}"
   ],
   "scripts": {
--- a/packages/labs/scoped-registry-mixin/package.json
+++ b/packages/labs/scoped-registry-mixin/package.json
@@ -25,7 +25,8 @@
   "files": [
     "/src/",
     "!/src/test/",
-    "/development/scoped-registry-mixin.{d.ts,d.ts.map,js,js.map}",
+    "/development/",
+    "!/development/test/",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/labs/scoped-registry-mixin/package.json [hunk 1]
```diff
@@ -25,6 +25,7 @@
   "files": [
     "/src/",
     "!/src/test/",
+    "/development/scoped-registry-mixin.{d.ts,d.ts.map,js,js.map}",
```
justinfagnani, line 28:
```suggestion
    "/development/",
    "!/development/test/",
```

So that if any new source files are added they'll get picked up too, and to not publish test sources.

justinfagnani (changes requested):

packages/labs/scoped-registry-mixin/package.json
sync: the reason I didn't do it like so was that there is also a test folder in the development folder (see screenshot)

Screen Shot 2021-04-06 at 6 25 54 am

Let me know if you still want me to change it ?

sync:
@googlebot I fixed it

sync:
@justinfagnani signed the google cla agreement twice and already got commits elsewhere I am not sure what’s happening with it, perhaps co authoring is causing it ?

sync:
This pr is fine with my cla https://github.com/Polymer/lit-html/pull/1727

justinfagnani (approved):
Thanks!

PR #1741, opened 2021-04-13
[labs/ssr] Replace karma with wtr for SSR integration tests

### LINKED ISSUE #1738: [ssr] Migrate Karma tests to Web Test Runner
WTR is used everywhere else and supports modules better than Karma.

kevinpschaaf:
This also tests SSR on Webkit and Firefox now (though not IE, since we don't have polyfills for SSR worked out yet).

Rather than proxy the SSR server through the dev server (which should be possible using `koa-proxies` middleware in the wtr config, although I couldn't seem to get it working), I just changed the code to enable CORS from the SSR server and load from there directly.

The extraneous changes are to get some pesky styling tests working cross-platform, since we'd only been testing on Chrome before.

Fixes #1738


justinfagnani:
Fixes #1738

justinfagnani (approved):

PR #1743, opened 2021-04-14
[lit-next] Rename hydrate & hydrate-support to "experimental"

kevinpschaaf:
We could also consider adding a top-level `console.warn` in `hydrate.js` (since `hydrate-support` also imports it) to the nature of:
> Lit hydration support is experimental. Experimental modules may undergo breaking changes within otherwise non-major releases.

WDYT?


abdonrd:
Isn't everything in the `labs` folder considered experimental?

kevinpschaaf:
> Isn't everything in the labs folder considered experimental?

Yes, however the `lit/hydrate.js` and `lit/hydrate-support.js` modules aren't themselves in labs, and decoupling them enough to move to a separate package would cause us to make more core library symbols public (and non-minifiable) than we'd like.

abdonrd:
> > Isn't everything in the labs folder considered experimental?
> 
> Yes, however the `lit/hydrate.js` and `lit/hydrate-support.js` modules aren't themselves in labs, and decoupling them enough to move to a separate package would cause us to make more core library symbols public (and non-minifiable) than we'd like.

Oh, you're totally right! I don't know why I thought everything was in labs. Thanks!

kevinpschaaf:
Yeah, the server-side SSR module itself is in labs, so you can't use hydrate without using `@lit-labs/ssr`, but as a pragmatic decision we're leaving hydration code in the core, and just want to be clear that we don't consider those API's to be totally stable yet, although the lit package itself will otherwise be.

packages/labs/ssr-client/package.json
```diff
@@ -33,7 +33,7 @@
     "build:watch": "rollup -c --watch",
     "build:ts": "tsc --build && treemirror development . '**/*.d.ts{,.map}'",
     "build:ts:watch": "tsc --build --watch",
-    "clean": "rm -rf {index,hydrate-support,hydrate}.{js,js.map,d.ts} directives/ development/ test/ *.tsbuildinfo",
+    "clean": "rm -rf {index}.{js,js.map,d.ts} directives/ development/ test/ *.tsbuildinfo",
```
sorvell, line 36: Was hydrate there mistakenly before?

sorvell (approved):
Looks good. Just one question.

packages/labs/ssr-client/package.json
kevinpschaaf: Yeah, vestigial from when I tried moving them there earlier, missed removing it.

justinfagnani (approved):

PR #1744, opened 2021-04-14
[labs/ssr] Add global render and ElementRenderer plugin API

kevinpschaaf:
Adds a `render-global` entrypoint which installs the DOM shim onto the nodejs global and exports `render` ready to use.  The test setup code was refactored a bit to run all integration tests both in vm/sandboxed mode and global mode. This should now "just work":

```ts
import {render} from '@lit-labs/ssr/lib/render-global.js';
import {someTemplate} from './some-template.js';
// ...
context.body = Readable.from(render(someTemplate(data)));
```

This PR also adds the ability to add/change ElementRenderers used via an optional `elementRenderers` field in `renderInfo`, which defaults to `[LitElementRenderer]`. The `ElementRenderer` API now includes `static matchesClass()` and `static matchesInstance()` methods for determining when it should be used (and the PR also adds a branding field to `LitElement` for this purpose).

Overriding the renderers might look something like this:
```ts
import {render} from '@lit-labs/ssr/lib/render-global.js';
import {ElementRenderer} from '@lit-labs/ssr/lib/element-renderer.js'
import {someTemplate} from './some-template.js';
// ...
class KaonRenderer extends ElementRenderer {
  static matchesClass(ctor) {
    return ctor.hasKaonBrandingFlag;
  }
  *renderShadow() { /* ... */ }
  // ...
}
// ...
context.body = Readable.from(render(someTemplate(data)), {elementRenderers: [KaonRenderer]});
```


justinfagnani:
@kevinpschaaf can you rebase or merge `main`?

packages/labs/ssr/src/test/integration/client/setup.ts
```diff
@@ -169,164 +172,170 @@ const assertHTML = (
...
-suite('basic', () => {
-  let container: HTMLElement;
-  let observer: MutationObserver;
-  let _mutations: Array<MutationRecord>;
+const modes = ['sandboxed', 'global'] as const;
+export const setupTest = async (
+  tests: SSRTestSuite,
+  testFile: string,
+  mode: typeof modes[number] = 'sandboxed'
+) => {
```
kevinpschaaf, line 180: The diff here is horrible but the main change is just sticking the whole suite creation in a parameterized function and calling it from the actual `*_test.js` files. The `mode` is used to switch which server endpoint is hit (global vs. sandbox) on L247.

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -160,5 +150,57 @@ export const getWindow = (
...
+        },
+      },
+    });
+  }
+
   return window;
 };
+
+export const installWindowOnGlobal = (props: {[key: string]: unknown} = {}) => {
+  if (globalThis !== globalThis.window) {
```
justinfagnani, line 189: Add comment. Is this to make it idempotent?

packages/labs/ssr/src/lib/render-global.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import './dom-shim-global.js';
+export * from './render-lit-html.js';
```
justinfagnani, line 8: I don't think this module or `dom-shim-global` are needed. Because we mutate the global, I'd rather have the importer call the install function so they know they're doing it. It's not much code.

packages/labs/ssr/src/lib/element-renderer.ts
```diff
@@ -8,16 +8,77 @@
...
 /**
  * An object that renders elements of a certain type.
  */
 export abstract class ElementRenderer {
-  constructor(public element: HTMLElement) {}
+  element!: HTMLElement;
+  tagName: string;
+
+  /**
+   * Should be implemented to return true when the given custom element class
```
justinfagnani, line 57: We're going to need some kind of precedence mechanism to work with subclassing. We want to match the most specific renderer we can.

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -15,12 +15,17 @@
...
+ * user SSR code. Includes a minimal shim of DOM APIs.
  *
+ * @param includeJSBuiltIns Whether certain standard JS context globals like
+ *  `console` and `setTimeout` should also be added to the global. Should
+ *  generally only be true when adding window to a sandboxed VM context that
+ *  starts with nothing.
  * @param props Additional properties to add to the window global
  */
 export const getWindow = (
+  includeJSBuiltIns = false,
```
justinfagnani, line 28: optional booleans are hard to read at callsites. Make an options bag?

packages/labs/ssr/src/lib/element-renderer.ts
```diff
@@ -8,16 +8,77 @@
...
+    _tagName: string,
+    _attributes: AttributesMap
+  ) {
+    return false;
+  }
+
+  constructor(tagName: string) {
+    this.tagName = tagName;
+    try {
+      this.element = new (customElements.get(this.tagName))();
```
justinfagnani, line 76: I'm not sure this should be in the base class. The super constructor must be called, and I could imagine renders that don't actually instantiate their elements.

packages/labs/ssr/src/lib/element-renderer.ts
```diff
@@ -8,16 +8,77 @@
...
 import {RenderInfo} from './render-lit-html.js';
 
+export type ElementRendererConstructor = (new (
+  tagName: string
+) => ElementRenderer) &
+  typeof ElementRenderer;
+
+type AttributesMap = Map<string, string>;
+
+export const getElementRenderer = (
```
justinfagnani, line 26: I think we're going to want some kind of renderer registry here so we can store renderers with some kind of precedence info. We could have a registerRenderer method that takes a renderer and a "base" renderer that it supersedes. Or we could offer `registerRenderer(tagName, renderer)` that has the highest precedence.

packages/labs/ssr/README.md
```diff
@@ -51,7 +60,83 @@ const ssrResult = await (renderModule(
...
+### Hydrating lit templates
+
+In order to "hydrate" lit templates, which is the process of having lit re-associate the expressions of a lit template with which nodes they should update in the DOM, the `experimental-hydrate` module is provided in the `lit` package. Prior to updating a server-rendered container using `render`, you should first call `hydrate` on that container using the same template and data that was used to render on the server:
+
+```js
+import {myTemplate} from './my-template.js';
+import {render, hydrate} from `lit/experimental-hydrate.js`;
+
+// Initial hydration required before render:
+const initialData = /* should be same data used to render on the server */;
```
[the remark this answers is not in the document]
kevinpschaaf: We currently require hydrating with same data, because we don't dirty-check or re-set every attribute or textContent during hydration. That's a choice, but because we prime the dirty checking with the values from hydration, they must start off in sync with the DOM. We discussed perf testing and/or adding a switch to allow possibly different data (which would require at least a dirty check of all values during hydration), but that's not yet implemented.

Re: formatting, the idea was this is a top-level module that's loading and hydrating the page on module load, and providing an update function to update it in the future.

packages/labs/ssr/README.md
```diff
@@ -51,7 +60,83 @@ const ssrResult = await (renderModule(
...
+
+Because the `hydrate` function above does not descend into shadow roots, it only works on one scope of the DOM at a time. To hydrate `LitElement` shadow roots, load the `lit/hydrate-support.js` module, which installs support for `LitElement` automatically hydrating itself when it detects it was server-rendered with declarative shadow DOM. This module should be loaded before the `lit` module is loaded, to ensure hydration support is properly installed.
+
+Put together, an HTML page that was server rendered and containing `LitElement`s in the main document might look like this:
+
+```js
+import {render} from '@lit-labs/ssr/lib/render-global.js';
+import './app-components.js';
+
+const ssrResult = render(html`
```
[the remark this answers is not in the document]
kevinpschaaf: I think it's fine for now. In a _typical_ WC-based app, custom elements are the top-level template owners for dynamic content on the page, not _another_ top level non-component actor, so on the server the top-level template is static (or at least, not dynamic on the client; it might use data to render out meta tags or something that won't change on the client). 

I get that this poses a conundrum for e.g. light-dom server rendering because in even when using `renderLight` and the data is coming from a highlander custom element as in `<x-app>${renderLight()}</x-app>`, the directive is invoked by the outer template. But we can get into that once we flesh out those patterns.

Hydrating the top-level template (implying there is dynamic data on the client) also means we'd need to add the `defer-hydration` configurability, because in that case even top-level elements should be deferred. So let's plan on tackling that when we flesh out light-dom rendering examples.

I can add a note for now indicating the outer template in the example is not dynamic and need not be hydrated on the client, but that there are other patterns.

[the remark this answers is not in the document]
justinfagnani: A note's fine for now

packages/labs/ssr/src/lib/dom-shim.ts
kevinpschaaf: This was in there from when I was experimenting with react, or trying to get Polymer running in JSDOM, or something. It's basically to avoid installing a DOM shim if there's already one there, but don't exactly recall where it was needed. Will add a comment.

packages/labs/ssr/README.md
```diff
@@ -1,38 +1,46 @@
 # @lit-labs/ssr
 
-Lit SSR server
+A server package for rendering Lit templates and components on the server.
```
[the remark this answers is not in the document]
kevinpschaaf: Updated

packages/labs/ssr/README.md
```diff
@@ -1,38 +1,46 @@
...
 ## Status
 
-`@lit-labs/ssr` is pre-release software, not quite ready for public consumption. As we develop it we are using it as a test bed to ensure that the next versions of `lit-html` and `lit-element` are SSR-ready.
+`@lit-labs/ssr` is pre-release software, not quite ready for public consumption. As we develop it we are using it as a test bed to ensure that new versions of `lit` (`lit-html` and `lit-element`) are SSR-ready. We expect that the foundational SSR support in this package will support a wide variety of use cases, from full-blown app rendering frameworks built on top of web components, to framework-specific plugins for rending custom elements in e.g. React or Angular, to pre-rendering plugins for static site generators like 11ty. Please stay tuned and file issues with use cases you'd like to see covered.
 
-## Developing
+## Server Usage
 
-### Setup
+### Rendering in the nodejs global scope
```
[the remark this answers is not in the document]
kevinpschaaf: Updated all

packages/labs/ssr/README.md
```diff
@@ -1,38 +1,46 @@
...
+import {render} from '@lit-labs/ssr/lib/render-global.js';
+import {myTemplate} from './my-template.js';
+
+//...
+
+const ssrResult = render(myTemplate(data));
+context.body = Readable.from(ssrResult);
+```
+
+### Sandboxed VM rendering
```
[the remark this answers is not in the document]
kevinpschaaf: 👍  Updated a bunch of places we called it sandbox to just "vm context" or "separate vm context", etc.

packages/labs/ssr/README.md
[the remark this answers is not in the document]
kevinpschaaf: Added note

packages/labs/ssr/src/lib/dom-shim.ts
kevinpschaaf: Yeah, better. Done.

packages/labs/ssr/src/lib/dom-shim.ts
kevinpschaaf: Added comment

packages/labs/ssr/src/lib/element-renderer.ts
kevinpschaaf: Discussed and agreed to handle in the future.

packages/labs/ssr/src/lib/element-renderer.ts
kevinpschaaf: Discussed and agreed to handle in the future.

packages/labs/ssr/src/lib/element-renderer.ts
kevinpschaaf: Good point. Moved back to LitElementRenderer.

packages/labs/ssr/src/lib/render-global.ts
kevinpschaaf: Oh, so I forgot, there's actually a practical issue here related to module ordering, which is what led me down this path in the first place. Because all the module imports get hoisted, you can't import and run `installGlobalDomShim()` _and_ load your component definitions in the same module. So the user would always need to make their own intermediate module like this, just to be able to do "import render & components" in the same module, which should be very common.

I went ahead with:
* `install-global-dom-shim.js`
* `render-with-global-dom-shim.js`

Pretty ugly, but let me know if the above reason makes sense and if it makes you want to have less unwieldily names as a result.

packages/labs/ssr/README.md
```diff
@@ -43,15 +51,99 @@ export const renderTemplate = (someData) => {
...
+// ...
+
+context.body = Readable.from(ssrResult);
 ```
+
+## Client usage
+
+### Hydrating lit templates
+
+In order to "hydrate" lit templates, which is the process of having lit re-associate the expressions of a lit template with which nodes they should update in the DOM, the `experimental-hydrate` module is provided in the `lit` package. Prior to updating a server-rendered container using `render`, you should first call `hydrate` on that container using the same template and data that was used to render on the server:
```
justinfagnani, line 72: You might just want to split this into two sentences: "Hydration is the process.... . In order to hydrate Lit templates..."

packages/labs/ssr/README.md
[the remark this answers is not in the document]
justinfagnani: Ah right, then change "should" to "must"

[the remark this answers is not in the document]
justinfagnani: As for formatting, I thought the statement was `initialData = hydrate(...)`. So maybe make an example data function:
```suggestion
/* must be same data used to render on the server */;
const initialData = getInitialAppData();
```

packages/labs/ssr/README.md
```diff
@@ -43,15 +51,99 @@ export const renderTemplate = (someData) => {
...
+const initialData = /* should be same data used to render on the server */;
+hydrate(myTemplate(initialData), document.body);
+
+// After hydration, render will efficiently update the server-rendered DOM:
+const update = (data) => render(myTemplate(data), document.body);
+```
+
+### Hydrating LitElements
+
+When `LitElement`s are server rendered, their shadow root contents are emitted inside a `<template shadowroot>`, also known as a [Declarative Shadow Root](https://web.dev/declarative-shadow-dom/), a new browesr feature that shipped in [Chrome 90](https://developer.chrome.com/blog/new-in-chrome-90/#declarative). Declarative shadow roots automatically attach their contents to a shadow root on the template's parent element when parsed. For browsers that do not yet implement declarative shadow root, there is a [`template-shadowroot`](https://github.com/webcomponents/template-shadowroot) ponyfill, described below.
```
justinfagnani, line 88: I might remove the reference to Chrome 90

justinfagnani, line 88: I might remove the reference to Chrome 90

justinfagnani (approved):

packages/labs/ssr/README.md
kevinpschaaf: Done

packages/labs/ssr/README.md
kevinpschaaf: Change it to just "shipping in Chrome"

PR #1746, opened 2021-04-14
Fixes #1739

### LINKED ISSUE #1739: [lit-element] ShadyCSS apply shim + polyfill-support.ts breaks when creating multiple instances of the same component.
# Description
When the Shady CSS apply shim is in use and multiple instances of the same component with styles that would be changed by the apply shim are created, the second component's cloned template is different from when the part indices were fetched, causing parts to attach to incorrect nodes. In the example below, the `BooleanAttributePart` for `some-attr` becomes associated with some other text node in the template, causing an error to be thrown when the part attempts to set the attribute.

# Lit versions

lit-element: 3.0.0-pre.4 lit-html: 2.0.0-pre.7

# Steps to Reproduce

1. Write this code...

```html
<!DOCTYPE html>
<body>
<outer-element></outer-element>
<outer-element></outer-element>

<script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.5.0/webcomponents-bundle.js"></script>
<script src="https://unpkg.com/@webcomponents/shadycss@1.10.2/apply-shim.min.js"></script>
<script type="module">
import 'https://unpkg.com/lit@2.0.0-pre.2/polyfill-support.js?module';
import {LitElement, css, html, render} from 'https://unpkg.com/lit@2.0.0-pre.2?module';

customElements.define('outer-element', class extends LitElement {
  render() {
    return html`
      <style>
        div {
          color: red;
        }
        inner-element {
          --some-mixin: {
            color: blue;
          };
        }
      </style>
      Outer element
      <div ?some-attr="${true}">${"Button Text"}</div>
      <inner-element></inner-element>
    `;
  }
});

customElements.define('inner-element', class extends LitElement {
  render() {
    return html`
      <style>
        div {
          @apply(--some-mixin);
        }
      </style>
      <div>inner element</div>
    `;
  }
});

</script>
</body>
```

2. See this output...

An exception with the message "TypeError: this.element.setAttribute is not a function" is thrown.

## Live Reproduction Link
https://jsbin.com/xekiqebojo/edit?html,output

Note: This error is not visible in the JSBin console pane. You need to open the real console.

## Expected Results
No error is thrown.

## Actual Results
An exception with the message "TypeError: this.element.setAttribute is not a function" is thrown.

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari

sorvell:
Under native Shadow DOM, it's critical that any ShadyCSS modified style (this is done only for @apply) be moved to the end of the template so it doesn't disturb part indices.

### DIFF
--- a/packages/lit-element/src/test/lit-element_styling_apply_test.ts
+++ b/packages/lit-element/src/test/lit-element_styling_apply_test.ts
@@ -66,20 +66,90 @@ import {assert} from '@esm-bundle/chai';
       }
     }
     customElements.define(name, E);
-    const el = document.createElement(name) as E;
-    container.appendChild(el);
+    const testInstance = async () => {
+      const el = document.createElement(name) as E;
+      container.appendChild(el);
+      // Workaround for Safari 9 Promise timing bugs.
+      (await el.updateComplete) && (await el.inner!.updateComplete);
+      await nextFrame();
+      const div = el
+        .shadowRoot!.querySelector('x-inner2')!
+        .shadowRoot!.querySelector('div');
+      assert.equal(
+        getComputedStyleValue(div!, 'border-top-width').trim(),
+        '10px'
+      );
+    };
+    await testInstance();
+    await testInstance();
+    await testInstance();
+  });
 
-    // Workaround for Safari 9 Promise timing bugs.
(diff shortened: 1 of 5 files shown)


kevinpschaaf (approved):
LGTM

PR #1747, opened 2021-04-14
Add description and badges to readme

justinfagnani:


kevinpschaaf (approved):
Somehow we got two "Labs" headers on L29; can you remove that? Otherwise looks good.

PR #1750, opened 2021-04-15
Updates from moving lit-next branch to main

justinfagnani:
* Mention lit-html-1.x branch in readme
* Fix old references to lit-next branch throughout repo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,6 @@ If you've already signed a CLA but are still getting bothered by the awfully ins
 ```bash
 git clone https://github.com/Polymer/lit-html.git
 cd lit-html
-git checkout lit-next
 npm install
 npm run bootstrap
 npm run build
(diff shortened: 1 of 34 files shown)


justinfagnani:
@aomarks and @kevinpschaaf I updated a few Tachometer `"ref"` settings from `"lit-next"` to `"main"`. I couldn't see any implicit references to `master` that would need to be updated to `lit-html-1.x`, since it looks like we're benchmarking against releases, but wanted to double check with you two.

aomarks (approved):

aomarks:
> @aomarks and @kevinpschaaf I updated a few Tachometer `"ref"` settings from `"lit-next"` to `"main"`. I couldn't see any implicit references to `master` that would need to be updated to `lit-html-1.x`, since it looks like we're benchmarking against releases, but wanted to double check with you two.

Yeah I think it should be fine, the github action is using the metadata from the PR so the branch shouldn't be hard coded much.

kevinpschaaf (approved):
LGTM. There are real IE failures in there, but they slipped in with https://github.com/Polymer/lit-html/pull/1746. I let Steve know.

PR #1757, opened 2021-04-16
[lit-html] Add and export type HTMLTemplateResult

Matsuuu:
As there is a matching one for SVGTemplateResult, it wouldn't hurt consistency to add one for HTMLTemplateResult

This would prevent having to define types like `export declare const html: (strings: TemplateStringsArray, ...values: unknown[]) => TemplateResult<1>;`


Matsuuu:
@googlebot I signed it!

justinfagnani (approved):

justinfagnani:
Thanks @Matsuuu !

PR #1759, opened 2021-04-16
[lit-html] Handle <template> elements in templates

### LINKED ISSUE #1737: [lit-html] Template literals with parts inside `<template>`s render incorrectly.
# Description

When rendering a template string where some of the parts are inside of a `<template>`, values are rendered into the wrong parts. ([`TemplateInstance#clone()`](https://github.com/Polymer/lit-html/blob/2be1fcf0a82a2bf368295d015613cd87c5ad96f6/packages/lit-html/src/lit-html.ts#L750) doesn't walk `<template>` content to find parts.)

# lit-html version

2.0.0-pre.7

# Steps to Reproduce

1. Write this code

```ts
import {html, render} from 'https://unpkg.com/lit-html@2.0.0-pre.7?module';
render(html`${'A'}<template>${'B'}</template>${'C'}`, document.body);
```

2. See this output...

```
AB
```

## Live Reproduction Link
https://jsbin.com/xojudijari/edit?html,output

## Expected Results
The page shows `AC`. (`'B'` should be rendered to a part inside a `<template>`.)

## Actual Results
The page shows `AB`.

justinfagnani:
Fixes #1737

### DIFF
--- a/packages/lit-html/package.json
+++ b/packages/lit-html/package.json
@@ -58,7 +58,7 @@
     "dev": "scripts/dev.sh",
     "prepublishOnly": "npm run check-version",
     "test": "npm run test:dev && npm run test:prod",
-    "test:dev": "cd ../tests && npx wtr '../lit-html/development/**/*_test.(js|html)'",
+    "test:dev": "cd ../tests && npx wtr '../lit-html/development/**/lit-html_test.js'",
     "test:prod": "MODE=prod npm run test:dev -- '../lit-html/src/test/version-stability_test.js'",
     "test:watch": "npm run test:dev -- --watch"
   },
--- a/packages/lit-html/src/test/lit-html_test.ts
+++ b/packages/lit-html/src/test/lit-html_test.ts
@@ -1873,8 +1873,8 @@ suite('lit-html', () => {
 
       log.length = 0;
       assertRender(
-        html`<div ${count('x', log)} a=${'a'}></div>`,
-        `<div a="a"></div>`
+        html`<div ${count('x', log)} a=${'a'}>${'A'}</div>${'B'}`,
+        `<div a="a">A</div>B`
       );
       assert.deepEqual(log, ['x:1']);
 
--- a/packages/lit-html/src/test/lit-html_test.ts
+++ b/packages/lit-html/src/test/lit-html_test.ts
@@ -2144,74 +2144,73 @@ suite('lit-html', () => {
         assertContent(`<div a="**[B:2:Y]##"></div>`);
       });
     });
-  });
 
-  const disconnectingDirective = directive(
-    class extends AsyncDirective {
-      log!: Array<string>;
-      id!: string;
+    const disconnectingDirective = directive(
+      class extends AsyncDirective {
+        log!: Array<string>;
+        id!: string;
 
-      render(log: Array<string>, id = '', value?: unknown, bool = true) {
-        this.log = log;
-        this.id = id;
-        return bool ? value : nothing;
+        render(log: Array<string>, id = '', value?: unknown, bool = true) {
+          this.log = log;
+          this.id = id;
+          return bool ? value : nothing;
+        }
+
+        disconnected() {
+          this.log.push('disconnected' + (this.id ? `-${this.id}` : ''));
+        }
+        reconnected() {
+          this.log.push('reconnected' + (this.id ? `-${this.id}` : ''));
+        }
       }
+    );
 
-      disconnected() {
-        this.log.push('disconnected' + (this.id ? `-${this.id}` : ''));
+    const passthroughDirective = directive(
+      class extends Directive {
+        render(value: unknown, bool = true) {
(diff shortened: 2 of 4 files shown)


packages/tests/web-test-runner.config.js
```diff
@@ -30,8 +30,8 @@ const browserPresets = {
   // Default set of Playwright browsers to test when running locally.
   local: [
     'chromium', // keep browsers on separate lines
-    'firefox', // to make it easier to comment out
-    'webkit', // individual browsers
+    // 'firefox', // to make it easier to comment out
+    // 'webkit', // individual browsers
```
kevinpschaaf, line 34: Remove?

packages/tests/web-test-runner.config.js
```diff
@@ -198,7 +198,7 @@ export default {
     }),
   ],
   // Only actually log errors and warnings. This helps make test output less spammy.
-  filterBrowserLogs: (type) => type === 'warn' || type === 'error',
+  // filterBrowserLogs: (type) => type === 'warn' || type === 'error',
```
kevinpschaaf, line 201: Remove?

packages/lit-html/package.json
```diff
@@ -58,7 +58,7 @@
     "dev": "scripts/dev.sh",
     "prepublishOnly": "npm run check-version",
     "test": "npm run test:dev && npm run test:prod",
-    "test:dev": "cd ../tests && npx wtr '../lit-html/development/**/*_test.(js|html)'",
+    "test:dev": "cd ../tests && npx wtr '../lit-html/development/**/lit-html_test.js'",
```
kevinpschaaf, line 61: Remove?

And note you can just do `npm run test:dev -- -- '../lit-html/development/**/lit-html_test.js'` on the command line to focus on one test if you don't want to use the CLI GUI for that.

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -391,24 +391,12 @@ suite('lit-html', () => {
...
-      );
-
-      // Currently fails:
-      // render(html`<${'A'}></ ${'A'}>${'B'}`, container);
-      // assert.equal(stripExpressionMarkers(container.innerHTML), '<><!-- -->B</>');
+      assert.throws(() => render(html`<${'A'}></ ${'A'}>`, container));
     });
 
-    test('binding after end tag name', () => {
+    test.skip('binding after end tag name', () => {
```
kevinpschaaf, line 399: Remove skip?

kevinpschaaf (commented):
This looks like it regresses benchmarks on top of code size.

I think I'd prefer marking this as a breaking change, and see if we get strong feedback on people having relied on this for "interesting" use cases enough to justify. The only one we've seen is using it to put comments in CSS in polymer templates via binding expressions (!!!), and for that, they could use `static` expressions.

graynorton:
Just noting that someone called out a specific use case that's not working: https://github.com/lit/lit/issues/1737#issuecomment-829224907

madfriend:
Sorry to be straightforward, but is there any chance on this PR getting merged? I stumbled upon the bug of nested `<template>` and it blocks the migration of our codebase from lit-html@1.2.1 to lit@2.

justinfagnani:
@madfriend we aren't sure. This has a negative perf impact on all templates, whether or not they have bindings in a `<template>` tag. We are considering suggesting that you use the [static expression support](https://lit.dev/docs/templates/expressions/#static-expressions) as a replacement for this functionality, since it scopes the perf impact to one templates.

madfriend:
@justinfagnani Are there any insights on why does it affect performance of all templates, not just the ones with `<template>` ? If I understood the suggested change correctly, we only increase the frontier of examined parts when we come across the template tag, otherwise the logic is pretty much the same. 

As for the static expressions, it just seems odd to use them as a workaround because internally lit-html stopped iterating over `<template>` children. It would mostly certainly be not easy to explain this to newcomers.

justinfagnani:
<template> "children" aren't really children - they're moved to a separate property that we have to check for, so the additional code path and conditionals to check whether the element is a template effect everyone, even if we don't recurse into the template content.

One reason why static makes sense for <template> is that changing the bound values in expressions in <templates> is of limited use - existing DOM created from those templates will not be updated, only the template itself. So if those values usually shouldn't and don't change, static templates optimize for that.

madfriend:
I see your point, but I am concerned that this is a breaking change for those who migrate from lit-html@1.2.1 to lit@2. Shouldn't this change be at least described in the docs (somewhere along the point about expressions in comments)?

justinfagnani:
Yes, I thought we had it listed here: https://lit.dev/docs/releases/upgrade/#lit-html but it look like we missed it.

justinfagnani:
I'm going to close this as too costly in size and perf for the majority of Lit templates that won't be binding inside `<template>`. We're going to recommend using static templates and will write docs to show that.

PR #1760, opened 2021-04-16
[all] Fix Husky after upgrade to v6

justinfagnani:


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
aomarks, line 1: what's this underscore?

aomarks (approved):

.husky/.gitignore
justinfagnani: I don't know, that was generated by the husky 4->6 upgrade script

abdonrd: That's because now Husky create a `.husky/_/husky.sh` file.
Related: https://blog.typicode.com/husky-git-hooks-javascript-config/

PR #1764, opened 2021-04-18
[lit-html] Don't allow classMap to override static classes

### LINKED ISSUE #1278: [lit-html] Clarify behavior of classMap
The test for classMap say that dynamic classes can not override and remove static classes, but this is only true on the first render. Subsequent renders will remove static classes. This is clearly broken, but what is the right behavior? To allow overriding or not? If we're going to change this, 2.0 is the time.

justinfagnani:
Fixes #1278


packages/lit-html/src/directives/class-map.ts
```diff
@@ -42,17 +43,30 @@ class ClassMapDirective extends Directive {
...
-    return Object.keys(classInfo)
-      .filter((key) => classInfo[key])
-      .join(' ');
+    // Add spaces to ensure separation from static classes
+    return (
+      ' ' +
+      Object.keys(classInfo)
+        .filter((key) => classInfo[key])
+        .join(' ') +
+      ' '
```
kevinpschaaf, line 52: The extra spaces are breaking the SSR tests (and the unit tests are failing first; I expect more to fail on the integration tests once the unit tests pass). Need to get those working.

FWIW, I don't feel strongly at all that we should support no-space rendering, but I assume this was supported in 1.x?

kevinpschaaf (changes requested):

packages/lit-html/src/directives/class-map.ts
justinfagnani: separating the classes this way is the right semantic thing to do - we don't want the first class in the classMap to be smooshed together with a static style if the user forgot a space. I'll update the tests

kevinpschaaf: My reasoning was that the "idiom" with lit being strings is string concatenation (you can't do `class="foo bar${list.join(' ')}"` and get something good, so why should `class="foo bar${classMap({dynamic: true})}"`)...?

But it's fine, I see it's needed for consistency at least, since you'd render`bardynamic` initially, but try to remove only `dynamic` on subsequent updates, so it wouldn't be consistent.

packages/lit-html/src/directives/class-map.ts
```diff
@@ -76,7 +90,10 @@ class ClassMapDirective extends Directive {
       // We explicitly want a loose truthy check of `value` because it seems
       // more convenient that '' and 0 are skipped.
       const value = !!classInfo[name];
-      if (value !== this._previousClasses.has(name)) {
+      if (
+        value !== this._previousClasses.has(name) &&
+        !this._staticClasses?.has(name)
```
sorvell, line 95: This is adding runtime overhead to what is potentially a hot-ish path. It doesn't appear to show up on existing benchmarks so seems ok.

packages/lit-html/src/directives/class-map.ts
```diff
@@ -42,17 +43,30 @@ class ClassMapDirective extends Directive {
...
+        .join(' ') +
+      ' '
+    );
   }
 
   override update(part: AttributePart, [classInfo]: DirectiveParameters<this>) {
     // Remember dynamic classes on the first render
     if (this._previousClasses === undefined) {
       this._previousClasses = new Set();
+      if (part.strings !== undefined) {
```
sorvell, line 60: Based on the restrictions that (1) using `classMap` means you can only have 1 dynamic part in the attribute, and (2) attribute bindings set the entire attribute value, this change should make `classMap` do the equivalent of emitting a string of the truthy keys.

sorvell (approved):

kevinpschaaf (approved):

PR #1765, opened 2021-04-18
[scoped-registry-mixin] adopt styles

manolakis:
As the adoption of styles is done by the method createShadowRoot of ReactiveElement, we must also adopt them inside the method createShadowRoot of ScopedRegistryMixin. Otherwise, components using this mixin are not going to have the styles adopted.

### DIFF
--- a/packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
+++ b/packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
@@ -13,11 +13,11 @@
  */
 
 import '@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js';
-import {LitElement, html} from 'lit';
+import {LitElement, html, css} from 'lit';
 import {ScopedRegistryHost} from '../scoped-registry-mixin';
 import {assert} from '@esm-bundle/chai';
 
-class SimpleGreeting extends ScopedRegistryHost(LitElement) {
+class SimpleGreeting extends LitElement {
   private name: String;
 
   static get properties() {
@@ -40,6 +40,14 @@ class ScopedComponent extends ScopedRegistryHost(LitElement) {
     'simple-greeting': SimpleGreeting,
   };
 
+  static get styles() {
+    return css`
+      :host {
+        color: #ff0000;
+      }
+    `;
+  }
+
   render() {
     return html` <simple-greeting
       id="greeting"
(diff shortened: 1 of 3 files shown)


manolakis:
I can't reproduce that error locally and everything seems to build properly.

could be that typescript is having some issues with package exports?

```bash
lerna success run Ran npm script 'build' in 15 packages in 97.7s:
lerna success - lit-benchmarks
lerna success - internal-scripts
lerna success - lit-element
lerna success - lit-html
lerna success - @lit/lit-starter-js
lerna success - @lit/lit-starter-ts
lerna success - lit
lerna success - @lit/localize-tools
lerna success - @lit/localize
lerna success - @lit/reactive-element
lerna success - @lit-labs/react
lerna success - @lit-labs/scoped-registry-mixin
lerna success - @lit-labs/ssr-client
lerna success - @lit-labs/ssr
lerna success - @lit-labs/task
```

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
```diff
@@ -108,4 +116,20 @@ suite('scoped-registry-mixin', () => {
     assert.isFalse($simpleGreeting instanceof SimpleGreeting);
     assert.isTrue($simpleGreeting instanceof HTMLElement);
   });
+
+  test(`simple-greeting should adopt the styles`, async () => {
```
justinfagnani, line 120: nit: I missed this in the previous review, but let's not put the test element name in the test name.

```suggestion
  test(`host element should apply static styles`, async () => {
```

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
```diff
@@ -24,6 +24,14 @@ class SimpleGreeting extends ScopedRegistryHost(LitElement) {
     return {name: {type: String}};
   }
 
+  static get styles() {
```
justinfagnani, line 27: Is there a reason why SimpleGreeting is a ScopedRegistryHost? It doesn't register any of its own children. I wonder if we could get better coverage of scenarios if SimpleGreeting just extended LitElement, and move these styles for the style test to ScopedComponent.

justinfagnani (commented):
Change and test look good, but unrelated to this PR I'd like to see SimpleGreeting extend LitElement so we know that combination works.

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
manolakis: done

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
manolakis: makes sense. I did the change.

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
```diff
@@ -108,4 +116,20 @@ suite('scoped-registry-mixin', () => {
     assert.isFalse($simpleGreeting instanceof SimpleGreeting);
     assert.isTrue($simpleGreeting instanceof HTMLElement);
   });
+
+  test(`host element should apply static styles`, async () => {
+    const $container = document.createElement('div');
```
justinfagnani, line 121: Remove the `$` prefixes

justinfagnani (approved):
LGTM with one small nit on variable names. Thanks!

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
manolakis: done

packages/labs/scoped-registry-mixin/src/test/scoped-registry-mixin_test.ts
manolakis: done

justinfagnani (approved):
Thanks @manolakis !

PR #1767, opened 2021-04-18
[localize] Upgrade to Lit 2 and add @localized decorator

aomarks:
- Upgrades to Lit 2
- Adds `@localized` decorator that adds a Lit 2 controller to the element, to listen for locale changes and update
- Adds `updateWhenLocaleChanges` function for users that don't write decorators
- Removes the `Localized` mixin in favor of above new APIs

#### Before

```ts
import {LitElement} from 'lit-element';
import {Localized} from '@lit/localize/localized-element.js';

class MyElement extends Localized(LitElement) {}
```

###### With decorators

```ts
import {LitElement, customElement} from 'lit';
import {localized} from '@lit/localize';

@localized()
@customElement('my-element');
class MyElement extends LitElement {}
```

###### Without decorators

```ts
import {updateWhenLocaleChanges} from '@lit/localize';

  constructor() {
    super();
    updateWhenLocaleChanges(this);
  }
}
```


packages/localize-tools/src/program-analysis.ts
```diff
@@ -360,7 +359,7 @@ interface Expression {
  * These substitutions are used to delineate template string literal expressions
  * embedded within HTML during HTML parsing in a way that is:
  *
- * [1] Valid anywhere a lit-html expression binding can go without changing the
+ * [1] Valid anywhere a lit expression binding can go without changing the
```
justinfagnani, line 362: I think here and throughout the proper capitalization is going to be "Lit" as in the name of the library, not the np package name. "lit-html` is both, and LitElement is similar to Lit in that we'd refer to it as LitElement even though the package name is "lit-element" (slightly different because LitElement is an actual class, but still).

packages/localize-tools/CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- **[BREAKING]** Lit dependency upgraded to v2.
+
+- **[BREAKING]** Replaces `Localized` mixin transform with `@localized`
```
justinfagnani, line 14: Looks like this should be in the next commit, if you're gong to keep them all.

packages/localize/src/lit-localize.ts
```diff
@@ -441,3 +446,114 @@ function generateId(template: TemplateLike): string {
...
+    elements,
+    finisher(clazz: typeof ReactiveElement) {
+      clazz.addInitializer(updateWhenLocaleChanges);
+    },
+  };
+};
+
+const legacyLocalized = (clazz: typeof ReactiveElement) => {
+  clazz.addInitializer(updateWhenLocaleChanges);
+  return clazz as any;
```
justinfagnani, line 558: Do you need to return the class? TS's decorate helper doesn't require it.

justinfagnani (approved):

PR #1772, opened 2021-04-20
Add TrustedTypes support to lit-next

rictic:
When running in a browser that supports [TrustedTypes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/trusted-types) this will create a lit-html policy and use it for blessing the content of tagged template literals before we hand it off to HTMLTemplateElement#innerHTML for parsing.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -22,6 +22,19 @@ const wrap =
     ? window.ShadyDOM!.wrap
     : (node: Node) => node;
 
+const trustedTypes = (globalThis as unknown as  Partial<Window>).trustedTypes;
+
+/**
+ * Our TrustedTypePolicy for HTML which is declared using the html template
+ * tag function.
+ *
+ * That HTML is a developer-authored constant, and is parsed with innerHTML
+ * before any untrusted expressions have been mixed in. Therefor it is
+ * considered safe by construction.
+ */
+const policy = trustedTypes?.createPolicy(
+    'lit-html', {createHTML: (s) => s, createScript: (s) => s});
+
 /**
  * Used to sanitize any value before it is written into the DOM. This can be
  * used to implement a security policy of allowed and disallowed values in
@@ -241,7 +254,7 @@ export interface CompiledTemplate extends Omit<Template, 'el'> {
   el?: HTMLTemplateElement;
 
   // The prepared HTML string to create a template element from.
-  h: string;
+  h: TrustedHTML;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -22,6 +22,19 @@ const wrap =
     ? window.ShadyDOM!.wrap
     : (node: Node) => node;
 
+const trustedTypes = (globalThis as unknown as  Partial<Window>).trustedTypes;
```
justinfagnani, line 25: did this actually save any bytes?

kevinpschaaf (approved):
LGTM

packages/lit-html/src/lit-html.ts
```diff
@@ -523,18 +536,22 @@ const getTemplateHtml = (
...
-    // ensure the last ChildPart has an end node, hence adding a comment if the
-    // last string was empty.
-    html + (strings[l] || '<?>') + (type === SVG_RESULT ? '</svg>' : ''),
+    htmlResult as TrustedHTML,
     attrNames,
   ];
 };
 
+
+const EMPTY_TEXT = trustedTypes?.emptyScript as unknown as '' ?? '';
```
justinfagnani, line 553: I would go ahead and use `||` here, since this is only run once on startup. Optimize for bytes.

packages/lit-html/src/lit-html.ts
```diff
@@ -523,18 +536,22 @@ const getTemplateHtml = (
           (attrNameEndIndex === -2 ? (attrNames.push(undefined), i) : end);
   }
 
+  let htmlResult: string|TrustedHTML =
+      html + (strings[l] || '<?>') + (type === SVG_RESULT ? '</svg>' : '');
+  if (policy) {
```
justinfagnani, line 541: I would prefer `(policy !== undefined)` if it doesn't add too many bytes. We use `!== undefined` enough that it should compress very well.

justinfagnani (approved):

kevinpschaaf:
Can you add a changelog entry before merging?

packages/lit-html/src/lit-html.ts
rictic: Yes before gzip, no difference after gzip

packages/lit-html/src/lit-html.ts
rictic: Switched to a ternary

packages/lit-html/src/lit-html.ts
rictic: Done

pshihn:
This change is giving me compilation issues in Typescript. See https://github.com/lit/lit/issues/1780

PR #1775, opened 2021-04-20
README tweaks

graynorton:
Slight messaging / presentation tweaks in preparation for launch, since this will be what people see when they look up `lit` on npm.

* Removed version number from heading (I couldn't find any comparable packages that included a version number)
* Grabbed the more official tagline and 1-phrase description from the monorepo
* Lightly edited intro text to de-emphasize the identity of `lit-html` as a separate thing, though resisted the temptation to make more edits further down for now

### DIFF
--- a/packages/lit/README.md
+++ b/packages/lit/README.md
@@ -1,14 +1,18 @@
-# Lit 2.0
+# Lit
 
-Fast, lightweight web components
+## Simple. Fast. Web Components.
+
+Lit is a simple library for building fast, lightweight web components.
+
+At its core is a boilerplate-killing component base class that provides reactive state, scoped styles, and a declarative template system that leads the pack in size, speed, and expressiveness.
 
 ## About this package
 
-The `lit` package contains everything needed to build Lit components with the LitElement base class, lit-html templates, and all first-party lit-html directives.
+The `lit` package contains everything needed to build Lit components: the LitElement base class, Lit templates, and all first-party Lit directives.
 
 Modules:
 
-- `lit`: The main module exports the core pieces needed for component development: `LitElement`, `html`, `css`, and the most
+- `lit`: The main module exports the core pieces needed for component development, including `LitElement`, `html`, and `css`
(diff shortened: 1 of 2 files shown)


packages/lit/README.md
```diff
@@ -1,14 +1,18 @@
-# Lit 2.0
+# Lit
 
-Fast, lightweight web components
+## Simple. Fast. Web Components.
+
+Lit is a simple library for building fast, lightweight web components.
+
+At its core is a boilerplate-killing component base class that provides reactive state, scoped styles, and a declarative template system that leads the pack in size, speed, and expressiveness.
```
aomarks, line 7: "At its core *Lit* is"

graynorton: I agree this phrasing is a bit of a mouthful, but I think it's correct, at least as intended.

Lit is a library. At its core is a base class.

(like: Earth is a planet. At its core is molten lava.)

aomarks: Oh I see! I didn't parse it like that. How about "At Lit's core"?

justinfagnani (approved):

jbingham:
Can we come up with an alternative to "leads the pack", both here and on lit.dev? It seems like a bit too marketing-y, and it's not obvious whether Lit leads, compared to which library, and measured how. It would be better not to invite comparisons and disputes about this. Better would be to just communicate small size, fast, expressive.

graynorton:
> Oh I see! I didn't parse it like that. How about "At Lit's core"?

> Can we come up with an alternative to "leads the pack", both here and on lit.dev? It seems like a bit too marketing-y, and it's not obvious whether Lit leads, compared to which library, and measured how. It would be better not to invite comparisons and disputes about this. Better would be to just communicate small size, fast, expressive.

Sure, I'll add a commit to make these changes.

aomarks (approved):

graynorton:
Will make a separate PR to tweak the same wording on lit.dev.

jbingham (approved):
LGTM

Issue #1752, opened 2021-04-15
[all] Setup GitHub Dependabot or Renovate to automated dependency updates


abdonrd:
Let's consider using GitHub Dependabot or Renovate for regular dependency updates in the new mono-repo.

- GitHub Dependabot: https://docs.github.com/en/code-security/supply-chain-security/about-dependabot-version-updates
- Renovate: https://github.com/marketplace/renovate

We could run the bot on a weekly basis and get PRs, that we would review and merge manually.
An example Renovate config can be found [here](https://github.com/modernweb-dev/web/blob/master/renovate.json).

stephenwade:
Hi @abdonrd! I worked with the Modern Web team to create their Renovate configuration. I'd be happy to help set up a Renovate config for the Lit repo.

I prefer Renovate over Dependabot because Renovate is more customizable. You can change almost every aspect of how Renovate works in your repository. For example, you could do any of these:
- Updates tested on a new branch. If they pass CI, they are automatically merged into the main branch. If not, Renovate opens a PR.
- Minor and patch updates automatically merged if they pass CI, and major updates opened as PRs.
- Group certain packages together into one PR (for example, all packages whose names contain "eslint").
- Run only on certain days of the week or only during certain times of day.
- Customize any setting only for certain updates (based on package name, dependency type, semver type, etc).

In the Modern Web repo, the team opted not to update minor and patch releases. They tried it for a couple weeks but I presume they found it to be too noisy. Their configuration now does not automerge anything and does not update minor or patch versions. It opens PRs for major version releases immediately and for lock file updates once per week.

How would you like your updates to work?

justinfagnani:
Dependabot is set up

PR #1784, opened 2021-04-21
[all] Update dependencies (only minor versions)

abdonrd:
Update all the dependencies.
Only minor versions, to avoid possible breaking changes.

This will help with: https://github.com/lit/lit/issues/1752


abdonrd:
As I just see #1785, maybe you (@rictic @aomarks) are interested in also merge this PR.

PR #1786, opened 2021-04-21
Add lit.dev banner on lit-html site.

arthurevans:
- Also redirect home page. Live on lit-html.polymer-project.org


aomarks (approved):

docs/_includes/version-banner.html
```diff
@@ -0,0 +1,7 @@
+<div class="version-banner">
+  <img src="/images/flame.svg">
+  <p>
+  lit-html is now part of the Lit library–see the new site at  <a href="https://lit.dev">lit.dev</a>.
+  This site documents lit-html 1.0, no longer the current version.
```
justinfagnani, line 5: Do we want to say "no longer the current version" before the final stable release? I think we probably want to flip it to latest on npm before saying that.

arthurevans:
Going to merge this and iterate on the language.

PR #1793, opened 2021-04-22
[labs/motion] add expand-collapse directive

LarsDenBakker:
I have an expand/collapse directive written in lit-html v1, I wanted to see how it would look like in v2 using the new directive API.

This type of animation is complicated because you can't do it using CSS or keyframes, because you want to animate to/from `display: none`. There are some hacks that people do, but it's often faulty or not accessible. The directive API is great for this type of animation.

This is a simple usage example:

```js
      class ExpandCollapseSimple extends LitElement {
        static styles = css`
          :host {
            display: block;
            margin-bottom: 16px;
          }
          .text {
            border: 1px solid black;
            width: 400px;
            padding: 8px;
          }
        `;

        static properties = {
          expanded: {type: Boolean},
        };

        constructor() {
          super();
          this.expanded = true;
        }

        render() {
          return html`
            <button @click=${this._toggleExpanded}>Expand</button>
            Simple example
            <div ${expandCollapse(this.expanded, {duration: 600})}>
              <div class="text">
                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
                eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut
                enim ad minim veniam, quis nostrud exercitation ullamco laboris
                nisi ut aliquip ex ea commodo consequat.
              </div>
            </div>
          `;
        }

        _toggleExpanded() {
          this.expanded = !this.expanded;
        }
      }

      customElements.define('expand-collapse-simple', ExpandCollapseSimple)
```

If you think it's useful, it could be added to labs. If not - that's also fine.


LarsDenBakker:
For SSR support, I don't think an element binding can return anything. It would be great if we can return a list of attributes to be set since we just need to set the styles. I thought I saw an `attr` template as one of the ideas for lit v2, but I can't find it now.

PR #1795, opened 2021-04-22
[all] Remove Polymer references and update links

abdonrd:
Remove a lot of Polymer references and update links


aomarks (approved):
thanks!

PR #1800, opened 2021-04-22
Fix nuke

aomarks:
The nuke script was broken, because `/bin/sh` doesn't necessarily support brace expansion, and because we weren't including `package-lock.json` files.

Also, don't auto-format `package-lock.json` files. Creates pointless churn and is confusing.

### DIFF
--- a/.prettierignore-sync
+++ b/.prettierignore-sync
@@ -7,6 +7,7 @@ packages/{*,labs/*}/{.gitignore,.prettierignore}
 packages/{lit-starter-ts,lit-starter-js}/.eslintignore
 
 [inline]
+package-lock.json
 packages/localize/examples/
 packages/localize-tools/testdata/
 packages/localize-tools/config.schema.json
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "format:prettier": "prettier \"**/*.{cjs,html,js,json,md,ts}\" --write",
     "ignore-sync": "ignore-sync .",
     "lint": "eslint \"**/*.{js,ts}\"",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "format:prettier": "prettier \"**/*.{cjs,html,js,json,md,ts}\" --write",
     "ignore-sync": "ignore-sync .",
     "lint": "eslint \"**/*.{js,ts}\"",
-    "nuke": "rm -rf node_modules packages/{*,labs/*}/node_modules && npm install && npm run bootstrap && npm run clean",
+    "nuke": "rm -rf node_modules package-lock.json && npm install && npx lerna exec 'rm -rf node_modules package-lock.json' && npx lerna bootstrap && npm run clean",
```
rictic, line 16: I think `npx` is unnecessary in an npm script

aomarks: true fixed

.prettierignore-sync
```diff
@@ -7,6 +7,7 @@ packages/{*,labs/*}/{.gitignore,.prettierignore}
 packages/{lit-starter-ts,lit-starter-js}/.eslintignore
 
 [inline]
+package-lock.json
```
rictic, line 10: This is going to result in some really big diffs of all the package-lock.json files as we go from the prettier diff versions to the formatting that npm writes

rictic (approved):

.prettierignore-sync
aomarks: One time yeah. (Also, you read package-lock diffs?)

rictic: sometimes!

PR #1814, opened 2021-04-26
[readme] Switch monorepo readme links to lit.dev and add logo

kevinpschaaf:


justinfagnani (approved):

csvn:
Just a thought about the logo. When using Github with dark mode, it looks like this:

image

It might be nice to have a outline on the black "Lit" text, so that it can be read for those who prefer dark mode.

justinfagnani:
Thanks for pointing that out @csvn!

This logo is already used in the lit package, so I think I'll merge this, then update that file to have a white background.

PR #1816, opened 2021-04-27
[lit-element] Add `this` parameter to the remaining functions in the hydrate module.

bicknellr:
`this` needs to be explicitly typed on these functions to prevent an internal build error.


justinfagnani:
What compiler setting causes an error without those annotations? We should enable it here.

justinfagnani:
Ping @bicknellr is there any way we can turn this rule on in our lint config?

justinfagnani (approved):
LGTM, though I'd really like to see us catch this error before importing.

bicknellr:
This is no longer blocking as we're not importing this internally - this is a completely new feature so there are no users there yet. These will pop up again whenever we do decide to pull this feature in and, like #1722, might be better handled in one big linter sweep.

PR #1817, opened 2021-04-27
Incrimination of kickCount in ReactControllerHost

### LINKED ISSUE #1815: [@lit-labs/react] ReactControllerHost._updated() not called after 2 kicks
# Description
The ReactControllerHost class does not update its self more than 2 times.
@ajmchambers was attempting to produce use the mouse position controller example as a hook in a react component, but the react component is only updating the coordinates once.

Doing some debugging I found this is because after the second kick, ReactControllerHost._updated() is not called. This leaves ReactControllerHost._updatePending set permanently to true, which means all calls to ReactControllerHost.requestUpdate() will fail. (See "Actual Results" section for my log output).

# Steps to Reproduce

1. Start with the usage code provided in [readme](https://github.com/lit/lit/tree/main/packages/labs/react#usage-1). (I changed the name of the react component)

```js
import * as React from 'react';
import {useController} from '@lit-labs/react/use-controller.js';
import {MouseController} from '@example/mouse-controller';

// Write a React hook function:
const useMouse = () => {
  // Use useController to create and store a controller instance:
  const controller = useController(React, (host) => new MouseController(host));
  // return the controller: return controller;
  // or return a custom object for a more React-idiomatic API:
  return controller.position;
};

// Now use the new hook in a React component:
const MousePositionComponent = (props) => {
  const mousePosition = useMouse();
  return (
    <pre>
      x: {mousePosition.x}
      y: {mousePosition.y}
    </pre>
  );
};

export default MousePositionComponent;
```

2. Create a controller such as the following:

```js
export class MouseController {
  host;
  pos = { x: 0, y: 0 };

  _onMouseMove = ({ clientX, clientY }) => {
    this.pos = { x: clientX, y: clientY };
    this.host.requestUpdate();
  };

  constructor(host) {
    this.host = host;
    host.addController(this);
  }

  hostConnected() {
    window.addEventListener("mousemove", this._onMouseMove);
  }

  hostDisconnected() {
    window.removeEventListener("mousemove", this._onMouseMove);
  }
}

```
3. Render it like so:
```jsx
import { StrictMode } from "react";
import ReactDOM from "react-dom";
import MousePositionReact from "./react/MousePosition";

const rootElement = document.getElementById("root");
ReactDOM.render(
  <StrictMode>
    <MousePositionReact />
  </StrictMode>,
  rootElement
);
```
4. Move the mouse and watch it update twice and stop.

## Reproduction Link
Original issue from @ajmchambers:
https://codesandbox.io/s/confident-jackson-l5kl1?file=/src/react/MousePosition.js

My recreation with console logging:
https://codesandbox.io/s/lucid-dust-nq56g?file=/src/react/useController.ts:604-623

Additional recreation with different event:
https://codesandbox.io/s/lingering-smoke-57s1y?file=/src/ClickController.js

## Expected Results
The mouse position keeps updating while the mouse is moved, and every time the component updates, ReactControllerHost._updated() would be called.

## Actual Results
The react component does not update more than 2 times, and ReactControllerHost._updated() stops being called after a couple updates.

Logging output from my reproduction with debugging:
- "request update" - [ReactControllerHost.requestUpdate() before checking _updatePending property. ](https://github.com/lit/lit/blob/e0c6e2c928ea5e50fae2b75f8c56ef4f2f47ad84/packages/labs/react/src/use-controller.ts#L59)
- "starting kick" - [ReactControllerHost.requestUpdate() right before updating _kick. ](https://github.com/lit/lit/blob/e0c6e2c928ea5e50fae2b75f8c56ef4f2f47ad84/packages/labs/react/src/use-controller.ts#L61)
- "update resolved" - [ReactControllerHost._updated() after where the _updatePending proprerty is turned back to false. ](https://github.com/lit/lit/blob/e0c6e2c928ea5e50fae2b75f8c56ef4f2f47ad84/packages/labs/react/src/use-controller.ts#L89)
- "component update" - inside the React component, immedietly before the return statement.

image

Vac1911:
**Fixes #1815** 

The _kickCount property is not linked to the react component, meaning the _kick function, does not change the value of _kickCount. I would consider: using a [proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) or If there is no reason to keep the _kickCount updated: removing the property from ReactControllerHost, and instead setting it to a random hash or timestamp in the component.

_Linted, formatted, and tested._

### DIFF
--- a/packages/labs/react/src/use-controller.ts
+++ b/packages/labs/react/src/use-controller.ts
@@ -32,7 +32,7 @@ class ReactControllerHost<C extends ReactiveController>
   /* @internal */
   _updatePending = true;
   private _resolveUpdate!: (value: boolean | PromiseLike<boolean>) => void;
-
+  // Count of the number of kicks performed
   private _kickCount: number;
   // A function to trigger an update of the React component
   private _kick: (k: number) => void;
@@ -59,7 +59,10 @@ class ReactControllerHost<C extends ReactiveController>
     if (!this._updatePending) {
       this._updatePending = true;
       // Trigger a React update by updating some state
-      microtask.then(() => this._kick(this._kickCount + 1));
+      microtask.then(() => {
+        this._kickCount++;
+        this._kick(this._kickCount);
+      });
     }
   }
 


justinfagnani:
Thanks for the fix! We'll definitely need tests though. I'm happy to do that soon, or you can if you want to.

Vac1911:
I'm not very familiar with testing in the way you are doing it. So if you wouldn't mind taking care of that, I would appreciate it.

justinfagnani:
Fixed with a test in https://github.com/lit/lit/pull/1820

Thanks for this PR though @Vac1911 !

PR #1819, opened 2021-04-28
Prevent `styleMap` from writing to the 'style' attribute.

bicknellr:
Internally, `styleMap` can't update the style attribute dynamically because it isn't allowed by the sanitizer. However, this directive already handles updates using the style property, so this change falls through to use that in all cases.

### DIFF
--- a/packages/lit-html/src/directives/style-map.ts
+++ b/packages/lit-html/src/directives/style-map.ts
@@ -66,10 +66,6 @@ class StyleMapDirective extends Directive {
 
     if (this._previousStyleProperties === undefined) {
       this._previousStyleProperties = new Set();
-      for (const name in styleInfo) {
-        this._previousStyleProperties.add(name);
-      }
-      return this.render(styleInfo);
     }
 
     // Remove old properties that no longer exist in styleInfo


packages/lit-html/src/directives/style-map.ts
```diff
@@ -66,10 +66,6 @@ class StyleMapDirective extends Directive {
 
     if (this._previousStyleProperties === undefined) {
```
kevinpschaaf, line 67: This entire `if` can go away and instead we start with L28 initializing `_previousStyleProperties: Set<string> = new Set();`

kevinpschaaf (commented):
Need to fix TS errors in the build, plus this one nit.

bicknellr:
The tests are failing because of https://github.com/lit/lit/issues/1920.

justinfagnani:
@bicknellr #1920 was about a directive returning `noChange` removing the attribute, while #1921 is about `nothing` removing the attribute. #1921 is intended behavior, but #1920 is not. I think #1921 should be closed and #1920 reopened.

justinfagnani:
Actually, looking into this some... I'm not sure #1920 isn't WAI too. The question is what should render if we get `noChange` on first render? I'll take this to that issue, but here one workaround is to return `undefined` instead of `noChange`. That would emulate any change we might make for #1920 which would be to initialize with `undefined` instead of `nothing`.

bicknellr:
This still has legit failures that I need to look at.

PR #1828, opened 2021-04-30
[labs/react] Omit 'children' from element type in props

### LINKED ISSUE #1826: [labs/react] Children have wrong type when "DOM" library is present in tsconfig.json
# Description

When the `dom` library is part of the `libs` configuration of `tsconfig.json`, a React component created using `createComponent` exposes a wrong type for the components `children` prop, resulting in an error:

```
components don't accept text as child elements. Text in JSX has the type 'string', but the expected type of 'children' is 'HTMLCollection | (HTMLCollection & string) | (HTMLCollection & number) | (HTMLCollection & false) | (HTMLCollection & true) | (HTMLCollection & ReactElement<...>) | (HTMLCollection & ReactNodeArray) | (HTMLCollection & ReactPortal) | undefined'.ts(2747)
```

This can be fixed be removing `dom`. Though, this might not be always the wanted solution.

# Steps to Reproduce

1. Add `dom` to the compiler option `lib` in `tsconfig.json`:

```json
{
    "compilerOptions": {
        "lib": [
            "dom"
        ],
        // ...
    }
}
```

2. Create a React component based on a `LitElement` Web Component:

```ts
import { html, LitElement } from "lit";
import { customElement } from "lit/decorators.js";
import React from "react";

@customElement("my-component")
export class MyComponent extends LitElement {
  render() {
    return html` <div><slot></slot></div> `;
  }
}

export const MyComponentReact = createComponent(
  React,
  "my-component",
  MyComponent
);
```

3. Use the component:

```ts
export default function App() {
  return (
    <div className="App">
      <MyComponentReact>
        Start editing to see some magic happen!
      </MyComponentReact>
    </div>
  );
}
```

4. Problem:

image

5. To resolve the problem momentarily, remove `dom` from the libraries in `tsconfig.json`

## Live Reproduction Link

https://codesandbox.io/s/strange-shadow-ryi9h

(I could not use your stackblitz template, since `tsconfig.json` cannot be modified there)

## Expected Results

The type for the `children` prop is `children?: React.ReactNode`, no matter if `dom` is part of the imported libraries or not.

## Actual Results

```
components don't accept text as child elements. Text in JSX has the type 'string', but the expected type of 'children' is 'HTMLCollection | (HTMLCollection & string) | (HTMLCollection & number) | (HTMLCollection & false) | (HTMLCollection & true) | (HTMLCollection & ReactElement<...>) | (HTMLCollection & ReactNodeArray) | (HTMLCollection & ReactPortal) | undefined'.ts(2747)
```

## Browsers Affected

This issue is not related to a specific browser.

## Further Notes

Interestingly, the type of the imported wrapper component looks as expected, but as soon as it is used within JSX, things go crazy.

image

justinfagnani:
Fixes #1826

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -123,8 +123,10 @@ export const createComponent = <I extends HTMLElement, E>(
 
   // Props the user is allowed to use, includes standard attributes, children,
   // ref, as well as special event and element properties.
+  // TODO: we might need to omit more properties from HTMLElement than just
+  // 'children', but 'children' is special to JSX, so we must at least do that.
   type UserProps = React.PropsWithChildren<
-    React.PropsWithRef<Partial<I> & Events<E>>
+    React.PropsWithRef<Partial<Omit<I, 'children'>> & Events<E>>
   >;
 
   // Props used by this component wrapper. This is the UserProps and the
--- a/packages/labs/react/src/test/create-component_test.tsx
+++ b/packages/labs/react/src/test/create-component_test.tsx
@@ -7,11 +7,14 @@
 import {ReactiveElement} from '@lit/reactive-element';
 import {property} from '@lit/reactive-element/decorators/property.js';
 import {customElement} from '@lit/reactive-element/decorators/custom-element.js';
-import * as ReactModule from 'react';
+import type * as ReactModule from 'react';
 import 'react/umd/react.development.js';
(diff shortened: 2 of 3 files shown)


rictic (approved):

swissmanu:
@justinfagnani is there already a plan for a prerelease of `@lit-labs/react` that will include this change?

PR #1830, opened 2021-05-01
feat: add directives.js to fix #1736

### LINKED ISSUE #1736: [lit-next] Exporting a `directives.js` just like decorators.js for all directives
## Description
- Who wants the functionality
   - Everyone.
- What it is they want
   - For consistency's sake, there is already a `decorators.js` exported that includes all decorators from the `/decorators` folders. There should be a `directives.js` that follows the same so that we can import `lit/directives` just like how we do with `lit/decorators`.
      ```ts
      import { cache } from 'lit/directives';
      // or import { cache } from 'lit/directives.js';
      ```
- Why they want it
   - For consistency's sake and the ease of importing >= 1 directive using just 1 import statement:
      ```ts
      import { cache, repeat } from 'lit/directives';
      // or import { cache, repeat } from 'lit/directives.js';
      ```
- Functional description of task/subtask
   1. Exporting all directives from the `/directives` folder just like `/decorators` in a `directives.js` at the root level.
   1. Include the following export map for `directives.js`:
      ```json
      ...
      "exports": {
         ...
          "./directives.js": {
             "default": "./directives.js"
          },
         ...
      }
      ```

## Acceptance criteria
What the card must do in order to accept it as complete. Acceptance Criteria must be concrete or measurable.

1. Multiple directives can be imported using just 1 import statement with node.js `exports` map
   ```ts
   import { cache, repeat } from 'lit/directives';
   ```
1. Multiple directives can be imported using just 1 import statement with node.js `exports` map and with `.js` extension
   ```ts
   import { cache, repeat } from 'lit/directives.js';
   ```
1. Multiple directives can be imported individually (existing behavior and it should preserve)
   ```ts
   import { cache } from 'lit/directives/cache';
   import { repeat } from 'lit/directives/repeat';

   // or import { cache } from 'lit/directives/cache.js';
   // or import { repeat } from 'lit/directives/repeat.js';
   
   ```

motss:
Fixes #1736.

## Changes
1. Add `directives.js` to re-export all directives from `lit-html/directives`
1. Add `./directives.js` in export map

### DIFF
--- a/packages/lit/package.json
+++ b/packages/lit/package.json
@@ -31,6 +31,9 @@
     "./directive.js": {
       "default": "./directive.js"
     },
+    "./directives.js": {
+      "default": "./directives.js"
+    },
     "./directives/": {
       "default": "./directives/"
     },
--- a/packages/lit/src/directives.ts
+++ b/packages/lit/src/directives.ts
@@ -1,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+ export * from 'lit-html/directives/async-append.js';
+ export * from 'lit-html/directives/async-replace.js';
+ export * from 'lit-html/directives/cache.js';
+ export * from 'lit-html/directives/class-map.js';
+ export * from 'lit-html/directives/guard.js';
+ export * from 'lit-html/directives/if-defined.js';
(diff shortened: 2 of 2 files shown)


justinfagnani:
I don't think we want to do this. The total collection of directives is much more heavyweight than decorators and there are some _very_ rarely used directives that we shouldn't encourage developers to load by default.

PR #1831, opened 2021-05-02
[localize] Refactor to isolate runtime implementations

aomarks:
⚠️ **NOTE** ⚠️: The only real new code in this PR is the `_installMsgImplementation` function and the references to it. Everything else is just moving existing code into new files. Git isn't showing this too well.

### Before
Before this PR,  all of the localize runtime logic was contained in the main `lit-localize.js` module, and it was directly integrated into the `msg` function in a way that wasn't amenable to tree-shaking.

### After
Now, the runtime implementation is factored out, and gets *installed* only when the end-user actually needs the runtime implementation. If an implementation is not installed, the `msg` function is a pure identity function.

(Note that for the time being, to avoid a breaking change, the main module still *exports* the runtime configuration function, but in a future breaking change I'll require importing directly from the new `init/` modules. This change at least unlocks tree-shaking and creates the new organizational structure.)

### Why?

- I'm planning to add an SSR-oriented runtime implementation very shortly, which will make `msg` able to context-switch when rendering multiple locales simultaneously. Since a number of things are different about this implementation, it makes the most sense to have two separate implementations of the `msg` function.

- It's completely valid to use the `msg` function even if you don't actually need to setup the localization runtime. Imagine a re-distributable component library, which uses `msg` to ensure it *could* be localized. Users who don't need localization shouldn't need to pay the cost of the full runtime -- all they need is an identity `msg` function.

- I'm also thinking eventually of implementing a "baked" runtime mode, which will be another kind of transformer output where all IDs are hard-coded, so that hashing isn't required. Again it will be preferable to just have a different `msg` implementation, and to omit all of the hashing code from the import graph.

- It's more readable.


packages/localize-tools/src/modes/transform.ts
```diff
@@ -498,25 +501,15 @@ class Transformer {
...
+    for (const decl of moduleSymbol.declarations) {
       if (
-        props.some(
-          (prop) =>
-            prop.escapedName === '_LIT_LOCALIZE_MSG_' ||
-            prop.escapedName === '_LIT_LOCALIZE_CONTROLLER_FN_' ||
-            prop.escapedName === '_LIT_LOCALIZE_DECORATOR_'
-        )
+        ts.isSourceFile(decl) &&
+        (decl.fileName.endsWith('/localize/lit-localize.d.ts') ||
```
aomarks, line 511: Note to self:

We also need to check for the new `/init/transform.d.ts` file so that if the user imports it that way (which I think will become the required method in an upcoming breaking change), we will remove the import.

We should really have a universal way to notice "any import from the localize package". Maybe we just look for `@lit/localize` in the path -- but this doesn't work for tests because we import directly with a relative path (but maybe we can add some special handling for that).

packages/localize/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
+- Removed accidental export of `_msg` (use `msg` instead).
```
justinfagnani, line 12: Do you want to add a mention of `msg()` being a no-op by default?

packages/localize/src/lit-localize.ts
```diff
@@ -351,209 +50,5 @@ export interface MsgOptions {
...
-      clazz.addInitializer(updateWhenLocaleChanges);
-    },
-  };
-};
-
-const legacyLocalized = (clazz: typeof ReactiveElement) => {
-  clazz.addInitializer(updateWhenLocaleChanges);
-  return clazz as any;
-};
+export const msg = ((template: TemplateLike, options?: MsgOptions) =>
```
justinfagnani, line 53: I don't know if you even need this redirection, unless you want to ensure that the configured msgImpl can't be called with different arguments or can have properties assigned to it.

justinfagnani (approved):

packages/localize/src/lit-localize.ts
aomarks: I think the indirection is required because otherwise we'd export the default implementation once, and then it could never be replaced.

justinfagnani: Exports are live bindings, but I guess could store `msg` in a variable before using? Some bundlers may make you opt-into live-binding support since it requires making all exports getters.

packages/localize/CHANGELOG.md
aomarks: There's no change here, it's always been this way. There just used to be an extra export of the same thing called `_msg`.

packages/localize/src/lit-localize.ts
aomarks: Ah! I should have known that. Done.

PR #1854, opened 2021-05-06
[reactive-element] Rename Warnings type to WarningKind. Add docs.

justinfagnani:
`Warnings` should have been named as a singular noun phrase.

I kept the old name too, though we might want to get rid of it as the number of people relying on it is probably close to 0.

Also added some docs.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -268,7 +268,13 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
  */
 const finalized = 'finalized';
 
-export type Warnings = 'change-in-update' | 'migration';
+/**
+ * A string representing on of the supported dev mode warnings classes.
+ */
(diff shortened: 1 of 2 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -268,7 +268,13 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
  */
 const finalized = 'finalized';
 
-export type Warnings = 'change-in-update' | 'migration';
+/**
+ * A string representing on of the supported dev mode warnings classes.
```
aomarks, line 272: One of

aomarks (approved):

justinfagnani:
@sorvell wdyt about removing the old `Warnings` name?

sorvell (approved):

sorvell:
Oh, we should add this to the changelog.

PR #1856, opened 2021-05-07
[lit-html] Clone templates such that they are inert while parts are created.

### LINKED ISSUE #1848: [lit-html] Part/node association can break if elements edit their own light DOM during construction.
# Description

The process of associating parts with nodes in [`TemplateInstance#_clone()`](https://github.com/lit/lit/blob/0ba0ab472fd8294d604b5ce8bc89569b8c8828bd/packages/lit-html/src/lit-html.ts#L777) can be broken by a custom element that edits its own light DOM during construction. `TemplateInstance#_clone()` uses `Document#importNode` to clone its fragment, which causes any cloned custom elements to upgrade synchronously. Any changes those custom elements make to the structure of the cloned fragment will happen before `TemplateInstance#_clone()` has had the opportunity to associate parts with nodes, which invalidates the indices used for this purpose and causes parts to become associated incorrectly.

# Steps to Reproduce

```ts
import {html, LitElement} from 'lit';

class OuterElement extends LitElement {
  static get properties() {
    return {
      count: {type: Number},
    };
  }

  constructor() {
    super();

    this.count = 0;

    for (let i = 1; i <= 3; i++) {
      setTimeout(() => this.count++, 1000 * i);
    }
  }

  render() {
    return html`
      <inner-element></inner-element>
      <b>count: ${this.count}</b>
    `;
  }
}

customElements.define('outer-element', OuterElement);

class InnerElement extends LitElement {
  constructor() {
    super();

    for (let i = 0; i < 5; i++) {
      const div = document.createElement('div');
      div.innerText = `Inserted div #${i + 1}.`;
      this.appendChild(div);
    }
  }

  render() { return html`<slot></slot>`; }
}

customElements.define('inner-element', InnerElement);
```

Then, put `<outer-element></outer-element>` somewhere on the page.

## Live Reproduction Link

https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4LnRzIiwiY29udGVudCI6ImltcG9ydCB7aHRtbCwgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcblxuY2xhc3MgT3V0ZXJFbGVtZW50IGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHN0YXRpYyBnZXQgcHJvcGVydGllcygpIHtcbiAgICByZXR1cm4ge1xuICAgICAgY291bnQ6IHt0eXBlOiBOdW1iZXJ9LFxuICAgIH07XG4gIH1cblxuICBjb25zdHJ1Y3RvcigpIHtcbiAgICBzdXBlcigpO1xuXG4gICAgdGhpcy5jb3VudCA9IDA7XG5cbiAgICBmb3IgKGxldCBpID0gMTsgaSA8PSAzOyBpKyspIHtcbiAgICAgIHNldFRpbWVvdXQoKCkgPT4gdGhpcy5jb3VudCsrLCAxMDAwICogaSk7XG4gICAgfVxuICB9XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYFxuICAgICAgPGlubmVyLWVsZW1lbnQ-PC9pbm5lci1lbGVtZW50PlxuICAgICAgPGI-Y291bnQ6ICR7dGhpcy5jb3VudH08L2I-XG4gICAgYDtcbiAgfVxufVxuXG5jdXN0b21FbGVtZW50cy5kZWZpbmUoJ291dGVyLWVsZW1lbnQnLCBPdXRlckVsZW1lbnQpO1xuXG5jbGFzcyBJbm5lckVsZW1lbnQgZXh0ZW5kcyBMaXRFbGVtZW50IHtcbiAgY29uc3RydWN0b3IoKSB7XG4gICAgc3VwZXIoKTtcblxuICAgIGZvciAobGV0IGkgPSAwOyBpIDwgNTsgaSsrKSB7XG4gICAgICBjb25zdCBkaXYgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdkaXYnKTtcbiAgICAgIGRpdi5pbm5lclRleHQgPSBgSW5zZXJ0ZWQgZGl2ICMke2kgKyAxfS5gO1xuICAgICAgdGhpcy5hcHBlbmRDaGlsZChkaXYpO1xuICAgIH1cbiAgfVxuXG4gIHJlbmRlcigpIHsgcmV0dXJuIGh0bWxgPHNsb3Q-PC9zbG90PmA7IH1cbn1cblxuY3VzdG9tRWxlbWVudHMuZGVmaW5lKCdpbm5lci1lbGVtZW50JywgSW5uZXJFbGVtZW50KTtcbiJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9pbmRleC5qc1wiPjwvc2NyaXB0PlxuPC9oZWFkPlxuPGJvZHk-XG4gIDxvdXRlci1lbGVtZW50Pjwvb3V0ZXItZWxlbWVudD5cbjwvYm9keT4ifV0

## Expected Results

The page should show five lines with the text "Inserted div #[some number]." followed by one line that says "count: 0". The value next to "count:" should update once per second for a few seconds.

## Actual Results

There is no value following the text "count:". Instead, the value appears after "Inserted div #2." and updates also render into that position.

## Browsers Affected

- Chrome
- Firefox
- Edge (Chromium)
- Safari 14

bicknellr:
This prevents custom elements inside the template from upgrading and potentially editing the cloned tree's structure before parts are associated with nodes in that tree, which can cause them to be associated with the wrong node.

- [ ] Tests

Fixes #1848.


justinfagnani:
Let's decide we want to do this first...

bicknellr:
+1, I think it might be interesting to see what happens to the benchmarks.

packages/lit-html/src/lit-html.ts
```diff
@@ -316,11 +316,11 @@ export interface RenderOptions {
...
   /**
-   * Node used for cloning the template (`importNode` will be called on this
-   * node). This controls the `ownerDocument` of the rendered DOM, along with
-   * any inherited context. Defaults to the global `document`.
+   * The node into which the template clone will be adopted. This controls the
+   * `ownerDocument` of the rendered DOM, along with any inherited context.
+   * Defaults to the global `document`.
    */
-  creationScope?: {importNode(node: Node, deep?: boolean): Node};
+  creationScope?: {adoptNode: Document['adoptNode']};
```
bicknellr, line 323: Maybe `Node & {adoptNode: Document['adoptNode']}` would be better here.

justinfagnani:
> +1, I think it might be interesting to see what happens to the benchmarks.

good point!

packages/lit-html/src/lit-html.ts
```diff
@@ -811,7 +811,7 @@ class TemplateInstance {
         nodeIndex++;
       }
     }
-    return fragment;
+    return (options?.creationScope ?? d).adoptNode(fragment);
```
justinfagnani, line 814: Here is where we used to call `customElements.upgrade()` too. I think we'll want some way to have a reference to a registry from the creationScope.

cc @kevinpschaaf @manolakis

bicknellr: > used to call `customElements.upgrade()`

Oh, that might be why the tests are failing.

bicknellr:
#1848 was closed as won't fix.

PR #1857, opened 2021-05-07
[reactive-element] Make unsafeCSS() use the CSSResult cache

### LINKED ISSUE #1846: [reactive-element] Should `unsafeCSS` use the `cssResultCache`?
# Description

Should multiple calls to `unsafeCSS` with the same content return the same shared `CSSResult`?
Currently, in Lit 2, this is the case for `css` tagged template literals, but not `unsafeCSS` calls. This means that ``css`${unsafeCSS(rules)}` `` will benefit from the cache (at the expense of some string concatenation) but not a direct call: `unsafeCSS(rules)`, besides both returning an _equivalent_ `CSSResult` object.

## Why is this important?

In Chromium browsers, a `CSSResult` will own a constructed stylesheet, sharing a `CSSResult` means constructing only one stylesheet, and as result parsing the CSS string only once. When that string is big, constructing the stylesheet could take dozens of milliseconds (tested in Chrome with CPU throttling ×4 and a 220K long string , constructing the stylesheet takes a bit more than 200ms; this quickly adds up when several LitElements use that same big chunk of CSS)

I was directed to file this issue on Slack: https://lit-and-friends.slack.com/archives/C03PF4L4L/p1620234012193100?thread_ts=1620232184.189500&cid=C03PF4L4L

# Steps to Reproduce

1. Write this code

```ts
import {css, unsafeCSS} from 'lit';

const styles = "* { outline: red dotted 1px; }";

console.log(css`* { outline: red dotted 1px; }` === css`* { outline: red dotted 1px; }`); // → true
console.log(unsafeCSS(styles) === unsafeCSS(styles)); // → false
console.log(css`${unsafeCSS(styles)}` === css`${unsafeCSS(styles)}`); // → true
console.log(css`* { outline: red dotted 1px; }` === unsafeCSS(styles)); // → false
console.log(css`* { outline: red dotted 1px; }` === css`${unsafeCSS(styles)}`); // → true
console.log(css`* { outline: red dotted ${unsafeCSS(42 - 41)}px; }` === css`${unsafeCSS(styles)}`); // → true
```

2. See this output... (see comments next to the `console.log()` above)

## Live Reproduction Link

https://stackblitz.com/edit/lit-html-t2raux?file=src/index.js

## Expected Results

Open the "Console" on the right pane of StackBlitz; there should be 6 lines with `true`.

## Actual Results

See "steps to reproduce" above for the actual results.

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

justinfagnani:
Fixes #1846

Also adds a test that `unsafeCSS()` is a valid top-level style. This means that the recommended pattern for including external CSS text can be `static styles = unsafeCSS(cssText)`, without using the `css` tag.

The PR also updates the `css` tag caching to have a fast path for strings with no interpolations.

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -53,15 +53,17 @@ export class CSSResult {
   }
 }
 
-/**
- * Wrap a value for interpolation in a [[`css`]] tagged template literal.
- *
- * This is unsafe because untrusted CSS text can be used to phone home
- * or exfiltrate data to an attacker controlled site. Take care to only use
- * this with trusted input.
- */
-export const unsafeCSS = (value: unknown) => {
-  return new CSSResult(String(value), constructionToken);
+const cssResultCache = new Map<string, CSSResult>();
+
+const getCSSResult = (cssText: string): CSSResult => {
+  let result = cssResultCache.get(cssText);
+  if (result === undefined) {
+    cssResultCache.set(
+      cssText,
+      (result = new CSSResult(cssText, constructionToken))
+    );
+  }
+  return result;
 };
 
 const textFromCSSResult = (value: CSSResultGroup | number) => {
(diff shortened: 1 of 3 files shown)


tbroyer (approved):

sorvell (approved):
LGTM

PR #1858, opened 2021-05-07
[wip] Reduce cache overhead of static templates.

bmeurer:
Following up on comments in https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/2876969 here's a proof-of-concept to reduce the performance overhead of using static.ts

The major issue that seems to stand out still is the tests for `_$litStatic$` which are going to pollute the engine's ICs. I'd suggest to maybe go with a `WeakMap` / `WeakSet` approach here instead.

cc @paullewis

### DIFF
--- a/packages/lit-html/src/static.ts
+++ b/packages/lit-html/src/static.ts
@@ -23,7 +23,7 @@ export const unsafeStatic = (value: string) => ({
   _$litStatic$: value,
 });
 
-const textFromStatic = (value: StaticValue) => {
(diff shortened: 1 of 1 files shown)


augustjk:
Merged in current main, resolving conflicts. Seems like this implementation is failing stability checks when using same tag name, e.g. https://github.com/lit/lit/blob/5c8b142552542ffa775b74074b8bd16f427a00fa/packages/lit-html/src/test/static_test.ts#L95-L96

PR #1869, opened 2021-05-10
[lit-html] Add `PropertyPart` to directives forward-compat file.

bicknellr:
This PR separates `PropertyPart` from `AttributePart` in the forward-compatible directive API. `PropertyPart` and `AttributePart` are separate types in both [1.x](https://github.com/lit/lit/blob/39a00486fee64277a630666fde08cddc9a2402c9/src/lib/parts.ts#L448) and [2.x](https://github.com/lit/lit/blob/f50dff4d9066c91cbca12c3b746dd8f7d3fb21e5/packages/lit-html/src/lit-html.ts#L1336).

### DIFF
--- a/src/directive.ts
+++ b/src/directive.ts
@@ -108,7 +108,7 @@ class PropertyPart extends AttributePart {
   constructor(legacyPart: legacyLit.PropertyPart) {
     super(legacyPart);
   }
-};
+}
 
 export type{BooleanAttributePart};
 class BooleanAttributePart {
--- a/src/directive.ts
+++ b/src/directive.ts
@@ -104,10 +104,6 @@ class AttributePart {
 export type{PropertyPart};
 class PropertyPart extends AttributePart {
   readonly type = PartType.PROPERTY;
-
-  constructor(legacyPart: legacyLit.PropertyPart) {
-    super(legacyPart);
-  }
 }
 
 export type{BooleanAttributePart};
--- a/src/directive.ts
+++ b/src/directive.ts
@@ -40,7 +40,8 @@ export interface AttributePartInfo {
   readonly tagName: string;
 }
 
-export type Part = ChildPart|AttributePart|BooleanAttributePart|EventPart;
+export type Part =
+    ChildPart|AttributePart|PropertyPart|BooleanAttributePart|EventPart;
 
 export type{ChildPart};
 class ChildPart {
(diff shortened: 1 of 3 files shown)


src/directive.ts
```diff
@@ -194,9 +202,9 @@ function legacyPartToPart(part: legacyLit.Part): Part {
     return new EventPart(part);
   } else if (part instanceof legacyLit.BooleanAttributePart) {
     return new BooleanAttributePart(part);
-  } else if (
-      part instanceof legacyLit.PropertyPart ||
-      part instanceof legacyLit.AttributePart) {
+  } else if (part instanceof legacyLit.PropertyPart) {
+    return new PropertyPart(part);
```
justinfagnani, line 206: Was this a bug that we never created a PropertyPart? Do we have test coverage here?

bicknellr:
It's exported here at least:

https://github.com/lit/lit/blob/8134face22ec3350ca7209587be16964d8c25b2d/packages/lit-html/src/lit-html.ts#L1335-L1336

src/directive.ts
bicknellr: I'm not sure if there was / wasn't a reason it wasn't originally included. I can't find tests for this file.

bicknellr: I've added a basic set of tests that check that `.type` of the `PartInfo` given to the directive's constructor is the expected value for each type of part. For `PropertyPart` specifically, there's nothing else to test.

src/directive.ts
```diff
@@ -96,9 +98,15 @@ class AttributePart {
...
-        PartType.PROPERTY :
-        PartType.ATTRIBUTE;
+  }
+}
+
+export type{PropertyPart};
+class PropertyPart extends AttributePart {
+  readonly type = PartType.PROPERTY;
+
+  constructor(legacyPart: legacyLit.PropertyPart) {
```
justinfagnani, line 108: fwiw, you don't need to write this constructor, even though leaving it out won't change the emit since the class needs a field initializer.

justinfagnani (dismissed):

src/directive.ts
bicknellr: Thanks, removed

bicknellr:
I think my push reset your review status.

justinfagnani (approved):

PR #1870, opened 2021-05-11
Adds decorators for `@compute`, `@observe`, and `@listen`

sorvell:
* Adds `@computed` decorator for creating computed properties. Decorator specifies an array of deps which are the names of dependent properties and a compute function which is called with dependency values and should return the computed property value.

* Adds `@observe` decorator for observing property changes. Decorator specifies an observe function which receives the current and previous value of the property as well as the name of the property. Note, the observed property must be a reactive property.

* Adds `@listen` decorator for declaratively adding event listeners. The decorator should be placed on an element property which is an event listener function. The decorator must provide an object which specifies the event `type`, `options`, and an optional `target`. The `target` defaults to the element itself or may be `root` to add the listener to the element's
 `renderRoot`, or an instance of `EventTarget`. If `target` is an `EventTarget`,
the listener is dynamically added and removed when the element is connected and disconnected.

### DIFF
--- a/packages/reactive-element/src/decorators/listen.ts
+++ b/packages/reactive-element/src/decorators/listen.ts
@@ -1,0 +1,165 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * IMPORTANT: For compatibility with tsickle and the Closure JS compiler, all
+ * property decorators (but not class decorators) in this file that have
+ * an @ExportDecoratedItems annotation must be defined as a regular function,
+ * not an arrow function.
+ */
+
+import {ReactiveElement} from '../reactive-element.js';
+import {decorateProperty} from './base.js';
+
+interface ListenInfo {
+  type: string;
+  prop: PropertyKey;
+  options?: EventListenerOptions;
+  target: '' | 'root' | EventTarget;
+}
+
+type Listeners = ListenInfo[];
+
+const hostListeners: WeakMap<typeof ReactiveElement, Listeners> = new WeakMap();
+const rootListeners: WeakMap<typeof ReactiveElement, Listeners> = new WeakMap();
+const targetedListeners: WeakMap<
+  typeof ReactiveElement,
+  Listeners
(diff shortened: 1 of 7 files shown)


packages/reactive-element/src/decorators/listen.ts
```diff
@@ -0,0 +1,165 @@
...
+ * from the DOM.
+ *
+ * @example
+ * ```ts
+ * class MyElement {
+ *   @listen('click')
+ *   private _clickHandler(e: Event) => {
+ *     this.clickedOn = e.target.localName;
+ *   }
+ *   @listen('keydown', window, {capture: true}))
```
jpzwarte, line 105: I noticed that the docs do not match the function definition. 

`listen('keydown', window, { capture: true })` 
vs 
`listen({ type: 'keydown', options: { capture: true }, target: window })`

abdonrd:
Any news with this?

abdonrd:
@sorvell what is the status of this? Thanks!

WickyNilliams:
Curious if there is still a plan to add these decorators? They would be very handy, and i expect many component libraries end up implementing these themselves.

I implemented an `observe` decorator in my design system today. I tried to adapt the one in the PR, but hit issues where `this` would be undefined inside the callback e.g.

```ts
@property()
@observe(() => {
  this.doSomething() // error `this` is undefined
})
foo: number = 0
```

I guess maybe the code is no longer compatible with Lit 2?

abdonrd:
Also interested on this!

tehbiga:
Sharing interest!
These are so far the major missing elements in Lit for me!

a11delavar:
If anyone is interested in these features (excluding `@computed`), check out the [lit wrapper](https://www.npmjs.com/package/@a11d/lit) I created. It contains [@eventListener](https://github.com/a11delavar/lit/tree/main/packages/Lit/eventListener) and [@updated](https://github.com/a11delavar/lit/blob/main/packages/Lit/updated) decorators.

justinfagnani:
Closing this for now, though we probably want to revisit these separately or as parts of signals integration.

PR #1876, opened 2021-05-12
Update lit-html.ts

Muffle10:
Saves around 6 bytes of data and allows enability.


justinfagnani:
Thanks for the PR, but these are build-time constants that are compiled out of the code in our various prod builds. They are not intended to be configurable. There is no code savings in prod builds because they are removed.

Issue #1834, opened 2021-05-02
[ssr] events are not registered


baboonmaj:
```json
"lit": "2.0.0-rc.1",
"@lit-labs/ssr": "^1.0.0-rc.1",
```

lit-element with a button and an event associated with it
```js
// my-element.js
import { LitElement, html, css } from 'lit'
export class MyElement extends LitElement {
  static get properties() {
    return {
      count: { type: Number }
    }
  }

  constructor() {
    super()
    this.count = 1;
  }

  onClick() {
    this.count = this.count + 1;
    this.requestUpdate();
  }

  render() {
    return html`
    <div style="padding: 16px; width:200px; border: 1px solid #777;">
      <button @click=${this.onClick}>
        Click Count: ${this.count}
      </button>
    </div>
    `
  }
}
window.customElements.define('my-element', MyElement)
```

which I am pre-rendering to `index.html`
```js
import { render } from "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { html } from "lit";
import "./my-element.js";
import { createWriteStream } from 'fs';

const generate = async (options) => {
  var wStream = createWriteStream('index.html');  
  const mainHtml = html`<my-element> </my-element>`

  const ssrResult = render(mainHtml);
  for await (const str of ssrResult) {
    wStream.write(str);
  }
  wStream.end();
}
export default generate;
```

The generated `index.html` contains just a button without any event handlers attached to it
```html
<!--lit-part fGXCEZN6E3o=--><my-element><template shadowroot="open"><!--lit-part YACaZRGRe3k=-->
    <div style="padding: 16px; width:200px; border: 1px solid #777;">
      <button ><!--lit-node 1-->
        Click Count: <!--lit-part-->1<!--/lit-part-->
      </button>
    </div>
    <!--/lit-part--></template></my-element><!--/lit-part-->
```

fgirardey:
I experienced the same issue this weekend, i don't know if its a known issue but i'll try to reproduce it on a dedicated repository 😊

kevinpschaaf:
This is working as intended, at least for the current approach to SSR. Event handlers are not serialized into the DOM, they will only be attached when `hydrate` is called on the client (done automatically when including `experimental-hydrate-support.js`.

We will be exploring patterns in SSR that allow for lazy hydration of DOM trees without missing events, but this is not currently in scope for SSR.

PR #1881, opened 2021-05-12
[ssr] Add global mode demo

aomarks:
Adds a global mode ssr demo. Previously there was only a vm-modules demo. The new demo is exactly the same as the other one, but modified to use global mode instead.

```
cd packages/labs/ssr
npm run demo:vm-modules
npm run demo:global
```


kevinpschaaf (approved):
LGTM

baboonmaj:
Sorry for spamming this area, but Thanks for putting up this demo.

I am trying to pre-render as opposed to ssr and like to have a little bit of help here.
have created an issue #1834 , but didnt get any response yet so trying if i can get lil help here

I like to prepare a `html` in advance which would also contain events. I was able to render the html using the ssr package, but not sure how the events will get associated with the elements.
Do i need to do something with hydrate ?

Can you please comment in #1834 and provide some pointers, and I can dig it further.

cc @aomarks @kevinpschaaf 

Thanks

Issue #1896, opened 2021-05-14
[localize] error in generated templates with multiple placeholder


jlesquembre:
# Description

Generated files (with `lit-localize extract/build`) are mixing the placeholders in the template

# Steps to Reproduce

1. Given this template:

```ts
  render() {
    return msg(html`
      <h1>Hello, ${this.name} ${this.surname}!</h1>
      <div>Current language: ${this.lang}</div>
      <button @click=${this._toogleLang} part="button">Toogle language</button>
      <slot></slot>
    `);
```
[source](https://github.com/jlesquembre/lit-element-localize-issue/blob/4121a6805e69d00ada87a73e95df6ad94490cda5/src/my-element.ts#L43-L49)

2. Those files are generated:
 
```xml
  <source>
      <ph id="0">&lt;h1></ph>Hello, <ph id="1">${this.name} ${this.surname}</ph>!<ph id="2">&lt;/h1>
      &lt;div></ph>Current language: <ph id="3">${this.lang}&lt;/div>
      &lt;button @click="${this._toogleLang}" part="button"></ph>Toogle language<ph id="4">&lt;/button>
      &lt;slot>&lt;/slot>
    </ph></source>
```
[source](https://github.com/jlesquembre/lit-element-localize-issue/blob/4121a6805e69d00ada87a73e95df6ad94490cda5/xliff/es.xlf#L6-L11)

```ts
    export const templates = {
      'hee42ce2388b5d52a': html`
      <h1>Hola, ${1} ${1}!</h1>
      <div>Idioma: ${3}</div>
      <button @click="${3}" part="button">Cambiar idioma</button>
      <slot></slot>
    `,
```
[source](https://github.com/jlesquembre/lit-element-localize-issue/blob/4121a6805e69d00ada87a73e95df6ad94490cda5/src/locales/es.ts#L11-L17)

Full code is here: https://github.com/jlesquembre/lit-element-localize-issue (I took [lit-element-starter-ts](https://github.com/PolymerLabs/lit-element-starter-ts) as a starting point)

You can run `npm run i18n:extract` and `npm run i18n:build` to regenerate the locale files, and `npm run build`, `npm run serve` to serve the site at http://localhost:8000/dev

## Expected Results

I expect to see `Hola John Doe!` if I select the `es` locale.

## Actual Results

I see `Hola Doe Doe!`

## Browsers Affected

I only tested it with Chrome, but I guess all browsers are affected

aomarks:
[...]

PR #1898, opened 2021-05-14
[localize] Fix duplicate and out-of-order localized template values

aomarks:
Fixes two bugs relating to localized template values:

1. When we generated localized templates, we replace expressions with numbers, representing the position of that value in the source template, to account for the possibility of expressions changing position (see https://github.com/lit/lit/pull/1625 for more details). However, we were incorrectly indexing using the position of the `<ph>` placeholder that *contained* the expression, instead of the index of the expression itself (note that a `<ph>` can contain any combination of HTML and zero or more placehlders). So the offsets could be wrong, and were sometimes duplicated.

2. We use the special string `_START_LIT_LOCALIZE_EXPR_${RANDOM}_` as a delimiter during internal template decomposition, but a `replace` operation was accidentally only matching the first occurrence, instead of all occurrences. So a template like `<a href="${url}/${base}">` previously contained this literal string.

Fixes https://github.com/lit/lit/issues/1896

### DIFF
--- a/packages/localize-tools/src/modes/runtime.ts
+++ b/packages/localize-tools/src/modes/runtime.ts
@@ -157,19 +157,19 @@ function generateLocaleModule(
     entries.push(`'${msg.name}': ${msgStr},`);
   }
   return `
-     // Do not modify this file by hand!
-     // Re-generate this file by running lit-localize
+    // Do not modify this file by hand!
+    // Re-generate this file by running lit-localize
 
-     ${importLit ? "import {html} from 'lit';" : ''}
-     ${importStr ? "import {str} from '@lit/localize';" : ''}
+    ${importLit ? "import {html} from 'lit';" : ''}
+    ${importStr ? "import {str} from '@lit/localize';" : ''}
 
-     /* eslint-disable no-irregular-whitespace */
-     /* eslint-disable @typescript-eslint/no-explicit-any */
+    /* eslint-disable no-irregular-whitespace */
+    /* eslint-disable @typescript-eslint/no-explicit-any */
 
-     export const templates = {
-       ${entries.join('\n')}
-     };
-   `;
+    export const templates = {
+      ${entries.join('\n')}
+    };
+  `;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/localize-tools/src/modes/runtime.ts
```diff
@@ -157,19 +157,19 @@ function generateLocaleModule(
...
+     /* eslint-disable @typescript-eslint/no-explicit-any */
 
-    export const templates = {
-      ${entries.join('\n')}
-    };
-  `;
+     export const templates = {
+       ${entries.join('\n')}
+     };
+   `;
```
rictic, line 172: intentional change? this indentation seems a little worse

packages/localize-tools/testdata/build-runtime-xliff/goldens/xliff/es-419.xlf
```diff
@@ -39,28 +39,22 @@
...
+<trans-unit id="h8d70dfec810d1eae">
+  <source><ph id="0">&lt;b></ph>Hello<ph id="1">&lt;/b></ph>! Click <ph id="2">&lt;a href="${urlBase}/${urlPath}"></ph>here<ph id="3">&lt;/a></ph>!</source>
+  <target><ph id="0">&lt;b></ph>Hola<ph id="1">&lt;/b></ph>! Clic <ph id="2">&lt;a href="${urlBase}/${urlPath}"></ph>aquí<ph id="3">&lt;/a></ph>!</target>
+</trans-unit>
 <trans-unit id="s03c68d79ad36e8d4">
   <note>Description of 0</note>
   <source>described 0</source>
   <target>described 0</target>
 </trans-unit>
-<trans-unit id="s03c68e79ad36ea87">
```
rictic, line 54: What's with dropping these `<trans-unit>` elements?

rictic (approved):

packages/localize-tools/src/modes/runtime.ts
aomarks: Don't know why that happened, undone.

Issue #1880, opened 2021-05-12
[all] Published `.d.ts.map` files have invalid `sources` field


aomarks:
Since we just copy `.d.ts.map` files from `development/` into the root, they have an incorrect relative path in the `sources` field.

E.g. for `lit/index.d.ts.map` we have `../src/index.ts` instead of `./src/index.ts`:

```json
{"version":3,"file":"index.d.ts","sourceRoot":"","sources":["../src/index.ts"],"names":[],"mappings":"AAAA;;;;GAIG;AAKH,OAAO,uBAAuB,CAAC;AAC/B,OAAO,UAAU,CAAC;AAElB,cAAc,4BAA4B,CAAC"}
```

(https://unpkg.com/browse/lit@2.0.0-rc.2/index.d.ts.map)

nicolejadeyee:
@aomarks hey Al, is this still an issue? Closing, please reopen if you see fit.

nicolejadeyee:
@aomarks is this still an issue?

aomarks:
> @aomarks is this still an issue?

Yes :)

PR #1899, opened 2021-05-15
[internal-tools] treemirror adjusts the sources array in *.map files to be relative to their new location

ajmchambers:
This relates to issue #1880.

Goal is for treemirror to adjust the sources array in the *.map files it's copying to be relative to their destination location.

```
/root
  /development
    /index.d.ts.map
    /folder/index.d.ts.map 
  /src
    /index.ts
    /folder/index.ts
```

After running in root `treemirror development . \"**/*.d.ts.map"`

```
/root
  /index.d.ts.map
  /folder/index.d.ts.map
```

sources array in `/root/index.d.ts.map` should have changed from: `['../src/index.ts']` to: `['./src/index.ts']`

sources array in `/root/folder/index.d.ts.map` should have changed from: `['../../src/folder/index.ts']` to: `['../src/folder/index.ts']`

Appears to be working for me on both windows and mac but would definitely appreciate another pair of eyes on it to make sure its working as expected.

### DIFF
--- a/packages/internal-scripts/package-lock.json
+++ b/packages/internal-scripts/package-lock.json
@@ -1,377 +1,508 @@
 {
-	"name": "internal-scripts",
-	"version": "1.0.0",
-	"lockfileVersion": 2,
-	"requires": true,
-	"packages": {
-		"": {
-			"version": "1.0.0",
-			"license": "BSD-3-Clause",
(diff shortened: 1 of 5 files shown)


ajmchambers:
This is old and probably irrelevant now so closing

PR #1903, opened 2021-05-17
[all] Update Prettier to v2.3.0

abdonrd:
Also fix the version, as https://github.com/webcomponents/polyfills/pull/449


package.json
```diff
@@ -33,7 +33,7 @@
     "ignore-sync": "^3.1.0",
     "lerna": "^4.0.0",
     "lint-staged": "^10.4.0",
-    "prettier": "^2.1.2",
+    "prettier": "2.3.0",
```
justinfagnani, line 36: Can we get a version range here like `^2.3.0`?

justinfagnani: Hmm... I see Prettier's advice, but it seems like a package-lock solve that just fine.

abdonrd: I fixed it as the @bicknellr recommended here https://github.com/webcomponents/polyfills/pull/449#issue-646113858:

>Apparently Prettier doesn't follow semver and may produce different formatting between updates within the same major version - they recommend pinning to an exact version: https://prettier.io/docs/en/install.html#summary . The recent 2.3.0 update changes the formatting in this project.

abdonrd: Updated!

PR #1904, opened 2021-05-18
Cast `window` to a new type (instead of extending it) to prevent collisions.

bicknellr:
This PR changes the `env.d.ts` files that extend the `Window` interface to create a new interface instead and cast `window` to that type whenever anything needs to read one of these extra properties. This prevents the types for the polyfill interfaces from conflicting internally. I've also removed the `ShadowRoot` property because it's already part of `Window`.


packages/lit-element/src/env.d.ts
```diff
@@ -30,10 +30,9 @@ interface ShadyDOM {
   wrap: (node: Node) => Node;
 }
 
-interface Window {
+interface LitExtraGlobals {
```
justinfagnani, line 33: If this interface extends `Window` you won't need the casts to `unknown`.

I can't quite get that to work though: https://www.typescriptlang.org/play?#code/FAFwngDgpgBA4gGwPYCMCGCYF4YHUCWAdgCZIDuMAZDONEgGYwDmy6CAKgBb4DOA3MGBEQUAE700AY1gES5AKIAPEKLQ8YUZVBLrEqDDADewGDHpIkALhiEArgFsUYgQF9BxKJIRpRsAG4+MGREpGTWsqFKKmoCwAAUwXIUanghCsqqPACUAHTmSAJAA

packages/lit-element/src/polyfill-support.ts
```diff
@@ -48,12 +48,14 @@ interface PatchableLitElement extends HTMLElement {
 }: {
   LitElement: PatchableLitElement;
 }) => {
+  const extraGlobals = window as unknown as LitExtraGlobals;
```
justinfagnani, line 51: Does this fully inline for a 0 byte change?

bicknellr: I'm not sure. If it doesn't, would you prefer that each use is individually casted?

packages/lit-element/src/env.d.ts
bicknellr: Yeah, I'm having trouble with this too. If I just say `extends Window` and remove the `as unknowns`, then I get this:

```
src/polyfill-support.ts(97,24): error TS2352: Conversion of type 'Window & typeof globalThis' to type 'LitExtraGlobals' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

But I can't seem to write whatever it _should_ extend properly:

```ts
interface LitExtraGlobals extends (Window & typeof globalThis) { ... }
```

```
src/env.d.ts(33,35): error TS2499: An interface can only extend an identifier/qualified-name with optional type arguments.
```

bicknellr: Oh, this works:

```ts
const extraGlobals = window as LitExtraGlobals & typeof globalThis;
```

It's longer but it doesn't have `unknown`. WDYT?

bicknellr: I just pushed some changes that extend `Window` here and make a new type to add the `globalThis`-ness into it so that `window` can be casted directly, PTAL.

packages/lit-element/src/polyfill-support.ts
bicknellr: The change is not fully inlined. Doing this seems to have made a couple of these files a bit smaller because `window` is now being aliased to a single letter variable in a few places:

Screen Shot 2021-05-18 at 18 45 21

(I couldn't think of a better way to show a character diff besides a screenshot...)

What do you typically do to compare sizes across changes?

justinfagnani: We have a checksize script for that. Looks like in the core files most references here are guarded in dev mode checks, so we're ok. Smaller is better too :)

justinfagnani (approved):

bicknellr:
I restarted the tests because the IE11 ones had more failures than the expected "timed out at the very end" problem.

bicknellr:
The new ones were also timeouts, but I'm not sure if they were newly introduced here somehow or if it was just a bad run.

bicknellr:
Ok, after restarting the tests this is only failing due to the usual timeouts in IE11 after running everything.

PR #1916, opened 2021-05-20
[reactive-element] Change css template tag to return CSSResult

### LINKED ISSUE #1894: [reactive-element] Revert css`` return type to CSSResult
We have a few too many users accessing `cssText` on the result of the `css` tag. The change to returning CSSResultGroup was intended to prevent TypeScript from narrowing the type of `static styles` and make subclassing easier, but this might not be panning out how we intended. It seems more pragmatic to roll back this change for now.

In the future we may want to deprecate `.cssText` and prepare for a time when `css` can return a CSSStyleSheet directly.

justinfagnani:
Fixes #1894

Also removes a couple of unnecessary type aliases, mainly just so we're not required to support those exports publicly.

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -102,7 +102,7 @@ export const unsafeCSS = (value: unknown) => {
 export const css = (
   strings: TemplateStringsArray,
   ...values: (CSSResultGroup | number)[]
-): CSSResultGroup => {
+): CSSResult => {
   const cssText =
     strings.length === 1
       ? strings[0]
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -15,21 +15,36 @@ export const supportsAdoptingStyleSheets =
   'adoptedStyleSheets' in Document.prototype &&
   'replace' in CSSStyleSheet.prototype;
 
+/**
+ * A CSSResult or native CSSStyleSheet.
+ *
+ * In browsers that support constructible CSS style sheets, CSSStyleSheet
+ * object can be used for styling along side CSSResult from the `css`
+ * template tag.
+ */
 export type CSSResultOrNative = CSSResult | CSSStyleSheet;
-
-export type CSSResultFlatArray = CSSResultOrNative[];
 
 export type CSSResultArray = Array<CSSResultOrNative | CSSResultArray>;
 
+/**
+ * A single CSSResult, CSSStyleSheet, or an array or nested arrays of those.
+ */
 export type CSSResultGroup = CSSResultOrNative | CSSResultArray;
 
 const constructionToken = Symbol();
 
+/**
+ * A container for a string of CSS text, that may be used to create a CSSStyleSheet.
+ *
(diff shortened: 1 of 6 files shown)


kevinpschaaf (approved):
I think this looks right, but maybe we should get an LGTM from @bicknellr ?

justinfagnani:
Oops, I had a bad rebase before pushing and lost the critical change here. I'm going to fix and force push so there's two easily scannable commits.

justinfagnani:
@kevinpschaaf can you PTAL with the correct commits? The first is what actually changes the return type of `css`. @bicknellr  this is just a revert to lit-element 2.x types, should be good.

packages/reactive-element/src/css-tag.ts
```diff
@@ -15,21 +15,36 @@ export const supportsAdoptingStyleSheets =
...
-
-export type CSSResultArray = Array<CSSResultOrNative | CSSResultArray>;
-
-export type CSSResultGroup = CSSResultOrNative | CSSResultArray;
+/**
+ * A single CSSResult, CSSStyleSheet, or an array or nested arrays of those.
+ */
+export type CSSResultGroup =
+  | CSSResultOrNative
+  | Array<CSSResultOrNative | CSSResultGroup>;
```
bicknellr, line 32: I think this is the same:
```suggestion
export type CSSResultGroup = CSSResultOrNative | Array<CSSResultGroup>;
```

packages/reactive-element/src/css-tag.ts
```diff
@@ -15,21 +15,36 @@ export const supportsAdoptingStyleSheets =
...
+ *
+ * In browsers that support constructible CSS style sheets, CSSStyleSheet
+ * object can be used for styling along side CSSResult from the `css`
+ * template tag.
+ */
 export type CSSResultOrNative = CSSResult | CSSStyleSheet;
 
-export type CSSResultFlatArray = CSSResultOrNative[];
-
-export type CSSResultArray = Array<CSSResultOrNative | CSSResultArray>;
```
bicknellr, line 22: `CSSResultArray` is exported in the old version and there are people using this internally - not tons though.

https://github.com/lit/lit-element/blob/2b398727dacb7694fbf38816824675d20838704f/src/lit-element.ts#L82-L83

packages/reactive-element/src/css-tag.ts
```diff
@@ -94,15 +116,17 @@ export const unsafeCSS = (value: unknown) => {
...
+ *
+ * For security reasons, only literal string values and number may be used in
+ * embedded expressions. To incorporate non-literal values [[`unsafeCSS`]] may
+ * be used inside an expression.
  */
 export const css = (
   strings: TemplateStringsArray,
   ...values: (CSSResultGroup | number)[]
-): CSSResultGroup => {
+): CSSResult => {
```
bicknellr, line 129: 👍

packages/reactive-element/src/css-tag.ts
justinfagnani: oh... I thought it was new. I'll add it back.

bicknellr (approved):

sorvell (approved):
LGTM

PR #1917, opened 2021-05-20
[reactive-element] Don't use instanceof to check for CSSResults

### LINKED ISSUE #1893: [reactive-element] CSSResults should be checked by property, not instanceof
This will enable `css` from one copy of reactive-element to work with other versions/copies.

See https://github.com/lit/lit/blob/main/packages/reactive-element/src/css-tag.ts#L70

justinfagnani:
Fixes #1893


kevinpschaaf (approved):
LGTM

sorvell (approved):
LGTM

PR #1918, opened 2021-05-20
[reactive-element] Don't cache CSSResults, cache style sheets

justinfagnani:
This might have a small perf and memory overhead improvement by not hitting a cache for small values that will just be interpolated as strings into larger CSSResults. Instead we cache just the heavy-to-compute object, the CSSStyleSheet itself.


kevinpschaaf (approved):
LGTM

sorvell (approved):
LGTM

Issue #1886, opened 2021-05-12
Transform decorators to readable JS


aomarks:
E.g. transform from:

```ts
@customElement('my-element');
class MyElement extends LitElement {
  @property()
  foo = 'bar';
}
```

to:

```js
class MyElement extends LitElement {
  static get properties() {
    return {
      foo: {}
    }
  }

  constructor() {
    super();
    this.foo = 'bar';
  }
}
customElements.define('my-element', MyElement);
```

Why?

- For lit.dev, we'd like to have a global TS/JS switch (https://github.com/lit/lit.dev/issues/332), and one of the obstacles for that is converting and maintaining every code example in both TS and JS. We can't just run tsc on the examples, because compiled decorator code is not human readable. With this transform, though, the majority of examples could be automatically converted and maintained going forward.

- We're also planning a transform that compiles to a more optimized form, e.g. where reactive property getter/setters are defined directly instead of by the lit runtime. Previously we had planned to write this transform to directly transform decorators, but if we have the intermediate transform described here, this second transform could instead consume readable javascript, which would then allow the optimized transform to work on code that was directly written with javascript too.

PR #1919, opened 2021-05-21
[transformers ] Add @lit/ts-transformers package with idiomatic decorator transform

aomarks:
Adds a `@lit/ts-transformers` package to contain TypeScript transforms.

This PR starts with the "idiomatic" decorator tansform, which transforms all Lit decorators into JavaScript idiomatic style.

The initial motivating use case for this transform is to automatically generate https://lit.dev examples in non-decorator JavaScript, controlled by a JS/TS switch (https://github.com/lit/lit.dev/issues/332).

There may also be some size/performance improvements by using this transform, but we'll want to benchmark before saying that -- plus there's more we can do for this goal, such as defining getters/setters for reactive properties too.

Fixes https://github.com/lit/lit/issues/1886

### Before

```ts
import {LitElement, html} from 'lit';
import {
  customElement,
  property,
  state,
  query,
  eventOptions,
} from 'lit/decorators.js';

@customElement('simple-greeting')
class SimpleGreeting extends LitElement {
  @property()
  name = 'Somebody';

  @state()
  private counter = 0;

  @query('#myButton')
  button?: HTMLButtonElement;

  render() {
    return html`
      <button id="myButton" @click=${this._onClick}>
        Hello ${this.name} (${this.counter})
      </button>
    `;
  }

  @eventOptions({capture: true})
  _onClick(event: Event) {
    this.counter++;
    console.log('click', event.target);
  }
}
```

### After

```ts
import {LitElement, html} from 'lit';

  static get properties() {
    return {
      str: {},
      num: {state: true},
    };
  }

  get button() {
    return this.renderRoot?.querySelector('#myButton');
  }

  constructor() {
    super(...arguments);
    this.name = 'Somebody';
    this.counter = 0;
    this._onClick = {
      handleEvent: (event) => {
        console.log('click', event.target);
      },
      capture: true,
    };
  }

  render() {
    return html`
      <button id="myButton" @click=${this._onClick}>
        Hello ${this.name} (${this.counter})
      </button>
    `;
  }
}

customElements.define('simple-greeting', SimpleGreeting);
```


packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (ts.isClassDeclaration(node)) {
+      return this._visitClassDeclaration(node);
+    }
+    return ts.visitEachChild(node, this.visit, this._context);
+  };
+
+  private _visitImportDeclaration(node: ts.ImportDeclaration) {
+    if (
+      ts.isStringLiteral(node.moduleSpecifier) &&
+      node.moduleSpecifier.text === 'lit/decorators.js'
```
rictic, line 104: Consider also checking for `'lit-element'` since that's probably still common, and `'`lit/decorators'` since that's how auto-imports will tend to get written

packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,263 @@
...
+          let arr = this._memberDecoratorVisitors.get(visitor.decoratorName);
+          if (arr === undefined) {
+            arr = [];
+            this._memberDecoratorVisitors.set(visitor.decoratorName, arr);
+          }
+          arr.push(visitor);
+          break;
+        }
+
+        default:
```
rictic, line 78: Recommend always wrapping cases in brackets, even when there's just one statement inside

packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,263 @@
...
+    // Class decorators
+    for (const decorator of class_.decorators ?? []) {
+      if (
+        !ts.isCallExpression(decorator.expression) ||
+        !ts.isIdentifier(decorator.expression.expression)
+      ) {
+        continue;
+      }
+      const decoratorName = decorator.expression.expression.getText();
+      const visitors = this._classDecoratorVisitors.get(decoratorName) ?? [];
```
rictic, line 123: Here and elsewhere, I think we should look to see where this was imported from, and only apply our transform if we're confident it was imported from lit.

I think all you need to do for this is get the symbol for decorator.expression.expression, then get its first declaration. That should be a node inside the import, for decorators we want to transform.

packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    // Class decorators
+    for (const decorator of class_.decorators ?? []) {
+      if (
+        !ts.isCallExpression(decorator.expression) ||
+        !ts.isIdentifier(decorator.expression.expression)
+      ) {
+        continue;
+      }
+      const decoratorName = decorator.expression.expression.getText();
```
rictic, line 122: Avoid calling getText(), as it's present on every node but isn't generally what you want (and it can throw for artificial nodes). We actually added a linter check to prevent calling getText in grimlock

We know that decorator.expression.expression is an Identifier so we can just get its text with the `text` property

packages/transformers/src/tests/idiomatic-test.ts
```diff
@@ -0,0 +1,392 @@
...
+    }
+  }
+  `;
+
+  const expected = `
+  import {LitElement} from 'lit';
+  class MyElement extends LitElement {
+    constructor() {
+      super(...arguments);
+      this._onClick = {
```
rictic, line 380: Hm, I don't think this is a safe change. Before _onClick was a function, now it's an object.

aomarks: Hmm true, this works well for a Lit `@event` binding, but it wouldn't work if the function was invoked in any other way, which we can't too easily know.

Another option is:

```ts
  _onClick(event) {
    console.log('click', event.target);
  }
}
Object.assign(MyElement.prototype._onClick, {once: true});
```

cc @justinfagnani since we discussed this a bit in the meeting

packages/transformers/src/lit-transformer.ts
aomarks: Replaced it with `if` because there are only two cases and it's not really more readable as a `switch` imo.

packages/transformers/src/lit-transformer.ts
aomarks: Done, removed all usage.

packages/transformers/src/tests/idiomatic-test.ts
aomarks: Hmm true, this works well for a Lit `@event` binding, but it wouldn't work if the function was invoked in any other way, which we can't too easily know.

Another option is:

```ts
  _onClick(event) {
    console.log('click', event.target);
  }
}
Object.assign(MyElement.prototype._onClick, {once: true});
```

cc @justinfagnani since we discussed this a bit in the meeting

rictic: Why not just:

```ts
MyElement.prototype._onClick.once = true;
```

packages/transformers/src/tests/compile-ts-fragment.ts
```diff
@@ -0,0 +1,153 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as ts from 'typescript';
+
+/**
+ * Filesystem caching for the TypeScript CompilerHost implementation used in
+ * `compileTsFragment`. Shared across multiple invocations to significantly
+ * speed up the loading of common files (e.g. TypeScript core lib d.ts files).
+ * Assumes files on disk never change.
```
rictic, line 13: Consider calling this out a bit more, I almost missed it. Maybe put it on its own line and do:

```
**Important Note**: Assumes files on disk never change
```

Maybe name it something that indicates that its caching makes assumptions, like `OptimisticallyImmutableCompilerHostCache` or `OverlyOptimisticCompilerHostCache`

packages/transformers/src/tests/idiomatic-test.ts
aomarks: Yep that works too -- though I could see an argument that it could be nice for our idiomatic style to have a slightly closer match between the way you set event handler options in e.g. `addEventListener(..., {once: true, capture: true})` using an object expression.

packages/transformers/src/tests/compile-ts-fragment.ts
aomarks: FYI just realized when I re-organized the commits I forgot to delete the other version of this file. So, this is not actually a new file, it just moved from `localize-tools` (though your comment is still valid). Added the delete to the newest commit since I don't want to invalidate the commits with comments anymore.

packages/transformers/src/lit-transformer.ts
aomarks: Done. Actually I think all of these are valid (!) so I'm mostly using prefix tests now:

```
'lit/decorators.js',
'lit/decorators',
'lit/decorators/custom-element.js',
'lit/decorators/custom-element',
'@lit/reactive-element/decorators.js',
'@lit/reactive-element/decorators',
'@lit/reactive-element/decorators/custom-element.js',
'@lit/reactive-element/decorators/custom-element',
'lit-element',
'lit-element/index.js',
'lit-element/index',
'lit-element/decorators.js',
'lit-element/decorators',
```

packages/transformers/src/lit-transformer.ts
aomarks: Done. Also made it understand `import {property as potato}` aliases, and only remove bindings that we actually are transforming (when importing from `lit-element` we especially need this, since `LitElement` etc. are imported from the same module).

kevinpschaaf:
Suggestion for more idiomatic TS -> JS without using decorators.
```
<div @click=${{passive: true, handleEvent: e => this.handleClick(e)}>
```

packages/transformers/src/tests/compile-ts-fragment.ts
aomarks: Added an "Important!" comment with explanation. Also just in case this wasn't clear, note this is in the test directory and used to speed up loading standard TS libs for unit tests.

packages/transformers/src/tests/idiomatic-test.ts
aomarks: Switched to e.g. `@click=${{handleEvent: (e) => this._onClick(e), capture: true}}` as long as the handler is `private`, see comment from Kevin (and offline discussion). If not `private`, back to modifying the prototype method directly after the class declaration.

aomarks:
> Suggestion for more idiomatic TS -> JS without using decorators.
> 
> ```
> <div @click=${{passive: true, handleEvent: e => this.handleClick(e)}>
> ```

Done, as long as the method is `private`. Otherwise, we'll annotate the prototype method directly (in case it's e.g. used by a subclass).

aomarks:
@rictic @justinfagnani @kevinpschaaf PTAL

I've addressed all comments, and made some additional improvements. Main items:

- The `@eventOptions` decorator now transforms corresponding references inside Lit template event binding to the canonical syntax we agreed on offline (`@click=${{handleEvent: (e) => this._onClick(e), capture: true}}`), as long as the method is `private`. If it's not `private`, we annotate the prototype method directly (in case e.g. it is used in an event binding by a subclass).

- We now actually track imports from Lit modules, and check for decorators/html function with a type checker symbol lookup, instead of making naming assumptions.

- We now skip full traversal of files if we don't see any relevant Lit imports right away.

packages/transformers/src/lit-transformer.ts
justinfagnani: The non-file-extension imports are not valid actually. They could be used by toolchains that don't (yet) support package exports, but I don't think we should support that. Those projects will break and have to fix imports when they update tools.

aomarks: > The non-file-extension imports are not valid actually. They could be used by toolchains that don't (yet) support package exports, but I don't think we should support that. Those projects will break and have to fix imports when they update tools.

If we don't support them, the behavior will be that we silently don't transform decorators if they are imported without an extension. Is that your preference? We could also have an explicit check for that style of import, but here seems like the wrong place to do that.

packages/transformers/src/tests/idiomatic-test.ts
```diff
@@ -0,0 +1,657 @@
...
+
+  const expected = `
+  import {LitElement} from 'lit';
+  class MyElement extends LitElement {
+    constructor() {
+      super(...arguments);
+      this.str = "foo";
+      this.num = 42;
+    }
+    static get properties() {
```
justinfagnani, line 104: we should put the properties block above the constructor.

justinfagnani, line 104: Ideally there's some space around the class members. Is that possible?

justinfagnani, line 104: Can we have an option to email `static properties = {...}`?

packages/transformers/src/tests/idiomatic-test.ts
```diff
@@ -0,0 +1,657 @@
...
+  `;
+
+  const expected = `
+  import {LitElement} from 'lit';
+  class MyElement extends LitElement {
+    constructor() {
+      super();
+      this.num = 42;
+    }
+    static get properties() {
```
justinfagnani, line 142: We shouldn't move the properties block and/or constructor if they already exist.

packages/transformers/src/tests/idiomatic-test.ts
```diff
@@ -0,0 +1,657 @@
...
+  import {LitElement} from 'lit';
+  class MyElement extends LitElement {
+    constructor() {
+      super(...arguments);
+      this.num = 42;
+      this.num2 = 24;
+    }
+    static get properties() {
+      return {
+        num: {state: true, attribute: false},
```
justinfagnani, line 176: `state: true` implies `attribute: false` already.

packages/transformers/src/tests/idiomatic-test.ts
```diff
@@ -0,0 +1,657 @@
...
+  `;
+
+  const expected = `
+  import {LitElement} from 'lit';
+  class MyElement extends LitElement {
+    get span() {
+      if (this.__span === undefined) {
+        this.__span = this.renderRoot?.querySelector('#mySpan');
+      }
+      return this.__span;
```
justinfagnani, line 223:
```suggestion
      return this.__span ??= this.renderRoot?.querySelector('#mySpan');
```

aomarks (commented):
All comments addressed @justinfagnani PTAL

packages/transformers/src/lit-transformer.ts
justinfagnani: I feel like we should error here. At first we'll be the ones using this and our code should be correct. If we find that we see valid requests to support extension-less imports, and it somehow would be better to fail, and it guaranteed to fail, somewhere else in the pipeline, then we could relax the strictness here. I don't want to perpetuate tools that do the wrong thing and will allow code that breaks when tools are upgraded.

packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,398 @@
...
+  Visitor,
+  ClassDecoratorVisitor,
+  MemberDecoratorVisitor,
+  GenericVisitor,
+} from './visitor.js';
+
+const unreachable = (x: never) => x;
+
+/**
+ * Configurable transformer for Lit.
```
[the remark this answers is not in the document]
justinfagnani: by "coalesce visitors" I mean that it's possible sometimes to write a batching visitor that calls visit on each of a set of visitors so they all go in one walk, rather than a series of walks. I think your code does this already.

packages/transformers/src/visitor.ts
```diff
@@ -0,0 +1,42 @@
...
+import type {LitClassContext} from './lit-class-context.js';
+import type {LitFileContext} from './lit-file-context.js';
+
+/** Union of all Lit transformer visitor types. */
+export type Visitor =
+  | ClassDecoratorVisitor
+  | MemberDecoratorVisitor
+  | GenericVisitor;
+
+/** Visit a Lit decorator that is applied to a class. */
```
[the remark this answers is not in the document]
justinfagnani: I meant as opposed to a decorated field, and for some reason when I read it before it seemed ambiguous to me and now doesn't. nvm!

packages/transformers/src/lit-transformer.ts
```diff
@@ -0,0 +1,398 @@
...
+   * Check the given import declaration to see if it imports any symbols from
+   * Lit modules that apply to the configured visitors. Return whether or not a
+   * relevant import was found.
+   */
+  private _analyzeImportDeclaration(node: ts.ImportDeclaration): boolean {
+    if (!ts.isStringLiteral(node.moduleSpecifier)) {
+      return false;
+    }
+    const specifier = node.moduleSpecifier.text;
+    if (
```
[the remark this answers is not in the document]
justinfagnani: yeah. +1

packages/transformers/src/tests/idiomatic-test.ts
aomarks: Discussed offline. Decided to leave as the getter for now, since that's what we document for JS at https://lit.dev/. Can revisit later.

packages/transformers/src/tests/idiomatic-test.ts
justinfagnani: Oh, that'd be good too.

packages/transformers/src/lit-transformer.ts
aomarks: Done. We now error if you don't import with  `.js` extension or the default module.

justinfagnani (approved):

packages/transformers/src/tests/idiomatic-test.ts
aomarks: Sent separately in https://github.com/lit/lit/pull/2030

PR #1922, opened 2021-05-21
Don't remove attributes with noChange on first render

### LINKED ISSUE #1920: Attribute content is cleared entirely if an expression in an `AttributePart` evaluates to `noChange` on first render
# Description

lit-html: `2.0.0-rc.3`
lit-element: `3.0.0-rc.2`

If one of the expressions to be rendered by an AttributePart is `noChange` on the first render, the entire content of the attribute is cleared even if there are other parts of the attribute statically written into the template (or controlled by other directives, I think).

I think this is because of these steps:

1. The `._$committedValue` property is initialized here and filled with `nothing`:

    https://github.com/lit/lit/blob/ec5ab1b394d8de27e1680294547d7cc1db166692/packages/lit-html/src/lit-html.ts#L1231

1. When the value produced is `noChange`, it is substituted with a value from `._$committedValue`:

    https://github.com/lit/lit/blob/ec5ab1b394d8de27e1680294547d7cc1db166692/packages/lit-html/src/lit-html.ts#L1292-L1295

    However, the value from  `._$committedValue` was never previously set because this is the first render, so it reads the `nothing` it was initialized to.

1. A couple lines down, if the value for that particular expression (`v`) is `nothing`, then the full value to render to the attribute (`value`, previously `"color: red; "` in the example below) is replaced with `nothing`[^1]:

    https://github.com/lit/lit/blob/ec5ab1b394d8de27e1680294547d7cc1db166692/packages/lit-html/src/lit-html.ts#L1298-L1299

# Steps to Reproduce

1. Write this code

```ts
import {html, render, noChange} from 'lit';

render(html`
  <div style="color: red; ${noChange}">
    This text should be red.
  </div>
`, document.body);
```

2. See this output...

The text "This text should be red." is rendered, but it isn't red.

## Live Reproduction Link

https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIHJlbmRlciwgbm9DaGFuZ2V9IGZyb20gJ2xpdCc7XG5cbnJlbmRlcihodG1sYFxuICA8ZGl2IHN0eWxlPVwiY29sb3I6IHJlZDsgJHtub0NoYW5nZX1cIj5cbiAgICBUaGlzIHRleHQgc2hvdWxkIGJlIHJlZC5cbiAgPC9kaXY-XG5gLCBkb2N1bWVudC5ib2R5KTsifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHNpbXBsZS1ncmVldGluZyBuYW1lPVwiV29ybGRcIj48L3NpbXBsZS1ncmVldGluZz5cbjwvYm9keT5cbiJ9XQ

## Expected Results

The rendered text is red.

## Actual Results

The rendered text is not red.

---

[^1]: I think this might qualify as a second bug: any directive of those associated with the same AttributePart can cause the entire attribute to be cleared if it alone returns `nothing` - even after the first render. Filed as https://github.com/lit/lit/issues/1921.

---

(Updated to remove discussion about directives, which aren't actually related.)

justinfagnani:
Fixes #1920

In attribute expressions `noChange` is now treated like `undefined` and `null` on first render, instead of like `nothing`. Previously the attribute was removed since we initialized attribute part values to `nothing`.

### DIFF
--- a/packages/lit-html/src/test/lit-html_test.ts
+++ b/packages/lit-html/src/test/lit-html_test.ts
@@ -965,7 +965,7 @@ suite('lit-html', () => {
       assertContent('<div>baz</div><div foo="bar"></div>');
     });
 
-    test('renders undefined in attributes', () => {
+    test('renders undefined in interpolated attributes', () => {
       render(html`<div attribute="it's ${undefined}"></div>`, container);
       assert.equal(
         stripExpressionComments(container.innerHTML),
@@ -980,6 +980,11 @@ suite('lit-html', () => {
 
     test('renders null in attributes', () => {
       render(html`<div attribute="${null as any}"></div>`, container);
+      assertContent('<div attribute=""></div>');
+    });
+
+    test('renders noChange in attributes', () => {
+      render(html`<div attribute="${noChange as any}"></div>`, container);
       assertContent('<div attribute=""></div>');
     });
 
@@ -1027,6 +1032,14 @@ suite('lit-html', () => {
     test('noChange works on one of multiple expressions', () => {
       const go = (a: any, b: any) =>
         render(html`<div foo="${a}:${b}"></div>`, container);
+
+      go('A', noChange);
+      assert.equal(
+        stripExpressionComments(container.innerHTML),
+        '<div foo="A:"></div>',
+        'A'
+      );
+
       go('A', 'B');
       assert.equal(
         stripExpressionComments(container.innerHTML),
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1022,6 +1022,29 @@ suite('lit-html', () => {
       assertContent('<div attribute=""></div>');
     });
 
+    test('renders empty string in attributes', () => {
+      render(html`<div attribute="${''}"></div>`, container);
+      assertContent('<div attribute=""></div>');
+    });
+
+    // In the PR for illustrative purposes. Will remove.
+    test.skip('renders noChange in attributes', () => {
```
sorvell, line 1031: Let's enable these tests and validate that the attribute is *not* rendered.

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -1066,6 +1089,14 @@ suite('lit-html', () => {
     test('noChange works on one of multiple expressions', () => {
       const go = (a: any, b: any) =>
         render(html`<div foo="${a}:${b}"></div>`, container);
+
+      go('A', noChange);
```
sorvell, line 1093: Let's clarity the name of the test, perhaps `noChange renders as empty string when used in interpolated attributes`.

sorvell (approved):
LGTM, just a couple of minor suggestions to improve the tests.

kevinpschaaf (approved):
LGTM

PR #1942, opened 2021-06-04
Add prefix to stable properties to avoid collisions with short user properties

### LINKED ISSUE #1925: Uncaught TypeError: this.L.has is not a function
# Description

If I define L property gets this.L.has is not a function

# Steps to Reproduce

1. Write this code

```ts
import {html, LitElement} from 'lit';
	@property({ type: Array }) L: number[] = [];
}
window.customElements.define('x-el', MyElement);
```

2. See this output...

## Live Reproduction Link

https://stackblitz.com/edit/lit-html-abwsxa

## Expected Results

use L as property, in lit-element@2 I can use it. If is a reserve property of lit, you should a **_ or #**

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

kevinpschaaf:
Add prefix to stable properties to avoid collisions with short user properties.

Also centralizes per-package class prefixes to validate prefixes aren't duplicated since the rollup configs tend to be copy/pasted and it'd be easy to forget to choose a new unique prefix for a new package.

Fixes #1925

### DIFF
--- a/packages/benchmarks/package-lock.json
+++ b/packages/benchmarks/package-lock.json
@@ -15,7 +15,7 @@
         "tachometer": "^0.5.9"
       },
       "devDependencies": {
-        "chromedriver": "^89.0.0",
+        "chromedriver": "^91.0.0",
         "typescript": "^4.2.3"
       }
     },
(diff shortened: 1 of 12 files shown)


sorvell (approved):

AndrewJakubowicz (approved):
Looks good to me. I also checked out locally and took a look at the minified build.

justinfagnani (approved):

PR #1948, opened 2021-06-08
[lit-html] Add flag to clear render container

justinfagnani:
This matches a change I'm adding to 1.x internally to not clear the container except when this flag is present.


kevinpschaaf (approved):
LGTM

PR #1954, opened 2021-06-11
fix(ifDefined): treat type as lit-analyzer in Lit 1 would

e111077:
Lit analyzer would see ifDefined function name and simply calculate the type as if it were the value without `undefined`. This fixes an issue where passing in a `DirectiveResult|undefined` was being calculated instead as `DirectiveResult | typeof nothing`. This meant lit-analyzer could not understand that this was a directive and meant to be ignored by the type checker.

externalization of cl/378967428

### DIFF
--- a/packages/lit-html/src/directives/if-defined.ts
+++ b/packages/lit-html/src/directives/if-defined.ts
@@ -12,4 +12,4 @@ import {nothing} from '../lit-html.js';
  *
  * For other part types, this directive is a no-op.
  */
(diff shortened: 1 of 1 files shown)


justinfagnani:
I think we need to fix lit-analyzer, since this is lying about the return type.

e111077:
Seems like lit-analyzer already checks each individual union type. This is an edge case anyway and the users' code can be rewritten. Will instead rewrite their code since I don't think it's prudent to spend as much time as I have already on legacy lit-analyzer code especially since it's just closure type checks which are internal-only anways.

PR #1955, opened 2021-06-12
[labs] Adds initial implementation of @lit-labs/context

benjamind:
This is the very first draft of an initial implementation of lit reactive controllers for the Context API. There's still a lot to do for this to be complete, but hopefully can be used as a good basis for moving forward with the propose API.

This version is implemented against the 'react-like' version of the context API outlined [here](https://github.com/webcomponents/community-protocols/pull/10#discussion_r622667567).

I set this up by cloning one of the other packages, so hopefully its mostly in line with coding style. Would love to hear feedback on this, and I'm sure we'll continue discussion on the Context API over in the community protocols group.


sorvell:
This is great. It may be a bit until we can review this thoroughly but the core proposal seems really good.

benjamind:
> This is great. It may be a bit until we can review this thoroughly but the core proposal seems really good.

No problem. I am away until next Tuesday now anyway. I will likely get the work done to align the proposal with this implementation then too.

packages/labs/context/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+export class MyApp extends LitElement {
+  private logger: Logger = {
+    log: (msg) => {
+      console.log(`[my-app] ${msg}`);
+    },
+  };
+  protected render(): TemplateResult {
+    return html`
+      <context-provider .context=${loggerContext} .value=${this.logger}>
```
justinfagnani, line 85: I'd like to offer a way to provide context to a subtree w/o an element. A directive seems like a good way to do this, and can wrap the entire template too:

```ts
  protected render(): TemplateResult {
    return provide({context: loggerContext, value: this.logger}, html`
      <my-thing></my-thing>
    `;
  }
```

Although I'd probably use the controller if I'm providing to the whole tree. I also think a `@provide` decorator would be nice:

```ts
export class MyApp extends LitElement {
  @provide(loggerContext)
  // @state() - @provide() is non-reactive by default? Use @state() to make it reactive...
  private logger: Logger = {
    log: (msg) => {
      console.log(`[my-app] ${msg}`);
    },
  };
```

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+      return;
+    }
+    ev.stopPropagation();
+    this.addCallback(ev.callback, ev.multiple);
+  };
+  hostConnected(): void {
+    this.host.addEventListener('context-request', this.onContextRequest);
+  }
+  hostDisconnected(): void {
+    this.clearCallbacks();
```
cuberoot, line 37: Do we need a call to `removeEventListener` here to match the call to `addEventListener` above?

packages/labs/context/README.md
benjamind: Yeah I like both of those additions. I also then a generic function for just providing context at any `HTMLElement` outside of any lit-specific template or class context would be useful (we have a case where we could do with this right now):

```ts
const view = document.querySelector('#some-view');
provideContext(view, loggerContext, logger);
```

packages/labs/context/README.md
```diff
@@ -0,0 +1,101 @@
...
+### Providing a Context
+
+Finally we want to be able to provide this context from somewhere higher in the DOM:
+
+```ts
+import {LitElement} from 'lit';
+import '@lit-labs/context/lib/elements/context-provider-element.js';
+import {loggerContext, Logger} from './my-logger.js';
+
+const loggerContext = createContext('logger', new Logger());
```
e111077, line 75: You are both importing and defining `loggerContext` in this example.

packages/labs/context/README.md
```diff
@@ -0,0 +1,101 @@
...
+import {LitElement} from 'lit';
+import {Logger, loggerContext} from './logger.js';
+
+export class MyElement extends LitElement {
+  @property({attribute: false})
+  public logger?: Logger;
+
+  public constructor() {
+    // add the consumer controller
+    this.addController(
```
justinfagnani, line 38: ContextConsumer already adds itself as a controller, so the `addController()` call isn't necessary.

Also, the patterns are new with controllers, but I think the pattern we're gravitating to is that controller contains its state and requests a host update when that state changes, then other code can just access it:

```ts
export class MyElement extends LitElement {
  private _loggerContext = new ContextConsumer(this, loggerContext);
  
  render() {
    this._loggerContext.value?.log('rendering...');
    return html`...`;
  }
}
```

In cases where code needs to know that _specifically_ this context was updated, the controller can take an optional callback.

packages/labs/context/README.md
justinfagnani: This is a different file from the one defining and exporting `loggerContext`. @benjamind maybe add file names before the code blocks to make that more clear?

packages/labs/context/README.md
justinfagnani: We'd need to figure out when this would be called and what state it needs to preserve. If that function adds a listener each time, we don't want to call it in render/update and add multiple listeners... We can keep weakmaps/sets for that. This might be good as a method on ContextProvider?

packages/labs/context/rollup.config.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {litProdConfig} from '../../../rollup-common.js';
+
+export default litProdConfig({
+  classPropertyPrefix: 'Ϋ',
```
justinfagnani, line 10: We moved away from greek prefixes. See the common Rollup config.

packages/labs/context/src/lib/context-event.ts
```diff
@@ -0,0 +1,40 @@
...
+     * A 'context-request' event can be emitted by any element which desires
+     * a context value to be injected by an external provider.
+     */
+    'context-request': ContextEvent<UnknownContext>;
+  }
+}
+
+/**
+ * A callback which is provided by a context requester and is called with the value satisfying the request.
+ * This callback can be called multiple times by context providers as the requested value is changed.
```
justinfagnani, line 15: Add `@param` jsdoc, especially for `dispose`

packages/labs/context/src/lib/create-context.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export type Context<T> = {
+  name: string;
+  initialValue?: T;
+};
+
+/**
+ * An unknown context type
+ */
+export type UnknownContext = Context<unknown>;
```
justinfagnani, line 12: Do we need this? `Context<unknown>` I understand when reading it, `UnknownContext` I have to look at the definition to know what it is, and it's only two less characters.

packages/labs/context/src/lib/decorators/context-provided.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
justinfagnani, line 3: 2017 -> 2021 here and elsewhere

packages/labs/context/src/lib/elements/context-provider-element.ts
```diff
@@ -0,0 +1,41 @@
...
+import {ContextProvider} from '../controllers/context-provider.js';
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'context-provider': ContextProviderElement<UnknownContext>;
+  }
+}
+
+@customElement('context-provider')
+export class ContextProviderElement<
```
justinfagnani, line 14: I think I'd rather skip this whole element in favor of an element directive that can make any element provide a value to a subtree. The `provide()` ChildPart directive I sketched above could also function as an element directive:

```
html`<div ${provide(loggerContext, logger)}><x-child></x-child><x-child></x-child></div>`
```

justinfagnani (changes requested):

packages/labs/context/README.md
benjamind: That pattern is an interesting one. I certainly agree with it where a context *has* to be provided, but if the context is designed as an optional pattern such that it can be used to fulfill a property value then I guess its not quite as straightforward.

For example in the `@contextProvided` decorator case, we're using the ContextConsumer controller to populate the property, a user of the component can just directly bind a property value or provide it via context. This seems very flexible to me.

packages/labs/context/README.md
benjamind: Trying to implement the directive, did your example above have an omission in it? I can't figure out how to make a directive that returns a `TemplateResult` directly? I'd have expected to have to wrap it in a `html` tag like so:

```ts
  protected render(): TemplateResult {
    return html`${provide({context: loggerContext, value: this.logger}, html`
      <my-thing></my-thing>
    `})`;
  }
  ```
  
  This tracks with the implementation I'm seeing in `until`.

Comment (benjamind) [reply packages/labs/context/README.md:85]: In the case of the ChildPart directive outlined here, would the directive attach the `context-request` event listener to the `parentNode` of the ChildPart or render a wrapper node around the content produced by the passed in template?

The `AttributePart` case is I think pretty straightforward, attach the listener to the `part.element` and remove it if the `part.element` changed?

File: packages/labs/context/README.md (hunk 4 above)
Comment (justinfagnani) [reply packages/labs/context/README.md:38]: I think in that case an element could use the callback, though the author should be very careful about how two different sources for a value interact. Most likely they want the local property to win, so either the callback should use logical nullish assignment, or a getter with a fallback to the ContextConsumer value:

```ts
  private _loggerContext = new ContextConsumer(this, loggerContext);
  
  private _logger?: Logger;
  get logger() {
    return this._logger ?? this._loggerContext.value;
  }
  set logger(v: Logger) {
    this._logger = v;
    this.requestUpdate();
  }
}
```

I think it'd be good if `@contextProvided` allowed this. Right now it looks like it always overwrites the local value.

packages/labs/context/README.md
justinfagnani: A DirectiveResult is a renderable value, so you don't need to wrap it. You can return a template from `render()` or `update()`.

For the ChildPart case there is a little complication about where to attach the listener, and to make sure that you only respond to events from the elements wrapped, so maybe that's not a good place to start.

I wouldn't do an AttributePart, I'd do ElementPart instead. I don't see a need for an attribute name.

benjamind: Ah I see. So we could skip support for Child part and just support ElementPart or direct return usage?

packages/labs/context/src/lib/elements/context-provider-element.ts
benjamind: Agreed, removed.

packages/labs/context/src/lib/decorators/context-provided.ts
benjamind: Done

packages/labs/context/src/lib/create-context.ts
benjamind: Agreed, reverted to `Context<unknown>` everywhere.

packages/labs/context/src/lib/context-event.ts
benjamind: Done, have not documented all functions yet, will do a doc pass after implemntation details settle.

packages/labs/context/rollup.config.js
benjamind: I think I've updated this correctly, let me know if not.

packages/labs/context/README.md
benjamind: I'm a little unsure how to go about this. Shoudl the `@contextProvided` directive add the property with getter/setter to the element if it doesn't already have a property? Is there any bad interaction that might have for `@property` decorator? Or should we just to the logical nullish assignment on receipt of a new context value instead?

packages/labs/context/README.md
benjamind: Done, I'm not sure what style to use here so I picked something that seemed to look half decent.

packages/labs/context/src/lib/controllers/context-provider.ts
benjamind: Fixed.

packages/labs/context/README.md
benjamind: I added a basic implementation of the `ElementPart` for `provide` directive. I am unsure how to extend this to use it as a direct returned templated result. Would it imply that the provide directive would render a passed in 3rd argument as the template? How do you get to the element from the template result inside the `update` function?

guillemcordoba:
Hey :D No intention to intrude, but any news about @lit-labs/context being merged? We are waiting for it with enthusiasm!

benjamind:
> Hey :D No intention to intrude, but any news about @lit-labs/context being merged? We are waiting for it with enthusiasm!

Good call out, I have let this stagnate as I've been caught up at work with many other things. I'll try and take another look at this asap. @justinfagnani I assume pushing ahead with this as labs release is still ok with the team?

packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -0,0 +1,74 @@
...
+    private callback: (value: ContextType<Ctx>, dispose?: () => void) => void,
+    private multiple: boolean = false
+  ) {
+    this.host.addController(this);
+  }
+
+  private dispose?: () => void;
+
+  hostConnected(): void {
+    this.host.dispatchEvent(
```
guillemcordoba, line 38: @benjamind 

In my port of this library [here](https://github.com/holochain-open-dev/context/tree/main/ui), I've had to add a `setTimeout` here because of an edge case:

- I have a parent element that is providing context to slotted elements:
```js
export class SynSession extends ScopedElementsMixin(LitElement) {
...
    render() {
    return html`
    <slot
      ${provide(
        synSessionContext,
        this.synStore?.sessionStore(this.sessionHash)
      )}
    ></slot>`;
  }

```

and then the context consumer is like: 

```js
export class SynFolks extends ScopedElementsMixin(LitElement) {
  @contextProvided({ context: synSessionContext, multiple: true })
  session!: SessionStore<any, any>;

  render() {
    if (!this.session) return html`There is no active session`;

    return html` Hi!`;
  }
}
```

and the html is like:

```html
  <syn-context>
    <syn-folks />
  </syn-context>
```

What happens is that the context consumer element gets connected before the provider does, and thus the event gets fired but not listened to. For now, the `setTimeout` workaround works, but I know it's not the most elegant solution.

blikblum: I handled this use case in this library: https://github.com/blikblum/wc-context

It tracks when a context request is not handled and re send the event after a microtask

benjamind: Elegant solutions, yeah we should work something like this in I guess. Thanks for the feedback!

justinfagnani: I can see this problem happening with the `provide()` directive, since it's not set up until rendering, which is at least a microtask after connected. but it shouldn't happen with ContextProvider since the listener is set up on connected. This could be a reason to recommend the controller over the directive. Perhaps the controller could take a selector to scope it to cover the subtree cases of the directive.

blikblum: FYI this occurs when provider is in a shadow Dom host and the consumer is slotted even when "providing" in connected callback.

The slotted element connected callback is fired before the  host connected callback.

renoirb: I also could fork this and play around. It took time to properly setup lit@2.0.0-rc  but it's working.

(I'm aware this is a lit implementation, but...)

say I have a team who will use the components, and I want them to be free to use plain JavaScript, or their framework of choice.

I wonder how we could make a provider builerplate that isn't bound to lit.
 I'll be spending time trying to make a Provider HTMLElement and probably see to have methods for updating fed in as argument.

packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -0,0 +1,74 @@
...
+      new ContextEvent(
+        this.context,
+        (value, dispose) => {
+          // some providers will pass a dispose function indicating they may provide future values
+          if (this.dispose) {
+            // if the dispose function changes this implies we have changed provider
+            if (this.dispose !== dispose) {
```
renoirb, line 45: That's actually a cool feature. Didn't realized that could be possible.

packages/labs/context/src/lib/controllers/context-container.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A simple class which stores a value, and triggers registered callbacks when the
+ * value is changed via its setter.
+ *
+ * An implementor might use other observable patterns such as MobX or Redux to get
+ * behavior like this. But this is a pretty minimal approach that will likely work
+ * for a number of use cases.
+ */
+export class ContextContainer<T> {
```
renoirb, line 27: This file could be used outside of Lit.

Has this been discussed in the CG of the context community-protocol breakout notes?

packages/labs/context/src/lib/directives/provide.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  Directive,
+  directive,
+  DirectiveParameters,
+  DirectiveResult,
+  ElementPart,
+  PartInfo,
+  PartType,
+} from 'lit/directive.js';
+import {noChange} from 'lit';
```
renoirb, line 10: When forking this experiment, and in now lit 2.0, `noChange` is not found. It is only accessible via `lit-html`.

packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -0,0 +1,74 @@
...
+            this.provided = true;
+            this.callback(value, dispose);
+          }
+
+          this.dispose = dispose;
+        },
+        this.multiple
+      )
+    );
+  }
```
manolakis, line 67: should we call the callback with the initialValue in case there is no value provided by a provider?

```js
        },
        this.multiple
      )
    );

    if (!this.provided && this.context.initialValue) {
      this.callback(this.context.initialValue);
    }
}
```

packages/labs/context/src/test/context-provided_test.ts
```diff
@@ -0,0 +1,72 @@
...
+  let provider: SimpleContextProvider;
+  let consumer: SimpleContextConsumer;
+
+  setup(async () => {
+    const container = document.createElement('div');
+    container.innerHTML = `
+       <simple-context-provider>
+         <simple-context-consumer></simple-context-consumer>
+       </simple-context-provider>
+     `;
```
renoirb, line 47: What if we want many `<simple-context-consumer></simple-context-consumer>` with some way of discriminating each of them?

```html
<simple-context-provider>
    <simple-context-consumer data-discriminant="foo"></simple-context-consumer>
    <simple-context-consumer data-discriminant="bar"></simple-context-consumer>
    <simple-context-consumer data-discriminant="baz"></simple-context-consumer>
</simple-context-provider>
```

That pattern would be useful in a micro-front-end app that knows a list of ids, but can't know data to fill each `<simple-context-consumer />`.

That's what I've been trying to do that prompted me to talk about it in https://github.com/webcomponents-cg/community-protocols/issues/21#issuecomment-934603937

renoirb: Here is an example benjamind/lit#1 for this use-case

benjamind:
@renoirb thanks for the notes, please be aware I am currently tied up in product launch issues and won't get to update lit-context until sometime in the next couple of weeks. At which point I'm going to make a complete laser focus on getting this finalized.

This PR has also fallen well behind main now so I am likely to close it and reopen after a rebase to the head to bring in lit2 support and cleanup a number of issues and changes which have come out of CG discussion.

benjamind:
I've made some major changes to the implementation, but I think where its at now is probably stable and would love if folks could find some time to re-review.

Most major change is a shift towards controllers and decorators:

[ContextConsumer](https://github.com/lit/lit/pull/1955/files#diff-17e264b35a8123e52c7a49d949caa3d9f1dd3b650b5bebbb510b420d0895f3ef) is a controller which emits `context-request` and executes a given callback when the request is satisfied, as per community protocol for these events.

[ContextProvider](https://github.com/lit/lit/pull/1955/files#diff-53fa6c8f234974eb9d4a8421925e380d5b2a34741a7e134c9328f722513d8021) is a controller which stores a value using a ContextContainer and handles the provider side of the protocol. The provider controller also emits a `context-provider` event to advertise the providers availability when connected to the dom this is used by ContextRoot described below.

[@contextRequest](https://github.com/lit/lit/pull/1955/files#diff-3b8229018c7bd57147eddc7bc480bd50c8700b5f3fc08b6a0ba7da2fe2b813a7) is a property decorator for requesting a context and binding its value to a property.

[@contextProvider](https://github.com/lit/lit/pull/1955/files#diff-bd5ec2cd4c201e79ace979f9822a562d218828f7a3cd1dca9f3e3a376a0a5ea0) the associated provider decorator, which shims the property setter it decorates and updates the ContextProvider controller on property change.

I've also added a [ContextRoot](https://github.com/lit/lit/pull/1955/files#diff-5853f7009a835f0d5ad58df308c815b46b035bbc0b199ce9eaddb7c610f8dae3) implementation which can track `context-request` and `context-provider` events to resolve unsatisfied requests when providers advertise their availabiilty. This provides a simple mechanism for resolving late-upgraded context providers without leaning too much into event pathing which on investigation has proven to be somewhat unreliable once you trace out all behaviors of slotted elements and portalling components used in various libraries.

Would love any feedback on this, and advise from the lit team on how to get this PR over the hurdle and committed.

LasseRosenow:
Hey @benjamind I really love this PR and am looking forward to it very much :) Thank you so much :)

Just a small thing I noticed:

We have the classes `ContextProvider` and `ContextConsumer`
Also we have the decorators `@contextProvide` and `@contextRequest`

Now I am wondering, why not name `@contextRequest` => `@contextConsumer`?
This would make it more consistent with the class names and also improve the general ease of use, as people would expect to have a consumer when there is a provider. :)

benjamind:
Agreed I think some naming consistency tweaks are in order. I wasn't sure which way to go specifically, Requestor / Provider or Consumer / Provider? I left it as is hoping someone would comment one preference over the other so thank you for bringing it up!

bennypowers:
prefer consumer/provider. it's more current IMO

bennypowers:
by-the-way i'm looking to implement context protocol in patternfly elements. so far i just copy-pasted from the protocol proposal but this is looking PHAT

packages/labs/context/src/lib/context-root.ts
```diff
@@ -0,0 +1,97 @@
...
+import {ContextKey} from './context-key.js';
+import {ContextRequest, ContextRequestEvent} from './context-request-event.js';
+import {ContextProviderEvent} from './controllers/context-provider.js';
+
+type UnknownContextKey = ContextKey<unknown, unknown>;
+
+/**
+ * Makes all object properties in a type into WeakRefs
+ */
+type MakeWeak<T> = {[P in keyof T]: T[P] extends object ? WeakRef<T[P]> : T[P]};
```
renoirb, line 16: Where does `WeakRefs` type comes from? When using an up to date TypeScript version, along with newest lit versions, including `@lit/reactive-element`, I don't *see* this type.

benjamind: I believe its in the `ESNext` and `ES2021` libs pulled in by setting the `lib` in `tsconfig.json`.

renoirb: Right. Probably our huge repo with many packages at work doesn't use ES2021.

packages/labs/context/src/lib/controllers/context-container.ts
renoirb: It's basically a "Container" or "Bottle" with list of listener for that. A bit like what vue-next does for dependency tracking.

packages/labs/context/src/lib/decorators/context-provider.ts
```diff
@@ -0,0 +1,78 @@
...
+ * ```ts
+ * import {loggerContext} from 'community-protocols/logger';
+ *
+ * class MyElement {
+ *   @contextProvided(loggerContext)
+ *   logger;
+ *
+ *   doThing() {
+ *     this.logger.log('thing was done');
+ *   }
```
renoirb, line 40: The specification says that ContextAPI *should not* consider *Dependency Injection*.  This looks like it.

Maybe we should change the examples to something that isn't part of what the specification. Say a component with two properties

```ts

export interface IUserData {
  readonly id: string
}

// Say another slice of state is about status
export interface IUserStatus extends IUserData {
  readonly presence: 'offline' | 'available' | 'away'
}

export interface IUserInfo extends IUserData {
  readonly firstName: string
}

export const UserInfoContext = createContext<IUserInfo>('user-info')

class UserInfoNameFormatter extends LitElement {
  @property()
  id = ''

  @state()
  userInfo

  public constructor() {
    super()
    new ContextConsumer(this, UserInfoContext, (value: IUserInfo) => {
      if (value.id === this.id) {
        this.userInfo = { ...value }
      }
    })
  }

  render() {
    return html`<div data-id=${this.id}>${this.firstName}</div>`
  }
}
 ```

**Comment by LasseRosenow** [comment]
Actually is there any eta, when this will be available it @lit-labs?

File: packages/labs/context/src/lib/controllers/context-consumer.ts (hunk 14 above)
Comment (renoirb) [reply packages/labs/context/src/lib/controllers/context-consumer.ts:99]: Makes sense. Anyway, a consumer or another who would care about a context may have different initial state values

File: packages/labs/context/src/lib/controllers/context-consumer.ts [hunk 18]
```diff
@@ -0,0 +1,100 @@
...
+  Context extends ContextKey<unknown, unknown>,
+  HostElement extends ReactiveElement
+> implements ReactiveController
+{
+  private provided = false;
+
+  constructor(
+    protected host: HostElement,
+    private context: Context,
+    private callback: (
```
renoirb, line 31: [Question] This signature seem to be used at a few places. Maybe it should have a name? Unless this inline type is for the typings rollup

packages/labs/context/src/lib/context-root.ts
```diff
@@ -0,0 +1,97 @@
...
+      // redispatch if we still have all the parts of the request
+      if (element && callback) {
+        element.dispatchEvent(
+          new ContextRequestEvent(ev.context, callback, subscribe)
+        );
+      }
+    });
+  };
+
+  private onContextRequest = (
```
sorvell, line 81: If an element is connected multiple times, it would add itself to this list multiple times, which isn't ideal. This seems like a corner case that can likely be ignored, but we should at least note it either in a comment here or associated issue.

packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ContextRequestEvent} from '../context-request-event.js';
+import {ContextKey, ContextType} from '../context-key.js';
+import {ReactiveController, ReactiveElement} from 'lit';
+
+export type DispatchOptions = 'connected' | 'updated' | 'raf' | 'microtask';
```
sorvell, line 11: For what use cases are these options needed?

packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -0,0 +1,100 @@
...
+            }
+            // if we don't support subscription, immediately unsubscribe
+            if (!this.subscribe) {
+              this.unsubscribe();
+            }
+          }
+
+          // only invoke callback if we are either expecting updates or have not yet
+          // been provided a value
+          if (!this.provided || this.subscribe) {
```
sorvell, line 89: Is this called if `subscribe` is false?

packages/labs/context/src/lib/controllers/context-container.ts
```diff
@@ -0,0 +1,77 @@
...
+  public get value(): T {
+    return this._value;
+  }
+  public set value(v: T) {
+    this.setValue(v);
+  }
+
+  public setValue(v: T, force = false) {
+    let changed = false;
+    if (v !== this._value) {
```
sorvell, line 40: Consider using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is.

packages/labs/context/src/lib/controllers/context-container.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A simple class which stores a value, and triggers registered callbacks when the
+ * value is changed via its setter.
+ *
+ * An implementor might use other observable patterns such as MobX or Redux to get
+ * behavior like this. But this is a pretty minimal approach that will likely work
+ * for a number of use cases.
+ */
+export class ContextContainer<T> {
```
sorvell, line 27: This is pretty generic, maybe call it `ValueNotifier` or something since it's not really related to `Context` per se.

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  implements ReactiveController
+{
+  constructor(
+    protected host: ReactiveElement,
+    private context: T,
+    initialValue?: ContextType<T>
+  ) {
+    super(initialValue);
+    this.host.addController(this);
+    this.attachListeners();
```
sorvell, line 53: Calling this in the constructor and in `hostConnected` doesn't make sense since it's double work. I guess the platform won't 2x add the listener, but the event will 2x dispatch, once potentially to outer space if the element is not connected at construction time.

It seems like the `request` event listener should be added here and the `provide` event should be dispatched in `hostConnected`.

packages/labs/context/src/lib/controllers/context-container.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  addCallback(callback: ContextCallback<T>, subscribe?: boolean): void {
+    if (subscribe) {
+      const record: CallbackRecord<T> = [
+        callback,
+        () => {
+          this.callbacks.delete(record);
+        },
+      ];
+      this.callbacks.add(record);
```
sorvell, line 69: Is there a reason we need to store the unsubscribe function here? The unfortunate result is that despite using a Set here, multiple calls with the same callback are not de-duplicated. This would be the case if the callback itself were stored in the Set.

sorvell (changes requested):
Think this is very close! Just a few suggestions.

packages/labs/context/src/lib/context-request-event.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface ContextRequest<Context extends ContextKey<unknown, unknown>> {
+  readonly context: Context;
+  readonly callback: ContextCallback<ContextType<Context>>;
+  readonly subscribe?: boolean;
+}
+
+/**
+ * An event fired by a context requester to signal it desires a specified context.
+ *
+ * A provider should inspect the `context` property of the event to determine if it has a value that can
```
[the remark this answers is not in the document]
benjamind: Yes it is the 'context key' unfortunately given that this is the event, it has to match the community protocol for the event shape defined here: https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md#the-context-request-event

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: Good question, I'm struggling to recall the specific uses we had for this. I suspect it was before I came up with the `context-root` solution, when it made sense to have different timings for some consumers where the providers were not yet available.

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: Yes, a consumer can request a context without subscribing for subsequent updates (`subscribe` is false). The second condition here caters for the 'multiple updates' case where a consumer 'subscribes' to the context for repeated update.

packages/labs/context/src/lib/controllers/context-provider.ts
benjamind: Hmm, I think you're right in that the `provide` event should only be dispatched in `hostConnected` but the event listener I think needs attachment in the constructor and in `hostConnected` since we remove the listener in disconnected?

packages/labs/context/src/lib/controllers/context-container.ts
benjamind: If we don't store the `Disposer` function here, then we'd need to create it in the `updateContext` loop  above to pass out the disposer function to each observer so it can remove itself if the subscription is complete. Unless I'm missing something?

packages/labs/context/src/lib/controllers/context-consumer.ts
blikblum: IMO the option to have a single context (updates consumer once) as specified in protocol is a mistake. The provided value may be not defined yet or be lazy loaded at the time the context event occurs. And requiring to the consumer know when is lazy loaded or not, creates a dependency to an implementation detail.

Also complicates the things a bit

The consumer, if only wants the first value, should call the unsubscribe at the time receives the first meaningful value

packages/labs/context/src/lib/controllers/context-container.ts
benjamind: Agreed, like that, renamed to `ValueNotifier` in 0248a2c

packages/labs/context/src/lib/controllers/context-container.ts
benjamind: Good catch! Updated in 0248a2c

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: I agree to some extent, defaulting to subscription rather than one-time request would simplify implementation a little. Though it would also have a potential overhead for all those context requests which are one-time only since the provider needs to create the unsubscribe function and it needs to be called. So you pay for that implementation simplicity with runtime overhead?

packages/labs/context/src/lib/context-root.ts
benjamind: Added a comment, shall I file tickets ahead of merging this PR?

packages/labs/context/src/lib/context-request-event.ts
```diff
@@ -0,0 +1,64 @@
...
+ * satisfy the request, calling the `callback` with the requested value if so.
+ *
+ * If the requested context event contains a truthy `subscribe` value, then a provider can call the callback
```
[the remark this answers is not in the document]
benjamind: Agreed it is an explicit optimization, added some more detail here.

packages/labs/context/src/context.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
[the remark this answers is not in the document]
benjamind: Fair, moved this up to the index.ts file in 0248a2c

packages/labs/context/README.md
```diff
@@ -0,0 +1,151 @@
+# @lit-labs/context
+
+## Overview
+
```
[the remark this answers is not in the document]
benjamind: Let me know if what I added her was sufficient, didn't want to get into too much detail.

packages/labs/context/README.md
benjamind: Was there something here still remaining @justinfagnani ?

packages/labs/context/src/lib/directives/provide.ts
benjamind: Directives have been removed as they have sequencing issues that are more easily solved via use of decorators and elements.

packages/labs/context/src/test/context-provided_test.ts
benjamind: I think we've agreed that this sort of interaction can be built atop the basic protocol based on what is being passed here. So resolving this conversation for now.

packages/labs/context/package.json
```diff
@@ -0,0 +1,67 @@
...
+      "default": "./index.js"
+    }
+  },
+  "files": [
+    "/src/",
+    "!/src/test/",
+    "/development/",
+    "!/development/test/",
+    "/context.{d.ts,d.ts.map,js,js.map}"
+  ],
```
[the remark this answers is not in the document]
benjamind: This just seems to be the pattern used in lit projects in general, @sorvell can you let me know if anything here should change.

packages/labs/context/src/lit-context.ts
```diff
@@ -0,0 +1,7 @@
+export {ContextCallback, ContextEvent} from './lib/context-event.js';
+
+export {createContext} from './lib/create-context.js';
```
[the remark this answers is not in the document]
benjamind: Resolving as outdated.

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: If you mean the callback signature, maybe, its now only used in 2 places one of which is readonly, so am leaving this as is for now.

packages/labs/context/README.md
```diff
@@ -0,0 +1,151 @@
...
+    this.provider.setValue(newLogger);
+  }
+}
+```
+
+## Known Issues
+
+### Late upgraded Context Providers
+
+In some cases you might have a context providing element that is upgraded late. LightDOM content below this provider may end up requesting a contenxt that is currently not provided by any provider.
```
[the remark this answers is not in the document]
benjamind: Fix applied in [0248a2c]

benjamind:
Think this is ready for another pass @sorvell, can someone also approve the CI workflows so we can get a test run going?

packages/labs/context/src/lib/controllers/context-consumer.ts
blikblum: Is the runtime overhead measurable?

Anyway, more important than implementation simplification is the decoupling from provider. 

Say you have a provider for `model` state. Initially is implemented as a sigleton / available from start.
Consumers are implemented with `multiple: false` 

Later the provider changes its implementation to lazy load `model`, sending the initial value as undefined. You need to change consumers implementation.

It's bad when provider and consumer are developed in same project. Now think when provider/consumer are developed by third party.

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: I have added the concept of a `ContextRoot` to this implementation that does something similar to these. It listens at its attached node for unhandled context request events, and also introduces the concept that providers should advertise themselves via a `context-provider` event. Once it sees a provider for an unhandled context request it re-dispatches the event at the source. Seems to work nicely without being too complex or needing to walk the DOM tree.

benjamind:
> Now I am wondering, why not name `@contextRequest` => `@contextConsumer`?
> This would make it more consistent with the class names and also improve the general ease of use, as people would expect to have a consumer when there is a provider. :)

@sorvell  what do you think of this suggestion?

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: Performance testing was going to be something the lit team would follow up with after getting this initial implementation merged. Might be worth exploring then.

I get your point here though. There is certainly some simplicity in only having the single model of subscription behavior. I'm open to changing it but we would need to get agreement from the community protocol group on that change. That might be easier to do so after we actually have this implementation published somewhere for people to try out.

You might also argue that a provider that switched to a lazy loaded value from an immediately provided value is breaking its initial contract, and that should really be done with a new context key?

renoirb:
Hi Benjamin!, I've noticed when you open the project in VSCode only in `packages/labs/context/` and see many import typing errors. Tests runs, but development and type checking doesn't work.

Here is what I changed to solve.

```patch
From 272f1a5699462dea53e192c620ab4f8450bd4008 Mon Sep 17 00:00:00 2001
From: Renoir Boulanger <contribs@renoirboulanger.com>
Date: Wed, 23 Mar 2022 19:52:35 -0400
Subject: [PATCH] local development typing errors

---
 packages/labs/context/tsconfig.json | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/packages/labs/context/tsconfig.json b/packages/labs/context/tsconfig.json
index 19e97352..74105763 100644
--- a/packages/labs/context/tsconfig.json
+++ b/packages/labs/context/tsconfig.json
@@ -22 +22,6 @@
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "paths": {
+      "lit/decorators": ["../../lit/decorators"],
+      "lit": ["../../lit"],
+      "@lit/*": ["../../"],
+    }
-- 
2.30.1 (Apple Git-130)
```

benjamind:
> Hi Benjamin!, I've noticed when you open the project in VSCode only in `packages/labs/context/` and see many import typing errors. Tests runs, but development and type checking doesn't work.
> 
> Here is what I changed to solve....

@sorvell can you comment on if this change is correct, I'm not sure the repo supports working in a single package like this in VSCode?

renoirb:
> @sorvell can you comment on if this change is correct, I'm not sure the repo supports working in a single package like this in VSCode?

That's what I'm doing at the moment, it works well.

1. VSCode opened in `packages/labs/context/`
2. In terminal below, run `./scripts/dev.sh`

And notice the lack of distracting red underlined text.

Screen Shot 2022-03-23 at 8 09 07 PM

packages/labs/context/README.md
```diff
@@ -0,0 +1,169 @@
...
+}
+```
+
+## Known Issues
+
+### Late upgraded Context Providers
+
+In some cases you might have a context providing element that is upgraded late. LightDOM content below this provider may end up requesting a context that is currently not provided by any provider.
+
+To solve this case we provide a `ContextRoot` class which can intercept and track unsatisfied `context-request` events and then redispatch these requests when providers are updated.
```
bennypowers, line 151: There's another problem here

```html
<early-provider>
  <late-provider>
    <eager-consumer>
  </late-provider>
</early-provider>
```

Oops! `<eager-consumer>` is stuck with `<early-provider>`, even though by rights it should get its own from `<late-provider>`

Does `ContextRoot` come equipped to deal with this?

My solution:

```ts
// consumer
hostConnected() {
  const event = new ContextEvent(this.context, this.contextCallback, true);
  this.host.dispatchEvent(event);
  contextEvents.set(this.host, event);
}

hostDisconnected() {
  this.dispose?.();
  this.dispose = undefined;
  contextEvents.delete(this.host);
}
```
```ts
//provider
/**
 * When a context provider connects, it listens for context-request events
 * it also fires all previously fired context-request events from their hosts,
 * in case this context provider upgraded after and is closer to a given consumer.
 */
hostConnected() {
  this.host.addEventListener('context-request', this.onChildContextEvent);
  for (const [host, fired] of contextEvents) {
    host.dispatchEvent(fired);
  }
}
```

LasseRosenow:
@benjamind 

> Agreed I think some naming consistency tweaks are in order. I wasn't sure which way to go specifically, Requestor / Provider or Consumer / Provider?

Just wanted to kindly ask if you made a decision as to where to go, as it seems like those inconsistencies are still present, but also I feel like most people prefer the provider - consumer approach judging by the comments

benjamind:
> @benjamind
> 
> > Agreed I think some naming consistency tweaks are in order. I wasn't sure which way to go specifically, Requestor / Provider or Consumer / Provider?
> 
> Just wanted to kindly ask if you made a decision as to where to go, as it seems like those inconsistencies are still present, but also I feel like most people prefer the provider - consumer approach judging by the comments

The controllers are called `ContextConsumer` and `ContextProvider`, we still have a `context-request` event which aligns with the community protocol. So I think this is all resolved now?

LasseRosenow:
> @LasseRosenow The controllers are called `ContextConsumer` and `ContextProvider`, we still have a `context-request` event which aligns with the community protocol. So I think this is all resolved now?

We still have the inconsistency with the decorator being called `contextRequest` while the class is called 'contextConsumer'

packages/labs/context/README.md
```diff
@@ -0,0 +1,169 @@
...
+      },
+      true // pass true to get updates if the logger changes
+    );
+  }
+
+  private doThing() {
+    this.logger?.log('a thing was done');
+  }
+}
+```
```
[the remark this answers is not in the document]
justinfagnani: +1

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -0,0 +1,77 @@
...
+  updateObservers = (): void => {
+    this.callbacks.forEach(([callback, disposer]) =>
+      callback(this._value, disposer)
+    );
+  };
+
+  addCallback(callback: ContextCallback<T>, subscribe?: boolean): void {
+    if (subscribe) {
+      const record: CallbackRecord<T> = [
+        callback,
```
justinfagnani, line 64: To avoid duplicating a record for the same callback, make `this.callbacks` a Map keyed by the callback with the disposer as the value. cc @sorvell

packages/labs/context/src/lib/controllers/context-container.ts
justinfagnani: We can use a map instead: https://github.com/lit/lit/pull/1955/files#r855421440

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  hostConnected(): void {
+    this.attachListeners();
+    // emit an event to signal a provider is available for this context
+    this.host.dispatchEvent(new ContextProviderEvent(this.context));
+  }
+  hostDisconnected(): void {
+    this.host.removeEventListener('context-request', this.onContextRequest);
+    this.clearCallbacks();
```
justinfagnani, line 77: We don't need to remove callbacks, right? All consumers in the subtree below will call their unsubscribe callbacks if they have them.

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+    ev.stopPropagation();
+    this.addCallback(ev.callback, ev.subscribe);
+  };
+
+  private attachListeners() {
+    this.host.addEventListener('context-request', this.onContextRequest);
+  }
+
+  hostConnected(): void {
+    this.attachListeners();
```
justinfagnani, line 71: We can remove this and only attach listeners to the host in the constructor. The cycle between the host and this controller can be GC'ed.

benjamind:
> > @LasseRosenow The controllers are called `ContextConsumer` and `ContextProvider`, we still have a `context-request` event which aligns with the community protocol. So I think this is all resolved now?
> 
> We still have the inconsistency with the decorator being called `contextRequest` while the class is called `contextConsumer`
> 
> Of course the event should keep it's name ;)

@LasseRosenow yes but I think `@contextConsumer` doesn't really describe what it does? Compare these two snippets:

```ts
@contextRequest({ context: loggerContext })
@property({attribute: false}
public logger?: Logger;
```
with
```ts
@contextConsumer({ context: loggerContext })
@property({attribute: false}
public logger?: Logger;
```

I might be willing to rename to `contextProvided` perhaps? This would align with some Adobe internal approaches to this.

```ts
@contextProvided({ context: loggerContext })
@property({attribute: false}
public logger?: Logger;
```

packages/labs/context/src/lib/context-root.ts
sorvell: Sure

packages/labs/context/src/lib/controllers/context-consumer.ts
sorvell: How about we file an issue to consider removing the options, given they may not be needed.

packages/labs/context/src/lib/controllers/context-provider.ts
sorvell: +1

packages/labs/context/src/lib/value-notifier.ts
sorvell: +1

justinfagnani:
Regarding the decorator name: I could see something like `@contextValue()` as well.

packages/labs/context/README.md
benjamind: After offline discussion marking this as resolved after all the changes we've made since this thread.

packages/labs/context/package.json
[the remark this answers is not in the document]
benjamind: Resolving this conversation for now as I think whats there currently conforms to the rest of the lit project, so will leave it to the lit team to revise this in future if needs be.

packages/labs/context/src/lib/decorators/context-provider.ts
benjamind: In the interests of merging I'm going to defer this, filed https://github.com/lit/lit/issues/2764 to track

packages/labs/context/src/lib/context-root.ts
benjamind: Filed https://github.com/lit/lit/issues/2765 to track

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: I have removed it in 79da874 after discussion with Justin. We can re-add in future if a user needs it.

packages/labs/context/src/lib/controllers/context-consumer.ts
benjamind: Filed https://github.com/lit/lit/issues/2762 to track adding performance tests.

packages/labs/context/src/lib/controllers/context-provider.ts
benjamind: Have removed in 79da874 as it was unnecessary, tests seemed to confirm this too!

packages/labs/context/README.md
benjamind: Have filed https://github.com/lit/lit/issues/2763 to track possibly investigating this. Not sure its actually solveable in a way that doesn't have performance impacts though and seems like a very very edge case to me?

packages/labs/context/README.md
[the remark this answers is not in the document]
benjamind: Updated in 79da874 added a `.value` property and also added a `requestUpdate` call in the ContextConsumer to support usage of `.value` in templates.

benjamind:
> Regarding the decorator name: I could see something like `@contextValue()` as well.

I renamed to `@contextProvided` in 79da874. I didn't like the overlap with the value of the context that `@contextValue` implies and `@contextProvided` does to me seem to state more what the intention is of the decorated property? We can revise in future if needs be.

packages/labs/context/src/lib/value-notifier.ts
benjamind: Done in 79da874, good call!

packages/labs/context/src/lib/controllers/context-provider.ts
benjamind: Agreed, removed in 79da874

packages/labs/context/src/lib/controllers/context-provider.ts
benjamind: Removed in 79da874

benjamind:
Phew! All issues addressed, I think this is ready to merge after CI passes?

sorvell:
Can we merge https://github.com/lit/lit/commit/79da87440b53a67aac74514962b9cad1e6243462 with this?

benjamind:
> Can we merge [79da874](https://github.com/lit/lit/commit/79da87440b53a67aac74514962b9cad1e6243462) with this?

Confused what do you mean?

justinfagnani:
@benjamind https://github.com/lit/lit/commit/79da87440b53a67aac74514962b9cad1e6243462 didn't make it into this PR somehow. I see that it's not a branch in the repo even, but it is a branch in your fork.

benjamind:
Well that's just weird. Will fix this morning.

benjamind:
I'm struggling to see the issue, the commit is showing for me in the commits associated with this PR: https://github.com/lit/lit/pull/1955/commits

Second from the bottom?

justinfagnani:
@benjamind huh... you're right seems fine. Maybe GitHub is just confusing us with the warning it puts at the top of a commit link from a PR from a work. It's warning that the commit is not part of this repository.

justinfagnani (approved):
LGTM. Thanks @benjamind !

sorvell (approved):
Huh, latest changes seem to show ok now.

Thanks for the additional context! LGTM

benjamind:
Success! Thanks everyone for the help and input on this one, it was a long time coming together but looking forward to this getting usage in the wild!

PR #1957, opened 2021-06-14
[lit-html] Update optional method syntax so it's compatible with tsickle.

AndrewJakubowicz:
## Why

Done as part of the lit next burndown. For internal details see cl/379343684.

Lit-html is transpiled by tsickle, and this optional type syntax is unsupported. The optional method is marked as unknown leading to internal test failures. Filing a bug (b/191086442) in tsickle in parallel to possibly handle this syntax in the future.

This unblocks lit-next being used internally.


rictic (approved):

AndrewJakubowicz:
The syntax is now supported! This change is no longer necessary.

PR #1959, opened 2021-06-15
[all] Change from greek prefixes to ASCII

### LINKED ISSUE #1958: Greek symbols result in "SyntaxError: Unexpected identifier 'i'"
# Description

The `Σ`, `Π`, etc. variable names in the codebase fails in mysterious ways when `<meta charset="utf8">` is not explicitly set. Browsers seem to default to this if not defined, but not the case in Safari Webviews.

For example, `self.Σi = 1` is interpreted as `self.Î i = 1`, causing `"SyntaxError: Unexpected identifier 'i'"`. Setting an explicit charset in the HTML page fixes the issue. A potential fix, other than removing the `Σ` symbols and similar, may be to not suffix (e.g. `Σi`, `Σo`), as the resulting space between the characters is what causes the issue. In other words, `self.Σ = 1` does not generate a syntax error.

Screen Shot 2021-06-14 at 5 18 32 PM

It was unexpected to debug this, as only a few ways of importing scripts generate sufficient line information in Safari tooling (and seems to inconsistently reflect charset state in tools, rendering as either `Σ` or `Î`) as well as loading up unminified builds, and not expecting internal variable names to affect the runtime.

# Steps to Reproduce

* Run this from a WKWebview: `<script>self.Σi = 1</script>`

Screen Shot 2021-06-14 at 5 44 45 PM

## Expected Results

No error.

## Actual Results

`"SyntaxError: Unexpected identifier 'i'"`.

## Browsers Affected

- [x] Safari WKWebview with no explicit charset: `"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)"`.

kevinpschaaf:
Greek prefixes require an explicit charset on some browser/webview environments, which was causing friction.

Fixes #1958


rollup-common.js
```diff
@@ -17,17 +17,17 @@ import virtual from '@rollup/plugin-virtual';
 // defined here rather than via an argument to litProdConfig() so we can
 // validate the list used by each package is unique (since copy/pasting the
 // individual package-based configs is common and error-prone)
-const STABLE_PROPERTY_PREFIX = 'Ξ';
+const STABLE_PROPERTY_PREFIX = '$A';
```
justinfagnani, line 20: I do wonder if this is unique enough...

sorvell (approved):

justinfagnani (approved):

rollup-common.js
sorvell: Maybe `$lit$` for this?

kevinpschaaf:
Size increase for lit-html:
```
Before: lit-html.js   7.26 KB  7.2 KB    3.08 KB   2.79 KB  
After:  lit-html.js   7.33 KB  7.33 KB   3.07 KB   2.8 KB    
```

rollup-common.js
kevinpschaaf: Added an add'l `_` to all prefixes.

justinfagnani (changes requested):

.changeset/fifty-trainers-check.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
kevinpschaaf: Whoops!

kevinpschaaf:
Removed pre.json and updated changelog message for this PR. @justinfagnani PTAL

.changeset/fifty-trainers-check.md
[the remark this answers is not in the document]
kevinpschaaf: Updated

PR #1960, opened 2021-06-15
[lit-html] Add spread directive

justinfagnani:
This needs more tests and I'm not entirely happy with using the private object this way, but the general approach works and is small.

Tests needed:
- [ ] - AsyncDirective lifecycle in all binding types
- [ ] - Multiple spread directives on an element
- [ ] - Removing the spread directive
- [ ] - Collisions between spread and normal bindings

### DIFF
--- a/packages/lit-html/src/directives/spread.ts
+++ b/packages/lit-html/src/directives/spread.ts
@@ -1,0 +1,104 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {noChange, nothing, _Σ as litHtmlPrivate} from '../lit-html.js';
+import {
+  directive,
+  Directive,
+  DirectiveParameters,
+  PartInfo,
+  PartType,
+  Part,
+  ElementPart,
+} from '../directive.js';
+
+const {
+  _AttributePart: AttributePartImpl,
+  _PropertyPart: PropertyPartImpl,
+  _BooleanAttributePart: BooleanAttributePartImpl,
+  _EventPart: EventPartImpl,
+  // _ElementPart: ElementPartImpl,
+} = litHtmlPrivate;
+
+export interface SpreadValues {
+  readonly [name: string]: unknown;
+}
+
+class SpreadDirective extends Directive {
+  _previousValues: Set<string> = new Set();
+  _parts = new Map<string, Part>();
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    if (partInfo.type !== PartType.ELEMENT) {
+      throw new Error(
+        'The `spread` directive must be used al element expression'
+      );
+    }
+  }
+
+  render(_values: SpreadValues) {
+    // TODO for SSR
+    throw new Error('Not implemented');
+  }
+
+  update(elementPart: ElementPart, [values]: DirectiveParameters<this>) {
+    // Remove old properties that no longer exist in styleInfo
+    // We use forEach() instead of for-of so that re don't require down-level
+    // iteration.
+    const {element} = elementPart;
+    this._parts!.forEach((part, name) => {
+      // If the name isn't in values or it's null/undefined
+      if (values[name] == null) {
+        // Clear the part, which will remove attributes and disconnect directives
+        part._$setValue(nothing);
+        // Remove the part
+        this._parts!.delete(name);
+      }
+    });
+
+    // Add or update properties
+    for (const name in values) {
+      const value = values[name];
+      if (value != null) {
+        let part = this._parts.get(name);
+        if (part === undefined) {
+          const m = /([.?@])?(.*)/.exec(name)!;
+          const ctor =
+            m[1] === '.'
+              ? PropertyPartImpl
+              : m[1] === '?'
+              ? BooleanAttributePartImpl
+              : m[1] === '@'
+              ? EventPartImpl
+              : AttributePartImpl;
+          this._parts.set(
+            name,
+            (part = new ctor(
+              element,
+              m[2],
+              ['', ''],
+              elementPart,
+              elementPart.options
+            ))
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/directives/spread.ts
```diff
@@ -0,0 +1,104 @@
...
+  _BooleanAttributePart: BooleanAttributePartImpl,
+  _EventPart: EventPartImpl,
+  // _ElementPart: ElementPartImpl,
+} = litHtmlPrivate;
+
+export interface SpreadValues {
+  readonly [name: string]: unknown;
+}
+
+class SpreadDirective extends Directive {
```
kevinpschaaf, line 30: We should consider whether this should be an `AsyncDirective` so that its `disconnected` callback can undo all the parts if you render a different value to the expression (or whether spread must be static), e.g....

```html
<div ${bool ? spread({foo, bar}) : nothing}></div>
``` 

To be fair, we don't do a good thing for e.g. `classMap` in this situation either (it would also need to be disconnectable), which seemed a bit overkill given the performance hit:
```html
<div class=${bool ? classMap({foo, bar}) : nothing}></div>
```

packages/lit-html/src/directives/spread.ts
```diff
@@ -0,0 +1,104 @@
...
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    if (partInfo.type !== PartType.ELEMENT) {
+      throw new Error(
+        'The `spread` directive must be used al element expression'
+      );
+    }
+  }
+
+  render(_values: SpreadValues) {
```
kevinpschaaf, line 43: I can imagine use cases where it'd be nice to let `values` be nullable, to avoid users from needing to always make a throwaway object in the alternate case:
```html
<div ${spread(maybeGotStuff ?? {})}></div>
```
or
```html
<div ${spread(shouldRenderStuff ? stuff : {})}></div>
```

Again, we don't do this in `classMap`/`styleMap`, but I think it's more likely to have patterns where entire bags are forwarded in (e.g. "`formControlAttributes`") vs. being generated by the user (otherwise the user could have just written the attribute bindings themselves).

But it does sort of beg the question what we expect `spread` to actually be used for in the OSS world.

packages/lit-html/src/directives/spread.ts
```diff
@@ -0,0 +1,104 @@
...
+    // TODO for SSR
```
kevinpschaaf, line 44: Do you have any nascent ideas here? Seems like we could go two ways here:
* Return an array of descriptors of the part types + values, and figure out how to reuse the `attribute-part` opcode handling to handle the per-part-specific stuff: https://github.com/lit/lit/blob/c8fe1d4c4a8b1c9acdd5331129ae3641c51d9904/packages/labs/ssr/src/lib/render-lit-html.ts#L612
* Return the array of _acutal_ part instances (iow, use the update code) that the `attribute-part` opcode handling can directly use, and figure out how to deal with the fact that you don't have an actual part to get the element to pass in `render` (which isn't technically a problem in SSR because _it_ doesn't need the element).


We do need to reuse code in the `attribute-part` opcode handling to ensure directives passed to spread work correctly.

packages/lit-html/src/directives/spread.ts
justinfagnani: I agree this is the most complete implementation, but given that it's more costly it might be better to keep things simple. I think this pattern _should_ suffice:

```
<div ${spread(bool ? {foo, bar} : undefined)}></div>
```

packages/lit-html/src/directives/spread.ts
justinfagnani: Agreed, as per the other thread.

I imagine this form of spread to be a lower level things that's often wrapped for better DX. An element might receive a bag of property or event bindings and need to add the correct prefix. `attr` will wrap this. Custom attribute-merging strategies will... do something. We need to make sure that's possible.

packages/lit-html/src/directives/spread.ts
justinfagnani: I'm not sure yet. I think something like just forwarding the object in an AttributeResult type might work, even if it's a bit weird to explain how that works or doesn't client-side.

I think figuring out hydration is the hard part here...

LarsDenBakker:
How do you feel about having a directive which spreads only properties, and one that spread binding like the current implementation? A very common use case is spreading an object of properties, often coming from some other source. Right now users would need to always transform the object to have all keys start with a `.`. Or they won't notice and always set attributes, which isn't great.

justinfagnani:
I think that's a good idea, but hopefully can be well expressed as a wrapper on this:

```ts
spreadProps = (props) => spread(Object.fromEntries(Object.entries().map(([k, v]) => ['.' + k, v])));
```

If the `Object.fromEntries(Object.entries()..)` bit is too much work we can look at customizing the name->part mapping with a callback.

LarsDenBakker:
@justinfagnani that wrapper would do quite a lot of work each render though, so would be nice to make that more convenient.

abdonrd:
Is there news on this? Thanks in advance!

The Carbon team is interested in use it:
https://github.com/carbon-design-system/carbon-web-components/issues/657#issuecomment-902666314

It's one of the main things they are waiting to update to Lit v2.

cc @jeffchew

daKmoR:
any chance this is going to happen within January 2022? 

we are thinking about waiting or hacking our own version for now 😅

samanime:
It'd be great to have this feature soon. I had upgraded without realizing it was missing and now I'll need to downgrade back.

If figuring out how to be deterministic when there are multiple objects with duplicate keys, why not simply limit it to a maximum of one spread operation per object and let them merge objects before. Then you don't have to try and come up with a deterministic approach for everyone, as they can solve it beforehand themselves.

This PR was started back 7 months ago and we still don't have a solution. =S

PR #1966, opened 2021-06-18
[all] Set up changesets

justinfagnani:
- Installs `@changesets/cli` so that it can be run with `npx changeset`
- Commits the result of `npx changeset init`
- Changed config:
  - "access": "public",  "baseBranch": "main",  "updateInternalDependencies": "minor",
- Adds a release action

I also set up on GitHub:
- Created a new environment named Changesets
- Generated and installed an NPM_TOKEN secret


.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
...
+          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
+          fetch-depth: 0
+
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+
+      - name: Install Dependencies
+        run: yarn
```
aomarks, line 26: Do we have to use yarn?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
...
+          # This expects you to have a script called release which does a build for your packages and calls changeset publish
+          publish: yarn release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Send a Slack notification if a publish happens
+        if: steps.changesets.outputs.published == 'true'
+        # You can do something when a publish happens.
+        run: my-slack-bot send-notification --message "A new version of ${GITHUB_REPOSITORY} was published!"
```
aomarks, line 41: Is this slack integration set up?

.github/workflows/release.yaml
justinfagnani: oops. Fixed I think. I'm not sure how we test this without running a release...

.github/workflows/release.yaml
justinfagnani: removed

aomarks (approved):

justinfagnani:
@kevinpschaaf I figured out how to get dependencies updated. I think this is ready to merge now

kevinpschaaf (approved):

PR #1967, opened 2021-06-18
`lit-html-1.x`: Run tests on GitHub actions

bicknellr:
This PR moves the tests and benchmarks for the `lit-html-1.x` branch from Travis to GitHub Actions.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+          node-version: 10
+
+      - name: Restore node_modules
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          # Bump the number after `lit-1.x-node-modules` to force a cache
+          # update. Note there are cache actions in below in `tests-sauce` and
+          # in `benchmarks.yml` that should all keep keys in sync.
```
justinfagnani, line 39: remove reference to benchmarks.yml

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+          npm run format && git diff --exit-code || (echo -e '\n\033[31mERROR:\033[0m
+          Project is not formatted. Please run "npm run format".' && false)
+
+  tests-local:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
```
justinfagnani, line 28: Try v2?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
```
justinfagnani, line 14: Let's try a supported Node LTS. 14 would be best if it works

bicknellr: Ok, bumped to 14

.github/workflows/tests.yml
bicknellr: bumped to v2

.github/workflows/tests.yml
bicknellr: removed

bicknellr:
I'm trying out updating the benchmarks in #1968. If the current run passes, I'll merge it here.

bicknellr:
I think this is ready for review now. Tests and benchmarks are now both running on GitHub Actions. The one thing left to do would be remove whatever's adding the "continuous-integration/travis-ci" check, but I don't have permission to do that as far as I can tell.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Build
+        run: npm run build
+
+      - name: Benchmark
+        run: npx tach --config tachometer.json --json-file results.json
+
+      - name: Report
+        uses: andrewiggins/tachometer-reporter-action@v2
+        with:
+          report-id: lit-html-1.x
```
bicknellr, line 45: I don't know what this field does. Should I remove it?

justinfagnani: @aomarks ?

justinfagnani (approved):

.github/workflows/benchmarks.yml
bicknellr: I'm going to try out a quick branch without this to see what happens.

bicknellr: Oh, just stumbled on this commit: https://github.com/andrewiggins/tachometer-reporter-action/commit/0f505a75d1451d2c19645731de1bfb6ce9e7d0dc

It's been removed from the docs, so I'm assuming it's not supported anymore. Also, nothing in my test branch (https://github.com/lit/lit/pull/1975) seems to have changed other than that the initial comment stating that the benchmarks were in progress didn't appear and the section title in the results comment changed from "lit-html-1.x" to "this-change, lit-html-1.x, previous-1.x-release".

justinfagnani (approved):

bicknellr:
I'm going to merge manually because the travis check is still 'pending'.

PR #1972, opened 2021-06-22
Makes properties that should remain unminified compatible with tools …

### LINKED ISSUE #1970: Stable names used for branding are renamed in Closure compiler
It looks like we need to keep these properties from being renamed by Closure:

```
 '_$litType$',
  '_$litDirective$',
  '_$litPart$',
  '_$litElement$',
  '_$cssResult$',
  ' _$setDirectiveConnected'
```

sorvell:
…other than rollup/terser

Fixes #1970.

Note, does not appear to impact build size at all.


kevinpschaaf (approved):

kevinpschaaf:
Actually, can you revert the changelog.md changes and instead do `npm run add changeset`?

PR #1979, opened 2021-06-24
Fix `getUpdateComplete` example snippet and clean up formatting for others.

### LINKED ISSUE #1614: [reactive-element] Type change but not docs change for `getUpdateComplete`
https://github.com/Polymer/lit-html/blob/d649560cf7a20333dd1af6aa800fb36b40cc9c8f/packages/reactive-element/src/reactive-element.ts#L1088-L1106

This was previously `Promise<void>` as per the documented use in the method comment, but is now `Promise<boolean>`. It's not immediately clear to someone extended this functionality as per the documented use case what value should be returned in such a context.

bicknellr:
The `getUpdateComplete` example is wrong because you're now required to return `Promise<boolean>`. However, my understanding of how it's actually meant to be implemented is pretty poor. Should it consider anything else before returning the result of `super.getUpdateComplete()`? Or, is a proper implementation really meant to return that value without change at all times? I've assumed the second here.

Also Fixes #1614


packages/reactive-element/src/reactive-element.ts
```diff
@@ -1168,11 +1168,15 @@ export abstract class ReactiveElement
    * ```js
    * class MyElement extends LitElement {
    *   async getUpdateComplete() {
```
AndrewJakubowicz, line 1170: Nit: override keyword?

AndrewJakubowicz (approved):
Amazing!

packages/reactive-element/src/reactive-element.ts
bicknellr: Added.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1164,7 +1164,7 @@ export abstract class ReactiveElement
    * before fulfilling this Promise. To do this, first await
    * `super.getUpdateComplete()`, then any subsequent state.
    *
-   * @return A promise of a boolean that indicates if the update resolved
+   * @return A promise of a boolean that resolves to true if the update resolved
```
justinfagnani, line 1151:
```suggestion
   * @return A promise of a boolean that resolves to true if the update completed
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1191,6 +1191,8 @@ export abstract class ReactiveElement
    * }
    * ```
    *
+   * @return A promise of a boolean that resolves to true if the update resolved
+   *     without triggering another update.
```
justinfagnani, line 1179:
```suggestion
   * @return A promise of a boolean that resolves to true if the update completed
   *     without triggering another update.
```

justinfagnani (approved):
LGTM, but I'd remove one of the "resolved" instances per method.

PR #1980, opened 2021-06-24
Ensure that `queryAssignedNodes` always returns an array

dfreedm:
If the `<slot>` element isn't rendered, the decorator can return null


justinfagnani:
LGTM, but can you wait a bit to merge so I can get #1966 in and we'll add changesets to this?

dfreedm:
SGTM. What's involved with adding a changeset?

dfreedm:
Changeset added

packages/reactive-element/src/decorators/query-assigned-nodes.ts
```diff
@@ -60,7 +60,7 @@ export function queryAssignedNodes(
           slotName ? `[name=${slotName}]` : ':not([name])'
         }`;
         const slot = this.renderRoot?.querySelector(slotSelector);
-        let nodes = (slot as HTMLSlotElement)?.assignedNodes({flatten});
+        let nodes = (slot as HTMLSlotElement)?.assignedNodes({flatten}) ?? [];
         if (nodes && selector) {
```
kevinpschaaf, line 64: The `nodes` test is no longer required here.

kevinpschaaf (changes requested):
Seems like we should maybe document this as a breaking change (and include it in the upgrade guide?).

I could see users having written code that tested for `null` that needs to be updated to `nodes.length === 0`.

dfreedm:
> Seems like we should maybe document this as a breaking change (and include it in the upgrade guide?).
> 
> I could see users having written code that tested for `null` that needs to be updated to `nodes.length === 0`.

The documentation for this API sort of implies that the return value was never null.

packages/reactive-element/src/decorators/query-assigned-nodes.ts
dfreedm: Done

justinfagnani:
> I could see users having written code that tested for `null` that needs to be updated to `nodes.length === 0`

Presumably they would have to _also_ check for `nodes.length === 0` for when the slot exists and has no assigned nodes, so I don't think this is really that much of a breaking change. I would be happy to see the change description updated to be a bit more clear and say that an empty array is now returned if the slot doesn't exist.

kevinpschaaf (approved):
Seems like we should maybe document this as a breaking change (and include it in the upgrade guide?).

I could see users having written code that tested for `null` that needs to be updated to `nodes.length === 0`.

sorvell (approved):

justinfagnani (approved):

Issue #1996, opened 2021-07-07
How to get the name of a custom element through the class?


luzhenqian:
I have a component, which is defined as follows:
```TS
@customElement("zt-checkbox")
export default class Checkbox extends LitElement {
  render() {
    return html`<div></div>`;
  }
}
```

I want to get a custom name by class(zt-checkbox), the pseudo code is as follows:

```ts
const name = Checkbox.customElementName// zt-checkbox
```

But the class does not contain this field, what should i do?

justinfagnani:
What's your use case?

We could have the `@customElement()` decorator write the class name somewhere on the class, but we're reluctant to do that because there's no standard property to use, so it wouldn't be available on elements built with other libraries, and the tag name wouldn't be there if you used `customElements.define()`, which is what you do in JS and completely valid in TypeScript as well. The property could also be wrong in cases where you subclass an element but don't use the decorator to register it.

For some background on adding this to the custom elements API see: https://github.com/WICG/webcomponents/issues/566

The most general thing to do is to patch CustomElementRegistry so you can get the tag name for any class:

```ts
const originalDefine = CustomElementRegistry.prototype.define;
const tagNameMap = new Map();
CustomElementRegistry.prototype = function(name, constructor, options) {
  tagNameMap.set(name, constructor);
  return originalDefine.call(this, name, constructor, options);
};
```

matthewp:
Is there an issue with doing: `new Checkbox().localName` ?

luzhenqian:
> 你的用例是什么？
> 
> 我们可以让`@customElement()`装饰器在类的某个地方写下类名，但我们不愿意这样做，因为没有标准属性可以使用，所以它不会在用其他库构建的元素上可用，并且标签名不会如果您使用`customElements.define()`，则不会在那里，这就是您在 JS 中所做的并且在 TypeScript 中也完全有效。在您对元素进行子类化但不使用装饰器注册它的情况下，该属性也可能是错误的。
> 
> 有关将此添加到自定义元素 API 的一些背景信息，请参阅：[WICG/webcomponents#566](https://github.com/WICG/webcomponents/issues/566)
> 
> 最常见的做法是修补 CustomElementRegistry 以便您可以获得任何类的标记名称：
> 
> ```ts
> const  originalDefine  =  CustomElementRegistry 。原型。定义；
> const  tagNameMap  =  new  Map ( ) ; 
> 自定义元素注册表。原型 = 函数（名称、 构造函数、 选项） { 
[...]

PR #1999, opened 2021-07-13
Add localName for lit-labs/ssr

matthewp:
This adds the localName to an element when registering. This means if you have an element constructor function you can find the tag name in order to render it.

```js
let tag = new MyElement().localName;
let instance = new LitElementRenderer(tag);
// ...
```

### DIFF
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -109,6 +109,14 @@ export const getWindow = ({
         observedAttributes:
           (ctor as CustomHTMLElement).observedAttributes ?? [],
       });
+      function localName() {
+        return name;
+      }
+      Object.defineProperty(ctor.prototype, 'localName', {
+        configurable: true,
+        enumerable: true,
+        get: localName,
+      });
     }
 
     get(name: string) {


packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -109,6 +109,14 @@ export const getWindow = ({
         observedAttributes:
           (ctor as CustomHTMLElement).observedAttributes ?? [],
       });
+      function localName() {
+        return name;
+      }
+      Object.defineProperty(ctor.prototype, 'localName', {
```
justinfagnani, line 115: Rather than add a property to the element, which could conceivably collide with a class property and wouldn't work with scoped registries, and may have problems with inheritance (ie https://github.com/lit/lit/issues/1996#issuecomment-875811141)

I'd rather see a method added the registry to retrieve the tag name. Since this seems somewhat likely to be added in standards in the future, it should have a definitely prollyfill-type name.

justinfagnani (commented):
Since this isn't used anywhere in this PR, what's the use case?

matthewp:
Use case is:

```js
import { MyElement } from './some-element.js';
import { LitElementRenderer from '@lit-labs/ssr/lib/lit-element-renderer.js';

let instance = new LitElementRenderer(new MyElement().localName);
// ... ssr this element
```

packages/labs/ssr/src/lib/dom-shim.ts
matthewp: So do you mean to add `localName` to the HTMLElement shim and then have it call out to the custom element registry? I could do that I guess, but I don't think it would work with scoped registries either.

justinfagnani: No, I mean add a method to CustomElementRegistry. Like:

```
const tagName = customElements.prollyfillGetTagName(MyElement);
```

The registry can store the names in a map.

matthewp: I would want to also add a getter on HTMLElement for localName, is that ok? The goal for me as an end-user is to only use standard APIs. Ideally I could use Lit with another shim (as I'm doing elsewhere), and that shim won't have `prollyfillGetTagName` but will have localName. 

So adding a `prollyfillGetTagName` and then using that within HTMLElement.prototype.localName makes sense to me.

justinfagnani: `localName` already exists on the prototype, but for instances and throws if you try to access it directly off the prototype object. Not throwing and returning a value from `MyElement.prototype.localName` is just as non-standard as a patched method on `CustomElementRegistry`. It's also not compatible with scoped registries where different instances with the same prototype might return different tag names.

This functionality could be written as a patch on `CustomElementRegistry.prototype.define` and a library function that takes and registry and constructor to get the name, and that would work on any standard implementation of CustomElementRegistry, even client-side. Would that suit your needs? What I like about that is that it's obvious that this is non-standard behavior from importing the library and the library can document that it needs to be patched in before elements are registered.

matthewp: There might be some confusion here. `localName` does not exist on the prototype in this shim. I am *not* wanting to access it off the prototype. I only want to access it on the instance through `new MyElement().localName`. I can even make it throw.

To be clear, all I want to do here is implement the standard `localName` behavior in this shim.

My first approach was to do it this way, but as you pointed out it's not correct in a number of ways. The shim is not correct in many other ways, so I didn't think correctness was a big concern of the shim. But I'm willing to make this as correct as you want it to be.  But I don't want to use a 3rd party package to access the localName as I don't need that in the browser or anywhere else that supports the Element interface.

matthewp:
It sounds like there isn't desire to add `localName` support to the shim, so closing.

packages/labs/ssr/src/lib/dom-shim.ts
kevinpschaaf: I think the meta point on this issue is that while it's trivial to add `localName` to the shim (and we could do that), the problem is that even if the shim had `localName`, newing up a CE _just_ to get its `localName` is too wasteful of a pattern to use for mapping a CE constructor to its tagName.

While that's maybe the only standard way to do it, it's just a bad pattern overall and likely worth patching customElements.define yourself to get that information while we wait on a better spec-based way to do it.

matthewp: I'm not sure I agree, creating an object itself is not expensive. It's possible that a constructor might do something expensive but it's generally advised against; most custom element constructors create a shadowRoot and do little else.

Patching customElements.define relies on script ordering, so the user has to run code in a particular order or otherwise they might define the element before the patch has been right, in which case you won't have your map.

PR #2002, opened 2021-07-13
Correct polyfill-support styling issues

### LINKED ISSUE #2000: [reactive-element] polyfill-support styling does not shim custom properties by firstUpdated/update
Styles should be fully applied by firstUpdate/update time, including any shimmed css custom properties. This is not the case because ShadyCSS's `styleElement` is [called after](https://github.com/lit/lit/blob/main/packages/reactive-element/src/polyfill-support.ts#L149-L153) `didUpdate` which calls firstUpdate/update.

### LINKED ISSUE #2001: [lit-html] polyfill-support should retain late added styles for back compat
Styles added after the first render should be retained. These styles will *not* be shimmed so they will leak to the document. However, this behavior is consistent with what happened in Lit 1 and since we're not fixing the core behavior (that late added styles are not properly shimmed), changing this does not seem worth it since it can break upgrading users.

sorvell:
Fixes #2000: ShadyCSS's styleElement is now called before firstUpdate/update so that code in these methods can properly assume styling (including shimmed custom properties) has fully applied.

Fixes #2001: Late added styles are now retained in templates. This behavior is consistent with what happened in Lit 1 and we're electing not to change it so that users that may be relying on this (admittedly weird and not well supported) behavior are not broken.

### DIFF
--- a/.changeset/beige-rules-kneel.md
+++ b/.changeset/beige-rules-kneel.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'@lit/reactive-element': patch
+---
+
+Fixes polyfill-support styling issues: styling should be fully applied by firstUpdated/update time; late added styles are now retained (matching Lit1 behavior)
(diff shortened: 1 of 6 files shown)


kevinpschaaf (approved):
LGTM

PR #2016, opened 2021-07-20
Replace local type declarations for polyfill APIs with those imported from the polyfills.

bicknellr:
This PR removes the local types describing the polyfill APIs in favor of the new TS externs from the polyfill repos.

- [x] https://github.com/webcomponents/polyfills/pull/457
- [x] Remove local tarball and use released polyfill packages instead.


bicknellr:
I'm going to mark this as ready for review despite this still using the local copy of the polyfills.

kevinpschaaf (approved):
LGTM, just one similar question to [lit-element#1208](https://github.com/lit/lit-element/pull/1208#pullrequestreview-723669333) as to why `ShadowRoot` is needed in env.d.ts: https://github.com/lit/lit/pull/2016/files#diff-6f666fa91821d7d88b3f61b91e6bf31a350cb7da17ec7f460c1407dc970c16eeR24

bicknellr:
(We talked in a meeting, I'll remove extra types in a follow up PR.)

PR #2028, opened 2021-07-28
[localize] Escape markup characters when generating HTML text content

aomarks:
Fixes a bug where we were not escaping `<`, `>`, and `&` characters to HTML entities when generating HTML text content in both runtime and transform mode.

cc @fetherston


kevinpschaaf (approved):
LGTM

fetherston:
Tests and fix make sense to me! Thanks for putting this together 🚀

PR #2034, opened 2021-08-04
[lit-html, reactive-element] Revert change to pause update cycle while disconnected

kevinpschaaf:
Reverts the change in Lit 2 to pause ReactiveElement's update cycle while the element is disconnected.

This change was originally added to Lit 2 to ensure that once AsyncDirectives are disconnected, they do not receive further `updates` until they are reconnected.

However, the change in Lit 2 to pause updates while the element was disconnected was found during internal testing to be subtly breaking in potentially dangerous ways, which could lead to memory leaks if the update cycle was relied on to clean up external event handlers/subscriptions.

Instead, the onus will shift to `AsyncDirective` authors to check the `this.isConnected` flag during `update` and avoid doing any potentially leaky work while disconnected, since they may never be reconnected in that state.

Note that calling `setValue` from an `asyncDirective` while disconnected will now synchronously commit the change (since it seems arbitrary now to not do that; that was only in to maintain the same guarantee we're removing above (that disconnected directives don't receive updates).

A follow-on PR (#2044) re-works the disconnection handling for first-party `AsyncDirectives` (`until`, `asyncAppend`, `asyncReplace`) to allow them to be GC'ed before their awaited promises resolve.


packages/lit-html/src/async-directive.ts
```diff
@@ -303,6 +302,7 @@ export abstract class AsyncDirective extends Directive {
     parent: Disconnectable,
     attributeIndex: number | undefined
   ) {
+    this.isConnected = (part.options as PrivateRenderOptions).isConnected;
```
sorvell, line 305: Should we consider naming the private key in renderOptions more obtusely since users can pass custom values here and we don't want to conflict?

justinfagnani: I think it's basically fine to conflict if we always overwrite. It's the same as ignore the user-provided value. If we only sometimes overwrote it'd be an issue, IMO.

kevinpschaaf: Since this is an implementation detail that we don't want anyone to even be reading from, I went ahead and renamed it to `_$isConnected` and made it a `stableProperty`.

packages/lit-html/src/experimental-hydrate.ts
```diff
@@ -114,6 +114,9 @@ export const hydrate = (
   container: Element | DocumentFragment,
   options: Partial<RenderOptions> = {}
 ) => {
+  options ??= {};
+  (options as PrivateRenderOptions)._$isConnected = true;
+
```
justinfagnani:
```suggestion
options = {...options, _$isConnected: true};
```

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -945,6 +958,8 @@ class ChildPart {
   private _textSanitizer: ValueSanitizer | undefined;
   /** @internal */
   _$parent: Disconnectable | undefined;
+  /** @internal */
+  __isConnected = true;
```
sorvell, line 962: Should this be initialized via options?

sorvell: Why is this distinct from this.options' `isConnected`?

packages/lit-html/src/lit-html.ts
```diff
@@ -394,9 +402,14 @@ export const render = (
       container.insertBefore(createMarker(), endNode),
       endNode,
       undefined,
-      options
+      options || {}
     );
   }
+  // Pass top-level part connection status to AsyncDirectives via options
+  (part.options as PrivateRenderOptions) = {
```
sorvell, line 409: Only copy this if there's no isConnected flag?

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -2066,18 +2067,37 @@ suite('lit-html', () => {
...
       assertContent(`<div>initial</div>`);
+      await promise;
+      assert.isFalse(aDirectiveInst.isConnected);
+      assertContent(`<div>resolved1</div>`);
       part.setConnected(true);
+      assert.isTrue(aDirectiveInst.isConnected);
       assertContent(`<div>resolved1</div>`);
     });
 
+    test('async directives rendered while disconnected in ChildPart', async () => {
```
sorvell, line 2086:
```suggestion
    test('async directives render while disconnected in ChildPart', async () => {
```

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -2781,157 +2781,6 @@ suite('ReactiveElement', () => {
     });
   });
```
sorvell, line 2783: Add a test to verify that updates happen while disconnected?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
```diff
@@ -394,9 +402,10 @@ export const render = (
       container.insertBefore(createMarker(), endNode),
       endNode,
       undefined,
-      options
+      options || {}
     );
   }
+  (part.options as PrivateRenderOptions)._$isConnected = part.__isConnected;
```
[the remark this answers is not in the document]
kevinpschaaf: Done (but copied only once, when we create the part).

packages/lit-html/src/experimental-hydrate.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
```diff
@@ -394,9 +402,14 @@ export const render = (
       container.insertBefore(createMarker(), endNode),
       endNode,
       undefined,
-      options
+      options || {}
```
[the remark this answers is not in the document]
kevinpschaaf: Updated to only copy the options when creating the part.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Moved the copy into the one-time part creation step.

packages/lit-html/src/lit-html.ts
kevinpschaaf: The connection state of the top-level part (the part created/returned by `render` that the user can call `part.setConnection(true/false)` on) is the source-of-truth for everything below it. The `options._$isConnected` flag is not the source of truth, but rather a communication channel between the top-level part and everything below it since they all get references to this one options bag passed down.

> Should this be initialized via options?

In theory, the `__isConnected` flag on any parts below the top-level part returned is never used (and the top-level part should never get an `options` bag with `_$isConnected` in it), so it doesn't really matter, but we could keep it in sync just for consistency.

kevinpschaaf: I think there's a flaw though... It's possible to get a reference to a child part that shares `options` with the top-level part, and then call `part.setConnected(false)` and change the connection state. From there, if you did `part.setValue(...)`, directives would still see the top-level connection state, not the local one.

packages/reactive-element/src/test/reactive-element_test.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Per design conversation, re-worked to use `get _$isConnected() { return this._$parent._$isConnected }` on all Disconnectables up to the `RootChildPart`, which holds the state for the entire tree.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -118,12 +118,19 @@ export const insertPart = (
...
+        part._$notifyConnectionChanged !== undefined &&
+        newConnectionState !== part._$isConnected;
       part._$reparentDisconnectables?.(containerPart);
       // Note that although `_$reparentDisconnectables` updates the part's
       // `_$parent` reference after unlinking from its current parent, that
       // method only exists if Disconnectables are present, so we need to
       // unconditionally set it here
       part._$parent = containerPart;
+      if (shouldNotifyConnectionChanged) {
+        part._$notifyConnectionChanged?.(newConnectionState);
```
justinfagnani, line 132: You don't need both the optional chaining here and the `part._$notifyConnectionChanged !== undefined` above.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -118,12 +118,19 @@ export const insertPart = (
     const endNode = wrap(part._$endNode!).nextSibling;
     const parentChanged = part._$parent !== containerPart;
     if (parentChanged) {
+      const newConnectionState = containerPart._$isConnected;
+      const shouldNotifyConnectionChanged =
+        part._$notifyConnectionChanged !== undefined &&
+        newConnectionState !== part._$isConnected;
```
justinfagnani, line 124: Do these have to be computed before the next line? I'd move them closer to the `part._$notifyConnectionChanged` call if possible.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -837,7 +835,6 @@ export abstract class ReactiveElement
     // connection promise
     if (this.__enableConnection) {
       this.__enableConnection();
-      this.__pendingConnectionPromise = this.__enableConnection = undefined;
```
justinfagnani, line 840: Can you write an unconditional `this.__enableConnection.?()` now?

justinfagnani, line 840: Is `__enableConnection` even set anymore? Can you remove it?

packages/lit-html/src/test/test-utils/memory.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+const canRunMemoryTests =
+  globalThis.performance?.memory?.usedJSHeapSize && window.gc;
+
+export const memorySuite = canRunMemoryTests ? suite : suite.skip;
+
+export const forceGC = () => {
+  window.gc();
```
justinfagnani, line 24: Do you know of any links that talk about using window.gc() in tests? Is 10 calls the most reliable way to do it? Would waiting a tick help?

packages/lit-html/src/lit-html.ts
```diff
@@ -1230,8 +1234,31 @@ class ChildPart {
...
 
+/**
+ * A top-level `ChildPart` returned from render that manages the connected state
+ * of `AsyncDirective`s created throughout the tree below it.
+ */
+export type {RootChildPart};
+class RootChildPart extends ChildPart {
+  private __isConnected = true;
+
+  get _$isConnected() {
```
justinfagnani, line 1245: I wonder if you could save a couple of bytes by making `_$isConnected` the property? You'd need defineProperty to not error on the base class not having a setter, so maybe not.

justinfagnani (changes requested):

packages/lit-html/src/lit-html.ts
```diff
@@ -1230,8 +1234,31 @@ class ChildPart {
   }
 }
 
+/**
+ * A top-level `ChildPart` returned from render that manages the connected state
+ * of `AsyncDirective`s created throughout the tree below it.
+ */
+export type {RootChildPart};
+class RootChildPart extends ChildPart {
```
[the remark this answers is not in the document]
kevinpschaaf: It didn't change the size, so I liked the clarity of RootChildPart.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Considered that, but defineProperty is also slow, in addition to bytes.

packages/lit-html/src/test/test-utils/memory.ts
kevinpschaaf: Not really. Calling it more than once is cargo-culted from somewhere, including that calling it in a loop doesn't have the same effect; I can't recall where from.

packages/lit-html/src/lit-html.ts
```diff
@@ -753,6 +754,7 @@ class Template {
 export interface Disconnectable {
```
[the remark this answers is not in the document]
kevinpschaaf: Tried, but it increased size.

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: Removed, was dead code.

packages/reactive-element/src/reactive-element.ts
kevinpschaaf: Removed, was dead code.

kevinpschaaf: Removed, was dead code.

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: `part._$isConnected` relies on the part's current parent, which is changed on the next line.

I refactored it a bit, PTAL

justinfagnani: looks good

packages/lit-html/src/lit-html.ts
justinfagnani: And we're TODO'ing setting the initial connected state in render because in LitElement it's almost always going to be first rendered when connected, yeah?

packages/lit-html/src/directive-helpers.ts
```diff
@@ -116,14 +116,26 @@ export const insertPart = (
...
       // Note that although `_$reparentDisconnectables` updates the part's
       // `_$parent` reference after unlinking from its current parent, that
       // method only exists if Disconnectables are present, so we need to
       // unconditionally set it here
       part._$parent = containerPart;
+      // Since the _$isConnected getter is somewhat costly, only
+      // read it once we know the subtree has directives that need
+      // to be notified
+      let newConnectionState;
+      if (
```
sorvell, line 132: Why was this not needed before? Was it just an omission? If so did we add test coverage?

Let's also add a comment here explaining that when a part is moved, its connection state needs to be updated to match the part it's moved into.

packages/lit-html/src/directive.ts
```diff
@@ -110,10 +110,14 @@ export abstract class Directive {
   _$disconnectableChildren?: Set<Disconnectable>;
   // This property needs to remain unminified.
   //@internal
-  ['_$setDirectiveConnected']?(isConnected: boolean): void;
+  ['_$notifyDirectiveConnectionChanged']?(isConnected: boolean): void;
 
   constructor(_partInfo: PartInfo) {}
 
+  get _$isConnected() {
```
sorvell, line 117: Add a comment here about how this works and why it's here.

packages/lit-html/src/directives/cache.ts
```diff
@@ -71,7 +77,6 @@ class CacheDirective extends Directive {
           clearPart(containerPart);
           insertPart(containerPart, undefined, cachedPart);
           setCommittedValue(containerPart, [cachedPart]);
-          cachedPart.setConnected(true);
```
sorvell, line 74: This is removed because `insertPart` handles it now?

packages/lit-html/src/directives/cache.ts
```diff
@@ -48,12 +54,12 @@ class CacheDirective extends Directive {
       if (cachedContainerPart === undefined) {
         const fragment = document.createDocumentFragment();
         cachedContainerPart = render(nothing, fragment);
+        cachedContainerPart.setConnected(false);
```
sorvell, line 57: This is set earlier here because `insertPart` synchronizes the connection state?

packages/lit-html/src/lit-html.ts
```diff
@@ -821,6 +823,10 @@ class TemplateInstance {
     this._$parent = parent;
   }
 
+  get _$isConnected() {
```
sorvell, line 826: Add a comment describing this.

packages/lit-html/src/test/test-utils/memory.ts
sorvell: Let's at least add a comment.

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
kevinpschaaf: Yeah, added TODO's, a skipped failing test for this (https://github.com/lit/lit/pull/2034/files#diff-fc790d086c06e1adeb4c976a60945f0b9b50aa86350a9c9c6b61618c48bac95dR522), and issue https://github.com/lit/lit/issues/2051

packages/lit-html/src/directive-helpers.ts
kevinpschaaf: Not an omission per se; prior to this the user could change any part's connection state, so it was up to the user moving a part to keep it in sync.

With this change, only the root part holds connection state, and everything under it is kept in sync / notified automatically. Added a test in `directive-helpers_test.ts`.

packages/lit-html/src/directive.ts
kevinpschaaf: Added a comment

packages/lit-html/src/directives/cache.ts
kevinpschaaf: Right

packages/lit-html/src/directives/cache.ts
kevinpschaaf: Right, previously we changed the connection state on the `ChildPart` that we inserted and moved between the cache and the container parts as we moved them. Now, `cachedContainerPart` is a `RootChildPart`, so we set it to "disconnected" once, and each time we move a part in and out, its connection state will be synced to the root part it is in.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Documented once in the interface and added comments for each one pointing to it.

packages/lit-html/src/async-directive.ts
```diff
@@ -388,21 +359,17 @@ export abstract class AsyncDirective extends Directive {
...
-        // this.__attributeIndex will be defined in this case, but
-        // assert it in dev mode
-        if (DEV_MODE && this.__attributeIndex === undefined) {
-          throw new Error(`Expected this.__attributeIndex to be a number`);
-        }
-        const newValues = [...(this.__part._$committedValue as Array<unknown>)];
-        newValues[this.__attributeIndex!] = value;
-        (this.__part as AttributePart)._$setValue(newValues, this, 0);
-      }
+    if (isSingleExpression(this.__part as unknown as PartInfo)) {
```
sorvell, line 362: The docs above this still say the value is queued. Need to update that.

packages/lit-html/src/test/test-utils/memory.ts
```diff
@@ -0,0 +1,34 @@
...
+    };
+  }
+}
+
+const canRunMemoryTests =
+  globalThis.performance?.memory?.usedJSHeapSize && window.gc;
+
+export const memorySuite = canRunMemoryTests ? suite : suite.skip;
+
+export const forceGC = () => {
```
sorvell, line 23: Still wan to add a comment about why this is being called 10x.

sorvell (changes requested):

packages/lit-html/src/test/test-utils/memory.ts
kevinpschaaf: This file actually goes with the next PR. Will test there and remove if unneeded.

packages/lit-html/src/async-directive.ts
kevinpschaaf: Good catch, done.

sorvell: Can't verify.

sorvell (approved):
Looks like the doc for `setValue` wasn't committed, but approving assuming you'll fix that.

packages/lit-html/src/lit-html.ts
```diff
@@ -1221,17 +1241,37 @@ class ChildPart {
...
-    this._$setChildPartConnected?.(false, true, from);
+    this._$notifyConnectionChanged?.(false, true, from);
     while (start && start !== this._$endNode) {
       const n = wrap(start!).nextSibling;
       (wrap(start!) as Element).remove();
       start = n;
     }
   }
+  /** @internal */
+  setConnected(isConnected: boolean) {
```
justinfagnani, line 1252: I think we should still add some docs that say this should only be called on root parts.

packages/lit-html/src/lit-html.ts
```diff
@@ -1221,17 +1241,37 @@ class ChildPart {
...
+    }
+  }
+}
+
+/**
+ * A top-level `ChildPart` returned from render that manages the connected state
+ * of `AsyncDirective`s created throughout the tree below it.
+ */
+export interface RootChildPart extends ChildPart {
+  setConnected(isConnected: boolean): void;
```
justinfagnani, line 1270: Add jsdocs here.

justinfagnani (approved):

justinfagnani:
LGTM, though I still prefer the name RootPart. RootChildPart just sounds a bit awkward and contradictory to me.

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

packages/lit-html/src/lit-html.ts
kevinpschaaf: Done

kevinpschaaf:
@justinfagnani Ok I think I changed my mind. I liked `RootChildPart` from the point of view of it being important for users to understand that it had all the API of a `ChildPart` plus the `RootPart` API (setConnected). But on further reflection, there's actually very-little-to-no useful `ChildPart` API that a user would need to know about/use on the value returned from `render`. So I think I'm good with that just being `RootPart`.

PR #2041, opened 2021-08-07
Remove unused type declarations and remove `LitExtendedWindow` and `LitExtraGlobals`.

bicknellr:
In this PR, I've removed as many declarations as I could from the `.d.ts` files as well as reverted (w/ conflict resolution) https://github.com/lit/lit/pull/1904. I think in https://github.com/lit/lit/pull/1904 I was trying to work around collisions with the polyfills' JS externs internally but, now that that is fixed, I think it was a mistake to not leave the other types (`litHtmlPlatformSupport`, etc.) on the global `Window` interface.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -42,7 +42,7 @@ if (DEV_MODE) {
 
   // Issue platform support warning.
   if (
-    (window as LitExtraGlobals).ShadyDOM?.inUse &&
+    window.ShadyDOM?.inUse &&
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (globalThis as any)['reactiveElementPlatformSupport'] === undefined
   ) {
--- a/.changeset/four-pianos-marry.md
+++ b/.changeset/four-pianos-marry.md
@@ -1,0 +1,7 @@
+---
+'lit-element': patch
+'lit-html': patch
+'@lit/reactive-element': patch
+---
+
+Remove some unnecessary internal type declarations.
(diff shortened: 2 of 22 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -42,7 +42,7 @@ if (DEV_MODE) {
 
   // Issue platform support warning.
   if (
-    (window as LitExtraGlobals).ShadyDOM?.inUse &&
+    window.ShadyDOM?.inUse &&
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (globalThis as any)['reactiveElementPlatformSupport'] === undefined
```
bicknellr, line 47: I think getting `reactiveElementPlatformSupport`, etc. off of the global shouldn't require `as any` anymore?

justinfagnani: Seems like it, did you try?

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts
bicknellr: Not yet. I'll try and then remove these if it works.

bicknellr: There's a bit more fallout than I expected around removing these casts, so I've put them in a separate PR with this PR as the base branch: #2043. Should I go ahead and merge this one or wait until the other one is ready too?

bicknellr: Internal builds seem to be fine with that new PR, so I'll just wait until a call is made there.

bicknellr: Based on our meeting this morning, I'm going to merge this PR before #2043 since it's already approved. Before that, I'll merge `main` here again to double check that nothing that would affect the tests has changed in the meantime.

PR #2043, opened 2021-08-09
Fix declarations to avoid `globalThis as any`.

bicknellr:
As far as I can tell, adding these properties to `interface Window` doesn't actually make them available on `globalThis`, which is why we needed to cast `globalThis as any` before. Further, the cast means that the types that were declared in this file weren't actually being used, so they didn't actually match. I haven't found another way to add something to `globalThis` besides `declare var`, which is what I did here, but this also causes TS to complain about the use of subscript syntax to retrieve them, so I've also removed that. I don't _think_ this effects minification because, externally, we don't use Closure for minification and, internally, TS declarations that end up affecting globals should be rewritten by Tsickle to externs that prevent them from being minified. Though, I might need to wrap these in a `declare global {...}` if I'm wrong.

- [x] Confirm minification isn't broken internally.


bicknellr:
AFAICT, minification internally is working with this change.

justinfagnani (approved):

PR #2044, opened 2021-08-10
[lit-html] Re-work async directives for better gc behavior

kevinpschaaf:
Re-works the disconnection handling for first-party `AsyncDirectives` (`until`, `asyncAppend`, `asyncReplace`) to use a WeakMap to connect an awaited promise's resolver and the directive that is awaiting it, so that a disconnected directive (and all of the DOM it references via its part) is not prevented from being GC'ed until its awaited promise resolves.

This effectively means they will have the previous behavior of deferring their commit until they are re-connected, but with the benefit of not leaking their DOM while disconnected.

Follow-on to #2034.


packages/lit-html/src/test/directives/async-append_test.ts
```diff
@@ -138,18 +139,106 @@ suite('asyncAppend', () => {
...
+      const big = () => new Array(10000).fill(0);
+      // Hold onto the iterables to prevent them from being gc'ed
+      const iterables: Array<TestAsyncIterable<string>> = [];
+      forceGC();
+      const heap = performance.memory.usedJSHeapSize;
+      for (let i = 0; i < 1000; i++) {
+        // Promise passed to until that will never resolve
+        const iterable = new TestAsyncIterable<string>();
+        iterables.push(iterable);
+        // Render the until into a `<span>` with a 10kb expando, to exaggerate
```
sorvell, line 225:
```suggestion
        // Render the directive into a `<span>` with a 10kb expando, to exaggerate
```

packages/lit-html/src/test/directives/async-append_test.ts
```diff
@@ -138,18 +139,106 @@ suite('asyncAppend', () => {
...
+    test('tree with asyncAppend cleared while iterables are pending', async () => {
+      const template = (v: unknown) => html`<div>${v}</div>`;
+      // Make a big array set on an expando exaggerate any leaked DOM
+      const big = () => new Array(10000).fill(0);
+      // Hold onto the iterables to prevent them from being gc'ed
+      const iterables: Array<TestAsyncIterable<string>> = [];
+      forceGC();
+      const heap = performance.memory.usedJSHeapSize;
+      for (let i = 0; i < 1000; i++) {
+        // Promise passed to until that will never resolve
```
sorvell, line 222:
```suggestion
        // Promise passed to directive that will never resolve
```

packages/lit-html/src/test/directives/async-replace_test.ts
```diff
@@ -225,18 +226,105 @@ suite('asyncReplace', () => {
...
+    test('tree with asyncReplace cleared while iterables are pending', async () => {
+      const template = (v: unknown) => html`<div>${v}</div>`;
+      // Make a big array set on an expando exaggerate any leaked DOM
+      const big = () => new Array(10000).fill(0);
+      // Hold onto the iterables to prevent them from being gc'ed
+      const iterables: Array<TestAsyncIterable<string>> = [];
+      forceGC();
+      const heap = performance.memory.usedJSHeapSize;
+      for (let i = 0; i < 1000; i++) {
+        // Promise passed to until that will never resolve
```
sorvell, line 308:
```suggestion
        // Promise passed to directive that will never resolve
```

packages/lit-html/src/test/directives/async-replace_test.ts
```diff
@@ -225,18 +226,105 @@ suite('asyncReplace', () => {
...
+        const iterable = new TestAsyncIterable<string>();
+        iterables.push(iterable);
+        // Render the until into a `<span>` with a 10kb expando, to exaggerate
```
sorvell, line 311:
```suggestion
        // Render the directive into a `<span>` with a 10kb expando, to exaggerate
```

sorvell (changes requested):

packages/lit-html/src/directives/async-append.ts
```diff
@@ -4,97 +4,43 @@
...
-      // transforms of iterables and async iterables requires a library,
-      // we accept a mapper function. This is especially convenient for
-      // rendering a template for each item.
-      if (mapper !== undefined) {
-        v = mapper(v, i);
-      }
-      const newPart = insertPart(part);
-      setChildPartValue(newPart, v);
-      i++;
+  // Override point for AsyncReplace to replace rather than append
```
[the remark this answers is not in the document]
kevinpschaaf: I made `AsyncAppend` subclass `AsyncReplace` since it's mostly identical code. As you can see in `AsyncAppend`, the only differences are:
* `validatePartType`: AsyncReplace can be used anywhere but AsyncAppend can only go in `ChildPart` position
* `update`: AsyncAppend needs to save the `__childPart` since it inserts new parts into it
* `commitValue`: AsyncAppend inserts new parts for each value, whereas AsyncReplace just calls `setValue` and overwrites the previous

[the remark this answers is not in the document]
kevinpschaaf: Oic, I had the comments backward. Fixed.

packages/lit-html/src/directives/async-append.ts
```diff
@@ -4,97 +4,43 @@
...
-    this._reconnectPromise = new Promise(
-      (resolve) => (this._reconnectResolver = resolve)
-    );
-  }
-
-  reconnected() {
-    // Resume iteration when reconnected
-    this._reconnectPromise = undefined;
-    this._reconnectResolver!();
+    const newPart = insertPart(this.__childPart);
```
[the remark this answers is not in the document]
kevinpschaaf: Creating and inserting a new part based on the next yielded value. This code is unchanged, just moved. Added a comment.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -5,74 +5,142 @@
...
+  private __mapper?: Mapper<unknown>;
+
+  constructor(partInfo: PartInfo) {
+    super(partInfo);
+    this.validatePartType(partInfo);
+  }
+
+  // Override point for AsyncReplace, which can only be used in ChildParts
+  protected validatePartType(_partInfo: PartInfo) {
+    // AsyncReplace can be used in any part type
```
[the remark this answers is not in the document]
kevinpschaaf: Ah, I had originally done `AsyncReplace extends AsyncAppend` which needed to loosen the part validation, but flipping it to `AsyncAppend extends AsyncReplace` lets `AsyncAppend` just do the stricter part validation in its constructor.

Removed.

packages/tests/web-test-runner.config.js
```diff
@@ -139,11 +139,18 @@ See https://wiki.saucelabs.com/display/DOCS/Platform+Configurator for all option
...
+    product: browser,
+    ...(browser === 'chromium'
+      ? {
+          launchOptions: {
+            args: ['--js-flags=--expose-gc', '--enable-precise-memory-info'],
+          },
+        }
+      : {}),
+  };
+  console.log(config);
```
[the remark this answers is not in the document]
kevinpschaaf: Removed

packages/lit-html/src/test/directives/async-append_test.ts
kevinpschaaf: Done

packages/lit-html/src/test/directives/async-append_test.ts
kevinpschaaf: Done

packages/lit-html/src/test/directives/async-replace_test.ts
kevinpschaaf: Done

packages/lit-html/src/test/directives/async-replace_test.ts
kevinpschaaf: Done

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -10,69 +10,127 @@ import {AsyncDirective} from '../async-directive.js';
...
-class AsyncReplaceDirective extends AsyncDirective {
-  private _value?: AsyncIterable<unknown>;
-  private _reconnectResolver?: () => void;
-  private _reconnectPromise?: Promise<void>;
+// When a directive is connected and actively awaiting a result from a promise,
+// the directive will be stored in `promiseToDirectiveMap`. The resolver will
+// get a reference back to the directive to perform the commit via the WeakMap
+// rather than closing over the directive instance, which would hold the
+// directive until the promise resolved.
+const promiseToDirectiveMap: WeakMap<Promise<unknown>, AsyncReplaceDirective> =
```
sorvell, line 18: What if the same iterable and thus promise is used in multiple directives? Does this need to be (can it be?) on the directive instance?

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -10,69 +10,127 @@ import {AsyncDirective} from '../async-directive.js';
...
+  private __awaitNextValue(nextPromise = this.__iterator!.next()) {
+    this.__nextPromise = nextPromise;
+    if (this.isConnected) {
+      // We still await the nextPromise even when disconnected (since if the
+      // directive reconnects it will need to handle the result), but only
+      // associate the directive to the promise when connected
+      promiseToDirectiveMap.set(nextPromise, this);
+    }
+    nextPromise.then((result) => {
+      const directive = promiseToDirectiveMap.get(nextPromise);
```
sorvell, line 75: Here `directive` is always the same as `this` isn't it? Does this mean we don't need to actually store the directive value and only need to know if this promise is mapped?

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -10,69 +10,127 @@ import {AsyncDirective} from '../async-directive.js';
...
-        await this._reconnectPromise;
-      }
-
-      // As a convenience, because functional-programming-style
-      // transforms of iterables and async iterables requires a library,
-      // we accept a mapper function. This is especially convenient for
-      // rendering a template for each item.
-      if (mapper !== undefined) {
-        v = mapper(v, i);
+  private __awaitNextValue(nextPromise = this.__iterator!.next()) {
```
sorvell, line 66: This is never called with the argument so let's remove it.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -10,69 +10,127 @@ import {AsyncDirective} from '../async-directive.js';
...
-    this.__iterate(mapper);
+    // If we're starting a new promise, stop tracking the last one, since its
+    // result is no longer relevant
+    if (this.__nextPromise !== undefined) {
+      promiseToDirectiveMap.delete(this.__nextPromise);
+    }
+    this.__mapper = mapper;
+    this.__iterable = iterable;
+    this.index = 0;
+    // Note, we use the iterator protocol manually (rather than for/await/of) so
```
sorvell, line 54: This is pretty tricky. It'd be really nice to have a `DisconnectablePromise` directive-helper to make this more generally available.

sorvell (changes requested):

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: The whole thing is not closing over `this` in the thenFn. Going through the WeakMap with the promise as the key is what breaks the dependency.

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Man, that's a good point (more for `until`, since I think that iterables are more likely to return fresh promises for each `next()` call). I can see if the WeakMaps can live on the directive instance if we pull it off first and only close over the WeakMap. Lemme try that.

packages/lit-html/src/directives/async-replace.ts
jridgewell: With a helper, you can actually continue using `for await`. The issue is that we're inside a concise method, which strongly holds onto the `this` value. If we defer to something like an `asyncForEach`, we've broken the inherent memory leak. We'll still need to indirect to get the `directve` value, but that's much simpler than writing manual iteration:

```js
async function asyncForEach(iterable, cb) {
  for await (const v of iterable) {
    if (await cb(v) === false) break;
  }
}
```

Adding in `=== false` allows us to break the iteration, if needed for any reason. And the `await cb(v)` would allow us to await a pause action, much like the current code:

<details>

<summary>AsyncPause</summary>

```js
class AsyncPause {
  _pause = null;
  _resolve = null;

  get() {
    return this._pause;
  }

  pause() {
    if (this._pause) return;
    this._pause = new Promise(resolve => {
      this._resolve = resolve;
    });
  }
  
  resume() {
    this._resolve?.();
    this._pause = null;
    this._resolve = null;
  }
}
```

</details>

```js
class AsyncDirective {
  _pauser = new AsyncPause();

  update(iterable) {
    const pauser = this._pauser;
    asyncForEach(iterable, async (value) => {
      if (pauser.get()) await pauser.get();

      // indirect to get directive here.
      if (directive._value !== iterable) return false;
    });
  }
}

**Comment by kevinpschaaf** [comment]
@jridgewell Cool, thanks for the inspiration here. I was actually working on factoring the bookkeeping out, but I think we decided to do that in a follow-on PR, since it can be an implementation detail change after we get the core behavior working.

File: packages/lit-html/src/directives/async-replace.ts (hunk 11 above)
Comment (kevinpschaaf) [reply packages/lit-html/src/directives/async-replace.ts]: Done

File: packages/lit-html/src/directives/async-replace.ts (hunk 9 above)
Comment (kevinpschaaf) [reply packages/lit-html/src/directives/async-replace.ts]: To address this, went ahead and refactored it into a `DisconnectableAwaiter` that wraps up the logic around indirecting the `this` ref to the directive, storing a value that resolves while disconnected, and flushing it back to the `thenFn` if resolved while disconnected.

File: packages/lit-html/src/directives/until.ts [hunk 13]
```diff
@@ -9,36 +9,51 @@ import {directive} from '../directive.js';
...
+// When a directive is disconnected and an awaited promise resolves, the
+// result will be stored in `promiseToResultMap` such that it can be retrieved
+// if/when it reconnects
+const promiseToResultMap: WeakMap<Promise<unknown>, unknown> = new WeakMap();
+
 const isPromise = (x: unknown) => {
   return !isPrimitive(x) && typeof (x as {then?: unknown}).then === 'function';
 };
 // Effectively infinity, but a SMI.
 const _infinity = 0x7fffffff;
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Added tests for same promise/iterable passed to multiple directive instances as well.

packages/lit-html/src/directives/disconnectable-awaiter.ts
```diff
@@ -0,0 +1,97 @@
...
+    // Note that if the directive is not connected when the promise is initially
+    // awaited, do not associte it with the promise, but still await it so that
+    // we can flush the result if/when the promise is reconnected to the
+    // directive. This is a small value-add that is easy to forget.
+    if (directive.isConnected) {
+      directiveMap.set(this, directive);
+    }
+    this._thenFn = thenFn;
+    promise.then((result: ResultType) => {
+      const weakDirective = directiveMap.get(this) as DirectiveType;
```
jridgewell, line 56: Because you're performing `disconnect()` and `reconnect()` manually, there shouldn't be a need to indirect through a `WeakMap`. Just storing this `directive` as an own property will be considerably faster. If the `DisconnectableAwaiter` gets collected, then own property will go with it the same as the weak key. And if it doesn't get collected, then both the property and key would stick around.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -7,72 +7,82 @@
...
+    // If we're starting a new promise, stop tracking the last one, since its
+    // result is no longer relevant
+    this.__nextAwaiter?.disconnect();
+    this.__mapper = mapper;
+    this.__iterable = iterable;
+    this.index = 0;
+    // Note, we use the iterator protocol manually (rather than for/await/of),
+    // which lets us use the `DisconnectableAwaiter` helper to disconnect &
+    // reconnect awaiting the next value from the iterable
+    this.__iterator = iterable[Symbol.asyncIterator]();
```
jridgewell, line 42: Nit: This is a slight change it behavior. `for await` will consume a sync iterator if the async iterator isn't define.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -7,72 +7,82 @@
...
-      // rendering a template for each item.
-      if (mapper !== undefined) {
-        v = mapper(v, i);
-      }
+  private __awaitNextValue() {
+    this.__nextAwaiter = new DisconnectableAwaiter(
+      this.__iterator!.next(),
+      this,
+      (directive, result) => directive.__commitResult(result)
+    );
```
jridgewell, line 52: This still seems to be a lot more complex than the alternative. Every loop requires a new object, 2 promises, and 2/3 closures.

The `AsyncPause` and `WeakRefWrapper` (could just be `PseudoWeakRef` without the fallback to real `WeakRef`) requires just 2 objects and a closure (all of which can be reused for every iteration) and internal promise allocation.

packages/lit-html/src/directives/disconnectable-awaiter.ts
kevinpschaaf: Doh, right.

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Reverted to `for await`

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Ok, I think that makes sense. Did another update to go with that approach. @justinfagnani @sorvell also PTAL

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -36,43 +53,49 @@ class AsyncReplaceDirective extends AsyncDirective {
   // because its return value must be `noChange`
   private async __iterate(mapper?: Mapper<unknown>) {
```
jridgewell, line 54: Nit: No longer async. If you want to return a promise still, we could `return forAwaitOf(…)` instead.
```suggestion
  private __iterate(mapper?: Mapper<unknown>) {
```

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -36,43 +53,49 @@ class AsyncReplaceDirective extends AsyncDirective {
...
-    const {_value: value} = this;
-    for await (let v of value!) {
-      // Check to make sure that value is the still the current value of
-      // the part, and if not bail because a new value owns this part
-      if (this._value !== value) {
-        break;
+    const {__value: value, __weakThis: weakThis, __pauser: pauser} = this;
+    forAwaitOf(value!, async (v: unknown) => {
+      // The while loop here handles the case that the connection state
+      // thrashes, causing the pauser to resume and then get re-paused
```
jridgewell, line 59: Good catch, I didn't think of that.

packages/lit-html/src/directives/until.ts
```diff
@@ -51,23 +63,44 @@ class UntilDirective extends AsyncDirective {
...
-      Promise.resolve(value).then((resolvedValue: unknown) => {
-        const index = this._values.indexOf(value);
-        // If state.values doesn't contain the value, we've re-rendered without
-        // the value, so don't render it. Then, only render if the value is
-        // higher-priority than what's already been rendered.
-        if (index > -1 && index < this._lastRenderedIndex) {
-          this._lastRenderedIndex = index;
-          this.setValue(resolvedValue);
+      // Note, the callback avoids closing over `this` so that the directive
+      // can be gc'ed before the promise resolves
```
jridgewell, line 70: Nit: do we want to copy this comment to the `AsyncReplaceDirective`, too?

jridgewell (approved):

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Good catch, it can be inlined back into update now.
Done.

packages/lit-html/src/directives/until.ts
kevinpschaaf: +1 Done

packages/lit-html/src/directives/async-helpers.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
justinfagnani, line 3: 2021

packages/lit-html/src/test/directives/async-append_test.ts
```diff
@@ -138,18 +142,144 @@ suite('asyncAppend', () => {
...
+      const template = (v: unknown) => html`<div>${v}</div>`;
+      // Make a big array set on an expando to exaggerate any leaked DOM
+      const big = () => new Array(10000).fill(0);
+      // Hold onto the iterables to prevent them from being gc'ed
+      const iterables: Array<TestAsyncIterable<string>> = [];
+      window.gc();
+      const heap = performance.memory.usedJSHeapSize;
+      for (let i = 0; i < 1000; i++) {
+        // Iterable passed to asyncAppend that will never yield
+        const iterable = new TestAsyncIterable<string>();
```
justinfagnani, line 264: Why do we create a new iterable per loop? Can we test with one iterable that never receives a value?

justinfagnani (approved):

packages/lit-html/src/directives/async-helpers.ts
kevinpschaaf: Done

packages/lit-html/src/directives/async-helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  deref() {
+    return this._ref;
+  }
+}
+
+/**
+ * A helper to pause and resume waiting on a condition in an async function
+ */
+export class Pauser {
```
sorvell: Perhaps bikeshedding, but it seems slightly better if this is called a `Player` and has `play` and `pause` methods. It could then have an `isPaused()` method which returns a promise that's resolved when `play` is called. Could that promise also just encapsulate the for loop that's currently used with `get`? The API seems less awkward that way.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -22,57 +34,64 @@ class AsyncReplaceDirective extends AsyncDirective {
...
+    const {__weakThis: weakThis, __pauser: pauser} = this;
+    // Note, the callback avoids closing over `this` so that the directive
+    // can be gc'ed before the promise resolves
+    forAwaitOf(value, async (v: unknown) => {
+      // The while loop here handles the case that the connection state
+      // thrashes, causing the pauser to resume and then get re-paused
+      while (pauser.get()) {
+        await pauser.get();
       }
+      const _this = weakThis.deref();
```
sorvell, line 47: Just for ultra clarity, suggest adding a comment here too that this avoids closing over `this` since the value is pulled out of the `weakThis` and that it's only there if the part is connected.

packages/lit-html/src/directives/async-replace.ts
```diff
@@ -22,57 +34,64 @@ class AsyncReplaceDirective extends AsyncDirective {
...
-      // the part, and if not bail because a new value owns this part
-      if (this._value !== value) {
-        break;
+    const {__weakThis: weakThis, __pauser: pauser} = this;
+    // Note, the callback avoids closing over `this` so that the directive
+    // can be gc'ed before the promise resolves
+    forAwaitOf(value, async (v: unknown) => {
+      // The while loop here handles the case that the connection state
+      // thrashes, causing the pauser to resume and then get re-paused
+      while (pauser.get()) {
```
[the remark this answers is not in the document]
sorvell: Or can this be collapsed into the class? See comment above.

packages/lit-html/src/test/directives/async-append_test.ts
kevinpschaaf: The iterable is passed into the `until` and then the entire part is cleared in the same loop. Each loop makes a new `span (with expando) <-- part <-- directive instance (with iterable) <-- awaited closure` which will leak if the closure strongly references the directive instance.

packages/lit-html/src/test/directives/test-async-iterable.ts
```diff
@@ -50,5 +50,6 @@ export class TestAsyncIterable<T> implements AsyncIterable<T> {
     // when devtools is closed. Waiting for rAF might be more reliable, but
     // this waits the minimum that seems reliable now.
     await Promise.resolve();
+    await Promise.resolve();
```
sorvell, line 53: Why was this added?

sorvell (changes requested):

sorvell (approved):
LGTM, consider my suggestions optional.

packages/lit-html/src/directives/async-helpers.ts
sorvell: Or maybe `Runner`?

packages/lit-html/src/directives/async-replace.ts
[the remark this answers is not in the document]
kevinpschaaf: `while (await pauser.get())` adds a microtask to every value when connected

packages/lit-html/src/directives/async-helpers.ts
kevinpschaaf: I started to implement it that way, but the final falsey check on `pauser.get()` must be synchronous to resuming, and you can't encapsulate that in an async function because the outside `await` adds a microtask between which `pauser.get()` might return truthy again and thus would be wrong.

packages/lit-html/src/directives/async-replace.ts
kevinpschaaf: Done

PR #2046, opened 2021-08-10
Throw when constructing an `EventPart` from a template attribute with extra content.

### LINKED ISSUE #2045: Extra text in attributes for `EventPart`s causes type confusion with regular `AttributePart`s.
# Description

When adding an event listener using the `@`-prefixed syntax in a template, if the browser-parsed attribute text contains more than just the marker text, the `AttributePart` constructor will set `.strings` to something other than `undefined` (and `._$committedValue` to an array of `nothing`):

https://github.com/lit/lit/blob/e6dc6a708adacec6a17a884784f821c3250d7532/packages/lit-html/src/lit-html.ts#L1276-L1278

Later, `TemplateInstance` tries to determine if a part is an `AttributePart`s by checking if it has non-`undefined` `.strings` and, if so, passes the array of all values to the part's `_$setValue`, which isn't what `EventPart`'s `_$setValue` expects:

https://github.com/lit/lit/blob/e6dc6a708adacec6a17a884784f821c3250d7532/packages/lit-html/src/lit-html.ts#L875-L876

# Steps to Reproduce

1. Write this code

```ts
import { html, render } from 'https://unpkg.com/lit-html@2.0.0-rc.3';

let message = 'not clicked';

const rerender = () => {
  const onClick = () => {
    message = 'clicked!';
    rerender();
  };

  render(
    html`
      <h1 @click="${onClick}ABC">${message}</h1>
    `,
    document.body
  );
};

rerender();
```

2. See this output...

No error is thrown or other warning given and the listener is not added. If you remove the "ABC" from the attribute, then the listener will be added as expected.

## Live Reproduction Link

https://jsbin.com/bifevafaho/edit?html,output

## Expected Results

The `EventPart` constructor should throw if the part's `strings` argument is anything other than `['', '']`.

## Actual Results

No error is thrown and the event listener is not added.

bicknellr:
Fixes #2045.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1432,6 +1432,21 @@ export type {EventPart};
 class EventPart extends AttributePart {
   readonly type = EVENT_PART;
 
+  constructor(
+    element: HTMLElement,
+    name: string,
+    strings: ReadonlyArray<string>,
+    parent: Disconnectable | undefined,
+    options: RenderOptions | undefined
+  ) {
+    super(element, name, strings, parent, options);
+
+    if (strings.length > 2 || strings[0] !== '' || strings[1] !== '') {
+      throw new Error('An event listener in a template has extra content ' +
+        'surrounding the listener expression.');
+    }
+  }
+
   // EventPart does not use the base _$setValue/_resolveValue implementation
   // since the dirty checking is more complex
   /** @internal */


packages/lit-html/src/lit-html.ts
```diff
@@ -1432,6 +1432,21 @@ export type {EventPart};
...
+    element: HTMLElement,
+    name: string,
+    strings: ReadonlyArray<string>,
+    parent: Disconnectable | undefined,
+    options: RenderOptions | undefined
+  ) {
+    super(element, name, strings, parent, options);
+
+    if (strings.length > 2 || strings[0] !== '' || strings[1] !== '') {
+      throw new Error('An event listener in a template has extra content ' +
```
XGHeaven, line 1445: I think it's better to give more information to help developer locate the position. Such as `name`/`strings`.

packages/lit-html/src/lit-html.ts
```diff
@@ -1432,6 +1432,21 @@ export type {EventPart};
...
 
+  constructor(
+    element: HTMLElement,
+    name: string,
+    strings: ReadonlyArray<string>,
+    parent: Disconnectable | undefined,
+    options: RenderOptions | undefined
+  ) {
+    super(element, name, strings, parent, options);
+
```
justinfagnani, line 1443: @bicknellr can you see if you put the entire body of the constructor inside an `if (DEV_MODE)` conditional if Terser will eliminate the constructor completely?

bicknellr: The constructor was apparently already here because it's used to assign `type`, but making this check depend on `DEV_MODE` also ended up converting its use of `...arguments` to into normal parameters:

Screen Shot 2021-08-16 at 17 13 27

bicknellr: Or, rather, adding this constructor with its regular arguments caused that change and checking `DEV_MODE` here prevented this new block from being added.

packages/lit-html/src/lit-html.ts
bicknellr: Pinning down the location might be a job better suited for lit-analyzer since there's not really a lot of context available. I did try this out and ended up going a bit overboard:

```diff
diff --git a/packages/lit-html/src/lit-html.ts b/packages/lit-html/src/lit-html.ts
index 07edbab8..d1e1e566 100644
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1445,9 +1445,18 @@ class EventPart extends AttributePart {
       DEV_MODE &&
       (strings.length > 2 || strings[0] !== '' || strings[1] !== '')
     ) {
+      const messageParts = [
+        {text: `    <${element.localName} ... @${name}="`},
+        {text: strings[0], isError: true},
+        {text: '${...}'},
+        {text: strings.slice(1).join('${...}'), isError: true},
+        {text: '" ...'},
+      ];
       throw new Error(
-        'An event listener in a template has extra content surrounding the ' +
-          'listener expression.'
+        `An event listener on a <${element.localName}> in a template has ` +
+          'extra content surrounding the listener expression:\n' +
+          messageParts.map(x => x.text).join('') + '\n' +
+          messageParts.map(x => (x.isError ? '^' : ' ').repeat(x.text.length)).join('')
       );
     }
   }
```

which spits out errors like this:

```
Error: An event listener on a <div> in a template has extra content surrounding the listener expression:
    <div ... @click="${...}}" ...
                           ^     
```

when the code looks like this:

```js
render(html`<div @click=${listener}}></div>`, container);
```

But there's no way to give a file and line number or print a more accurate representation of the source with what's available, as far as I can tell.

packages/lit-html/src/lit-html.ts
```diff
@@ -1496,6 +1496,26 @@ export type {EventPart};
...
+    if (
+      DEV_MODE &&
+      (strings.length !== 2 || strings[0] !== '' || strings[1] !== '')
+    ) {
+      throw new Error(
+        'An event listener in a template must have exactly one expression ' +
```
justinfagnani, line 1513: Use the name in the error for a little more hint to the developer as to where the problem is.

packages/lit-html/src/lit-html.ts
justinfagnani: that seems ok... just a few more chars

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
bicknellr: Updated, here's an example:
``A `<div>` has a `@click=...` listener with invalid content. Event listeners in templates must have exactly one expression and no surrounding text.``
WDYT?

justinfagnani: +1 sounds great

PR #2050, opened 2021-08-13
Fix documentation examples so they have syntax highlighting

AndrewJakubowicz:
## Context

Lit.dev pulls the API documentation out of the jsDoc comments. While auditing lit.dev I found some examples that aren't rendering correctly. Some of these examples also don't render correctly in the IDE. This is a small cleanup PR that unifies the example syntax.

## Why

Fixes some example code in the jsDocs on both lit.dev and hover IDE documentation, i.e.

Screen Shot 2021-08-13 at 11 05 39 AM

## How

There is two kinds of example syntax I could identify in the codebase. Inline triple backticks with a language annotation, and the `@example` attribute.
Currently lit.dev doesn't render the `@example` attribute (Filed: https://github.com/lit/lit.dev/issues/434), but does render inline code fences. Updated docs to use code fences with language declaration.

## Testing

Tested manually in VS code by hovering over the properties and eyeballing the tooltips. Risk should be very minor as this is a small documentation cleanup.

### DIFF
--- a/packages/lit-html/src/directives/cache.ts
+++ b/packages/lit-html/src/directives/cache.ts
@@ -88,7 +88,7 @@ class CacheDirective extends Directive {
  *
  * Example:
  *
- * ```
+ * ```js
  * let checked = false;
  *
  * html`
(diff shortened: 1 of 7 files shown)


aomarks (approved):
Nice!

justinfagnani (approved):

PR #2056, opened 2021-08-19
[lit-html, lit-element] Add initialIsConnected to RenderOptions. Fixes #2051

### LINKED ISSUE #2051: [lit-html] No way to initially render a part that is disconnected
# Description
While refactoring the AsyncDirective disconnection handling in https://github.com/lit/lit/pull/2034 and removing the Lit 2 change to remove pausing updates while disconnected, it's now possible for a LitElement to perform its initial render while disconnected. Since there is no way to tell a part to initially render in a disconnected state, those directives will be out of sync.

# Steps to Reproduce

Failing test added in https://github.com/lit/lit/pull/2034/files#diff-fc790d086c06e1adeb4c976a60945f0b9b50aa86350a9c9c6b61618c48bac95dR522

kevinpschaaf:
Adds the ability to set the initial connected state for the top-level part being rendered. The `part.setConnected()` method must be used subsequent to initial render to change the connected state of the part. Set to `false` if the initial render occurs in a disconnected tree and `AsyncDirective`s should see `isConnected === false` for their initial render.

Updates `LitElement` to set this flag in `connected`/`disconnectedCallback`. I went with just adding it into the `renderOptions` bag at the connection edges, rather than tracking connection state via a separate flag and assigning it in every render, since the edges happen less frequently.

This change ticks up the lit-html size by about ~20 bytes. Code golf ideas appreciated. The name is unfortunately long, but couldn't come up with something significantly shorter and still meaningful.

```
lit-html.js                     7.56 KB   7.56 KB   3.14 KB   2.86 KB   
lit-html.js                     7.63 KB   7.63 KB   3.16 KB   2.88 KB   
```

Fixes #2051

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -134,6 +134,7 @@ export class LitElement extends ReactiveElement {
    * @category lifecycle
    */
   connectedCallback() {
+    this.renderOptions.initialIsConnected = true;
     super.connectedCallback();
     this.__childPart?.setConnected(true);
   }
@@ -142,6 +143,7 @@ export class LitElement extends ReactiveElement {
    * @category lifecycle
    */
   disconnectedCallback() {
+    this.renderOptions.initialIsConnected = false;
     super.disconnectedCallback();
     this.__childPart?.setConnected(false);
   }
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -341,6 +341,14 @@ export interface RenderOptions {
    * any inherited context. Defaults to the global `document`.
    */
   creationScope?: {importNode(node: Node, deep?: boolean): Node};
+  /**
+   * The initial connected state for the top-level part being rendered.
+   * The `part.setConnected()` method must be used sebsequent to initial render
+   * to change the connected state of the part. Set to `false` if the initial
+   * render occurs in a disconnected tree and `AsyncDirective`s should see
+   * `isConnected === false` for their initial render.
+   */
+  initialIsConnected?: boolean;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


jridgewell:
Could this be done in the other direction? Instead of initializing to `true`, let it be `false`, and wait for the `connectedCallback` signal to tell it otherwise?

I'm out of date on the larger codebase, but I think only `AsyncDirective` reads it, then it shouldn't matter much (they're already async).

kevinpschaaf:
> Could this be done in the other direction? Instead of initializing to true, let it be false, and wait for the connectedCallback signal to tell it otherwise?

@jridgewell Unfortunately, no: `AsyncDirective`s still have their `update` called sync to render, and we want `this.isConnected` to return the correct value.

packages/lit-element/src/lit-element.ts
```diff
@@ -128,6 +128,7 @@ export class LitElement extends ReactiveElement {
    * @category lifecycle
    */
   connectedCallback() {
+    this.renderOptions.isConnected = true;
```
sorvell, line 131: A user could disconnect the element in their `connectedCallback` before calling `super.connectedCallback` and the state would then be wrong. This is a pretty corner case so not sure it matters, but alternatively, we could check `isConnected` directly at render time.

packages/lit-html/src/lit-html.ts
```diff
@@ -341,6 +349,14 @@ export interface RenderOptions {
...
    */
   creationScope?: {importNode(node: Node, deep?: boolean): Node};
+  /**
+   * The initial connected state for the top-level part being rendered.
+   * The `part.setConnected()` method must be used subsequent to initial render
+   * to change the connected state of the part. Set to `false` if the initial
+   * render occurs in a disconnected tree and `AsyncDirective`s should see
+   * `isConnected === false` for their initial render.
+   */
+  isConnected?: boolean;
```
sorvell, line 359: If it's missing, the directives are considered connected? That seems odd: the default is `true` but if you include the option, you typically would pass `true`. What about `isDisconnected` or `initializeAsDisconnected`?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
kevinpschaaf: The value is `??` checked, so _if_ it's provided, the value is respected, whichever way it's set. Seems less odd than flipping the polarity for the value that's used to control `isConnected`.

packages/lit-element/src/lit-element.ts
kevinpschaaf: You're suggesting to hit the DOM binding for every element's first render?

kevinpschaaf: At least in native, adding this to update doesn't seem to show up, so I guess we'll go with that...?
```ts
    if (!this.hasUpdated) {
      this.renderOptions.isConnected = this.isConnected;
    }
 ```
image

PR #2060, opened 2021-08-19
Enable noImplicitOverride in our tsconfigs, requiring us to add the `override` modifier on overridden fields

rictic:
This seems like a net win for clarity (more verbiage, but it lets the reader know when a method field has special meaning to the superclass), and it's easy on the code author because TS flags it immediately and the editor plugin offers to add the `override` keyword for you


justinfagnani (approved):

justinfagnani (changes requested):
Oops. un-approving since the tests are broken!

rictic:
Tests fixed! PTAL

justinfagnani (approved):

PR #2061, opened 2021-08-19
[monorepo] Add changeset to checks

kevinpschaaf:
Adds enforcing that all PR's need changesets, to streamline releases.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,21 @@ name: Tests
 on: [push, pull_request]
 
 jobs:
+
+  changeset:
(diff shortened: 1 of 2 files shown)


justinfagnani (approved):

bicknellr:
Shouldn't the changeset check have failed since this PR doesn't include a changeset?

kevinpschaaf:
@bicknellr Good question! Changes to the top-level monorepo are exempt, as it turns out. In https://github.com/lit/lit/pull/2061/commits/b1e3b28391cf0f9052c725a200f682e86cf5272a I tested changing lit-html's README to confirm it does fail.

PR #2065, opened 2021-08-19
@query and @queryAll return null or [] before first update

### LINKED ISSUE #2062: [ReactiveElement] `query`/`queryAll` should return similar results as Lit1
It was not an intended breaking change to alter the behavior of these decorators. However, they are subtly different. In particular if properties mapped to them are accessed before the first update (which should be a rare corner case), the returned result can be different.

In Lit2, `renderRoot` is created in `connectedCallback` and as a result these early accesses of properties with these decorators now return `undefined`, whereas previously they would return `null` or `empty NodeList`. This can easily be normalized to the Lit1 behavior to ensure less migration churn.

sorvell:
Fixes #2062. To match Lit1 behavior, the `@query` decorator returns `null` (rather than `undefined`) if a decorated property is accessed before first update. Note, when the `cache` option is used, the first result is always cached and it must be either `null` or the found element. Likewise, a `@queryAll` decorated property returns `[]` rather than `undefined`.

### DIFF
--- a/packages/reactive-element/src/decorators/query.ts
+++ b/packages/reactive-element/src/decorators/query.ts
@@ -61,13 +61,11 @@ export function query(selector: string, cache?: boolean) {
           ) {
             (this as unknown as {[key: string]: Element | null})[
               key as string
-            ] = this.renderRoot?.querySelector(selector);
+            ] = this.renderRoot?.querySelector(selector) ?? null;
           }
-          return (
-            (this as unknown as {[key: string]: Element | null})[
-              key as string
-            ] ?? null
-          );
+          return (this as unknown as {[key: string]: Element | null})[
+            key as string
+          ];
         };
       }
       return descriptor;
--- a/packages/reactive-element/src/decorators/query.ts
+++ b/packages/reactive-element/src/decorators/query.ts
@@ -46,7 +46,7 @@ export function query(selector: string, cache?: boolean) {
     descriptor: (name: PropertyKey) => {
       const descriptor = {
         get(this: ReactiveElement) {
-          return this.renderRoot?.querySelector(selector);
+          return this.renderRoot?.querySelector(selector) ?? null;
         },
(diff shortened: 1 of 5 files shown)


packages/reactive-element/src/decorators/query.ts
```diff
@@ -63,9 +63,11 @@ export function query(selector: string, cache?: boolean) {
               key as string
             ] = this.renderRoot?.querySelector(selector);
           }
-          return (this as unknown as {[key: string]: Element | null})[
-            key as string
-          ];
+          return (
+            (this as unknown as {[key: string]: Element | null})[
+              key as string
+            ] ?? null
```
justinfagnani, line 69: should we initialize the cache property to `null` so that we avoid later hidden class changes and then check for `null` instead of `undefined` above?

kevinpschaaf (approved):
LGTM

packages/reactive-element/src/decorators/query.ts
kevinpschaaf: There's no real place to do that right? You can't do it at decorator time since it's an own property on the instance. And there's not a good initializer spot to do it before the accessor is run the first time ... 🤷 .

sorvell: This wasn't quite correct. The intended behavior is to always cache the first result and that the result must be either null or the found element.

justinfagnani (approved):

aomarks:
This is a bit of a shame in terms of decorator -> vanilla JS documentation/transformer, since it requires adding ` ?? null` or ` ?? []` everyplace we transform `@query` and `@queryAll`, which doesn't read as nicely, and I don't think a person would usually think to write that otherwise. It's also a little less intuitive behaviorally since you'd expect it to match the types of the built-in `query` and `queryAll`.

Any chance we could keep this as a breaking change, or do we think the cost is too high?

PR #2066, opened 2021-08-20
[lit-html] Issue a DEV_MODE warning for setValue on a disconnected ChildPart

### LINKED ISSUE #2063: [lit-html] Improve ergonomics for malformed child parts
When a user manually modifies DOM under Lit's control, it's possible to eject a part's `startNode` from the DOM. If Lit subsequently re-renders this part, the render can fail because the `startNode` has no `parentNode` and therefore cannot insert content.

In Lit1, this would silently fail because there was a check for `startNode.parentNode` which had to pass before any part value could be set. 

In Lit2, this check was removed so there will be an exception now. We may not want to put the check back, but at the least we should generate a better error in dev mode.

sorvell:
Issues a DEV_MODE warning if a ChildPart cannot accept values because its marker nodes have been ejected from the part's `parentNode`.

Fixes #2063

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1041,6 +1041,11 @@ class ChildPart implements Disconnectable {
   }
 
   _$setValue(value: unknown, directiveParent: DirectiveParent = this): void {
+    if (DEV_MODE && this.parentNode === null) {
+      throw new Error(
+        `This \`ChildPart\` has no \`parentNode\` and therefore cannot accept the value of ${value}. This likely means the element containing the part was manipulated in an unsupported way outside of Lit's control such that the part's marker nodes were ejected from DOM. For example, setting the element's \`innerHTML\` or \`textContent\` can do this.`
+      );
+    }
     value = resolveDirective(this, value, directiveParent);
     if (isPrimitive(value)) {
       // Non-rendering child values. It's important that these do not render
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -1041,6 +1041,11 @@ class ChildPart implements Disconnectable {
   }
 
   _$setValue(value: unknown, directiveParent: DirectiveParent = this): void {
+    if (DEV_MODE && this.parentNode === null) {
+      throw new Error(
+        `This \`ChildPart\` has no \`parentNode\` and therefore cannot accept the value of ${value}. This likely means the element containing the part was manipulated in an unsupported way outside of Lit's control such that the part's marker nodes were ejected from DOM. For example, setting the element's \`innerHTML\` or \`textContent\` can do this.`
```
kevinpschaaf, line 1046: Might be a little clearer?
```suggestion
        `The value ${value} was rendered to a ChildPart whose marker nodes are no longer in the DOM. This likely means the element containing the part was manipulated in an unsupported way outside of Lit's control such that the part's marker nodes were ejected from DOM. For example, setting the element's \`innerHTML\` or \`textContent\` can do this.`
```

kevinpschaaf (approved):
Please `npm run changeset add`.

kevinpschaaf (changes requested):
This caused tests to fail, PTAL.

packages/lit-html/src/lit-html.ts
justinfagnani: I'm not sure about including the value... It could be a complex object like a TemplateResult. Since we can't get to the source position of the template, we could probably do best by printing out the template strings.

justinfagnani (approved):

kevinpschaaf (approved):

PR #2067, opened 2021-08-20
[ts-transformers] Preserve blank lines and fix constructor ordering

aomarks:
Adds two new transformers to the `@lit/ts-transformers` package: `preserveBlankLinesTransformer` and `constructorCleanupTransformer`.

The goal of these transformers is to address the loss of formatting that occurs in any standard TypeScript compile. We care about preserve formatting because we're using these transforms to generate idiomatic JavaScript from TypeScript that we will present in lit.dev documentation (https://github.com/lit/lit.dev/issues/332).

The problems these address are:

1. TypeScript never emits blank newlines. The `preserveBlankLinesTransformer` transformers addresses this by inserting `//__BLANK_LINE_PLACEHOLDER_G1JVXUEBNCL6YN5NFE13MD1PT3H9OIHB__` comments wherever a blank line appeared in the original source. These can comments can then be trivially replaced with newlines as a post-transform step.

2. TypeScript often moves the `constructor` to the top of the class, above `static` members, even when the constructor was present in the original source code. The `constructorCleanupTransformer` addresses this by checking the original source position of `constructors`, and moving them back to their original position. If the constructor is completely synthetic (meaning it doesn't have an original source position), we do a best effort by moving it below the final `static` member of the class, and inserting a blank-line placeholder comment.

As applied to the idiomatic decorator transform, that means we now generate code like this ...

```ts

  static get properties() {
    return {
      str: {},
      num: {type: Number, attribute: false},
    };
  }

  constructor() {
    super(...arguments);
    this.str = "foo";
    this.num = 42;
  }
}
```

... instead of this:

```ts
  constructor() {
    super(...arguments);
    this.str = "foo";
    this.num = 42;
  }
  static get properties() {
    return {
      str: {},
      num: {type: Number, attribute: false},
    };
  }
}
```

Also includes a bugfix related to import statement order.


packages/ts-transformers/src/tests/preserve-blank-lines-test.ts
```diff
@@ -0,0 +1,249 @@
...
+    import 'foo';
+    class Foo {
+      foo() {}
+      bar() {}
+    }
+  `;
+  checkTransform(input, expected);
+});
+
+test('blank line after imports', () => {
```
AndrewJakubowicz, line 80: Is it possible for there to be a file with only new lines, or begins with newlines before the code?

packages/ts-transformers/src/preserve-blank-lines.ts
```diff
@@ -0,0 +1,207 @@
...
+      }
+    }
+
+    if (!anyBlankLinesAdded) {
+      return;
+    }
+
+    // The TypeScript APIs around manipulating comments are a bit crude.
+    // Original source comments are represented differently to "synthetic"
+    // comments (ones created during a transform) .We don't have the ability to
```
justinfagnani, line 158: Bad spacing around period

justinfagnani (approved):
Wow, very nice! Learned a bit about the compiler reviewing this!

packages/ts-transformers/src/tests/preserve-blank-lines-test.ts
AndrewJakubowicz: Does the transforming code handle syntax errors, or is that not applicable?

Thank you! This code change is fantastic!

aomarks: > Is it possible for there to be a file with only new lines, or begins with newlines before the code?

Good questions! Investigated this, and found that comments at the beginning and end of the file are weird!

Comments at the beginning of the file are associated with the `SourceFile` *and* the first child node. Comments at the end of the file are associated with the `EndOfFileToken`. In both cases, it is not possible to change the default comment-emitting behavior (which drops blank lines) -- you can't use the `setTextRange` trick, and you also can't use `addSyntheticComment` (they have no effect). So unless we can find another trick, there isn't much we can do here. I'm not too worried, since dropping blank lines at the very beginning and end of the file doesn't seem like that big a deal. Added some `.skip` tests with a `TODO` to document this in case we want to revisit it later.

This did help me find and fix a bug that affected an important case, though, which is something like:

```ts
/** @license ... */

const foo = 0;
```

While there is a comment at the top of the file here, so we couldn't deal with any blank lines *above* it -- we can still insert a newline above the `const foo = 0;` to recover *that* blank line. We were sort of doing this before, but in a way that actually caused the leading comment to be emitted twice. Fixed now with a test.

> Does the transforming code handle syntax errors, or is that not applicable?

We shouldn't need to do anything special. If TypeScript encounters a syntax error, it will try its best to produce a valid AST, which is what we'll transform, and the user will see a syntax error.

PR #2070, opened 2021-08-20
Throw instead of rendering error on security issue

rictic:
This is easier for users to understand what's going on and where the problem is. Upstreaming cl/369366746

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1085,12 +1085,17 @@ class ChildPart implements Disconnectable {
             if (parentNodeName === 'STYLE') {
               message =
                 `Lit does not support binding inside style nodes. ` +
-                `This is a security risk, as style injection attacks can exfiltrate data and spoof UIs. ` +
-                `Consider instead using css\`...\` literals to compose styles, and accomplishing dynamicism by mutating the DOM rather than styles`;
+                `This is a security risk, as style injection attacks can ` +
+                `exfiltrate data and spoof UIs. ` +
+                `Consider instead using css\`...\` literals ` +
+                `to compose styles, and make do dynamic styling with ` +
+                `css custom properties, ::parts, <slot>s, ` +
+                `and by mutating the DOM rather than stylesheets.`;
             } else {
               message =
                 `Lit does not support binding inside script nodes. ` +
-                `This is a security risk, as it could allow arbitrary code execution.`;
+                `This is a security risk, as it could allow arbitrary ` +
+                `code execution.`;
             }
           }
           throw new Error(message);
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -1080,13 +1080,20 @@ class ChildPart implements Disconnectable {
...
+          let message = 'Forbidden';
+          if (DEV_MODE) {
+            if (parentNodeName === 'STYLE') {
+              message =
+                `Lit does not support binding inside style nodes. ` +
+                `This is a security risk, as style injection attacks can exfiltrate data and spoof UIs. ` +
+                `Consider instead using css\`...\` literals to compose styles, and accomplishing dynamicism by mutating the DOM rather than styles`;
+            } else {
+              message =
+                `Lit does not support binding inside script nodes. ` +
```
justinfagnani, line 1092: This is fine as an existing restriction, but I'd like to dig into this more with the ISE team.

First, in client-side rendering <script> tags in Lit templates will not execute (in SSR they can, but we could make sure they don't).

Then <script> tags can be used for other purposes than executable JS, mainly because they're raw text elements and can contain unescaped "<" characters. The Playground Elements use <script> tags for file content. I've seen syntax highlighters that do the same. I could see cases where you want those to be dynamic...

packages/lit-html/src/lit-html.ts
```diff
@@ -1080,13 +1080,20 @@ class ChildPart implements Disconnectable {
...
-            )
-          );
-          return;
+          let message = 'Forbidden';
+          if (DEV_MODE) {
+            if (parentNodeName === 'STYLE') {
+              message =
+                `Lit does not support binding inside style nodes. ` +
+                `This is a security risk, as style injection attacks can exfiltrate data and spoof UIs. ` +
+                `Consider instead using css\`...\` literals to compose styles, and accomplishing dynamicism by mutating the DOM rather than styles`;
```
justinfagnani, line 1089: Can you beak this line up to be < 80 cols?

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
rictic: Yeah, we could add some additional checks here, e.g. that there's a `type` field of a non-javascript type.

It is the case that these script tags don't typically run, but are we certain that there isn't a way to make them run? e.g. render into an unattached document fragment, then attach that? Something something inert template and clone into document?

packages/lit-html/src/lit-html.ts
rictic: Done!

packages/lit-html/src/lit-html.ts
justinfagnani: I'm not 100% sure if cloning drops the script-created bit, but I'd like to be! Seems like good tests to eventually add.

PR #2072, opened 2021-08-21
Fix noted TODOs

sorvell:
* Fixes doc link TODOs in LitElement and ReactiveElement
* Fixes styling tests to include static bindings in lit-html and LitElement
* Removes `@queryAssignedNodes` shim of `Element.matches` since this is now addressed via polyfills.

### DIFF
--- a/packages/reactive-element/src/decorators/query-assigned-nodes.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-nodes.ts
@@ -14,13 +14,6 @@
 import {ReactiveElement} from '../reactive-element.js';
 import {decorateProperty} from './base.js';
 
-// TODO(sorvell): Remove when https://github.com/webcomponents/polyfills/issues/397 is addressed.
-// x-browser support for matches
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const ElementProto = Element.prototype as any;
-const legacyMatches =
-  ElementProto.msMatchesSelector || ElementProto.webkitMatchesSelector;
-
 /**
  * A property decorator that converts a class property into a getter that
  * returns the `assignedNodes` of the given named `slot`. Note, the type of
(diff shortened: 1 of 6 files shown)


packages/reactive-element/src/decorators/query-assigned-nodes.ts
```diff
@@ -14,13 +14,6 @@
 import {ReactiveElement} from '../reactive-element.js';
 import {decorateProperty} from './base.js';
 
-// TODO(sorvell): Remove when https://github.com/webcomponents/polyfills/issues/397 is addressed.
-// x-browser support for matches
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const ElementProto = Element.prototype as any;
-const legacyMatches =
-  ElementProto.msMatchesSelector || ElementProto.webkitMatchesSelector;
```
kevinpschaaf, line 22:
```
tests:  ❌ @queryAssignedNodes > returns assignedNodes for slot that contains text nodes filtered by selector when Element.matches does not exist
tests:       TypeError: node.matches is not a function
tests:         at ../reactive-element/src/decorators/query-assigned-nodes.ts:60:32
tests:         at Array.filter (<anonymous>)
tests:         at HTMLElement.get [as footerAssignedItems] (../reactive-element/src/decorators/query-assigned-nodes.ts:57:24)
tests:         at o.<anonymous> (../reactive-element/src/test/decorators/queryAssignedNodes_test.ts:221:40)
```
😭

kevinpschaaf (commented):
Failing test on matches, not sure what's up since it seems like https://github.com/webcomponents/polyfills/pull/400 was merged long ago.

Also cc: @arthurevans @aomarks since this is the first time we're putting website URLs in code (references to docs from devmode warnings). It would be really nice to indirect through a URL shortener so we could update links if they change. Would it be easy to make a redirect config on lit.dev so the code could link to `lit.dev/msg/0001` or `lit.dev/msg/method-removed` and have that redirect to a configurable spot?

kevinpschaaf:
Also, changeset please

kevinpschaaf (approved):
I added https://github.com/lit/lit/issues/2078 to track whether we want to change deep lit.dev links to a flat/shortened redirectable scheme.

PR #2088, opened 2021-08-24
Update all packages to `typescript@^4.3.5`.

bicknellr:


bicknellr:
This is already in progress in https://github.com/lit/lit/pull/2060.

PR #2106, opened 2021-08-28
[ts-transformers] Rewrite static styles and properties class fields as getters

aomarks:
**Note to reviewer**: I recommend looking at each commit separately, I tried to break it up for readability.

Transforms:

```ts
  static styles = css\`p { color: red; }\`;
  static properties = {foo: {type: Number}};
}
```

Into:

```ts
  static get styles() {
    return css\`p { color: red; }\`;
  }
  static get properties() {
    return {foo: {type: Number}};
  }
}
```

Instead of:

```ts
}
MyElement.styles = css\`p { color: red; }\`;
MyElement.properties = {foo: {type: Number}};
```

This transform is useful because Lit is not compatible with standard class fields, so standard class field emit needs to be disabled (`useDefineForClassFields: false`). However, this means statics turn into `MyClass.foo = ...` assignments following the class declaration, which has poor readability compared to a getter.

### DIFF
--- a/packages/ts-transformers/src/constructor-cleanup.ts
+++ b/packages/ts-transformers/src/constructor-cleanup.ts
@@ -6,6 +6,7 @@
 
 import * as ts from 'typescript';
 import {BLANK_LINE_PLACEHOLDER_COMMENT} from './preserve-blank-lines';
+import {isStatic} from './util';
 
 /**
  * TypeScript transformer which improves the readability of the default
@@ -88,11 +89,7 @@ const cleanupClassConstructor = (
     // common style.
     newCtorIdx = 0;
     for (let i = class_.members.length - 1; i >= 0; i--) {
-      const isStatic =
-        class_.members[i].modifiers?.find(
-          (modifier) => modifier.kind === ts.SyntaxKind.StaticKeyword
-        ) !== undefined;
-      if (isStatic) {
+      if (isStatic(class_.members[i])) {
         newCtorIdx = i;
         break;
       }
--- a/packages/ts-transformers/src/idiomatic.ts
+++ b/packages/ts-transformers/src/idiomatic.ts
@@ -16,6 +16,7 @@ import {QueryAsyncVisitor} from './idiomatic/query-async.js';
 import {QueryAssignedNodesVisitor} from './idiomatic/query-assigned-nodes.js';
 import {EventOptionsVisitor} from './idiomatic/event-options.js';
 import {StaticStylesVisitor} from './idiomatic/static-styles.js';
+import {StaticPropertiesVisitor} from './idiomatic/static-properties.js';
 
 /**
  * TypeScript transformer which transforms all Lit decorators to their idiomatic
@@ -71,6 +72,7 @@ export default function idiomaticLitDecoratorTransformer(
       new QueryAssignedNodesVisitor(context),
       new EventOptionsVisitor(context, program),
       new StaticStylesVisitor(context, program),
+      new StaticPropertiesVisitor(context, program),
     ]);
     return (file) => {
       return ts.visitNode(file, transformer.visitFile);
--- a/packages/ts-transformers/src/idiomatic.ts
+++ b/packages/ts-transformers/src/idiomatic.ts
@@ -15,6 +15,7 @@ import {QueryAllVisitor} from './idiomatic/query-all.js';
 import {QueryAsyncVisitor} from './idiomatic/query-async.js';
 import {QueryAssignedNodesVisitor} from './idiomatic/query-assigned-nodes.js';
 import {EventOptionsVisitor} from './idiomatic/event-options.js';
+import {StaticStylesVisitor} from './idiomatic/static-styles.js';
 
 /**
  * TypeScript transformer which transforms all Lit decorators to their idiomatic
@@ -69,6 +70,7 @@ export default function idiomaticLitDecoratorTransformer(
       new QueryAsyncVisitor(context),
       new QueryAssignedNodesVisitor(context),
       new EventOptionsVisitor(context, program),
+      new StaticStylesVisitor(context, program),
     ]);
     return (file) => {
       return ts.visitNode(file, transformer.visitFile);
--- a/packages/ts-transformers/src/idiomatic/static-properties.ts
+++ b/packages/ts-transformers/src/idiomatic/static-properties.ts
@@ -1,0 +1,70 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as ts from 'typescript';
+import {isStatic, extendsReactiveElement} from '../util.js';
+
+import type {LitFileContext} from '../lit-file-context.js';
+import type {GenericVisitor} from '../visitor.js';
+
+/**
+ * Transform:
+ *
+ *   static properties = {foo: {type: Number}};
+ *
+ * Into:
+ *
+ *   static get properties() {
+ *     return {foo: {type: Number}};
+ *   }
+ *
+ * This transform is useful because Lit is not compatible with standard class
+ * fields, so standard class field emit needs to be disabled
+ * (useDefineForClassFields: false). However, this means statics turn into
+ * `MyClass.foo = ...` assignments following the class declaration, which has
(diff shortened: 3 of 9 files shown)


justinfagnani:
All current browsers support instance and static class fields... I wonder if we should do this one by default or not.

aomarks:
> And would a different approach be to transform decorated fields await to the static class properties block and constructor initializers? That would preserve all other fields.

Sorry can you clarify? Do you mean turn on standard class field emit, but then have a post-built-in-transform that converts instance field initializers to constructor assignments?

bathos:
> Lit is not compatible with standard class fields

Probably missing specific context here, but this surprised me since we’ve been using lit with JS and static and instance fields (i.e., with defineProperty semantics, not set) seemingly without problems for two years. When does it not work exactly? Is it an issue that started/starts in lit 3? (We’re still on 2.5.1, so this could be important for us when we next upgrade...)

aomarks:
> > Lit is not compatible with standard class fields
> 
> Probably missing specific context here, but this surprised me since we’ve been using lit with JS and static and instance fields (i.e., with defineProperty semantics, not set) seemingly without problems for two years. When does it not work exactly? Is it an issue that started/starts in lit 3? (We’re still on 2.5.1, so this could be important for us when we next upgrade...)

Hi @bathos. If you use standard class fields with any version of lit-element, those properties will not be reactive, meaning `requestUpdate()` won't be automatically called when they change.

Here's a playground example for [Lit 2 (lit-element 3)](https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4LmpzIiwiY29udGVudCI6ImltcG9ydCB7aHRtbCwgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcblxuZXhwb3J0IGNsYXNzIFhIZWxsbyBleHRlbmRzIExpdEVsZW1lbnQge1xuICBzdGF0aWMgcHJvcGVydGllcyA9IHtcbiAgICBuYW1lOiB7fVxuICB9O1xuXG4gIC8vIGRvZXMgbm90IHdvcmtcbiAgbmFtZSA9ICdkZWZhdWx0JztcblxuICAvLyB3b3Jrc1xuICAvL2NvbnN0cnVjdG9yKCkge1xuICAvLyAgc3VwZXIoKTtcbiAgLy8gIHRoaXMubmFtZSA9ICdkZWZhdWx0JztcbiAgLy99XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYEhlbGxvICR7dGhpcy5uYW1lfWA7XG4gIH1cbn1cbmN1c3RvbUVsZW1lbnRzLmRlZmluZSgneC1oZWxsbycsIFhIZWxsbyk7XG5cbnNldFRpbWVvdXQoKCkgPT4ge1xuICBkb2N1bWVudC5xdWVyeVNlbGVjdG9yKCd4LWhlbGxvJykubmFtZSA9ICd1cGRhdGVkJztcbn0sIDUwMCk7In0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL2luZGV4LmpzXCI-PC9zY3JpcHQ-XG48eC1oZWxsbz48L3gtaGVsbG8-In0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjIuMC4wLXJjLjJcIlxuICB9XG59In1d), and for [lit-element 2](https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4LmpzIiwiY29udGVudCI6ImltcG9ydCB7aHRtbCwgTGl0RWxlbWVudH0gZnJvbSAnbGl0LWVsZW1lbnQnO1xuXG5leHBvcnQgY2xhc3MgWEhlbGxvIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHN0YXRpYyBwcm9wZXJ0aWVzID0ge1xuICAgIG5hbWU6IHt9XG4gIH07XG5cbiAgLy8gZG9lcyBub3Qgd29ya1xuICBuYW1lID0gJ2RlZmF1bHQnO1xuXG4gIC8vIHdvcmtzXG4gIC8vY29uc3RydWN0b3IoKSB7XG4gIC8vICBzdXBlcigpO1xuICAvLyAgdGhpcy5uYW1lID0gJ2RlZmF1bHQnO1xuICAvL31cblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgSGVsbG8gJHt0aGlzLm5hbWV9YDtcbiAgfVxufVxuY3VzdG9tRWxlbWVudHMuZGVmaW5lKCd4LWhlbGxvJywgWEhlbGxvKTtcblxuc2V0VGltZW91dCgoKSA9PiB7XG4gIGRvY3VtZW50LnF1ZXJ5U2VsZWN0b3IoJ3gtaGVsbG8nKS5uYW1lID0gJ3VwZGF0ZWQnO1xufSwgNTAwKTsifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vaW5kZXguanNcIj48L3NjcmlwdD5cbjx4LWhlbGxvPjwveC1oZWxsbz4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl4yLjUuMVwiXG4gIH1cbn0ifV0).

This issue is documented [here](https://lit.dev/docs/components/properties/#declaring-properties-in-a-static-properties-field) and [here](https://lit.dev/docs/components/decorators/#avoiding-issues-with-class-fields).

Since you say you've been using them for 2 years with no problem, I wonder if you are actually transpiling class fields to the old semantics using Babel or TypeScript?

aomarks:
> Hi @bathos. If you use standard class fields with any version of lit-element, those properties will not be reactive, meaning `requestUpdate()` won't be automatically called when they change.

Also I should clarify that this is only an issue for *instance* fields, i.e. the reactive data properties of the element. The *static* fields: `styles` and `properties` can safely use standard static class fields.

bathos:
The examples here are for static fields, though.

(I shouldn't have mentioned instance fields. I understand why those wouldn't work, though we _do_ use them in source as well; decorator transforms lead to them being realized as accessors on the prototype instead, despite the syntax.)

aomarks:
> The examples here are for static fields, though.
> 
> (I shouldn't have mentioned instance fields. I understand why those wouldn't work, though we _do_ use them in source as well; decorator transforms lead to them being realized as accessors on the prototype instead, despite the syntax.)

Ah yeah, understood. This PR is a little bit of a roundabout:

- Because standard *instance* fields are not supported, we are setting the `useDefineForClassFields` flag to `false` so that TypeScript doesn't emit new-semantics class fields.
- However, that flag controls both instance *and* static fields. There is no way to ask TypeScript to emit standard static fields without also doing instance fields.
- That means if you have written a static class field, TypeScript will transform it to a `MyElement.styles ...` assignment after the class. This doesn't look very readable (which is the main goal of this set of transforms), so we convert it to a getter instead.

I think Justin's idea is pretty good, though --instead of all this, have an after-transform that turns reactive class fields back into constructor assignments. That way we could run with `useDefineForClassFields` enabled, giving us standard static class fields, but with instance class fields repaired to workaround this Lit incompatibility.

aomarks:
> I mean if you transform the decorated fields to _not_fields_ then you can keep `useDefineForClassFields` as `true`. So for properties, in addition to adding to the static properties object, replace the field declaration with just a constructor initializer.

Oh right, we can even do this as a before-transform. I was imagining we had to do it as an after transform for some reason. Sounds good, I like this much better, I will close this PR and send another one that does this.

aomarks:
> Oh right, we can even do this as a before-transform. I was imagining we had to do it as an after-transform for some reason. Sounds good, I like this much better, I will close this PR and send another one that does that.

Ready for review in https://github.com/lit/lit/pull/2116

PR #2109, opened 2021-08-29
[ts-transformers] Clean up unnecessary `super(...arguments)` calls

aomarks:
When synthesizing a subclass constructor, TypeScript always includes a `super(...arguments)` call.

However, it is often unnecessary to include the `...arguments` argument.

This PR updates the `constructor-cleanup` transformer to remove the `...arguments` argument when none of the ancestor classes have any parameters in their constructor.

Note this is not 100% safe, since an ancestor constructor could directly use `arguments` in its body (which we may not be able to detect because we don't have access to bodies via `.d.ts` files) but that should be fairly rare. For the main use case we're targeting (generating lit.dev samples), it seems like a reasonable readability/correctness tradeoff.

### DIFF
--- a/packages/ts-transformers/src/constructor-cleanup.ts
+++ b/packages/ts-transformers/src/constructor-cleanup.ts
@@ -5,7 +5,8 @@
  */
 
 import * as ts from 'typescript';
-import {BLANK_LINE_PLACEHOLDER_COMMENT} from './preserve-blank-lines';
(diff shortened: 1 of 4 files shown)


aomarks:
Closing for now, is going to need a re-base due to some other changes. Will re-open when ready.

justinfagnani:
Seems like a reasonable transform - we shouldn't support using `arguments` when `...arguments` works and is included in the signature.

justinfagnani (approved):

PR #2112, opened 2021-08-30
[ReactiveElement] Throws rather than warns in dev mode for class fields that shadow reactive properties

### LINKED ISSUE #2104: [ReactiveElement] Make shadowed properties throw instead of warn in DEV_MODE
The warning is likely not enough since your element is really in a broken state if this happens. Instead we should throw.

sorvell:
Fixes #2104. The element is in a broken state in this case, so rather than warn, we throw.


kevinpschaaf (approved):
LGTM

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1092,16 +1092,13 @@ export abstract class ReactiveElement
           }
         );
         if (shadowedProperties.length) {
-          issueWarning(
+          throw new Error(
```
justinfagnani, line 1095: @aomarks what should we do for errors now that `issueWarning` takes a warning code?

justinfagnani (approved):
Code looks good, just a question about using the new message URLs with errors.

packages/reactive-element/src/reactive-element.ts
kevinpschaaf: The error can just include the full URL; don't need to go through issueWarning, right?

aomarks: Yeah that seems fine, just put the code URL directly.

kevinpschaaf: Done

PR #2113, opened 2021-08-31
[all] Upgrade all package locks, includes TypeScript 4.4.2

aomarks:
- Upgrade all package locks. Includes TypeScript 4.3.5 -> 4.4.2.
- Typing fixes for TypeScript 4.4.2 (mostly https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables)
- Adds `devDependency` on new `@types/web-ie11` package to address microsoft/TypeScript-DOM-lib-generator#1068
- Upgrade `xmldom` to new `@xmldom/xmldom` due to xmldom/xmldom#271
- Pin the localize transform example to TypeScript ~4.3.5 due to rollup/plugins#983
- Add missing `@types/node` dependency to `internal-scripts` (not sure why this wasn't an error before -- maybe the Node types were coming via a transitive dependency which was removed by an upgrade).
- Minor `@lit/localize-tools` changes: don't run `build` as part of `test` script, use `@lit/ssr` bare module in test instead of relative path, trivial config file schema change from dependency upgrade.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -438,16 +438,12 @@ if (ENABLE_EXTRA_SECURITY_HOOKS) {
   }
 }
 
-const walker = (
-  d.createTreeWalker as (
-    root: Node,
-    whatToShow: number,
-    filter: NodeFilter | null,
-    // All 4 parameters are required for IE11, but TypeScript 4.4 removed the
-    // entityReferenceExpansion parameter. This is the previous signature.
-    entityReferenceExpansion?: boolean
-  ) => TreeWalker
-)(d, 129 /* NodeFilter.SHOW_{ELEMENT|COMMENT} */, null, false);
+const walker = d.createTreeWalker(
+  d,
+  129 /* NodeFilter.SHOW_{ELEMENT|COMMENT} */,
+  null,
+  false
+);
 
 let sanitizerFactoryInternal: SanitizerFactory = noopSanitizer;
 
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -438,12 +438,16 @@ if (ENABLE_EXTRA_SECURITY_HOOKS) {
   }
 }
 
-const walker = d.createTreeWalker(
-  d,
-  129 /* NodeFilter.SHOW_{ELEMENT|COMMENT} */,
-  null,
-  false
-);
+const walker = (
+  d.createTreeWalker as (
+    root: Node,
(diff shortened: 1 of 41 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -438,12 +438,16 @@ if (ENABLE_EXTRA_SECURITY_HOOKS) {
...
+const walker = (
+  d.createTreeWalker as (
+    root: Node,
+    whatToShow: number,
+    filter: NodeFilter | null,
+    // All 4 parameters are required for IE11, but TypeScript 4.4 removed the
+    // entityReferenceExpansion parameter. This is the previous signature.
+    entityReferenceExpansion?: boolean
+  ) => TreeWalker
+)(d, 129 /* NodeFilter.SHOW_{ELEMENT|COMMENT} */, null, false);
```
rictic, line 450: This seems awkward in comparison to just depending on the IE11 types.

Another nice part about depending on IE11 types is that if/when we drop IE11 support, we can drop those types and the error will let us know we can drop this hack

packages/localize/examples/transform/package.json
```diff
@@ -20,6 +20,6 @@
     "rollup": "^2.41.1",
     "rollup-plugin-summary": "^1.3.0",
     "rollup-plugin-terser": "^7.0.2",
-    "typescript": "^4.3.5"
+    "typescript": "~4.3.5"
```
rictic, line 23: Is this what we want to do here long term?

rictic (approved):

packages/localize/examples/transform/package.json
aomarks: Nope, but the typescript rollup plugin is totally broken with TypeScript 4.4, see https://github.com/rollup/plugins/issues/983

rictic: I see. Too bad there's no comments allowed in json files :/

packages/lit-html/src/lit-html.ts
aomarks: True, done.

We will need to add this dependency, or write a patch, when we next import into google3, though.

aomarks: PTAL @rictic

packages/localize/examples/transform/package.json
justinfagnani: We do want to keep the typescript constraint as a `~x.y.z` though, so this is great. We should have had that before, but relied on the package lock...

PR #2114, opened 2021-08-31
[lit-html]  In dev mode, indicate if a part is placed inside an element which cannot contain parts

### LINKED ISSUE #1717: [lit-element] [lit-next] Issue with textarea exposing a property
# Description
When I expose a property using `<textarea>` content,  the behavior is weird

# Lit versions
`lit: "2.0.0-pre.2"`

# Steps to Reproduce
- create a lit-element with a property (lets call it `name`)
- show this property using textarea
  ```
  <textarea> ${this.name} </textarea>
  ```
- now use this property in any other elements attribute 
  ```
  <div data-name="${this.name}"> </div>
  ```

## Live Reproduction Link
https://stackblitz.com/edit/lit-html-next-s3zjyy

## Expected Results
should render and not throw a console error

## Actual Results
do not render and throw a console error

## Browsers Affected
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
> Not tested in other browser, but seems it will be the same case with others too

> In case it matters, it works with `lit-element - 2.4.0`

sorvell:
Fixes #1717 and #1737 .

Parts are not supported inside the `template` or `textarea` tags. In dev mode, we indicate if parts are placed here so the developer can remove them.


packages/lit-html/src/lit-html.ts
```diff
@@ -547,6 +578,15 @@ const getTemplateHtml = (
         }
       } else if (regex === tagEndRegex) {
         if (match[ENTIRE_MATCH] === '>') {
+          // If we're entering a tag which should not contain internal parts,
+          // setup the regex for leaving the tag. Note, the existence of the
+          // regex indicates that the warning should be issued.
+          if (DEV_MODE && warnPartsInsideTag!) {
+            warnPartsInsideEndRegex = new RegExp(
+              `</${warnPartsInsideTag}`,
+              'g'
```
kevinpschaaf, line 587: Why the 'g'?

packages/lit-html/src/lit-html.ts
```diff
@@ -512,10 +525,18 @@ const getTemplateHtml = (
...
-
     // The conditions in this loop handle the current parse state, and the
     // assignments to the `regex` variable are the state transitions.
     while (lastIndex < s.length) {
+      // If we are moving outside of a tag that should not contains
+      // internal parts, then reset that warning state.
+      if (
+        DEV_MODE &&
+        warnPartsInsideEndRegex! &&
+        s.match(warnPartsInsideEndRegex!)
```
kevinpschaaf, line 536: Because templates can nest, technically this would break since the tracking isn't balanced, right?

```js
html`<template><template></template>${foo}</template>`
```

kevinpschaaf: Seems like we'd need a stack, which is starting to get pretty fancy... Maybe timebox and just leave a TODO/issue if there's not an easy fix? I'm not sure we should spend a lot of time on this since it's just a DEV_MODE issue for a pretty corner case.

packages/lit-html/src/lit-html.ts
```diff
@@ -236,6 +235,15 @@ const doubleQuoteAttrEndRegex = /"/g;
  */
 const rawTextElement = /^(?:script|style|textarea)$/i;
 
+/**
+ * List of elements that do not support internal parts. This is needed for
+ * issuing dev mode warnings.
+ */
+let unsupportedInternalPartsElement: RegExp;
+if (DEV_MODE) {
+  unsupportedInternalPartsElement = /^(?:textarea|template)$/i;
```
kevinpschaaf, line 244: Use literal regex in the DEV_MODE block?

kevinpschaaf (approved):
LGTM

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -2015,6 +2015,92 @@ suite('lit-html', () => {
...
+        assert.throws(() => {
+          render(
+            html`<template>
+            <div><div><div><div>${'test'}</div></div></div></div>
+            </template>`,
+            container
+          );
+        });
+
+        // attr part deep inside
```
justinfagnani, line 2040: Add test for element expression

justinfagnani (changes requested):

justinfagnani (approved):

packages/lit-html/src/test/lit-html_test.ts
sorvell: Done.

PR #2117, opened 2021-09-01
[starter-kits] Update starter kits to use open-wc CEM analyzer

kevinpschaaf:
Updates the API docs templating to the new custom-elements.json manifest format. Adds basic method printing, which was a TODO in the previous code. Output for the included sample element (source below) looks like the image below:

```ts
/**
 * @license
 * Copyright 2019 Google LLC
 */

import {LitElement, html, css} from 'lit';
import {customElement, property} from 'lit/decorators.js';

/**
 * An example element.
 *
 * @fires count-changed - Indicates when the count changes
 * @slot - This element has a slot
 * @csspart button - The button
 */
@customElement('my-element')
export class MyElement extends LitElement {
  static override styles = css`
    :host {
      display: block;
      border: solid 1px gray;
      padding: 16px;
      max-width: 800px;
    }
  `;

  /**
   * The name to say "Hello" to.
   */
  @property()
  name = 'World';

  /**
   * The number of times the button has been clicked.
   */
  @property({type: Number})
  count = 0;

  override render() {
    return html`
      <h1>${this.sayHello(this.name)}!</h1>
      <button @click=${this._onClick} part="button">
        Click Count: ${this.count}
      </button>
      <slot></slot>
    `;
  }

  private _onClick() {
    this.count++;
    this.dispatchEvent(new CustomEvent('count-changed'));
  }

  /**
   * Formats a greeting
   * @param name The name to say "Hello" to
   * @returns A greeting directed at `name`
   */
  sayHello(name: string): string {
    return `Hello, ${name}`;
  }
}

declare global {
  interface HTMLElementTagNameMap {
    'my-element': MyElement;
  }
}
```

image


packages/lit-starter-js/docs-src/api.11ty.cjs
```diff
@@ -2,89 +2,131 @@
...
-  render(data) {
-    const customElements = data.api['11tydata'].customElements;
-    const tags = customElements.tags;
-    return `
+   render(data) {
+     const manifest = data.api['11tydata'].customElements;
+     const elements = manifest.modules.reduce(
+       (els, module) =>
+         els.concat(
+           module.declarations?.filter((dec) => dec.customElement) ?? []
```
justinfagnani: So the most correct thing to do here is to get the exports, filter for custom element defines, then resolve those to their exports... but in the starter kit we know we're exporting all the declarations, so it'll have the same effect. We might just want to do the more proper thing later to account for projects that grow from this.

justinfagnani (approved):

packages/lit-starter-js/docs-src/api.11ty.cjs
kevinpschaaf: Yeah let's tackle that later.

Issue #2118, opened 2021-09-01
[ssr] installWindowOnGlobal shim globalThis.global does not inherit any of the existing builtins like Promise


fetherston:
# Description

The DOM shim for @lit-labs/ssr does not inherit the builtin Node APIs from the environment onto the global object. Some third party libraries like koa [expect them to be there](https://github.com/stream-utils/raw-body/blob/0e1291f1d6cbc9ee8e16f893d091e417841c95e5/index.js#L89).

# Steps to Reproduce

1. Write this code

```ts
import { installWindowOnGlobal } from '@lit-labs/ssr/lib/dom-shim';
installWindowOnGlobal();
if (!global.Promise) {
    throw new Error('global.Promise is missing');
}
```

2. See this output...
It'll throw because global.Promise is undefined.

## Live Reproduction Link

N/A

## Expected Results

global.Promise should be the native Node Promise API

## Actual Results

No error should be thrown.

## Browsers Affected

N/A

PR #2120, opened 2021-09-01
[ssr] Don't assign DOM shim window.global to window

aomarks:
This means that `globalThis.global` will retain its Node built-ins, and `globalThis.global === globalThis` will hold, whereas before it would lose anything we didn't explicitly set on `window`, so something like `global.Promise` would be `undefined`.

This `window.global` assignment seems to have been done for compatibility with `node-fetch`, but I tried a `fetch` from the `demo/global/module.js` module, and the fetch seemed to work fine. Not sure if there is another invocation path I missed, but it seems ok?

Fixes https://github.com/lit/lit/issues/2118

cc @fetherston

### DIFF
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -141,14 +141,12 @@ export const getWindow = ({
 
     // Set below
     window: undefined as unknown,
-    global: undefined as unknown,
 
     // User-provided globals, like `require`
     ...props,
   };
 
   window.window = window;
-  window.global = window; // Required for node-fetch
 
   if (includeJSBuiltIns) {
     Object.assign(window, {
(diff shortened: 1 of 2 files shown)


packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -141,14 +141,12 @@ export const getWindow = ({
...
     // Set below
     window: undefined as unknown,
-    global: undefined as unknown,
 
     // User-provided globals, like `require`
     ...props,
   };
 
   window.window = window;
-  window.global = window; // Required for node-fetch
```
justinfagnani, line 151: What about node-fetch here? Should we do something like this?
```ts
Object.assign(window.global, window):
```

aomarks: See PR description -- unless you know a reason it actually fails?

If we do need it, I think we can do `window.global = globalThis`

justinfagnani (approved):

aomarks:
This will go out with the next batch of Lit RC releases, probably a little later this week.

PR #2123, opened 2021-09-01
[react] Add React.HTMLAttributes to component props

### LINKED ISSUE #2026: React wrapped components don't get HTML event types
# Description

After wrapping custom elements with `@lit-labs/react`, the resulting React components do not have types for HTML events. This causes TypeScript to throw an error, even though the syntax is correct and functions properly.

image

# Steps to Reproduce

1. Create any Lit-based custom element
2. Use `@lit-labs/react` to create a React-wrapper component
3. Import the React component in a `create-react-app --template typescript` project
4. Add `onClick`, `onMouseDown`, or any global HTMLElement event
5. Notice the TypeScript error

## Live Reproduction Link

As this is potentially a bug in `@lit-labs/react`, the repro requires custom elements that are wrapped for React and then consumed in a React app. The following is a repo created using the following commands.

Repo: https://github.com/claviska/react-wrapper-types-repro

It's worth noting that I used `@shoelace-style/react` for the wrapped components, but I'm seeing the same behavior in another library that also uses `@lit-labs/react`.

Here's the command to create the repo from scratch:

```
npx create-react-app --template typescript react-wrapper-types-repro
npm i @shoelace-style/react
```

The only modifications are in `App.tsx`:

- Added Shoelace - https://github.com/claviska/react-wrapper-types-repro/blob/master/src/App.tsx#L5-L6
- Used a Shoelace button - https://github.com/claviska/react-wrapper-types-repro/blob/master/src/App.tsx#L16

The error appears on line 16. Note that adding `@ts-ignore` will suppress the error and `onClick` works as expected.

## Expected Results

HTML events such as `onClick`, `onMouseDown`, etc. will be allowed by TypeScript.

## Actual Results

Line 16, where `onClick` is used, results in a TypeScript error.

CleanShot 2021-07-27 at 08 38 22@2x

Alt text:

```
Type '{ children: string; onClick: () => void; type: "primary"; }' is not assignable to type 'IntrinsicAttributes & Partial<Pick<SlButton, "button" | "slot" | "style" | "title" | "animate" | "circle" | "accessKey" | "accessKeyLabel" | "autocapitalize" | ... 263 more ... | "handleClick">> & { ...; } & { ...; } & RefAttributes<...>'.
  Property 'onClick' does not exist on type 'IntrinsicAttributes & Partial<Pick<SlButton, "button" | "slot" | "style" | "title" | "animate" | "circle" | "accessKey" | "accessKeyLabel" | "autocapitalize" | ... 263 more ... | "handleClick">> & { ...; } & { ...; } & RefAttributes<...>'. Did you mean 'onclick'?ts(2322)

(JSX attribute) onClick: () => void
```

## Browsers Affected

N/A

justinfagnani:
Fixes #2026

The React.HTMLAttributes type contains the built-in properties and event handlers and such, including the `onFoo` synthetic event handlers.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -178,7 +178,7 @@ export const createComponent = <I extends HTMLElement, E>(
         setProperty(
           this._element,
           prop,
-          this.props[prop as Exclude<keyof ComponentProps, symbol>],
+          this.props[prop as keyof ComponentProps],
           oldProps ? oldProps[prop as keyof ComponentProps] : undefined,
           events
         );
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -126,7 +126,11 @@ export const createComponent = <I extends HTMLElement, E>(
   // TODO: we might need to omit more properties from HTMLElement than just
   // 'children', but 'children' is special to JSX, so we must at least do that.
   type UserProps = React.PropsWithChildren<
-    React.PropsWithRef<Partial<Omit<I, 'children'>> & Events<E>>
+    React.PropsWithRef<
+      Partial<Omit<I, 'children'>> &
+        Events<E> &
+        React.HTMLAttributes<HTMLElement>
+    >
   >;
 
   // Props used by this component wrapper. This is the UserProps and the
@@ -174,7 +178,7 @@ export const createComponent = <I extends HTMLElement, E>(
         setProperty(
           this._element,
           prop,
-          this.props[prop as keyof ComponentProps],
+          this.props[prop as Exclude<keyof ComponentProps, symbol>],
           oldProps ? oldProps[prop as keyof ComponentProps] : undefined,
           events
         );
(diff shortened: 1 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -174,7 +178,7 @@ export const createComponent = <I extends HTMLElement, E>(
         setProperty(
           this._element,
           prop,
-          this.props[prop as keyof ComponentProps],
+          this.props[prop as Exclude<keyof ComponentProps, symbol>],
```
kevinpschaaf, line 181: What is this doing?

justinfagnani: I was getting an error that symbol couldn't be used in an index type. I scanned, but couldn't figure which part of React.HTMLAttributes<HTMLElement> was including a symbol key, so I just... excluded it

justinfagnani: I removed it and don't see the error now 🤷

PR #2124, opened 2021-09-01
[ts-transformers] Emit static properties as class fields

aomarks:
1. Updates the transformers so that we emit `static properties = { ... }` instead of `static get properties() { return { ... }; }`.

2. If there is an existing `static properties` field *or* getter, then we will augment whatever is already there with new properties, without changing the syntax.

3. Found and fixed a new weird bug that only cropped up after this change for some reason, relating to the difference between `class.decorators = []` and `class.decorators = undefined`. When `decorators` is an empty array, there is sometimes some no-op decorator transform cruft left around by the transform, like this:
   ```ts
   MyElement = __decorate([], MyElement)
   ```
   I believe this is due to some conditionals in the built-in transform like `if (class.decorators)` as opposed to `if (class.decorators && class.decorators.length > 0)`.

### DIFF
--- a/packages/ts-transformers/src/lit-transformer.ts
+++ b/packages/ts-transformers/src/lit-transformer.ts
@@ -367,11 +367,9 @@ export class LitTransformer {
   ) {
     const f = this._context.factory;
     const properties = [
-      ...(existingProperties?.properties
-        ? existingProperties.properties.map((prop) =>
-            cloneNode(prop, {factory: this._context.factory})
-          )
-        : []),
+      ...(existingProperties?.properties.map((prop) =>
+        cloneNode(prop, {factory: this._context.factory})
+      ) ?? []),
       ...newProperties.map(({name, options}) =>
         f.createPropertyAssignment(
           f.createIdentifier(name),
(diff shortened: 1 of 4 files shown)


packages/ts-transformers/src/lit-transformer.ts
```diff
@@ -345,14 +356,17 @@ export class LitTransformer {
...
+      name: string;
+      options?: ts.ObjectLiteralExpression;
+    }>
   ) {
     const f = this._context.factory;
     const properties = [
-      ...(existingProperties
-        ? existingProperties.map((prop) =>
+      ...(existingProperties?.properties
+        ? existingProperties.properties.map((prop) =>
```
justinfagnani, line 371: I don't think you need the ternary:
```ts
existingProperties?.properties.map(...) ?? []
```

justinfagnani (approved):

packages/ts-transformers/src/lit-transformer.ts
aomarks: Ty done

PR #2128, opened 2021-09-02
[lit-html] Add test for sync AsyncDirective.setValue() call

### LINKED ISSUE #1541: [lit-next] Test for synchronous call of DisconnectableDirective#setValue in #update
Directives may call setValue synchronously in their update method in order to render and then postprocess output.

justinfagnani:
Fixes #1541 

This is just some assurance that `setValue()` works synchronously as well, since we don't want to add code to prevent it .

### DIFF
--- a/packages/lit-html/src/test/lit-html_test.ts
+++ b/packages/lit-html/src/test/lit-html_test.ts
@@ -2197,8 +2197,10 @@ suite('lit-html', () => {
 
     test('async directive can call setValue synchronously', () => {
       assertRender(
-        html`<div foo=${syncAsyncDirective('test')}></div>`,
-        '<div foo="test"></div>'
+        html`<div foo=${syncAsyncDirective('test')}>${syncAsyncDirective(
+          'test'
+        )}</div>`,
+        '<div foo="test">test</div>'
       );
     });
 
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -2185,6 +2186,22 @@ suite('lit-html', () => {
...
+        render(x: string) {
+          this.setValue(x);
+          return noChange;
+        }
+      }
+    );
+
+    test('async directive can call setValue synchronously', () => {
+      assertRender(
+        html`<div foo=${syncAsyncDirective('test')}></div>`,
```
kevinpschaaf, line 2200: Can you add the directive in ChildPart position also? I expect it's going to error in SSR.

Just looked at the code, and it's safe for attribute parts because we [no-op AttributePart._commitValue](https://github.com/lit/lit/blob/main/packages/lit-html/src/private-ssr-support.ts#L67) before running directives, but ChildPart doesn't have a convenient way to do that; we just call `resolveDirective` assuming it will just return a value and won't ever actually touch the dummy part, but then `setValue` in render actually does.

If that does error like I expect, then one option is to just say `setValue` can only be called synchronously from `update` not render, which would solve it. Alternatively, we could make a similar private-ssr-support helper to no-op `_$setValue` on ChildPart and return any committed value when resolving directives. This is probably the more robust bet.

kevinpschaaf (changes requested):

kevinpschaaf (approved):
LGTM, will file issue on sync setValue in SSR

kevinpschaaf:
Filed in https://github.com/lit/lit/issues/2143

PR #2132, opened 2021-09-03
Prepare next prerelease

justinfagnani:


sorvell (approved):
Some of the changelog messages include info about other packages, but this is a pretty minor thing so nbd imo.

PR #2133, opened 2021-09-03
[ts-transformers] Misc housekeeping before publish

aomarks:
Getting this package ready to be published.

1. New README
2. New file layout, with internal stuff under `internal/`
3. Output modules to root of package. Adds `index.js`. Replaces default exports with named ones. New usage is:
    ```ts
    import {
      idiomaticDecoratorsTransformer,
      preserveBlankLinesTransformer,
      constructorCleanupTransformer,
    } from '@lit/ts-transformers';
    ```
4. Stricter TypeScript settings
5. Don't publish `src/`. Include `.ts` sources directly in `.js.map` files
6. Fix bug with `queryAsync` transform (oops, async getters don't exist)
7. Fix more bugs with decorator cruft (a method with `decorators = []` instead of `decorators = undefined` can cause problems).
7. Misc small style things


packages/ts-transformers/tsconfig.json
```diff
@@ -4,17 +4,22 @@
...
     "moduleResolution": "node",
     "lib": ["es2020", "DOM"],
+    "forceConsistentCasingInFileNames": true,
+
     "strict": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitOverride": true,
+    "noImplicitReturns": true,
+    "noPropertyAccessFromIndexSignature": true,
+    "noUncheckedIndexedAccess": true,
```
justinfagnani, line 16: This is one of the strict options I'm unsure of. It seems like it mostly resulted in adding non-null assertions. Did you find the other cases it caught useful - were they actual bugs?

.changeset/swift-dryers-invent.md
```diff
@@ -0,0 +1,2 @@
+---
```
justinfagnani, line 1: I think in this case we can just override the bot

justinfagnani (approved):
Very nice commits! Don't squash, preserve these.

.changeset/swift-dryers-invent.md
aomarks: This is actually what the bot says to do for this case in the CI output: `npx changeset --empty`

packages/ts-transformers/tsconfig.json
aomarks: It did actually catch some bugs! In e.g. the `customElement` decorator transform we had:

```ts
const [arg0] = decorator.expression.arguments;
if (!ts.isStringLiteral(arg0)) {
  return;
}
```

If the decorator had no arguments, `ts.isStringLiteral()` would throw, because it tries to access `.kind` on `undefined`. Turning on `noUncheckedIndexedAccess` showed me that we needed an extra guard here.

Even though type checking would also catch this particular error -- since the `customElement` decorator is typed to require an argument -- compilation would still crash during transform, since type checking doesn't block transformation.

It does require adding `!`s on a lot of safe for loops, but I feel it's worth it. I've definitely made mistakes in this area in the past too.

PR #2134, opened 2021-09-03
[reactive-element] Add CSP compatibility for element styles

### LINKED ISSUE #2025: Violation of "style-src 'self'" CSP when injecting component styles
# Description

When using a Lit web component on a page served from a web server having the `style-src` CSP set to `'self'`, current versions of Firefox and Safari report a CSP violation and prevent loading the components styles.

Surprisingly, Chrome works perfectly.

Tested browsers:
- ❌ Firefox: macOS, 90.0.2 (64-bit)
- ❌ Safari: macOS, Version 14.1.1 (16611.2.7.1.4)
- ✅ Chrome: macOS, 92.0.4515.107 (Official Build) (x86_64)

# Steps to Reproduce

TL;DR: Run following command to bring up the reproduction example on port `3000`:

```shell
docker run --it -rm -p 3000:80 ghcr.io/swissmanu/firefox-csp-lit:latest
```

The reproduction of this issue requires a specific web server setup. Please see https://github.com/swissmanu/firefox-csp-lit where I prepared a minimal reproducible example.
It consists of the actual code as well as ready-to-use Docker image serving that code via nginx having the regarding HTTP headers set.

## Expected Results

Styles get injected without violating any CSP, unrelated to the browser which the web component is used in.

## Actual Results

Along with an unstyled button, the Firefox dev tools console shows a CSP violation error:

Screenshot 2021-07-27 at 10 52 47

## Browsers Affected

- [ ] Chrome
- [X] Firefox
- [ ] Edge (untested)
- [X] Safari 11
- [X] Safari 10
- [ ] IE 11 (untested)

justinfagnani:
This PR lets devs set a `window.litNonce` property which is used to apply a CSP nonce to apply to generated `<style>` tags.

This lets an app set a unique-per-request nonce in server-generated HTML:

```html
<script>
  window.litNonce = 'abc123';
</script>
```

Which will work with a CSP policy of:

```
Content-Security-Policy: style-src 'nonce-abc123';
```

The indirection of specifying a global to hold the actual nonce is needed to make it easy to have a static JS build and a per-request varying nonce value. Importing the ReactiveElement class to set the static is potentially difficult to do for bundled, non-JS-modules-using apps.

Fixes #2025

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -150,7 +150,9 @@ export const adoptStyles = (
   } else {
     styles.forEach((s) => {
       const style = document.createElement('style');
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       if (nonceProperty && (window as any)[nonceProperty]) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         style.setAttribute('nonce', (window as any)[nonceProperty]);
       }
       style.textContent = (s as CSSResult).cssText;
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -134,10 +134,14 @@ export const css = (
  * the shadowRoot should be placed *before* any shimmed adopted styles. This
  * will match spec behavior that gives adopted sheets precedence over styles in
  * shadowRoot.
+ *
+ * @param nonceProperty The name of a global property that stores a CSP nonce
+ *   to use for any generated <style> tags.
  */
 export const adoptStyles = (
   renderRoot: ShadowRoot,
-  styles: Array<CSSResultOrNative>
+  styles: Array<CSSResultOrNative>,
+  nonceProperty?: string
 ) => {
   if (supportsAdoptingStyleSheets) {
     (renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) =>
@@ -146,6 +150,9 @@ export const adoptStyles = (
   } else {
     styles.forEach((s) => {
       const style = document.createElement('style');
+      if (nonceProperty && (window as any)[nonceProperty]) {
+        style.setAttribute('nonce', (window as any)[nonceProperty]);
+      }
       style.textContent = (s as CSSResult).cssText;
       renderRoot.appendChild(style);
     });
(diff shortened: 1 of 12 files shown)


samaniehsan:
@justinfagnani  thanks for your attention to this issue. we are currently using lit elements version 1.4
What version of lit elements would support this work-around?

justinfagnani:
This would be for `lit-element@3.0` / `lit@2.0`

packages/reactive-element/src/css-tag.ts
```diff
@@ -146,6 +150,11 @@ export const adoptStyles = (
   } else {
     styles.forEach((s) => {
       const style = document.createElement('style');
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (nonceProperty && (window as any)[nonceProperty]) {
```
sorvell, line 154:
```suggestion
      if (nonceProperty && (window as any)[nonceProperty] !== undefined) {
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -303,6 +303,35 @@ export abstract class ReactiveElement
   extends HTMLElement
   implements ReactiveControllerHost
 {
+  /**
+   * The name of a global property that stores a CSP nonce to use for any
+   * generated <style> tags.
+   *
+   * Element styles are implemented with <style> tags when the browser doesn't
+   * support adopted StyleSheets. To use <such style> tags with the style-src
```
sorvell, line 311:
```suggestion
   * support adopted StyleSheets. To use such <style> tags with the style-src
```

sorvell (changes requested):

sorvell:
Would it be better from a responsibility perspective to just have something like: `window.listStyleCallback` e.g. `(style) => style;` which we call if it's defined when we're creating a style. The user would then implement this if desired and set the `nonce` as needed.  This puts the onus for actually dealing with and security the nonce with the user?

justinfagnani:
> Would it be better from a responsibility perspective to just have something like: window.litCreateStyleCallback e.g. (style) => style; which we call if it's defined when we're creating a style. The user would then implement this if desired and set the nonce as needed. This puts the onus for actually dealing with and security the nonce with the user?

The only CSP related thing they could do in such a callback is set the `nonce` attribute, so I think we should just do this for them. There's precedent in other frameworks and tools of having a global that's written to `<script>` and `<style>`. I think we should just do the simple thing like they're doing.

justinfagnani:
I limited the tests to just the reactive-element styling tests by adding a new `reactive-element-csp_test.html` file with both the CSP header added via a `<meta>` tag and a script that sets the global.

justinfagnani:
cc @arthurevans for this new feature. I need to file a docs issue still.

sorvell (approved):

rictic (approved):

PR #2135, opened 2021-09-04
[reactive-element] Add addInitializer() docs

justinfagnani:
We need to add something similar to the lifecycle docs for lit.dev too.


AndrewJakubowicz (approved):
LGTM! Great documentation improvement!

Small non actionable notes:

Using the code example I tried to add a simple decorator in the lit playground to play with this API. I copied the code and immediately tried to use `@myDecorator()` instead of the correct `@myDecorator`.  I did this because of my expectation from other decorators that I should put params on the end.

It may be helpful to prompt how to use the example decorator.

PR #2145, opened 2021-09-09
Add dev mode specific versions of `polyfill-support`

### LINKED ISSUE #2142: Shady DOM dosen't work with modern bundlers
# Description

If you use a simple HTML with unpkg.com, shady DOM works

```html
<!DOCTYPE html>
<body>
    <outer-element></outer-element>
    <script>
        customElements.forcePolyfill = true;
        ShadyDOM = {force: true};
        ShadyCSS = {shimcssproperties: false};
    </script>
    <script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.6.0/webcomponents-bundle.js"></script>
    <script type="module">
        import 'https://unpkg.com/lit@2.0.0-rc.4/polyfill-support.js?module';
        import {LitElement, css, html} from 'https://unpkg.com/lit@2.0.0-rc.4?module';

        customElements.define(
          "outer-element",
          class extends LitElement {
            static get styles() {
              return [
                css`
                  :host {
                    color: orange;
                  }
                  div {
                    color: red;
                  }
                `
              ];
            }
            render() {
              return html`
                -> Outer element should be color orange...
                <div>${"Button Text Red"}</div>
              `;
            }
          }
        );
    </script>
</body>
```

But with bundler o dev servers (like vitejs) dosen't work

# Steps to Reproduce

1. Write this code

HTML
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Shady Dom Lit Element Simple Starter Kit</title>
    <script>
        customElements.forcePolyfill = true;
        ShadyDOM = {force: true};
        ShadyCSS = {shimcssproperties: true};
    </script>
</head>
<body>
    <outer-element></outer-element>
    <lit-app></lit-app>
    <script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.6.0/webcomponents-bundle.js"></script>
    <script type="module" src="/src/lit-app.ts"></script>
</body>
</html>
```
```ts
import 'lit/polyfill-support.js';
import { LitElement, html, css } from 'lit';
import { customElement } from 'lit/decorators.js';

@customElement('lit-app')
export class LitApp extends LitElement {
  static styles = [
    css`
      :host {
        display: block;
        color: blue;
      }
      div.red {
        color: red;
      }
    `
  ];

  render() {
    return html`
      <h1>Hello from TS</h1>
      <div class="red">${'Button Text Red'}</div>
    `;
  }
}

```

2. See this output...

## Live Reproduction Link

https://stackblitz.com/edit/vite-zv4cxk

## Expected Results

Generate Shady CSS scope Styles in header

## Actual Results

Don't generate shady scope styles in header

## Browsers Affected

- [ x ] Chrome 93
- [ x ] Firefox
- [ x ] Edge
- [ x ] Safari 11
- [ x ] Safari 10
- [ x ] IE 11

sorvell:
Fixes #2142. Allows users to a dev and prod versions of Lit that both support `polyfill-support` via loading dev and prod versions of `polyfill-support`. Previously only 1 `polyfill-support` was allowed.

Note, this also changes the globals used from the older `PlatformSupport` to `PolyfillSupport`.

### DIFF
--- a/packages/lit-element/src/polyfill-support.ts
+++ b/packages/lit-element/src/polyfill-support.ts
@@ -42,7 +42,12 @@ interface PatchableLitElement extends HTMLElement {
   renderOptions: RenderOptions;
 }
 
-globalThis.litElementPlatformSupport ??= ({
+// Note, explicitly use `var` here so that this can be re-defined when
+// bundled.
+// eslint-disable-next-line no-var
+var DEV_MODE = true;
+
+globalThis[`litElementPolyfillSupport${DEV_MODE ? `DevMode` : ``}`] ??= ({
   LitElement,
 }: {
   LitElement: PatchableLitElement;
--- a/.changeset/pretty-hotels-taste.md
+++ b/.changeset/pretty-hotels-taste.md
@@ -1,0 +1,7 @@
+---
+'lit-element': patch
+'lit-html': patch
+'@lit/reactive-element': patch
+---
+
+Adds dev mode specific versions of `polyfill-support`. This allows users to load prod and dev versions of Lit with polyfill support. To enable this, both the prod and dev versions of `polyfill-support` should be loaded.
--- a/packages/benchmarks/package-lock.json
+++ b/packages/benchmarks/package-lock.json
@@ -12,7 +12,7 @@
 				"tachometer": "^0.5.9"
 			},
 			"devDependencies": {
-				"chromedriver": "^91.0.0",
+				"chromedriver": "^92.0.0",
 				"typescript": "^4.3.5"
 			}
 		},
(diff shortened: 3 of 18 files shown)


justinfagnani (approved):

packages/lit-element/src/polyfill-support.ts
```diff
@@ -42,7 +42,12 @@ interface PatchableLitElement extends HTMLElement {
   renderOptions: RenderOptions;
 }
 
-globalThis.litElementPlatformSupport ??= ({
+// Note, explicitly use `var` here so that this can be re-defined when
+// bundled.
+// eslint-disable-next-line no-var
+var DEV_MODE = true;
```
kevinpschaaf, line 48: Seems bad that polyfill supports are being bundled together without iffe's such that this would error as a `const`, but afaiu that's safe because anytime we'd actually have conflicts between them they'd error, so probably not worth the time to fix now.

kevinpschaaf (approved):

PR #2146, opened 2021-09-11
Work around https://crbug.com/993268

rictic:
Upstreaming cl/396041127


packages/reactive-element/src/reactive-element.ts
```diff
@@ -974,6 +977,13 @@ export abstract class ReactiveElement
       if (attrValue == null) {
         this.removeAttribute(attr);
       } else {
+        if (attrValue === '' && attr.startsWith('on') && trustedTypes) {
```
justinfagnani, line 980: I think if this is done in [defaultConverter.toAttribute](https://github.com/lit/lit/pull/2146/files#diff-305d9543c5dd0eef82d2583a84422cfa2a1786f1f466692485e6dac5eb7d43baR226) it would be more similar to the fix in lit-html.ts.

justinfagnani: @rictic ping

justinfagnani: Here's what I'm talking about: https://github.com/lit/lit/compare/tt-crbug...tt-crbug-2

This removes any new conditionals from the __propertyToAttribute path. The question is whether this is a problem only for `type: Boolean` properties, where we induce an empty string for truthy values. I'm assuming that if the user passes in an actual empty string, it's on them to make sure it's a trusted value.

PR #2147, opened 2021-09-11
[ts-transformers] Fix bug with unsafe property access on import node

aomarks:
There was a spot where we assumed an `importClause.namedBindings` wouldn't be `undefined`. This caused an exception in one import scenario.

Also release this as 1.0.1


.changeset/cuddly-mugs-repeat.md
```diff
@@ -0,0 +1,2 @@
+---
```
justinfagnani, line 1: I don't think we need empty changesets... we can just merge with a failing changeset check

justinfagnani (approved):

.changeset/cuddly-mugs-repeat.md
aomarks: It's nice to see the overall green checkmark though, right? And they'll get deleted automatically on next publish?

(Not that we ever get a green checkmark, because our IE tests are permanently failing?)

Issue #1150, opened 2020-04-28
Problem with lit-html in IE11


niallcollins:
Hi all,

Working with lit-html and really like it. Works perfectly for me in all browsers except IE11 where the following error occurs:

Message: Syntax Error
directive.js
Line No: 68
Code: export const directive = <F extends DirectiveFactory>(f: F): F =>

Should I have any polyfills or specific setup to run on IE11? Please ask any questions and assistance would be appreciated as not working IE11 is a deal breaker for me. Thanks

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [X] IE 11

justinfagnani:
lit-html is distributed as modern JavaScript. To run in very old browsers like IE11 you'll need to compile it to their particular level of JavaScript. Babel is well suited to this, as well as many bundlers which include it. See also: https://lit-html.polymer-project.org/guide/tools

niallcollins:
Thanks for feedback. I already use Babel so will look to use this. An examples of projects available that may have this setup in place that I could review?

kevinpschaaf:
The open-wc website has good docs on using rollup + babel to compile modern JS to work on older browsers: https://open-wc.org/building/building-rollup.html

arthurevans:
For IE11, you'll need to load a template polyfill. There's one in the lit-html repo, but it's pretty sketchy. I'd recommend the one from the webcomponents polyfills (https://github.com/webcomponents/polyfills/tree/master/packages/template).

For a simple working example, here's what I came up with: https://github.com/PolymerLabs/lit-html-build
Deployed here: https://wizardly-davinci-e343a6.netlify.app/  ... Work on IE11 for me.

sorvell:
Closing based on the comments above. If the problem persists after implementing the recommendations above, please fell free to re-open this issue or open a new issue with more info.

PR #2152, opened 2021-09-14
Remove dynamic naming for `DEV_MODE` polyfill support functions.

bicknellr:
These dynamic lookups cause their expression to have an implicit `any` type, which is disallowed internally.

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -168,9 +168,10 @@ export class LitElement extends ReactiveElement {
 globalThis.litElementHydrateSupport?.({LitElement});
 
 // Apply polyfills if available
-globalThis[`litElementPolyfillSupport${DEV_MODE ? `DevMode` : ``}`]?.({
-  LitElement,
-});
+const polyfillSupport = DEV_MODE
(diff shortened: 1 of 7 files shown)


packages/lit-element/src/lit-element.ts
```diff
@@ -196,7 +197,7 @@ if (DEV_MODE) {
     };
     warnRemovedOrRenamed(this, 'render');
     warnRemovedOrRenamed(this, 'getStyles', true);
-    warnRemovedOrRenamed(this.prototype, 'adoptStyles');
+    warnRemovedOrRenamed(this.prototype as {}, 'adoptStyles');
```
bicknellr, line 200: This should be in a different PR.

bicknellr: Removed.

justinfagnani (approved):
LGTM, but there are non-timeout related IE11 errors. They don't look relates to this PR either, but I'm not sure when they might have crept in.

bicknellr:
`scoped-registry-mixin > host element should apply static styles` has been flaky in main for a bit now. I went through some of our recent runs and added [a comment about this in #1150](https://github.com/lit/lit/issues/1550#issuecomment-926244492).

PR #2155, opened 2021-09-14
[@lit-labs/react] Fix displayName

### LINKED ISSUE #2154: [@lit-labs/react] Wrong displayName for components.
# Description

Hello! 

All React components created with `@lit-labs/react` `createComponent` have all the same displayName. Which makes it hard to debug and doesn't work in [react-element-to-jsx-string](https://www.npmjs.com/package/react-element-to-jsx-string).

before

I have a PR with fix ready, this issue is to cover it.

# Steps to Reproduce

1. Write this code

```ts
import { html, LitElement } from "lit";
import { customElement } from "lit/decorators.js";
import * as React from "react";

@customElement("my-label")
class Label extends LitElement {
  render() {
    return html/* html */ `<label class="label"><slot></slot></label>`;
  }
}

const ReactComponent = createComponent(React, "my-label", {});

console.assert(ReactComponent.displayName === "Label");

```

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

g12i:
Fixes #2154

It also makes Lit React components work properly with [react-element-to-jsx-string](https://www.npmjs.com/package/react-element-to-jsx-string).

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -169,6 +169,8 @@ export const createComponent = <I extends HTMLElement, E>(
     private _userRef?: React.Ref<unknown>;
     private _ref?: React.RefCallback<I>;
 
+    static displayName = elementClass.name;
+
     private _updateElement(oldProps?: ComponentProps) {
       if (this._element === null) {
         return;
@@ -249,11 +251,17 @@ export const createComponent = <I extends HTMLElement, E>(
     }
   }
 
-  return React.forwardRef((props?: UserProps, ref?: React.Ref<unknown>) =>
-    createElement(
-      ReactComponent,
-      {...props, __forwardedRef: ref} as ComponentProps,
-      props?.children
-    )
+  const ForwardedComponent = React.forwardRef(
+    (props?: UserProps, ref?: React.Ref<unknown>) =>
+      createElement(
+        ReactComponent,
+        {...props, __forwardedRef: ref} as ComponentProps,
+        props?.children
+      )
   );
+
+  // To ease debugging in the React Developer Tools
+  ForwardedComponent.displayName = ReactComponent.displayName;
+
+  return ForwardedComponent;
 };
--- a/packages/labs/react/src/test/create-component_test.tsx
+++ b/packages/labs/react/src/test/create-component_test.tsx
@@ -83,10 +83,7 @@ suite('createComponent', () => {
   const renderReactComponent = async (
     props?: ReactModule.ComponentProps<typeof BasicElementComponent>
   ) => {
-    window.ReactDOM.render(
-      <BasicElementComponent {...props}/>,
-      container
-    );
+    window.ReactDOM.render(<BasicElementComponent {...props} />, container);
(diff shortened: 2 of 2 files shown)


g12i:
@googlebot I signed it!

justinfagnani:
I wonder if this is a good default that will actually work with react-element-to-jsx-string.

```ts
class AwesomeThingElement extends LitElement {}

const AwesomeThing =  createComponent(
    window.React,
    'awesome-thing',
    AwesomeThingElement);
```

So now `AwesomeThing` has a display name of `AwesomeThingElement`. I guess element will usually be defined in a separate module, so you can do:

```ts
class AwesomeThingElement extends LitElement {}
```
```ts
import {AwesomeThingElement as AwesomeThingElementClass} from './awesome-element.js';

const AwesomeThingElement =  createComponent(
    window.React,
    'awesome-thing',
    AwesomeThingElementClass);
```

Which will ensure that the class and exported component have the same name. That might be worth documenting... wdyt?

g12i:
Hey @justinfagnani!

I'm not sure if I understand :(. Are you suggesting that the user would expect the `displayName` to be the same as the variable name, as this is how it usually works? Yes, this makes sense. I forgot about it as for me the React component was default export, so I assumed it should simply be the same.

In my view the `class ReactComponent` should mimic Lit's class `.name`.

```tsx
static displayName = elementClass.name;
```

And then if we follow [React's HOC guide](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components) it return look like this:

```tsx
  function forwardRef(props?: UserProps, ref?: React.Ref<unknown>) {
    return createElement(
      ReactComponent,
      { ...props, __forwardedRef: ref } as ComponentProps,
      props?.children
    );
  }

  forwardRef.displayName = `Lit(${elementClass.name})`;

  return React.forwardRef(forwardRef);
```

Which creates nice output in React Dev Tools

Screenshot 2021-09-15 at 10 30 47

But it's ugly in `react-element-to-jsx-string`.

obraz

The only way to make it work in `react-element-to-jsx-string` is to explicitly define `.displayName` of the exported Object (the one returned by `React.forwardRef`). 

This one looks more like a `react-element-to-jsx-string` problem 😕 .

Here's my playground: https://github.com/g12i/lit-playground

g12i:
Alternatively `displayName` could be a function parameter
```tsx
const createComponent = <I extends HTMLElement, E>(
  React: typeof ReactModule,
  tagName: string,
  elementClass: Constructor<I>,
  events?: StringValued<E>,
  displayName?: string
) => {
// ...


  class ReactComponent extends Component<ComponentProps> {
    private _element: I | null = null;
    private _elementProps!: { [index: string]: unknown };
    private _userRef?: React.Ref<unknown>;
    private _ref?: React.RefCallback<I>;

    static displayName = displayName ?? elementClass.name;
```

g12i:
Hey @justinfagnani any thoughts?

sorvell:
Optionally allowing the user to set the `displayName` sounds good.

PR #2160, opened 2021-09-14
Don't throw the shadowed properties dev-mode error when a property is intentionally not reactive.

bicknellr:
This PR prevents the dev-mode error about shadowed properties from being thrown if the property is configured not to have a generated descriptor. This includes setting `noAccessor` in the properties block / `@property` decorator and returning undefined from `getPropertyDescriptor`.

This PR also adds `undefined` as a possible return type for `getPropertyDescriptor`, which is documented as if it should be able to do this, but currently isn't allowed due to the inferred return type. This change required adding a few `!` assertions in some existing tests.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -601,7 +601,7 @@ export abstract class ReactiveElement
     name: PropertyKey,
     key: string | symbol,
     options: PropertyDeclaration
-  ) {
+  ): PropertyDescriptor | undefined {
     return {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       get(): any {
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -38,6 +38,8 @@ let requestUpdateThenable: (name: string) => {
 
 let issueWarning: (code: string, warning: string) => void;
 
+let prototypeToReactivePropertyNames: WeakMap<{}, Set<PropertyKey>>;
+
 if (DEV_MODE) {
   // Ensure warnings are issued only 1x, even if multiple versions of Lit
   // are loaded.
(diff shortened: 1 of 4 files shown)


packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -3075,4 +3075,95 @@ suite('ReactiveElement', () => {
       assert.equal(el2.attrValue, 'custom');
     });
   });
+
+  if (DEV_MODE) {
```
sorvell, line 3079: This should be moved to `reactive-element_dev_mode_test`

sorvell (changes requested):

packages/reactive-element/src/test/reactive-element_test.ts
bicknellr: Ok, I moved these into that file.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1140,7 +1153,13 @@ export abstract class ReactiveElement
         const shadowedProperties: string[] = [];
         (this.constructor as typeof ReactiveElement).elementProperties.forEach(
           (_v, p) => {
-            if (this.hasOwnProperty(p) && !this.__instanceProperties?.has(p)) {
+            if (
```
sorvell, line 1156: Rather than loop over `elementProperties`, this can just loop over `prototypeToReactivePropertyNames.get(this.constructor.prototype)`, right?

sorvell (changes requested):

packages/reactive-element/src/reactive-element.ts
bicknellr: Yeah, I think you're right: in `createProperty`, the `elementProperties` entry is always set before the `prototypeToReactivePropertyNames` entry might be set, so `prototypeToReactivePropertyNames` should always be the subset we want here. Updated.

Issue #1883, opened 2021-05-12
[all] Expand jsdoc documentation


aomarks:
The jsdoc documentation for all the core libraries are currently very lacking, as can be seen by browsing through https://lit.dev/docs/api/LitElement/ and related pages.

We should sweep through every API and a whole bunch of missing documentation.

AndrewJakubowicz:
Note the list below was generated from the lit.dev API docs. There's duplication between ReactiveElement and LitElement because of inheritance. Haven't deduped the list.

Publicly generated JSDoc symbols to audit/improve (top level):

## LitElement

 - [ ] [LitElement](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/lit-element/src/lit-element.ts#L86) - /packages/lit-element/src/lit-element.ts#L86
 - [ ] [RenderOptions](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/lit-html/src/lit-html.ts#L375) - /packages/lit-html/src/lit-html.ts#L375

## ReactiveElement

 - [ ] [ReactiveElement](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/reactive-element/src/reactive-element.ts#L314) - /packages/reactive-element/src/reactive-element.ts#L314
 - [ ] [UpdatingElement](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/lit-element/src/lit-element.ts#L56) - /packages/lit-element/src/lit-element.ts#L56
 - [ ] [ComplexAttributeConverter](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/reactive-element/src/reactive-element.ts#L117) - /packages/reactive-element/src/reactive-element.ts#L117
 - [ ] [PropertyDeclaration](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/reactive-element/src/reactive-element.ts#L141) - /packages/reactive-element/src/reactive-element.ts#L141
 - [ ] [PropertyDeclarations](https://github.com/lit/lit/blob/4a5369c8c9b282c4f47a70126da11810e9754745/packages/reactive-element/src/reactive-element.ts#L213) - /packages/reactive-element/src/reactive-element.ts#L213
[...]

PR #2161, opened 2021-09-15
Expand jsdoc documentation for reactive-element and lit-element.

AndrewJakubowicz:
### Context

Add some documentation for some public undocumented methods in lit-element and reactive-element. Addresses github issue: https://github.com/lit/lit/issues/1883

### How

Used lit.dev documentation to guide the jsdocs, and addressed some class members that had no documentation on the lit.dev/docs/api pages.

Added some documentation for previously undocumented:
 - `LitElement.connectedCallback()`
 - `LitElement.disconnectedCallback()`
 - `ReactiveElement.hasUpdated`
 - `ReactiveElement.addController`
 - `ReactiveElement.removeController`

### Testing

Tested with mouse over in VsCode.

Thank you!

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -759,11 +759,15 @@ export abstract class ReactiveElement
   private __updatePromise!: Promise<boolean>;
 
   /**
+   * Internal flag marking a pending update.
    * @category updates
    */
   isUpdatePending = false;
 
   /**
+   * Is set to `true` after the first update. Required for compatibility with
+   * server-side rendering as the element code cannot assume that `renderRoot`
+   * exists before the element `hasUpdated`.
    * @category updates
    */
   hasUpdated = false;
@@ -817,6 +821,9 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Install a [[`ReactiveController`]] onto the element's reactive update
+   * cycle. LitElement automatically calls into any installed controllers during
+   * its lifecycle callbacks.
    * @category controllers
    */
   addController(controller: ReactiveController) {
@@ -831,6 +838,7 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Remove a [[`ReactiveController`]] from the element.
    * @category controllers
    */
   removeController(controller: ReactiveController) {
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -821,8 +821,8 @@ export abstract class ReactiveElement
   }
 
   /**
-   * Install a `ReactiveController` onto the element's reactive update cycle.
-   * LitElement automatically calls into any installed controllers during its
+   * Registers a `ReactiveController` to participate in the element's reactive update cycle.
+   * The element automatically calls into any registered controllers during its
    * lifecycle callbacks.
    * @category controllers
    */
@@ -838,7 +838,7 @@ export abstract class ReactiveElement
   }
 
   /**
-   * Remove a `ReactiveController` from the element.
+   * Removes a `ReactiveController` from the element.
    * @category controllers
    */
   removeController(controller: ReactiveController) {
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -759,15 +759,15 @@ export abstract class ReactiveElement
   private __updatePromise!: Promise<boolean>;
 
   /**
-   * Internal flag marking a pending update.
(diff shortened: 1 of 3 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -817,6 +821,9 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Install a [[`ReactiveController`]] onto the element's reactive update
```
aomarks, line 824: Maybe we should start using e.g. `{@link ReactiveController}`, since that will work in VSCode too. We will need to change or update this regexp to extract them for our lit.dev API docs: https://github.com/lit/lit.dev/blob/e2d7fd3625fb021c539348d7ad06121e871a8ebd/packages/lit-dev-tools/src/generate-api-docs-data.ts#L730

aomarks (approved):
Nice docs!

Link thing is optional -- maybe something we should do as a search/replace across everything.

packages/reactive-element/src/reactive-element.ts
aomarks: (Optional, maybe we should do this all at once with a search/replace).

AndrewJakubowicz: Great point. Will remove links from this change and tracking: https://github.com/lit/lit.dev/issues/503

packages/reactive-element/src/reactive-element.ts
```diff
@@ -817,6 +821,9 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Install a `ReactiveController` onto the element's reactive update cycle.
+   * LitElement automatically calls into any installed controllers during its
+   * lifecycle callbacks.
```
arthurevans, line 826: "Install... onto" seems a little awkward? Maybe:

```suggestion
   * Registers a `ReactiveController` to participate in the element's reactive update cycle.
   * The element automatically calls into any registered controllers during its
   * lifecycle callbacks.
```
This matches the language in the guide more closely. Only thing I don't like is the term "registered controllers" which sounds kinda weird. 

For the second sentence, we could say, "`LitElement`" (the class) or "Lit" (the library) but in this case I thought maybe it was clearer to just say, "the element" because we're specifically talking about a single instance (this instance invokes its registered controllers).

packages/reactive-element/src/reactive-element.ts
```diff
@@ -831,6 +838,7 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Remove a `ReactiveController` from the element.
```
arthurevans, line 841:
```suggestion
   * Removes a `ReactiveController` from the element.
```

arthurevans (approved):
LGTM. Added a few minor suggestions.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -759,11 +759,15 @@ export abstract class ReactiveElement
   private __updatePromise!: Promise<boolean>;
 
   /**
+   * Internal flag marking a pending update.
    * @category updates
    */
   isUpdatePending = false;
 
   /**
+   * Is set to `true` after the first update. Required for compatibility with
```
justinfagnani, line 768: I'd remove the "required for..." and just say "The element cannot assume..."

packages/reactive-element/src/reactive-element.ts
```diff
@@ -817,6 +821,9 @@ export abstract class ReactiveElement
   }
 
   /**
+   * Registers a `ReactiveController` to participate in the element's reactive update cycle.
+   * The element automatically calls into any registered controllers during its
+   * lifecycle callbacks.
```
justinfagnani, line 826: Add "If the element is connected when `addController()` is called, the controller's `hostConnected()` callback will be immediately called.

AndrewJakubowicz:
Is this ok to merge with the failing tests? None of them seem related to jsdocs?

aomarks:
> Is this ok to merge with the failing tests? None of them seem related to jsdocs?

The changeset failure is because there's no changeset. Might actually make sense here, since it is a change to the distributd package. You can use `npx changeset` for a wizard that lets you create an entry.

I think the Sauce tests always fail, something has been up with those for weeks. IE11 busted? Not sure.

aomarks:
> The changeset failure is because there's no changeset. Might actually make sense here, since it is a change to the distributed package, and a changeset will signal that there is a pending change that hasn't been released yet, and eventually make its way into the CHANGELOG. You can use `npx changeset` for a wizard that lets you create an entry.

Just something general like "Additional documentation on Lit APIs" or something seems fine.

PR #2162, opened 2021-09-16
Additional documentation for `html` and `nothing`

AndrewJakubowicz:
Addresses github issue: https://github.com/lit/lit/issues/1883

### Context

Documenting lit-html's `html` and `nothing`. The API documentation is sparse so I've borrowed heavily from the original [lit-html polymer website](https://lit-html.polymer-project.org/guide/writing-templates).

Let me know if there is too much (especially in the `html` tag case) and if we only want to keep a smaller amount or something different. Intent was to share that there isn't string concatenation going on under the hood.

Identified as currently on lit.dev there is minimal documentation and they are heavily used in Lit: https://lit.dev/docs/api/templates/

### Testing

Tested with mouse over in VsCode.

Thank you!

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -337,18 +337,21 @@ export const noChange = Symbol.for('lit-noChange');
 /**
  * A sentinel value that signals a ChildPart to fully clear its content.
  *
- * The values `undefined`, `null`, and the empty string `''` all render an empty
- * text node. To render nothing at all, use `nothing` provided by Lit:
- *
  * ```ts
- * import {html, nothing} from 'lit';
- *
  * const button = html`${
  *  user.isAdmin
  *    ? html`<button>DELETE</button>`
  *    : nothing
  * }`;
  * ```
+ *
+ * Prefer using `nothing` over other falsy values as it provides a consistent
+ * behavior between various expression binding contexts.
+ *
+ * In child expressions, `undefined`, `null`, `''`, and `nothing` all behave the
+ * same and render no nodes. In attribute expressions, `nothing` _removes_ the
+ * attribute, while `undefined` and `null` will render an empty string. In
+ * property expressions `nothing` becomes `undefined`.
  */
 export const nothing = Symbol.for('lit-nothing');
 
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -310,6 +310,54 @@ const tag =
 /**
  * Interprets a template literal as an HTML template that can efficiently
  * render to and update a container.
+ *
+ * ```ts
+ * const header = (title: string) => html`<h1>${title}</h1>`;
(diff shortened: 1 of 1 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -310,6 +310,54 @@ const tag =
 /**
  * Interprets a template literal as an HTML template that can efficiently
  * render to and update a container.
+ *
+ * ```ts
+ * const header = (title: string) => html`<h1>${title}</h1>`;
+ * ```
+ *
+ * The `html` tag takes the literal strings and embedded expression values for
```
aomarks, line 318: Does this text come from lit.dev, or is it new?

It does seem like there's a downside to having **too detailed** documentation here. Otherwise we are maintaining the same documentation in two different places, which will definitely get out of sync.

I wonder if we should do something a little more concise, and then to the relevant lit.dev section?

justinfagnani: Yeah, I don't think we need to describe how this works. I think we might hit a couple of points like:
- this returns a description of the DOM to render as a value
- Maybe that the value is a TemplateResult (but even that's pretty drilled in)
- It captures the expression values and the static template strings
- The result needs to be passed to render() to do anything
- If a template result comes from the same expression as a previously rendered result, it's updated instead of replaced
- Point to the template docs on lit.dev

AndrewJakubowicz: Agreed! Do we have this info on lit.dev (I couldn't find it). My reference was: https://lit-html.polymer-project.org/guide/writing-templates

Is the plan that lit.dev contains all the documentation for various packages?

justinfagnani: Yes, lit.dev is completely replacing the other sites.

AndrewJakubowicz: I opened issue: https://github.com/lit/lit.dev/issues/504, since not sure if we have a lit.dev location for these inner concepts yet.

Made much more succinct in follow up commit. Thank you!

packages/lit-html/src/lit-html.ts
```diff
@@ -327,6 +336,19 @@ export const noChange = Symbol.for('lit-noChange');
 
 /**
  * A sentinel value that signals a ChildPart to fully clear its content.
+ *
+ * The values `undefined`, `null`, and the empty string `''` all render an empty
```
justinfagnani, line 340: This changed in Lit 2. In child expressions, `undefined`, `null`, `''`, and `nothing` all render the same thing now - no nodes at all. In attribute expressions `undefined`, etc., will render an empty string and `nothing` will _remove_ the attribute. In property expressions `nothing` becomes `undefined`.

AndrewJakubowicz: This is incredible info thank you. Will add.

justinfagnani (approved):

PR #2173, opened 2021-09-20
copyedit README

arthurevans:
Minor edits for consistency following the renaming.

### DIFF
--- a/packages/labs/motion/README.md
+++ b/packages/labs/motion/README.md
@@ -13,10 +13,10 @@ $ npm install @lit-labs/motion
 ## Animate directive
 
 The `animate` directive can be used to animate DOM elements from one lit render
-to the next. If the `animate` element between renders, it will perform a "tweening"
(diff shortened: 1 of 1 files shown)


arthurevans:
A couple of notes I didn't feel sure enough of to add to this PR:

The README mentions "standard KeyframeAnimationOptions." I wanted to link to that API, but I can't find it on MDN. TypeScript docs include KeyframeAnimationOptions, which extends KeyframeEffectOptions. MDN only mentions the latter.  The only difference I can find between the two is that KeyframeAnimationOptions includes an `id` field.

Also, was wondering whether it would be worthwhile linking the keyframe properties (`in` and `out`) with Keyframe Formats page (https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Keyframe_Formats#syntax). I think that's what's expected here.

sorvell (approved):

Issue #2183, opened 2021-09-22
[localize] Don't require a tsconfig especially for JS projects


aomarks:
Currently all localize projects require an existing `tsconfig.json`. We use this to determine which file paths to analyze, and which TypeScript compiler options to set for analysis.

However, this shouldn't be required, since we can just generate our own default config internally.

This is especially important for JS users, who currently have to create a tsconfig with `allowJs: true` just in order for localize analysis/build to work, which is unnecessary friction.

We will need an alternate way to specify which files to analyze, though.

Related to https://github.com/lit/lit/issues/2177

Issue #2182, opened 2021-09-22
[localize] Option to emit JS instead of TS


aomarks:
Currently the modules generated by localize are always TS. We should have the option to emit JS. Actually, there is literally only one TypeScriptism in any of the generated modules (`as const`), so this should be quite simple.

Currently, a JS user would have to run `tsc` just to remove this type annotation.

Related to https://github.com/lit/lit/issues/2177

Issue #2177, opened 2021-09-21
[localize] Provide README documentation and example for @lit/localize without TS


aomarks:
Compiling a JS project should be possible with the `allowJs` option, but we should prove this out with an example, and document it at https://lit.dev/docs/libraries/localization/.

### Discussed in https://github.com/lit/lit/discussions/2176

<div type='discussions-op-text'>

<sup>Originally posted by **mario-aleo** September 21, 2021</sup>
Do @lit/localize support plain JS projects?

I can't manage to setup @lit/localize on a project that isn't using TS.

Even though https://lit.dev/docs/libraries/localization/ has a **JS** option on exemples, I can run `lit-localize extract` because the needed ` tsconfig.json` can't find any file to work with :(

</div>

aomarks:
Related lit.dev issue: https://github.com/lit/lit.dev/issues/512

PR #2188, opened 2021-09-23
[localize] Improve vanilla JS ergonomics and documentation

aomarks:
**Note to reviewer**: I highly recommend looking at each commit separately.

- **Don't require a `tsconfig.json`**. Adds an `inputFiles` field, and makes the `tsConfig` field optional when `inputFiles` is specified. When `tsConfig` is not specified, a default config is used that will include `.js` files. When both are set, `inputFiles` takes precedence. 

- **Option to emit generated modules as `.js` files**.  Adds a new `output.language` setting for runtime mode locale generation, and automatically detects the filetype based on the file extension when using `output.localeCodesModule`.

- **Adds JS versions of example projects**: `examples/runtime-js` and `examples/transform-js`.

- **Documented vanilla JS usage in README**.

- (Unrelated) Fixed the `$schema` property that is automatically added to `@lit/localize-tools` config files. It was previously pointing at the incorrect URL.

Fixes https://github.com/lit/lit/issues/2183
Fixes https://github.com/lit/lit/issues/2182
Fixes https://github.com/lit/lit/issues/2177


packages/localize/README.md
```diff
@@ -224,15 +222,14 @@ lit-localize supports two output modes: _transform_ and _runtime_.
...
-> signatures to identify calls to `msg` and other APIs during analysis of your
-> code. Casting a lit-localize function to a type that does not include its
-> annotation will prevent lit-localize from being able to extract and transform
-> templates from your application. For example, a cast like
-> `(msg as any)("Hello")` will not be identified. It is safe to re-assign
-> lit-localize functions or pass them as parameters, as long as the distinctive
-> type signature is preserved. If needed, you can reference each function's
-> distinctive type with e.g. `typeof msg`.
+> Note that lit-localize relies on distinctive, annotated type signatures to
+> identify calls to `msg` and other APIs during analysis of your code. Casting a
```
justinfagnani, line 226: Casting is a TS concept. What's the equivalent guidance for JS?

packages/localize/README.md
```diff
@@ -224,15 +222,14 @@ lit-localize supports two output modes: _transform_ and _runtime_.
...
+> lit-localize function to a type that does not include its annotation will
+> prevent lit-localize from being able to extract and transform templates from
+> your application. For example, a cast like `(msg as any)("Hello")` will not be
+> identified. It is safe to re-assign lit-localize functions or pass them as
```
justinfagnani, line 230: I assume it's _not_ safe to reassign in JS?

packages/localize-tools/src/modes/runtime.ts
```diff
@@ -58,6 +58,21 @@ async function runtimeOutput(
   config: Config,
   runtimeConfig: RuntimeOutputConfig
 ) {
+  let fileExtension;
+  const language = runtimeConfig.language ?? 'ts';
```
justinfagnani, line 62: Is this the real default? Seems like we should only default to TS when there's a tsconfig file, but that may be happening elsewhere.

packages/localize-tools/src/types/modes.d.ts
```diff
@@ -11,21 +11,32 @@ export interface RuntimeOutputConfig {
   mode: 'runtime';
 
   /**
-   * Output directory for generated TypeScript modules. Into this directory will
-   * be generated a <locale>.ts for each `targetLocale`, each a TypeScript
-   * module that exports the translations in that locale keyed by message ID.
+   * Language for emitting generated modules. Options:
+   *
+   * - "ts": Default. Emit TypeScript modules with ".ts" file extension.
```
justinfagnani, line 16: Yeah, as with the other comment, I think js should be the default unless there's a tsconfig, although... maybe that default can change in a tool that wraps the transformer for usage outside TS toolchains?

packages/localize/README.md
aomarks: You can use a `/** @type import("@lit/localize").msg */` JSDoc comment.

I've actually moved this whole section to an FAQ item at the bottom of this README, and added examples of how to do the type cast in JS and TS. I don't think this issue actually needs to be so prominent, since it's a bit distracting and scary, and it only applies if you're doing something outside of our documentation that should be pretty unusual -- and even then most simple re-export and re-assign patterns will work fine in JS and TS.

packages/localize/README.md
aomarks: Actually most simple patterns like a re-export or re-assign work just fine, since TS will follow references the same way it does in a .ts file.

packages/localize-tools/src/modes/runtime.ts
aomarks: Good point! I was thinking we'd need to default to `ts` to avoid a breaking change, but since `tsConfig` was required before, we can make `js` the default when it is not set, without a breaking change.

Done.

packages/localize-tools/src/types/modes.d.ts
aomarks: Done.

packages/localize/README.md
```diff
@@ -609,6 +625,54 @@ export default locales.map(({locale, localeTransformer}) => ({
 }));
 ```
 
+### JavaScript example
```
justinfagnani, line 628: Not for this PR, but I'd consider breaking the readme up into multiple files, or paring it down and putting most of this on lit.dev

justinfagnani (approved):

packages/localize/README.md
aomarks: I agree! I plan to move all of this content to https://lit.dev/docs/libraries/localization/ pretty soon

PR #2197, opened 2021-09-28
[labs/motion] Add meaningful error when not used on LitElement

alfonsogarciacaro:
Because directives can be used with standalone templates, users can get confused when `animate` is not working (I had to debug the code to find the problem). Not sure if this the best way to do the check but it'd be nice to offer a more meaningful error when the host `LitElement` cannot be found. It'd be also desirable to put in the README, I can see a PR to include a mention if that's OK.

### DIFF
--- a/packages/labs/motion/src/animate.ts
+++ b/packages/labs/motion/src/animate.ts
@@ -192,6 +192,9 @@ export class Animate extends AsyncDirective {
     const firstUpdate = this._host === undefined;
     if (firstUpdate) {
       this._host = part.options?.host as LitElement;
+      if (this._host == null) {
+        throw new Error("animate directive must be used in a LitElement");
+      }
       this._host.addController(this);
       this.element = part.element;
       nodeToAnimateMap.set(this.element, this);


packages/labs/motion/src/animate.ts
```diff
@@ -192,6 +192,9 @@ export class Animate extends AsyncDirective {
     const firstUpdate = this._host === undefined;
     if (firstUpdate) {
       this._host = part.options?.host as LitElement;
+      if (this._host == null) {
+        throw new Error("animate directive must be used in a LitElement");
```
justinfagnani, line 196: The host just has to implement the ReactiveControllerHost interface. So it can be a ReactiveElement or other implementation.

alfonsogarciacaro: Thanks for checking @justinfagnani! Yes, that's what I thought at the beginning. I've implemented [HookComponents](https://fable.io/Fable.Lit/docs/hook-components.html) (similar to VirtualComponent in Haunted) using `AsyncDirective` in lit 2, and tried to implement the [ReactiveControllerHost interface](https://lit.dev/docs/api/controllers/#ReactiveControllerHost). But this was not enough for `animate` because it depends on methods that are not on the interface but come from ReactiveElement class, like `hasUpdated` or `updateCompleted`: https://github.com/lit/lit/blob/fba16cc1ceac30184514d32cd8402f9086e0eb43/packages/labs/motion/src/animate.ts#L445-L453

Actually I wanted to ask about this, because if Reactive Controllers stick to the ReactiveControllerHost interface they could be used more broadly (e.g. you could create a React hook that accepts reactive controllers), but if controllers are going to depend on LitElement at the end this won't be the case (and it makes the `ReactiveControllerHost` interface somewhat useless).

augustjk:
closing in favor of https://github.com/lit/lit/pull/2239

Issue #2178, opened 2021-09-21
[all] Fully automate the release process


justinfagnani:
We have many of the pieces in place for fully automated releases, but we need to close the loop. We should require that Lit 2.1, and hopefully patch releases before then, are fully automated.

This requires at least:
- [ ] A custom changesets changelog generator
- [ ] A script to update the `litHtmlVersion`, etc. globals
- [ ] Adding Git tags
- [ ] Making GitHub releases
- [ ] Uploading single-file builds (ie #1257 ) to GitHub for distribution

bicknellr:
All of the tasks above are done. Landing https://github.com/lit/lit.dev/pull/673 is the last thing here.

PR #2205, opened 2021-09-29
Add custom changelog generator function

justinfagnani:
Part of #2178 

As part of testing this I noticed that we're generating changelogs for the localize examples. We need to figure out how to exclude those for full automation.

### DIFF
--- a/.changeset/changelog-lit.cjs
+++ b/.changeset/changelog-lit.cjs
@@ -1,0 +1,116 @@
+const {
+  getInfo,
+  getInfoFromPullRequest,
+} = require('@changesets/get-github-info');
+
+// Forked from: https://github.com/atlassian/changesets/blob/main/packages/changelog-github/src/index.ts
+// Remove the "Thanks!" message, as it's almost always self-congratuatory to our team
+// TODO: add back "Thanks!" for external contributors
+
+const repo = 'lit/lit';
+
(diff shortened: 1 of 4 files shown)


.changeset/changelog-lit.cjs
```diff
@@ -0,0 +1,116 @@
+const {
```
aomarks, line 1: Add license header, I think this is good?

```
/**
 * @license
 * Copyright 209 Ben Conolly
 * SPDX-License-Identifier: MIT
 */
```

.changeset/changelog-lit.cjs
```diff
@@ -0,0 +1,116 @@
...
+  getInfo,
+  getInfoFromPullRequest,
+} = require('@changesets/get-github-info');
+
+// Forked from: https://github.com/atlassian/changesets/blob/main/packages/changelog-github/src/index.ts
```
aomarks, line 6: nbd but would be nice in future to add a commit before this one which is the pristine version, so that we can see the diff

(I did a diff manually and lgtm)

aomarks (approved):

.changeset/changelog-lit.cjs
justinfagnani: yeah, good idea

PR #2207, opened 2021-09-30
Add script to update global version variables

justinfagnani:
Part of #2178

This script runs after `changeset version` and copies the new versions into the source files that have these global version variables.

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -269,7 +269,6 @@ export const _$LE = {
 
 // IMPORTANT: do not change the property name or the assignment expression.
 // This line will be used in regexes to search for LitElement usage.
-// TODO(justinfagnani): inject version number at build time
 (globalThis.litElementVersions ??= []).push('3.0.0');
 if (DEV_MODE && globalThis.litElementVersions.length > 1) {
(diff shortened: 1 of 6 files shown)


packages/lit-element/src/lit-element.ts
```diff
@@ -269,8 +269,7 @@ export const _$LE = {
 
 // IMPORTANT: do not change the property name or the assignment expression.
 // This line will be used in regexes to search for LitElement usage.
-// TODO(justinfagnani): inject version number at build time
-(globalThis.litElementVersions ??= []).push('3.0.0');
+(globalThis.litElementVersions ??= []).push('0.0.0');
```
justinfagnani, line 272: Oops, I need to roll this back

scripts/update-version-variables.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
aomarks, line 3: 2021

kevinpschaaf (approved):

justinfagnani:
@aomarks I'll move or remove that script. It might be nice to have an option that just checks that the versions are in sync.

PR #2239, opened 2021-10-11
Remove reference to LitElement in lit-labs/motion

alfonsogarciacaro:
The animate controller should be compatible with any component implementing `ReactiveControllerHost` not only LitElement, see #2197

### DIFF
--- a/.changeset/cold-windows-pretend.md
+++ b/.changeset/cold-windows-pretend.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/motion': patch
+---
+
+Make `animate` directive compatible with any element implementing `ReactiveControllerHost` instead of just `LitElement`.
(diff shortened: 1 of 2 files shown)


leolorenzoluis:
@justinfagnani Thoughts?

aomarks:
cc @sorvell

RhondaP750:
I have no idea what any of this means, but I am trying to learn.

justinfagnani:
I'm mostly ok with this. My one concern is that we don't find a need for having a LitElement host in the future.

Well, and that I have some doubts the usefulness of this change. This is a lit-html directive that only works if the host is a ReactiveElement. What other base classes besides LitElement meet that criteria? Did someone extend ReactiveElement to use lit-html, but not use LitElement? If so, I wonder why?

OnurGumus:
Any updates for this?

augustjk (approved):
Sorry for taking so long to get to this. The changes here make sense.
I rebased to current main, resolving conflict, and added a changeset.
Thank you very much for the contribution!

Issue #2246, opened 2021-10-13
[labs/ssr] Cannot SSR a dynamic tag with `unsafeStatic`


shakyShane:
https://gist.github.com/shakyShane/423b19e9f479b06369eb90f4f503310a

The following stand-alone code examples both cause the error `Error: unexpected final partIndex: 0 !== 1`. 

I'm using `unsafeStatic` in a controlled way (eg: not user-provided data).

```js
import "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { render as renderDomShim } from "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { html, unsafeStatic } from 'lit/static-html.js'
import { LitElement } from 'lit'

/**
 * Rendering a static `<p>` fails
 *
 * ```
 * Error: unexpected final partIndex: 0 !== 1
 * ```
 */
const staticP = unsafeStatic('p');
for (const part of renderDomShim(html`<${staticP}>hello</${staticP}>`)) {
  console.log(part);
}

/**
 * Rendering a custom element also fails
 */
customElements.define('any-thing', class extends LitElement {
  render() {
    return html`<p>Hello world</p>`
  }
})

const element = unsafeStatic('any-thing');
for (const part of renderDomShim(html`<${element}>hello</${element}>`)) {
  console.log(part);
}
```

Any help will be greatly appreciated :)

shakyShane:
@justinfagnani thanks for the clarity. I've worked around it with a patch for the time being, but who knows what that patch might otherwise break 😢

justinfagnani:
@shakyShane I can't speak to the priority of this quite yet, but I am ramping up on a bunch of SSR work. I will try to take a closer look soon.

justinfagnani:
btw, @shakyShane just so you're aware, SSR support for static templates might be very tricky. It's probably similar to React and `dangerouslySetInnerHtml` where for hydration you'll need to overwrite the whole template instance.

dunqan:
@justinfagnani is there any update for it? even partial one, i.e. to be able to render dynamic tags without hydration support (or with some drawback, like dropping server side rendered shadow DOM, and just recreate it on client side completely)?

If not, what would be the best approach for now, to have some data driven dynamic tags with Lit, working both on SSR and client side?

tpluscode:
I just found this and would really appreciate if dynamic tags worked in SSR

@shakyShane can you share your patch? It could just work for me for the time being too

[...]

PR #2247, opened 2021-10-13
added failing test case for #2246

shakyShane:
I added this failing test case for #2246 to speed up the debugging process :)


augustjk:
Original issue was fixed with https://github.com/lit/lit/pull/3825 which added a separate integration test for static html.

PR #2261, opened 2021-10-24
[labs/motion] Add Spring and Spring2D controllers

justinfagnani:
Adds `SpringController` and `SpringController2D` that you can use to simulate springs that drive host updates, like so:

```ts
@customElement('goo-element')
export class GooElement extends LitElement {
  _mouse = new MouseController(this);
  _spring1 = new SpringController2D(this, fast);

  render() {
    // Set the spring to go to the mouse
    this._spring1.toPosition = this._mouse.position;
    // Position a div based on the current position of the spring.
    return html`
      <div
        class="b1"
        style=${positionStyle(this._spring3.currentPosition)}
      ></div>
    `;
  }
}

const fast = {
  stiffness: 1200,
  damping: 400,
};

const positionStyle = ({x, y}: Position2D) =>
  styleMap({
    transform: `translate3d(${x}px,${y}px,0) translate3d(-50%,-50%,0)`,
  });
```

### DIFF
--- a/packages/labs/motion/README.md
+++ b/packages/labs/motion/README.md
@@ -133,22 +133,24 @@ The Spring and Spring2D controllers simulate physical springs with the [Wobble](
 To use a spring you set the `fromPosition` and `toPosition` properties, which puts the spring into tension, and the end of the spring into motion. `currentPosition` will update of the course of the spring simulation, and the spring controller will trigger an element update so that the element can read `currentPosition` and render using it.
 
 ```ts
+import {SpringController2D} from '@lit-labs/motion/spring.js';
+
 @customElement('goo-element')
 export class GooElement extends LitElement {
   // Both the MouseController and SpringController2D will
   // trigger a render when the mouse moves or the spring updates
-  _mouse = new MouseController(this);
-  _spring1 = new SpringController2D(this, fast);
+  #mouse = new MouseController(this);
+  #spring = new SpringController2D(this, fast);
 
   render() {
     // Set the spring to go to the mouse
-    this._spring1.toPosition = this._mouse.position;
+    this.#spring.toPosition = this.#mouse.position;
 
     // Position a div based on the current position of the spring.
     return html`
       <div
         class="b1"
-        style=${positionStyle(this._spring3.currentPosition)}
+        style=${positionStyle(this.#spring.currentPosition)}
       ></div>
     `;
   }
--- a/packages/labs/motion/package.json
+++ b/packages/labs/motion/package.json
@@ -21,6 +21,11 @@
       "types": "./development/index.d.ts",
       "development": "./development/index.js",
       "default": "./index.js"
+    },
+    "./spring.js": {
+      "types": "./development/spring.d.ts",
+      "development": "./development/spring.js",
+      "default": "./spring.js"
     }
   },
   "files": [
@@ -29,7 +34,8 @@
     "/index.{d.ts,d.ts.map,js,js.map}",
     "/animate.{d.ts,d.ts.map,js,js.map}",
     "/animate-controller.{d.ts,d.ts.map,js,js.map}",
-    "/position.{d.ts,d.ts.map,js,js.map}"
+    "/position.{d.ts,d.ts.map,js,js.map}",
+    "/spring.{d.ts,d.ts.map,js,js.map}"
   ],
   "scripts": {
     "start": "wireit",
--- a/packages/labs/motion/src/test/spring_test.ts
+++ b/packages/labs/motion/src/test/spring_test.ts
@@ -3,15 +3,16 @@
  * Copyright 2021 Google LLC
  * SPDX-License-Identifier: BSD-3-Clause
  */
-import {LitElement} from 'lit';
-import {customElement} from 'lit/decorators.js';
-import {assert} from 'chai';
+
 import {
-  SpringConfig,
   Spring2DConfig,
+  SpringConfig,
   SpringController,
   SpringController2D,
-} from '../spring.js';
+} from '@lit-labs/motion/spring.js';
+import {assert} from 'chai';
+import {LitElement} from 'lit';
+import {customElement} from 'lit/decorators.js';
 
 suite('Spring', () => {
   let container: HTMLElement;


packages/labs/motion/src/demo/styles.css.ts
```diff
@@ -0,0 +1,83 @@
...
+    left: 50px;
+    width: 50px;
+    height: 50px;
+  }
+
+  .hooks-filter {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    filter: url('#goo');
```
AndrewJakubowicz, line 80: Nice!

packages/labs/motion/src/spring.ts
```diff
@@ -0,0 +1,175 @@
...
+  private _host: ReactiveControllerHost & HTMLElement;
+
+  protected _spring: Spring;
+  private _toValue: number;
+
+  constructor(
+    host: ReactiveControllerHost & HTMLElement,
+    options?: SpringConfig
+  ) {
+    (this._host = host).addController(this);
```
AndrewJakubowicz, line 29: Does this do the assignment followed by the `addController`? Neat!

packages/labs/motion/src/spring.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  get toValue() {
+    return this._toValue;
+  }
+
+  set toValue(toValue: number) {
+    this._toValue = toValue;
+    this._spring.updateConfig({toValue});
+    if ((this._host as HTMLElement).isConnected) {
```
AndrewJakubowicz, line 72: Does this need the type assertion since it has type: `ReactiveControllerHost & HTMLElement`?

packages/labs/motion/src/spring.ts
```diff
@@ -0,0 +1,175 @@
...
+      toValue: this._toPosition.x,
+    });
+    this._ySpring = new SpringController(host, {
+      ...options,
+      fromValue: options?.fromPosition?.y,
+      toValue: this._toPosition.y,
+    });
+  }
+
+  get currentPosition() {
```
AndrewJakubowicz, line 124: Nit: Maybe an explicit return type, otherwise pretty self documenting.

packages/labs/motion/src/spring.ts
```diff
@@ -0,0 +1,175 @@
...
+   */
+  get isAtRest(): boolean {
+    return this._xSpring.isAtRest && this._ySpring.isAtRest;
+  }
+
+  /**
+   * Whether or not the spring is currently emitting values.
+   *
+   * Note: this is distinct from whether or not it is at rest.
+   * See also `isAtRest`.
```
AndrewJakubowicz, line 150: Nit: JsDoc reference to `isAtRest` for API docs link. Doesn't need to be acted upon, but I think `[[isAtRest]]` will add a reference.

Ref: https://typedoc.org/guides/doccomments/#symbol-references

packages/labs/motion/src/test/spring_test.ts
```diff
@@ -0,0 +1,140 @@
...
+        this.spring = new SpringController2D(this, options);
+      }
+    }
+
+    test('spring is initially stopped, and initial params', async () => {
+      const el = new SpringController2DTest({
+        fromPosition: {x: 100, y: 100},
+        toPosition: {x: 50, y: 50},
+      });
+      // This one isn't what you might expect, but is what Wobble returns
```
AndrewJakubowicz, line 95: Why? Isn't the spring in tension?

packages/labs/motion/src/test/spring_test.ts
```diff
@@ -0,0 +1,140 @@
...
+import {customElement} from 'lit/decorators.js';
+import {assert} from '@esm-bundle/chai';
+import {
+  SpringConfig,
+  Spring2DConfig,
+  SpringController,
+  SpringController2D,
+} from '../spring.js';
+
+suite('Spring', () => {
```
AndrewJakubowicz, line 16: Maybe not required, but what happens if the component is connected. The spring runs for a frame. Then the component is disconnected (calling `stop()`). Then what happens when the component is reconnected? Does the spring continue from where it left off?

AndrewJakubowicz (approved):
Missing domain knowledge so did a syntax and manual test code review.

The demo is so amazing. Otherwise very minor nits and clarifications mostly for my own understanding. This is great!

From an API standpoint my only question is around integration with `animate()` directive. I ask this as an alternative to needing to supply a `styleMap` directive for the spring. Non blocking.

Great stuff!

packages/labs/motion/README.md
```diff
@@ -110,6 +110,45 @@ in the host element: `isPlaying` returns true if any `animate`'s are
...
+}
+
+const fast = {
+  stiffness: 1200,
+  damping: 400,
+};
+
+const positionStyle = ({x, y}: Position2D) =>
+  styleMap({
+    transform: `translate3d(${x}px,${y}px,0) translate3d(-50%,-50%,0)`,
```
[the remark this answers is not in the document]
justinfagnani: Animate sets up a CSS animation, which can't simulate a spring and doesn't create new JS values or callbacks to read them from. The Spring controller is a separate animation technique where the values are computed in JS, and rendered to CSS every frame. They aren't particularly compatible.

packages/labs/motion/package.json
```diff
@@ -61,7 +62,9 @@
     "typescript": "^4.3.5"
   },
   "dependencies": {
-    "lit": "^2.0.0"
+    "@lit/reactive-element": "^1.0.1",
+    "lit": "^2.0.0",
```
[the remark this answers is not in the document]
justinfagnani: The thought was that SpringController only needs ReactiveElement, though as a package dependency this doesn't help since we're already depending on `lit`.

packages/labs/motion/src/demo/spring2d.ts
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  constructor(host: ReactiveControllerHost) {
+    this.host = host;
+    host.addController(this);
+    window.addEventListener('mousemove', this._onMouseMove);
+  }
+
+  hostConnected() {
+    window.addEventListener('mousemove', this._onMouseMove);
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/labs/motion/src/spring.ts
justinfagnani: fixed

packages/labs/motion/src/spring.ts
justinfagnani: Replaced with `{@link}`

packages/labs/motion/src/test/spring_test.ts
justinfagnani: ah, true

packages/labs/motion/src/test/spring_test.ts
justinfagnani: yes, it'll continue where it left off, since `start()` is called. I think this a simple choice. There may a way to advance time so the animation acts like it was running, but I don't think that's that useful.

justinfagnani:
@AndrewJakubowicz thanks for the review from so long ago!

I addressed you comments and then rebased onto the 3.0 branch.

I have one important change I want to make to the 2D controller before I think this is mergable.

bencbradshaw:
Very cool controller and demo! Code looks great. Documentation is thorough enough that I could confidently use the controller.

bencbradshaw (approved):

packages/labs/motion/README.md
```diff
@@ -110,6 +110,69 @@ in the host element: `isPlaying` returns true if any `animate`'s are
...
+| `mass`                      | `number`  | The mass of the object attached to the end of the spring. Defaults to `1`.                                            |
+| `initialVelocity`           | `number`  | The initial velocity (in units/ms) of the object attached to the spring. Defaults to `0`.                             |
+| `allowsOverdamping`         | `boolean` | Whether or not the spring allows "overdamping" (a damping ratio > 1). Defaults to `false`.                            |
+| `overshootClamping`         | `boolean` | False when overshooting is allowed, true when it is not. Defaults to `false`.                                         |
+| `restVelocityThreshold`     | `number`  | When spring's velocity is below restVelocityThreshold, it is at rest. Defaults to `.001`.                             |
+| `restDisplacementThreshold` | `number`  | When the spring's displacement (current value) is below restDisplacementThreshold, it is at rest. Defaults to `.001`. |
+
+The SpringController2D constructor's `Spring2DConfig` has the same properties except `fromValue` and `toValue`, which are replaces with 2D equivalents:
```
AndrewJakubowicz, line 169:
```suggestion
The SpringController2D constructor's `Spring2DConfig` has the same properties except `fromValue` and `toValue`, which are replaced with 2D equivalents:
```

packages/labs/motion/src/demo/styles.css.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import {css} from 'lit';
+
+export const styles = css`
+  :host {
+    font-family: Arial, 'sans serif';
```
AndrewJakubowicz, line 10: No text is used in the spring demo – can optionally remove.

packages/labs/motion/demo/spring2d.html
```diff
@@ -0,0 +1,39 @@
...
+    <script src="../development/demo/spring2d.js" type="module"></script>
+    <style>
+      html,
+      body,
+      goo-element {
+        width: 100%;
+        height: 100%;
+        margin: 0;
+        padding: 0;
+        background-color: white;
```
AndrewJakubowicz, line 24: background-color can optionally be removed, it is handled in the demo by the full screen element.

packages/labs/motion/src/demo/styles.css.ts
```diff
@@ -0,0 +1,83 @@
...
+    display: block;
+  }
+  .hooks-main {
+    width: 100%;
+    height: 100%;
+    background: lightblue;
+    overflow: hidden;
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
AndrewJakubowicz, line 20: This can be deleted because the elements within this container end up being absolutely positioned.

packages/labs/motion/src/demo/styles.css.ts
AndrewJakubowicz, line 1: Optionally, you can copy paste these simplified styles:

Removed duplicates. Used `inset: 0px` instead of width/height. Removed font. Tested in the demo.

```ts
/**
 * @license
 * Copyright 2021 Google LLC
 */
import {css} from 'lit';

export const styles = css`
  :host {
    display: block;
  }

  .hooks-main {
    width: 100%;
    height: 100%;
    background: lightblue;
    overflow: hidden;
  }

  .hooks-filter {
    position: absolute;
    inset: 0px;
    filter: url('#goo');
    overflow: hidden;
  }

  .b1,
  .b2,
  .b3 {
    position: absolute;
    will-change: transform;
    border-radius: 50%;
    background: lightcoral;
    box-shadow: 10px 10px 5px 0px rgba(0, 0, 0, 0.75);
    opacity: 0.6;
  }

  .b1 {
    width: 170px;
    height: 170px;
  }

  .b2 {
    width: 350px;
    height: 350px;
  }

  .b3 {
    width: 200px;
    height: 200px;
  }

  .b1::after,
  .b2::after,
  .b3::after {
    content: '';
    position: absolute;
    top: 50px;
    left: 50px;
    width: 50px;
    height: 50px;
    border-radius: 50%;
    background: rgba(255, 255, 255, 0.8);
  }

  .b2::after {
    top: 70px;
    left: 70px;
    width: 70px;
    height: 70px;
  }
`;

```

packages/labs/motion/src/spring.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  get currentPosition(): Position2D {
+    return {x: this._xSpring.currentValue, y: this._ySpring.currentValue};
+  }
+
+  /**
+   * The spring's current velocity in units / ms.
+   */
+  // TODO(justinfagnani): should velocity be a vector?
+  get currentVelocity(): number {
```
AndrewJakubowicz, line 131: `velocity` implies vector. The scalar equivalent would be `speed`.
Should the name of this accessor be changed to `get currentSpeed`?

AndrewJakubowicz (approved):
Nice! Proposed some simplifications to the demo styles.

Also I think I'd prefer we use the scalar name `speed` instead of `velocity` when in 2d since I think that's what is currently being calculated.

packages/labs/motion/src/demo/spring2d.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  override render() {
+    // This is the chain of updates starting with the current mouse position and
+    // flowing through each spring
+    this._spring1.toPosition = this._mouse.position;
+    this._spring2.toPosition = this._spring1.currentPosition;
+    this._spring3.toPosition = this._spring2.currentPosition;
+
+    return html`
+      <svg style="position: absolute; width: 0; height: 0">
```
AndrewJakubowicz, line 73: This can be simplified to:

```
<svg style="display: none;">
```

PR #2267, opened 2021-10-27
fix: make willUpdate lifecycle hook protected

### LINKED ISSUE #2266: [reactive-element] Make `willUpdate` lifecycle method protected
# Description

Unlike other lifecycle hooks, `willUpdate` is public which confusingly pollutes the API of the components.
Let's consider making it protected to align with other methods such as `updated`, `firstUpdated` etc.

web-padawan:
## What I did

Changed `willUpdate()` to be `protected`, same as related hooks: `update()`, `updated()` and `firstUpdated()`.

Fixes #2266 

Main question: should we consider this change safe for a patch release? It might break someone's code 🤔

### DIFF
--- a/.changeset/empty-cobras-lick.md
+++ b/.changeset/empty-cobras-lick.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Make `willUpdate` lifecycle hook protected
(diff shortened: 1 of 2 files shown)


justinfagnani:
I personally think this is small enough and shouldn't break correct code so we an consider it non-breaking.

justinfagnani (approved):

Issue #2271, opened 2021-10-28
[localize] Prefer `<x>` to `<ph>` in generated XLIFF files


aomarks:
I tried uploading @lit/localize generated XLIFF files to 3 localization services, and found that support for `<ph>` tags was lacking, while support for `<x>` tags was good.

The difference between the two tags in the XLIFF 1.2 spec is quite unclear to me, but following what actually works in real services seems is the obvious way to go.

- [`<ph>`](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#ph): Placeholder - The <ph> element is used to delimit a sequence of native stand-alone codes in the translation unit.
- [`<x>`](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#x): Generic placeholder - The <x/> element is used to replace any code of the original document.

## crowdin

`<ph>` works but isn't optimal. `<ph>` tags are automatically converted to `<x>` tags for display in the translation tool, and then restored to original `<ph>` tags when the results are downloaded. This works, but it means that there is very little context available to understand what the placeholders mean. If we use `<x>` tags from the start, then we can use the `equiv-text` attribute to display the HTML markup/code. It also seems that an `example` attribute will be preserved and displayed (even though that is not in the spec), which might be a good way to represent the `example` annotation that we'll be adding (https://github.com/lit/lit/issues/1890).

## phrase

`<ph>` tags are broken because they get escaped. `<x>` tags are preserved correctly.

Input:
```xml
<source>Hello <ph id="ph0">NAME</ph>, how are you?</source>
<source>Hello <x id="NAME" example="an example"/>, how are you?</source>
```

Output:
```xml
<target xml:lang="es">Hola &lt;ph id="ph0"&gt;NAME&lt;/ph&gt;, como estas?</target>
<target xml:lang="es">Hola ‪<x id="NAME" example="an example"/>‬, como estas?</target>
```

## lokalise

`<ph>` tags are broken because their opening/closing tags get stripped. `<x>` tags are preserved correctly.

Input:
```xml
<source>Hello <ph id="ph0">NAME</ph>, how are you?</source>
<source>Hello <x id="NAME" example="an example"/>, how are you?</source>
```

Output:
```xml
<target>Hola NAME, como estas?</target>
<target>Hola <x id="NAME" example="an example"/>, como estas?</target>
```

PR #2275, opened 2021-10-29
[localize] Add <x equiv-text> XLIFF placeholder style and use by default

aomarks:
See https://github.com/lit/lit/issues/2271 for more context.

XLIFF is the XML format we use to represent extracted templates/strings that need translation. XLIFF specifies multiple ways of encoding placeholders (for representing HTML markup and dynamic expressions). The differences according to the spec are a bit confusing:

- [`<ph>`](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#ph): "Placeholder - The <ph> element is used to delimit a sequence of native stand-alone codes in the translation unit."
- [`<x>`](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#x): "Generic placeholder - The <x/> element is used to replace any code of the original document."

Previously we were using `<ph>`, because the spec seemed to match what we need, and I was primed by XLB (Google's very similar format) which also uses `<ph>` tags for this purpose. However, I found that in practice translation tools seem to have much better support for the XLIFF `<x>` tag (I have tested crowdin, phrase, and lokalise). Additionally, `<x>` is the approach used by Angular (see https://angular.io/guide/i18n-example), so it is likely that translation tools/services have been already tested with Angular style message extraction.

This is possibly breaking because it changes the default from `<ph>` to `<x>`, but upgrading both source and translated messages to `<x>` will happen automatically when the user next runs `lit-localize extract`. We retain the ability to use `<ph>` tags by setting a new config file setting.

Fixes https://github.com/lit/lit/issues/2271

cc @fetherston fyi

### DIFF
--- a/packages/localize-tools/src/formatters/xliff.ts
+++ b/packages/localize-tools/src/formatters/xliff.ts
@@ -11,7 +11,7 @@ import type {Config} from '../types/config.js';
 import type {XliffConfig} from '../types/formatters.js';
 import type {Locale} from '../types/locale.js';
 import {Formatter} from './index.js';
-import {KnownError} from '../error.js';
+import {KnownError, unreachable} from '../error.js';
 import {
   Bundle,
   Message,
@@ -110,6 +110,15 @@ export class XliffFormatter implements Formatter {
           contents.push(child.nodeValue || '');
         } else if (
           child.nodeType === doc.ELEMENT_NODE &&
+          child.nodeName === 'x'
+        ) {
+          const phText = getNonEmptyAttributeOrThrow(
+            child as Element,
+            'equiv-text'
+          );
+          contents.push({untranslatable: phText});
+        } else if (
+          child.nodeType === doc.ELEMENT_NODE &&
           child.nodeName === 'ph'
         ) {
           const phText = child.childNodes[0];
@@ -118,7 +127,9 @@ export class XliffFormatter implements Formatter {
             !phText ||
             phText.nodeType !== doc.TEXT_NODE
           ) {
-            throw new KnownError(`Expected <ph> to have exactly one text node`);
+            throw new KnownError(
+              `Expected <${child.nodeName}> to have exactly one text node`
+            );
           }
           contents.push({untranslatable: phText.nodeValue || ''});
         } else {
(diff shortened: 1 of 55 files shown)


fetherston:
Looks like the message ids remain the same right?

aomarks:
> Looks like the message ids remain the same right?

Yep, this will have no effect on message IDs. If you upgrade and do nothing else, the next time you run `extract` the placeholders will change to the new syntax, both in the `<source>` and `<target>` elements. If that confuses some part of the localization pipeline for some reason, you can switch back to the old style with `placeholderStyle: "ph"`.

packages/localize-tools/src/formatters/xliff.ts
```diff
@@ -118,7 +127,9 @@ export class XliffFormatter implements Formatter {
             !phText ||
             phText.nodeType !== doc.TEXT_NODE
           ) {
-            throw new KnownError(`Expected <ph> to have exactly one text node`);
+            throw new KnownError(
+              `Expected <${child.nodeName}> to have exactly one text node`
```
rictic, line 131: Feel free to punt, but this seems like a good candidate for a diagnostic

packages/localize-tools/src/tests/e2e/extract-xliff-fresh-ph.test.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
rictic, line 3: Here and elsewhere: 2021

packages/localize-tools/testdata/build-runtime-xliff-ph/goldens/tsout/empty.txt
```diff
@@ -0,0 +1 @@
+git won't track an empty directory
```
rictic, line 1: (is this necessary? looks like there are files in this dir)

rictic (approved):

packages/localize-tools/src/formatters/xliff.ts
aomarks: We're not in a TypeScript context here -- do you mean something that displays similar to how a TypeScript diagnostic displays with a text snippet and underline etc.?

packages/localize-tools/src/tests/e2e/extract-xliff-fresh-ph.test.ts
aomarks: This is just a rename, so it's copyright date shouldn't change (right?). And the other new files are also clones.

packages/localize-tools/testdata/build-runtime-xliff-ph/goldens/tsout/empty.txt
aomarks: Done

PR #2279, opened 2021-11-02
docs: Make version banner more apparent for lit-html 1.x

e111077:
same treatment as lit/lit-element#1211


arthurevans (approved):
LGTM, ship it! :shipit:

PR #2283, opened 2021-11-04
Fix formatting and linting

justinfagnani:


.eslintignore
```diff
@@ -165,3 +165,4 @@ packages/labs/task/index.*
 packages/labs/task/task.*
 
 packages/localize/testdata/
+packages/localize/examples/
```
aomarks, line 168: why shouldn't we lint the examples?

PR #2286, opened 2021-11-09
[localize] Update READMEs to point at new full docs at lit.dev

aomarks:
Now that we have full docs at https://lit.dev/docs/localization/overview/ we don't need all of this detail in the READMEs.


packages/localize-tools/README.md
```diff
@@ -1,13 +1,40 @@
-# @lit/localize-tools
```
abdonrd, line 1: This is right?

aomarks: Oops, fixed, thanks!

justinfagnani (approved):

PR #2288, opened 2021-11-09
[ssr] Improve module loader and add a module cache that track imports.

justinfagnani:
This make the module loader a class so that it can hold instance state about the cache, current vmContextID. This way more than one loader could be instantiated, which would be important for hosting multiple apps. This is working towards being able to introspect and modify the module cache too.


packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
+/**
```
aomarks, line 1: Would be nice in future to do a rename commit separately. For this PR, GitHub doesn't understand that there was a move, so the diff is useless.

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+ *  - import.meta.url support
+ *  - Dynamic import() that functions the same as static imports
+ *
+ * There are some behaviors specific to lit-html. Mainly that imports of certain
+ * directives are redirected to Node/SSR compatible implementations.
+ */
+export class ModuleLoader {
+  private static _nextVmContextId = 0;
+
+  // TODO (justinfagnani): document the need for this
```
aomarks, line 55: Document now? :)

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+    this._context = vm.createContext(contextGlobal);
+  }
+
+  /**
+   * Imports a module given by `path` into a new VM context with `contextGlobal` as the
+   * global object.
+   *
+   * @param specifier
+   * @param referrer
+   * @param contextGlobal The object that will become the global, via vm.createContext
```
aomarks, line 82: `contextGlobal` parameter doesn't exist, and the other two `@param`s have no content. Just remove the jsdoc I guess.

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+   * Module instance, and maintains the module cache.
+   *
+   * Used directly by `importModule` and by the linker and dynamic import()
+   * support function.
+   */
+  private async _loadModule(
+    specifier: string,
+    referrer: string
+  ): Promise<ImportResult> {
+    if (/:\d+$/.test(referrer)) {
```
aomarks, line 113: Isn't `:0` a valid filename?

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+    if (moduleRecord !== undefined) {
+      return {
+        path: modulePath,
+        module: await moduleRecord.evaluated,
+      };
+    }
+
+    const modulePromise = (async () => {
+      const source = await fs.readFile(modulePath, 'utf-8');
+      // TODO: store and re-use cachedData:
```
aomarks, line 138: Comment still relevant? It is cached now, right? Or if there's another layer of caching we could do, maybe elaborate on the comment (there is nothing called `cachedData`).

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+        importModuleDynamically: this._importModuleDynamically,
+        context: this._context,
+        identifier: this._getIdentifier(modulePath),
+      });
+    })();
+
+    moduleRecord = {
+      path: modulePath,
+      imports: [],
+      evaluated: modulePromise!,
```
aomarks: Are the `!`s on `modulePromise` and `modulePath` around here needed? Kinda looks like they shouldn't be null/undefined already.

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+  private async _loadBuiltInModule(specifier: string) {
+    let moduleRecord = this.cache.get(specifier);
+    if (moduleRecord !== undefined) {
+      return {
+        path: specifier,
+        module: await moduleRecord.evaluated,
+      };
+    }
+    // Provide basic support for built-in modules (needed for node shims of
+    // DOM API's like fetch)
```
aomarks, line 170: nit: APIs

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+      }
+      const referrer = identifier.split(/:\d+$/)[0];
+      return (await this._loadModule(specifier, referrer)).module;
+    } else {
+      throw new Error('unknown module source');
+    }
+  };
+
+  private _getIdentifier(modulePath: string) {
+    return 'file://' + modulePath + `:${this._vmContextId}`;
```
aomarks:
```js
return `file://${modulePath}:${this._vmContextId}`;
```

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  private _getBuiltInIdentifier(specifier: string) {
+    return specifier + `:${this._vmContextId}`;
```
aomarks, line 232:
```js
return `${specifier}:${this._vmContextId}`;
```

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+}
+
+/**
+ * Resolves specifiers using web-ish Node module resolution. Web-compatible
+ * full URLs are passed through unmodified. Relative and absolute URLs
+ * (starting in `/`, `./`, `../`) are resolved relative to `referrer`. "Bare"
+ * module specifiers are resolved with the 'resolve' package.
+ *
+ * This replaces some lit-html modules with SSR compatible equivalents. This is
```
aomarks, line 242: Nit: s/lit-html/Lit

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+ * currently hard-coded, but should instead be done with a configuration object.
+ */
+export const resolveSpecifier = (specifier: string, referrer: string): URL => {
```
aomarks: Does this need to be exported?

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+    // those will be absolute to the file system.
+    return new URL(specifier);
+  } catch (e) {
+    if (referrer === undefined) {
+      throw new Error('referrer is undefined');
+    }
+    if (
+      specifierMatches(specifier, 'lit') ||
+      specifierMatches(specifier, 'lit-html') ||
+      specifierMatches(specifier, 'lit-element')
```
aomarks, line 263: add @lit/reactive-element?

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+    ) {
+      // Override where we resolve lit-html from so that we always resolve to
+      // a single version of lit-html.
+      referrer = import.meta.url;
```
aomarks, line 267: Does this force one version because it causes us to resolve relative to the top-level node_modules directory instead of a nested one?

What if there are two conflicting versions of Lit?

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: Ah shoot, sorry

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+    const referencingModulePath = new URL(referrer).pathname;
+    const modulePath = resolve.sync(specifier, {
+      basedir: path.dirname(referencingModulePath),
+      moduleDirectory: ['node_modules'],
+      extensions: ['.js'],
+      // Some packages use a non-standard alternative to the "main" field
+      // in their package.json to differentiate their ES module version.
+      packageFilter: (packageJson: PackageJSON) => {
+        packageJson.main =
+          packageJson.module ?? packageJson['jsnext:main'] ?? packageJson.main!;
```
aomarks, line 277: I think `?? index.js` should be the final fallback?

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -0,0 +1,279 @@
...
+      specifierMatches(specifier, 'lit') ||
+      specifierMatches(specifier, 'lit-html') ||
+      specifierMatches(specifier, 'lit-element')
+    ) {
+      // Override where we resolve lit-html from so that we always resolve to
+      // a single version of lit-html.
+      referrer = import.meta.url;
+    }
+    const referencingModulePath = new URL(referrer).pathname;
+    const modulePath = resolve.sync(specifier, {
```
aomarks: Will this handle package exports?

.eslintignore
```diff
@@ -149,9 +149,13 @@ packages/labs/scoped-registry-mixin/scoped-registry-mixin.*
 packages/labs/ssr/demo/
 packages/labs/ssr/lib/
 packages/labs/ssr/node_modules/
-packages/labs/ssr/test/
+packages/labs/ssr/test/**/*.js
```
aomarks, line 152: Are these changes needed? Maybe they belong in another PR that adds tests? Same for `.prettierignore`.

aomarks (changes requested):

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: I can remove this. It was useful for flushing out bugs...

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: This is referring to Node's vm module cachedData, which should speed up repeated module creation in cases where we create fresh vm contexts. It's not related to our module cache. I updated the link and it seems like they added cachedData for modules now, so we can do this.

packages/labs/ssr/src/lib/module-loader.ts
aomarks: I guess I was wondering about the prefixing scheme overall, although I suppose it doesn't actually require that there are never filenames that match the prefix pattern.

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: Yes.

We have a bit of work to do here to test behavior with multiple versions of Lit. Since Lit 2.0 we don't require instanceof anymore, so we might not even need this. It wouldn't be used in the "global" mode (where we path Node's global and don't load modules like this) anyway, but I don't want to change this quite yet.

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: `resolve` should already implement that then. Here we're just adding in some non-standard behavior to look at the `module` and `jsnext:main` fields.

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: Oops, those are from before a refactor that changed them to `const`.

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: I'm using it into another project

packages/labs/ssr/src/lib/module-loader.ts
justinfagnani: I don't know... I don't think this package has been updated in a while.

.eslintignore
justinfagnani: This PR does add tests. Some of your comments were on older version files though, so maybe you didn't see them?

justinfagnani:
@aomarks PTAL

.eslintignore
aomarks: Oops yeah I missed them somehow.

aomarks (approved):

PR #2294, opened 2021-11-12
[ssr] Add customElementRendered callback

justinfagnani:
This allows callers to see what tags are rendered as they are streamed. An important use case is to enable on-demand bundling of just the modules required to hydrate the page or a subset of the page.

### DIFF
--- a/.changeset/tasty-zebras-double.md
+++ b/.changeset/tasty-zebras-double.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': minor
+---
+
+Add customElementRendered callback to RenderInfo so that callers can konw what elements were rendered.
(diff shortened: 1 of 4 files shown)


kevinpschaaf (approved):

kevinpschaaf:
Chromedriver needs a bump for benchmarks it seems.

.changeset/tasty-zebras-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': minor
+---
+
+Add customElementRendered callback to RenderInfo so that callers can konw what elements were rendered.
```
arthurevans, line 5:
```suggestion
Add customElementRendered callback to RenderInfo so that callers can know what elements were rendered.
```

PR #2300, opened 2021-11-12
[starter-kits] Fix starter kits so `npm run serve` isn't `npm run docs:serve`

AndrewJakubowicz:
### Context

Currently you cannot serve your component from the starter kit. You always get the generated docs.

This change makes it so when you run `npm run serve`, you get your component. Docs are still available via the `nom run docs:serve` command.

Currently: `npm run serve` and `npm run docs:serve` give you the same output.

### Why

Because it's confusing and has resulted in the following efforts to fix:
 - https://github.com/lit/lit-element-starter-ts/issues/58
   - In the linked video: https://www.youtube.com/watch?v=86tet76u0uk&ab_channel=MiklosNemeth , they are unable to figure out how to serve their component. This is fixed by this change.
- This PR on the original repo: https://github.com/lit/lit-element-starter-ts/pull/57 fixes the serve command explicitly.

### Repro bug

1. Make a new template based on the starter kit.
1. `npm i`
1. `npm run build`
1. `npm run serve` -> get docs page.
1. Follow instructions in readme and navigate to `/dev/index.html` and get a 404.

### Impact

Fixes serve command so it matches what expectation is set in the readme. Following links in readme, i.e. `/dev/index.html` return your component instead of 200.

### DIFF
--- a/.changeset/metal-news-sin.md
+++ b/.changeset/metal-news-sin.md
@@ -1,0 +1,6 @@
+---
+'@lit/lit-starter-js': patch
+'@lit/lit-starter-ts': patch
+---
+
+Fix starter kits so `npm run serve` serves the root directory, and add a link to the `/dev/index.html` component example from `/`.
--- a/packages/lit-starter-js/index.html
+++ b/packages/lit-starter-js/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Lit Starter Kit</title>
+  </head>
+  <body>
+    <a href="/dev/index.html">Component Demo</a>
+  </body>
+</html>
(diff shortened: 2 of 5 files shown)


e111077 (approved):

justinfagnani (approved):

AndrewJakubowicz:
This required a little more work mostly to guide folks in the right direction.

Previous behavior. `npm run serve` served the docs directory. The result is that you can never get to your /dev/index.html demo. You are instead siloed to docs.

New behavior: `npm run serve` serves the root directory. This allows for the `/dev/index.html` route to be available and work. And it correctly shows an incrementor with the slotted content.

However the pitfall of the new behavior is that web/dev-server prompts you to navigate to localhost:8000/ which is a 404. As very light guidance, I've added an html file that links to the demo, rather than having a 404 experience.

@justinfagnani PTAL at new changes. I manually tested all the build rules. Thank you!

justinfagnani:
Looks good. Still need a changeset though.

AndrewJakubowicz:
Done. Thank you.

justinfagnani (approved):

e111077:
Can you please also follow up on the TS and JS issues. There seem to be other issues that seem to be related.

AndrewJakubowicz:
I followed up on all the issues that seem related and will follow up again once this is released into those repos.

Issue #2301, opened 2021-11-13
[labs/task] `render` does not return a useful type


mohe2015:
# Description

When using `render` of a `@lit-labs/task` `Task` the method is typed as having a return type of `unknown`.

# Steps to Reproduce

1. Write this code

```ts
import type { Task } from "@lit-labs/task/task.js";

const task: Task = undefined!;

// Type 'unknown' is not assignable to type 'number'.ts(2322)
const test: number = task.render({
  complete: () => 1
})
```

2. The code does not typecheck because it tries to assign type `unknown` to `number`. (Adding the three other handlers does not help as you can see in how the source code is typed)

mohe2015:
Sorry for bumping but I think this is one of the main annoyances and from my attempt you can see that the current thing is really hard to type

sorvell:
We'll likely want to introduce a generic that's set when the task is defined and piped to the render return type.

e111077:
I think this should be fixable in TS, but the underlying issue here is that `Task.render` is not intended to be used outside of rendering. With the latest release, `Task.value` should not change value needlessly to `undefined` anymore and be more "stable". Does `Task.value` fit your use case?

additionally, pedantic issue, in your case, your typing is wrong – it should be:

```ts
const test: number|undefined = task.render({
  complete: () => 1
})
```

taylor-vann:
I have an existing PR that addresses this issue. It's been a while though. I'll update and post.

https://github.com/lit/lit/pull/2821

justinfagnani:
I'm not sure there's anything very useful to do here. We could type `render()` as returning the union of the renderer methods passed to it. That's about the most accurate thing we can do, but I'm not sure what value that has.

PR #2302, opened 2021-11-13
[@lit-labs/task] Type return value of `render`

mohe2015:
Fixes https://github.com/lit/lit/issues/2301

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -46,6 +46,10 @@ export interface TaskConfig<T extends unknown[], R> {
   autoRun?: boolean;
 }
 
+type GetReturnType<Type> = Type extends (...args: any) => infer V
+  ? V
+  : undefined;
+
 // TODO(sorvell): Some issues:
 // 1. When task is triggered in `updated`, this generates a ReactiveElement
 // warning that the update was triggered in response to an update.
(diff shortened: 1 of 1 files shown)


mohe2015:
I signed the CLA! (I hope)

packages/labs/task/src/task.ts
```diff
@@ -30,11 +30,11 @@ export const initialState = Symbol();
 
 export type TaskStatus = typeof TaskStatus[keyof typeof TaskStatus];
 
-export type StatusRenderer<R> = {
-  initial?: () => unknown;
-  pending?: () => unknown;
-  complete?: (value: R) => unknown;
-  error?: (error: unknown) => unknown;
+export type StatusRenderer<R, Q> = {
```
justinfagnani, line 33: nit: `R` here stands for "result" I think. Rather than use the next letter in the alphabet, let's do `V` for the render value.

justinfagnani: It'd be very reasonable for the render callbacks to return different types. I'd default the type parameter to `unknown`: `<R, V = unknown>`

packages/labs/task/src/task.ts
```diff
@@ -188,9 +188,6 @@ export class Task<T extends [...unknown[]] = any, R = any> {
         return renderer.complete?.(this.value!);
       case TaskStatus.ERROR:
         return renderer.error?.(this.error);
-      default:
```
justinfagnani, line 191: Let's not remove this

justinfagnani (changes requested):

justinfagnani:
Thanks for the PR @mohe2015! Left a few comments

packages/labs/task/src/task.ts
mohe2015: It didn't typecheck for me with that part (and also threw errors with my change when I removed a case) but maybe you can help me fix that if that would be the case now again.

On that note `npm run lint` and `npm run format` throws 100s of errors for me. Does the code need to be unstaged or so?

packages/labs/task/src/task.ts
mohe2015: > It'd be very reasonable for the render callbacks to return different types. I'd default the type parameter to `unknown`: `<R, V = unknown>`

To me it seems like I still get typecheck errors now even with a default also added in the method directly. Maybe that's the reason it's typed as `unknown`. As an alternative I was thinking about `V1,V2,V3,V4` and then the resulting one is the union of them.

mohe2015:
If anyone can tell me how to do this in a clean way (feel free to create your own PR) I would love to know. I think such extremely strong typings can create way more powerful+reliable software.

mohe2015:
I tried again pretty long and still failed to find a way without casting. I believe that typescript is simply not strong enough. The optional methods make it impressively hard to type it correctly.

How realistic is it that something similar to the current change would be merged?

mohe2015:
Also is there some way to test this with my own project? I tried in the past but simply linking didn't seem to work. Or is that supposed to work? I think that would be really useful to add to the top-level README.

PR #2307, opened 2021-11-19
Add an additional template object check

### LINKED ISSUE #2306: Strong brand checks on template strings arrays
We should make TemplateResults harder to forge.

rictic:
This makes it much harder for untrusted values in a databinding to masquerade as internal lit types. Specifically, this ensures that a template result value can't be obtained from JSON.parse.

Fixes #2306

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -653,6 +653,22 @@ const getTemplateHtml = (
   const htmlResult: string | TrustedHTML =
     html + (strings[l] || '<?>') + (type === SVG_RESULT ? '</svg>' : '');
 
+  // A security check to prevent spoofing of Lit template results.
+  // In the future, we may be able to replace this with Array.isTemplateObject,
+  // though we might need to make that check inside of the html and svg
+  // functions, because precompiled templates don't come in as
+  // TemplateStringArray objects.
+  if (!Array.isArray(strings) || !strings.hasOwnProperty('raw')) {
+    let message = 'invalid template strings array';
+    if (DEV_MODE) {
+      message =
+        `Internal Error: expected template strings to be an array ` +
+        `with a 'raw' field. Please file a bug at ` +
+        `https://github.com/lit/lit/issues/new?template=bug_report.md ` +
+        `and include information about your build tooling, if any.`;
+    }
+    throw new Error(message);
+  }
   // Returned as an array for terseness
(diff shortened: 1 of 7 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -653,6 +653,22 @@ const getTemplateHtml = (
   const htmlResult: string | TrustedHTML =
     html + (strings[l] || '<?>') + (type === SVG_RESULT ? '</svg>' : '');
 
+  // A security check to prevent spoofing of Lit template results.
+  // In the future, we may be able to replace this with Array.isTemplateObject,
+  // though we might need to make that check inside of the html and svg
+  // functions, because precompiled templates don't come in as
+  // TemplateStringArray objects.
+  if (!Array.isArray(strings) || !strings.hasOwnProperty('raw')) {
```
justinfagnani, line 661: There are two other potentially reasonable places to do this:
  1. In the Template constructor, which will be called the same number of times as getTemplateHtml
  2. In `ChildPart._$setValue`, which is called more often but where our actual runtime duck-typing lives and might indicate other type-checks that need to be tightened. The nice this about this location is that we wouldn't need to throw, we could just fall through to generic object handling.

Either of those would leave getTemplateHtml open for use with arrays of strings for other utility purposes. I think it's worth ruling them out at least before committing.

rictic: > Either of those would leave getTemplateHtml open for use with arrays of strings for other utility purposes

getTemplateHtml is a core part of the security boundary, and we should think very carefully before calling it with something other than a TemplateStringsArray. 

For `ChildPart._$setValue`, I think it's fine to continue to disallow directly rendering objects with '_$lit...$' keys.

The thing I like about doing it here is that, by default it's good to co-locate security checks with blessing the value as secure with policy.createHtml

If we change our minds later, I don't think it'd be a disruptive change to move this check elsewhere.

justinfagnani: Seem reasonable. I do wonder if the policy usage wouldn't also be good to put in the Template constructor, but it's effectively the same except for outside callers of `getTemplateHtml`.

justinfagnani (changes requested):
Looks good, but we need a test that an error is thrown for a JSON-produced template result.

justinfagnani (approved):

PR #2309, opened 2021-11-21
Add basic event support

justinfagnani:
This is a start of what it would look like to add event support to SSR. It uses Node's `Event` and `EventTarget` classes and extends `EventTarget` to have a parent that it can dispatch to to implement bubbling. `stopPropagation()`, etc., should work, but aren't tested yet.

The big open questions are:
* Can we have an event path of just the custom elements without materializing the built-ins that would be on the path with a full DOM structure? This seems ok if the only event patterns we support are for cross-custom-element communication, ie context.
* Where do elements install event listeners? I added a call to `connectedCallback()` in here, but that only doesn't throw because none of the test elements do anything with the DOM in `connectedCallback()`. That wouldn't be viable in general.
* Do we need to implement the capture phase?
* How could we implement the proper composed path? We would need some concept of slotted elements being in the custom element instance stack.

### DIFF
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -37,7 +37,42 @@ export const getWindow = ({
     return attrs;
   };
 
-  class Element {}
+  /**
+   * Extends EventTarget to have a parent reference and adds event propagation.
+   */
+  class EventTargetWithParent extends EventTarget {
+    __eventTargetParent: EventTarget | undefined;
+
+    override dispatchEvent(event: Event): boolean {
+      // TODO (justinfagnani): This doesn't implement capture at all.
+      // To implement capture we'd need to patch addEventListener to track the
+      // capturing listeners separately, then call into a capture method here
+      // which would supercall before processing any capturing listeners.
+
+      // First dispatch the event on this instance
+      let canceled = super.dispatchEvent(event);
+
+      // Then conditionally bubble up. cancelBubble is true if a handler
+      // on this instance called event.stopPropagation()
+      if (!event.cancelBubble && this.__eventTargetParent !== undefined) {
+        canceled &&= this.__eventTargetParent.dispatchEvent(event);
+      }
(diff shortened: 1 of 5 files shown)


kevinpschaaf:
Wonder if there's an alternative that would be more compatible with bring-your-own-shim and keeping the `ElementRenderer` just an interface where instead of requiring the instance to hook up the fake parent, we instead:
* Add `ElementRenderer`:`dispatchEvent` (which in ours will just call `instance.dispatchEvent()`
* Add a _patch_ of `dispatchEvent` on whatever `Element.prototype` exists (either our dom shim, which would just have `Element` extend `EventTarget`, or a real DOM shim like jsdom, which would have a more correct implementation)
* Our patch would super to the existing `dispatchEvent`, and then if it should bubble AND there's no `this.parentNode` (which is always the case in our DOM shim, and would be the case when using jsdom and you go up and hit a LitElementRenderer instance), then look in the (somehow-made-accessible) open custom elements stack and dispatch to the next element up.

It seems like that might avoid punting on `ElementRenderer` just yet.

We of course (maybe as part of this), should try running with a full DOM shim to make sure even the basic stuff works there, and specifically that event dispatch would work.

EisenbergEffect:
Pinging @nicholasrice from FAST to take a look. My cursory glance leads to me to think this will work for us as well.

nicholasrice:
I am definitely in support of exposing dispatchEvent. The largest use-case for us will be context and dependency-injection APIs. Our DI infrastructure already is implemented using events on the client, and we have a context-like feature that could be implemented on the server using events. We also have plans to implement general-purpose context APIs, which would obviously benefit as well.


> * Can we have an event path of just the custom elements without materializing the built-ins that would be on the path with a > full DOM structure? This seems ok if the only event patterns we support are for cross-custom-element communication, ie context. 

I think this constraint is suitable for our implementations and the majority of our components. It seems in-line with other code we have that no-ops operations that would provide references to native elements, such as ref directives, slotted element inspection, etc.

Issue #2296, opened 2021-11-12
[infra] Lit logo image that works in dark mode / on other backgrounds


TarunavBA:
# Description

The Lit Logo is not visible properly in GitHub dark mode. 

# Steps to Reproduce

1. Go to GitHub.com/lit/lit
2. Change the GitHub theme to dark
2. See that the lit logo is not visible.

## Expected Results

The logo should be both dark and light mode compatible

## Actual Results

The logo is not visible

## Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

thescientist13:
might be helpful to include some screenshots in the description to make it a bit easier to understand / see the issue.

TarunavBA:
Sure @thescientist13 .

TarunavBA:
Here it is :
![Screenshot_2021-11-13-10-22-07-708](https://user-images.githubusercontent.com/89698728/141606108-a2b0b6fb-a37b-4143-960f-d171ba0c18aa.jpeg)

TarunavBA:
Any updates on this @thescientist13 ?

thescientist13:
I'm not really a graphic designer nor am I an owner of the Lit repo, so my ability to help is very limited other than to help triage and assist from the outside. 😅 

That said, your screenshot is helpful as it it looks like it is the text aspect to the image that doesn't render properly in dark mode, naturally, given the text is dark. 

I saw Adam Wathan (creator of Tailwind) had a similar [question / challenge](https://twitter.com/adamwathan/status/1438096197672132617).  I'm not sure what he means by sticker, but effectively he added a white border to the text so that in light mode, it looks the same, but in dark mode, the text has a stroke around it to keep it from getting lost in dark mode.


<img width="616" alt="Screen Shot 2021-11-16 at 10 00 03 AM" src="https://user-images.githubusercontent.com/895923/142009446-42091b0e-9b9e-4820-9ef6-3d0501e1de3a.png">

<img width="615" alt="Screen Shot 2021-11-16 at 9 58 47 AM" src="https://user-images.githubusercontent.com/895923/142009476-cb079727-b256-4bec-809a-52566db744ca.png">

----

Maybe that would be helpful for the Lit team to consdier?

TarunavBA:
Hey @justinfagnani I know a good friend of mine who could help us. Are you ok if he helps?

justinfagnani:
[...]

PR #2314, opened 2021-11-24
Add dark theme version of the Lit logo for GitHub READMEs

aomarks:
As of just now, GitHub supports showing a different image for light vs dark theme: https://twitter.com/ashtom/status/1463554157932023808

This adds a new dark theme version of the logo, which is the same except that the Lit text is white, and uses the new `#gh-light-mode-only` and `#gh-dark-mode-only` fragments to switch between them.

This also updates the logo to the new more compact version of the logo SVG from https://github.com/lit/lit.dev/pull/578.

Fixes https://github.com/lit/lit/issues/2296


aomarks:
@justinfagnani: This is going to churn the `lit` package again, since I updated the logo to the new more compact version. The original reason we added the logo to the NPM package was so that it would show up on npmjs.com -- but then we learned that npmjs.com doesn't actually read those images from relative paths, it finds images from GitHub automatically. **So should we instead remove the logo from the NPM package, to save the bytes?**

aomarks:
Hmm, I guess a downside of this is that the version on npmjs.com is going to show both images, since it doesn't know about the special fragments.

I guess we could run a script that removes the dark mode version before we publish to NPM.

aomarks:
> Hmm, I guess a downside of this is that the version on npmjs.com is going to show both images, since it doesn't know about the special fragments.

Sounds like NPM and VS Code will both support this too eventually https://twitter.com/ashtom/status/1463567599246065666

justinfagnani:
Until npm supports this, could we add a white stroke/border/background to the logo image to make it work in both?

graynorton:
I consulted Ocupop on this topic before the holidays and they did some exploration, eventually concluding that they weren't happy with the white stroke / border solution from a visual brand integrity point of view.

Their recommendation was to use the dark-mode-compatible version of the logo if we have the ability to switch, and to use a white rectangular background if we don't:

https://gyazo.com/18b45a5c99da57a77235c2a7337f5fd3

justinfagnani:
@aomarks do we know if npm supports the dark mode image convention yet?

augustjk:
Superseded by https://github.com/lit/lit/pull/3457

Issue #2292, opened 2021-11-11
[lit/reactive-element] Add @queryAssignedElements() decorator


AndrewJakubowicz:
## Description

Tracking issue to propose adding `@queryAssignedElements(slotName?: string, flatten?: boolean)` which can be used as a declarative alternative to [`slot.assignedElements(options?: {flatten: true})`](https://html.spec.whatwg.org/multipage/scripting.html#dom-slot-assignedelements-dev).

### Why they want it

We currently support [`@queryAssignedNodes()`](https://lit.dev/docs/api/decorators/#queryAssignedNodes), but don't have a queryAssignedElements alternative.

Currently it is possible to work around the `flatten: false` case of queryAssignedElements by using `@queryAssignedNodes('', true, '*')`.

The impact of having this decorator is a declarative and readable way to work with slotted elements.

### Functional description of task/subtask

If we wanted to implement this so it was similar to the existing `@queryAssignedNodes`, we could implement this API: 

 - `@queryAssignedElements(slotName?: string, flatten?: boolean, selector?: string)`

We could also implement this decorator with an API that's close to the spec, with the slotName passed in as the first argument:

 - `@queryAssignedElements(slotName?: string, options?: {flatten: boolean})`

### Acceptance criteria

A declarative `@queryAssignedElements()`, or decision to not move forward with this proposal.
Thank you!

PR #2327, opened 2021-12-01
[reactive-element] Add queryAssignedElements decorator

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/2292

Adds `queryAssignedElements` decorator and tests (that similarly match the `queryAssignedNodes` tests).

This change proposes a new options bag API which lets us avoid needing to use and empty string if targeting an unnamed slot.

Note that there is temporarily an asymmetry between this decorator and `queryAssignedNodes` which will be addressed with https://github.com/lit/lit/issues/2332.

### How

 - The first commit introduces the new decorator, with tests matching closely what the `queryAssignedNodes` tests. In fact the fixture is almost identical.
 - Then I add the imports closely following the `query-assigned-nodes` references in the codebase.
 - Finally I updated the ts-transformer such that it can transform queryAssignedElements into JS.

### Testing

Change is tested with unit tests, as well as transformer tests.

### DIFF
--- a/.changeset/famous-feet-kneel.md
+++ b/.changeset/famous-feet-kneel.md
@@ -1,0 +1,9 @@
+---
+'lit': minor
+'lit-element': minor
+'lit-html': minor
+'@lit/reactive-element': minor
+'@lit/ts-transformers': minor
+---
+
+Add `queryAssignedElements` decorator for a declarative API that calls `HTMLSlotElement.assignedElements()` on a specified slot. `selector` option allows filtering returned elements by a CSS selector.
--- a/.changeset/famous-feet-kneel.md
+++ b/.changeset/famous-feet-kneel.md
@@ -1,9 +1,8 @@
 ---
 'lit': minor
 'lit-element': minor
-'lit-html': minor
 '@lit/reactive-element': minor
 '@lit/ts-transformers': minor
 ---
 
-Add `queryAssignedElements` decorator for a declarative API that calls `HTMLSlotElement.assignedElements()` on a specified slot. `selector` option allows filtering returned elements by a CSS selector.
+Add `queryAssignedElements` decorator for a declarative API that calls `HTMLSlotElement.assignedElements()` on a specified slot. `selector` option allows filtering returned elements with a CSS selector.
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -1,0 +1,80 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * IMPORTANT: For compatibility with tsickle and the Closure JS compiler, all
+ * property decorators (but not class decorators) in this file that have
+ * an @ExportDecoratedItems annotation must be defined as a regular function,
+ * not an arrow function.
+ */
+
+import {ReactiveElement} from '../reactive-element.js';
+import {decorateProperty} from './base.js';
+
+export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
+  /**
+   * A string name of the slot. Leave empty for the default slot.
+   */
+  slotName?: string;
+  /**
+   * A string which filters the results to elements that match the given css
+   * selector.
+   */
+  selector?: string;
+}
+
+/**
+ * A property decorator that converts a class property into a getter that
+ * returns the `assignedElements` of the given `slot`. Provides a declarative
+ * way to use
+ * [`slot.assignedElements`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements).
+ *
+ * @param options Object that sets options for nodes to be returned. See
+ *     [MDN parameters section](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements#parameters)
+ *     for available options. Also accepts two more optional properties,
+ *     `slotName` and `selector`.
+ * @param options.slotName Name of the slot. Undefined or empty string for the
+ *     default slot.
+ * @param options.selector Element results are filtered such that they match the
+ *     given css selector.
+ *
+ * ```ts
+ * class MyElement {
+ *   @queryAssignedElements({ slotName: 'list' })
+ *   listItems;
+ *   @queryAssignedElements()
+ *   unnamedSlotEls;
+ *
+ *   render() {
+ *     return html`
+ *       <slot name="list"></slot>
+ *       <slot></slot>
+ *     `;
+ *   }
+ * }
+ * ```
+ * @category Decorator
+ */
+export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
+  const {slotName, selector} = options ?? {};
+  return decorateProperty({
+    descriptor: (_name: PropertyKey) => ({
+      get(this: ReactiveElement) {
+        const slotSelector = `slot${
+          slotName ? `[name=${slotName}]` : ':not([name])'
+        }`;
+        const slot = this.renderRoot?.querySelector(slotSelector);
+        const elements = (slot as HTMLSlotElement).assignedElements(options);
+        if (selector) {
+          return elements.filter((node) => node.matches(selector));
+        }
+        return elements;
+      },
+      enumerable: true,
+      configurable: true,
+    }),
+  });
+}
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -18,7 +18,7 @@ export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
   /**
    * A string name of the slot. Leave empty for the default slot.
    */
-  slotName?: string;
+  slot?: string;
   /**
    * A string which filters the results to elements that match the given css
    * selector.
@@ -35,15 +35,15 @@ export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
  * @param options Object that sets options for nodes to be returned. See
  *     [MDN parameters section](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements#parameters)
  *     for available options. Also accepts two more optional properties,
- *     `slotName` and `selector`.
- * @param options.slotName Name of the slot. Undefined or empty string for the
+ *     `slot` and `selector`.
+ * @param options.slot Name of the slot. Undefined or empty string for the
  *     default slot.
  * @param options.selector Element results are filtered such that they match the
  *     given css selector.
  *
  * ```ts
  * class MyElement {
- *   @queryAssignedElements({ slotName: 'list' })
+ *   @queryAssignedElements({ slot: 'list' })
  *   listItems;
  *   @queryAssignedElements()
  *   unnamedSlotEls;
@@ -59,15 +59,13 @@ export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
  * @category Decorator
  */
 export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
-  const {slotName, selector} = options ?? {};
+  const {slot, selector} = options ?? {};
   return decorateProperty({
     descriptor: (_name: PropertyKey) => ({
       get(this: ReactiveElement) {
-        const slotSelector = `slot${
-          slotName ? `[name=${slotName}]` : ':not([name])'
-        }`;
-        const slot = this.renderRoot?.querySelector(slotSelector);
-        const elements = (slot as HTMLSlotElement).assignedElements(options);
+        const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
+        const slotEl = this.renderRoot?.querySelector(slotSelector);
+        const elements = (slotEl as HTMLSlotElement).assignedElements(options);
         if (selector) {
           return elements.filter((node) => node.matches(selector));
         }
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -16,12 +16,11 @@ import {decorateProperty} from './base.js';
 
 export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
   /**
-   * A string name of the slot. Leave empty for the default slot.
+   * Name of the slot. Leave empty for the default slot.
    */
   slot?: string;
   /**
-   * A string which filters the results to elements that match the given css
-   * selector.
+   * CSS selector used to filter the elements returned.
    */
   selector?: string;
 }
@@ -32,6 +31,8 @@ export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
  * way to use
  * [`slot.assignedElements`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements).
  *
+ * Note, the type of this property should be annotated as `Array<HTMLElement>`.
+ *
  * @param options Object that sets options for nodes to be returned. See
  *     [MDN parameters section](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements#parameters)
  *     for available options. Also accepts two more optional properties,
@@ -39,14 +40,14 @@ export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
  * @param options.slot Name of the slot. Undefined or empty string for the
  *     default slot.
  * @param options.selector Element results are filtered such that they match the
- *     given css selector.
+ *     given CSS selector.
  *
  * ```ts
  * class MyElement {
  *   @queryAssignedElements({ slot: 'list' })
- *   listItems;
+ *   listItems!: Array<HTMLElement>;
  *   @queryAssignedElements()
- *   unnamedSlotEls;
+ *   unnamedSlotEls?: Array<HTMLElement>;
  *
  *   render() {
  *     return html`
@@ -64,8 +65,9 @@ export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
     descriptor: (_name: PropertyKey) => ({
       get(this: ReactiveElement) {
         const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
-        const slotEl = this.renderRoot?.querySelector(slotSelector);
-        const elements = (slotEl as HTMLSlotElement).assignedElements(options);
+        const slotEl =
+          this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
+        const elements = slotEl?.assignedElements(options) ?? [];
         if (selector) {
           return elements.filter((node) => node.matches(selector));
         }
--- a/packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
+++ b/packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
@@ -1,0 +1,195 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {queryAssignedElements} from '../../decorators/query-assigned-elements.js';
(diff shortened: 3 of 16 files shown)


packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,60 @@
...
+ *       <slot name="list"></slot>
+ *     `;
+ *   }
+ * }
+ * ```
+ * @category Decorator
+ */
+export function queryAssignedElements(
+  slotName = '',
+  options?: AssignedNodesOptions | undefined
```
justinfagnani, line 45: Two things...
1. I wonder if we should fix the queryAssignedNodes API and take an options bag here? That way it's as easy to specify AssignedNodesOptions without a slot name as it is to specify a slot name without options.
2. Since queryAssignedNodes takes a selector to filter elements, should this?

AndrewJakubowicz: Yeah I agree that we may as well make this more useful, especially as the other decorator provides a filter. Below are two API options. Do you have a preference or suggestion?

### Option 1

```ts
interface QueryAssignedElementsOptions {
  /**
   * Name of the slot. Leave undefined for unnamed default slot.
   */
  slotName?: string,
  /**
   * If it exists, is passed directly into `HTMLSlotElement.assignedElements(options)`.
   */
  options?: AssignedNodesOptions,
  /**
   * A string which filters the results to elements that match the given css selector.
   */
  selector?: string
}

function queryAssignedElements(
  options?: QueryAssignedElementsOptions
) { ... }
```

### Option 2

Flatten the options bag (but there could be collisions in the future if more options than `flatten` are added), i.e.: `interface QueryAssignedElementsOptions extends AssignedNodesOptions { /* slotName, selector */ }`



Does it also make sense to migrate other decorators with or shortly after this one with the options bag approach for consistency, or do we need to wait for a certain release?

Thank you!

justinfagnani: I like option 2.

I think it'd be ideal to update `queryAssignedNodes` as well, but in a non-breaking way by type-checking the first argument and adding a TS override.

AndrewJakubowicz: Updated this commit. Also created a follow up task for `queryAssignedNodes`: https://github.com/lit/lit/issues/2332

packages/ts-transformers/src/internal/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,255 @@
...
+      return {slotName: '', selector: ''};
+    }
+    const findStringLiteralFor = (key: string): string => {
+      const propAssignment = opts.properties.find(
+        (p) => p.name && ts.isIdentifier(p.name) && p.name.text === key
+      );
+      if (
+        propAssignment &&
+        ts.isPropertyAssignment(propAssignment) &&
+        ts.isStringLiteral(propAssignment.initializer)
```
aomarks, line 95: Should probably throw if it's not a string literal, because it could also be a variable and we'd silently ignore that here.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+        <slot></slot>
+      `;
+    }
+  }
+  customElements.define('assigned-elements-el-2', E);
+
+  const defaultSymbol = Symbol('default');
+  const headerSymbol = Symbol('header');
+  class S extends RenderingElement {
+    @queryAssignedElements() [defaultSymbol]!: Element[];
```
justinfagnani, line 65: This is really just testing that `decorateProperty()` works with symbols. That should be tested once, not with every decorator. If you want to be extra sure that this works in case it's refactored, one symbol-keyed property among the rest in the existing classes should suffice.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+      this.assignedEls3 = this.renderRoot.querySelector(
+        'assigned-elements-el-symbol'
+      ) as S;
+    }
+  }
+  customElements.define(generateElementName(), C);
+
+  setup(async () => {
+    container = document.createElement('div');
+    container.id = 'test-container';
```
justinfagnani, line 121: doesn't seem used

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+    document.body.appendChild(container);
+    el = new C();
+    container.appendChild(el);
+    await el.updateComplete;
+    await el.assignedEls.updateComplete;
```
justinfagnani, line 126: Use Promise.all() and await all the elements? Or await a setTimeout?

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  teardown(() => {
+    if (container !== undefined) {
```
justinfagnani, line 130: `container?.remove();` and you probably don't need to null-out the variable

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+  });
+
+  test('returns assignedElements for slot', () => {
+    // Note, `defaultAssigned` does not `flatten` so we test that the property
+    // reflects current state and state when nodes are added or removed.
+    assert.deepEqual(el.assignedEls.defaultAssigned, [el.div]);
+    const child = document.createElement('div');
+    const text1 = document.createTextNode('');
+    el.assignedEls.appendChild(text1);
```
justinfagnani, line 142: `el.assignedEls.append(text1, child)`

packages/ts-transformers/src/internal/decorators/query-assigned-elements.ts
AndrewJakubowicz: Done thank you!

justinfagnani (changes requested):

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import {ReactiveElement} from '../reactive-element.js';
+import {decorateProperty} from './base.js';
+
+export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
+  /**
+   * A string name of the slot. Leave empty for the default slot.
+   */
+  slotName?: string;
```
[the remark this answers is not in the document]
AndrewJakubowicz: Done!

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,195 @@
...
+    footerAssignedFiltered!: Element[];
+
+    override render() {
+      return html`
+        <slot></slot>
+        <slot name="footer"></slot>
+      `;
+    }
+  }
+  customElements.define('assigned-elements-el', D);
```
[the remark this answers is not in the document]
AndrewJakubowicz: Agree!

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: Great point. Left the default symbol case (for extra certainty) but removed the header symbol.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: Confirmed.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: These need to be sequential so opted into using `setTimeout`.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: Updated this everywhere relevant.

.changeset/famous-feet-kneel.md
```diff
@@ -0,0 +1,9 @@
+---
+'lit': minor
+'lit-element': minor
+'lit-html': minor
```
justinfagnani, line 4: does lit-html need to be minor?

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
justinfagnani: I think you can remove this element and just make one of the fields in one of the other classes a symbol.

.changeset/famous-feet-kneel.md
AndrewJakubowicz: It does not, thank you! Only `reactive-element` & `ts-transformers` have logic changes. `lit` and `lit-element` re-export the new decorator so I assume they should remain.

Thanks for catching this.

packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -253,8 +253,7 @@ export const resolveSpecifier = async (
     if (
       specifierMatches(specifier, 'lit') ||
       specifierMatches(specifier, 'lit-html') ||
-      specifierMatches(specifier, 'lit-element') ||
-      specifierMatches(specifier, '@lit/reactive-element')
+      specifierMatches(specifier, 'lit-element')
```
justinfagnani, line 256: Talked on chat: I think you should leave this and add `@lit/reactive-element` to ssr's dependencies

AndrewJakubowicz: Perfect suggestion. Worked! Thank you!

AndrewJakubowicz:
Tests are passing. The only tests failing is something to do with Sauce not connecting:

```
ests: Error: 3 Dec 01:36:24 - Sauce Connect could not establish a connection.
tests: 3 Dec 01:36:24 - Please check your firewall and proxy settings.
tests: 3 Dec 01:36:24 - Goodbye.
tests:     at Socket.<anonymous> (/home/runner/work/lit/lit/packages/tests/node_modules/saucelabs/build/index.js:322:25)
tests:     at Socket.emit (events.js:400:28)
tests:     at Socket.emit (domain.js:475:12)
tests:     at addChunk (internal/streams/readable.js:293:12)
tests:     at readableAddChunk (internal/streams/readable.js:267:9)
tests:     at Socket.Readable.push (internal/streams/readable.js:206:10)
tests:     at Pipe.onStreamRead (internal/stream_base_commons.js:188:23)
```

I think this is unrelated to my changes.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: Done!

.changeset/famous-feet-kneel.md
```diff
@@ -0,0 +1,8 @@
+---
+'lit': minor
+'lit-element': minor
+'@lit/reactive-element': minor
+'@lit/ts-transformers': minor
+---
+
+Add `queryAssignedElements` decorator for a declarative API that calls `HTMLSlotElement.assignedElements()` on a specified slot. `selector` option allows filtering returned elements with a CSS selector.
```
aomarks, line 8: Update this since the `selector` option is different now?

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+/*
+ * IMPORTANT: For compatibility with tsickle and the Closure JS compiler, all
+ * property decorators (but not class decorators) in this file that have
+ * an @ExportDecoratedItems annotation must be defined as a regular function,
+ * not an arrow function.
+ */
+
+import {ReactiveElement} from '../reactive-element.js';
```
aomarks, line 14:
```suggestion
import type {ReactiveElement} from '../reactive-element.js';
```

packages/ts-transformers/src/internal/decorators/query-assigned-elements.ts
aomarks: Actually just occured to me that `selector` could be used as-is whatever kind of node it is, since it just goes directly into the `filter` arguments, right?

And you could maybe also handle `slot` being an identifier or whatever too by generating `` `slot[name=${node}]` `` in the case that it's not a string?

aomarks (changes requested):
This is great! Awesome to do get the transformer code in too.

.changeset/famous-feet-kneel.md
AndrewJakubowicz: Thanks for the comment! Would you be able to elaborate? Are you referring to the removal of the Node filter?

In `queryAssignedNodes` the selector changes the return type from `Node[]` to `Element[]`, whilst in this decorator the selector shouldn't change the return type. A docs update to `queryAssignedNodes` is tracked in https://github.com/lit/lit/issues/2171.

packages/reactive-element/src/decorators/query-assigned-elements.ts
AndrewJakubowicz: Nice 🎉

.changeset/famous-feet-kneel.md
aomarks: Oh sorry, I misread this -- I was thinking this was written before you added the `options` parameter, but when I read it again it's fine (I think I read `selector` option as `selector` parameter).

packages/ts-transformers/src/internal/decorators/query-assigned-elements.ts
```diff
@@ -251,4 +264,26 @@ export class QueryAssignedElementsVisitor implements MemberDecoratorVisitor {
...
     );
   }
+
+  /**
+   * @param slot Expression that evaluates to the slot name.
+   * @returns Template string node representing `slot[name=${slot}]`
+   */
+  private createNamedSlotSelector(slot: ts.Expression): ts.TemplateExpression {
+    const factory = this._factory;
+    return factory.createTemplateExpression(
```
aomarks, line 274: Consider checking if slot is a string literal here, and generating a pre-interpolated string here (e.g. `"slot[name=foo]"` instead of `"slot[name=${"foo"}]`). That will be faster (no need for a runtime interpolation), and also more like what a human would usually work (important because the main use case of the transformers is actually docs generation).

AndrewJakubowicz: Genius comment. Thank you. Added the special handling for string literals and no substitution template literals. I.e. "foo" or `foo`. Also updated tests so both are tested.

aomarks (approved):
:100:

packages/reactive-element/src/decorators/query-assigned-elements.ts
justinfagnani: did this get changed?

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,78 @@
...
+import {decorateProperty} from './base.js';
+
+export interface QueryAssignedElementsOptions extends AssignedNodesOptions {
+  /**
+   * A string name of the slot. Leave empty for the default slot.
+   */
+  slot?: string;
+  /**
+   * A string which filters the results to elements that match the given css
```
justinfagnani, line 23: capitalize "CSS"

justinfagnani: I don't think you need the phrase with the type like "A string...", I think you can say "A CSS selector used to filter the elements returned..."

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,78 @@
...
+ *     `slot` and `selector`.
+ * @param options.slot Name of the slot. Undefined or empty string for the
+ *     default slot.
+ * @param options.selector Element results are filtered such that they match the
+ *     given css selector.
+ *
+ * ```ts
+ * class MyElement {
+ *   @queryAssignedElements({ slot: 'list' })
+ *   listItems;
```
justinfagnani, line 47: I think we should show the recommended type annotation here and below: `listItems?:  Array<HTMLElement>`

And we should add a note about the return type to the JSDoc, like in `queryAssignedNodes`, but correct (not `NodeListOf`)

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @category Decorator
+ */
+export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
+  const {slot, selector} = options ?? {};
+  return decorateProperty({
+    descriptor: (_name: PropertyKey) => ({
+      get(this: ReactiveElement) {
+        const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
+        const slotEl = this.renderRoot?.querySelector(slotSelector);
+        const elements = (slotEl as HTMLSlotElement).assignedElements(options);
```
justinfagnani, line 68: I think the cast is obscuring that `slotEl` may be undefined or null. Try:
```ts
        const slotEl = this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
        const elements = slotEl.assignedElements(options);
```

And then you'll have to handle the other cases.

justinfagnani (changes requested):

packages/reactive-element/src/decorators/query-assigned-elements.ts
AndrewJakubowicz: Done. Great comment.

packages/reactive-element/src/decorators/query-assigned-elements.ts
AndrewJakubowicz: Done!

packages/reactive-element/src/decorators/query-assigned-elements.ts
AndrewJakubowicz: Done! Handled with the `?? []` used in query-assigned-nodes. Great catch!

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -0,0 +1,80 @@
...
+ *     `slot` and `selector`.
+ * @param options.slot Name of the slot. Undefined or empty string for the
+ *     default slot.
+ * @param options.selector Element results are filtered such that they match the
+ *     given CSS selector.
+ *
+ * ```ts
+ * class MyElement {
+ *   @queryAssignedElements({ slot: 'list' })
+ *   listItems?: Array<HTMLElement>;
```
justinfagnani, line 48: The `?` can be `!` since the getter always returns an array.

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
```diff
@@ -0,0 +1,162 @@
...
+    assert.equal(el.assignedEls.footerNotFlattenedSlot.length, 1);
+    assert.equal(el.assignedEls.footerNotFlattenedSlot?.[0]?.tagName, 'SLOT');
+
+    child2.remove();
+    flush();
+    assert.deepEqual(el.assignedEls.footerAssigned, [child1]);
+  });
+
+  test('always returns an array, even if the slot is not rendered', () => {
+    assert.isArray(el.missingSlotAssignedElements);
```
justinfagnani, line 147: I'm surprised that this test was here before the change to always return an array. Was it failing? If it was passing before... why?

justinfagnani (approved):

packages/reactive-element/src/test/decorators/queryAssignedElements_test.ts
AndrewJakubowicz: It was passing because it was selecting: `<slot slot="footer"></slot>`.

PR #2331, opened 2021-12-02
[labs/router] Add initial implementation of labs/router

justinfagnani:
This adds a router implemented as a pair of reactive controllers: `Router` and `Routes`.

Features:
- Declarative route configuration
- Routes declared with:
  -  A [URLPattern](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) to match locations
  - A `render()` method to render a template when the route matches
- Outlets: Automatically renders the currently matched route to a Lit template
- Nested routes
- Dynamic mutable route configurations
- Async enter() callbacks

Example:

```ts
  private _routes = new Routes(this, [
    {path: '/', render: () => html`<h1>Home</h1>`},
    {path: '/projects', render: () => html`<h1>Projects</h1>`},
    {path: '/about', render: () => html`<h1>About</h3>`},
  ]);
  render() {
    return html`
      <header>...</header>
      <main>${this._routes.outlet}</main>
      <footer>...</footer>
    `;
  }
}
```


packages/labs/router/src/lib/url-pattern.d.ts
```diff
@@ -0,0 +1,43 @@
+declare class URLPattern {
```
aomarks, line 1:
```suggestion
/**
 * @license
 * Copyright 2021 Google LLC
 * SPDX-License-Identifier: BSD-3-Clause
 */
 
declare class URLPattern {
```

(Or if this is copied from somewhere else, add its copyright header)

aomarks:
```suggestion
// See https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1199 to track adding URLPattern types to the standard TypeScript DOM types.

declare class URLPattern {
```

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export type RouteConfig = StringRouteConfig | URLPatternRouteConfig;
+
+/**
+ * A reactive controller that performs location-based routing using a
+ * configuration of URL patterns and associated render callbacks.
+ */
+export class Routes implements ReactiveController {
+  protected _host: ReactiveControllerHost & HTMLElement;
```
aomarks, line 28: Could we make some of these `readonly`? (And do we care about doing that as a rule?)

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+        render: route.render,
+        path: (route as StringRouteConfig).path,
+        pattern:
+          (route as URLPatternRouteConfig).pattern ??
+          new URLPattern({pathname: (route as StringRouteConfig).path}),
+      });
+    }
+  }
+
+  goto(pathname: string) {
```
aomarks, line 79: Add JSDoc

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+      this._onRoutesConnected
+    );
+    // TODO: connect to parent router
+    const event = new RoutesConnectedEvent(this);
+    this._host.dispatchEvent(event);
+    this._onDisconnect = event.onDisconnect;
+  }
+
+  hostDisconnected() {
+    // When this child routes constroller is disconnected because a parent
```
aomarks, line 143:
```suggestion
    // When this child routes controller is disconnected because a parent
```

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    const childRoutes = e.routes;
+    this._childRoutes.push(childRoutes);
+
+    e.stopImmediatePropagation();
+    e.onDisconnect = () => {
+      // Remove route from this._childRoutes:
+      this._childRoutes?.splice(
+        this._childRoutes.indexOf(childRoutes) >>> 0,
```
aomarks, line 163: Comment the right shift? Is the reason to turn `-1` into a no-op?

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+      composed: true,
+      cancelable: false,
+    });
+    this.routes = routes;
+  }
+}
+
+declare global {
+  interface HTMLElementEventMap {
+    'routes-connected': RoutesConnectedEvent;
```
aomarks, line 210: No strong opinion, but should we use `lit-routes-connected` for a lower chance of event name collision? We use `lit-localize-status` in localize.

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/**
+ * Returns the tail of a pathname groups object. This is the match from a
+ * wildcard at the end of a pathname pattern, like `/foo/*`
+ */
+const getTailGroup = (groups: {[key: string]: string}) => {
+  let tailKey: string | undefined;
+  for (const key of Object.keys(groups)) {
+    if ((/\d+/.test(key) && tailKey === undefined) || key > tailKey!) {
```
aomarks, line 182: Should this be:

```js
if (/\d+/.test(key) && (tailKey === undefined || key > tailKey)) {
```

?

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,82 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-export {};
+import {ReactiveControllerHost} from 'lit';
```
aomarks, line 7:
```suggestion
import type {ReactiveControllerHost} from 'lit';
```

aomarks: Should we get into the habit of declaring our type-only imports?

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,82 @@
...
+  }
+
+  override hostDisconnected() {
+    super.hostDisconnected();
+    window.removeEventListener('click', this._onClick);
+    window.removeEventListener('popstate', this._onPopState);
+  }
+
+  private _onClick = (e: MouseEvent) => {
+    const isNonNavigationClick =
```
aomarks, line 44: Crazy that all this low-level stuff has to get checked! Feels like there are some missing platform APIs for intercepting navigations.

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,82 @@
...
+    if (href === '' || href.startsWith('mailto:')) {
+      return;
+    }
+
+    const location = window.location;
+    const origin = location.origin || location.protocol + '//' + location.host;
+    if (anchor.origin === origin) {
+      return;
+    }
+
```
aomarks, line 74: Do we also need any special handling for hash-only navigations?

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,82 @@
...
+    const href = anchor.href;
+    if (href === '' || href.startsWith('mailto:')) {
+      return;
+    }
+
+    const location = window.location;
+    const origin = location.origin || location.protocol + '//' + location.host;
```
aomarks, line 70: Could it be beneficial to cache this `origin` value as a module scoped variable, since it should never change (right?)?

aomarks (changes requested):
Cool!

packages/labs/router/src/router.ts
jpzwarte: `typescript-eslint` has support for this (including auto-fix): 

```javascript
    '@typescript-eslint/consistent-type-imports': [
      'error',
      {
        prefer: 'type-imports'
      }
    ],
```

packages/labs/router/src/router.ts
justinfagnani: There's a really nice App History API in origin trial right now!

packages/labs/router/src/routes.ts
justinfagnani: ah, good point

packages/labs/router/src/routes.ts
justinfagnani: ooh yes!

packages/labs/router/src/router.ts
justinfagnani: most likely, yes

packages/labs/router/src/router.ts
justinfagnani: +1

packages/labs/router/src/router.ts
justinfagnani: There's a whole lot to do around query params and fragments. Going to resolve this to move TODO tasks to a doc or issues.

packages/labs/router/src/lib/url-pattern.d.ts
justinfagnani: This file will likely go away by the time the PR is done, by including the URLPattern polyfill.

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+    if (pathname?.startsWith('.')) {
+      throw new Error('Not implemented');
+    }
+    pathname ??= this._currentPathname;
+    return (this._parentRoutes?.link() ?? '') + pathname;
+  }
+
+  /**
+   * Navigates to this routes controller to `pathname`.
```
aomarks, line 100:
```suggestion
   * Navigates this routes controller to `pathname`.
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
```
aomarks, line 4: Maybe add a disclaimer that this isn't yet released?

Unless we're planning not to merge this to main until it's ready to release (I think I'd advocate for merging even before we're ready for a preview release, so that we keep the tests running with mainline dev).

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+}
+```
+
+Routes can be nested: a route path can include a trailing `/*` pattern to match against a prefix, and will automatically propagate that prefix to Routes controllers defined in child elements.
+
+The general shape of the API includes:
+
+- A `Router` controller that's used as a top-level singleton to set up event listeners
+- A `Routes` controller for declaring routes inside components
+- Declaration of routes with `URLPatterns` and render callbacks
```
aomarks, line 35:
```suggestion
- Declaration of routes with [`URLPattern`](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) and render callbacks
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+### Routes
+
+Routes is the main interface into the router API. A Routes controller contains route definitions and the templates that each route renders:
+
+```ts
+class MyElement extends LitElement {
+  private routes = new Routes(this, [
+    {path: '/', render: () => html`<h1>Home</h1>`},
+    {path: '/projects', render: () => html`<h1>Projects</h1>`},
+    {path: '/about', render: () => html`<h1>About</h3>`},
```
aomarks, line 78:
```suggestion
    {path: '/about', render: () => html`<h1>About</h1>`},
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+}
+```
+
+The second argument is the route configuration: an array of `RouteConfig` objects.
+
+#### RouteConfig
+
+A RouteConfig contains at the minimum the pattern to match URLs against and a template to render. Names can be provided to reference routes for link generation.
+
+There are two types of RoutConfig: `PathRouteConfig` and `URLPatternRouteConfig`:
```
aomarks, line 89:
```suggestion
There are two types of `RouteConfig`s: `PathRouteConfig` and `URLPatternRouteConfig`:
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+  render: ({id}) => html`<x-profile .profileId=${id}></x-profile>`
+}
+```
+
+#### Outlets
+
+An outlet is where a routes object renders the currently selected route's template. It can be used anywhere in the host element's template:
+
+```ts
+html`${this.routes.outlet}`;
```
aomarks, line 137:
```suggestion
html`<main>${this.routes.outlet}</main>`;
```

Since you wouldn't usually need the `html` wrapper unless you also had some markup to wrap around it. Sticking in a `main` might be a common top-level wrapping pattern.

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+```
+
+#### `goto()`
+
+`goto(url: string)` is a programmatic navigation API. It should be able to take full URLs for top-level navigation and relative URLs for navigation within a nested route space.
```
aomarks, line 142:
```suggestion
`goto(url: string)` is a programmatic navigation API. It takes full URLs for top-level navigation and relative URLs for navigation within a nested route space.
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+
+`goto(name: string, params: object)` _(not implemented)_ allows navigation via named routes. The name and params are scoped to the Routes object it's called on, though nested routes can be triggered by a "tail" parameter - the match of a trailing /\*parameter (See tail groups).
```
aomarks, line 144:
```suggestion
`goto(name: string, params: object)` _(not implemented)_ allows navigation via named routes. The name and params are scoped to the Routes object it's called on, though nested routes can be triggered by a "tail" parameter - the match of a trailing `/*` parameter (See tail groups).
```

packages/labs/router/README.md
```diff
@@ -4,4 +4,193 @@ A router for Lit.
...
+
+Components need to generate links to resources within the app. It's desirable to not require that link generation has global knowledge of URLs, so links should be able to be generated with either only local information, or with abstracted parameters. `routes.link()` helps generate different kinds of links:
+
+##### Relative links
+
+Relative links are relative to the parent route and its current state, and can be specified with a path string or name and parameters.
+
+Examples
+
+- `this._routes.link(user.id)` - within a `<x-user>` component that has a route pattern like ':id', this would link to another user profile.
```
aomarks, line 156:
```suggestion
- `this._routes.link(user.id)` - within a `<x-user>` component that has a route pattern like `":id"`, this would link to another user profile.
```

aomarks (approved):
Nice!

packages/labs/router/README.md
```diff
@@ -2,6 +2,212 @@
...
+
+In this example, the page can handle URLs `/foo`, `/child/foo` and `/child/bar`.
+
+## TODO
+
+### Server router integration
+
+This client-side router is intended to be able to integrate with server-side routing, both configuration-based routers and convention-based routers like file-based routers.
+
+Current ideas for integration include a fallback route handler (instaled by the app) that delegates to a server API call that returns additional route configuration dynamically. The minimal API needed in the client router are:
```
kenchris, line 210: installed (two l's)

packages/labs/router/README.md
```diff
@@ -2,6 +2,212 @@
...
+}
+```
+
+### Routes
+
+Routes is the main interface into the router API. A Routes controller contains route definitions and the templates that each route renders:
+
+```ts
+class MyElement extends LitElement {
+  private routes = new Routes(this, [
```
kenchris, line 81: So this is Route*S* and not Route*R* as above?  Mistake?

kenchris: I see you have Router inheriting from Routes. As the names look so similar, it might make sense to point out this difference or find a better name for one of the two

rictic (commented):
oops, somewhat old comments that I apparently forgot to publish

packages/labs/router/src/lib/url-pattern.d.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(init: URLPatternInit | string, baseURL?: string);
```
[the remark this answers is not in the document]
justinfagnani: I deleted this file in favor of the types from the polyfill

packages/labs/router/src/routes.ts
```diff
@@ -163,14 +167,24 @@ export class Routes implements ReactiveController {
...
       }
       const pattern = getPattern(route);
-      const r = pattern.exec({pathname});
-      this._currentParams = r?.pathname.groups ?? {};
-      tailGroup = getTailGroup(this._currentParams);
+      const result = pattern.exec({pathname});
+      const params = result?.pathname.groups ?? {};
+      tailGroup = getTailGroup(params);
+      if (typeof route.enter === 'function') {
+        const success = await route.enter(params);
```
rictic, line 179: what happens if route.enter rejects?

rictic, line 179: what happens if another navigation begins before this one is finished (e.g. the user clicks a link, nothing happens for a bit, so they click it a few more times)?

rictic, line 179: This is a pit of failure for UX

Easy to block a route transition on a process that may be slow (and worse: fast in dev and slow in prod). In that case, as I read this, the user clicks a link and literally nothing happens

If we go with this design, we'll want to expose hooks so that apps can display a sign of life during the navigation (even if it's just an indeterminate progress bar along the top of the window)

justinfagnani: That should fine here since none of the state for matching this route is recorded unless enter() returns non-false. Obviously we need a lot more tests...

justinfagnani: There's a TODO at the top of `goto()` related to this too.

We need to consider a few use cases here - it'd be good to gather realistic ones from the community. I can imagine cases where you want a new navigation to cancel a pending one, and some cases where you want a pending navigation to block new navigations (some kinds of navigating form submission). And forms may bring up a case where you want to block navigations from arbitrary code outside of the router. And maybe all of that _shouldn't_ be in the router because the browser doesn't disable links and what you want is a user-affordance like a modal with a scrim that disables all the visible DOM.

I'll add something to the design doc.

justinfagnani: There's an intersection with Task here in that technically we have the same four async states: initial, pending, success, failure. I'd say for a router you really want to only expose pending and success to users (404 should be success for the 404 route).

This is one of the many things we need to collect use-cases for, probably as a user-story/design-requirement for the UX a developer is trying to create. Some cases will want to not do anything until the new route is ready to render. Native links are mostly like this (though we'll want to integrate with the [pending task protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/pending-task.md) so a page could build a pending state affordance). Some cases will want to transition to a pending/loading UI, then transition to the rendered route.

rictic: (nit, typo in the pending task protocol doc: https://github.com/webcomponents-cg/community-protocols/pull/28)

rictic: Yeah, I think this is largely a thing to solve in defaults, affordances, and documentation

Ooh, looking at app history, we could potentially make use of the navigate event's transitionWhile method to use the browser's ordinary affordances for indicating that it's mid-navigation.

If that's the case though, I think that (at some point, perhaps not when app history initially launches) the navigation will probably also be cancellable by the user

sorvell (changes requested):

packages/labs/router/README.md
```diff
@@ -2,6 +2,255 @@
...
 ## Overview
 
-_This project is just starting. Details TBD._
+`@lit-labs/router` is a component-oriented router API vended as reactive controllers. Routes are configured as part of component definitions, and integrated into the component lifecycle and rendering.
+
+Usage will generally look like this, with a configuration in a reactive controller, and rendering done via route-specific render callbacks and an "outlet" to use in the main render() method:
+
+```ts
+class MyElement extends LitElement {
+  private _routes = new Routes(this, [
```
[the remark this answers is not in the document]
kevinpschaaf: The `this` in the render fns is still bound to the element here, so ``() => html`<h1>Home: ${this.someProp}</h1>` `` works, right?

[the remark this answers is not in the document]
justinfagnani: Yes, that's right.

packages/labs/router/README.md
```diff
@@ -2,6 +2,255 @@
...
+- Extraction of URL pattern parameters into data objects passed to render callbacks
+- A `routes.outlet` that renders the current route's render callback
+- A `routes.link()` method to generate URLs to use in `<a>` tags, etc.
+- A `routes.goto()` method for performing a navigation
+
+## API
+
+### Router
+
+A Router is a controller (a subclass of the Routes controller) for use at the top-level of an application. It's main purpose is to set up global `click` and `popstate` event listeners (which should be installed only once on a page). It can optionally contain route definitions as a convenience.
```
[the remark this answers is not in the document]
justinfagnani: I had considered making `Router` _not_ extend `Routes` and just be a thing that installs the event listeners. For simple cases though a component would need to install both a `Router` and a `Routes`, and that seemed unfortunate.

React Router also has this Router/Routes division. In their case they have browser, server, and browser/fragment implementations. We're going to try to run the same router on the client and server, but one could imagine a Router implementation that uses the App History API instead of catching click and popstate events.

I'm hoping we can document that "Router goes with the app, Routes go with nested components" and that'll make sense.

packages/labs/router/README.md
```diff
@@ -2,6 +2,255 @@
...
+}
+```
+
+The second argument is the route configuration: an array of `RouteConfig` objects.
+
+#### RouteConfig
+
+A RouteConfig contains at the minimum the pattern to match URLs against and a template to render. Names can be provided to reference routes for link generation.
+
+There are two types of `RouteConfig`s: `PathRouteConfig` and `URLPatternRouteConfig`:
```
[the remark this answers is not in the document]
justinfagnani: This is there because if we make URLPattern objects for you, it seems good to let you bring your own. It's only the difference between:

```ts
{
  path: '/foo/:id',
  render({id}) { ... }
}
```

and

```ts
{
  pattern: new URLPattern({pathname: '/foo/:id'}),
  render({id}) { ... }
}
```

The key additional feature that making your own URLPattern allows now is to match on query params. That seems niche enough to relegate to bring-you-own-pattern-object.

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,75 @@
...
+ * This class extends Routes so that it can also have a route configuration.
+ *
+ * There should only be one Router instance on a page, since the Router
+ * installs global event listeners on `window` and `document`. Nested
+ * routes should be configured with the `Routes` class.
+ */
+export class Router extends Routes {
+  override hostConnected() {
+    super.hostConnected();
+    window.addEventListener('click', this._onClick);
```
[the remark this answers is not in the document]
justinfagnani: Depends how closely we want to match native <a>/click events. I think stop propagation preventing routing is fine, though that's not how the browser works. If we want to be closer to native behavior and did capture I'm not sure how and when we'd check for `defaultPrevented`?

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,307 @@
...
+const patternCache = new WeakMap<PathRouteConfig, URLPattern>();
+
+const isPatternConfig = (route: RouteConfig): route is URLPatternRouteConfig =>
+  (route as URLPatternRouteConfig).pattern !== undefined;
+
+const getPattern = (route: RouteConfig) => {
+  if (isPatternConfig(route)) {
+    return route.pattern;
+  }
+  let pattern = patternCache.get(route);
```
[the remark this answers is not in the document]
justinfagnani: Not sure. They're parsed so maybe, but if not cached they'd only be parsed on navigations, which is usually infrequent.

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,307 @@
...
+      throw new Error('Not implemented');
+    }
+    pathname ??= this._currentPathname;
+    return (this._parentRoutes?.link() ?? '') + pathname;
+  }
+
+  /**
+   * Navigates this routes controller to `pathname`.
+   *
+   * This does not navigate parent routes, so it isn't (yet) a general page
```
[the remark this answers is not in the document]
justinfagnani: I'll raise this in the design doc if I haven't already. The API shape here should ideally support both 1) navigating to a route within sub-routes scope without knowledge of the global URL space, and 2) routing to anywhere in the global URL space if you do have that knowledge. There are a bunch of tradeoffs and couplings around that.

packages/labs/router/README.md
```diff
@@ -2,6 +2,255 @@
...
+  private _routes = new Routes(this, [
+    {path: '/', render: () => html`<h1>Home</h1>`},
+    {path: '/projects', render: () => html`<h1>Projects</h1>`},
+    {path: '/about', render: () => html`<h1>About</h3>`},
+  ]);
+
+  render() {
+    return html`
+      <header>...</header>
+      <main>${this._routes.outlet}</main>
```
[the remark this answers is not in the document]
justinfagnani: Yeah, I can make that change. I considered `this._routes.render()`, but I feel like we're getting a lot of `render()` methods that need to be differentiated. `renderOutlet()` is a bit longer, but more precise.

packages/labs/router/src/routes.ts
justinfagnani: I think `enter()` (and maybe an `exit()`) is good because it lets `render()` stay synchronous, can reject a route transition, and signals that the route is in a pending state, which we might want to expose to `render()` so that it can rendering a loading message.

We also had a request to add middleware to the router. I need to go back and check the examples, but it seemed like middleware might be able to just hook enter/exit and render is still just part of the route configuration.

thepassle: > We also had a request to add middleware to the router.

https://lit-and-friends.slack.com/archives/C03PF4L4L/p1638556109035200?thread_ts=1638548761.018500&cid=C03PF4L4L

justinfagnani: omg, thanks for adding that link. I was just about to ask you where that was :)

packages/labs/router/src/router.ts
```diff
@@ -4,4 +4,75 @@
...
+    window.addEventListener('popstate', this._onPopState);
```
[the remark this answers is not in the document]
kevinpschaaf: Nvm, missed the assignment to arrow function.

packages/labs/router/README.md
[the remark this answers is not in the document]
sorvell: What about just `outlet()` ?

packages/labs/router/README.md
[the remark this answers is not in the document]
sorvell: Seems ok I guess.

packages/labs/router/README.md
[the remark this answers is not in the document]
sorvell: Sounds good.

packages/labs/router/README.md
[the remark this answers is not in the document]
sorvell: OK.

packages/labs/router/src/routes.ts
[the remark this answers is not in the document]
sorvell: OK.

packages/labs/router/src/router.ts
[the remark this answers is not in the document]
sorvell: it's bound... `_onPopState = () => ...`

sorvell (approved):

packages/labs/router/src/routes.ts
thepassle: Some more context:

Here's a more up to date example. Arguably some of these things could/should even be baked into the router, like resetting focus and setting the page title, and would be configurable in the router configuration

```js
const routes = [
  {
    path: '/shop',
    component: 'shop-page',
    action: middleware(
      title('shop'), // sets page title
      offline, // redirects user to the offline screen
      auth, // checks auth, if not authed redirect to offline screen
      lazy(() => import('../pages/shop-page/index.js')), // lazy load
      scrollIntoView, // scrolls a position
      restoreFocus // resets focus
    )
  },
];
```

Here's the code for the middleware function, its not very big:

```js
async function invokeMiddlewares(middlewares) {
  if (!middlewares.length) return;

  const mw = middlewares[0];

  return mw(async () => {
    await invokeMiddlewares(middlewares.slice(1));
  });
}

export function middleware(...middlewares) {
  return () => {
    invokeMiddlewares(middlewares);
  };
}
```

abdonrd: Regarding the title update...
I think we will want to update more things like the description and the meta tags.

packages/labs/router/README.md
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/router/README.md
```diff
@@ -2,6 +2,255 @@
...
+```ts
+{
+  path: '/*',
+  render: (params) => html`<h1>Not found: params[0]</h1>`,
+  enter: async (params) => {
+    const path = params[0];
+    const dynamicRoute = getDynamicRoute(path);
+    if (dynamicRoute) {
+      const {routes} = this._router;
+      routes.splice(routes.langth - 1, 0, dynamicRoute);
```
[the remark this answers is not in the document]
justinfagnani: The splice is needed to insert before the fallback. We talked about making fallback a separate option so that this isn't needed.

packages/labs/router/src/routes.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    // TODO (justinfagnani): generalize this to handle query params and
+    // fragments. It currently only handles path names because it's easier to
+    // completely disregard the origin for now. The click handler only does
+    // an in-page navigation if the origin matches anyway.
+    let tailGroup: string | undefined;
+    if (this.routes.length === 0) {
+      // If a routes controller has none of its own routes it acts like it has
+      // one route of `/*` so that it passes the whole pathname as a tail
+      // match.
```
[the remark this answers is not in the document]
justinfagnani: I think this works actually, because a fallback that loads from the server should return false in it's enter, which will prevent any update to router state - those updates will instead come from the server-fetched route. I have a test that simulates that for one level, we can expand for multiple levels.

PR #2335, opened 2021-12-04
[lit-html] Add when, map, join, and range directives.

justinfagnani:
These helpers are simple functions, not directives, that help make a template read a bit more declaratively when containing conditionals and loops. They're not strictly necessary, but we've found a number of users ask for them or write their own.

Adds:
* `when()`: sugar for ternaries. Especially nice when there's no else clause
* `map()`: Evaluate a function of items in an iterable. Stylistically more "declarative" and works with iterables.
* `join()`: Interleaves the items of an iterable with a joiner value or result of calling a joiner function.
* `range()`: Returns an iterable that contains number in a range. Useful when combined with `map()`.


packages/lit-html/src/helpers.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
aomarks, line 3:
```suggestion
 * Copyright 2021 Google LLC
```

packages/lit-html/src/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * When `condition` is true, returns the result of calling `trueCase()`, else
+ * returns the result of calling `falseCase()` if `falseCase` is defined.
+ *
+ * This is a convenience wrapper around a ternary expression that makes it a
+ * little nicer to write an inline conditional without an else.
```
aomarks, line 12: Would love a usage example in each of the jsdocs.

You can use ` ```ts ` blocks and those will render highlighted in our generated API docs!

packages/lit-html/src/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Returns an iterable containing the result of calling `f(value)` on each
+ * value in `items`.
+ */
+export function* map<T>(
+  items: Iterable<T>,
+  f: (value: T, index: number) => unknown
+) {
+  const i = 0;
+  for (const value of items) {
+    yield f(value, i);
```
aomarks, line 30:
```suggestion
    yield f(value, i++);
```

?

.changeset/new-guests-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': minor
+---
+
+Add when, map, join, and range tempalte helpers.
```
aomarks, line 5:
```suggestion
Add `when`, `map`, `join`, and `range` template helpers.
```

aomarks (changes requested):

packages/lit-html/src/helpers.ts
justinfagnani: fixed

packages/lit-html/src/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+ *     ${when(this.user, () => html`User: ${this.user.username}, () => html`Sign In...`)}
+ *   `;
+ * }
+ * ```
+ */
+export const when = (
+  condition: boolean,
+  trueCase: () => unknown,
+  falseCase?: () => unknown
+): unknown => (condition ? trueCase() : falseCase?.());
```
aomarks, line 28: How about this to get stronger types?

```ts
function when<T>(condition: boolean, trueCase: () => T): T|undefined;
function when<T, F>(condition: boolean, trueCase: () => T, falseCase: () => F): T|F;
function when(
  condition: boolean,
  trueCase: () => unknown,
  falseCase?: () => unknown
) {
  return condition ? trueCase() : falseCase?.();
}
```

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAdwBYFMwB4AqA+ACggQBMZYEAuRAIzjgBt0BDMAGkSgCcR0BhZgGd01AgEpEAXjyIcY6jgA+4EumAww6EgG4AUKEgUkaTLg4AxQsTBkj1OoxbtOPfkJGJxUmTg7BmDMICwqIS0ojm8rKK5noG0PDGGGAEuoiI1raJ9vRMrGxpLrzBHl7h4ADWYHDI7IX+gW7CAPyh3oiV1bW6EgDehVzoUCBcSJnkiYjNRU3oXtQNQe7NAHTiegC+urrWglCIAB5SKMkE3LwcZTIAjJdhMgBEwPQPYtpAA

(Also maybe there's a more compact way to do that)

aomarks: Or even this???

```ts
function when<T>(condition: true, trueCase: () => T): T;
function when<T>(condition: false, trueCase: () => T): undefined;
function when<T>(condition: boolean, trueCase: () => T): T|undefined;
function when<T, F>(condition: true, trueCase: () => T, falseCase: () => F): T;
function when<T, F>(condition: false, trueCase: () => T, falseCase: () => F): F;
function when<T, F>(condition: boolean, trueCase: () => T, falseCase: () => F): T|F;
function when(
  condition: boolean,
  trueCase: () => unknown,
  falseCase?: () => unknown
) {
  return condition ? trueCase() : falseCase?.();
}
```

packages/lit-html/src/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+ * incrementing by `step`.
+ *
+ * If `start` is omitted, the range starts at `0`. `step` defaults to `1`.
+ *
+ * @example
+ *
+ * ```ts
+ * render() {
+ *   return html`
+ *     ${map(range(8), () => html`<div class="cell"></div>)}
```
aomarks, line 104:
```suggestion
 *     ${map(range(8), () => html`<div class="cell"></div>`)}
```

packages/lit-html/src/test/helpers_test.ts
```diff
@@ -0,0 +1,151 @@
...
+    container = document.createElement('div');
+  });
+
+  suite('when', () => {
+    test('true case, without false case', () => {
+      assertRender(
+        when(true, () => html`X`),
+        'X'
+      );
+    });
```
aomarks, line 30: Add a test that `` when(false, () => html`X`) `` returns `undefined`?

aomarks (changes requested):

packages/lit-html/src/test/helpers_test.ts
```diff
@@ -0,0 +1,151 @@
...
+    test('end < start', () => {
+      assert.deepEqual([...range(2, 1)], []);
+    });
+
+    test('custom step', () => {
+      assert.deepEqual([...range(0, 10, 3)], [0, 3, 6, 9]);
+    });
+
+    test('negative step', () => {
+      assert.deepEqual([...range(0, -3, -1)], [0, -1, -2]);
```
kevinpschaaf, line 148: Yikes, does e.g. `range(0, 10, -1)` this need any infinite-loop protection? At least in DEBUG mode?

kevinpschaaf: I'd suggest matching the behavior of range here: https://github.com/google/closure-library/blob/master/closure/goog/array/array.js#L1570

packages/lit-html/src/helpers.ts
justinfagnani: Oh nice. I didn't do this because it's unlikely to matter in template usages, but it's easy enough and nice to do anyway.

I got this as a refinement of option 2:
```ts
function when<T, F>(condition: true, trueCase: () => T, falseCase?: () => F): T;
function when<T, F = undefined>(condition: false, trueCase: () => T, falseCase?: () => F): F;
function when(
  condition: boolean,
  trueCase: () => unknown,
  falseCase?: () => unknown
) {
  return condition ? trueCase() : falseCase?.();
}
```

packages/lit-html/src/test/helpers_test.ts
justinfagnani: Added a case to this test. There was already one test for infinitely loop protection in the "end < start" case above.

The reason this already works is because of 1) the `<` test, which means that if `end < start` the loop breaks immediately, and 2) the switch of the operands for the `<` test when `step` is negative, which applies that same test correctly in that case. It's the ternary in here:

```ts
  for (let i = start; step > 0 ? i < end : end < i; i += step) {
  ```

File: packages/lit-html/src/helpers.ts (hunk 5 above)
Comment (aomarks) [reply packages/lit-html/src/helpers.ts]: This doesn't seem to work when the type is `boolean` -- only `true` or `false`:

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAdwBYFMwB4AqAaRAMQD4AKCBAExlgQC5EoAnEdA51gYQEMBndBqQCUiALzFE+RMG4Abfj34B+QSPFEhDHAG4AUKEi0kaTLgKExicJXTAYYdJTIUw1Iwxny2jFukUDEYTEJKU8FPnQVQLUJQk0iPQNoeGMMMFJdREQXNxSGACM4OFl0bjA8TJ8uCNVgqzAAazA4ZHLKsL8IqKD1cCaWsF0RAG9KpnQoECYkHJoUxCUqzv4gjzlw5QA6YT0AX11dF14oRG4ARgYwEABbfPQmSxN0jm8eiTOhPSOT7gAmBmstnsjkeaVIHQIb0QHy+CGOpwAzJcbncmAAfQF2ByUUGYUgvU68RCFYqlcrROowg7fYkXRBXW73XHPXyQmLQtl1ABEwCKXM+hzhJ3y-0QxyY9gA5szwetXuyzpz1Dy+QKafkkfSUfc0eKpTKCXxiUUSmUle9zYgVXB+dogA

File: packages/lit-html/src/test/helpers_test.ts (hunk 7 above)
Comment (aomarks) [reply packages/lit-html/src/test/helpers_test.ts:35]: (Sorry for nits)

I think this new `assertRender` should be in its own test function?

Also some of the descriptions might be a little unclear. E.g. there's "true case, with false case" and "false case", which are the same except for the first parameter. So I guess it's unclear to me if "case" is referring to the first parameter or the second two parameters.

Suggestions:

```
true with false case
false with false case
true without false case
false without false case
```

Or:

```
when(true, T)
when(false, T)
when(true, T, F)
when(false, T, F)
```

File: packages/lit-html/src/helpers.ts [hunk 9]
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * This is a convenience wrapper around a ternary expression that makes it a
+ * little nicer to write an inline conditional without an else.
+ *
+ * @example
+ *
+ * ```ts
+ * render() {
+ *   return html`
+ *     ${when(this.user, () => html`User: ${this.user.username}, () => html`Sign In...`)}
```
aomarks, line 19:
```suggestion
 *     ${when(this.user, () => html`User: ${this.user.username}`, () => html`Sign In...`)}
```

aomarks:
We had this discussion offline, writing up here:

From a usage perspective, these new helpers seem to be in the same category as directives. The difference is really just an implementation detail -- that these ones don't need to use the `Directive` class.

E.g. it could be difficult for users to remember that `repeat` is imported from `directives/repeat.js`, but `join` is imported from `helpers.js`. To users who aren't familiar with the implementations, there would probably be no obvious reason why we've organized them this way.

Some options:

1. Keep the distinction. Helpers implemented with `Directive` are imported from `directives/<directive>.js`, otherwise from `helpers.js` (except for `ifDefined` which doesn't use `Directive` but is already under `directives/`).

2. Grant these new helpers the "directive" title, even though they technically aren't.

3. Move existing directives _and_ these new helpers into a new `helpers/` or similar folder, and add re-exports to make non-breaking. "Helpers" becomes our broad term for template-utility-functions, and "directives" are for the specific type of helper that needs deeper control into how Lit renders.

aomarks (changes requested):

graynorton:
I like this option:

> 3. Move existing directives _and_ these new helpers into a new `helpers/` or similar folder, and add re-exports to make non-breaking. "Helpers" becomes our broad term for template-utility-functions, and "directives" are for the specific type of helper that needs deeper control into how Lit renders.

packages/lit-html/src/test/helpers_test.ts
justinfagnani: fixed

justinfagnani:
Moved the helpers into directives/

aomarks (approved):
LGTM!

packages/lit-html/src/test/helpers_test.ts
kevinpschaaf: 👍

justinfagnani:
@kevinpschaaf are you good with this now?

packages/lit-html/src/directives/when.ts
```diff
@@ -0,0 +1,45 @@
...
+  trueCase: () => T,
+  falseCase?: () => F
+): T;
+export function when<T, F = undefined>(
+  condition: false,
+  trueCase: () => T,
+  falseCase?: () => F
+): F;
+export function when<T, F = undefined>(
+  condition: boolean,
```
kevinpschaaf, line 35: Type-wise, do we _only_ want to support `boolean` here and not truthy/falsey? I'd probably argue for it being looser, so it's not worse than using ternary.

justinfagnani: changed

kevinpschaaf (approved):

Issue #2317, opened 2021-11-25
[@lit-labs/task] Task function never call if there are no deps


justinfagnani:
# Description

There are two related problems with using Task with no deps (an always executed once task):
  1. new Task() doesn't support being called with no deps function
  2. A deps function of `() => []` causes the task function to never be called

# Steps to Reproduce

1. Write this code

```ts
  new Task(this, fn, () => [])
```

2. See this output...

## Live Reproduction Link
[...]

PR #2336, opened 2021-12-05
[@lit-labs/task] Refines when tasks are run

### LINKED ISSUE #2323: [@lit-labs/task] Tasks that don't start immediately
It'd be useful to be able to define a task that doesn't start immediately. The task could start on some action / event, or when some other async process is done.

I could imagine two non-exclusive ways to do this:
  1. Have a method to start a task, along with an option to not immediately run a task. That option could be a particular set of deps, as in https://github.com/lit/lit/issues/2317. Maybe `() => []` always runs the task, and `undefined` creates a manually run task.
  2. Support Promises as deps and only run a task when the promises resolve. This could be tricky when dealing with Promise identity (new Promise instances wrapping the same underlying value) and needing to always wait a microtask to see the Promise value. Some of the same issues that `Task` solves vs `until()`.

sorvell:
Fixes #2323 and #2317.
* Tasks automatically run whenever their arguments change. A task with no arguments does not run; a task with an empty array of arguments runs once; otherwise tasks runs when its arguments array members differs from is previous arguments. 
* The automatic running of tasks based on arguments can be disabled by setting `autoRun` to `false`. This can be done via the task config or by setting the property on the task itself.
* Call `run()` (optionally specifying task arguments) to explicitly run the task.

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -151,12 +151,30 @@ export class Task<T extends [...unknown[]] = any, R = any> {
 
   protected async performTask() {
     const args = this._getArgs?.();
-    const argsDirty = this._argsDirty(args);
-    if (this.autoRun && argsDirty) {
+    if (this.shouldRun(args)) {
       this.run(args);
     }
   }
 
+  /**
+   * Determines if the task should run when it's triggered as part of the
+   * host's reactive lifecycle. Note, this is not checked when `run` is
+   * explicitly called. A task runs automatically when `autoRun` is `true` and
+   * either its arguments change.
+   * @param args The task's arguments
+   * @returns
+   */
+  protected shouldRun(args?: T) {
+    return this.autoRun && this._argsDirty(args);
+  }
+
+  /**
+   * A task runs when its arguments change, as long as the `autoRun` option
+   * has not been set to false. To explicitly run a task outside of these
+   * conditions, call `run`. A custom set of arguments can optionally be passed
+   * and if not given, the configured arguments are used.
+   * @param args optional set of arguments to use for this task run
+   */
   async run(args?: T) {
     args ??= this._getArgs?.();
     if (
@@ -177,8 +195,7 @@ export class Task<T extends [...unknown[]] = any, R = any> {
     this._host.requestUpdate();
     const key = ++this._callId;
     try {
-      // TODO(sorvell): fix type so args can be optional.
-      result = await this._task(args as unknown as T);
+      result = await this._task(args!);
     } catch (e) {
       error = e;
     }
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -10,7 +10,13 @@ import {ReactiveControllerHost} from '@lit/reactive-element/reactive-controller.
 export type TaskFunction<D extends [...unknown[]], R = any> = (
   args: D
 ) => R | typeof initialState | Promise<R | typeof initialState>;
-export type DepsFunction<D extends [...unknown[]]> = () => D;
+export type ArgsFunction<D extends [...unknown[]]> = () => D;
+
+export type CanRunFunction = (
+  canRun: () => boolean,
+  status: TaskStatus,
+  args?: unknown[]
+) => boolean;
 
 /**
  * States for task status
@@ -36,6 +42,12 @@ export type StatusRenderer<R> = {
   complete?: (value: R) => unknown;
   error?: (error: unknown) => unknown;
 };
+
+export interface TaskConfig<T extends unknown[], R> {
+  task: TaskFunction<T, R>;
+  args?: ArgsFunction<T>;
+  canRun?: CanRunFunction;
+}
 
 // TODO(sorvell): Some issues:
 // 1. When task is triggered in `updated`, this generates a ReactiveElement
@@ -65,14 +77,20 @@ export type StatusRenderer<R> = {
  * object with optional corresponding state method to easily render values
  * corresponding to the task state.
  *
+ * The task is run automatically when its arguments change; however, this can
+ * be customized by passing a `canRun` function. It receives arguments of
+ * a default `canRun()` function, the task's status, and the values of the
+ * task's arguments.
+ *
  * class MyElement extends ReactiveElement {
  *   url = 'example.com/api';
  *   id = 0;
  *   task = new Task(
- *     this,
- *     ([url, id]) =>
- *       fetch(`${this.url}?id=${this.id}`).then(response => response.json()),
- *     () => [this.id, this.url]
+ *     this, {
+ *       task: ([url, id]) =>
+ *         fetch(`${this.url}?id=${this.id}`).then(response => response.json()),
+ *       args: () => [this.id, this.url]
+ *     }
  *   );
  *
  *   update(changedProperties) {
@@ -86,13 +104,15 @@ export type StatusRenderer<R> = {
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export class Task<T extends [...unknown[]] = any, R = any> {
-  private _previousDeps: T = [] as unknown as T;
+  private _previousArgs: T = [] as unknown as T;
   private _task: TaskFunction<T, R>;
-  private _getDependencies: DepsFunction<T>;
+  private _args: ArgsFunction<T>;
   private _callId = 0;
   private _host: ReactiveControllerHost;
   private _value?: R;
   private _error?: unknown;
+  // Internal signal for being fully done processing a task, including host updates.
+  private _isIdle = true;
   status: TaskStatus = TaskStatus.INITIAL;
 
   /**
@@ -107,13 +127,21 @@ export class Task<T extends [...unknown[]] = any, R = any> {
 
   constructor(
     host: ReactiveControllerHost,
-    task: TaskFunction<T, R>,
-    getDependencies: DepsFunction<T>
+    task: TaskFunction<T, R> | TaskConfig<T, R>,
+    args?: ArgsFunction<T>
   ) {
     this._host = host;
     this._host.addController(this);
+    if (typeof task === 'object') {
+      const taskConfig = task as TaskConfig<T, R>;
+      args = taskConfig.args;
+      task = taskConfig.task;
+      if (taskConfig.canRun !== undefined) {
+        this._canRun = taskConfig.canRun;
+      }
+    }
     this._task = task;
-    this._getDependencies = getDependencies;
+    this._args = args || (((_x: unknown) => []) as unknown as ArgsFunction<T>);
     this.taskComplete = new Promise((res, rej) => {
       this._resolveTaskComplete = res;
       this._rejectTaskComplete = rej;
@@ -121,12 +149,15 @@ export class Task<T extends [...unknown[]] = any, R = any> {
   }
 
   hostUpdated() {
-    this._completeTask();
+    this.runTask();
   }
 
-  private async _completeTask() {
-    const deps = this._getDependencies();
-    if (this._isDirty(deps)) {
+  protected async runTask() {
+    const args = this._args();
+    const argsDirty = this._argsDirty(args);
+    const canRun = () => this.defaultCanRun(argsDirty, args);
+    if (this._canRun(canRun, this.status, args)) {
+      this._isIdle = false;
       if (
         this.status === TaskStatus.COMPLETE ||
         this.status === TaskStatus.ERROR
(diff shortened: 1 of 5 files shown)


justinfagnani:
I still like the idea that a Task with no args function is no auto-run at all, and a task with an empty args array is auto-run because there's no args to dirty-check:

```ts
new Task(host, fn); // not auto-run
new Task(host, fn, () => []); // auto-run
```

`canRun` is still a good feature, but I'm not sure a default of running every host update is what users would expect from no providing args function, and it _looks_ like it's very involved to make a manually run task because `canRun` needs to vary returning `true` or `false` since it's checked even in the manual call to `runTask()`. So I think we'd still need a simple way to make a manually-run Task.

Is this how we'd have to do it with `canRun` now?
```ts
  private _task = new Task(this, {task: async() => {...}, canRun: () => this._canRunTask});
  private @state _canRunTask = false;
  private async _runTask() {
    this._canRunTask = true;
    this._task.runTask();
    await this._task.taskComplete;
    this._canRunTask = false;
  }
  
  private _onRunButtonClick() {
    this._runTask();
  }
}
```

I'd like to see something more like:
```ts
  private _task = new Task(this, async() => {...});
  
  private _onRunButtonClick() {
    this._task.run();
  }
}
```

packages/labs/task/src/task.ts
```diff
@@ -107,26 +127,37 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
 
   hostUpdated() {
-    this._completeTask();
+    this.runTask();
   }
 
-  private async _completeTask() {
-    const deps = this._getDependencies();
-    if (this._isDirty(deps)) {
+  protected async runTask() {
```
justinfagnani, line 155: Can this be made public and named `run()`? Or is there another public method to manually run a task?

packages/labs/task/src/task.ts
```diff
@@ -86,13 +104,15 @@ export type StatusRenderer<R> = {
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export class Task<T extends [...unknown[]] = any, R = any> {
-  private _previousDeps: T = [] as unknown as T;
+  private _previousArgs: T = [] as unknown as T;
   private _task: TaskFunction<T, R>;
-  private _getDependencies: DepsFunction<T>;
+  private _args: ArgsFunction<T>;
```
justinfagnani, line 109: It'd still prefix this with `_get` since it's a function to get the args, and not an array of the args like `_previousArgs`

packages/labs/task/src/task.ts
```diff
@@ -86,13 +104,15 @@ export type StatusRenderer<R> = {
...
   private _callId = 0;
   private _host: ReactiveControllerHost;
   private _value?: R;
   private _error?: unknown;
+  // Internal signal for being fully done processing a task, including host updates.
+  private _isIdle = true;
```
justinfagnani, line 115: Shouldn't idle always correspond to one of the existing TaskStatus's? Like `idle = this.status !== TaskStatus.PENDING`

justinfagnani (changes requested):

sorvell:
Thinking about this again, I think probably `canRun` should be removed because it's a bit cumbersome as you pointed out. I think instead:
* tasks without args should never "auto" run
* we should have something like `disableAutoRun` to make tasks with no args (empty or not provided) never auto-run (auto-run with no args is problematic since we signal status changes with requestUpdate)
* expose `run()` to manually run.

packages/labs/task/src/task.ts
sorvell: Removed.

packages/labs/task/src/task.ts
sorvell: Done.

packages/labs/task/src/task.ts
sorvell: Done.

packages/labs/task/src/test/task_test.ts
```diff
@@ -308,4 +325,41 @@ suite('Task', () => {
     await new Promise((r) => setTimeout(r, 0));
     assert.equal(el.task.status, TaskStatus.INITIAL, 'new initial');
   });
+
+  test('task without args runs when not pending', async () => {
+    assert.equal(el.noArgsTaskValue, undefined);
```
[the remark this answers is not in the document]
sorvell: Done.

packages/labs/task/src/task.ts
```diff
@@ -102,71 +113,92 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
-          this.status = TaskStatus.INITIAL;
+  protected async performTask() {
+    const args = this._getArgs?.();
+    const argsDirty = this._argsDirty(args);
+    if (this.autoRun && argsDirty) {
+      this.run(args);
+    }
+  }
+
+  async run(args?: T) {
```
justinfagnani, line 160: oh, nice to include `args` here. Should document that it defaults to the args fn.

justinfagnani: Add jsdoc

justinfagnani: Should there be an option that controls whether the task is always run here, canceling a pending task, or if it is a no-op for a pending task? Similar to `shouldRun` in the auto-run case I guess.

packages/labs/task/src/task.ts
```diff
@@ -102,71 +113,92 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
+    this._error = undefined;
+    this._value = undefined;
+    let result!: R | typeof initialState;
+    let error: unknown;
+    // Request an update to report pending state.
+    this._host.requestUpdate();
+    const key = ++this._callId;
+    try {
+      // TODO(sorvell): fix type so args can be optional.
+      result = await this._task(args as unknown as T);
```
justinfagnani, line 181: Does this work? `this._task(args!)`

packages/labs/task/src/test/task_test.ts
```diff
@@ -137,27 +103,129 @@ suite('Task', () => {
...
+    el.requestUpdate();
+    await tasksUpdateComplete();
     assert.equal(el.task.status, TaskStatus.INITIAL);
-    assert.equal(el.taskController.value, undefined);
-    assert.equal(el.taskController.status, TaskStatus.INITIAL);
+    assert.equal(el.taskValue, undefined);
   });
 
-  test('pending and complete status', async () => {
+  test('tasks with args run initially', async () => {
```
justinfagnani, line 116: I think you need a test for `{args: () => []}`. Seems like that might not run with the current implementation of `_argsDirty`.

packages/labs/task/src/task.ts
```diff
@@ -102,71 +113,92 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
+      args = taskConfig.args;
+      task = taskConfig.task;
+      if (taskConfig.autoRun !== undefined) {
+        this.autoRun = taskConfig.autoRun;
+      }
+    }
     this._task = task;
-    this._getDependencies = getDependencies;
+    this._getArgs = args;
     this.taskComplete = new Promise((res, rej) => {
```
justinfagnani, line 142: Should this change for manual run tasks so that it's a resolved promise?

packages/labs/task/src/task.ts
```diff
@@ -194,16 +226,11 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
-      if (notEqual(dep, previousDeps[i])) {
-        return true;
-      }
-      i++;
-    }
-    return false;
+  private _argsDirty(args?: T) {
+    const prev = this._previousArgs;
+    this._previousArgs = args;
+    return Array.isArray(args) && Array.isArray(prev)
```
justinfagnani, line 232: I think if `_previousArgs` is undefined (or maybe state is initial), this should always return true.

packages/labs/task/src/test/task_test.ts
```diff
@@ -137,27 +103,129 @@ suite('Task', () => {
...
 
-  test('error status', async () => {
+  test('tasks with args run when args change', async () => {
+    const el = getTestElement({args: () => [el.a, el.b]});
+    await renderElement(el);
+    el.resolveTask();
+    await tasksUpdateComplete();
+    assert.equal(el.task.status, TaskStatus.COMPLETE);
+    assert.equal(el.taskValue, `a,b`);
+    el.a = 'a1';
```
justinfagnani, line 134: it'd be nice to break up the major steps of the test with comments.

packages/labs/task/src/test/task_test.ts
```diff
@@ -137,27 +103,129 @@ suite('Task', () => {
...
     await tasksUpdateComplete();
     assert.equal(el.task.status, TaskStatus.COMPLETE);
-    assert.equal(el.taskController.status, TaskStatus.COMPLETE);
-    assert.equal(el.taskValue, 'result: foo, bar');
-    assert.equal(el.taskControllerValue, 'result: 0');
+    assert.equal(el.taskValue, `a,b`);
   });
 
-  test('error status', async () => {
+  test('tasks with args run when args change', async () => {
```
justinfagnani, line 127: It'd be good to also have a basic test that tasks don't run when args haven't changed.

justinfagnani (changes requested):

packages/labs/task/src/task.ts
sorvell: Doc added. Let's consider those other options in a follow-on issue.

packages/labs/task/src/task.ts
sorvell: Yup, done.

packages/labs/task/src/task.ts
sorvell: Added protected `shouldRun` which checks initial state.

packages/labs/task/src/test/task_test.ts
sorvell: Added.

packages/labs/task/src/test/task_test.ts
sorvell: Added.

packages/labs/task/src/test/task_test.ts
sorvell: Done.

packages/labs/task/src/task.ts
```diff
@@ -10,7 +10,7 @@ import {ReactiveControllerHost} from '@lit/reactive-element/reactive-controller.
 export type TaskFunction<D extends [...unknown[]], R = any> = (
   args: D
 ) => R | typeof initialState | Promise<R | typeof initialState>;
-export type DepsFunction<D extends [...unknown[]]> = () => D;
+export type ArgsFunction<D extends [...unknown[]]> = () => D;
```
justinfagnani, line 13: this is a breaking change btw, and maybe the only one. Do you want to export a `DepsFunction` alias so we don't have to release a new major?

packages/labs/task/src/task.ts
```diff
@@ -102,71 +113,109 @@ export class Task<T extends [...unknown[]] = any, R = any> {
...
+  autoRun = true;
+
   private _resolveTaskComplete!: (value: R) => void;
   private _rejectTaskComplete!: (e: unknown) => void;
 
   constructor(
     host: ReactiveControllerHost,
-    task: TaskFunction<T, R>,
-    getDependencies: DepsFunction<T>
+    task: TaskFunction<T, R> | TaskConfig<T, R>,
```
justinfagnani, line 127: Let's write this as two overloads: one with task and getArgs and the other with just a config.

packages/labs/task/src/task.ts
justinfagnani: ah, right

packages/labs/task/src/task.ts
sorvell: Done.

packages/labs/task/src/task.ts
sorvell: Can you explain why?

packages/labs/task/src/task.ts
sorvell: Done.

packages/labs/task/src/task.ts
justinfagnani: Just wondering what `await task.taskComplete` should do for a manually run task? The task is idle, but it's not necessarily ever going to run. In some ways it seems like a task that has completed some times, and might or might not run again, and in that case `taskComplete` would be a resolved promise.

justinfagnani (approved):

packages/labs/task/src/task.ts
sorvell: Let's handle this as a follow on. Filed https://github.com/lit/lit/issues/2347.

Issue #2165, opened 2021-09-17
[lit-html] Add a "keyed" directive


straversi:
## Description

Sometimes it is useful to cache a node against a certain key, invalidating it when the key changes. Here is a concrete example:

```ts
render() {
  return html`<div ${flip(...)}>${this.name}</div>`
}
```

I want to use `flip` to "animate-in" this div whenever name changes. This won't work unless the div is distinct for each "name" change.

A workaround is to use `repeat` with a single-item array and the key function. However, this is cumbersome. A standalone "keyed" directive would help a lot!

(👋😃)

PR #2337, opened 2021-12-06
[lit-html] Add keyed directive

justinfagnani:
Adds a `keyed(key, value)` directive that clears a part if the key changes. This is useful for ensuring stateful DOM is cleared and with animations that require adding or removing elements for fly-in / fly-out animations.

Fixes https://github.com/lit/lit/issues/2165


packages/lit-html/src/directives/keyed.ts
```diff
@@ -0,0 +1,37 @@
...
+      setCommittedValue(part);
+      this.key = k;
+    }
+    return v;
+  }
+}
+
+/**
+ * Associates a renderable value with a unique key. When the key changes, the
+ * part is cleared forcing a new render, even if the value - such as a template
```
kevinpschaaf, line 30:
```suggestion
 * previous DOM is removed and disposed before rendering the next value, even if the value - such as a template
```

I think this clarifies the exact behavior better?

packages/lit-html/src/test/directives/keyed_test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const div = container.firstElementChild;
+    assert.equal(stripExpressionMarkers(container.innerHTML), '<div></div>');
+    assert.equal((div as any).foo, 1);
+
+    // Rerendering with same key should reuse the DOM
+    go(1);
+    const div2 = container.firstElementChild;
+    assert.equal(stripExpressionMarkers(container.innerHTML), '<div></div>');
+    assert.equal((div2 as any).foo, 1);
+    assert.strictEqual(div, div2, 'X');
```
kevinpschaaf, line 34: The `X` here is extraneous, right?

packages/lit-html/src/directives/keyed.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  render(k: unknown, v: unknown) {
+    this.key = k;
+    return v;
+  }
+
+  override update(part: ChildPart, [k, v]: DirectiveParameters<this>) {
+    if (k !== this.key) {
+      // Clear the part before returning a value. This forces a re-render.
+      setCommittedValue(part);
```
kevinpschaaf, line 21: Man, had to look up the API for what no 2nd argument does. I remember now we defaulted it to a RESET_VALUE sigil that ensures any dirty check will fail. Maybe clarify in the comment that omitting the 2nd arg is what forces a re-render.

kevinpschaaf (approved):
Small nits.

packages/lit-html/src/test/directives/keyed_test.ts
justinfagnani: Oh yeah, it's an assertion label for debugging. Will remove.

packages/lit-html/src/directives/keyed.ts
justinfagnani: done

PR #2341, opened 2021-12-07
Add choose() directive

justinfagnani:


packages/lit-html/src/directives/choose.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Chooses and evaluates a template function from a list based on matching
+ * the given `value` to a case.
+ *
+ * Cases are structured as `[caseValue, func]`. `value` is matched to
+ * `caseValue` by strict equality. The first match is selected.
+ *
+ * This is similar to a switch statement, but as an expression and without
+ * fallthrough.
```
aomarks, line 15: Example code for docs?

packages/lit-html/src/directives/choose.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export const choose = <T, V>(
+  value: T,
+  cases: Array<[T, () => V]>,
+  defaultCase?: () => V
+) => {
+  for (const c of cases) {
+    const caseValue = c[0];
+    const fn = c[1];
```
aomarks, line 24: Could put the `const fn = ` inside the `if` case, since it's unused otherwise.

packages/lit-html/src/directives/choose.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * This is similar to a switch statement, but as an expression and without
+ * fallthrough.
+ */
+export const choose = <T, V>(
+  value: T,
+  cases: Array<[T, () => V]>,
+  defaultCase?: () => V
+) => {
+  for (const c of cases) {
```
aomarks, line 22: For string cases, this could be an object index. But I assume it's an array search because we want to support more complex keys?

aomarks (approved):
LGTM, just a couple small comments

packages/lit-html/src/directives/choose.ts
```diff
@@ -0,0 +1,30 @@
...
+    const caseValue = c[0];
+    const fn = c[1];
+    if (caseValue === value) {
```
mohe2015, line 25: On phone right now but maybe use the compare function that's also used in lit-labs/task which can also compare NaN?

packages/lit-html/src/directives/choose.ts
justinfagnani: yeah... and we're matching Wit's implementation for now. We could add an overload for objects for a nicer syntax syntax for string keys.

packages/lit-html/src/directives/choose.ts
justinfagnani: We don't use `Object.is()` in core yet because it requires another polyfill in IE - which we still support. We could use it here without it technically being a breaking change, but it would be harder to describe which polyfills we require.

So I think that we probably don't want to do that and if you need to equality compare against numbers that might be `NaN` that you either convert the number to a sentinel first, or use some other control-flow construct. Maybe ternary before the `choose()`. It also seems to me like if you're dealign with `NaN` you might want to be using inequalities which would rule out `choose()` anyway.

These new helpers are simple and meant to cover many but not all use cases, and we hope people can write their own if they need to customize their behavior.

In the future I could see a `match()` directive that would allow for any comparison:

```ts
match(x, [
  [(x) => x < 0, html`negative`],
  [(x) => x > 0, html`positive`],
  [(x) => x === 0, html`zero`],
  [(x) => Object.is(x, Nan), html`oops`],
])
```

mohe2015: I thought about https://github.com/lit/lit/blob/fcc2b3d0054e69e6f76588ea9f440117b6d0deed/packages/reactive-element/src/reactive-element.ts#L280 but I agree that this is probably overkill as this directive is mostly a convenience for the common use cases.

PR #2343, opened 2021-12-08
Add a license header to the `keyed` directive.

bicknellr:
This file is missing a license header.


justinfagnani (approved):
oops! good catch

Issue #2321, opened 2021-11-27
[ssr] No easy interop with 3rd party custom elements


shakyShane:
# Description

If I load into memory (via a bundler for example) a 3rd-party custom element script such as [@github/clipboard-copy-element](https://www.npmjs.com/package/@github/clipboard-copy-element) - Lit will not render its attributes since it cannot find a suitable elementRenderer.

This is a huge foot-gun since a user of my framework would expect Lit to simply skip over something it cannot render (which it does) but to preserve it's inline attributes in the resulting HTML. 

It occurs here:

https://github.com/lit/lit/blob/dcab56b0eeef6e05969c3682d3275a30ddf18d97/packages/labs/ssr/src/lib/render-lit-html.ts#L695-L699

# Steps to Reproduce

1. Write this code

```js
import "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import "@github/clipboard-copy-element";
import { render } from "@lit-labs/ssr";
import { html } from "lit";

const input = html`
  <clipboard-copy for="login" shane="here">
    Copy
    <span class="notice" hidden>Copied!</span>
  </clipboard-copy>
`;

for (let string of render(input)) {
  console.log(string);
}
```

Output: 

```html
<!--lit-part uzegxRF9tt8=-->

  <clipboard-copy  
>

    Copy
    <span class="notice" hidden>Copied!</span>
  </clipboard-copy>

<!--/lit-part-->

```

Is there a go-to solution for this problem, outside of a custom elementRenderer?

Thanks :)

shakyShane:
Actually, even a custom element renderer will not work for custom elements that do work inside of their constructors.

shakyShane:
For the example above, the following works and allows the custom-element to be server-rendered with it's attributes.

```ts
class MyRenderer extends ElementRenderer {
  constructor(tagName) {
    super(tagName);
    const attrs = [];
    this.element = {
      attributes: attrs,
      getAttribute: (qualifiedName) => {
        return attrs.find((x) => x.name === qualifiedName);
      },
      setAttribute: (qualifiedName, value) => {
        attrs.push({ name: qualifiedName, value });
      },
    };
  }
  attributeChangedCallback(name, old, value) {
    // no-op
  }

  static matchesClass(arg1, arg2) {
    return true;
  }
}
```

justinfagnani:
I think this is just a bug in that if there is no renderer available, we should always write attributes.

shakyShane:
[...]

PR #2344, opened 2021-12-08
[ssr] Fix unknown element attribute rendering

aomarks:
Fixes a bug where static attributes on custom elements would not render if the custom element was unregistered or did not have a renderer.

I considered a few ways to solve this and picked [2]. Open to switching to [1] or doing something else:

1. Don't emit `custom-element-*` opcodes at all for un-renderable elements. However, this would make `getTemplateOpcodes` dependent on `renderInfo`, and it seemed kind of nice to preserve the feature of `getTemplateOpcodes` being only a function of the template. If we did this, seems like we'd want to add the renderer `instance` to the `custom-element-open` opcode, so that we don't look up the renderer twice.

2. Add a new `FallbackRenderer` class which just tracks and renders attributes. **This is what I picked**, since it seemed pretty undisprutive, and it's sort of nice to not have to handle the `undefined` case as specially in the opcode render loop.

3. Add special tracking for unrenderable elements using a bag of attributes directly in the `customElementInstanceStack` stack, and add fallback logic to render static attributes in the `custom-element-attributes` switch case. This just seemed like a less clean way of writing [2].

Fixes https://github.com/lit/lit/issues/2321

### DIFF
--- a/.changeset/khaki-squids-join.md
+++ b/.changeset/khaki-squids-join.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix bug where static attributes did not render for unknown elements
--- a/packages/labs/ssr/src/lib/element-renderer.ts
+++ b/packages/labs/ssr/src/lib/element-renderer.ts
@@ -28,10 +28,10 @@ export const getElementRenderer = (
   tagName: string,
   ceClass: typeof HTMLElement | undefined = customElements.get(tagName),
   attributes: AttributesMap = new Map()
-): ElementRenderer | undefined => {
+): ElementRenderer => {
   if (ceClass === undefined) {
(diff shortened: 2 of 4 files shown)


justinfagnani (approved):
🎉

kevinpschaaf (approved):
That reminds me, another reason to go for (2) over (1) is that it preserves the ability for an element renderer to match based on per-instance attributes (we pass `op.staticAttributes` for the instance to the `getElementRenderer` call at opcode execution time); looking up the renderer at opcode generation time would remove that flexibility. I thought we might use that for e.g. a `client-side-only` attribute to choose a `FallbackRenderer` that doesn't SSR).

So this seems good, at least until we refine the `ElementRenderer` matching concept and and how we decide to do client-side only components.

Issue #1998, opened 2021-07-09
Bring escape-html into labs/ssr?


matthewp:
I've been experimenting with getting LitElements to render in Deno and am mostly able to get it to work (importing with Skypack URLs). The only blocker as far as I can tell is the usage of the Node.js `'module'` module for the sake of requiring 'escape-html'.

I can understand not wanting to take on the maintenance burden of officially supporting Deno, but you could pretty easily unofficially support it just by bringing escape-html into the codebase. It's an old module that hasn't been updated in a long time and is a single small function which is unlikely to need changes in the future, so I think it's a good candidate for vendoring. 

Happy to send a PR if you're agreeable.

matthewp:
This would likely help solve https://github.com/lit/lit/issues/1983 as well.

PR #2346, opened 2021-12-08
[ssr] Replace escape-html dependency with our own version

aomarks:
The `escape-html` package is not an ES module, and it's a pretty trivial function to just write ourselves. See https://unpkg.com/browse/escape-html@1.0.3/index.js for reference. This new implementation should have the exact same output.

Advantages: getting rid of the only non-esm dependency helps with running SSR in deno and service-workers, and drops our dependency surface area and package size.

Fixes https://github.com/lit/lit/issues/1998

### DIFF
--- a/packages/labs/ssr/src/lib/util/escape-html.ts
+++ b/packages/labs/ssr/src/lib/util/escape-html.ts
@@ -1,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const replacements = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  // Note &apos; was not defined in the HTML4 spec, and is not supported by very
+  // old browsers like IE8, so a codepoint entity is used instead.
+  "'": '&#39;',
+};
+
+/**
+ * Replaces characters which have special meaning in HTML (&<>"') with escaped
+ * HTML entities ("&amp;", "&lt;", etc.).
+ */
+export const escapeHtml = (str: string) =>
+  str.replace(
+    /[&<>"']/g,
+    (char) => replacements[char as keyof typeof replacements]
+  );
--- a/.changeset/fresh-eggs-check.md
+++ b/.changeset/fresh-eggs-check.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Remove dependency on escape-html (which is not an ES module)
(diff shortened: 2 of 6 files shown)


packages/labs/ssr/src/lib/util/escape-html.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Note &apos; was not defined in the HTML4 spec, and is not supported by very
+  // old browsers like IE8, so a codepoint entity is used instead.
+  "'": '&#39;',
+};
+
+/**
+ * Replaces characters which have special meaning in HTML (&<>"') with escaped
+ * HTML entities ("&amp;", "&lt;", etc.).
+ */
+export const escapeHtml = (str: string) =>
```
justinfagnani, line 21: Since we use this in quoted attribute value position can we double-check that this is the right escape algorithm for that? I seem to recall there are difference escape requirements in different contexts, but don't know exactly off the top of my head.

justinfagnani (approved):

packages/labs/ssr/src/lib/util/escape-html.ts
```diff
@@ -0,0 +1,25 @@
...
+  str.replace(
+    /[&<>"']/g,
+    (char) => replacements[char as keyof typeof replacements]
```
AndrewJakubowicz, line 24: This is really elegant.

AndrewJakubowicz (approved):
Missing domain knowledge, but syntactically this LGTM!

Is this code in a tested path?

packages/labs/ssr/src/lib/util/escape-html.ts
aomarks: Testing locally and some stack overflow posts seem to support this scheme. I do wonder why we need to quote anything except for & and ", though, since it seems that [<>'] should be safe inside an " attribute string.

aomarks: The spec isn't super precise about this, it seems

https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

```
Attribute values are a mixture of text and character references, except with the additional restriction that the text cannot contain an ambiguous ampersand.
```

The definition of text:
```
Text is allowed inside elements, attribute values, and comments. Extra constraints are placed on what is and what is not allowed in text based on where the text is to be put, as described in the other sections.
```

But maybe there's another part of the spec I should be looking at.

justinfagnani: It's ok. This keeps the current behavior. Just something that crossed my mind with your fallback renderer PR.

aomarks:
> Is this code in a tested path?

Yep, it's in the code path of most rendering, so it's covered by most tests. Confirmed that breaking the implementation fails some tests.

PR #2356, opened 2021-12-11
Testing sauce labs fix

AndrewJakubowicz:
Increase the test stability of Lit core.

This PR was prompted by the long running sauce ie11 test failures. This has prevented Lit from being green for many months. There are three known issues with sauce labs and I've addressed two of them in this change.

The three issues are:
 1. A polyfill test failure when testing ie11 due to scoped element registry not supporting ie11.
 2. Random test file failures in sauce-ie11 with error `Browser tests did not start after 60000ms`.
 3. Sauce proxy connection failure.

The first issue (1.) was addressed by doing a feature check on `window.ShadowRoot && (window as any).ShadowRootInit;` and optionally running `suite.skip`.

The second issue was addressed by increasing `testsStartTimeout`. This is required because `testsStartTimeout` is measured from when the test file is enqueued. In ie11 where we test sequentially, this means the total time of the whole test suite must start within `testsStartTimeout`. Running tests sequentially takes about 2 minutes, so we'd constantly reach our 1 minute timeout on tests that had been waiting on the queue for a long time.

All other fixes are extremely minor.

### Validation

This was validated and tested locally, and then checked against Github Actions.


AndrewJakubowicz:
Ported over changes that seemed to improve the testing situation to https://github.com/lit/lit/pull/2364.

Still lots of room for improvement!

PR #2360, opened 2021-12-14
Improve docs to more clearly link to the options interface for queryAssignedNodes

AndrewJakubowicz:
Raised in https://github.com/lit/lit.dev/pull/623

This change makes the link to the options object clearer for both queryAssignedElements and queryAssignedNodes.

Initially wanted to link from the `@param options` jsdoc, but it's not yet supported on Lit.dev.

It's easier to document this currently with the TypeScript types and a link in the description documentation. Once https://github.com/lit/lit.dev/issues/503 is addressed we can update this with `@link` and `@param` providing the best of both worlds. 

Tested manually with lit.dev api generation.

This now shows up like so and the link can be clicked. Information is de-duped:
Screen Shot 2021-12-14 at 2 33 13 PM

### DIFF
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -16,10 +16,16 @@ import {decorateProperty} from './base.js';
 import type {ReactiveElement} from '../reactive-element.js';
 import type {QueryAssignedNodesOptions} from './query-assigned-nodes.js';
 
+/**
+ * Options for the {@link queryAssignedElements} decorator. Extends from the
+ * options that can be passed into
+ * [HTMLSlotElement.assignedElements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements).
+ */
 export interface QueryAssignedElementsOptions
   extends QueryAssignedNodesOptions {
   /**
-   * CSS selector used to filter the elements returned.
+   * CSS selector used to filter the elements returned. For example, a selector
+   * of `".item"` will only include elements with the `item` class.
    */
   selector?: string;
 }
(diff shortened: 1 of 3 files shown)


aomarks (approved):
Nice!

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -50,9 +56,7 @@ export interface QueryAssignedElementsOptions
  * Note, the type of this property should be annotated as `Array<HTMLElement>`.
  *
  * @param options Object that sets options for nodes to be returned. See
- *     [MDN parameters section](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements#parameters)
- *     for available options. Also accepts two more optional properties,
- *     `slot` and `selector`.
+ *     {@link QueryAssignedElementsOptions} for all available options.
  * @param options.slot Name of the slot. Undefined or empty string for the
```
justinfagnani, line 60: Should we even keep this style of deep property docs when we should have docs on the QueryAssignedElementsOptions interface? Seems like this is duplication that will need to be synced.

AndrewJakubowicz: It's true. It also makes for a lot of info duplication. Removing the `options.slot`.

AndrewJakubowicz: Done!

AndrewJakubowicz:
I de-duped the information, and also made sure that it generates correctly on Lit.dev. In the generated docs there is now a link to the interface which contains more context.

justinfagnani (approved):

aomarks (approved):
Great!

Issue #2272, opened 2019-03-07
[lit] Static build for CDN distribution


slightlyoff:
Per a F2F conversation with @justinfagnani, wanted to capture agreement to add a single, modern-browser-exclusive build of LitElement and the most frequently used lit-html directives to the distributed NPM module. The goal is to enable easy inclusion of LitElement from a single source file.

## Rationale

LitElement, when built, is tiny! That isn't apparent when using it from, e.g., unpkg. The need for the Polymer CLI tools also increases the friction to getting started with the edit/compile cycle that historically makes web development so much fun.

Instead of trying to boil the whole ocean by providing every build variant, or having to compromise on size to accomodate legacy, the idea here is to get developers started quickly while providing an easier path to "real" LitElement usage via local builds.

Part of this approach might need to add guards to the build to give up when loaded in the presence of existing LitElement (to prevent over-inclusion) and console logging to flag that this isn't the most efficient approach.

jimmont:
Is it possible to publish the ES modules as javascript with the bare module-name paths resolved to local paths and published at https://lit.github.io/lit/ (or root there)? And expose that branch for cloning to allow using locally? 

Minification and other transformations aren't needed in my case as the projects I'd be using this with would have their own bundling/processing steps.

This would allow using Lit without requiring additional tooling (npm, rollup, etc, etc) and side-step those risks and overhead.

The user story would be:
As a developer I want static importable javascript files so that I can use Lit directly without additional package/transform tooling, except for a web browser and git.

jimmont:
> As a developer I want static importable javascript files so that I can use Lit directly without additional package/transform tooling, except for a web browser and git.

following up to my previous comment I've been working through a solution to provide own static modules, including Lit, without using npm directly; transforming modules to use as static resources and provide for installation using git in mixed projects where the module provided is expected to work with npm; this is specific to how I consume Lit related dependencies, this issue and the user story mentioned here;

the script to import, as part of an evolving idea to solve for the given user-story, essentially copies the modules from unpkg and saves locally:
https://github.com/sourdough/starter/blob/main/tools/external.importer.js

the transformed dependencies dumped into a repo for checkout (including Lit):
https://github.com/sourdough/wildtype

an npm based sample project for testing this scenario and solution:
https://github.com/sourdough/batch

an npm managed dependency which installs the above wildtype modules:
https://github.com/sourdough/ingredient

I was expecting consuming projects to have their own process which would optimize for their development and release scenarios.
[...]

Issue #1257, opened 2020-09-12
single bundled, downloadable distribution please


khoan:
would be handy for developing offline, e.g. long train rides with choppy interweb.

Thanks.

justinfagnani:
Once you install with npm you can work offline, just like if you downloaded a bundle you could work offline. What are the different constraints for you?

FYI, we're unlikely to pursue strategies here that only work for leaf modules and not for modules that depend on lit-html, especially lit-element. If we claim you can use lit-html w/o a package manager or decent dev server, but you can't when using lit-element, then we've exaggerated on a practical level that I'm not comfortable with. Since lit-element imports lit-html, we need a solution that allows for resolving bare import specifiers.

Right now `npm i` and a dev server like `es-dev-server` do that quite well and work offline, so we need more information about why that wouldn't work.

jimmont:
hey @justinfagnani, the general assertion and problem here seems to be the added requirement of non-static set of assets, specifically a transformative process or service, or anything that doesn't just work in a browser given whatever environment it's pulled into. Bare imports, package management and whatever middleware were/are generally assumed to be distinct from consuming most frontend javascript. In my case I'd prefer to serve up from Deno and get rid of Node entirely. Preferences aren't so much the point as what isolates concerns and eases adoption. If that's not an aim of lit-things then this probably isn't much use to comment about, and is a common issue in frontend projects--implementation requires some toolkit. When I've migrated to various platforms (rpi arm64, arm32, apple-silicon, various major versions of Nodejs) these create problems and in my opinion unnecessary complexity that I've been blocked by several times. Distributing a bundle of directly importable javascript files is ideal.

justinfagnani:
Our plan is to publish a few pre-configured bundles to lit.dev

kevinpschaaf:
> Our plan is to publish a few pre-configured bundles to lit.dev

Specifically, I think we'll plan on adding two bundles:
* A `lit.min.js` bundle that bundles and exports only the code currently loaded by the `lit` entrypoint (i.e., excludes decorators, directives, polyfill-support, etc.)
* A `lit.max.js` bundle that exports the kitchen sink -- including decorators, directives, and polyfill-support

thescientist13:
So it seems this would this effectively address the request made in https://github.com/lit/lit-element/issues/603?

PR #2361, opened 2021-12-15
Add 'all' bundle and ESM bundles.

bicknellr:
This adds a new ['max' bundle](https://github.com/lit/lit/issues/1257#issuecomment-924389119) (name needs bikeshedding) that reexports everything and also starts generating ESM versions of both the 'min' and 'max' bundles.

This is part of the work towards closing #2272.


packages/lit/src/index.max.ts
```diff
@@ -0,0 +1,41 @@
...
+export * from './directives/unsafe-svg.js';
+export * from './directives/until.js';
+export * from './directives/when.js';
+export {
+  html as staticHtml,
+  literal,
+  svg as staticSVG,
+  unsafeStatic,
+  withStatic,
+} from './static-html.js';
```
bicknellr, line 41: Rollup was nice enough to complain that re-exporting everything from this file as-is would cause a collision. I don't think there's a way to say "re-export everything, but rename just these specific things", so I had to list them all out here. Fortunately there's not many.

packages/lit/src/index.max.ts
```diff
@@ -0,0 +1,41 @@
...
+export * from './directives/map.js';
+export * from './directives/range.js';
+export * from './directives/ref.js';
+export * from './directives/repeat.js';
+export * from './directives/style-map.js';
+export * from './directives/template-content.js';
+export * from './directives/unsafe-html.js';
+export * from './directives/unsafe-svg.js';
+export * from './directives/until.js';
+export * from './directives/when.js';
```
bicknellr, line 34: Maybe we should add a `directives.js` that groups all of these, like `decorators.js`?

justinfagnani: I don't think we should.

The rationale for the current setup is that there are more directives and they are less likely to be used in a known combination than the decorators. There were even fewer decorators at first and the core were pretty likely to be used together: `@customElement`, `@property`, and `@state` at least. Decorators also have to be used with a build chain, so they're more likely to be used with tree-shaking optimizers.

I'd rather move towards separating the decorators, but we can't do that now.

packages/lit/src/index.max.ts
justinfagnani: Leave a comment here and in lit-html/static.js that this export needs to be updated if new exports are added to that file.

packages/lit/src/index.max.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import './polyfill-support.js';
```
justinfagnani, line 7: I think this file should have a console.warn() about not using it in production and it being much larger than using the files from npm.

packages/lit/rollup.config.js
```diff
@@ -58,5 +58,20 @@ export default litProdConfig({
       output: 'lit.min',
       name: 'Lit',
     },
+    {
+      file: 'index',
+      output: 'lit.min.esm',
```
justinfagnani, line 63: Let's not use the `esm` suffix here.

packages/lit/rollup.config.js
```diff
@@ -58,5 +58,20 @@ export default litProdConfig({
...
+      format: 'es',
+    },
+    {
+      file: 'index.max',
+      output: 'lit.max',
```
justinfagnani, line 68: what format is this? Let's add a suffix for this one.

justinfagnani (changes requested):

packages/lit/rollup.config.js
bicknellr: The `output` field of each of these will get `.js` appended by `litMonoBundleConfig`, here: https://github.com/lit/lit/blob/81cc2aa0e825fafa49b8f1cce7a784123714835e/rollup-common.js#L434

Before these changes, we were only generating `lit.min.js`. (I don't think this file was ever part of the npm package, so we can probably rename it safely.) After, the bundles are now `lit.min.js`, `lit.min.esm.js`, `lit.max.js`, and `lit.max.esm.js`. You mentioned not wanting to use `.esm` above, is that because you were thinking it was the final extension (i.e. no `.js`) or would you prefer the ESM bundles to be the no-indication default and maybe have a `.umd` infix for the UMD ones intead?

packages/lit/rollup.config.js
bicknellr: (see other comment)

packages/lit/src/index.max.ts
bicknellr: Ok, I think I'll also add a flag to disable it though.

bicknellr:
**Bundle names bikeshedding thread**

I used `.max` here since Kevin used it as a straw-man once for the full-featured bundles being added here, but `.max` is confusing when the options are {UMD, ESM} &cross; {fewest features, all features}, where all of those are minified, which is what `.min` usually represents. (i.e. `.max` wouldn't be the opposite end of the axis `.min` usually refers to.)

So the situation in this PR is currently:

| | UMD | ESM |
| - | - | - |
| fewest features | `lit.min.js` | `lit.min.esm.js` |
| all features | `lit.max.js` | `lit.max.esm.js` |

One possible option, using `.full` instead of `.max` and where ESM gets the shorter name:

| | UMD | ESM |
| - | - | - |
| fewest features | `lit.min.umd.js` | `lit.min.js` |
| all features | `lit.min.full.umd.js` | `lit.min.full.js` |

I know it's just a file name, but four extensions feels a bit much... Yet I also still want to replace `.min` with `.bundle` or something. *shrug*

@justinfagnani @kevinpschaaf Thoughts?

kevinpschaaf:
For the one axis, I like `umd` for the umd build and no moniker for the `esm` build. 2nd choice would be monikers for both (e.g. `lit.umd.js` and `lit.esm.js`, which might just be clearer).

If we're not going to also publish a non-minified version for each (and just include sourcemaps) then it seems like we could also drop the `min` part. So would:

`lit.js`
`lit.umd.js`
+
`lit.full.js`
`lit.full.umd.js`

work?

justinfagnani:
I think I would consider "all" for the max/full build, no bundle-type modifier for standard JS modules:

| | UMD | ESM |
| - | - | - |
| fewest features | `lit.min.umd.js` | `lit.min.js` |
| all features | `lit.all.min.umd.js` | `lit.all.min.js` |

packages/lit/rollup.config.js
justinfagnani: yes, I meant no suffix before the extension for standard modules and a `umd` suffix for this bundle. Taking it to the main thread though.

packages/lit/src/index.max.ts
bicknellr: Added.

justinfagnani (approved):
Awesome!

packages/lit/rollup.config.js
```diff
@@ -53,10 +53,27 @@ export default litProdConfig({
     {
       file: 'polyfill-support',
     },
```
bicknellr, line 55: I noticed this earlier today and thought it might have implications for the naming scheme used below since the polyfill support module is in a UMD wrapper. However, I think this is fine because it actually works in both ESM and UMD contexts as it has no exports (it only writes globals).

bicknellr:
Everything's passed after some reruns (miraculously including the IE11 tests?) and I've double checked that this doesn't add anything to `lit` using `npm pack`, so I'll go ahead and merge.

AndrewJakubowicz:
Wow the ie11 tests passed! Did you need to retry any of the github actions? This is great feedback/followup on https://github.com/lit/lit/pull/2364!

Thank you!

bicknellr:
I had to retry some of them a few times but it was much nicer being able to restart more granularly.

[The timeout](https://github.com/lit/lit/runs/4663254143?check_suite_focus=true#step:9:185) looks like it happened during part of the first `npm run test` here:

https://github.com/lit/lit/blob/cc700d9e3325a0f71acf60b2f753c64427571d7f/.github/workflows/sauce-ie11.yaml#L47

And the success is because the second attempt passed? Also, it looks like both the [dev build](https://github.com/lit/lit/runs/4663254143?check_suite_focus=true#step:9:321) and [prod build](https://github.com/lit/lit/runs/4663254143?check_suite_focus=true#step:9:349) ran during that second attempt. Should it have done something different?

AndrewJakubowicz:
Exceptional it worked as intended! Forgot I added that!

5310:
Pardon me, but where can I find this `lit.all.min.esm.js` or `lit.all.min.js`, which ever is the new ESM bundle? 

[Unpkg](https://unpkg.com/browse/lit@2.1.0/) doesn't seem to have any such file even after 2.1.0 was pushed today, which no longer works with Skypack.

bicknellr:
These aren't being distributed anywhere just yet. If you want to try them out in the meantime, building this repo (clone, then `npm ci && npm run bootstrap && npm run build`) will place them in `packages/lit/`.

PR #2364, opened 2021-12-16
[test] Improve robustness of our integration testing

AndrewJakubowicz:
### Context

We have not had a green build in a long time. This is some work to improve the tests. There are three sources of failure that need addressing.

1. Tests that always fail on ie11. These were opted out.
2. Sauce test flakes through timeouts - increasing the start timeout mostly resolves this.
3. Sauce proxy connection failure (another case of the test flakes).

In this PR I've addressed 1, and tried to improve 2 by playing with the configurations. 3 is still an issue. I've added a workaround so it's possible to retry a single sauce test instead of all 4.

### Tested

Through Github actions and manual sauce tests.

In order to make it possible to feasibly get to `green`, I've split the Github Action jobs so they can be rerun in isolation. This means that with the 50% failure rate we don't have to restart all 4 sauce tests.

Does this need a test package changelog?


.github/workflows/tests.yml
```diff
@@ -46,12 +46,12 @@ jobs:
...
         with:
-          node-version: 14
+          node-version: 16
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
 
-      # Installs system dependencies needed for playwright.
-      - uses: microsoft/playwright-github-action@v1
+      - name: Install playwright deps
+        run: npx playwright install-deps
```
justinfagnani, line 54: Do you have a theory on why this helped? Leave a comment?

justinfagnani (approved):
Awesome!

.github/workflows/tests.yml
AndrewJakubowicz: I'm not convinced this helped. Instead I did this while I was updating the libraries since it was recommended: https://github.com/microsoft/playwright-github-action#%EF%B8%8F-you-dont-need-this-github-action-%EF%B8%8F

It's also symmetric with [lit.dev](https://github.com/lit/lit.dev/blob/main/.github/workflows/integration-tests.yml#L17).

PR #2370, opened 2021-12-21
Replace jsdoc symbol links with @linkcode instead of square brackets

AndrewJakubowicz:
### Context

Currently we support [[`symbol`]] in our documentation to create an automated hyperlink in our generated lit.dev API documentation. However [[`symbol`]] isn't recognized by VS code and ends up being rendered directly into the documentation.

Instead if we use `{@link symbol}` or `{@linkcode symbol}` then we get automatic hyperlinks and jump to definition in the IDE hover documentation if the linked symbol is in scope. If the symbol is not in scope the IDE experience is similar to currently where the link text is directly printed into the documentation.

This change should follow lit.dev's change https://github.com/lit/lit.dev/pull/629 which adds logic to handle transforming `@link` and `@linkcode`.

### Testing

This was tested manually by hovering over the symbols, and using this PR to generate Lit.dev documentation.

### Open questions

 - Some of the links reference symbols that aren't imported or in scope. Is it worth importing the symbols as type only for the purpose of documentation?

#### In editor, hover documentation on main branch:

Screen Shot 2021-12-21 at 2 52 24 PM

#### In editor, hover documentation with this change:

Screen Shot 2021-12-21 at 2 40 56 PM

#### Generated documentation on lit.dev (for both):

Screen Shot 2021-12-21 at 2 49 40 PM

### DIFF
--- a/.changeset/wise-taxis-refuse.md
+++ b/.changeset/wise-taxis-refuse.md
@@ -1,0 +1,7 @@
+---
+'lit-element': patch
+'@lit/reactive-element': patch
+---
+
+Replace square bracket links with the `linkcode` JSDoc tag.
+Editors will create a jump to definition hyperlink for the linkcode tag if the identifier is in scope.
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -5,15 +5,15 @@
  */
 
 /**
- * The main LitElement module, which defines the [[`LitElement`]] base class and
- * related APIs.
+ * The main LitElement module, which defines the {@linkcode LitElement} base
+ * class and related APIs.
  *
  *  LitElement components can define a template and a set of observed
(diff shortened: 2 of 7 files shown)


aomarks:
> * Some of the links reference symbols that aren't imported or in scope. Is it worth importing the symbols as type only for the purpose of documentation?

Hmmm... maybe? What are some examples?

aomarks:
Just needs a changeset

aomarks (approved):
Nice!

AndrewJakubowicz:
> Hmmm... maybe? What are some examples?

In the screenshot for LitElement, note that `property` isn't linked in the VS Code screenshot. One way this can be "fixed" is by adding a type only import: `import type {property} from '@lit/reactive-element/decorators.js';` so the identifier is in scope.

justinfagnani (approved):

PR #2376, opened 2022-01-05
Set observers version to 0.0.0 so that changesets releases it as 1.0.0

aomarks:


bicknellr (approved):
This hasn't been pushed to npm before so LGTM.

Issue #2383, opened 2021-12-31
[lit] Lit Skypack support


codecycletechnologies:
Hello there,

We tried to use lit element from https://www.skypack.dev/view/lit.
We are receiving below error if we try with the given sample code in documentation.

```
Property 'render' in type 'MyElement' is not assignable to the same property in base type 's'.
  Type '() => { _$litType$: any; strings: any; values: any[]; }' is not assignable to type '() => symbol'.
    Type '{ _$litType$: any; strings: any; values: any[]; }' is not assignable to type 'symbol'.deno-ts(2416)
(method) MyElement.render(): {
    _$litType$: any;
    strings: any;
    values: any[];
}

```

However if we add `as unknown as symbol` at the end of render method then this type error is getting resolved.

```
 render () {
        return html`Web Components are <span>${ this.mood }</span>!` as unknown as symbol;
 }
```

It would be great if skypack import work in similar way as npm lit module render method. 
And if render method needs some changes in skypack then it should be available in documentation.

aomarks:
For reference, it sounds like the snippet being run is the one from https://github.com/lit/lit/tree/main/packages/lit#overview

```ts
import {LitElement, html, css} from 'lit';
import {customElement, property} from 'lit/decorators.js';

// Registers the element
@customElement('my-element')
export class MyElement extends LitElement {
  // Styles are applied to the shadow root and scoped to this element
  static styles = css`
    span {
      color: green;
    }
  `;

  // Creates a reactive property that triggers rendering
  @property()
  mood = 'great';

  // Render the component's DOM by returning a Lit template
  render() {
    return html`Web Components are <span>${this.mood}</span>!`;
  }
}
```

aomarks:
@codecycletechnologies Can you give a little more information about how you are evaluating this code? Are you using deno, or some other build tools?

taylor-vann:
Hey @codecycletechnologies thanks for reaching out!

Wanted to ask a few clarifying questions :) Sorry if they're basic

1) How are we importing LIt? And are we loading the `d.ts` file as well?

Something like 

```
// @deno-types="https://cdn.skypack.dev/lit?dts"
import * as coolLib from "https://cdn.skypack.dev/lit";
```

2) Mind if we see a code snippet or a repo with this issue in action?

albertosantini:
I cannot reproduce it, but I think there is a misunderstanding,

For example, https://jsbin.com/vizayohedu/edit?html,output

```
[...]

PR #2384, opened 2022-01-06
Fix missing decorators/query-assigned-elements.js file

aomarks:
We forgot to update the `rollup.config.js` files to include the new `query-assigned-elements.js` decorator file in the NPM packages.

May help with https://github.com/lit/lit/issues/2383, but I think this is a new problem on top of an existing Deno-related problem with Skypack.


AndrewJakubowicz (approved):
Thank you!

PR #2386, opened 2022-01-06
Add missing caret to SSR's reactive-element dependency range

aomarks:
Not sure why this was pinned to 1.1.0 -- it should use a ^ range like everything else.


AndrewJakubowicz (approved):
LGTM!

PR #2392, opened 2022-01-07
Improve queryAssignedElements decorator browser compatibility

AndrewJakubowicz:
### Context

Internally queryAssignedElements was using the newer `HTMLSlotElement.assignedElements` method. This newer method doesn't have ShadyDOM polyfill support, but would still leave some browsers unsupported if it did have polyfill support.

Using [Modern browser breakdown](https://lit.dev/docs/tools/requirements/#modern-browser-breakdown) as a guide, `assignedElements` is not supported until Firefox 66. This leaves Firefox versions 64 and 65 without support.

tl;dr `queryAssignedElements` doesn't work on older browsers.

### How

Gracefully fallback on `assignedNodes` if required.  The behavior of the decorator remains unchanged.

### Testing

This is a private implementation change, so the existing unit tests should cover this. No functional changes.
Also testing via cl/420366110

Size of decorators before changes (on main):

```
Name                                   Size     Minified  Gzipped  Brotli   
----------------------------------------------------------------------------
decorators/query-assigned-elements.js  74 B     73 B      87 B     62 B     
decorators/query-assigned-nodes.js     71 B     70 B      86 B     63 B 

decorators/query-assigned-elements.js (DEV)  519 B     518 B     380 B    323 B
decorators/query-assigned-nodes.js (DEV)     593 B     592 B     442 B    362 B 
```

Size of decorators after change:

```
Name                                   Size      Minified  Gzipped  Brotli   
-----------------------------------------------------------------------------
decorators/query-assigned-elements.js  74 B    73 B      87 B     62 B     
decorators/query-assigned-nodes.js     71 B    70 B      86 B     63 B

decorators/query-assigned-elements.js (DEV)  685 B     684 B     464 B    391 B    
decorators/query-assigned-nodes.js (DEV)     606 B     605 B     421 B    361 B 
```

The release size is the same after this change, only dev has slightly increased in size.

### DIFF
--- a/.changeset/stupid-rockets-deliver.md
+++ b/.changeset/stupid-rockets-deliver.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Fix `@queryAssignedElements` decorator so internal implementation is compatible with Lit's browser support matrix.
--- a/.changeset/stupid-rockets-deliver.md
+++ b/.changeset/stupid-rockets-deliver.md
@@ -2,4 +2,6 @@
 '@lit/reactive-element': patch
 ---
 
-Fix `@queryAssignedElements` decorator so internal implementation is compatible with Lit's browser support matrix.
+Fix `@queryAssignedElements` decorator so it is compatible with legacy browsers.
+Uses `HTMLSlotElement.assignedElements` if available with a graceful fallback
+on `HTMLSlotElement.assignedNodes` which is supported by polyfills.
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -23,10 +23,10 @@ const slotAssignedElements =
   window.HTMLSlotElement?.prototype.assignedElements ??
   function slotAssignedElementsPolyfill(
     this: HTMLSlotElement,
-    opts: AssignedNodesOptions
+    opts?: AssignedNodesOptions
   ) {
     return this.assignedNodes(opts).filter(
-      (node) => node.nodeType === Node.ELEMENT_NODE
+      (node): node is Element => node.nodeType === Node.ELEMENT_NODE
     );
   };
 
@@ -82,7 +82,7 @@ export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
         const slotEl =
           this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
         const elements =
-          slotEl !== null ? slotAssignedElements.call(slotEl, options) : [];
+          slotEl != null ? slotAssignedElements.call(slotEl, options) : [];
         if (selector) {
           return elements.filter((node) => node.matches(selector));
         }
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -67,11 +67,18 @@ export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
         const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
         const slotEl =
           this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
-        const elements = slotEl?.assignedElements(options) ?? [];
+        // HTMLSlotElement.assignedElements would be preferred here, however
+        // there isn't a good path forward regarding browser polyfill support.
+        const elements = slotEl?.assignedNodes(options) ?? [];
         if (selector) {
-          return elements.filter((node) => node.matches(selector));
+          return elements.filter(
+            (node) =>
+              node.nodeType === Node.ELEMENT_NODE &&
+              (node as Element).matches(selector)
+          );
+        } else {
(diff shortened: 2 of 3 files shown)


packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -67,11 +67,18 @@ export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
         const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
         const slotEl =
           this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
-        const elements = slotEl?.assignedElements(options) ?? [];
+        // HTMLSlotElement.assignedElements would be preferred here, however
+        // there isn't a good path forward regarding browser polyfill support.
+        const elements = slotEl?.assignedNodes(options) ?? [];
```
aomarks, line 72: What do you think about effectively including a tiny module-scoped polyfill directly in this file, so that we'll still use built-in `assignedElements` when available?

Maybe something like this?


```ts
const slotAssignedElements =
  HTMLSlotElement.prototype.assignedElements ??
  function slotAssignedElementsPolyfill(
    this: HTMLSlotElement,
    opts: AssignedNodesOptions
  ) {
    return this.assignedNodes(opts).filter(
      (node) => node.nodeType === Node.ELEMENT_NODE
    );
  };

...

const elements =
  slotEl !== null ? slotAssignedElements.call(slotEl, options) : [];
```

It has the advantage of using the native method when available instead of the filter. Presumably this is faster... though I don't know for sure.

It does mean that when a selector is provided, and the polyfill is used, there will be 2 filters instead of 1. But probably preferring slightly better performance on newer browsers is the direction we should lean in?

AndrewJakubowicz: Would the `HTMLSlotElement` exist on IE11, for example when shadyDOM is in use?

aomarks: Good question, I think not reading https://github.com/webcomponents/polyfills/blob/0d119eadcf382f4e5ed865c5b96e87c9350e2d1e/packages/shadydom/src/patch-prototypes.js#L72, but I'm not sure...

@bicknellr @kevinpschaaf can you confirm?

If not I suppose you can do `window?.HTMLSlotElement?.prototype?.assignedElements` instead?

AndrewJakubowicz: Done! Great suggestion to measure bundle size!

bicknellr: I'm not sure, I assumed Shady DOM would need to create `HTMLSlotElement` if it doesn't exist but I can't find where it does that. cc @sorvell

AndrewJakubowicz: I can also add that depending on HTMLSlotElement directly fails some tests. Using `window.HTMLSlotElement?.prototype.assignedElements` works.

aomarks (approved):
Looks good to me! Would be nice to get a stamp from @kevinpschaaf too on this approach.

.changeset/stupid-rockets-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Fix `@queryAssignedElements` decorator so internal implementation is compatible with Lit's browser support matrix.
```
aomarks, line 5: Could rephrase this to be a little more useful to users -- specifically it fixes the decorator for a certain range of Firefox versions, right?

AndrewJakubowicz: Sorry the Firefox version stuff was a complexity when deciding where to put the `assignedElements` "polyfill". Either in Lit directly, or in ShadyDOM.

The Firefox version in the support matrix means for full legacy browser & modern browser support, we should put the polyfill in the decorator.

I'll update the changelog to be more straightforward. LMK if it's better. Thank you!!!

AndrewJakubowicz:
Note, this has passed tgp.

kevinpschaaf (approved):
LGTM

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -67,7 +81,8 @@ export function queryAssignedElements(options?: QueryAssignedElementsOptions) {
         const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
         const slotEl =
           this.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
-        const elements = slotEl?.assignedElements(options) ?? [];
+        const elements =
+          slotEl != null ? slotAssignedElements.call(slotEl, options) : [];
```
sorvell, line 85: I suggest making this take an element so you don't have to use call. It's simpler and should be just slightly faster.

sorvell (changes requested):

packages/reactive-element/src/decorators/query-assigned-elements.ts
AndrewJakubowicz: Done! Thank you!

sorvell (approved):

PR #2401, opened 2022-01-13
Add a devlog events system

rictic:
When running Lit in dev mode, and when window.emitLitDebugLogEvents is true, we will emit 'lit-debug' events on window. These events will be fired during rendering and as part of the ReactiveElement update lifecycle, giving internal information about rendering that may be useful for debug tools and visualizers.

Production builds will not be affected, all of the devlog code is compiled out when DEV_MODE is false.

Note that this is an unstable API which may change or even be removed in the future, even on patch releases.

### DIFF
--- a/.changeset/chilly-spies-brush.md
+++ b/.changeset/chilly-spies-brush.md
@@ -1,6 +1,0 @@
----
-'lit-html': patch
-'@lit/reactive-element': patch
----
-
-Added a devlog events system that may be used for debugging and visualizing Lit's internals.
--- a/.changeset/chilly-spies-brush.md
+++ b/.changeset/chilly-spies-brush.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'@lit/reactive-element': patch
+---
+
+Added a devlog events system that may be used for debugging and visualizing Lit's internals.
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -201,7 +201,10 @@ const debugLogEvent = DEV_MODE
       }
     }
   : undefined;
-let debugLogId = 0;
+// Used for connecting beginRender and endRender events when there are nested
+// renders when errors are thrown preventing an endRender event from being
+// called.
+let debugLogRenderId = 0;
 
 /**
  * `true` if we're building for google3 with temporary back-compat helpers.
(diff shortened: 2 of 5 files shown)


.changeset/chilly-spies-brush.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit-html': patch
```
aomarks, line 2: Seems like this should be a minor bump, since it's adding a new API (albeit only in dev mode)?

rictic: My thinking is that this is establishing a pattern that this is an unstable API that's free to change at any time, including in patch releases. On the other hand, that doesn't mean that we _have_ to only do so in patch releases. I think I find that convincing, will update.

packages/lit-html/src/lit-html.ts
```diff
@@ -11,6 +11,198 @@ const DEV_MODE = true;
...
+      if (shouldEmit) {
+        window.dispatchEvent(
+          new CustomEvent<LitUnstable.DebugLog.Entry>('lit-debug', {
+            detail: event,
+          })
+        );
+      }
+    }
+  : undefined;
+let debugLogId = 0;
```
aomarks, line 204: Should this be called `debugLogRenderId` for clarity (at first I thought it would count all log events, but it's actually just renders).

aomarks, line 204: Will this variable be removed by Terser and Closure when not in DEV_MODE?

packages/lit-html/src/lit-html.ts
```diff
@@ -422,10 +614,19 @@ export const render = (
   container: HTMLElement | DocumentFragment,
   options?: RenderOptions
 ): RootPart => {
+  const renderId = DEV_MODE ? debugLogId++ : 0;
   const partOwnerNode = options?.renderBefore ?? container;
   // This property needs to remain unminified.
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let part: ChildPart = (partOwnerNode as any)['_$litPart$'];
+  debugLogEvent?.({
```
aomarks, line 622: Are Terser and Closure smart enough to remove these statements entirely when debugLogEvent is undefined?

aomarks (changes requested):
Very cool!

packages/lit-html/src/lit-html.ts
rictic: +1

rictic: Yep!

packages/lit-html/src/lit-html.ts
rictic: Yep!

aomarks (approved):

PR #2404, opened 2022-01-14
Push bundles to lit/bundles during release.

bicknellr:
This PR adds a couple new steps to the release workflow which will push bundles to the [lit/bundles](https://github.com/lit/bundles) repo. New bundles are pushed whenever `lit` appears in [the `publishedPackages` output of `changesets/action`](https://github.com/changesets/action#outputs). lit/bundles currently contains a single empty commit with tag `empty`; the new steps check out the repo at that commit, copy in the bundles and their source maps, and then create and push a new commit tagged with the version of `lit` that was published.

[lit/bundles](https://github.com/lit/bundles) is currently private in case we need to trash the entire repo and try again. Also, this PR should finish the work for #2272, but I don't want to give GitHub the actual auto-close-on-merge phrase since we should watch it once or twice after merging before we make the repo public.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -10,6 +10,10 @@ lerna-debug.log
 .DS_Store
 .vscode/
 
+# `/bundles` is used during the release step as the checkout location for the
+# bundles repo. (https://github.com/lit/bundles)
+/bundles
+
 packages/benchmarks/generated/
 packages/benchmarks/generator/build/
 packages/benchmarks/generator/generated/
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -76,14 +76,7 @@ jobs:
           cp ../packages/lit/lit.min.umd.js.map .
 
           # Stage the bundles, create the commit, tag it, and push.
-          git add lit.all.min.js
-          git add lit.all.min.js.map
-          git add lit.all.min.umd.js
-          git add lit.all.min.umd.js.map
-          git add lit.min.js
-          git add lit.min.js.map
-          git add lit.min.umd.js
-          git add lit.min.umd.js.map
+          git add .
           git config user.name "Lit Release Bot"
           git config user.email "lit-release-bot@google.com"
           git commit -m "Bundles for lit@${LIT_VERSION}"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,3 +37,60 @@ jobs:
           HUSKY: 0
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Checkout bundles repo
+        uses: actions/checkout@v2
+        if: steps.changesets.outputs.published
+        with:
+          repository: lit/bundles
+          ref: empty
+          path: bundles
+          token: ${{ secrets.LIT_RELEASE_BOT_TOKEN }}
+
+      - name: Push bundles
+        if: steps.changesets.outputs.published
+        run: |
+          # Extract the version of `lit` that was published or the empty string.
+          LIT_VERSION=$(npm run --silent extract-published-lit-version <<EOF
+            ${{ steps.changesets.outputs.publishedPackages }}
+          EOF
+          )
+
+          # Don't create a bundle commit if `lit` wasn't published.
+          if [[ -z "$LIT_VERSION" ]]; then
+            exit
+          fi
+
+          # Enter the bundle repo directory.
+          pushd bundles
+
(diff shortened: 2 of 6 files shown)


.eslintignore
```diff
@@ -10,6 +10,10 @@ lerna-debug.log
 .DS_Store
 .vscode/
 
+# `/bundles` is used during the release step as the checkout location for the
```
aomarks, line 13: This file gets auto-generated by the `ignore-sync` script, so you only need to put it in the `.gitignore` file and run `npm run ignore-sync`. (See https://github.com/foray1010/ignore-sync#readme)

.github/workflows/release.yaml
```diff
@@ -37,3 +37,60 @@ jobs:
...
+          ref: empty
+          path: bundles
+          token: ${{ secrets.LIT_RELEASE_BOT_TOKEN }}
+
+      - name: Push bundles
+        if: steps.changesets.outputs.published
+        run: |
+          # Extract the version of `lit` that was published or the empty string.
+          LIT_VERSION=$(npm run --silent extract-published-lit-version <<EOF
+            ${{ steps.changesets.outputs.publishedPackages }}
```
aomarks, line 55: So this will only work once we update our changeset logic to bump the lit package every time any of the core packages are released, right?

.github/workflows/release.yaml
```diff
@@ -37,3 +37,60 @@ jobs:
...
+          git add lit.min.umd.js
+          git add lit.min.umd.js.map
+          git config user.name "Lit Release Bot"
+          git config user.email "lit-release-bot@google.com"
+          git commit -m "Bundles for lit@${LIT_VERSION}"
+          git tag "v${LIT_VERSION}"
+          git push origin "v${LIT_VERSION}"
+
+          # Exit the bundle repo directory.
+          popd
```
aomarks, line 96: Is pushd/popd necessary? Does each step not reset its cwd?

.github/workflows/release.yaml
```diff
@@ -37,3 +37,60 @@ jobs:
...
+
+          # Don't create a bundle commit if `lit` wasn't published.
+          if [[ -z "$LIT_VERSION" ]]; then
+            exit
+          fi
+
+          # Enter the bundle repo directory.
+          pushd bundles
+
+          # Checkout the empty root commit (with tag `empty`).
```
aomarks, line 67: Is `empty` a tag that you previously created?

.github/workflows/release.yaml
```diff
@@ -37,3 +37,60 @@ jobs:
...
+          cp ../packages/lit/lit.all.min.js.map .
+          cp ../packages/lit/lit.all.min.umd.js .
+          cp ../packages/lit/lit.all.min.umd.js.map .
+          cp ../packages/lit/lit.min.js .
+          cp ../packages/lit/lit.min.js.map .
+          cp ../packages/lit/lit.min.umd.js .
+          cp ../packages/lit/lit.min.umd.js.map .
+
+          # Stage the bundles, create the commit, tag it, and push.
+          git add lit.all.min.js
```
aomarks, line 81: Would `git add .` work so that we don't need to enumerate the list of files twice? There wouldn't be any additional files, since this is always starting from an empty directory, right?

aomarks (changes requested):
Nice! Just a few comments.

.github/workflows/release.yaml
bicknellr: I'm not sure if the working directory is preserved across steps. I haven't been able to find anywhere that the docs call this out explicitly but, given that [`working-directory`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsrun) exists, it seems more likely you're right that they reset. Either way, since the parts outside of the pushd/popd here don't depend on the working directory, I've removed them and added `working-directory: bundles` to this step.

.github/workflows/release.yaml
bicknellr: Yes, the intent here is to have the bundles and their version tags match the `lit` package releases on npm: iff the release workflow publishes `lit` to npm, then we should push the bundles generated during that same workflow.

.eslintignore
bicknellr: Yeah, these are the changes that script made. These files aren't ignored, so I assume we want them checked in?

aomarks: Oh I see, sorry I think the comments made me think they were manually checked copied to each place. NM.

aomarks (approved):

.eslintignore
bicknellr: OIC, yeah I was surprised it copied the comments also.

Issue #1897, opened 2021-05-14
[localize] Generate IDs for XLB placeholders


aomarks:
In XLIFF, we generate an ID for each placeholder from its position in the source template. This lets us reliably track placeholders if they change position in a translated template.

For XLB, we don't currently generate an ID at all, so we use the placeholder text itself as the ID. This is fine most of the time, but could fail in some weird edge cases where two identical placeholders appear, but their order actually matters (e.g. a template with an expression `${i++}`).

Maybe we can use the XLB `name` attribute for this, but we need to think about what this would mean for translators, because this might cause them to see a numeric ID instead of the placeholder contents.

PR #2405, opened 2022-01-14
[localize-tools] Improve handling of duplicate msg id

### LINKED ISSUE #2140: [localize] Improve handling of translation messages with same id
## Description

- Who wants the functionality
Your friends at Reddit 😄 

- What it is they want
The following code produces the below error when running `lit-localize extract`
```js
msg(str`${mother} said hello`)
msg(str`${father} said hello`)
....
error TS2324: Message ids must have the same default text wherever they are used
```

- Why they want it
To improve developer ergonomics of using Lit/Localize.

## Acceptance criteria

**At a minimum:** the error message is more clear and actionable. 
**Ideally:** Allow the same message in multiple places, as long as the strings + description are the same.

augustjk:
Fixes #2140 and #1897.

- For messages to be considered same:
  - No longer check for expressions to be identical.
  - Do check that desc be identical.
- Display a more helpful aggregate error message for all messages with the same id.
- For XLB format, add placeholder index to `name` attribute of `<ph>` tags.
- Utilize the placeholder indexes embedded in tags for determining placeholder locations during transform mode.

### DIFF
--- a/.changeset/many-singers-complain.md
+++ b/.changeset/many-singers-complain.md
@@ -4,4 +4,4 @@
 
 **BREAKING** Update analysis to consider messages with same id **and** description to be identical (but no longer checks for expressions to be same) and improve error message on finding incompatible duplicates.
 
-`lit-locazlie extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable to message or add different `id` options to differentiate them.
+`lit-localize extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable message or add different `id` options to differentiate them.
--- a/.changeset/many-singers-complain.md
+++ b/.changeset/many-singers-complain.md
@@ -1,0 +1,7 @@
+---
+'@lit/localize-tools': minor
+---
+
+**BREAKING** Update analysis to consider messages with same id **and** description to be identical (but no longer checks for expressions to be same) and improve error message on finding incompatible duplicates.
+
+`lit-locazlie extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable to message or add different `id` options to differentiate them.
--- a/packages/localize-tools/src/formatters/xlb.ts
+++ b/packages/localize-tools/src/formatters/xlb.ts
@@ -76,7 +76,6 @@ class XlbFormatter implements Formatter {
       const contents: Array<string | Placeholder> = [];
       for (let j = 0; j < msg.childNodes.length; j++) {
         const child = msg.childNodes[j];
-        let phIdx = 0;
         if (child.nodeType === doc.TEXT_NODE) {
           contents.push(child.nodeValue || '');
         } else if (
@@ -91,8 +90,9 @@ class XlbFormatter implements Formatter {
           ) {
             throw new KnownError(`Expected <ph> to have exactly one text node`);
           }
-          const index =
-            Number((child as Element).getAttribute('name')) || phIdx++;
+          const index = Number(
+            getNonEmptyAttributeOrThrow(child as Element, 'name')
+          );
           contents.push({untranslatable: phText.nodeValue || '', index});
         } else {
           throw new KnownError(
--- a/packages/localize-tools/src/formatters/xlb.ts
+++ b/packages/localize-tools/src/formatters/xlb.ts
@@ -76,6 +76,7 @@ class XlbFormatter implements Formatter {
       const contents: Array<string | Placeholder> = [];
       for (let j = 0; j < msg.childNodes.length; j++) {
         const child = msg.childNodes[j];
+        let phIdx = 0;
         if (child.nodeType === doc.TEXT_NODE) {
           contents.push(child.nodeValue || '');
         } else if (
@@ -90,7 +91,9 @@ class XlbFormatter implements Formatter {
           ) {
             throw new KnownError(`Expected <ph> to have exactly one text node`);
           }
-          contents.push({untranslatable: phText.nodeValue || ''});
+          const index =
+            Number((child as Element).getAttribute('name')) || phIdx++;
+          contents.push({untranslatable: phText.nodeValue || '', index});
         } else {
           throw new KnownError(
             `Unexpected node in <msg>: ${child.nodeType} ${child.nodeName}`
@@ -127,12 +130,14 @@ class XlbFormatter implements Formatter {
       }
       indent(messagesNode, 2);
       messagesNode.appendChild(messageNode);
+      let phIdx = 0;
       for (const content of contents) {
         if (typeof content === 'string') {
           messageNode.appendChild(doc.createTextNode(content));
         } else {
           const {untranslatable} = content;
           const ph = doc.createElement('ph');
+          ph.setAttribute('name', String(phIdx++));
           ph.appendChild(doc.createTextNode(untranslatable));
           messageNode.appendChild(ph);
         }
--- a/packages/localize-tools/src/messages.ts
+++ b/packages/localize-tools/src/messages.ts
@@ -66,6 +66,7 @@ export interface Bundle {
  */
 export interface Placeholder {
   untranslatable: string;
+  index?: string;
   // TODO(aomarks) Placeholders can also have names and examples, to help the
   // translator understand the meaning of the placeholder. We could
   // automatically add names for common markup patterns like START_BOLD and
--- a/packages/localize-tools/src/messages.ts
+++ b/packages/localize-tools/src/messages.ts
@@ -66,7 +66,7 @@ export interface Bundle {
  */
 export interface Placeholder {
   untranslatable: string;
-  index?: number;
+  index: number;
   // TODO(aomarks) Placeholders can also have names and examples, to help the
   // translator understand the meaning of the placeholder. We could
   // automatically add names for common markup patterns like START_BOLD and
--- a/packages/localize-tools/src/messages.ts
+++ b/packages/localize-tools/src/messages.ts
@@ -66,7 +66,7 @@ export interface Bundle {
  */
 export interface Placeholder {
   untranslatable: string;
-  index?: string;
+  index?: number;
   // TODO(aomarks) Placeholders can also have names and examples, to help the
   // translator understand the meaning of the placeholder. We could
   // automatically add names for common markup patterns like START_BOLD and
--- a/packages/localize-tools/src/modes/transform.ts
+++ b/packages/localize-tools/src/modes/transform.ts
@@ -335,10 +335,12 @@ class Transformer {
     let {template} = templateResult.result;
 
     const placeholdersByIndex = new Map<string, Placeholder>();
-    let idx = 0;
-    for (const content of contents) {
-      if (typeof content === 'object') {
-        placeholdersByIndex.set(String(idx++), content);
+    {
+      let idx = 0;
+      for (const content of contents) {
+        if (typeof content === 'object') {
+          placeholdersByIndex.set(String(idx++), content);
+        }
       }
     }
 
@@ -352,11 +354,15 @@ class Transformer {
     const sourceExpressions = new Map<string, ts.Expression>();
     if (ts.isTemplateExpression(template)) {
       for (const span of template.templateSpans) {
-        // TODO(aomarks) Support less brittle/more readable placeholder keys.
-        const key = this.sourceFile.text
-          .slice(span.expression.pos, span.expression.end)
-          .replace(/{ /g, '{')
-          .replace(/ }/g, '}');
+        const printer = ts.createPrinter({
+          newLine: ts.NewLineKind.LineFeed,
+          noEmitHelpers: true,
+        });
+        const key = printer.printNode(
+          ts.EmitHint.Unspecified,
+          span.expression,
+          this.sourceFile
+        );
         sourceExpressions.set(key, span.expression);
       }
     }
@@ -391,15 +397,21 @@ class Transformer {
           })
           .join('');
 
-        template = parseStringAsTemplateLiteral(templateLiteralBody);
+        const parseResult = parseStringAsTemplateLiteral(templateLiteralBody);
+        template = parseResult.template;
         if (ts.isTemplateExpression(template)) {
           const newParts = [];
           newParts.push(template.head.text);
           for (const span of template.templateSpans) {
-            const expressionKey = templateLiteralBody
-              .slice(span.expression.pos - 1, span.expression.end - 1)
-              .replace(/{ /g, '{')
-              .replace(/ }/g, '}');
+            const printer = ts.createPrinter({
+              newLine: ts.NewLineKind.LineFeed,
+              noEmitHelpers: true,
+            });
+            const expressionKey = printer.printNode(
+              ts.EmitHint.Unspecified,
+              span.expression,
+              parseResult.file
+            );
             const sourceExpression = sourceExpressions.get(expressionKey);
             if (sourceExpression === undefined) {
               throw new Error(
--- a/packages/localize-tools/src/modes/transform.ts
+++ b/packages/localize-tools/src/modes/transform.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {Message, makeMessageIdMap} from '../messages.js';
+import {Message, Placeholder, makeMessageIdMap} from '../messages.js';
 import {writeLocaleCodesModule} from '../locales.js';
 import type {Locale} from '../types/locale.js';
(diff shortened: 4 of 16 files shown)


.changeset/many-singers-complain.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/localize-tools': minor
+---
+
+**BREAKING** Update analysis to consider messages with same id **and** description to be identical (but no longer checks for expressions to be same) and improve error message on finding incompatible duplicates.
+
+`lit-locazlie extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable to message or add different `id` options to differentiate them.
```
aomarks, line 7:
```suggestion
`lit-localize extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable to message or add different `id` options to differentiate them.
```

packages/localize-tools/src/tests/analysis.unit.test.ts
```diff
@@ -342,7 +342,28 @@ test('error: different message contents', () => {
       },
     ],
     [
-      '__DUMMY__.ts(4,5): error TS2324: Message ids must have the same default text wherever they are used',
+      '__DUMMY__.ts(3,5): error TS2324: The translation message with ID greeting was defined in 2 places, but with different strings or descriptions. If these messages should be translated together, make sure their strings and descriptions are identical, and consider factoring out a common variable or function. If they should be translated separately, add one or more {id: "..."} overrides to distinguish them.',
+    ]
+  );
+});
+
+test('error: same message contents with different desc', () => {
```
aomarks, line 350: Mind adding one more test case for `same id and desc with different expression is not error`?

packages/localize-tools/src/program-analysis.ts
```diff
@@ -684,7 +696,6 @@ function contentEqual(
     if (typeof b !== 'object') {
```
aomarks, line 696: Could simplify this now to `if (typeof a === 'object' && typeof b !== 'object')`

aomarks (changes requested):
Great! A few small comments but otherwise LGTM.

packages/localize-tools/src/tests/analysis.unit.test.ts
```diff
@@ -347,6 +347,22 @@ test('error: different message contents', () => {
...
+    import {msg, str} from '@lit/localize';
+    const name = "friend";
+    const otherName = "pal";
+    msg(str\`Hello \${name}\`);
+    msg(str\`Hello \${otherName}\`);
+  `;
+  checkAnalysis(src, [
+    {
+      name: 'saed7d3734ce7f09d',
+      contents: ['Hello ', {untranslatable: '${name}'}],
```
aomarks, line 361: Hmm... Seeing this example has actually made me realize there is a bit of a problem with this change.

Take this example:

```ts
msg(`Hello ${friend}`)
msg(`Hello ${pal}`)
```

Both msg calls will map to the same message id, so there will only be one XLIFF entry for both calls, which will look like this:

```xml
<trans-unit id="saed7d3734ce7f09d">
  <source>Hello <x equiv-text="${friend}"/></source>
</trans-unit>
```

Note that we put the actual source code of the expression inside the `<x>` tag. This is because we want to provide some information to the translator about what that `<x>` placeholder means, so that they can decide where it should go in the translated version (translators can move, but not change `<x>` tags). 

But, we had to pick one of the expressions from the two `msg` calls that correspond to this message id. I expect that currently we'd whichever one we encounter first, which is determined by the order that we traversed the module graph.

This is already a bit of a problem, because it means changing around the module graph could cause surprising churn in the generated XLIFF files, which would be noisy and could trigger unnecessary re-translation. Ideally this would be stable and predictable.

We could prefer shorter expressions? Lexicographic?

---

The larger problem, though, is when we go to produce our output in transform mode.

Let's say we have two `msg` calls with 2 expressions, that get assigned the same id:

```ts
msg(`foo ${a} bar {b}`)
msg(`foo ${c} bar {d}`)
```

So as before, the first `msg` call we encountered with that id wins as the "canonical" one, and we generate XLIFF like this:

```xml
<source>foo <x equiv-text="${a}"/> bar <x equiv-text="${b}"/></target>
```

Now let's say we get a translation back where the order of the `<x>` tags was reversed (which could happen for reasons like a difference in subject/verb/object ordering between two languages):

```xml
<target>xxx <x equiv-text="${b}"/> yyy <x equiv-text="${a}"/></target>
```

In transform mode, we would _want_ to generate this:

```ts
msg(html`xxx <b>${b}</b> yyy {a}`)
msg(html`xxx <b>${d}</b> yyy {c}`)
```

But... we'll hit a problem. When generating translated `msg` calls, we currently rely on the fact that the expressions in the source file text will match the ones in the `equiv-text` XLIFF attributes. This is because we need to be able to find the AST node that corresponds to each expression, and insert it into the placeholders in the newly generated translated template.

This logic happens here: https://github.com/lit/lit/blob/5ac025bf9610adb7069ef8d88ed6bd96ff730f2f/packages/localize-tools/src/modes/transform.ts#L374

The problem is that in the second `msg` call, the source file text won't match the `equiv-text` attributes. We're trying to find the positions of "d" and "c", but we only see "a" and "b". We didn't previously have this problem because we enforced that the source text for all expressions were identical whenever two `msg` calls had the same id.

---

So, I think we'd need to do something smarter here to make this work, where instead of mapping `source expression -> xliff expression`, we instead do `source expression -> canonical source expression -> xliff expression`. We should be able to do the new `source -> canonical` mapping by using the _offset_ of the expressions.

So in this case we would figure out that "c" in the second `msg` call corresponds to "a" in the first `msg` call because it is the first expression. Same idea for "d" corresponding to "b". Now we have the _canonical_ expression values, and we will be able to match the `equiv-text` attributes correctly.

augustjk: I see. So if I understood this correctly, we need to decide on some rule to determine what the canonical source expressions are going to be so the generated xliff is stable, but also we need to account for the ordering of the expression and be able to map that to a different set of source expressions from a message with the same id for transform mode. (For runtime mode this isn't an issue because we're already replacing expressions with indexes?)

aomarks: > I see. So if I understood this correctly, we need to decide on some rule to determine what the canonical source expressions are going to be so the generated xliff is stable, but also we need to account for the ordering of the expression and be able to map that to a different set of source expressions from a message with the same id for transform mode. (For runtime mode this isn't an issue because we're already replacing expressions with indexes?)

Yep, I think that's exactly right!

It would be nice to include a test that verifies that runtime mode also works when a translation changes expression order, but I think you're correct that it will be unaffected by this issue.

packages/localize-tools/src/messages.ts
```diff
@@ -66,6 +66,7 @@ export interface Bundle {
  */
 export interface Placeholder {
   untranslatable: string;
+  index?: string;
```
aomarks, line 69: Number instead of string here I think?

packages/localize-tools/src/modes/transform.ts
```diff
@@ -331,9 +331,17 @@ class Transformer {
     if (templateResult.error) {
       throw new Error(stringifyDiagnostics([templateResult.error]));
     }
-    const {tag} = templateResult.result;
+    const {tag, contents} = templateResult.result;
     let {template} = templateResult.result;
 
+    const placeholdersByIndex = new Map<string, Placeholder>();
```
aomarks, line 337: Can use a number as a Map key here instead of string

packages/localize-tools/src/modes/transform.ts
```diff
@@ -331,9 +331,17 @@ class Transformer {
...
+    let idx = 0;
+    for (const content of contents) {
```
aomarks, line 339: I think here I'd probably do:

```ts
for (let i = 0; i < contents.length; i++) {
  const content = contents[i];
  ...
```

to keep the index variable more tightly scoped. I suppose you could also do:

```ts
{
  let idx = 0;
  for (const content of contents) {
    ...
}
```

but I find the first firm a more readable because it's a classic pattern, as opposed to the block which is a bit unusual

packages/localize-tools/src/modes/transform.ts
```diff
@@ -345,10 +353,10 @@ class Transformer {
     if (ts.isTemplateExpression(template)) {
       for (const span of template.templateSpans) {
         // TODO(aomarks) Support less brittle/more readable placeholder keys.
```
aomarks, line 355: Could remove this TODO now, since you've fixed it!

packages/localize-tools/src/modes/transform.ts
```diff
@@ -345,10 +353,10 @@ class Transformer {
...
-        const key = this.sourceFile.text.slice(
-          span.expression.pos,
-          span.expression.end
-        );
+        const key = this.sourceFile.text
+          .slice(span.expression.pos, span.expression.end)
+          .replace(/{ /g, '{')
```
aomarks, line 358: Curious about what's going on here, why is it needed? Seems like there could be some cases where this could break, such as `${"{ "}` (a literal string expression that would match the same regex).

packages/localize-tools/src/tests/transform.unit.test.ts
```diff
@@ -298,15 +307,15 @@ test('msg(string(msg(string))) translated', () => {
...
+    '`Hola &lt;b&gt;Mundo&lt;/b&gt;!`;',
     {
       messages: [
         {
           name: 'foo',
           contents: [
-            'Hola ',
-            {untranslatable: '<b>${msg("World", {id: "bar"})}</b>'},
-            '!',
+            'Hola <b>',
```
aomarks, line 316: Nice catch!

aomarks (changes requested):

packages/localize-tools/src/messages.ts
augustjk: Here and for the comment below, I chose string since when these are added as attributes on the placeholder tags in the translation, they become strings and I'm eventually looking up by that. I guess I could cast that to a number when comparing too, so it could be either.

packages/localize-tools/src/modes/transform.ts
augustjk: I do want to keep `index` more tightly scoped but the former wouldn't work because it's not every `content` should increment, just the ones that are objects. So I think I'll go with the latter. Good call that curly braces can be used on their own for scoping!

packages/localize-tools/src/modes/transform.ts
augustjk: As we extract from the source code we run into this part https://github.com/lit/lit/blob/main/packages/localize-tools/src/program-analysis.ts#L332. If we get an object literal within an expression (like an options object to an inner `msg` call), the typescript printer will add spaces after/before opening/closing brackets regardless of what the source had and we store that in the `untranslatable` portion of a placeholder.

As we transform, we reconstruct the message by stitching together the translation contents, including the placeholder (which now includes bracket spaces). If the source code does not have bracket spaces within expressions, the transform errors as we can't find a matching expression now.

I'll try to think on it more but a way to make them congruent without forcing these replaces isn't coming to me right now.

augustjk: If we know the expression in the translation will always have bracket spaces (and I don't see an option for that in the typescript printer), we could take the source expression and run it through prettier with `bracketSpacing: true` to make them consistent.

aomarks: I see! Nicely sleuthed.

I'm thinking we might want to switch from using `printer.printNode` (https://github.com/lit/lit/blob/13b0a411a3c002540a687d0d70b499f801bf39e6/packages/localize-tools/src/program-analysis.ts#L332) to using the same source text extraction approach that we use in `transform.ts` (https://github.com/lit/lit/blob/13b0a411a3c002540a687d0d70b499f801bf39e6/packages/localize-tools/src/modes/transform.ts#L348). So:

```ts
file.text.slice(span.expression.pos, span.expression.end);
```

Seems like that would give us identically formatted text in both places.

augustjk: Hah, yeah that does it. The only concern for me then is that this will change the `extract` behavior as well and potentially change the placeholders in the extracted xliff/xlb. If a user originally had no bracket spaces, they'd have to re-extract and update the placeholders to match the new source for any already translated portions.

aomarks: That's true. I suppose the other option is to use `printer.printNode` in that transform mode spot then. Should have the same effect.

packages/localize-tools/src/modes/transform.ts
```diff
@@ -341,59 +341,115 @@ class Transformer {
...
-            typeof content === 'string'
-              ? escapeTextContentToEmbedInTemplateLiteral(content)
-              : content.untranslatable
-          )
+          .map((content) => {
+            if (typeof content === 'string') {
+              return escapeTextContentToEmbedInTemplateLiteral(content);
+            }
+            if (content.index === undefined) {
+              throw new Error(
```
aomarks, line 377: I think this will currently throw on a transform when the input is XLB, since we won't have indexes in those files. I think that means https://github.com/lit/lit/issues/1897 needs to be fixed as part of this PR too or else a release would break all XLB use cases.

(To be honest, as far as I know there is only one customer of XLB, and I believe they are on a much older version anyway -- so it's unlikely anybody would be hit by this. However we should still not introduce the breakage [and side note maybe reach out to them to see what their status is!]).

packages/localize-tools/src/messages.ts
```diff
@@ -66,6 +66,7 @@ export interface Bundle {
  */
 export interface Placeholder {
   untranslatable: string;
+  index?: number;
```
aomarks, line 69: This should be required (related to other comment).

aomarks (changes requested):
Looks great! I think we're good to go with https://github.com/lit/lit/issues/2140, but I believe https://github.com/lit/lit/issues/1897 will need to be fixed as part of this PR too to avoid breaking the XLB use case.

packages/localize-tools/src/modes/transform.ts
aomarks: Just to clarify, this will need to be a breaking change for existing XLB files, but we should at least make sure it works for fresh builds.

Existing XLB files could then be migrated manually. (I think if we had a lot of XLB users, I would say we should provide a backwards-compatibility function that tries to make a best guess at the index from the contents of the `<ph>` tag when the index is missing, but I don't think we need it given the small [1] number of users).

packages/localize-tools/src/formatters/xlb.ts
```diff
@@ -90,7 +91,9 @@ class XlbFormatter implements Formatter {
           ) {
             throw new KnownError(`Expected <ph> to have exactly one text node`);
           }
-          contents.push({untranslatable: phText.nodeValue || ''});
+          const index =
+            Number((child as Element).getAttribute('name')) || phIdx++;
+          contents.push({untranslatable: phText.nodeValue || '', index});
```
augustjk, line 96: @aomarks this is an attempt for keeping existing xlb compatibility (dumbly assigning indexes as they appear, since matching by content seemed like a lot more work) but I wonder if it's actually more dangerous than throwing since it can silently mess up the order of the placeholders in the translation.

aomarks: Yeah, good thought, but I agree that it could actually be harmful because it would silently generate incorrect templates in the case that a placeholder was repositioned. I think it's fine to throw here.

.changeset/many-singers-complain.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/localize-tools': minor
+---
+
+**BREAKING** Update analysis to consider messages with same id **and** description to be identical (but no longer checks for expressions to be same) and improve error message on finding incompatible duplicates.
+
+`lit-localize extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable message or add different `id` options to differentiate them.
```
aomarks, line 7:
```suggestion
`lit-localize extract` will now error if multiple messages had the same text but different `desc` option. Be sure to add the same `desc` option for these messages to be considered the same translatable message or add different `id` options to differentiate them.

**BREAKING** XLB files now require `name` attributes on all `<ph>` elements. If you encounter errors relating to missing `name` attributes, run `lit-localize extract` to re-generate your XLB files from source with new `name `attributes.
```

aomarks (approved):
Awesome! Looks great. Just one suggestion to mention the XLB breaking change in the changelog too.

.changeset/many-singers-complain.md
aomarks: Oops sorry, I must have just scrolled past that by mistake. Nope, what you already have is great!

PR #2406, opened 2022-01-14
[ssr] Fix typo in ssr readme. Fixes #2381

### LINKED ISSUE #2381: [ssr] README has an inaccuracy
# Description

The client usage hydration part of the README tries to import `render` from `experimental-hydrate.js` but it has no render function.

```ts
import {render, hydrate} from `lit/experimental-hydrate.js`;
...
const update = (data) => render(myTemplate(data), document.body);
```

This fails.

This appears to work.

```ts
import {render} from `lit-html`;
import {hydrate} from `lit/experimental-hydrate.js`;
...
const update = (data) => render(myTemplate(data), document.body);
```

kevinpschaaf:


packages/labs/ssr/README.md
```diff
@@ -73,8 +73,8 @@ context.body = Readable.from(ssrResult);
 
 ```js
 import {myTemplate} from './my-template.js';
-import {render, hydrate} from `lit/experimental-hydrate.js`;
-
+import {render} from `lit-html`;
```
aomarks, line 76: Why not from `lit`?

aomarks (approved):

packages/labs/ssr/README.md
kevinpschaaf: No reason, I think? I'm guessing there was a timing thing with when we added lit and that just didn't get updated? I don't think there's a reason.

aomarks (approved):

PR #2407, opened 2022-01-15
Add an ADR for bundle distribution.

bicknellr:
This ADR describes why bundles are distributed by committing them to a separate repo.

We should probably come up with a template and expected workflow for these if we want them to be a regular thing. I just copied the basic structure from the example in [the 18F site you linked](https://18f.gsa.gov/2021/07/06/architecture_decision_records_helpful_now_invaluable_later/), but I don't think that example is intended to be totally prescriptive. (Also, I did exactly the wrong thing for the file name according to them.)


ADRs/2022-01-14_bundle-distribution.md
```diff
@@ -0,0 +1,27 @@
...
+The bundles will be distributed by committing them to a new repo
+(https://github.com/lit/bundles) at the same time `lit` is published to npm.
+
+## Consequences
+
+- The separate repo allows them to be tagged with unprefixed versions (vs.
+  package-prefixed tags, e.g. `lit@2.0.0`), which is necessary for compatibility
+  with jsDelivr's scheme for specifying an imprecise version when requesting a
+  file from a repo on GitHub. (See https://jsdelivr.com/features for more info.)
+- Avoiding including the bundles in the npm packages makes it clear that they
```
kevinpschaaf, line 25: Maybe include something like: 
>This avoids a risk that users might publish a reusable package depending on a bundle rather than lit source, leading to duplication of the lit library when users install such packages alongside others that do correctly depend on lit source.

justinfagnani (approved):

rictic (approved):

PR #2420, opened 2022-01-19
Use a different GitHub action for project issue automation

aomarks:
Turns out https://github.com/peter-evans/create-or-update-project-card doesn't support the new beta versions of projects. I found a similar library that does: https://github.com/leonsteinhaeuser/project-beta-automations. Switching to that.

### DIFF
--- a/.github/workflows/add-issues-to-project.yml
+++ b/.github/workflows/add-issues-to-project.yml
@@ -1,16 +1,16 @@
 # GitHub action that automatically adds newly filed issues to the Lit project.
-# See https://github.com/peter-evans/create-or-update-project-card for documentation.
+# See https://github.com/leonsteinhaeuser/project-beta-automations for documentation.
 on:
   issues:
     types: [opened]
 jobs:
-  createCard:
+  add-issue-to-project:
     runs-on: ubuntu-latest
     steps:
-      - name: Create project card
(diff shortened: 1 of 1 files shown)


.github/workflows/add-issues-to-project.yml
```diff
@@ -1,16 +1,16 @@
...
+        uses: leonsteinhaeuser/project-beta-automations@v1.1.0
         with:
-          token: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
-          project-location: lit
-          project-number: 4
-          column-name: No Status
+          gh_token: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
+          organization: lit
+          project_id: 4
+          resource_node_id: ${{ github.event.issue.node_id }}
```
bicknellr, line 16: Does this need a line like `status_value: No Status` to indicate what status newly opened issues should be given? Or does that happen automatically and 'No Status' is an out-of-band concept?

aomarks: "If left unspecified, new items are added without an explicit status, and existing items are left alone." - https://github.com/leonsteinhaeuser/project-beta-automations

bicknellr (approved):

PR #2422, opened 2022-01-19
Add type declarations for the bundles.

bicknellr:
- [ ] Add the type bundles to the commit pushed into the lit/bundles repo.


package.json
```diff
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@changesets/cli": "^2.16.0",
     "@changesets/get-github-info": "^0.5.0",
+    "@microsoft/api-extractor": "^7.19.4",
```
bicknellr, line 29: I had trouble installing this as a dev dependency in `packages/lit`. AFAICT, it works due to npm's default lookup behavior but maybe it should be moved into `packages/lit/package.json`?

PR #2424, opened 2022-01-20
Use graphql API directly to automate adding issues to project

aomarks:
The workflow I was trying from https://github.com/leonsteinhaeuser/project-beta-automations didn't work (filed https://github.com/leonsteinhaeuser/project-beta-automations/issues/27).

But I realized we could use the graphql API directly to get more control here. This new action is based on the example at https://docs.github.com/en/issues/trying-out-the-new-projects-experience/automating-projects#example-workflow-authenticating-with-a-personal-access-token, just slightly modified and simplified.

I've tested this in my personal repo and it works, so hopefully it will work for us too.

### DIFF
--- a/.github/workflows/add-issues-to-project.yml
+++ b/.github/workflows/add-issues-to-project.yml
@@ -1,5 +1,7 @@
 # GitHub action that automatically adds newly filed issues to the Lit project.
-# See https://github.com/leonsteinhaeuser/project-beta-automations for documentation.
+#
+# Based on the example at
+# https://docs.github.com/en/issues/trying-out-the-new-projects-experience/automating-projects#example-workflow-authenticating-with-a-personal-access-token
 on:
   issues:
     types: [opened]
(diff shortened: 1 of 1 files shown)


.github/workflows/add-issues-to-project.yml
```diff
@@ -1,16 +1,43 @@
...
-      - name: Add issue to Lit project
-        uses: leonsteinhaeuser/project-beta-automations@v1.1.0
-        with:
-          gh_token: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
-          organization: lit
-          project_id: 4
-          resource_node_id: ${{ github.event.issue.node_id }}
+      # Finds the global entity ID for the project and adds it to an environment variable
+      # for the next step to consume.
+      - name: Get project ID
```
AndrewJakubowicz, line 14: Is the global entity ID for a project static? I.e. could be stored directly in secrets?

AndrewJakubowicz (approved):
LGTM

.github/workflows/add-issues-to-project.yml
aomarks: It is, and it's not a secret so it could also just be inlined here. However it's a weird identifier that's a bit hard to find because you have to do an API query to get it, so it's more convenient to just query for it directly here. That way this script can be updated easily, used in other contexts etc.

PR #2430, opened 2022-01-21
Provide better paths for sources in the bundles' source maps.

bicknellr:
By default, the paths in the source map are relative paths from the bundle to the included sources as they were on disk at the time the bundle was generated. This causes the developer tools to display the original sources at seemingly unrelated locations. This PR transforms them so that the original locations appear to be in a single directory with the same name as the bundle with the same structure as the `packages` directory.

For example, rather than the source map for `lit.min.js` saying that the main `lit-html` entrypoint is located at `../lit-html/src/lit-html.ts`:

```
+ lit-html (generated by source map)
| + src
|   + lit-html.ts
+ lit-element (generated by source map)
| + ...
+ dir-with-bundle
  + lit.min.js
```

It will say that it is at `./lit.min.js/lit-html/src/lit.html.ts` instead:
 
 ```

+ dir-with-bundle
   + lit.min.js
   + lit.min.js (generated by source map)
     + lit-html
     | + src
     |   + lit-html.ts
     + lit-element
       + ...
```

 (tl;dr: `path.join(bundleName, relativeFromPackagesDirToSource)`)

### DIFF
--- a/packages/lit/rollup.config.js
+++ b/packages/lit/rollup.config.js
@@ -6,6 +6,27 @@
 
 import {litProdConfig} from '../../rollup-common.js';
 import {createRequire} from 'module';
+import * as path from 'path';
+
+/**
+ * Takes a `relativeSourcePath` and `sourcemapPath` - with the same semantics as
+ * those provided to the function given to Rollup's
+ * [`output.sourcemapPathTransform` config option][1] - and produces a relative
+ * path to the source file from the `packages` directory of this repo.
+ *
+ * [1]: https://rollupjs.org/guide/en/#outputsourcemappathtransform
+ */
+const makeRelativeToPackagesDir = (relativeSourcePath, sourcemapPath) => {
+  const absoluteSourcePath = path.resolve(
+    path.join(path.dirname(sourcemapPath), relativeSourcePath)
+  );
+  const absolutePackagesDirPath = path.resolve(path.join(__dirname, '..'));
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/lit/rollup.config.js [hunk 1]
```diff
@@ -58,11 +79,30 @@ export default litProdConfig({
...
       format: 'es',
+      sourcemapPathTransform: (relativeSourcePath, sourcemapPath) => {
+        // Convert the paths of the sources to appear as descendants of a
+        // directory with the same name as the bundle. (By default, the paths in
+        // the source map are relative paths from the bundle to the included
+        // sources as they were on disk at the time the bundle was generated.
+        // This causes the developer tools to display the original sources at
+        // seemingly unrelated locations.)
+        return path.join(
+          'lit.min.js',
```
justinfagnani, line 90: Do we want the bundle name in there? If that's left off won't the path be relative to the folder containing the bundle, and be pretty good? Seems maybe weird to treat the file as a folder too. I wonder if that would trip up tools that try to display a file hierarchy?

bicknellr: I tried loading it in Chrome and Firefox and they both seem to be fine with having a file and directory with the same name. IMO, it's clearer to keep them in this position rather than making them siblings because the folder is shown right next to the bundle vs. being potentially spread out across the parent folder.

justinfagnani (approved):

Issue #2129, opened 2021-09-02
[react] React wrapped components do not offer any way to customise the type of events dispatched


WickyNilliams:
# Description

`createComponent` from `@lit-labs/react` does not offer any way to customise the type of events being dispatched. As can be seen in the following lines, `Event` is always used:

https://github.com/lit/lit/blob/efe88ba5a31bae117f8a4c8ddf111068368cf249/packages/labs/react/src/create-component.ts#L87-L89

# Steps to Reproduce

Consider some code in a lit component like:

```js
this.dispatchEvent(
  new CustomEvent<string>('some-event', {
    detail: "foobar",
    bubbles: true,
    composed: true,
  })
);
```

And a react wrapper like:

```ts
export const MyComponent = createComponent(React, "my-component", MyLitComponent, {
  onSomeEvent: "some-event",
});
```

TS generates a type like this from the react wrapper:

```ts
onSomeEvent: ((e: Event) => unknown) | undefined;
```

`Event` is incompatible with `CustomEvent<string>` in type terms, so I _cannot_ re-type at the call site like this without an error:

```ts
{/* does not work! */}
<MyComponent
  onSomeEvent={(e: CustomEvent<string>) => console.log(e.detail)}
/>
```

It would be nice if `createComponent` supported some form of generics for defining event types, since I can likely auto-generate this from my custom elements manifest (at least to some extent: seems the event type there is only `CustomEvent` and not `CustomEvent<string>`)

## Live Reproduction Link

N/A

## Expected Results

TS should know that `e` here is `CustomEvent<string>`, without needing to explicitly state the type:

```ts 
<MyComponent
  onSomeEvent={(e) => console.log(e.detail)}
/>
```

## Actual Results

I have been forced to do this, which is a shame, as we lose all type safety:

```ts
<MyComponent
  onSomeEvent={(e: any) => console.log("who needs types, anyway?!", e.detail)}
/>
```

## Browsers Affected

N/A

rickardlindahl:
I would also like to see this feature implemented.

As a workaround you can type cast the event like this:

```typescript
<MyComponent
  onSomeEvent={(e) => {
    const customEvent = e as CustomEvent<string>;
    console.log("You can now access the detail prop without errors", customEvent.detail);
  }}
/>
```

PR #2445, opened 2022-01-25
[@lit-labs/react] support custom events

taylor-vann:
In regards to issue #2129

Allow Custom Events to be passed as arguments in event listeners.


taylor-vann:
@justinfagnani @sorvell 
This PR is ready for review :)

packages/labs/react/src/create-component.ts
```diff
@@ -6,6 +6,20 @@
...
+  [K in keyof R]?: EventListeners;
+};
+
+type ReservedReactProperties =
+  | 'children'
+  | 'localName'
+  | 'ref'
+  | 'style'
+  | 'className';
+
```
justinfagnani, line 22: To avoid repeating the property names you can infer the type from a const array: https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgUwguA3BATASgghsKABThAAcUoBPAOVwFskYBeGAbQCgYYByYACwCWAGwyIw3ADSceQkMFxDaDSdO6IAZiq7dolIQi09gQ3BAhKDUgLoxTMUJCgBudg+jwkKdNjwFiZCgFGFjAEAHcYAGUEKAAKAEoXdipyGABRfQYwKAAVSnIAHhyYBAAPKAQwDAgYHFwMcCFKAEE4OFxKAoBXMABrMBAwsAA+YeZpYrKKqpq6hrAm1vbOgTB1FHTMytz8hGHpLgB+TYQsnfIDmAAuGFD0OCSUhFrPNEw6vxJyOCoLZhOznlCk8QOoPMg3j58EQvhQaPQkPtXOB3KQvhYbjgId4PjCAj94Qx-tx1CAQNwnEA

packages/labs/react/src/create-component.ts
```diff
@@ -16,84 +30,77 @@ const reservedReactProperties = new Set([
...
-      // Otherwise just update the listener with new value
-    } else {
-      handler.handleEvent = listener;
-    }
-    // Remove listener if one exists and value is undefined
-  } else if (handler !== undefined) {
+
+  // Remove listener if one exists
+  const handler = events.get(event);
+  if (handler !== undefined) {
```
justinfagnani, line 56: The point of the current code was to avoid thrashing the DOM with `addEventListener`/`removeEventListener` calls when fresh closures are passed on repeat renders.

Imagine this JSX:

```tsx
<SomeWebComponent onFoo={(e) => console.log(e)}></SomeWebCompoent>
```

Every render that will pass a new function to `SomeWebComponent.onFoo`. With the current code `addEventListener` is only called once.

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -44,6 +44,10 @@ class BasicElement extends ReactiveElement {
   fire(name: string) {
     this.dispatchEvent(new Event(name));
   }
+
+  ice(name: string) {
```
justinfagnani, line 48: What's this for?

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -250,6 +255,26 @@ suite('createComponent', () => {
     const onBar = (e: Event) => {
       barEvent = e;
     };
+
+    const onCustomBar = (e: CustomEvent<string>) => {
```
justinfagnani, line 259: I don't think we need a separate check for `CustomElement`, just narrow the type of event used in `onFoo` and check that a property in the event type but not in `Event` is accessible without a type-error.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
```diff
@@ -6,6 +6,20 @@
 
 import * as ReactModule from 'react';
 
+type Constructor<I> = {new (): I};
+type CustomEventListener = (e: CustomEvent) => void;
```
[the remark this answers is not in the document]
taylor-vann: This was to add a `EventListener | CustomEventListener` type union for our event map.

I couldn't find a custom event listener in the default ts lib, down to change if one is availble though!

packages/labs/react/src/create-component.ts
taylor-vann: i'll try!

packages/labs/react/src/create-component.ts
taylor-vann: Nice! I was trying to mimic that behavior but i guesss i biffed it.

The aim was to leverage the diff check in `create-component.ts` line 81. My logic was, if there was an update to the event listener, we'd get pass that diff check and remove / add the new event listener.

taylor-vann: I think accounting for repeat closures is not 'react-like' but i'm open about this. My rationale is: If a user relies on currying data in a closure, but we don't update those closures, their closure is stale.

Also in a react review I might say:

`<SomeWebComponent onFoo={(e) => console.log(e)}></SomeWebCompoent>`

should be

```
const logger = createCallback((e) => console.log(e)), []);

return (
  <SomeWebComponent onFoo={(e) => console.log(e)}></SomeWebCompoent>
)
```

because in react that closure will be updated every render

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: that was me having fun with a function name, fire and ice. will change to whatever is best though

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: sounds good!

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: It was to have a separate function to dispatch a custom event for a test

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
taylor-vann: can confirm
https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L3693-L3705
CustomEvent is declared outside and unlinked to the dom map.

I assume it's expected that users with typecast their event handlers with `CustomEvent<type>`. Adding events to the DOM Event map feels a little extra. So my thoughts were "let's add event listeners to our accepted types"

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: I can't get the visual behavior to happen
I think this is more about what types the labs package narrows, not what events can be added.

https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L5113
https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L3693-L3700

taylor-vann: However, I _can_ get the visual behavior when I remove this file's CustomEventListener

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: removed new tests.

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: removed new tests.

packages/labs/react/src/create-component.ts
taylor-vann: reverted logic back to original

packages/labs/react/src/create-component.ts

taylor-vann:
this PR is ready for another review :)

WickyNilliams:
So what is the result of this? That you can instead accept `CustomEvent` as a param? Can you manually configure types beyond that? e.g if you have `class MyEvent extends Event {}`

taylor-vann:
@WickyNilliams Yes :)

Currently a user could extend an event like you pointed out (probably a more durable typing).

This PR would allow someone to provide an event listener _and_ a custom event listener.

packages/labs/react/src/create-component.ts
taylor-vann: Resolving because I think the behavior you pointed out is accounted for in 'react philosophy'.

A fresh closure would thrash the dom in react and I believe the same behavior should be expected in ours as well.

taylor-vann:
Component properties now reflect in the react component props through the props map.

```
const propsMap = {
  onFoo: 'foo', // <- expects (e: KeyboardEvent) => {}
  onBar: 'bar', // <- expects (e: CustomEvent<string>) => {}
};
```

sorvell (changes requested):
It seems like there are some stylistic changes in this PR that aren't directly related to fixing the core issue of typing the event arguments. Please separate those changes into a separate PR and focus this PR exclusively to fixing the referenced issue. That will make it easier to review. Thanks.

taylor-vann (commented):
lots of changes, needs another look

taylor-vann:
@sorvell I can make another PR that's no problem

taylor-vann:
@justinfagnani @sorvell 
more concise PR #2648

taylor-vann:
closing for: #2648 2648

PR #2456, opened 2022-01-26
[ssr] Fix typing issues for users

aomarks:
Fix TypeScript typing issues for users of `@lit-labs/ssr`. `@lit-labs/ssr` doesn't currently compile cleanly when imported by TypeScript users with `skipLibCheck:false`, because our public typings rely on typings that aren't being provided:

- Added dependency on `@types/node` because `url.URL` is part of the `ModuleLoader` API.

- Tightened up the types in our `vm.Module` augmentations.

- Provided a new `VmModule` interface that we now use instead of `vm.Module` on all public APIs, because consumers do not have access to `vm.Module` typings (they are not included in `@types/node`), and we don't want to augment them for our consumers.


packages/labs/ssr/src/lib/module-loader.ts
```diff
@@ -11,6 +11,8 @@ import * as vm from 'vm';
 import resolveAsync from 'resolve';
 import {builtinModules} from 'module';
 
+import './typings/node.js';
```
justinfagnani, line 14: I think we could define our own type for use in the public APIs (only `ModuleRecord`?) where we now have references to `vm.Module`.

aomarks: Added a new `VmModule` type for public APIs.

packages/labs/ssr/package.json
```diff
@@ -33,38 +33,39 @@
...
     "koa-static": "^5.0.0",
-    "koa": "^2.7.0",
     "mocha": "^9.1.1",
     "typescript": "^4.3.5",
     "uvu": "^0.5.1"
   },
   "dependencies": {
     "@lit-labs/ssr-client": "^1.0.0",
     "@lit/reactive-element": "^1.1.0",
+    "@types/node": "^17.0.12",
```
augustjk, line 65: i'm not sure how much this matters but is node v17 the recommended version for ssr? perhaps we should update the `engines.node` value in `package.json` as well.

aomarks: It doesn't require 17. Might technically work from v12 in global mode, since that's when es modules were added. ES module experimental flag was removed in v13.2.0/v12.17.0. Experimental VM module support was added in v13.0.0/v12.16.0.

I suppose you could make the case that we should be using the oldest support version of the typings? A problem with that, though, is that I'm not sure how well updated the older major versions are with typing bug fixes. I suspect bug-fixes are just added to the latest version.

justinfagnani (approved):

packages/labs/ssr/package.json
augustjk: Looking at version history on npm, it does seem like they've done patch updates for even v12 fairly recently. https://www.npmjs.com/package/@types/node/v/12.20.42 was published 10 days ago. If we don't go as far back as 12, still maybe using typings for even/LTS version makes more sense?

aomarks: Active LTS seems reasonable. Switched to 16.

packages/labs/ssr/package.json
```diff
@@ -33,38 +33,39 @@
...
     "@types/command-line-args": "^5.0.0",
-    "@types/koa__router": "^8.0.2",
+    "@types/koa": "^2.0.49",
     "@types/koa-cors": "*",
     "@types/koa-static": "^4.0.1",
-    "@types/koa": "^2.0.49",
+    "@types/koa__router": "^8.0.2",
     "@types/mocha": "^9.0.0",
-    "@types/node-fetch": "^2.5.7",
     "@types/node": "^16.7.8",
```
augustjk, line 45: Just noticed we also have this in `devDependencies`. We should remove from here and keep the one in `dependencies`?

aomarks: Good catch, thanks. Removed.

augustjk (approved):

PR #2457, opened 2022-01-26
[reactive-element] Expand jsdoc documentation, and misc code sample syntax fixes.

AndrewJakubowicz:
Part of https://github.com/lit/lit/issues/1883#issuecomment-1018777254

This PR was originally to make some minor fixes to the reactive-element code samples, but I expanded it slightly to also include documentation for `willUpdate`, `firstUpdated`, and the `attributeChangedCallback`.

The documentation updates are scoped to the reactive-element package.

### DIFF
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -17,7 +17,7 @@ import type {ReactiveElement} from '../reactive-element.js';
 import type {QueryAssignedNodesOptions} from './query-assigned-nodes.js';
 
 /**
- * A tiny module scoped polyfill for HTMLSlotElement.prototype.assignedElements.
+ * A tiny module scoped polyfill for HTMLSlotElement.assignedElements.
  */
 const slotAssignedElements =
   window.HTMLSlotElement?.prototype.assignedElements != null
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -17,7 +17,7 @@ import type {ReactiveElement} from '../reactive-element.js';
 import type {QueryAssignedNodesOptions} from './query-assigned-nodes.js';
 
 /**
- * A tiny module scoped polyfill for HTMLSlotElement.assignedElements.
+ * A tiny module scoped polyfill for HTMLSlotElement.prototype.assignedElements.
  */
 const slotAssignedElements =
   window.HTMLSlotElement?.prototype.assignedElements != null
@@ -33,7 +33,7 @@ const slotAssignedElements =
 /**
  * Options for the {@linkcode queryAssignedElements} decorator. Extends the
  * options that can be passed into
- * [HTMLSlotElement.assignedElements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements).
+ * [slot.assignedElements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements).
  */
 export interface QueryAssignedElementsOptions
   extends QueryAssignedNodesOptions {
--- a/packages/reactive-element/src/decorators/query-assigned-elements.ts
+++ b/packages/reactive-element/src/decorators/query-assigned-elements.ts
@@ -33,7 +33,7 @@ const slotAssignedElements =
 /**
  * Options for the {@linkcode queryAssignedElements} decorator. Extends the
  * options that can be passed into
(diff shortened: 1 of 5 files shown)


augustjk (approved):
Nice! Helpful comments and examples!

packages/reactive-element/src/decorators/query-assigned-nodes.ts
```diff
@@ -79,7 +79,8 @@ export function queryAssignedNodes(
  *
  * Note the type of this property should be annotated as `Array<Node>` if used
  * without a `selector` or `Array<HTMLElement>` if a selector is provided.
- * Please use `@queryAssignedElements` if using a CSS selector is desired.
+ * Please use {@linkcode queryAssignedElements @queryAssignedElements} if using
+ * a CSS selector is desired.
```
arthurevans, line 83: This comment makes sense in the context of deprecating the arg to queryAssignedNodes and creating the new queryAssignedElements... But to a user coming in without this background, it doesn't make as much sense. 

That is, why use `queryAssignedElements`? Because I want to see only elements... and (maybe) filter them using a CSS selector.

```suggestion
 * Use {@linkcode queryAssignedElements @queryAssignedElements} to
 * list only elements, and optionally filter the element list using a CSS selector. 
 ```
 
Given that this only shows up in code suggestions for the deprecated form... maybe this suggestion is overkill. (Although I think some people are currently doing `queryAssignedNodes('', true, '*')` to get elements, in which case they don't really want the selector, they just want queryAssignedElements.)
 
If you don't take this suggestion, maybe just make the language active ("if using a CSS selector is desired" => "if you are using a CSS selector")

**Review by arthurevans — APPROVED** [review]
LGTM. Some minor very optional suggestions. Please dismiss them if you don't feel they make sense in the context where this doc will appear.

File: packages/reactive-element/src/decorators/query-assigned-nodes.ts (hunk 1 above)
Comment (AndrewJakubowicz) [reply packages/reactive-element/src/decorators/query-assigned-nodes.ts:83]: I like this suggestion! It's clearer!

File: packages/reactive-element/src/decorators/query-assigned-elements.ts [hunk 2]
```diff
@@ -17,7 +17,7 @@ import type {ReactiveElement} from '../reactive-element.js';
 import type {QueryAssignedNodesOptions} from './query-assigned-nodes.js';
 
 /**
- * A tiny module scoped polyfill for HTMLSlotElement.assignedElements.
+ * A tiny module scoped polyfill for HTMLSlotElement.prototype.assignedElements.
```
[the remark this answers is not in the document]
AndrewJakubowicz: Applied!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -964,6 +964,9 @@ export abstract class ReactiveElement
 
   /**
    * Synchronizes property values when attributes change.
+   *
+   * Specifically, when an attribute is set, the corresponding property is set.
+   * You should rarely need to implement this callback.
```
justinfagnani, line 969: Since this is not our interface, I think we should point to the MDN docs: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks

I think saying that it's rarely needed to be overridden is good, but the most important thing is saying that if you do you must call super.

And not needed now, but I'd also document the method by what it does from the platform (and hopefully we'd be consistent with the other CE callback's docs). Something like "Called whenever one of the element's attributes is changed. This is one of the standard custom element lifecycle callbacks."

packages/reactive-element/src/decorators/query-assigned-nodes.ts
```diff
@@ -18,7 +18,7 @@ import type {ReactiveElement} from '../reactive-element.js';
 
 /**
  * Options for the {@linkcode queryAssignedNodes} decorator. Extends the options
- * that can be passed into [HTMLSlotElement.assignedNodes](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedNodes).
+ * that can be passed into [slot.assignedNodes](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedNodes).
```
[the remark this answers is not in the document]
AndrewJakubowicz: Can revert! Thank you!

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: This is super useful. I can address.

AndrewJakubowicz: Added these notes. PTAL! Thank you!

justinfagnani (approved):

Issue #2352, opened 2021-12-09
[11ty] Release eleventy (11ty) plugin


aomarks:
Some work started already in https://github.com/lit/lit/tree/lit-next-ssr-11ty

PR #2463, opened 2022-01-28
[@lit-labs/eleventy-plugin-lit] Switch to ESM, add tests, misc cleanup

aomarks:
- Switch from using `require()` to `await import()` to load component definitions, so that we can load ES modules. This means the UMD or CJS requirement is removed.

  Major caveat here is that it doesn't work with Eleventy's `--watch` mode, because the ESM cache can't be cleared. **In a followup, I'll be switching to isolated VM modules.** (Also note that the previous approach of using `require()` didn't work either, though we could have cleared the cache because Node does expose APIs for the `require()` cache).

- Add some integration tests.

- Miscellaneous infrastructure cleanup.

Part of https://github.com/lit/lit/issues/2352


.changeset/gentle-dolphins-draw.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
augustjk, line 2: Empty changeset file?

packages/labs/eleventy-plugin-lit/test/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "__note__": "Switches Node into module mode for tests",
+  "type": "module"
+}
```
augustjk, line 4: If we need this just for tests, what do you think about adding `NODE_OPTIONS='--input-type=module'` in front of the test command to achieve that? We can also use [cross-env](https://www.npmjs.com/package/cross-env) for cross-platform safe way of setting environment variables if wanted.

.changeset/gentle-dolphins-draw.md
aomarks: Yep, since it's not released, it's unhelpful to write incremental changes in the public changelog (because from the user's perspective, nothing changed). `npx changeset --empty` creates an empty one which is what the failing changeset action suggests to do to silence it.

packages/labs/eleventy-plugin-lit/test/package.json
aomarks: The `"type":"module"` approach seems simpler to me, and it means `npx uvu` will work normally, instead of having to remember to deal with extra envs.

augustjk (approved):

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -0,0 +1,183 @@
...
+      import { html, LitElement } from 'lit';
+      class MyElement extends LitElement {
+        render() {
+          return html\`<b>shadow content</b>\`;
+        }
+      }
+      customElements.define('my-element', MyElement);
+    `,
+  });
+  assert.equal((await rig.exec('eleventy --config .eleventy.cjs')).code, 0);
```
justinfagnani, line 140: amazing test rig! 🎉

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -0,0 +1,183 @@
...
+  assert.equal(
+    await rig.read('_site/index.html'),
+    // TODO(aomarks) The way we inject <head> and <body> doesn't seem quite
+    // right. Investigate.
+    normalize(`
+      <html><head></head><!--lit-part BRUAAAUVAAA=--><!--lit-part 7fcMKuWrg1g=--><body><h1>Heading</h1>
```
justinfagnani, line 146: SSR test utils so these aren't as fragile would eventually be nice

justinfagnani (approved):

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -27,9 +27,14 @@ module.exports = {
...
             .unsafeHTML;
-          for (const module of options.componentModules ?? []) {
-            require(path.join(process.cwd(), module));
-          }
+          // TODO(aomarks) This doesn't work in Eleventy --watch mode, because
+          // the ES module cache never gets cleared. Switch to isolated VM
+          // modules so that we can control this.
+          await Promise.all(
+            (options.componentModules ?? []).map(
+              (module) => import(path.join(process.cwd(), module))
```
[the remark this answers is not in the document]
aomarks: > You might want path.resolve() here?

Done. (Because of canonicalization? I'm not sure if that affects the module identity for the cache, but maybe?)

> For security, might you want to also make sure that the path is in the project root? Or are there cases you want to support reaching out of the cwd?

I don't think we can really define what the project root is (cwd?), and I don't really see why we shouldn't be able to reach of it anyway. In general, an eleventy build config can access any files it wants to that the OS lets it. Why should we introduce a boundary?

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -0,0 +1,183 @@
...
+  private readonly _tempDir = pathlib.join(tempRoot, String(Math.random()));
+
+  async setup() {
+    // Simulate installing @lit-labs/eleventy-plugin with a symlink to the root
+    // of this package.
+    await fs.mkdir(pathlib.join(this._tempDir, 'node_modules', '@lit-labs'), {
+      recursive: true,
+    });
+    await fs.symlink(
+      '../../../../..',
```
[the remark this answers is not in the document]
aomarks: Done

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
aomarks: Agreed!

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
aomarks: Thanks! I like being able to see all the files defined together within the test function, instead of having to deal with `testdata/` directories which is the other way we could do this.

Only real downside is not having syntax highlighting. Could have a no-op `html` tag for HTML files, but (at least in my VSCode setup) there is no support for `js` or `md` tag highlighting.

PR #2475, opened 2022-02-01
[@lit-labs/eleventy-plugin-lit] Remove unnecessary marker comments

aomarks:
Trims Lit SSR `<!--lit-part-->` and `<?>` marker comments before and after the top-level synthetic template that is rendered for each Eleventy HTML file. I believe these serve no purpose, since the top-level template that we synthetically render could never be hydrated.

This also avoids the problem of marker comments appearing before `<!doctype>` declarations, which I think may have been why we had some special handling around trying to detect or synthesize `<body>` tags previously. So I was able to remove this part of the code as well.

Also includes some miscellaneous minor cleanup.

Part of https://github.com/lit/lit/issues/2352

### DIFF
--- a/packages/labs/eleventy-plugin-lit/src/index.ts
+++ b/packages/labs/eleventy-plugin-lit/src/index.ts
@@ -13,7 +13,8 @@
 // So instead we use TypeScript's ESM output mode, but explicitly write
 // require() calls for the CommonJS modules we import.
 //
-// See https://github.com/microsoft/TypeScript/issues/43329 for more details.
+// See https://github.com/microsoft/TypeScript/issues/43329 and
+// https://github.com/microsoft/TypeScript#22321 for more details.
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const path = require('path') as typeof import('path');
--- a/packages/labs/eleventy-plugin-lit/src/index.ts
+++ b/packages/labs/eleventy-plugin-lit/src/index.ts
@@ -31,35 +31,35 @@ module.exports = {
     eleventyConfig.addTransform(
       'render-lit',
       async (content: string, outputPath: string) => {
-        if (outputPath.endsWith('.html')) {
-          const render = (
-            await import('@lit-labs/ssr/lib/render-with-global-dom-shim.js')
-          ).render;
-          const html = (await import('lit')).html;
-          const unsafeHTML = (await import('lit/directives/unsafe-html.js'))
-            .unsafeHTML;
-          // TODO(aomarks) This doesn't work in Eleventy --watch mode, because
-          // the ES module cache never gets cleared. Switch to isolated VM
-          // modules so that we can control this.
-          await Promise.all(
-            (options.componentModules ?? []).map(
-              (module) => import(path.resolve(process.cwd(), module))
-            )
-          );
-          let head, body, tail;
-          const page = content.match(/(.*)(<body.*<\/body>)(.*)/);
-          if (page) {
-            [head, body, tail] = page;
-          } else {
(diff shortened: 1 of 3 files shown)


packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -4,8 +4,19 @@
...
+// calls.
+//
+// So instead we use TypeScript's ESM output mode, but explicitly write
+// require() calls for the CommonJS modules we import.
+//
+// See https://github.com/microsoft/TypeScript/issues/43329 for more details.
+
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const path = require('path');
+const path = require('path') as typeof import('path');
```
justinfagnani, line 19: Does `import path = require('path');` work without the cast?

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -20,35 +31,36 @@ module.exports = {
...
+        );
+
+        const rendered = iterableToString(render(html`${unsafeHTML(content)}`));
+        // Lit SSR includes special comment markers around templates to identify
+        // them during hydration. However, it's not possible for the top-level
+        // template that we've just rendered to ever be hydrated, because it was
+        // only defined ephemerally right now. So the comments serve no purpose
+        // in this case and can be removed. This also avoids the issue of
+        // comments appearing before <!doctype> declarations.
+        const outerMarkersTrimmed = rendered
```
justinfagnani, line 60: Rather than removing the markers after they're already emitted, could we add an option to SSR to not render the markers?

packages/labs/eleventy-plugin-lit/src/index.ts
aomarks: Nope, it returns `any`

aomarks:
```
Import assignment cannot be used when targeting ECMAScript modules.
```

And we're emitting ES modules instead of CommonJS because we need dynamic `import`. Due to https://github.com/microsoft/TypeScript/issues/43329 you can't emit dynamic `import()` when targeting Common JS.

justinfagnani: oh, right... https://github.com/microsoft/TypeScript/issues/22321

packages/labs/eleventy-plugin-lit/src/index.ts
aomarks: Started on this, but the issue I have hit is that there are two layers of markers:

1. For the outer template
2. For the `unsafeHTML` directive

So we can add an `outerMarkers:false` option to `render`, but that's only going to handle layer 1 and not 2. We don't want the option it to be generally recursive, since that would break all hydration. I suppose it could take a `depth` parameter, and only omit comments up to that depth.

Once we switch to vm modules, I think we won't have the `unsafeHTML` directive anymore, so this problem will go away.

justinfagnani: seems fine for now, but I do wonder if we're going to want ways to control marker output more generally if we ever have SSR-only rendering.

justinfagnani (approved):

packages/labs/eleventy-plugin-lit/src/index.ts
aomarks: Maybe a `markerStartDepth` parameter would work. `0` by default, `1` if you don't care about top-level, `2` for top-level + e.g. `unsafeHtml`, `-1` to skip all?

PR #2478, opened 2022-02-01
Hard code tunnel id. What happens?

AndrewJakubowicz:


AndrewJakubowicz:
The tunnel doesn't seem to work? It seems to still be the randomly generated nano id.

PR #2480, opened 2022-02-02
[@lit-labs/eleventy-plugin-lit] Use VM modules and fix path vs url bug

aomarks:
- Switches from global rendering to isolated VM modules.

- This allows Eleventy `--watch` mode to work. Fixes https://github.com/lit/lit/issues/2476

- Fixes a bug in `ModuleLoader` where we would fail to resolve lit-related modules because of the use of a ` file://` URL where a path was required. Refactored variable names to make this clearer.

- Switch to `fileURLToPath` and `pathToFileURL` instead of manual `file://` string manipulation in `ModuleLoader` to convert between paths and file URLs, because those functions handle escaping correctly.

- Refactored to use `eleventyConfig.on('eleventy.before')` to de-duplicate work that can be done per-build versus per-html-file.

### DIFF
--- a/packages/labs/eleventy-plugin-lit/src/index.ts
+++ b/packages/labs/eleventy-plugin-lit/src/index.ts
@@ -61,6 +61,9 @@ module.exports = {
           renderModulePath
         )
       ).module.namespace.render as typeof contextifiedRender;
+      // TOOD(aomarks) We could also directly synthesize an html TemplateResult
+      // instead of doing so via the the unsafeHTML directive. The directive is
+      // performing some extra validation that doesn't really apply to us.
       contextifiedUnsafeHTML = (
         await loader.importModule(
           'lit/directives/unsafe-html.js',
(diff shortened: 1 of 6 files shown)


packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -104,6 +104,53 @@ test.after.each(async ({rig}) => {
...
+ * the given timeout expires.
+ */
+const retryUntilTimeElapses = <T>(
+  timeout: number,
+  pollInterval: number,
+  fn: () => Promise<T>
+): Promise<T> => {
+  const start = performance.now();
+  let lastError: unknown;
+  return new Promise<T>((resolve, reject) => {
```
justinfagnani, line 118: Would an async function work? Might be more readable:

```ts
while (performance.now() - start < timeout) {
  try {
    return await fn();
  } catch (err) {
    lastError = err;
    await new Promise((r) => setTimeout(r, pollInterval));
  }
}
throw lastError ?? new Error('Timed out immediately'));
```

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -27,38 +27,60 @@ module.exports = {
...
+
+    const renderModulePath = path.join(process.cwd(), 'arbitrary.js');
+    const resolvedComponentModules = componentModules.map((module) =>
+      path.resolve(process.cwd(), module)
+    );
+
+    let contextifiedRender: (value: unknown) => IterableIterator<string>;
+    let contextifiedUnsafeHTML: (value: string) => unknown;
+
+    // Create a fresh context before each build, so that our module cache resets
```
justinfagnani, line 46: This is where you can use logic similar to what I pasted to reuse a ModuleLoader instance and clear just the changed files in the cache - if you can get the changed files from eleventy. It would be good to measure any perf impact of caching unchanged modules.

justinfagnani (approved):

packages/labs/eleventy-plugin-lit/src/index.ts
aomarks: Added a TODO.

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
aomarks: Good call, done.

augustjk (approved):

PR #2482, opened 2022-02-02
Better types for PropertyValues<T>

### LINKED ISSUE #2481: PropertyValues type does not work with keyof this
PropertyValues is defined as this:

```ts
export type PropertyValues<T = any> = keyof T extends PropertyKey
  ? Map<keyof T, unknown>
  : never;
```

but this causes errors because of `keyof T extends PropertyKey`, since the inferred type is `Map<(string|symbol|number) & keyof T>`. When a type is `keyof T`, it fails to match.

```ts
protected override willUpdate
(props: PropertyValues<this>) {
  const properties: Array<keyof this> = ['focused', 'populated'];
  for (const property of properties) {
    if (props.has(property)) {  // ERROR: Argument of type 'keyof this' is not assignable to parameter of type 'PropertyKey & keyof this'
      // ...
    }
  }
}
```

Originally reported by @asyncLiz

justinfagnani:
Fixes #2481 

This is technically a breaking change in that:

```ts
class E extends LitElement {
  declare foo: number;
  update(changedProperties: PropertyValues<this>) {
    changedProperties.get('bar'); // Error
  }
}
```

This should be an ok breaking change as it catches a legitimate problem. The thing to look out for here is whether this causes unexpected breaks. I don't quite understand what the conditional type was doing here, but since one branch was `never`, removing that should't cause any new errors.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -223,9 +223,18 @@ type AttributeMap = Map<string, PropertyKey>;
  * interface corresponding to the declared element properties.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type PropertyValues<T = any> = keyof T extends PropertyKey
-  ? Map<keyof T, unknown>
-  : never;
+export interface PropertyValues<T = any> extends Map<keyof T, unknown> {
+  get<K extends keyof T>(k: K): T[K];
+  forEach(
+    callbackfn: <K extends keyof T>(
+      value: T[K],
+      key: K,
+      map: Map<K, unknown>
+    ) => void,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    thisArg?: any
+  ): void;
+}
 
 export const defaultConverter: ComplexAttributeConverter = {
   toAttribute(value: unknown, type?: unknown): unknown {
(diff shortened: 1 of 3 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -223,9 +223,18 @@ type AttributeMap = Map<string, PropertyKey>;
  * interface corresponding to the declared element properties.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type PropertyValues<T = any> = keyof T extends PropertyKey
-  ? Map<keyof T, unknown>
-  : never;
+export interface PropertyValues<T = any> extends Map<keyof T, unknown> {
+  get<K extends keyof T>(k: K): T[K];
```
aomarks, line 227: How about `keys`, `values`, and `entries`? And I suppose `set` and `delete` would probably indicate something bad going on, but for completeness?

BTW here's a similar interface that I made in playground elements: https://github.com/google/playground-elements/blob/b896927494d63fe99850200c2b502d503b149e3e/src/playground-code-editor.ts#L17

justinfagnani: `has`, `delete`, and `keys` only use the key generic, so those don't ned any changes. I updated the value generic argument to `Map<keyof T, T[keyof T]>` so that `values`, are taken care of, and overrode `set()` to be generic itself.

aomarks: Cool! Would be nice to include test statements for `keys`, `values`, `entries`, `set`, and `delete` in the test file, too.

aomarks (approved):

packages/reactive-element/src/reactive-element.ts
justinfagnani: Added a few of those, especially for `set()`, but I think there are diminishing returns on re-checking that the value generic is doing the right thing everywhere it's used in the base interface.

aomarks (approved):

Issue #2248, opened 2021-10-14
[starter-kits] Automate starter kit releases


justinfagnani:
We need a script to copy the start kit files into the start kit repos and commit.

Part of #2178

PR #2496, opened 2022-02-04
Push the contents of the starter kits into their respective repos.

bicknellr:
This PR checks out the starter kits, pulls the contents of their individual directories in the main lit repo into the root of their individual repos, and commits / pushes the new content to `main`. This process is only gated on whether or not there has been any change in a starter kit's directory - the push does NOT depend on (e.g.) a version bump.

- [x] Merge #2516

Part of the work for https://github.com/lit/lit/issues/2248.


justinfagnani:
Could we test this by pushing to a different branch than `main`?

bicknellr:
These actions need some kind of condition to be triggered on. Currently, they'll run every time this workflow runs, which means during both the "make the version PR" and "publish packages" phases. I'm looking for a decent way to detect when its being run in the "publish packages" phase but, AFAICT, `changesets/action` only sets `published` to true if a non-zero number of packages were published - i.e. it doesn't make the "make the version PR" and "publish packages" phases distinguishable if zero packages were published. This might be good enough for our purposes but it would be nice not to have these pushes blocked on some other package having publishable changes.

bicknellr:
Maybe these shouldn't be tied to releases? They might fit better in a workflow that doesn't have any automatic triggers.

bicknellr:
Or, maybe they should be in a workflow that automatically triggers on every push to `main` since the starter kits aren't versioned.

.github/workflows/starter-kits.yaml
```diff
@@ -0,0 +1,82 @@
...
+          # Read the contents of the TS starter kit directory from the imported
+          # commit into the index root, ignoring any conflicts (`--reset`) and
+          # updating the work tree too (`-u`).
+          git read-tree --reset -u "$IMPORT_REF":packages/lit-starter-ts
+          # If there are no changes, skip the rest of this step.
+          if git diff --quiet --exit-code; then
+            exit 0
+          fi
+          # Stage the changes, commit, and push.
+          git add .
```
AndrewJakubowicz, line 51: Is this required?

bicknellr:
Thanks for taking the time to review!

PR #2508, opened 2022-02-07
The `published` output of the changesets action is the string 'false'.

### LINKED ISSUE #2514: Changesets GH action `published` output is a string 'false'
The actions output is the string `'false'`, so this evaluates as true:

https://github.com/lit/lit/blob/937388e247a26ee201f58be1c4fa536635b2fc5c/.github/workflows/release.yaml#L61

bicknellr:
I assumed this output was a boolean, but it seems like it's actually a string, which caused these steps to run more often than necessary.

https://github.com/lit/lit/runs/5097992845?check_suite_focus=true#step:8:21

Fixes #2514.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,7 +48,7 @@ jobs:
         uses: actions/checkout@v2
         # GitHub Actions outputs are strings and the string 'false' converts to
         # `true` because it isn't the empty string.
-        if: steps.changesets.outputs.published && steps.changesets.outputs.published != 'false'
+        if: steps.changesets.outputs.published == 'true'
         with:
           repository: lit/dist
           ref: empty
@@ -58,7 +58,7 @@ jobs:
       - name: Push bundles to dist repo
         # GitHub Actions outputs are strings and the string 'false' converts to
         # `true` because it isn't the empty string.
-        if: steps.changesets.outputs.published && steps.changesets.outputs.published != 'false'
+        if: steps.changesets.outputs.published == 'true'
         working-directory: dist
         run: |
           # Extract the version of `lit` that was published or the empty string.
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Checkout dist repo
         uses: actions/checkout@v2
-        if: steps.changesets.outputs.published
+        if: steps.changesets.outputs.published && steps.changesets.outputs.published != 'false'
         with:
           repository: lit/dist
           ref: empty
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -46,15 +46,15 @@ jobs:
 
       - name: Checkout dist repo
         uses: actions/checkout@v2
-        if: steps.changesets.outputs.published
+        if: steps.changesets.outputs.published && steps.changesets.outputs.published != 'false'
```
aomarks, line 49: Maybe link to an issue if you file it?

aomarks (approved):

.github/workflows/release.yaml
bicknellr: So, I thought this was an issue with their action but, after some more digging, I'm not so sure. AFAICT, action outputs are all supplied by writing to standard out in a special format, but the actual content of that output is just an arbitrary string.

Here's where the changesets action uses `@actions/core` to set an output:

https://github.com/changesets/action/blob/1ed9bc24b7a56462c183eb815c8f4bdf0e2e5785/src/index.ts#L91

... which calls:

https://github.com/actions/toolkit/blob/15e23998268e31520e3d93cbd106bd3228dea77f/packages/core/src/core.ts#L190-L193

... which calls:

https://github.com/actions/toolkit/blob/15e23998268e31520e3d93cbd106bd3228dea77f/packages/core/src/command.ts#L23-L30

This makes me think that the issue is really more that GitHub Actions i/o is all string based and not really related to the changeset action itself. Though, I did run across this comment which points out this [`fromJSON`](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson) function as a pretty clean way of handling it:

https://github.community/t/boolean-evaluation-workflow-yaml/173565/2

Though, I'm not sure how `fromJSON` reacts to things that don't parse as JSON and I haven't found anything in the docs about that yet. WDYT? Should I switch over to `fromJSON`?

aomarks: I think I'd just keep what you have here, which seems like the simplest solution, add a comment about why, and file an issue on changesets to point out the gotcha. Their docs do say "A boolean value to indicate whether a publishing is happened or not", so at the very least the README should be updated, since it's not a boolean: https://github.com/changesets/action

justinfagnani, line 49: Since this is a string, do we want to do `== 'true'` instead? I'd rather be defensive against values other than `'true'` and `'false'` and avoid the double negative of "not false".

justinfagnani (changes requested):

.github/workflows/release.yaml
bicknellr: Yeah, that makes more sense. Updated.

aomarks: Oops, I think this was my suggestion -- I thought that `published` was either the list of published packages, or `false`. But actually it probably is `"true"` or `"false"`, and the separate `publishedPackages` output is the list of packages. So just checking `"true"` sounds good to me too.

https://github.com/changesets/action#outputs

aomarks (approved):

justinfagnani (approved):

PR #2509, opened 2022-02-08
[meta] Fix invisible changeset step in release action by changing its id

aomarks:
For some surprising reason, changing the `id` of the changesets step in our release workflow from `"changesets"` to *anything else* (I picked `"cs"`) fixes the problem where the step is invisible from the list of steps in the GitHub UI.

See https://github.com/changesets/action/issues/149


AndrewJakubowicz (approved):
Wow! Great find!

augustjk (approved):
Great find! 🕵️

Issue #2393, opened 2022-01-07
[ssr] Lit SSR breaks Sass with polyfill


FredKSchott:
# Description

The required Lit SSR global window polyfill breaks Sass, and potentially other tools. cc @jaydanurwin who ran into this issue while trying to record a video on using Lit + Sass in Astro.

# Steps to Reproduce

1. Write this code in Node.js

```
await import('@lit-labs/ssr/lib/install-global-dom-shim.js');
// then, Sass fails to load
await import('sass');
```

2. See error: `"Cannot assign to read only property 'fs' of object '#<Object>'`

Relevant: https://github.com/mbullington/node_preamble.dart/issues/26

jaydanurwin:
Thanks for filing this @FredKSchott. It's not a complete project breaker but it would be nice to use Sass in Astro alongside Lit SSR components.

jonathantneal:
Noting the bug is probably with Sass. https://github.com/mbullington/node_preamble.dart/issues/26

aomarks:
https://github.com/lit/lit/issues/2391 is a similar issue where the presence of the window polyfill breaks the behavior of another library.

Maybe we could swap in and out the window polyfill only when are rendering (using AsyncLocalStorage?) as a general fix for this class of issue.

justinfagnani:
This shouldn't be an issue if we load all the code into a fresh VM context, like we do with the JS module loader. Since we want a version that can work without the `--experimental-vm-modules` flag, maybe there's a way to use plain dynamic `import()` in a classic script vm context so that we get a fresh global, just without the module loading hooks.

aomarks:
> This shouldn't be an issue if we load all the code into a fresh VM context, like we do with the JS module loader. Since we want a version that can work without the `--experimental-vm-modules` flag, maybe there's a way to use plain dynamic `import()` in a classic script vm context so that we get a fresh global, just without the module loading hooks.

Unfortunately this is not possible because dynamic `import()` in any isolated vm context always requires the `--experimental-vm-modules` flag. From the [docs](https://nodejs.org/api/vm.html#class-vmscript):

```
importModuleDynamically <Function> Called during evaluation of this module when import()
is called. If this option is not specified, calls to import() will reject with
`ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`. This option is part of the experimental modules API.
We do not recommend using it in a production environment.
```

[...]

PR #2510, opened 2022-02-08
[ssr] Replace window proxy in global DOM shim

augustjk:
Addresses part of the problem surfaced by #2393.

The proxy was causing unexpected behavior when attempting to assign to `window.global` after the shim is installed. Making `globalThis.window` a reference to `globalThis` should achieve the same thing the proxy was meant to handle.

### DIFF
--- a/.changeset/silent-carrots-sneeze.md
+++ b/.changeset/silent-carrots-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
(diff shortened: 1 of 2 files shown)


kevinpschaaf (approved):
LGTM

aomarks (approved):

aomarks:
Nice!

justinfagnani:
@kevinpschaaf what was the original reason for the Proxy? I kind of remember something about it vs `globalThis.window = window`...

PR #2515, opened 2022-02-09
Update `@custom-elements-manifest/analyzer` in the starter kits.

bicknellr:
I've just bumped the packages here. The `custom-elements.json` that's committed to the individual repos is out of date so I thought it might be due to this, but I think it just hasn't been regenerated in a long time. Either way, I think we settled on removing it from the individual repos.


AndrewJakubowicz (approved):
LGTM!

Is removing the manifest part of this change?

bicknellr:
> Is removing the manifest part of this change?

No, this is just updating the package. I think the extra `custom-elements.json` will be removed automatically during the next release after https://github.com/lit/lit/pull/2496 is merged (still WIP).

justinfagnani (approved):

PR #2516, opened 2022-02-09
[starter-kits] Upstream changes from the individual starter kit repos.

bicknellr:
These repos include a few extra typo fixes and their own issue templates that aren't in the main repo. After this, the starter kit packages in the main repo should be ready to be written over the individual repos, without losing anything unexpected. (FWIW, I'm expecting the next release following https://github.com/lit/lit/pull/2496 to delete the `custom-elements.json` in those repos.)


packages/lit-starter-js/.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
```
justinfagnani, line 1: are these going to be cloned with the repo since this is a template repo?

bicknellr: Yes, these should get copied over during the `git read-tree ...` steps in #2496.

bicknellr: I've been working on the basis that the starter kit packages in this repo should exactly match the individual repos. They're slightly different at the moment so this PR is upstream the things we want to keep since #2496 will start overwriting them.

AndrewJakubowicz: oh... the ISSUE_TEMPLATE config gets used when "Use this template" is pressed from the starter kit. This means anyone who uses our starter kit template ends up with issues pointed at the Lit monorepo.

There doesn't seem to be a way to omit files from the "use this template" option: https://github.community/t/can-you-ignore-files-folders-when-making-a-repo-from-a-template/3279

Maybe we should keep issues on the starter kits turned on and delete this config?

AndrewJakubowicz (approved):
LGTM!

packages/lit-starter-js/.github/ISSUE_TEMPLATE/config.yml
bicknellr: Oh! Thanks for the extra context, Andrew; I originally thought Justin was referring to the release workflow that would copy from here to the individual starter kit repos (vs. users cloning the starter kit repos).

I added the issue templates here because those repos currently contain these files in them and have issues enabled, so I just assumed we wanted them. If we don't want them, I can delete the issue templates here and they'll be removed from the repos during a release once #2496 has landed.

Should delete them / disable issues on those repos?

justinfagnani: Yeah, I don't think there should be any templates here. Thanks for clarifying @AndrewJakubowicz!

AndrewJakubowicz: Right now lets remove them, especially since I only added them a week ago: https://github.com/lit/lit-element-starter-ts/commit/9b52767bf6efda0d9432a0ad78c32fb1cb6d44ba

justinfagnani (changes requested):
will be good when the issue templates are removed

bicknellr:
I removed the issue templates.

Issue #2513, opened 2022-02-08
The new PropertyValues type definition causes breaking compile errors.


mattpowell:
# Description

A [change](https://github.com/lit/lit/pull/2482) in `@lit/reactive-element` at version `1.2.2` causes previously working TypeScript compilations to fail.

# Steps to Reproduce

**tl;dr**: Use typescript to compile the following code with `@lit/reactive-element` version `1.2.2`:
```typescript
import {LitElement} from 'lit';

export class Component extends LitElement {
  override updated(properties: Map<string, string>) {
    // noop
  }
}
```

For more detail see: https://gist.github.com/mattpowell/f772e99ddefaad6ebf1190d5bd60b112

## Expected Results

Compilation is successful and there are no type errors.

(Upgrading from 1.2.1 to 1.2.2 should be backwards compatible and previously allowed types/function signatures should still be valid)

## Actual Results

An error is thrown:
```
component.ts:4:12 - error TS2416: Property 'updated' in type 'Component' is not assignable to the same property in base type 'LitElement'.
  Type '(properties: Map<string, string>) => void' is not assignable to type '(_changedProperties: PropertyValues<any>) => void'.
    Types of parameters 'properties' and '_changedProperties' are incompatible.
      Type 'PropertyValues<any>' is not assignable to type 'Map<string, string>'.
        Types of property 'forEach' are incompatible.
          Type '(callbackfn: <K extends string | number | symbol>(value: any, key: K, map: Map<K, any>) => void, thisArg?: unknown) => void' is not assignable to type '(callbackfn: (value: string, key: string, map: Map<string, string>) => void, thisArg?: any) => void'.
            Types of parameters 'callbackfn' and 'callbackfn' are incompatible.
              Types of parameters 'key' and 'key' are incompatible.
                Type 'K' is not assignable to type 'string'.
                  Type 'string | number | symbol' is not assignable to type 'string'.
                    Type 'number' is not assignable to type 'string'.
```

Also, see [comment here](https://gist.github.com/mattpowell/f772e99ddefaad6ebf1190d5bd60b112?permalink_comment_id=4058032#gistcomment-4058032).

mattpowell:
This was linked in the PR that merged the type change, but linking it here for completeness: https://github.com/vmware/clarity/issues/6654

justinfagnani:
btw, for bug reports it'd be nice to see the code and error inline to the bug rather than behind a link. It makes it easier to glance and see what the bug is about.

[...]

PR #2518, opened 2022-02-09
Fix PropertyValues<any> to be compatible with Map<string, string> and subtyping

justinfagnani:
Fixes https://github.com/lit/lit/issues/2513

This still preserves the new behavior of strongly-typed Maps when the type parameter is specified, so there still could be some breaks for cases where people use `PropertyValues<this>` but then access keys that are not in the type. These _should_ be invalid uses that we want to catch, but we need to be on the lookout for unexpected breaks or ones that aren't easily fixable.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -218,33 +218,29 @@ type PropertyDeclarationMap = Map<PropertyKey, PropertyDeclaration>;
 
 type AttributeMap = Map<string, PropertyKey>;
 
-// This condition is here so that if the type parameter T is not specified, or
-// is `any`, the type will include `Map<PropertyKey, unknown>`. Since T is not
-// given in the uses of PropertyValues in this file, all uses here fallback to
-// meaning `Map<PropertyKey, unknown>`, but if a developer uses
-// `PropertyValues<this>` (or any other value for T) they will get a
-// strongly-typed Map type.
 /**
- * A Map of property keys to values. If the type parameter T is specified as
- * a non-any, non-unknown type, the Map will be a strongly-typed
- * PropertyValueMap that associates the map keys with their corresponding
- * property type on T.
+ * A Map of property keys to values.
+ *
+ * Takes an optional type parameter T, which when specified as a non-any,
+ * non-unknown type, will make the Map more strongly-typed, associating the map
+ * keys with their corresponding value type on T.
  *
  * Use `PropertyValues<this>` when overriding ReactiveElement.update() and
  * other lifecycle methods in order to get stronger type-checking on keys
  * and values.
  */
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz:
Discussed offline. Found another breaking case with subtyping. Shared a minimal repro with Justin.

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -3130,4 +3130,47 @@ suite('ReactiveElement', () => {
...
+  });
+
+  test('PropertyValues<T> works with subtyping', () => {
+    // This test only checks compile-type behavior. There are no runtime
+    // checks.
+    class A extends ReactiveElement {
+      foo!: number;
+      override update(changedProperties: PropertyValues<A>) {
+        const n: number = changedProperties.get('foo');
+        console.log(n);
```
AndrewJakubowicz, line 3162: Nit: console.log required?

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -3130,4 +3130,47 @@ suite('ReactiveElement', () => {
...
+      }
+    }
+    class B extends A {
+      bar!: string;
+      override update(changedProperties: PropertyValues<B>) {
+        const s: string = changedProperties.get('bar');
+        console.log(s);
```
AndrewJakubowicz, line 3169: Nit: console.log required?

AndrewJakubowicz (approved):
Excellent, great types! Also passed internal tests. Shared offline.

justinfagnani:
@AndrewJakubowicz @aomarks PTAL

packages/reactive-element/src/reactive-element.ts
```diff
@@ -219,21 +219,32 @@ type PropertyDeclarationMap = Map<PropertyKey, PropertyDeclaration>;
...
+ * A Map of property keys to values.
+ *
+ * Takes an optional type parameter T, which when specified as a non-any,
+ * non-unknown type, will make the Map more strongly-typed, associating the map
+ * keys with their corresponding value type on T.
+ *
+ * Use `PropertyValues<this>` when overriding ReactiveElement.update() and
+ * other lifecycle methods in order to get stronger type-checking on keys
+ * and values.
  */
```
aomarks, line 231: Is the jsdoc comment going to get associated with this type correctly, given that there's a `//` commen below it? I think if you can hover over it in vscode, it should be fine. Less certain that typedoc would also work.

aomarks (approved):

packages/reactive-element/src/reactive-element.ts
justinfagnani: yeah, it works fine.

Issue #2502, opened 2022-02-07
[meta] Changeset release script PRs fails CLA check


aomarks:
Since https://github.com/lit/lit/pull/2465, we use `lit-robot` for changesets. We had hoped that this would make those PRs pass the Google CLA check, because `lit-robot` is allowlisted as a known good robot. However, it still fails.

Example: https://github.com/lit/lit/pull/2500/checks

```
The following contributors were found for this pull request:

❌ https://github.com/lit/lit/commit/6d58996417b6173a193203b64d58d120bcd28b54 PR Opener: @lit-robot
❌ https://github.com/lit/lit/commit/6d58996417b6173a193203b64d58d120bcd28b54 Author: @github-actions[bot] <github-actions[bot]​@users.noreply.github.com>

(Only the first commit for a unique contributor is listed.)
```

AndrewJakubowicz:
Fixed in https://github.com/lit/lit/pull/2543

Fixed by https://github.com/lit/lit/pull/2540

PR #2519, opened 2022-02-09
Use Lit Robot's info as the global name and email for git operations during releases.

bicknellr:
This PR sets the global git email and name to those of our robot account so that it doesn't have to be set in each individual repo before committing.


aomarks (approved):

aomarks:
Cool! Maybe this will fix https://github.com/lit/lit/issues/2502

justinfagnani:
This is draft... is it intended for review?

bicknellr:
> Cool! Maybe this will fix #2502

Huh, maybe!

> This is draft... is it intended for review?

I didn't mark it as ready since it's based on #2496, which is still pending, but I certainly won't stop you if you want to look it over early

bicknellr:
Actually no... I'd rather see that other PR work without changing the existing git setup.

bicknellr:
Closing: this was added as part of #2540.

PR #2523, opened 2022-02-11
`working-directory` only works for steps with `run`. Use `with.cwd` instead for `changesets/action`.

bicknellr:
Apparently, `working-directory` is only applicable to jobs that use `run` and not those that specify an action with `uses`. Fortunately, changesets/action supports [a `cwd` input](https://github.com/changesets/action#inputs) for the same purpose.

Here's where `working-directory` is mentioned in the docs:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -45,6 +45,8 @@ jobs:
         id: cs
         uses: changesets/action@v1
         with:
+          # The `working-directory` option is only available for steps that use
+          # `run`. This `cwd` option is the same, but specific to this action.
           cwd: lit
           version: npm run version
           publish: npm run release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -38,14 +38,14 @@ jobs:
...
       - name: Create Release Pull Request or Publish to npm
-        working-directory: lit
         # The id of this step must not be "changesets", or else the step will be invisible
         # in the list of steps from the GitHub UI when the action runs (though it will still
         # run, and its output will appear in the raw logs). Unknown why this is the case,
         # see https://github.com/changesets/action/issues/149 for discussion.
         id: cs
         uses: changesets/action@v1
         with:
+          cwd: lit
```
AndrewJakubowicz, line 48: Is it worth adding a comment about `working-directory` only working with `run` and not `uses`?

AndrewJakubowicz (approved):

.github/workflows/release.yaml
bicknellr: sgtm, added.

AndrewJakubowicz (approved):

PR #2535, opened 2022-02-14
[starter-kits] Add a link to the starter templates so users will know where they should file issues.

bicknellr:
I'm going to transfer all of the issues in the starter kit repos to this one and then disable issues. These links should help people find the original repo so they know where to file issues if they need to.

Part of #2536.


AndrewJakubowicz (approved):
Awesome!

bicknellr:
Thanks for the review!

Issue #2391, opened 2022-01-07
[ssr] SSR rendering in NextJS


K4CZP3R:
# Description

I'm trying to render a lit component via SSR in NextJS, the default behavior is that it fails to import the component because there is no `window` property at the server-side rendering.

To fix that, I've imported `@lit-labs/ssr/lib/install-global-dom-shim.js` before importing my lit component. 

Unfortunately, some packages relay their SSR-detection on the availability of `window` property. 
And because of the installed shim,  packages will fail to detect the valid environment and will be executing client-side code on the server.

The shim installed by `@lit-labs/ssr` does not contain all of the functions of the real, client-sided, `window` property.

In my case, the package `styled-jsx` tries to execute the following line of code: 
`var node = this._isBrowser && document.querySelector('meta[property="csp-nonce"]');`.

This results in an exception because the `_isBrowser` property is true on the server-side.
Hardcoding this property results in compiling without any problems.

# Steps to Reproduce

1. Write this code

element.ts
```ts
import "@lit-labs/ssr/lib/install-global-dom-shim.js";
import {html, css, LitElement} from 'lit';
import {customElement, property} from 'lit/decorators.js';

@customElement('simple-greeting')
export class SimpleGreeting extends LitElement {
  static styles = css`p { color: blue }`;

  @property()
  name = 'Somebody';

  render() {
    return html`<p>Hello, ${this.name}!</p>`;
  }
}
```

index.js
```ts
import "element"

export default function Home(){
  return ( <simple-greeting></simple-greeting> )
}
```

2. `npm run build`

## Expected Results

Use of lit element in NextJS using SSR (which is enabled by default), without adding any new server-side specific functionality to the element itself.

## Actual Results

Exception is thrown when builiding:
```error - TypeError: document.querySelector is not a function```

**Comment by mohsenasfiamaersk** [comment]
We have exactly the same issue, I've made this codesandbox which replicates the issue (just for reference it's a modified edition of [this](https://codesandbox.io/s/itj0x?file=/pages/index.tsx)):
https://codesandbox.io/s/nextjs-lit-forked-cymhs?file=/components/Button.client.tsx

**Comment by mohsenasfiamaersk** [comment]
Apparently we can also add the shim by injecting it to the `pages/_document.js` like below, but that requires changing the webpack config to copy the shim to the bundle as well, which I haven't managed to do yet.
```
import Document, { Html, Head, Main, NextScript } from 'next/document';
import raw from 'raw.macro';
[...]

PR #2538, opened 2022-02-15
Resolves document.querySelector not a function when doing SSR with Next.js

mohsenasfia:
Addresses #2391 and resolves the issue when trying to render lit components with SSR in `NextJS` where the `@lit-labs/ssr/lib/install-global-dom-shim.js` is necessary to be imported, but the compilation results in the following error in `styled-jsx` package: 
`TypeError .. document.querySelector is not a function`
[Code sandbox repo which reproduces the issue](https://codesandbox.io/s/nextjs-lit-forked-cymhs?file=/components/Button.client.tsx)

### DIFF
--- a/.changeset/healthy-cars-grab.md
+++ b/.changeset/healthy-cars-grab.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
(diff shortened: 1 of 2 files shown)


augustjk:
Thank you for the contribution!

We're taking a stance to keep the DOM shim as minimal as possible. Rather than adding no-op methods to suppress errors, we believe it best for component code to be more intentional about what happens when it does not have access to DOM APIs.

Since this PR and the referenced issue has been opened, we've added node builds that can be safely imported (but not server-rendered) without the global DOM shim, and an `isServer` environment checker for alleviating errors due to browser code running in the server.

PR #2540, opened 2022-02-15
Work around a `changesets/action` bug with its `cwd` option.

### LINKED ISSUE #2539: Release broken: `changesets/action` has a bug with cwd
`changesets/action` calls into git before changing directories as specified with `cwd`, which means that it fails if the root directory is not a git repo:

https://github.com/changesets/action/blob/898d125cee6ba00c6a11b6cadca512752c6c910c/src/index.ts#L19-L34

bicknellr:
`changesets/action@v1` has a bug where it attempts to set the git user info locally before changing to the directory specified with `cwd`.[^1] This means that it will fail at this step if the root directory is not a git repo. To work around this, this PR initializes a new git repo in the root directory and sets Lit Robot's git user info globally.

Fixes #2539.

[^1]: https://github.com/changesets/action/blob/898d125cee6ba00c6a11b6cadca512752c6c910c/src/index.ts#L19-L34

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,10 +14,21 @@ jobs:
     environment: Changesets
     runs-on: ubuntu-latest
     steps:
+      - name: Use Lit Robot's name and email for all git actions
+        run: |
+          git config --global user.name "Lit Robot"
+          git config --global user.email "lit-robot@google.com"
+
       - name: Setup Node.js 16
         uses: actions/setup-node@v2
         with:
           node-version: 16
+
+      # changesets/action@v1 currently has a bug where it attempts to set the
+      # local git user info before changing directories for `cwd`, which means
+      # it will fail if the root directory is not a git repo.
+      - name: Initialize empty root repo
+        run: git init
 
       - name: Checkout Lit Repo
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -14,11 +14,22 @@ jobs:
...
+          git config --global user.email "lit-robot@google.com"
+
       - name: Setup Node.js 16
         uses: actions/setup-node@v2
         with:
           node-version: 16
 
+      # changesets/action@v1 currently has a bug where it attempts to set the
+      # local git user info before changing directories for `cwd`, which means
+      # it will fail if the root directory is not a git repo.
```
AndrewJakubowicz, line 29: May be worth adding a TODO to a Lit issue or changesets issue since this is a workaround comment. To track cleaning this up.

AndrewJakubowicz (approved):
And this can be tested by running the release action? Can we run the release action from a branch (I assume not)? Thank you!

.github/workflows/release.yaml
bicknellr: This is now #2542.

PR #2548, opened 2022-02-16
The bundle push step no longer runs in the `lit` repo, so this script needs to be called directly.

bicknellr:
This also removes the `extract-published-lit-version` from the repo's npm scripts and adds an explicit exit code to the `exit`.


augustjk (approved):
LGTM

PR #2550, opened 2022-02-17
Consistently use a 60s test timeout

rictic:
We set a 60s test timeout in packages/tests/web-test-runner.js but in other places we go with the default of 2000. This can lead to spurious failures like https://github.com/lit/lit/runs/5225253370


.changeset/chatty-foxes-remember.md
```diff
@@ -0,0 +1,11 @@
...
+'lit': patch
+'lit-element': patch
+'lit-html': patch
+'@lit/lit-starter-js': patch
+'@lit/lit-starter-ts': patch
+'@lit/localize': patch
+'@lit/reactive-element': patch
+---
+
+[internal] Consistently use a longer timeout when running tests.
```
AndrewJakubowicz, line 11: Does this timeout change require a public version bump?

rictic: I don't think it does, but the changesets github check fails if there isn't a changeset file. I suppose I should just let it fail though :)

AndrewJakubowicz (approved):
Yay for test improvements! Thank you!

Issue #2549, opened 2022-02-17
[labs/ssr] ModuleLoader linker fails with multiple concurrent entrypoints


aomarks:
Given two modules that both import `lit`, this approach works fine:

```ts
new ModuleLoader(...);
for (const mod of modules) {
  await loader.importModule(module, referrer);
}
```

But this approach consistently fails:

```ts
new ModuleLoader(...);
await Promise.all(modules.map((module) =>
  loader.importModule(module, referrer));
);
```

with an error like:

```
Error: request for '@lit/reactive-element' is not yet fulfilled
        at SourceTextModule.link (node:internal/vm/module:198:17)
        at ModuleLoader.importModule (/home/aomarks/work/wireit-lit/packages/labs/ssr/src/lib/module-loader.ts:107:7)
        at async Promise.all (index 1)
        at AsyncEventEmitter.<anonymous> (/home/aomarks/work/wireit-lit/packages/labs/eleventy-plugin-lit/src/index.ts:85:7)
        at async Promise.all (index 0)
        at async Eleventy.executeBuild (/home/aomarks/work/wireit-lit/packages/labs/eleventy-plugin-lit/node_modules/@11ty/eleventy/src/Eleventy.js:971:7)
```

Decent chance there is a timing issue in the `ModuleLoader` class https://github.com/lit/lit/blob/908f451d1444955571206588646efd04457d8d02/packages/labs/ssr/src/lib/module-loader.ts#L68

aomarks:
cc @justinfagnani I spent a while poking around at `ModuleLoader`, but couldn't identify the problem here. I'll dig into this more deeply probably tomorrow, but just pinging in case this seems familiar to you, or something comes to mind.

aomarks:
cc @justinfagnani @kevinpschaaf I spent a while poking around at `ModuleLoader`, but couldn't identify the problem here yet. I'll dig into this more deeply probably tomorrow, but just pinging in case this seems familiar to you, or something comes to mind.

PR #2551, opened 2022-02-17
[@lit-labs/eleventy-plugin-lit] Temporary fix for ModuleLoader concurrency issue

aomarks:
There is some underlying issue in our ModuleLoader class whereby
loading multiple modules concurrently can cause errors like:

```
"request for '@lit/reactive-element' is not yet fulfilled
```

There is likely some kind of race condition there, but I haven't
yet identified it. Tracking at https://github.com/lit/lit/issues/2549.

To address this problem for the Eleventy plugin in the short term,
this temporarily makes module loading serial instead of concurrent.

### DIFF
--- a/.changeset/few-books-exist.md
+++ b/.changeset/few-books-exist.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': patch
+---
+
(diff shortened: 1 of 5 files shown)


e111077 (approved):
quick turnaround!

AndrewJakubowicz (approved):

PR #2553, opened 2022-02-17
Echo steps throughout the bundle pushing process.

bicknellr:
This PR adds a few extra echos throughout the bundle push step to help clarify what's happening during this step in the logs.


augustjk (approved):
Nice!

PR #2564, opened 2022-02-17
Update changeset for lit release

augustjk:
Adding `lit` to changeset since core libraries are getting updated

### DIFF
--- a/.changeset/perfect-pens-itch.md
+++ b/.changeset/perfect-pens-itch.md
@@ -1,4 +1,5 @@
 ---
+'lit': minor
 'lit-element': minor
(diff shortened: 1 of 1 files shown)


aomarks (commented):
<strike>I think we should have a changeset file for it too, which is similar/identical to the change to the core libs, so that the changelog gets updated correctly.</strike>

aomarks:
> I think we should have a changeset file for it too, which is similar/identical to the change to the core libs, so that the changelog gets updated correctly.

Sorry, ignore me, I misread the diff.

aomarks (approved):

PR #2568, opened 2022-02-18
SSR with Nuxt cannot find module 'parse5’

mohsenasfia:
Resolves the issue happening in `Nuxt2` after building and running the app nuxt/framework#2559 nuxt/nuxt.js#12894:
`Cannot find module 'parse5’, Require stack: /node_modules/@lit-labs/ssr/lib/util/parse5-utils.js` 

Issue is reproduced in this repo, you can see it by running `npm run serve` and then by browsing the page: https://github.com/mohsenasfia/nuxt2


steveworkman:
+1 for this fix, simply swapping over to the native module import seems to work in Nuxt. Digging into the module resolution output it appears that even though it's searching through the right folders, it never finds `parse5`

```sh
MODULE 57469: load native module module
MODULE 57469: Module._load REQUEST parse5 parent: /node_modules/@lit-labs/ssr/lib/util/parse5-utils.js
MODULE 57469: looking for "parse5" in ["/node_modules/@lit-labs/ssr/lib/util/node_modules","/node_modules/@lit-labs/ssr/lib/node_modules","/node_modules/@lit-labs/ssr/node_modules","/node_modules/@lit-labs/node_modules","/node_modules","/Users/xxxxx/.node_modules","/Users/steve/.node_libraries","/Users/xxxxx/.nvm/versions/node/v16.15.0/lib/node"]
```

danielroe:
The Nuxt 3 issue in the stackblitz above is caused by the polyfill being stripped from the final build output, likely for the same cause as https://github.com/nuxt/framework/issues/2561.

steveworkman:
@mohsenasfia the underlying Nuxt bug has been fixed, so we should be able to get somewhere with Nuxt integration. I think we can drop this PR as it's not the solution to the problem, https://github.com/nuxt/framework/issues/2561 was!

Issue #2367, opened 2021-12-18
[labs/task] Allow setting initial data


mohe2015:
# Description

If you do navigation preloading or similar you may already have the data for the task at component creation time. Then it would make sense to be able to either set that data manually at Task creation time or using some attributes.

The simplest but probably also most unclean change would be to make `_value`

https://github.com/lit/lit/blob/57a53e1c85216a4255e9f700402ef6cbe37a4167/packages/labs/task/src/task.ts#L108

 public and name it `value`

# Steps to Reproduce

```ts
import {Task} from '@lit-labs/task';

this._apiTask = new Task(
  this,
  taskFunction,
  () => /** @type {[URLSearchParams]} */ ([this.history.url.searchParams])
);
 if (this.initial !== undefined) {
  this._apiTask.status = TaskStatus.COMPLETE;
  // TODO FIXME don't use this private attribute that got minified
  this._apiTask.P = this.initial;
}
```

sorvell:
We are planning to add a mechanism for setting an initial value. This will be independent of the task's status. In other words, the task can have a value and be in the initial status.

PR #2573, opened 2022-02-20
[@lit-labs/task] Allow providing initial state

mohe2015:
Closes https://github.com/lit/lit/issues/2367

cc @sorvell

Mostly to find out whether this way is fine.

Didn't run the tests because on NixOS it doesn't seem to easily work.

Also need to test this. Last time it didn't work to link this package but I will see.

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -44,6 +44,9 @@ export interface TaskConfig<T extends unknown[], R> {
   task: TaskFunction<T, R>;
   args?: ArgsFunction<T>;
   autoRun?: boolean;
+  initialStatus?: TaskStatus;
+  initialValue?: R;
+  initialError?: unknown;
 }
 
 // TODO(sorvell): Some issues:
@@ -107,7 +110,7 @@ export class Task<T extends [...unknown[]] = any, R = any> {
   private _host: ReactiveControllerHost;
   private _value?: R;
   private _error?: unknown;
-  status: TaskStatus = TaskStatus.INITIAL;
+  status: TaskStatus;
 
   /**
    * A Promise that resolve when the current task run is complete.
@@ -142,6 +145,9 @@ export class Task<T extends [...unknown[]] = any, R = any> {
       typeof task === 'object' ? task : ({task, args} as TaskConfig<T, R>);
     this._task = taskConfig.task;
     this._getArgs = taskConfig.args;
+    this.status = taskConfig.initialStatus ?? TaskStatus.INITIAL;
+    this._value = taskConfig.initialValue;
+    this._error = taskConfig.initialError;
     if (taskConfig.autoRun !== undefined) {
       this.autoRun = taskConfig.autoRun;
     }


mohe2015:
Seems to work in my project (I had to `ln -s` this because `npm link` didn't want to). Currently I'm not interested in writing a test though unfortunately.

mohe2015:
Any chance to get a review please? Or is there anything I can help with?

mohe2015:
Your answer is totally sufficient I just wanted some information on the state of this. Thank you.

> @mohe2015 we'd need tests to merge this, so if you don't want to write any we'll have to wait until one of us can take over the PR.

If I should find the time to write some I will notify here in advance so there is no duplicate work. I had issues in the past with the setup on my operating system though which further demotivates me.

> I also wonder about the design of allowing all the initial state to be set manually like this. Could the first run of the task function not just complete early and return the same value? We should take that to the issue for discussion though.

I also thought about this. I thought that it would need two rerenders instead of one and the following approach also exhibits that behaviour. If you have an idea on how to do this without a rerender I would love to know and that may make this PR uneccessary. Also I don't know how bad that rerender actually is but I would expect it would be possible it causes a flash of content although I couldn't see one yet.
```typescript
export const pwProjects = async (url: URL) => {
  const result = await taskFunction("/api/v1/projects", url, "projects");
  return html`<pw-projects .initial=${result} prefix="projects"></pw-projects>`;
};


export class PwEntityList<
  P extends keyof typeof entityRoutes,
  X extends string
> extends PwForm<P> {

  // ...

  override render() {
    // ...
    if (this.initialRender) {
      this._task = new Task(this, {
        task: async () => {
          if (this.initialRender) {
            this.initialRender = false;
            
            // this is not set yet because its passed using properties
            console.log(this.initial)
            if (this.initial) {
              return this.initial;
            }  
          }
  
          const data = parseRequestWithPrefix(
            this.url,
            this.prefix,
            this.history.url
          );
  
          const formDataEvent = new CustomEvent<
            z.infer<typeof entityRoutes[P]["request"]>
          >("myformdata", {
            bubbles: false,
            detail: data[this.prefix] ?? {},
          });
          this.form.value?.dispatchEvent(formDataEvent);
  
          const result = await myFetch<P>(
            "GET",
            this.url,
            formDataEvent.detail,
            {}
          );
  
          HistoryController.goto(
            new URL(
              `?${encodeURIComponent(
                JSON.stringify({
                  ...data,
                  [this.prefix]: formDataEvent.detail,
                })
              )}`,
              window.location.href
            ),
            {},
            true
          );
  
          return result;
        },
      });
  
      void this._task.run();
    }
    // ...
}
```
```
rerender pw-app 0.221305971394413
rerender pw-entitylist /api/v1/projects 0.09831745708584716 
rerender pw-app 0.09127497369239312
rerender pw-app 0.34908154859564
rerender pw-entitylist /api/v1/users 0.23022775624359948
Object { success: true, data: {…} }
rerender pw-entitylist /api/v1/users 0.5374328084204616
```

mohe2015:
I'm gonna close this because I think the alternative if acceptable.

Issue #2572, opened 2022-02-19
[ssr] undefined attributes throw for lit elements


daKmoR:
# Description

Trying to ssr an undefined value/attribute of a LitElement throws.

# Steps to Reproduce

1. Write this code

```ts
import { render } from "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { html, LitElement } from 'lit';

class MyEl extends LitElement {
  static get properties() {
    return {
      message: { type: String }
    }
  }
}

customElements.define('my-el', MyEl);

const template = ({ message }) => html`
  <p>${message}</p>
  <my-el message=${message}></my-el>
`;

const ssrResult = render(template({ message: undefined }));
//                                  👆 using `message: 'foo'` works as expected only undefined throws
let ssrStr = '';
for (const ssrPart of ssrResult) {
  ssrStr += await ssrPart;
}
console.log(ssrStr);
```

2. execute `node file.js`

## Expected Results

no error thrown

## Actual Results

```
file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/util/escape-html.js:19
export const escapeHtml = (str) => str.replace(/[&<>"']/g, (char) => replacements[char]);
                                       ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at escapeHtml (file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/util/escape-html.js:19:40)
    at LitElementRenderer.renderAttributes (file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/element-renderer.js:88:40)
    at renderAttributes.next (<anonymous>)
    at renderTemplateResult (file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:467:33)
    at renderTemplateResult.next (<anonymous>)
    at renderValue (file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:354:16)
    at renderValue.next (<anonymous>)
    at render (file:///Users/html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:336:12)
    at render.next (<anonymous>)
    at file:///Users/html/test-ssr/undefined.js:21:12
```

## Suggested changed

in here https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/element-renderer.ts#L146

```diff
    *renderAttributes() {
        if (this.element !== undefined) {
            const { attributes } = this.element;
[...]

PR #2580, opened 2022-02-23
fix(ssr): handle undefined/null element attribute values

daKmoR:
closes https://github.com/lit/lit/issues/2572

### DIFF
--- a/packages/labs/ssr/src/test/lib/render-lit_test.ts
+++ b/packages/labs/ssr/src/test/lib/render-lit_test.ts
@@ -218,6 +218,46 @@ test('element with property', async () => {
   );
 });
 
+test('element with attribute', async () => {
+  const {render, elementWithAttribute} = await setup();
+  const result = await render(elementWithAttribute('bar'));
+  // TODO: we'd like to remove the extra space in the start tag
+  assert.is(
+    result,
+    `<!--lit-part ZI1U/5CYP1o=--><test-property  foo="bar"><!--lit-node 0--><template shadowroot="open"><!--lit-part UNbWrd8S5FY=--><main><!--lit-part-->bar<!--/lit-part--></main><!--/lit-part--></template></test-property><!--/lit-part-->`
+  );
+});
+
+test('element with an empty string attribute', async () => {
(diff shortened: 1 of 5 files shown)


packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -218,6 +218,36 @@ test('element with property', async () => {
   );
 });
 
+test('element with attribute', async () => {
```
aomarks, line 221: Could you add a case for `'element with empty string attribute'` too?

aomarks (approved):
Looks great! Thanks for the PR. Just one small request to add a quick extra test case as well.

packages/labs/ssr/src/test/lib/render-lit_test.ts
daKmoR: added 👍

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -218,6 +218,46 @@ test('element with property', async () => {
...
```
justinfagnani, line 221: These tests should be added here: https://github.com/lit/lit/blob/main/packages/labs/ssr/src/test/integration/tests/basic.ts#L883

This way we get server rendering with an assertion, then client hydration that checks that there's no mutations.

justinfagnani (changes requested):

packages/labs/ssr/src/test/lib/render-lit_test.ts
daKmoR: for "normal" rendering it's already there [AttributePart accepts undefined](https://github.com/lit/lit/blob/main/packages/labs/ssr/src/test/integration/tests/basic.ts#L940) and that worked before already

it's about the rendering of attributes of an element... so I added it to `LitElement: Attribute binding`

aomarks (approved):

daKmoR:
this bug is quite annoying - anything still blocking this fix? 🤔

PR #2582, opened 2022-02-25
missing comma in example code

drewp:


drewp:
Actually I suggest writing the example with args arranged like this:
```
private _load = new Task(this, {
    args: () => [this.sus],       // "when these attributes change..."
    task: ([sus]) =>              // "then run this task, which takes the same params"
      fetch("users" + (sus ? "?sus=1" : ""))
        .then((resp) => resp.json())
        .then((j) => j.users),    // "since by the time we're down here, we're thinking about return vals, not params"
  });
```

aomarks (approved):
Thanks for the fix!

aomarks (changes requested):
Ah, would you be able to use the `npx changeset` command to indicate that this is a patch change with a brief description (e.g. "Fix typo in readme sample code"). Thanks!

aomarks:
> I can't really justify the spare time for this process. I already lost the checkout, so to propose this one-char fix would now take relearning the syntax for git branches (because I use hg), etc. Maybe there was no checkout, and I did it all on github?

No problem, I'll add the changeset. Thanks for the fix!

AndrewJakubowicz (approved):
Thank you! Added changeset. Re-requested review from @aomarks

aomarks (approved):

PR #2590, opened 2022-03-01
Remove saucelabs tests

taylor-vann:
- experimenting with removing sauce labs for our tests
- github actions have matured into full featured CI/CD
- AFAICT we currently create a new saucelabs proxy per package, we are rate-limited after a certain number of saucelab proxies are created

Solution:
- we are testing per browser per os
- move over to custom github actions
- then refine WTR and other testing capabilities to improve testing speed


taylor-vann:
duplicate PR

PR #2591, opened 2022-03-02
[@lit-labs/eleventy-plugin-llit] Add option to render using worker threads

### LINKED ISSUE #2484: [11ty] Use worker threads instead of vm modules

augustjk:
Resolves #2484

### Changes
- Added option `mode: 'vm' | 'worker'` (defaults to 'vm'). If mode is set as 'worker' uses a worker thread to render the component modules.
- Caveat of using worker threads is that the source files of the user modules need to be clear to Node that they're modules either with the extension `.mjs` or must be in a subdirectory with a `package.json` with `"type": "module"`. It can't be in the project root's `package.json` since eleventy needs to be commonjs.

### Todo
- [x] Update readme with documentation about this.
- [x] Discussion on whether this should be an option or just replace the 'vm' flow. There may be implications for how we do data flow for future roadmap items.
- [x] Add/modify tests


packages/labs/eleventy-plugin-lit/src/util/render-in-worker-thread.ts
```diff
@@ -0,0 +1,23 @@
...
+if (parentPort === null) {
+  throw new Error(
+    'render-in-worker-thread.js must only be run in a worker thread'
+  );
+}
+
+(async function () {
+  for (const module of workerData as string[]) {
+    await import(module);
+  }
```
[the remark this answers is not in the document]
aomarks: Related to this and the other comment about handling parallel requests, right now the protocol between the worker and host is kind of minimal/implicit:

- Worker data is imports
- Host -> worker message is render request
- Worker -> host message is render response

WDYT about creating a more explicit protocol? Maybe something like this?:

```js
type Message = InitializeRequest | InitializeResponse | RenderRequest | RenderResponse;

interface InitializeRequest {
  type: 'initialize-request';
  imports: string[];
}

...

interface RenderResponse {
  type: 'render-response';
  id: number;
  rendered: string;
}
```

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -14,23 +14,26 @@
...
 
 module.exports = {
   configFunction: function (
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     eleventyConfig: any,
-    {componentModules}: LitPluginOptions = {}
+    {componentModules, mode = 'vm'}: LitPluginOptions = {}
   ) {
-    if (require('vm').Module === undefined) {
+    if (mode === 'vm' && require('vm').Module === undefined) {
```
aomarks, line 36: Might be nice for readability to define two separate functions here, instead of `if` statements within one big function.

This particular check could be moved directly into that function. I think also the `renderModulePath` variable below is only used in the `vm` case too.

aomarks (changes requested):
Great start!

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -68,65 +71,122 @@ ${reset}`
...
+      eleventyConfig.addTransform(
+        'render-lit',
+        async (content: string, outputPath: string) => {
+          if (!outputPath.endsWith('.html')) {
+            return content;
+          }
+
+          const renderedContent: string = await new Promise((resolve) => {
+            worker.once('message', (data: string) => {
+              resolve(data);
```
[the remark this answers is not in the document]
augustjk: Good point, I haven't considered that. I'll see if I can find out more how eleventy handles that. The request id to resolve sounds like a good plan.

packages/labs/eleventy-plugin-lit/src/util/render-in-worker-thread.ts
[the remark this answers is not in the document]
augustjk: Yes, I think that's a great idea to enhance how we utilize the message plus it leaves it extensible. Thanks!

packages/labs/eleventy-plugin-lit/src/index.ts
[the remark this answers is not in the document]
augustjk: Some testing with logging and adding random timers in rendering does confirm that eleventy does build the pages in parallel and if they come back in the wrong order, the output is messed up. Great catch! I'll definitely work on utilizing unique request ids so we don't mix up results.

packages/labs/eleventy-plugin-lit/README.md
```diff
@@ -46,14 +46,15 @@ npm i @lit-labs/eleventy-plugin-lit
 
 ### Register plugin
 
-Edit your `.eleventy.js` config file to register the Lit plugin:
+Edit your `.eleventy.cjs` config file to register the Lit plugin:
```
aomarks, line 49: I would just keep this as `.js`, since that's the default extension most people will have when using eleventy.

packages/labs/eleventy-plugin-lit/README.md
```diff
@@ -72,28 +92,25 @@ module.exports = function (eleventyConfig) {
 Use the `componentModules` setting to tell the plugin where to find the
 definitions of your components.
 
-Pass an array of paths to `.js` files containing Lit component definitions.
+Pass an array of paths to `.js` or `.mjs` files containing Lit component definitions.
 Paths are interpreted relative to to the directory from which the `eleventy`
 command is executed.
 
-Each `.js` file should be a JavaScript module (ESM) that imports `lit` with a
+Each `.js` or `.mjs` file should be a JavaScript module (ESM) that imports `lit` with a
```
aomarks, line 99: Same here, drop `.mjs`

packages/labs/eleventy-plugin-lit/README.md
```diff
@@ -72,28 +92,25 @@ module.exports = function (eleventyConfig) {
...
-Node flag enabled. This flag is available in Node versions `12.16.0` and above.
-This flag allows the lit plugin to render components without modifying any
-global state, and enables compatibility with Eleventy's `--watch` mode.
-
-```sh
-NODE_OPTIONS=--experimental-vm-modules eleventy
-```
+1. Add `{"type": "module"}` to your base `package.json` and make sure the
+   eleventy config file end with the `.cjs` extension.
+1. Make all component files end with the `.mjs` extensions.
```
aomarks, line 111: I think I would remove the `.mjs` option; the other two options are better.

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -4,41 +4,124 @@
...
+function configureVm(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  eleventyConfig: any,
+  resolvedComponentModules: string[]
+) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  if (require('vm').Module === undefined) {
+    // Show a more friendly error message if the --experimental-vm-modules
+    // flag is missing.
+    // TODO(aomarks) Rendering in a worker would remove the need for this flag.
```
aomarks, line 119: Can remove this comment.

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -50,83 +133,117 @@ module.exports = {
...
-      const {ModuleLoader} = await import('@lit-labs/ssr/lib/module-loader.js');
-      const window = getWindow({includeJSBuiltIns: true});
-      const loader = new ModuleLoader({global: window});
-      // TODO(aomarks) Replace with concurrent Promise.all version once
-      // https://github.com/lit/lit/issues/2549 has been addressed.
-      for (const module of resolvedComponentModules) {
-        await loader.importModule(module, renderModulePath);
+    switch (mode) {
+      case 'worker': {
+        return configureWorker(eleventyConfig, resolvedComponentModules);
```
aomarks, line 236: I think I would remove the `return`s here and below, since nothing is returned.

packages/labs/eleventy-plugin-lit/src/index.ts
```diff
@@ -4,41 +4,124 @@
...
+    worker.on('error', (err) => {
       console.error(
-        `${yellow}
+        'Unexpected error while rendering lit component in worker thread',
+        err
+      );
+      throw err;
+    });
+
+    let requestResolve: Function;
```
aomarks, line 39: Narrower type:

```suggestion
    let requestResolve: () => void;
```

packages/labs/eleventy-plugin-lit/src/util/render-in-worker-thread.ts
```diff
@@ -0,0 +1,44 @@
...
+
+if (parentPort === null) {
+  throw new Error(
+    'render-in-worker-thread.js must only be run in a worker thread'
+  );
+}
+
+parentPort.on('message', async (message: Message) => {
+  switch (message.type) {
+    case 'initialize-request': {
```
aomarks, line 21: Add a guard that only one `initialize-request` can ever be sent for the lifetime of a worker?

aomarks (changes requested):
Fantastic! Just some relatively minor comments.

packages/labs/eleventy-plugin-lit/README.md
```diff
@@ -72,28 +92,25 @@ module.exports = function (eleventyConfig) {
 Use the `componentModules` setting to tell the plugin where to find the
 definitions of your components.
 
-Pass an array of paths to `.js` files containing Lit component definitions.
+Pass an array of paths to `.js` or `.mjs` files containing Lit component definitions.
```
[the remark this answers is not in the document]
augustjk: That makes sense, and is a great point about this being primarily authored for the web.

packages/labs/eleventy-plugin-lit/src/index.ts
augustjk: I was debating between doing a `return` here versus adding a `break;` after each of these. Now I agree that semantically `return` makes less sense as nothing's returned.

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -171,370 +178,345 @@ test('without plugin', async ({rig}) => {
...
-      module.exports = function (eleventyConfig) {
-        eleventyConfig.addPlugin(litPlugin, {
-          componentModules: ['./js/my-element.js'],
-        });
-      };
-    `,
+const modes = ['worker', 'vm'] as const;
 
-    // js
+const getMyElementDefinitionAndConfigByMode = (mode: typeof modes[number]) => {
```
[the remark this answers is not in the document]
augustjk: Right! Without needing a parameter, it can just be an object rather than a function.

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
```diff
@@ -171,370 +178,340 @@ test('without plugin', async ({rig}) => {
...
+    (mode === 'vm' ? 'NODE_OPTIONS=--experimental-vm-modules ' : '') +
+    'eleventy --config .eleventy.cjs';
 
-    // initial component definition
-    'js/my-element.js': `
-      import { html, LitElement } from 'lit';
-      class MyElement extends LitElement {
-        render() {
-          return html\`INITIAL\`;
+  test('basic component in markdown file', async ({rig}) => {
```
aomarks, line 212: Sorry I didn't think of this before -- Could you add a new test to confirm that we can process multiple pages simultaneously, which would cover the `id` tracking feature? Maybe the pages could be generated so that you can easily make a large number of them (e.g. 25 pages, each with an incrementing number in it).

(Sorry if there is one, but I don't think I see it).

aomarks (changes requested):

packages/labs/eleventy-plugin-lit/src/test/eleventy-plugin-test.ts
augustjk: Oh, yes. Good call out.

aomarks (approved):
Awesome!

Issue #2544, opened 2022-02-15
[meta] Changeset action fails on "Version Packages" PR


AndrewJakubowicz:
During release, a Version Packages PR is created.

The changeset action fails with error:

```
> changeset
> changeset "status" "--since=main"

🦋  error Some packages have been changed but no changesets were found. Run `changeset add` to resolve this error.
🦋  error If this change doesn't need a release, run `changeset add --empty`.
```

See previous Version Package pull requests as reference:

 - https://github.com/lit/lit/runs/5207461948?check_suite_focus=true
    - From https://github.com/lit/lit/pull/2543

### Why?

Speculatively, the "Version Packages" PR deletes all the changeset markdown files, and bumps some version numbers. Thus when `changeset` checks against the `main` branch, it finds changes without a changeset markdown file associated.

kevinpschaaf:
Should we file an issue on changesets? Seems like we're using their stuff in the way they intend?

PR #2596, opened 2022-03-03
[meta] Fix changeset action failing on release PR

AndrewJakubowicz:
Fixes https://github.com/lit/lit/issues/2544

### Context

This check was added in https://github.com/lit/lit/pull/2061 and ensures that changes to packages require a changeset file. See docs on the command: https://github.com/changesets/changesets/blob/main/packages/cli/README.md#status

However in the case of an automated release PR, we create a situation which fails this check. This is because the release PR deletes all changeset files, and also modifies packages by bumping version numbers and updating the CHANGELOG.md files (reference PR: https://github.com/lit/lit/pull/2565/files).

Thus this check will (and does) always fail in a release PR.

### Fix

Skip this check if `lit-robot` authored the PR.

### Alternative considered

We could create an empty changeset file as part of the automated release PR. I opted for this route as I think it makes sense for the automated release PR to not have a changeset file, but can be convinced otherwise.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
       - run: git checkout main
       - run: git checkout ${{ github.sha }}
       - run: npm ci
-      # Enforce that all PR's that change packages need changesets. Changes
+      # Enforce that all PRs that change packages need changesets. Changes
       # without changesets result in this job failing.
       - run: npm run changeset status -- --since=main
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -21,6 +24,8 @@ jobs:
       - run: git checkout main
       - run: git checkout ${{ github.sha }}
       - run: npm ci
+      # Enforce that all PR's that change packages need changesets. Changes
```
aomarks, line 27:
```suggestion
      # Enforce that all PRs that change packages need changesets. Changes
```

aomarks (approved):
nice!

justinfagnani (approved):
nice!

PR #2597, opened 2022-03-04
Fetch all history for the starter kit repos during checkout in the sync workflow.

bicknellr:
The starter kit repos were not pushed after merging #2496 because the checked out repos are shallow: https://github.com/lit/lit/runs/5415481001?check_suite_focus=true#step:5:28

### DIFF
--- a/.github/workflows/starter-kits.yaml
+++ b/.github/workflows/starter-kits.yaml
@@ -30,6 +30,7 @@ jobs:
           ref: release-automation-test-main
           path: lit-element-starter-ts
           token: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
+          depth: 0
 
       - name: Push changes to lit-element-starter-ts repo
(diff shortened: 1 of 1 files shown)


.github/workflows/starter-kits.yaml
```diff
@@ -30,26 +30,33 @@ jobs:
           ref: release-automation-test-main
           path: lit-element-starter-ts
           token: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
+          # Fetch all history, otherwise we can't push to this repo.
+          depth: 0
```
bicknellr, line 34: This should be `fetch-depth`. *facepalm*

bicknellr: Fixed.

AndrewJakubowicz (approved):
Nice! Also like the logging as well!

PR #2598, opened 2022-03-04
Set `fetch-depth: 0` on all repos when syncing the starter kits.

bicknellr:
The source repo needs to not be shallow for the fetch to succeed without [`--update-shallow`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---update-shallow) ([`.git/shallow` docs here](https://git-scm.com/docs/shallow)). I think this likely means that #2597 wasn't necessary, but I'm not inclined to revert it since I'm worried it would affect the push.

### DIFF
--- a/.github/workflows/starter-kits.yaml
+++ b/.github/workflows/starter-kits.yaml
@@ -22,6 +22,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: lit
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
Change makes sense! 🤞

PR #2600, opened 2022-03-04
[@lit/reactive-element] Fix a lint warning in the tests.

bicknellr:
This unused variable is intentional because the assignment's type error is used to check the type of the expression, but the lint warning appears at the end of every PR's diff because it isn't handled.


AndrewJakubowicz (approved):
LGTM - and since test only change agree that it doesn't need a changeset.

kevinpschaaf (approved):
LGTM

bicknellr:
I should've refreshed - thanks for the reviews!

PR #2635, opened 2022-03-14
Don't clone nodes as part of the debug event logger

### LINKED ISSUE #2632: Child element constructor is called twice
# Description

When you render a child Lit element, the constructor is called twice.

This has been causing some problems: particularly with the `web-dialog` component that is throwing exceptions when it's rendered inside of a Lit element. It's logging a console error on the second element creation.

Minimal repro: https://github.com/RandomEngy/lit-double-ctor-repro

I write a console message each time the child element class instance is constructed and you see two entries for `Sub element created.`

Using lit `2.2.0`.

# Steps to Reproduce

1. Render a child Lit element as part of a parent lit element.
2. Observe the number of times the child element constructor is called.

## Expected Results

Constructor is called once

## Actual Results

Constructor is called twice

## Browsers Affected

- [x] Chrome
- [ ] Firefox
- [x] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

(Possibly more, only tested in Edge and Chrome)

rictic:
Integrations that need to clone nodes can do so themselves

Fixes #2632

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -189,16 +189,19 @@ interface DebugLoggingWindow {
  * Compiled out of prod mode builds.
  */
 const debugLogEvent = DEV_MODE
-  ? (event: LitUnstable.DebugLog.Entry) => {
+  ? () => {
       const shouldEmit = (window as unknown as DebugLoggingWindow)
         .emitLitDebugLogEvents;
       if (shouldEmit) {
+        return undefined;
+      }
+      return (event: LitUnstable.DebugLog.Entry) => {
         window.dispatchEvent(
           new CustomEvent<LitUnstable.DebugLog.Entry>('lit-debug', {
             detail: event,
           })
         );
-      }
+      };
     }
   : undefined;
 // Used for connecting beginRender and endRender events when there are nested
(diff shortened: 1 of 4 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -189,16 +189,19 @@ interface DebugLoggingWindow {
  * Compiled out of prod mode builds.
  */
 const debugLogEvent = DEV_MODE
-  ? (event: LitUnstable.DebugLog.Entry) => {
+  ? () => {
       const shouldEmit = (window as unknown as DebugLoggingWindow)
         .emitLitDebugLogEvents;
       if (shouldEmit) {
+        return undefined;
```
justinfagnani, line 196: Should the condition here be inverted?

packages/lit-html/src/lit-html.ts
```diff
@@ -189,16 +189,19 @@ interface DebugLoggingWindow {
...
+      }
+      return (event: LitUnstable.DebugLog.Entry) => {
```
justinfagnani, line 198: Do you think it's that much better to be lazier here rather than removing the clone() calls? Why does the element need to be cloned in the first place? I think that could still fail in some cases when the event are enabled.

justinfagnani (approved):

PR #2642, opened 2022-03-15
Add an additional security brand check to StaticValues

rictic:
Similar to #2307

### DIFF
--- a/.changeset/seven-bottles-sing.md
+++ b/.changeset/seven-bottles-sing.md
@@ -1,5 +1,6 @@
 ---
 'lit-html': patch
+'lit': patch
 ---
 
-Add an additional security brand check to StaticValues; Similar to #2307
+Add an additional security brand check to StaticValues; Similar to [#2307](https://github.com/lit/lit/pull/2307)
--- a/.changeset/seven-bottles-sing.md
+++ b/.changeset/seven-bottles-sing.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Add an additional security brand check to StaticValues; Similar to #2307
--- a/packages/lit-html/src/static.ts
+++ b/packages/lit-html/src/static.ts
@@ -78,7 +78,7 @@ interface StaticValue {
 
   /**
    * A value that can't be decoded from ordinary JSON, make it harder for
-   * a attacker-controlled data that goes through JSON.parse to produce a valid
+   * attacker-controlled data that goes through JSON.parse to produce a valid
    * StaticValue.
    */
   r: RegExp;
--- a/packages/lit-html/src/static.ts
+++ b/packages/lit-html/src/static.ts
@@ -7,7 +7,7 @@
 // Any new exports need to be added to the export statement in
 // `packages/lit/src/index.all.ts`.
 
-import { html as coreHtml, svg as coreSvg, TemplateResult } from './lit-html.js';
+import {html as coreHtml, svg as coreSvg, TemplateResult} from './lit-html.js';
 
 interface StaticValue {
   /** The value to interpolate as-is into the template. */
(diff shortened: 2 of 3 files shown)


.changeset/seven-bottles-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
```
AndrewJakubowicz, line 2: Nit: Add `lit` patch as well.

packages/lit-html/src/static.ts
```diff
@@ -55,14 +56,33 @@ const textFromStatic = (value: StaticValue) => {
...
+  return (value as Partial<StaticValue>)?.['_$litStatic$'];
+};
+
+interface StaticValue {
+  /** The value to interpolate as-is into the template. */
+  _$litStatic$: string;
+
+  /**
+   * A value that can't be decoded from ordinary JSON, make it harder for
+   * a attacker-controlled data that goes through JSON.parse to produce a valid
```
AndrewJakubowicz, line 81:
```suggestion
   * attacker-controlled data that goes through JSON.parse to produce a valid
```

.changeset/seven-bottles-sing.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add an additional security brand check to StaticValues; Similar to #2307
```
AndrewJakubowicz, line 5: Worth making #2307 a markdown link?

AndrewJakubowicz (approved):
LGTM, but should also get a LGTM from someone else with domain knowledge here.

packages/lit-html/src/static.ts
```diff
@@ -23,8 +23,9 @@ import {html as coreHtml, svg as coreSvg, TemplateResult} from './lit-html.js';
  * Static values can be changed, but they will cause a complete re-render
  * since they effectively create a new template.
  */
-export const unsafeStatic = (value: string) => ({
+export const unsafeStatic = (value: string): StaticValue => ({
   ['_$litStatic$']: value,
+  r: /_/,
```
justinfagnani, line 28: It's clever that this is short, but using a RegExp is maybe a little obscure? Also `typeof x === {aLiteral}` is hghly optimized in JS VMs.

What about:

```ts
const staticBrand = Symbol();
export const unsafeStatic = (value: string): StaticValue => ({
  ['_$litStatic$']: value,
  r: staticBrand,
});
```

Then:
```ts
if (!(typeof (value as Partial<StaticValue>)?.r !== 'symbol')) {
```

rictic: IE11 doesn't have native Symbol, right, so `typeof Symbol() === 'symbol'` can't work, can it? https://caniuse.com/mdn-javascript_builtins_symbol

justinfagnani: We already require the Symbol polyfill.

`Symbol.for()` might be good like you mentioned before too, then you can do `(value as Partial<StaticValue>)?.r === staticBrand`

jridgewell: Since these are each brand new instances of RegExp, and not a shared value, I'm assuming you're just trying to guard against JSON injections. In that case, you can take advantage of `constructor: undefined` to get the same benefits without needing an object (or a `instanceof` check, which is very slow). JSON cannot inject an `undefined` value, and `constructor` on all instances, so `value.constructor === undefined` guarantees the `value` was not constructed from JSON.

rictic: That's a really cool trick @jridgewell! Ended up going with Symbol.for('') for clarity and to avoid hitting VM edge cases.

.changeset/seven-bottles-sing.md
rictic: +1, done

packages/lit-html/src/static.ts
rictic: Oh, erp, didn't push all my changes, should be there now

packages/lit-html/src/static.ts
```diff
@@ -9,6 +9,43 @@
...
+  r: typeof brand;
+}
+
+/**
+ * Prevents JSON injection attacks.
+ *
+ * The goals of this brand:
+ *   1) fast to check
+ *   2) code is small on the wire
+ *   3) multiple versions of Lit in a single page will all produce mutually
```
AndrewJakubowicz, line 30: How does this work? Is it possible to add this case as a test?

justinfagnani: It's difficult to add tests for multiple versions. It works before `Symbol.for(s)` uses a global cache keyed on `s`, so `Symbol.for('x') === Symbol.for('x')`

justinfagnani (approved):
🔒 💂

packages/lit-html/src/static.ts
AndrewJakubowicz: Oh! TIL!

AndrewJakubowicz (approved):
Awesome! Great comment on the `brand`.

PR #1701, opened 2021-03-26
[lit-next] Labs CLI w/ localize commands

justinfagnani:

PR #2647, opened 2022-03-19
Add @lit-labs/cli package

justinfagnani:
Supercedes: #1701

WIP: There's still some cleanup to do in package setup, readme, etc.

Also adds a localize command


AliMD:
After `polymer-cli`, I am so excited to see @lit/cli
Could you please take some notes about your plan and have we can help? 
Is there any discussion/notes about this?

For me `polymer-cli` was more than simple tools, It was best practices to build, watch, deploy and test. Please follow this goal not just simple tools for dummies!

packages/labs/cli/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "typescript": "~4.6.2",
+    "uvu": "^0.5.3"
+  },
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "files": [
+    "/bin",
+    "/lib",
+    "/npm-shrinkwrap.json"
```
aomarks, line 42: Shrinkwrap? What year is it?

packages/labs/cli/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "stdout-stderr": "^0.1.13",
+    "typescript": "~4.6.2",
+    "uvu": "^0.5.3"
+  },
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "files": [
+    "/bin",
+    "/lib",
```
aomarks, line 41: I slightly prefer seeing relative paths here. Makes it more obvious it's relative to the package, rather than absolute filesystem root. (Even though npm re-roots it).

```suggestion
    "bin",
    "lib",
```

rictic (changes requested):

packages/labs/cli/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "/npm-shrinkwrap.json"
+  ],
+  "homepage": "https://github.com/lit/lit",
+  "keywords": [
+    "lit"
+  ],
+  "license": "BSD-3-Clause",
```
aomarks, line 48: Very nit, but it seems really common to have license, homepage, types etc. at the top, and dependencies at the bottom.

packages/labs/cli/src/bin/lit.ts
```diff
@@ -0,0 +1,29 @@
+#!node
```
aomarks, line 1: The shebang should be `#!/usr/bin/env node` right? Because PATH is not consulted when evaluating shebangs.

packages/labs/cli/src/lib/command.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export type CommandOptions = {
+  [name: string]: unknown;
+};
+
+/**
+ * A CLI command definition.
+ *
+ * This object only contains the metadata about a command and its command line
+ * options, not the actual command implementation. The module it's defined in
```
aomarks, line 21:
```suggestion
 * options, not the actual command implementation. The module its defined in
```

packages/labs/cli/src/lib/commands/help.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import chalk from 'chalk';
+import commandLineUsage from 'command-line-usage';
+
+import {globalOptions} from '../options.js';
+import {Command, CommandOptions} from '../command.js';
+import {LitCli} from '../lit-cli.js';
+
+const CLI_TITLE = chalk.bold.underline('Lit CLI');
+// const CLI_DESCRIPTION = '...';
```
aomarks, line 15: Remove?

packages/labs/cli/src/lib/commands/help.ts
```diff
@@ -0,0 +1,112 @@
...
+const CLI_USAGE = 'Usage: `lit <command> [options ...]`';
```
aomarks, line 16: Maybe we don't want the backticks inside of this string? I feel like that's not so common.

packages/labs/cli/src/lib/commands/localize.ts
```diff
@@ -0,0 +1,47 @@
...
+        },
+      ],
+      async run({config}: {config: string}, console: Console) {
+        const build = await import('../localize/build.js');
+        await build.run(config, console);
+      },
+    },
+  ],
+  async run() {
+    throw new Error('must use extract or build');
```
aomarks, line 45: Error could be a little more friendly, like `Please use "lit localize extract" or "lit localize build"`

aomarks (changes requested):

packages/labs/cli/package.json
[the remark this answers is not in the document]
justinfagnani: There's not a great place to document it, so I added a `prepublishOnly` step that calls `npm shrinkwrap`

packages/labs/cli/src/lib/command.ts
rictic: `it's` is correct here

packages/labs/cli/src/bin/lit.ts
justinfagnani: I think you're right, but this works for me. Changed.

packages/labs/cli/src/lib/command.ts
aomarks: Oh yeah, lol sorry.

packages/labs/cli/src/bin/lit.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitCli} from '../lib/lit-cli.js';
+
+process.on('uncaughtException', (error: null | undefined | Partial<Error>) => {
```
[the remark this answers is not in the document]
justinfagnani: I guess? I'll remove

packages/labs/cli/package.json
justinfagnani: ha ha. shrinkwraps are publishable to npm and use when installing at a top level. Their only use-case now is basically CLIs: https://docs.npmjs.com/cli/v7/configuring-npm/npm-shrinkwrap-json

packages/labs/cli/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "files": [
+    "/bin",
+    "/lib",
+    "/npm-shrinkwrap.json"
+  ],
+  "homepage": "https://github.com/lit/lit",
```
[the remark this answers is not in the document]
justinfagnani: I didn't see a way to make an actual empty exports that validated, so I just exported index.js so at least people can't import other modules.

packages/labs/cli/src/lib/command.ts
```diff
@@ -0,0 +1,56 @@
...
+ * the actual implementation.
+ */
+export interface Command {
+  /**
+   * The main name of the command.
+   */
+  name: string;
+
+  /**
+   * Any aliases, such as shortnads, for the command.
```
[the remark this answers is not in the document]
justinfagnani: 😳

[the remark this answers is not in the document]
justinfagnani: We might have `gen` and `generate`? We can remove now and add later.

packages/labs/cli/src/lib/command.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  aliases?: string[];
+
+  description: string;
+
+  options?: OptionDefinition[];
```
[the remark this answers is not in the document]
justinfagnani: Right now the CLI is set up for eager metadata and lazy execution. I think we can make the metadata lazy too once we add a plugin system. Right now this is simple.

packages/labs/cli/src/lib/command.ts
```diff
@@ -0,0 +1,56 @@
...
+   * Documentation to append onto the output of `lit help commandName`.
+   */
+  getUsageSections?(): Section[];
+}
+
+/**
+ * A command may return a CommandResult to indicate an exit code.
+ */
+export interface CommandResult {
+  exitCode: number;
```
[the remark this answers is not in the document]
justinfagnani: It's for allowing a command to fail in a specific way. We did this in the Polymer CLI, but I can't remember exactly when we used it.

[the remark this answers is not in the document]
justinfagnani: Removed.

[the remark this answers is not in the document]
justinfagnani: actually, added it back for the localize command to use.

packages/labs/cli/package.json
```diff
@@ -3,17 +3,20 @@
...
   "version": "0.0.0",
   "author": "Google LLC",
+  "license": "BSD-3-Clause",
   "bin": {
     "lit": "./bin/lit.js"
   },
   "bugs": "https://github.com/lit/lit/issues",
   "type": "module",
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
aomarks, line 13: Do we want types here given that it's not importable?

packages/labs/cli/package.json
```diff
@@ -3,17 +3,20 @@
...
   "repository": "lit/lit",
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc --watch",
-    "test": "uvu test '.*_test\\.js$'"
+    "test": "uvu test '.*_test\\.js$'",
+    "prepublishOnly": "npm run shrinkwrap"
```
aomarks, line 19: Should be prepack I think?

aomarks (changes requested):

packages/labs/cli/package.json
justinfagnani: Done

packages/labs/cli/package.json
justinfagnani: Looks like it. Thanks.

aomarks (approved):
:tada:

rictic (commented):
Why rename the .d.ts files to .ts files?

packages/labs/cli/src/lib/lit-cli.ts
```diff
@@ -0,0 +1,151 @@
...
+      return {
+        commandName: [...parentCommandNames, commandName].join(' '),
+        command,
+        argv: parsedArgs.argv,
+      };
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      // We need a valid command name to do anything. If the given
+      // command is invalid, run the generalized help command.
+      if (error.name === 'INVALID_COMMAND') {
```
rictic, line 142: prefer an instanceof here, and for error to be unknown rather than any

rictic (changes requested):

justinfagnani:
@rictic 

> Why rename the .d.ts files to .ts files?

They weren't importable from another package otherwise.

packages/labs/cli/src/lib/lit-cli.ts
justinfagnani: I'm not sure there is a class to check... I'll look.

justinfagnani: The class is just `Error`, so there's nothing good to instanceof

justinfagnani: Oh... yeah meant that the catch() argument might not be an Error. Done.

rictic (approved):

PR #2648, opened 2022-03-19
[@labs/react] type event callbacks

taylor-vann:
This PR is a retcon of PR #2445 

- users explicitly define event types in event callbacks
- default event type expected from a PropName map is `(e: Event) => void`
- user can declare a event map interface like `{propName: EventType, ...}`
- will be expressed as `{prop: (e: EventType) => void, ...}` in the React Component Props
- user prop type is used despite existing or expected React Type
- example: `{'onChange': Event}` will eclipse `React.FormEvent<HTMLInputElement>` with a DOM `Event` 

This should allow clients to explicitly define event types in react event callbacks like the example below:
```Typescript
  interface MyElementEvents {
    onClick: MouseEvent;  
    onKeyDown: KeyboardEvent;
  }

  const myElementPropNames = {
    onClick: 'click', // (e: MouseEvent) => void
    onKeyDown: 'keydown',  // (e: KeyboardEvent) => void
  }; 
  
  const myComponent = createComponent<MyComponent, MyElementEvents>(
    myElementPropNames,
    ...,
  )
```

Without generics, the typing defaults to `Event`:
```Typescript
  const myElementPropNames = {
    onClick: 'click', // (e: Event) => void
    onKeyDown: 'keydown',  // (e: Event) => void
  }; 
  
  const myComponent = createComponent(
    myElementPropNames,
    ...,
  )  
```

Undefined, 'missing', or types that don't extend from `Event` also default to `Event` in the EventMap:
```Typescript
  interface MyElementEvents {
    onClick: MouseEvent; 
    onNotAnEventType: any; 
  }

  const myElementPropNames = {
    onClick: 'click', // (e: MouseEvent) => void
    onKeyDown: 'keydown',  // defaults to (e: Event) => void
    onNotAnEventType: 'not-event-type',  // defaults to (e: Event) => void
  }; 
  
  const myComponent = createComponent<MyComponent, MyElementEvents>(
    myElementPropNames,
    ...,
  )
```

### DIFF
--- a/packages/labs/react/README.md
+++ b/packages/labs/react/README.md
@@ -57,14 +57,14 @@ React component.
 
 #### Typescript
 
-Event callback types can be refined by type casting with `EventHandler`. The
+Event callback types can be refined by type casting with `EventName`. The
 type cast helps `createComponent` correlate typed callbacks to property names in
 the event property map.
 
-Uncasted EventHandlers will fallback to `(e: Event) => void`.
+Non-casted event names will fallback to an event type of `Event`.
 
 ```ts
-import type {EventHandler} from '@lit-labs/react';
+import type {EventName} from '@lit-labs/react';
 
 import * as React from 'react';
 import {createComponent} from '@lit-labs/react';
@@ -75,7 +75,7 @@ export const MyElementComponent = createComponent(
   'my-element',
   MyElement,
   {
-    onClick: 'pointerdown' as EventHandler<PointerEvent>,
+    onClick: 'pointerdown' as EventName<PointerEvent>,
     onChange: 'input',
   }
 );
@@ -96,9 +96,9 @@ Event callbacks will match their type cast. In the example below, a
 ```
 
 NOTE: This type casting is not associated to any component property. Be
-careful to use the corresponding type dispatched or bubbling from the
-webcomponent. Incorrect type casts might result in events with missing
-properties.
+careful to use the corresponding type dispatched or bubbled from the
+webcomponent. Incorrect types might result in additional properties, missing
+properties, or properties of the wrong type.
 
 ## `useController`
 
--- a/packages/labs/react/README.md
+++ b/packages/labs/react/README.md
@@ -57,9 +57,9 @@ React component.
 
 #### Typescript
 
-EventHandlers can be typecasted using the `EventHandler` type included in
+Event callbacks can be typecasted using the `EventHandler` type included in
 this package. The cast helps `createComponent` correlate callback types to
-property names from an event prop name map.
+property names from an event property map.
 
 Uncasted EventHandlers will fallback to `(e: Event) => void`.
 
@@ -75,7 +75,7 @@ export const MyElementComponent = createComponent(
   'my-element',
   MyElement,
   {
-    onClick: 'pointerdown' as EventHandler<PointerEvent>
+    onClick: 'pointerdown' as EventHandler<PointerEvent>,
     onChange: 'input',
   }
 );
@@ -86,11 +86,15 @@ EventHandlers in a component's props will match their typecast. A
 
 ```tsx
 <MyElementComponent
-  onClick={(e: PointerEvent) => { console.log('DOM PointerEvent called!')}}
-  onChange={(e: Event) => { console.log(e)}}
+  onClick={(e: PointerEvent) => {
+    console.log('DOM PointerEvent called!');
+  }}
+  onChange={(e: Event) => {
+    console.log(e);
+  }}
 />
 ```
- 
+
 NOTE: This type cast is not associated to any actual component property. Be
 careful to use the corresponding type dispatched from the webcomponent.
 Incorrect type casts might result in events without expected properties.
--- a/packages/labs/react/README.md
+++ b/packages/labs/react/README.md
@@ -57,8 +57,9 @@ React component.
 
 #### Typescript
 
-Event callbacks can be typed using the included `EventHandler` type. By using a
-typecast, `createComponent` correlates callback types to property names from an event property map.
+Event callback types can be refined by type casting with `EventHandler`. The
+type cast helps `createComponent` correlate typed callbacks to property names in
+the event property map.
 
 Uncasted EventHandlers will fallback to `(e: Event) => void`.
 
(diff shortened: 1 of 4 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -84,16 +84,34 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
-  [P in keyof S]?: (e: Event) => unknown;
-};
-
 type StringValued<T> = {
   [P in keyof T]: string;
 };
 
 type Constructor<T> = {new (): T};
 
+type EventString<T = unknown> = string & {
```
justinfagnani, line 93: Specify the type parameter constraint and default to `Event`

```suggestion
type EventString<T extends Event = Event> = string & {
```

packages/labs/react/src/create-component.ts
```diff
@@ -84,16 +84,34 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
+  __event_type: T;
+};
+
+/***
+ * Typecast that curries an Event type through an EventString. The React
+ * equivelant to an event listeners is an `EventHandler`. The goal of the type
+ * name is to hint that corresponding EventHandler callback in their props will
+ * be of a certain type.
+ */
+export type EventHandler<T = unknown> = T extends Event
```
justinfagnani, line 103: I don't think you need this, especially if you constrain the type parameter in `EventString`.

packages/labs/react/README.md
```diff
@@ -55,6 +55,49 @@ React component.
 />
 ```
 
+#### Typescript
+
+Event callbacks can be typed using the included `EventHandler` type. By using a
```
justinfagnani, line 60: I would name the public type `EventName`. `EventHandler` sounds like it's the type of the actual event handler function.

packages/labs/react/README.md
```diff
@@ -55,6 +55,49 @@ React component.
...
+typecast, `createComponent` correlates callback types to property names from an event property map.
+
+Uncasted EventHandlers will fallback to `(e: Event) => void`.
```
justinfagnani, line 63: The fallback is that the event type is `Event`, right?

```suggestion
Non-casted event names will fallback to an event type of `Event`.
```

justinfagnani (changes requested):
Looks really good @taylor-vann !

Just some naming suggestions and a typing simplification.

packages/labs/react/src/create-component.ts
taylor-vann: Done :)

packages/labs/react/src/create-component.ts
taylor-vann: Done :)

packages/labs/react/README.md
taylor-vann: Done :)

packages/labs/react/README.md
taylor-vann: Done :)

packages/labs/react/src/create-component.ts
```diff
@@ -84,16 +84,34 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
+  ? EventString<T>
+  : never;
+
+type EventHandlerRecord = Record<string, EventString | string>;
```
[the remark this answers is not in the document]
taylor-vann: Done :)

packages/labs/react/src/create-component.ts
```diff
@@ -84,16 +84,34 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
+
+type EventHandlerMap<R extends EventHandlerRecord> = {
```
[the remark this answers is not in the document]
taylor-vann: Done :)

packages/labs/react/README.md
```diff
@@ -55,6 +55,49 @@ React component.
...
+  }}
+  onChange={(e: Event) => {
+    console.log(e);
+  }}
+/>
+```
+
+NOTE: This type casting is not associated to any component property. Be
+careful to use the corresponding type dispatched from the webcomponent.
+Incorrect type casts might result in events with missing properties.
```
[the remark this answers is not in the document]
taylor-vann: Done :)

taylor-vann (commented):
Changes addressed :)

packages/labs/react/src/create-component.ts
```diff
@@ -84,16 +84,32 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
+  __event_type: T;
+};
+
+/***
+ * Typecast that curries an Event type through an EventString. The React
+ * equivelant to an event listeners is an `EventName`. The goal of the type
+ * name is to hint that corresponding EventName callback in their props will
+ * be of a certain type.
+ */
+export type EventName<T> = T extends Event ? EventString<T> : never;
```
justinfagnani, line 103: I don't think you don't need this type at all. Once you have the type constraint of `T extends Event` TypeScript will fail for any type parameter that doesn't extend `Event` so you could never take the else clause here. I tried this locally and it works fine.

Remove this type, rename `EventString` to `EventName`, and export that.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
taylor-vann: That makes sense :) very cool!

taylor-vann (commented):
addressed changes :)

justinfagnani (approved):
Awesome!

taylor-vann:
Thanks @justinfagnani @WickyNilliams @sorvell for helping me get this across the finish line

WickyNilliams:
Really nice how small this change was in the end! Thanks for this, very much appreciated

WickyNilliams:
No rush, but i'm curious what the plan/schedule is for releasing this? Thanks again!

taylor-vann:
https://github.com/lit/lit/pull/2700

@WickyNilliams sorry for the delay but this change has made it to the release in the version mentioned above :)

WickyNilliams:
Awesome thanks!

PR #2661, opened 2022-03-23
Give a clearer error message when rendering into null/undefined

rictic:


AndrewJakubowicz (approved):

packages/lit-html/src/lit-html.ts
```diff
@@ -637,6 +637,13 @@ export const render = (
   container: HTMLElement | DocumentFragment,
   options?: RenderOptions
 ): RootPart => {
+  if (DEV_MODE && container == null) {
```
justinfagnani, line 640: This feels like runtime type-checking here, which we don't do in most other places. What's an example of a problematic call? Would a check that the container is a HTMLElement or DocumentFragment be better for that case?

rictic: The case that came up was:

```typescript
render(html`...`, document.body);
```

justinfagnani: So because `bundle.js` is a sync script and it contains some top-level `render(html`...`, document.body)` call and `document.body` doesn't exist yet? The `<some-lit-element>` has nothing to do with it right?

Just wondering here because there are probably a lot of places in our API you could incorrectly pass `null`. This seems ok though.

justinfagnani (approved):

packages/lit-html/src/lit-html.ts
rodrigob: Yes `<some-lit-element>` was there to justify the inclusion of lit in the `bundle.js`. The correct code should be

```
<html> 
<head></head>
<body>
  <some-lit-element></some-lit-element>
  <script src="./bundle.js"></script>
</body>
</html>
```

PR #2662, opened 2022-03-23
[@lit-labs/ssr] Implement HTMLElement.shadowRoot

### LINKED ISSUE #2652: [ssr] Consider supporting `HTMLElement.shadowRoot` in SSR dom-shim.
# Description
I noticed today that the default dom-shim in `@lit-labs/ssr` does not implement the `HTMLElement.shadowRoot` property. Specifically, invoking `attachShadow()` with an open shadow-root configuration does not cause the created shadow root for the element to be exposed from the `shadowRoot` property; [it only returns the shadow root object.](https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/dom-shim.ts#L63)

I know that you're trying to keep the dom-shim lean, but I was wondering if you would consider a PR that implements the [shadow root getter behavior](https://dom.spec.whatwg.org/#dom-element-attachshadow).

For context, I'm working on FAST's SSR implementation. We're hoping to align to the same dom-shim constraints as Lit, however, FAST only caches shadow root references for elements defined with a closed shadow root. In open shadow root cases, the shadow root will simply be retrieved from the element's `shadowRoot` property anytime the shadow root needs to be manipulated or inspected. This is currently causing a core code path to throw.

I'm sure we can adjust our runtime or dom-shim code to accommodate this gap, but I thought I would file an issue with ya'll first to see if this is a feature you're interested in taking. I'm happy to put up a PR if it is.

nicholasrice:
This PR implements a public readonly property, `shadowRoot`, on the `HTMLElement` class defined in the dom-shim. This property will evaluate to the object created and returned by `attachShadow()` when `attachShadow()` is invoked with an open shadow root mode (`host.attachShadow({mode: "open"}`). Otherwise, `shadowRoot` will evaluate to `null`. This behavior is intended to match the DOM behavior defined [here](https://dom.spec.whatwg.org/#dom-element-attachshadow).

Let me know if changes are need; I'm happy to make any updates. I've also added one question inline that should probably get attention before merging. Thanks in advance!

Fixes #2652

### DIFF
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -46,6 +46,10 @@ export const getWindow = ({
         value,
       }));
     }
+    #shadowRoot: null | ShadowRoot = null;
+    get shadowRoot() {
+      return this.#shadowRoot;
+    }
     abstract attributeChangedCallback?(
       name: string,
       old: string | null,
@@ -60,8 +64,12 @@ export const getWindow = ({
     hasAttribute(name: string) {
       return attributesForElement(this).has(name);
     }
-    attachShadow() {
-      return {host: this};
+    attachShadow(init?: ShadowRootInit) {
+      const shadowRoot = {host: this};
+      if (init && init.mode === 'open') {
+        this.#shadowRoot = shadowRoot;
+      }
+      return shadowRoot;
     }
     getAttribute(name: string) {
       const value = attributesForElement(this).get(name);
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -64,7 +64,7 @@ export const getWindow = ({
     hasAttribute(name: string) {
       return attributesForElement(this).has(name);
     }
-    attachShadow(init?: ShadowRootInit) {
+    attachShadow(init: ShadowRootInit) {
       const shadowRoot = {host: this};
       if (init && init.mode === 'open') {
         this._shadowRoot = shadowRoot;
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -46,9 +46,9 @@ export const getWindow = ({
         value,
       }));
     }
-    #shadowRoot: null | ShadowRoot = null;
+    private _shadowRoot: null | ShadowRoot = null;
     get shadowRoot() {
-      return this.#shadowRoot;
+      return this._shadowRoot;
     }
     abstract attributeChangedCallback?(
       name: string,
@@ -67,7 +67,7 @@ export const getWindow = ({
     attachShadow(init?: ShadowRootInit) {
       const shadowRoot = {host: this};
       if (init && init.mode === 'open') {
-        this.#shadowRoot = shadowRoot;
+        this._shadowRoot = shadowRoot;
       }
       return shadowRoot;
     }
--- a/packages/labs/ssr/src/test/lib/dom-shim_test.ts
+++ b/packages/labs/ssr/src/test/lib/dom-shim_test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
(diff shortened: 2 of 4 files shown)


packages/labs/ssr/src/test/lib/dom-shim_test.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
```
aomarks, line 3:
```suggestion
 * Copyright 2022 Google LLC
```

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -60,8 +64,12 @@ export const getWindow = ({
     hasAttribute(name: string) {
       return attributesForElement(this).has(name);
     }
-    attachShadow() {
-      return {host: this};
+    attachShadow(init?: ShadowRootInit) {
```
nicholasrice, line 67: I've added this as an optional param because I believe this would technically be a breaking change if it were required, even though the platform requires the init argument. I'm happy to update this to align with the platform if that is preferred.

packages/labs/ssr/src/test/lib/dom-shim_test.ts
```diff
@@ -0,0 +1,32 @@
...
+test('elements without an attached shadow root should expose a null value from the "shadowRoot" property', () => {
+  class UnattachedShadowRoot extends HTMLElement {}
+  const element = new UnattachedShadowRoot();
+  assert.is(element.shadowRoot, null);
+});
+test('elements defined with an open shadow root should expose it\'s shadow root from the "shadowRoot" property', () => {
+  class OpenShadowRoot extends HTMLElement {}
+  const element = new OpenShadowRoot();
+  const shadow = element.attachShadow({mode: 'open'});
+  assert.is(shadow, element.shadowRoot);
```
aomarks, line 23: May as well also check this while we're at it:

```suggestion
  assert.is(shadow, element.shadowRoot);
  assert.is(shadow.host, element);
```

packages/labs/ssr/src/lib/dom-shim.ts
aomarks: I think we should probably just make it required, to match the platform. It seems like it would be unusual to be call this method _only_ in an SSR context -- it will usually be called by a library that is already web compatible. So I think we can make it required without issuing a breaking change.

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -46,6 +46,10 @@ export const getWindow = ({
         value,
       }));
     }
+    #shadowRoot: null | ShadowRoot = null;
```
aomarks, line 49: I'm inclined to use `private _shadowRoot` here, because AFAIK we haven't been using native class fields anywhere in our code base so far. And since we're targeting ES2020 still, the output is a little gnarly (https://www.typescriptlang.org/play?target=7#code/MYGwhgzhAEBmD29oG8CwAoa0DEAjMATtALzQB2AriCANwYZYLwAUAlCg1tAQKYAuFAmWh8AFgEsIAOjyE6maAF8MioA).

@justinfagnani WDYT about starting to use native class fields in our tools?

aomarks (changes requested):
Awesome, thanks for the PR!

aomarks:
> ###  No Changeset found

Please run `npx changeset` in the top-level of the git repo, and mark `@lit-labs/ssr` as having a minor bump, with a brief description.

> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Please read and sign the Google CLA (if you agree!), see https://github.com/lit/lit/pull/2662/checks?check_run_id=5668663752 for details.

packages/labs/ssr/src/lib/dom-shim.ts
nicholasrice: Done. I'll leave this thread open so it's easier to find for other reviewers, in case they have thoughts.

packages/labs/ssr/src/lib/dom-shim.ts
nicholasrice: Sure thing, renamed.

aomarks (approved):

aomarks:
Thanks for the PR!

PR #2676, opened 2022-03-30
[@lit-labs/analyzer] Add analyzer package and start of Analyzer class

justinfagnani:


packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -0,0 +1,104 @@
...
+    throw new Error('LitElement not found');
+  };
+
+  getLitElementModule = () => {
+    const files = this.program.getSourceFiles();
+    for (const file of files) {
+      if (
+        file.fileName.endsWith('/node_modules/lit-element/lit-element.d.ts')
+      ) {
+        // hopefully there's only one of these...
```
rictic, line 81: In many cases there will be multiple. I think we should have a set of LitElement class declarations.

Hm, or in order to be lazier, maybe we should instead look at each of the super classes' base types and spot check it to see if it's a LitElement, rather than proactively tracking down the declaration(s) in the expectation that later we'll want to know it

justinfagnani: Yeah, that may be nice with a cache of types that are already subclasses of LitElement

packages/labs/analyzer/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/lit/lit.git",
+    "directory": "packages/labs/analyzer"
+  },
+  "main": "index.js",
+  "type": "module",
+  "scripts": {
+    "build": "tsc --build",
+    "install-test-deps": "cd test-files/basic-elements && npm ci",
```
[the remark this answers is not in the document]
rictic: it may be enough to express "before running tests, if any of the package.jsons in test-files have changed, then install test deps"

rictic (approved):

packages/labs/analyzer/src/test/analyzer_test.ts
```diff
@@ -0,0 +1,38 @@
...
+import ts from 'typescript';
+import * as path from 'path';
+
+import {Analyzer} from '../lib/analyzer.js';
+
+test('Basic elements - isLitElement', () => {
+  const basicElementsPackagePath = new URL(
+    '../test-files/basic-elements',
+    import.meta.url
+  ).pathname;
```
aomarks, line 18: Won't this always be a `/` delimited path? So it won't work on Windows. I think you want:

```ts
import * as pathlib from 'path';
import {fileURLToPath} from 'url';

const __filename = fileURLToPath(import.meta.url);
const __dirname = pathlib.dirname(__filename);
const basicElementsPackagePath = pathlib.resolve(__dirname, '..', 'test-files', 'basic-elements');
```

packages/labs/analyzer/test-files/basic-elements/src/element-a.ts
```diff
@@ -0,0 +1,18 @@
+import {LitElement, html, css} from 'lit';
```
aomarks, line 1: Even test files should probably have copyright statement?

packages/labs/analyzer/src/test/analyzer_test.ts
aomarks, line 18: Speaking of Windows support, I think it would be a good idea to start testing on Windows right away, since it's easier to handle Windows edge cases incrementally instead of in a big pass afterwards. It's pretty to set up Windows CI with GitHub Actions, see https://github.com/lit/wireit/blob/97b7a9f3932597a8c91d3a4d406fcdd7251d234d/.github/workflows/tests.yml#L9

(Another PR fine, I just recommend doing it sooner rather than later)

packages/labs/analyzer/package.json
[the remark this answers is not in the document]
aomarks: Maybe a path `devDependency` dependency would work? Not sure if it matters that that could cause some dependencies to be installed/deduped into `analyzer/node_modules` instead of `basic-elements/node_modules`?

packages/labs/analyzer/src/test/analyzer_test.ts
aomarks: I guess this would work too, a bit simpler:

```ts
import {fileURLToPath} from 'url';

fileURLToPath(new URL(
  '../test-files/basic-elements',
  import.meta.url
).pathname));
```

packages/labs/analyzer/tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compilerOptions": {
+    "incremental": true,
+    "composite": true,
+    "tsBuildInfoFile": "./lib/.tsbuildinfo",
+    "target": "es2020",
```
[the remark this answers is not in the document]
rictic: Ah yeah, we might also want to use real private fields in the analyzer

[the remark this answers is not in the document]
justinfagnani: Yeah, I think Node 14 is reasonable for now. It'll go out of maintenance 2023-04-23.

Class fields, including private, are es2022 though. tsc's downlevel of them isn't great either, IMO, so I'd rather not.

aomarks (approved):

packages/labs/analyzer/package.json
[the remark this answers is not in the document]
justinfagnani: that did work. It did also hoist `@lit/...` into the analyzer's node_modules, but I think that's ok.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -81,10 +81,7 @@ export class Analyzer {
   };
 
   private _isLitElementModule = (file: ts.SourceFile) => {
-    return (
-      file.fileName.startsWith(this.packageRoot) &&
-      file.fileName.endsWith('/node_modules/lit-element/lit-element.d.ts')
-    );
+    return file.fileName.endsWith('/node_modules/lit-element/lit-element.d.ts');
```
rictic, line 84: just curious, did something come up that prompted this change?

justinfagnani: Yeah: `lit-element` got hoisted out of the test-project when I made the test project a devDep of analyzer. We probably want to support hoisted installs for monorepos and things like pnpm anyway.

Issue #2593, opened 2022-03-03
[@lit-labs/react]: when React ref is passed to a wrapped component attribute `l="[Object object]` ends up in DOM


WickyNilliams:
# Description

When a React ref is passed a React-wrapper component, then an attribute like `l="[Object object]` ends up in the DOM.

# Steps to Reproduce

1. Write this code

```ts
import { useRef } from "react";
import { MyReactInput } from "./web-component";

export default function App() {
  const ref = useRef();

  return (
      <MyReactInput ref={ref} />
  );
}

```

2. See this output...

## Live Reproduction Link

Used Codesandbox as its what I am familiar with:

https://codesandbox.io/s/goofy-lederberg-bjiwkb?file=/src/App.tsx:794-857

## Expected Results

No unexpected attributes added to DOM.

## Actual Results

```html
<my-input l="[Object object]"></my-input>
```

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 14

Didn't test the rest:

- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

taylor-vann:
closing this issue, the forwardedRef will no longer be attached to the webcomponent itself

WickyNilliams:
Just to confirm, I'm no longer seeing this issue with the latest version. Thanks!

taylor-vann:
Fantastic! Thanks for confirming this for us @WickyNilliams

PR #2678, opened 2022-03-30
[@lit-labs/react] - remove renegade properties

taylor-vann:
Currently our react wrapper will attach the '__forwardedRef' react component property to the referenced element itself.

This PR:
- creates a property sieve (i'm assuming there might be other rogue properties in the future)
- skips over properties in the property sieve, doesn't pass sieve props to element
- add tests to confirm sieve properties aren't added

solves issue:
#2593

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -14,10 +14,6 @@ const reservedReactProperties = new Set([
   'className',
 ]);
 
-const propertySieve = new Set([
-  "__forwardedRef",
-]);
-
 const listenedEvents: WeakMap<
   Element,
   Map<string, EventListenerObject>
@@ -259,7 +255,7 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
       // iterate again when setting properties.
       this._elementProps = {};
       for (const [k, v] of Object.entries(this.props)) {
-        if (propertySieve.has(k)) continue;
+        if (k === '__forwardedRef') continue;
 
         if (elementClassProps.has(k)) {
           this._elementProps[k] = v;
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -14,6 +14,10 @@ const reservedReactProperties = new Set([
   'className',
 ]);
 
+const propertySieve = new Set([
+  "__forwardedRef",
+]);
+
 const listenedEvents: WeakMap<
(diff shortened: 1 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -255,6 +259,8 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
       // iterate again when setting properties.
       this._elementProps = {};
       for (const [k, v] of Object.entries(this.props)) {
+        if (propertySieve.has(k)) continue;
```
justinfagnani, line 262: I wouldn't generify this yet. A `=== '__forwardedRef'` check is going to be much faster than a `has()` check. We can change later, and for lest than 7-10 keys an array in check will be faster than a Set too.

taylor-vann: sounds good to me!

taylor-vann (commented):
changes addressed

sorvell:
@taylor-vann Can you explain more how/why the current implementation results in the behavior noted in #2593? Thanks.

taylor-vann:
> @taylor-vann Can you explain more how/why the current implementation results in the behavior noted in #2593? Thanks.

for sure @sorvell!

The main take away is we should skip over properties that shouldn't be attached to our webcomponent in our react component's `render` method.

When a ref is passed to our React component, the ref prop will change from `undefined` to defined and drive a render.

During `render` there needs to be a sieve between react props and webcomponent attributes / properties that shouldn't be added to the webcomponent.

Currently, `__forwardRef` will be added to `props` on render

```TS
render() {
  ...
  for (const [k, v] of Object.entries(this.props)) {
    if (elementClassProps.has(k)) {
      this._elementProps[k] = v;
    } else {
      // forwardedRef is sent to react class props
      props[k === 'className' ? 'class' : k] = v;
    }
}
```

Afterwards, when `onComponentDidMount` on `componentDidUpdate` we update the webcomponent with the react those props:
```TS
  override componentDidMount() {
    // props now include forwarded ref
    this._updateElement();
  }
```

`_updateElement` then sets every property from the react props on the element as attributes or event listeners via `setProperty()`, including `__forwardedRef`.

Right now we're only watching the one property `__fowardedRef` with an `if`, but if we find more we could have a `Set` of skippable properties.

kevinpschaaf (approved):
LGTM

Issue #2683, opened 2022-03-30
[infra] Create more unique suacelabs tunnel identifiers


taylor-vann:
## Description

Subtask of #2638 (Step 1)

Create more unique tunnel identifiers for better communication between separate github actions and saucelabs

## Acceptance criteria

4 tunnels are created rrather than 2
one for each saucelabs run per github action event
(currently 2)

justinfagnani:
Closing as outdated.

Issue #2638, opened 2022-03-15
Improve repo test experience / infra


taylor-vann:
## Mitigate False Test Failures (especially from saucelabs)

All the devs want a great testing and CI experience but saucelabs has consistently failed as a reliable infrastructure for CI.
IMO we abstract too much responsibility away from our repo to our tooling. It is difficult to observe responsibility in failures outside of our test suite (ie which tool failed)

Lit should have more control over its own builds and tests.

With github actions maturing, more OS are available for testing this should be possible.

The goal of this issue / task is to explore the following options:
- fix saucelabs proxy issues
- isolate saucelabs to only IE tests on windows
- remove our dependence on saucelabs entirely

## Acceptance criteria

The goal is to provide a CI experience that only (mostly) fails because of test failures, not test runner failures or proxy timeouts.

PR #2684, opened 2022-03-30
[@tests] use more unique identifiers for saucelabs tunnels

taylor-vann:
Saucelabs tunnels should have a 1:1 correlation with a github action.

As in:
lit-tests push tunnel
lit-tests pull_requests tunnel

Rather than running both actions on one tunnel.

This PR is to address #2683, subtask of #2638


.github/workflows/sauce-ie11.yaml
```diff
@@ -42,5 +42,7 @@ jobs:
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           BROWSERS: preset:sauce-ie11
-        # Retry once automatically.
-        run: npm run test || npm run test
+        run: |
+          export SAUCE_BUILD_ID=lit-tests__${GITHUB_EVENT_NAME}__${GITHUB_SHA}
+          export SAUCE_TUNNEL_ID=${SAUCE_BUILD_ID}_tunnel
```
aomarks, line 47: Is there a page we could link to that describes `SAUCE_TUNNEL_ID`? Is that read by web-test-runner? (I couldn't find it from a quick google).

aomarks, line 47: What was being used as the sauce tunnel ID before?

Was the same exact tunnel being used for every invocation of GitHub actions? 

What does Sauce do when it gets two different connections over the same tunnel ID? Does the second attempt to create a tunnel with the ID block until the first tunnel on that ID shuts down? Or does Sauce labs try and disconnect the first one? Or something else?

taylor-vann: Ooops! Forgot to track a file change for wtr, will update shortly

taylor-vann: Before only two tunnels were being used. Both push and pull requests for one action would run on a single tunnel.

It's a little difficult to garner _why_ from the logs, but this change greatly reduces the drops with a log similar to `could not connect to proxy, check internet connection` 

It's also a pattern material-web implements in their tests. This change is the minimal env variables to become unique.

aomarks: Ahh, I was trying to find this variable, but couldn't find it anywhere.

It looks like `SAUCE_TUNNEL_ID` is defined in MWC's karma config, here: https://github.com/material-components/material-web/blob/c96cb2e0cbb0e691e7a3d9ae7c7e91d2b59b12e5/karma.conf.js#L125

So, it's not going to have any effect here, right?

taylor-vann: Appears we had a pretty successful run regardless

taylor-vann: Exactly, I forgot to include the wtf update in this commit, it's effect should be nil at the moment. But next commit it will be included

taylor-vann: This won't change how IE times out but it prevents a single tunnel from running multiple GitHub actions (I think currently it affects multiple PRs not just one PR because we all use the same build name)

PR #2691, opened 2022-03-31
[lit-html] Fix `ref` bug when auto-bound class method used as as callback

### LINKED ISSUE #2677: [lit-html] Ref callbacks incorrect when unbound class methods used as callbacks
# Description

User reported bug: Ref callbacks are called incorrectly (with `undefined` when they shouldn't be) when the same callback is passed to multiple `ref`s. Particularly, if an unbound `LitElement` method is passed to `ref`, this will result in incorrect callbacks when multiple instances of that element are used.

## Live Reproduction Link

https://codesandbox.io/s/lit-ref-8lwb73?file=/src/index.js

## Expected Results

Ref callback only called once for initial render for each element instance.

## Actual Results

First element ref callback called correctly. Second element callback first called with `undefined`, then with the node, which is unexpected.

kevinpschaaf:
The `WeakMap` used to store the last element called to a given callback is keyed on the callback function. Since unbound LitElement class methods passed as callbacks will all have the same references between class instances, this will result in incorrect tracking of the last element per unique callback, given that callbacks called on different context's should be considered unique.

The change here uses nested WeakMaps to effectively double-key the lookup on both function _and_ context.

Fixes #2677.

### DIFF
--- a/.changeset/short-cooks-deliver.md
+++ b/.changeset/short-cooks-deliver.md
@@ -1,5 +1,6 @@
 ---
 'lit-html': patch
+'lit': patch
 ---
 
 Fixs `ref` bug when auto-bound class method used as as callback could incorrectly receive `undefined`.
--- a/.changeset/short-cooks-deliver.md
+++ b/.changeset/short-cooks-deliver.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Fixs `ref` bug when auto-bound class method used as as callback could incorrectly receive `undefined`.
--- a/.changeset/short-cooks-deliver.md
+++ b/.changeset/short-cooks-deliver.md
@@ -3,4 +3,4 @@
 'lit': patch
 ---
 
-Fixs `ref` bug when auto-bound class method used as as callback could incorrectly receive `undefined`.
+Fixes `ref` bug when auto-bound class method used as a callback could incorrectly receive `undefined`.
--- a/packages/lit-html/src/directives/ref.ts
+++ b/packages/lit-html/src/directives/ref.ts
@@ -33,7 +33,7 @@ interface RefInternal {
 // has already been rendered to a new spot. It is double-keyed on both the
 // context (`options.host`) and the callback, since we auto-bind class methods
 // to `options.host`.
-const lastElementForCallback: WeakMap<
+const lastElementForContextAndCallback: WeakMap<
   object,
   WeakMap<Function, Element | undefined>
 > = new WeakMap();
@@ -43,7 +43,7 @@ export type RefOrCallback = Ref | ((el: Element | undefined) => void);
 class RefDirective extends AsyncDirective {
   private _element?: Element;
   private _ref?: RefOrCallback;
-  private _context: object | undefined;
+  private _context?: object;
 
   render(_ref: RefOrCallback) {
     return nothing;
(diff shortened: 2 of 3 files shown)


packages/lit-html/src/directives/ref.ts
```diff
@@ -30,16 +30,20 @@ interface RefInternal {
 
 // When callbacks are used for refs, this map tracks the last value the callback
 // was called with, for ensuring a directive doesn't clear the ref if the ref
-// has already been rendered to a new spot
-const lastElementForCallback: WeakMap<Function, Element | undefined> =
-  new WeakMap();
+// has already been rendered to a new spot. It is double-keyed on both the
+// callback and the context (options.host), since we auto-bind class methods to
+// options.host.
+const lastElementForCallback: WeakMap<
```
sorvell, line 36: It seems like the WeakMap setup would make more sense if it was keyed off the host first.

sorvell (changes requested):

packages/lit-html/src/directives/ref.ts
kevinpschaaf: Done

.changeset/short-cooks-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
```
AndrewJakubowicz, line 2: Nit: Please add lit monorepo patch

I.e.:

```
'lit-html': patch
'lit': patch
```

.changeset/short-cooks-deliver.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fixs `ref` bug when auto-bound class method used as as callback could incorrectly receive `undefined`.
```
AndrewJakubowicz, line 6:
```suggestion
Fixes `ref` bug when auto-bound class method used as a callback could incorrectly receive `undefined`.
```

.changeset/short-cooks-deliver.md
kevinpschaaf: THANK YOU! 😌

packages/lit-html/src/test/directives/ref_test.ts
```diff
@@ -140,41 +140,76 @@ suite('ref', () => {
...
-    queriedEl = container.firstElementChild;
-    assert.equal(queriedEl?.tagName, 'DIV');
-    assert.deepEqual(host.calls, ['DIV', undefined, 'SPAN', undefined, 'DIV']);
+    // Test first instance
+    const host1 = new Host();
+    testRef(host1);
+
+    // Test second instance
+    const host2 = new Host();
+    testRef(host2);
```
AndrewJakubowicz, line 204: Non actionable - checking understanding. The previous behavior would have added `undefined` to the first testRef assertion for the second host giving `[...initialCalls, undefined, 'DIV']);`, because it would accidentally treat the different Hosts as the same context?

kevinpschaaf: Exactly. If we take the fix out, the test would fail on the 2nd host seeing its callback called with `undefined` before it is called with its div, which is what's seen in the original repro: https://codesandbox.io/s/lit-ref-8lwb73?file=/src/index.js

packages/lit-html/src/directives/ref.ts
```diff
@@ -74,19 +74,19 @@ class RefDirective extends AsyncDirective {
...
+      // both the context and the callback, since we allow passing unbound
       // functions that are called on options.host, and we want to treat
       // these as unique "instances" of a function.
-      let lastElementForContext = lastElementForCallback.get(this._ref);
+      const context = this._context ?? globalThis;
+      let lastElementForContext = lastElementForCallback.get(context);
       if (lastElementForContext === undefined) {
         lastElementForContext = new WeakMap();
-        lastElementForCallback.set(this._ref, lastElementForContext);
+        lastElementForCallback.set(context, lastElementForContext);
```
AndrewJakubowicz, line 84: (no action required - comment because I got a bit confused with the naming here but was able to figure it out.)

I think I found it surprising that `lastElementForCallback` has a context as key and in my mind returns a `lastElementForCallback` weakmap.

Could having the global map be called `lastElementForContextAndCallback`, then here it would return `lastElementForCallback`.

Then later (L:89) `lastElementForCallback.set(this._ref, element);` can be used.

AndrewJakubowicz (approved):
Reviewed - although primarily syntax and logic surrounding the issue, since lack expertise in this domain.
This is awesome! TIL! Test is fantastic, and only comments are nits and clarifying my own understanding.

sorvell (approved):

packages/lit-html/src/directives/ref.ts
kevinpschaaf: Yeah, making the names make sense with the double-keying is hard, but I think this is an improvement! Done.

packages/lit-html/src/directives/ref.ts
```diff
@@ -30,16 +30,20 @@ interface RefInternal {
...
+  WeakMap<Function, Element | undefined>
+> = new WeakMap();
 
 export type RefOrCallback = Ref | ((el: Element | undefined) => void);
 
 class RefDirective extends AsyncDirective {
   private _element?: Element;
   private _ref?: RefOrCallback;
-  private _context: unknown;
+  private _context: object | undefined;
```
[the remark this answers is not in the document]
kevinpschaaf: Done!

AndrewJakubowicz (approved):
Brilliant!

PR #2692, opened 2022-03-31
[localize-tools] Keep expressions in attribute values for transform mode

### LINKED ISSUE #2574: [localize] transform mode doesn't handle template strings with quotes very well
Input: `.placeholder=${msg(`What's your question?`)}`

Output: `\n        .placeholder=What's your question?\n`

Workaround: `.placeholder="${msg(`What's your question?`)}"`

augustjk:
Fixes #2574

No longer subsume expressions into text if they follow a `=` with the assumption being they are placed as a value of a html attribute, allowing `lit-html` to parse them properly.

This PR also corrects an oversight from #2405 where we were returning the old `template` rather than `newTemplate` if we had a simple string literal after applying translation.

### DIFF
--- a/packages/localize-tools/src/modes/transform.ts
+++ b/packages/localize-tools/src/modes/transform.ts
@@ -448,7 +448,7 @@ class Transformer {
           'Internal error: string literal cannot be html-tagged'
         );
       }
-      return template;
+      return newTemplate;
     }
 
     // We may have ended up with template expressions that can be represented
@@ -534,7 +534,22 @@ class Transformer {
       return true;
     };
 
-    for (const span of template.templateSpans) {
+    for (let i = 0; i < template.templateSpans.length; i++) {
+      const span = template.templateSpans[i];
+      // A span preceded by `=` can be an attribute so skip subsume and
+      // keep it as an expression to produce valid lit-html template
+      // TODO(augustinekim) Consider optimizing to regular quoted string for
+      // regular html attributes
+      if (
+        (i === 0
+          ? template.head.text
+          : template.templateSpans[i - 1].literal.text
+        ).endsWith('=')
+      ) {
+        fragments.push(ts.visitNode(span.expression, this.boundVisitNode));
+        fragments.push(span.literal.text);
+        continue;
+      }
       let expression = span.expression;
       // Can we directly subsume this span?
       if (!subsume(expression)) {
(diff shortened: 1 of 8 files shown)


packages/localize-tools/src/tests/transform.unit.test.ts
```diff
@@ -353,6 +353,28 @@ test('msg(string(<b>msg(string)</b>)) translated', () => {
   );
 });
 
+test('html(msg(string)) with msg as attr value', () => {
```
aomarks, line 356: Could you add one with a `.bar` property attribute, too? Same result, just for completeness.

packages/localize-tools/src/modes/transform.ts
```diff
@@ -448,7 +448,7 @@ class Transformer {
           'Internal error: string literal cannot be html-tagged'
         );
       }
-      return template;
+      return newTemplate;
```
aomarks, line 451: Oops! Do the tests in this PR fail if this is switched back to `template`? If not, is there a test we could add that would cover it?

aomarks (changes requested):
Awesome! Just a small request for an extra test, and a question.

packages/localize-tools/src/modes/transform.ts
augustjk: Upon taking a closer look now, it seems we never enter this code block if a translation does occur because then `newTemplate` would always be a `ts.TemplateLiteral`. The only time we enter this block is when we don't have to translate, i.e. generating original locale's version of the source, in which case `newTemplate` would not have changed from it's original assignment from `template` here:
https://github.com/lit/lit/blob/f78b4cc0285c4df5248324367453a5c4a0f3d1d6/packages/localize-tools/src/modes/transform.ts#L335
So.. it was technically an oversight but of no consequence, hence none of our existing tests failed. For readability's sake I think keeping `newTemplate` here still makes sense.

aomarks: sgtm

packages/localize-tools/src/tests/transform.unit.test.ts
augustjk: Done, and also added additional tests for having multiple attributes in a single template expression so as to test the logic for looking at both the head and the middle.

aomarks: Nice!

aomarks (approved):
Looks great

PR #2695, opened 2022-04-01
Remove note from @lit-labs/router readme

abdonrd:
The `@lit-labs/router` it's now at npm:

https://npmjs.com/package/@lit-labs/router


abdonrd:
I guess we don't need a changeset for this.

justinfagnani (approved):

PR #2698, opened 2022-04-01
Separate local and saucelabs tests

taylor-vann:
- use `tests/` dir to explicitly run tests sent to saucelabs
- add package.json command `tests:remote` to run saucelabs

### DIFF
--- a/.github/workflows/sauce-ie11.yaml
+++ b/.github/workflows/sauce-ie11.yaml
@@ -43,4 +43,4 @@ jobs:
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           BROWSERS: preset:sauce-ie11
         # Retry once automatically.
-        run: npm run test || npm run test
+        run: npm run test:remote || npm run test:remote
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "nuke": "rm -rf node_modules && npm install && lerna exec 'rm -rf node_modules' && lerna bootstrap && npm run clean",
     "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean",
     "test": "lerna run test --ignore lit --ignore tests --ignore @lit-labs/motion --concurrency 1 --stream",
+    "test:remote": "cd tests && npm run test",
     "prepare": "husky install",
     "changeset": "changeset",
     "version": "npm run changeset version && npm run update-version-vars",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "lint": "eslint \"**/*.{js,ts}\"",
     "nuke": "rm -rf node_modules && npm install && lerna exec 'rm -rf node_modules' && lerna bootstrap && npm run clean",
     "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean",
-    "test": "lerna run test --ignore lit --ignore lit-html --ignore lit-element --ignore @lit/reactive-element --ignore @lit-labs/motion --concurrency 1 --stream",
+    "test": "lerna run test --ignore lit --ignore tests --ignore @lit-labs/motion --concurrency 1 --stream",
     "prepare": "husky install",
     "changeset": "changeset",
     "version": "npm run changeset version && npm run update-version-vars",
--- a/packages/tests/package.json
+++ b/packages/tests/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run test:dev && npm run test:prod",
     "test:dev": "npx wtr",
-    "test:prod": "MODE=prod npm run test:dev",
+    "test:prod": "TEST_PROD_BUILD=true npm run test:dev",
     "test:watch": "npm run test:dev -- --watch"
   },
   "devDependencies": {
--- a/packages/tests/web-test-runner.config.js
+++ b/packages/tests/web-test-runner.config.js
@@ -163,9 +163,18 @@ const require = createRequire(import.meta.url);
 export default {
   rootDir: '../',
   // Note this file list can be overridden by wtr command-line arguments.
+  // Tests run by lerna should override command-line arguments.
+  // Sauce tests are sent in one go to reduce proxy attempts.
   files: [
     '../lit-html/development/**/*_test.(js|html)',
+    '../labs/observers/development/**/*_test.(js|html)',
+    '../labs/react/development/**/*_test.(js|html)',
+    '../labs/router/development/**/*_test.js',
+    '../labs/scoped-registry-mixin/development/**/*_test.(js|html)',
+    '../labs/ssr/development/**/*_test.(js|html)',
+    '../labs/task/development/**/*_test.(js|html)',
     '../lit-element/development/**/*_test.(js|html)',
+    '../lit-html/development/**/*_test.(js|html)',
     '../reactive-element/development/**/*_test.(js|html)',
   ],
   nodeResolve: true,


package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "eslint \"**/*.{js,ts}\"",
     "nuke": "rm -rf node_modules && npm install && lerna exec 'rm -rf node_modules' && lerna bootstrap && npm run clean",
     "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean",
-    "test": "lerna run test --ignore lit --ignore lit-html --ignore lit-element --ignore @lit/reactive-element --ignore @lit-labs/motion --concurrency 1 --stream",
+    "test": "lerna run test --ignore lit --ignore tests --ignore @lit-labs/motion --concurrency 1 --stream",
```
kevinpschaaf, line 18: For the local test run, if we kept `tests` but ignored all the packages that `tests` runs, it seems like that might be more efficient since it's just one `wtr` invocation, although since the tunnels are the slow part of spinning up/down wtr and we don't have those on local, not sure it's a big deal? It's certainly less fiddly to do what you have here, so we don't have to curate a big ignore list of packages.

packages/tests/web-test-runner.config.js
```diff
@@ -56,16 +56,30 @@ function makeSauceLauncherOnce() {
   if (!sauceLauncher) {
     const user = (process.env.SAUCE_USERNAME || '').trim();
     const key = (process.env.SAUCE_ACCESS_KEY || '').trim();
-    if (!user || !key) {
-      throw new Error(
-        'To test on Sauce, set the SAUCE_USERNAME' +
-          ' and SAUCE_ACCESS_KEY environment variables.'
-      );
+    const tunnelIdentifier = (process.env.TUNNEL_IDENTIFIER || '').trim();
```
aomarks, line 59: Suggest `SAUCE_TUNNEL_ID` for consistency with the other variables, and for clarity.

packages/tests/web-test-runner.config.js
```diff
@@ -56,16 +56,30 @@ function makeSauceLauncherOnce() {
...
+      {
+        user,
+        key,
+      },
+      {
+        build: tunnelIdentifier,
+      },
+      {
+        tunnelIdentifier,
+        sharedTunnel: true,
```
aomarks, line 79: Could you add brief comments explaining `sharedTunnel` and `noRemoveCollidingTunnels` and why we want to set them here?

package.json
kevinpschaaf: I do see one test in the `lit` package (and some in `labs/motion`); seems like we should not ignore those if we go with this approach.
```suggestion
    "test": "lerna run test --ignore tests --concurrency 1 --stream",
```

packages/tests/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run test:dev && npm run test:prod",
     "test:dev": "npx wtr",
-    "test:prod": "MODE=prod npm run test:dev",
+    "test:prod": "TEST_PROD_BUILD=true MODE=prod npm run test:dev",
```
aomarks, line 9: What does `TEST_PROD_BUILD` mean? I don't see it actually being consumed anywhere in our code (though I do see it's used in other `test:prod` npm scripts).

packages/tests/web-test-runner.config.js
```diff
@@ -163,14 +171,22 @@ const require = createRequire(import.meta.url);
...
-    '../lit-html/development/**/*_test.(js|html)',
+    '../labs/observers/development/**/*_test.(js|html)',
+    '../labs/react/development/**/*_test.(js|html)',
+    '../labs/router/development/**/*_test.js',
+    '../labs/scoped-registry-mixin/development/**/*_test.(js|html)',
+    '../labs/ssr/development/**/*_test.(js|html)',
+    '../labs/task/development/**/*_test.(js|html)',
     '../lit-element/development/**/*_test.(js|html)',
+    '../lit-html/development/**/*_test.(js|html)',
     '../reactive-element/development/**/*_test.(js|html)',
```
kevinpschaaf, line 185: There's a test in `lit`, and some in `motion`:
```suggestion
    '../reactive-element/development/**/*_test.(js|html)',
    '../lit/development/**/*_test.(js|html)',
    '../labs/motion/development/**/*_test.(js|html)',
```

kevinpschaaf (changes requested):

.github/workflows/sauce-ie11.yaml
```diff
@@ -37,10 +41,15 @@ jobs:
...
 
-      - name: Test
-        env:
-          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
-          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
-          BROWSERS: preset:sauce-ie11
-        # Retry once automatically.
-        run: npm run test || npm run test
+      - name: Create tunnel name
+        run: echo "TUNNEL_IDENTIFIER=lit-tests__${GITHUB_SHA}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
taylor-vann: i think that's a good suggestion :) I was debating whether or not. I think two tunnels per PR isn't so bad. will change.

.github/workflows/sauce-ie11.yaml
```diff
@@ -37,10 +41,15 @@ jobs:
...
+
+      - name: Test Dev
+        run: |
+          cd packages/tests
+          npm run test:dev || npm run test:dev
```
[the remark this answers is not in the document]
taylor-vann: oh! forgot about that will do

.github/workflows/sauce.yaml
```diff
@@ -6,12 +6,17 @@ jobs:
   tests-sauce:
     # Don't run on forks. Sauce tests need access to secrets.
     if: github.repository == 'lit/lit'
-
+
```
aomarks, line 9:
```suggestion

```

package.json
taylor-vann: @labs/motion is currently failing on its own, I think that's why it's being ignored currently.
But adding the lit test is definitely a change i should make

packages/labs/analyzer/index.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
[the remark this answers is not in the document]
taylor-vann: sounds good to me!

aomarks (changes requested):

.github/workflows/sauce-ie11.yaml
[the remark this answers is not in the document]
taylor-vann: Can always reduce later :)

.changeset/eighty-dogs-occur.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
taylor-vann: oh nice! the changeset checks failed so I added one, is there a way to upload changes in `tests` that doesn't need a changeset?

.github/workflows/sauce-ie11.yaml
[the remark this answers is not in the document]
taylor-vann: sounds good!

packages/tests/package.json
taylor-vann: I saw it in the `test` command of a few packages. Was mainly covering my bases. if it's not consumed anywhere, we can get rid of it!

So that's what i'll do :)

packages/tests/web-test-runner.config.js
```diff
@@ -4,218 +4,36 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {createRequire} from 'module';
```
kevinpschaaf, line 7: I would say, let's keep the config together so that the BROWSERS env variable is usable as it was before, since this will likely change people's workflows (the tests package is useful when locally developing, not just CI)...

For example, today I'll often do
```
BROWSERS=chromium npm run test:watch
```
which will run all the tests locally just on Chrome as I'm developing. And then can change to
```
BROWSERS=preset:sauce npm run test
```
to sanity check on older browsers before pushing up a PR.

.github/workflows/sauce.yaml
taylor-vann: I think i got that covered :) there's been some changes so I'll rerequest a review

kevinpschaaf (changes requested):

packages/tests/package.json
taylor-vann: I removed the TEST_PROD_BUILD from all package.jsons

Since TEST_PROD_BUILD is not used anywhere, those packages are just running DEV tests twice.
Those packages now echo "prod tests identical to dev tests, skipping"

Down to change, but I don't see any reason to run tests twice

packages/tests/web-test-runner.config.js
taylor-vann: For sure!

packages/tests/web-test-runner.config.js
taylor-vann: Sounds good

packages/tests/web-test-runner.config.js
```diff
@@ -163,14 +171,22 @@ const require = createRequire(import.meta.url);
 export default {
   rootDir: '../',
   // Note this file list can be overridden by wtr command-line arguments.
+  // Tests run by lerna should override command-line arguments.
```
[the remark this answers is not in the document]
taylor-vann: Will fix

packages/tests/web-test-runner.config.js
```diff
@@ -163,14 +171,22 @@ const require = createRequire(import.meta.url);
...
+    '../labs/ssr/development/**/*_test.(js|html)',
+    '../labs/task/development/**/*_test.(js|html)',
     '../lit-element/development/**/*_test.(js|html)',
+    '../lit-html/development/**/*_test.(js|html)',
     '../reactive-element/development/**/*_test.(js|html)',
   ],
   nodeResolve: true,
-  concurrency: Number(process.env.CONCURRENT_FRAMES || 6), // default cores / 2
-  concurrentBrowsers: Number(process.env.CONCURRENT_BROWSERS || 2), // default 2
+  concurrency: Number(process.env.CONCURRENT_FRAMES || 1), // default cores
```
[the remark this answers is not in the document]
taylor-vann: yah the code looks pretty self explanitory

packages/tests/web-test-runner.config.js
taylor-vann: Added the lit test :)

packages/tests/web-test-runner.config.js
taylor-vann: Thanks for the explanation!

I've added that functionality back into the split config files :)

So you can go:
```
BROWSERS=chromium npm run test:watch
```
and when you need to run sauce tests, it's the same
```
BROWSERS=sauce npm run test:watch
```

If we don't wan't split config files that's fine too!

But I like how separating configs isolates a particular kind of logic. As in:
- web-test-runner.saucelabs.config.js is all about saucelabs settings
- web-test-runner.config is all about local test settings
- wtr-config.js provides shared settings that can be overwritten in either file.

When we separate these configs, a lot of the logic we do to parse details out of "browser strings" starts to look odd. The changes I've uploaded show how we can have more explicit configs without parsing strings for details.

We don't have to go 'preset:local' anymore, no more 'preset' or parsing details out of a string. To me that feels kinda awkward and although it's a neat, long term I think it's difficult to maintain or extend. A plain pojo is just as readable.

My pitch would be 'no magic' / parsed configs. WTR configs might be best served by have explicit lists of browsers and tests.

Like i said though down to change! <3

packages/tests/package.json
taylor-vann: resolving for now! gonna need another review because some files have changed :)

.changeset/eighty-dogs-occur.md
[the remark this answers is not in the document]
aomarks: You can run `npx changeset --empty` to generate an empty one, that way the check will pass.

.github/workflows/sauce.yaml
aomarks: Looks like it's still there.

BTW I use the `"files.trimTrailingWhitespace": true` VSCode setting to do this cleanup automatically on every save.

.github/workflows/tests.yml
```diff
@@ -69,6 +69,6 @@ jobs:
 
       - name: Test
         env:
-          BROWSERS: preset:local
```
aomarks, line 72: Is this removed because `preset:local` is the default?

aomarks (changes requested):

.github/workflows/tests.yml
taylor-vann: yes :)

packages/tests/web-test-runner.config.js
kevinpschaaf: @aomarks Seems ok... I think you set up the browser string parsing scheme?

.changeset/eighty-dogs-occur.md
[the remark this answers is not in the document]
taylor-vann: wanna check that i successfully mad an empty changeset before i ask for another review :)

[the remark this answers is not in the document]
taylor-vann: kewl think i got that old changeset replaced with an empty one

.github/workflows/sauce.yaml
taylor-vann: oh interesting!
i thought yaml files required an trailing space
(github gives me an interresting symbol / suggestion without trailing space)

taylor-vann (commented):
addressed changes, further review needed :)

taylor-vann (commented):
addressed

taylor-vann (commented):
addresed

taylor-vann (commented):
addressed

packages/tests/README.md
```diff
@@ -0,0 +1,234 @@
+# Tests
```
kevinpschaaf, line 1: Thank you for adding, this!

packages/tests/README.md
```diff
@@ -0,0 +1,234 @@
...
+    '../<package_name>/test/**/*_test.(js|html)',
+];
+```
+
+#### Prod mode
+
+Not all packages run in production mode remotely.
+
+For example, packages found in [labs]('../labs/README.md') are inherently
+not meant for production and have no obligation to be tested in `PROD` mode.
```
kevinpschaaf, line 131: Note this is just de-facto; we do make production builds of the labs packages and there's nothing inherently keeping a labs package from adding a DEV_MODE debug warning or check, where we would want to test in both dev and prod -- just most haven't needed that yet. So I would err on just testing everything the same way (i.e. in both dev and prod).

I just saw this discussion: https://github.com/lit/lit/pull/2698#discussion_r844464088, and I actually get the feeling that
```
   "test:prod": "TEST_PROD_BUILD=true npm run test:dev",
```
was meant to be
```
    "test:prod": "MODE=prod npm run test:dev",
```
which is what's found in the code packages, and these packages got left behind after a variable rename or something (the `MODE` env variable is what triggers the prod/dev source remapping: https://github.com/lit/lit/blob/main/packages/tests/web-test-runner.config.js#L15-L18).

@sorvell @justinfagnani Does this match your understanding? Any reason to treat labs packages differently?

packages/tests/web-test-runner.config.js
```diff
@@ -4,218 +4,45 @@
...
-      user,
-      key,
-    });
-  }
-  return sauceLauncher;
+const browsers = [];
+if (browserPresets[requestedBrowsers] !== undefined) {
+  browsers.push(
+    playwrightLauncher({
+      product: requestedBrowsers,
```
kevinpschaaf, line 28: Note, the previous interface for `BROWSERS` was a comma-separated list of browsers (hence the plural) parsed by 
```js
const browsers = (process.env.BROWSERS || 'preset:local')
  .split(',')
  .map(parseBrowser)
  .flat();
```

I don't think this code works with a list of browsers anymore, just one browser or `local`. I don't know everyone's personal workflows for running tests, but I'd normally err on the side of not changing stuff like that. @justinfagnani @sorvell opinions? Do you use this env variable?

packages/tests/README.md
kevinpschaaf: Assuming we actually want to just run everything in dev and prod, then I think this means the `defFiles` / `prodFiles` distinction would go away.

kevinpschaaf: From blame, it looks like that's what happened: `TEST_PROD_BUILD` was renamed to `MODE` and the labs testing of prod mode got left behind: https://github.com/lit/lit/commit/c8b41b7bef2d30d3515d6942ece66fb46e967e29.

Iow I don't think it's intentional that we're not testing `labs` in prod mode.

taylor-vann: I see! makes sense (i was assuming a lot there)
I'll update the prod tests with `MODE` and remove any kind of "we don't run prod tests" statements

packages/tests/web-test-runner.config.js
taylor-vann: I'll add the multi browser support back in :)

My main goals here with the different configs is two fold:
- split saucelabs configurations from local configurations
- avoid recursion in a config file

I kinda took a liberty with the `preset:` keyword. They're all presets so that distinction didn't land for me. Plus when you use it `BROWSERS=preset:local`, we just substring `preset:` out anyways.
```
const preset = browser.substring('preset:'.length);
const entries = browserPresets[preset];
```

If we wanna keep the `preset:` prefix, i can add that support back in too!

packages/tests/README.md
taylor-vann: then we'll keep it!

packages/tests/README.md
taylor-vann: I've added back `MODE=prod` back for labs packages, everything is running prod tests now!

packages/tests/web-test-runner.config.js
taylor-vann: I've added back this functionality
`preset:local` and `preset:sauce` work again
user can define lists of browsers  ie `BROWERS=chromium,firefox`

packages/tests/README.md
taylor-vann: Removed all mentions of "we don't always run prod tests", it very much says "we run prod tests" now :)

taylor-vann (commented):
ive addressed changes and resolved feedback comments :) thanks @kevinpschaaf

packages/tests/README.md
aomarks: FYI we already have some testing documentation at https://github.com/lit/lit/blob/main/CONTRIBUTING.md#tests. Would be nice to unify.

aomarks:
It seems like there's ongoing experimentation happening in this PR, which makes me a little unsure about when/how to review. But there are already some great improvements that it feels like we could be landing!

What do you think about splitting this PR apart into some more narrowly scoped PRs, and landing them incrementally, and then continuing with experimentation to get IE passing in a separate non-PR-associated branch? It would be cool to land incremental improvements, even if we don't get IE to pass right away, and would help with reviewing too, since there would be less to try and understand all at once.

Some ideas for some more narrowly scoped PRs:

- The new SAUCE_TUNNEL_ID improvements
- Fixing the issue where we were running more tests than necessary in the sauce job
- Adding missing tests
- Re-organization of config files
- Updated docs

taylor-vann:
@aomarks sounds good to me!

Long as the reviews for the smaller PRs get done during my spring I don't have a problem with smaller PRs.

These last few pushes were me checking what breaks and doesn't in our sauce configs, I have to push it to GitHub in order to view. Everything passes locallt

taylor-vann:
I'm going to upload a few more changes and then freeze this PR for review.
I'll also leave details about what sepcifically these changes do, and what I'm aiming for, and what plans for the future are.

taylor-vann:
making this a draft so i can isolate what tests consistently pass on saucelabs

taylor-vann:
keeping this in a draft phase and hopefully will parse these out into separate tasks

There is a core chunk of work: decrease sauce tunnels
Dx could be improved: recursion in a config

taylor-vann:
I'll be breaking this into separate PRs:

But before I do, I'd like to point out that everything passes with this current setup
1 tunnel
4 "proxies" / "connections" to the 1 tunnel (ie dev, ie prod, sauce dev, sauce prod)

The caveat is that only lit-html tests provable run successfully. I cannot confirm other tests pass consistently. But with this setup I can confirm our tests will run in all browsers in saucelabs and locally only running lit-html

Which tells me our tests are failing or are flakey. I hope this is not the case but this is the conclusion I have after working on this for a month. We have failing tests. We need to identify which ones fail or flake. It is currently very obtuse to find that out.
The purpose of this effort is to give better fails, allowing us to unbreak ourselves.

But I'd like future reviewers to keep in mind, despite how much we reduce complexity or despite how _confident_ we feel our tests are successful that we should let the actual tests tell us if they should pass or fail rather than our confidence. We might not have "all green" to start adding changes. We need to be okay with this <3

It'll get better, is my main takeaway

taylor-vann:
I can confir that `lit-html` and `reactive-element` tests will pass
onwards

taylor-vann:
`lit-element` consitently proves itself to 'hang' and cause other tests to timeout and fail.
removing and moving on.

taylor-vann:
separating browsers got all checkmarks ;_;

taylor-vann:
"With this setup, we strongly recommend applying a --tunnel-identifier, which will make it easier for tracking tunnels. Otherwise, test traffic initiated using your account will use an unnamed tunnel automatically."
https://docs.saucelabs.com/secure-connections/sauce-connect/setup-configuration/high-availability/

main might all be running on the same tunnel if we do not identify one

taylor-vann:
removing lit-element tests and running jobs per browser give consistent and clear successes.
will try with dev and prod mode smushed together tomorrow.

differences from main:
- run dev and prod modes separate
- unique identifiers for build based on gitaction+gitsha+mode
- shared tunnel per commit

PR #2729, opened 2022-04-13
Remove the empty root repo from the release workflow.

### LINKED ISSUE #2542: Remove empty repo at the root of the release workflow
The release workflow creates a new git repo in the workspace root to work around a bug in the changesets action where git commands are run before changing directories for the `cwd` option. We should stop creating this repo once this bug is fixed.

https://github.com/lit/lit/blob/b8ceafb00651b9a4b5021ac5a3d1d960ceb10e6b/.github/workflows/release.yaml#L27-L31

- [x] https://github.com/changesets/action/issues/159

bicknellr:
This removes the empty root repo and tells `changesets/action` not to set the git user in the repo it's working on, which we already do at the beginning of the workflow.

Fixes #2542.


bicknellr:
Also, what title prefix should I use for these kinds of PRs? Maybe "[meta]" or "[infra]"?

.github/workflows/release.yaml
```diff
@@ -60,6 +54,7 @@ jobs:
           # The `working-directory` option is only available for steps that use
           # `run`. This `cwd` option is the same, but specific to this action.
           cwd: lit
+          setupGitUser: false
```
aomarks, line 57: Add comment explaining this?

aomarks (changes requested):

.github/workflows/release.yaml
bicknellr: Ok, added.

aomarks (approved):

PR #2732, opened 2022-04-15
Enforce use of file extensions in imports

### LINKED ISSUE #2644: lit-html package with type="module" uses extensionless imports in /directive.d.ts causing typescript to fail.
Note the differences in https://unpkg.com/browse/lit-html@2.2.1/directive.d.ts
vs https://unpkg.com/browse/lit@2.2.1/directive.d.ts

the lit directive.d.ts specifies an extension.
the lit-html directive.d.ts does not, which causes the typescript transpiler to fail!

note: https://unpkg.com/browse/lit-html@2.2.1/package.json specifies itself as "type": module which is why extensionless imports fail.

justinfagnani:
Fixes #2644 and adds an eslint plugin to enforce extensions going forward.

### DIFF
--- a/.eslintignore-sync
+++ b/.eslintignore-sync
@@ -6,3 +6,4 @@ packages/{*,labs/*}/{.gitignore,.eslintignore}
 
 [inline]
 packages/localize/testdata/
+packages/localize/examples/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,11 +9,16 @@
     "ecmaVersion": 2020,
     "sourceType": "module"
   },
-  "plugins": ["eslint-plugin-no-only-tests", "@typescript-eslint"],
+  "plugins": [
+    "eslint-plugin-import",
+    "eslint-plugin-no-only-tests",
+    "@typescript-eslint"
+  ],
   "env": {
     "browser": true
   },
   "rules": {
+    "import/extensions": ["error", "always", {"mjs": "never"}],
     "no-prototype-builtins": "off",
     "no-only-tests/no-only-tests": "error",
     "@typescript-eslint/ban-types": "off",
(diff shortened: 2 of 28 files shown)


.eslintrc.json
```diff
@@ -9,11 +9,16 @@
...
+  "plugins": [
+    "eslint-plugin-import",
+    "eslint-plugin-no-only-tests",
+    "@typescript-eslint"
+  ],
   "env": {
     "browser": true
   },
   "rules": {
+    "import/extensions": ["error", "always", {"mjs": "never"}],
```
justinfagnani, line 21: The `.mjs` exclusion is for imports of `uvu/assert`. The TypeScript version we're using doesn't resolve `uvu/assert.mjs`. That might be fixed in a newer version.

.eslintignore-sync
```diff
@@ -6,3 +6,4 @@ packages/{*,labs/*}/{.gitignore,.eslintignore}
 
 [inline]
 packages/localize/testdata/
+packages/localize/examples/
```
justinfagnani, line 9: This wasn't ignored before and the examples were causing all kinds of lint errors. Are we correctly running lint on CI? @aomarks

.eslintrc.json
justinfagnani: Weird... locally this fixes the problems with `uvu/assert`. On CI it doesn't. 🤷‍♂️

rictic: I can repro the issue locally by doing:

```
git clone git@github.com:lit/lit.git
git checkout extensions
npm ci
npm run lint
```

rictic (approved):

.eslintrc.json
justinfagnani: Well, I couldn't get whatever was causing things to pass locally to work on CI, so I just eslint-ignore'ed those imports.

PR #2733, opened 2022-04-15
[@tests] add prod mode to labs packages

taylor-vann:
Labs packages will test with prod and dev modes

Previously `TEST_PROD_MODE=true` was used to flag prod mode.
This has been changed to `MODE=prod`


aomarks (approved):

aomarks:
Nice! Thanks for the small scoped PR.

It's expected that the Sauce tests are still failing, so that's fine. The changesets failure is also fine, since this doesn't affect any published packaged. (We can also do `npx changeset --empty` to make it green).

taylor-vann:
added an empty changeset! thanks @aomarks

aomarks (approved):

PR #2743, opened 2022-04-15
[@tests] Add dev and prod test commands

taylor-vann:
Some packages need dev tests
Others need prod tests

This PR adds `test:prod` and `test:dev` commands to packages where they are missing.

It also adds a dev job and prod job to github actions.


packages/localize-tools/package.json
```diff
@@ -22,7 +22,9 @@
     "clean": "git clean -dXn | sed 's/^Would remove //' | grep -v node_modules | xargs rm -rf",
     "build": "npm run clean && npm run generate-json-schema && tsc --build",
     "generate-json-schema": "typescript-json-schema tsconfig.schema.json ConfigFile --include=src/types/config.d.ts --required --noExtraProps > config.schema.json",
-    "test": "npm run test:unit && npm run test:check-tsc",
+    "test": "npm run test:dev && npm run test:prod",
+    "test:dev": "npm run test:unit && npm run test:check-tsc",
```
aomarks, line 26: It feels a little weird to have `test:dev` and `test:prod` in packages where that distinction doesn't have any meaning.

It also seems like it would be easy to mistakenly create a new package (like `cli`) which has a `test` target, and assume that will automatically get run during CI, because that is the widely used convention that most packages used, but with this scheme it won't get run at all.

So I would suggest a slight variation:

- Use `test` for `prod` tests, and for packages where there is no dev vs prod distinction.

- Use `test:dev` only in the packages where that distinction exists.

Also note that `lerna run <script>` only runs `<script>` in the packages that have that script, so we don't need to define no-op `test:dev` scripts in packages without the prod vs dev distinction.

aomarks (changes requested):

justinfagnani:
@taylor-vann taking a look now, but an initial question I have is how this helps the Sauce and IE situations? It'd be good to document that in this commit so we know why it was made.

taylor-vann:
the original commit this PR is apart of has an explanation and has already been reviewed.
https://github.com/lit/lit/pull/2698#issue-1190328705

These smaller PRs were supposed to be about getting this submitted in incremental steps and to provide better observability. If we could keep conversations about this PR in  https://github.com/lit/lit/pull/2698 that would be really helpful.

This is not supposed to represent an entire PR or explain itself
@aomarks @justinfagnani

taylor-vann:
If this incremental approach isn't possible, id rather keep everything in one PR to stay organized

taylor-vann:
lol all tests passed but they weren't supposed to yet! ;_;

packages/localize-tools/package.json
taylor-vann: If there is no distinction or discernable difference between DEV and PROD then we should not be running two sets of tests.

If there is a functional and _testable_ difference between the patterns and behaviors of DEV and PROD mode, then we should separate these environments / contexts / whatever.

Isolating DEV failures and PROD failures is critical.

If we are incorrectly using `staging` terminology and the only difference is whether DEV logs more output than PROD, then we should only test PROD and leave DEV as a `--verbose` mode to help with tests.

taylor-vann:
closing. breaking these up isn't working for me, the changes are difficult to explain without other PRs or changes present

Issue #2009, opened 2021-07-15
@lit-labs/motion tests failing


aomarks:
`@lit-labs/motion` is not included in the lit package.json `test` script, so it is not currently tested in CI (see https://github.com/lit/lit/pull/1919/files which improves this script to not require enumerating packages, but currently has to exclude motion since it is failing). It is currently failing.

Repro:

```
~/work/lit npx lerna run test --scope @lit-labs/motion
lerna notice cli v4.0.0
lerna info versioning independent
lerna notice filter including "@lit-labs/motion"
lerna info filter [ '@lit-labs/motion' ]
lerna info Executing command in 1 package: "npm run test"
lerna ERR! npm run test exited 1 in '@lit-labs/motion'
lerna ERR! npm run test stdout:

> @lit-labs/motion@1.0.0-rc.2 test
> npm run test:dev && npm run test:prod


> @lit-labs/motion@1.0.0-rc.2 test:dev
> cd ../../tests && npx wtr '../labs/motion/development/**/*_test.js'

Using development builds

Chromium: |██████████████████████████████| 0/2 test files | 0 passed, 0 failed
Firefox:  |██████████████████████████████| 0/2 test files | 0 passed, 0 failed
Webkit:   |                              | 0/2 test files | 0 passed, 0 failed

Running tests...

Running 2 test files...

../labs/motion/development/test/flip_test.js:

 ❌ Flip > animates layout change (failed on Firefox)
      AssertionError: expected -1.600006103515625 to equal 0
      + expected - actual
      
      --1.600006103515625
      +0
      
      at r (../labs/motion/node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:9574:12)
      at n.exports<[3]</t.exports/i.prototype.assert (../labs/motion/node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:250:12)
      at n.exports<[6]</t.exports/r.equal (../labs/motion/node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:4365:9)
      at ../labs/motion/src/test/flip_test.ts:148:11

 ❌ Flip > adjusts for ancestor position (failed on Firefox)
      AssertionError: expected { Object (left, top, ...) } to deeply equal { Object (left, top, ...) }
      + expected - actual
      
       {
      -  "height": 1.9999984105438164
[...]

PR #2748, opened 2022-04-19
[tests] Consolidate additional browser tests into tests/ package

taylor-vann:
Running `npm run test` will execute all tests in dev mode and all tests in prod mode now.

Lerna will not run individual packages.

This should reduce the amount of saucelabs tunnels called _implicitly_

Further PRs will adjust DX and make our configs more clear


packages/tests/web-test-runner.config.js
```diff
@@ -164,8 +164,14 @@ export default {
   rootDir: '../',
   // Note this file list can be overridden by wtr command-line arguments.
   files: [
-    '../lit-html/development/**/*_test.(js|html)',
+    '../labs/observers/development/**/*_test.(js|html)',
+    '../labs/react/development/**/*_test.(js|html)',
+    '../labs/router/development/**/*_test.js',
+    '../labs/scoped-registry-mixin/development/**/*_test.(js|html)',
+    '../labs/ssr/development/**/*_test.(js|html)',
```
aomarks, line 171: I think that `ssr` won't work in this consolidated `tests/` configuration, because it needs an extra command-line flag, and it needs to register its special SSR plugin:

https://github.com/lit/lit/blob/3e181bcb3d969775eda799fd6fcae1ead843225b/packages/labs/ssr/package.json#L19

https://github.com/lit/lit/blob/3e181bcb3d969775eda799fd6fcae1ead843225b/packages/labs/ssr/web-test-runner.config.js#L20

So for now I would suggest not including `ssr` in this consolidated config, instead letting it run in its own standalone `wtr` invocation as before.

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "eslint \"**/*.{js,ts}\"",
     "nuke": "rm -rf node_modules && npm install && lerna exec 'rm -rf node_modules' && lerna bootstrap && npm run clean",
     "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean",
-    "test": "lerna run test --ignore lit --ignore lit-html --ignore lit-element --ignore @lit/reactive-element --ignore @lit-labs/motion --concurrency 1 --stream",
+    "test": "cd packages/tests && npm run test",
```
aomarks, line 18: IIUC, it seems like this might cause our `tests-local` job to stop including any of our Node tests. For example the `ssr`, `localize-tools`, `ts-transformers` packages all have Node tests which aren't handled by the consolidated `tests/` package.

https://github.com/lit/lit/blob/3e181bcb3d969775eda799fd6fcae1ead843225b/.github/workflows/tests.yml#L74

One potential short term fix for this PR could be to add some more `--ignore` statements to the `lerna` command, instead of dropping the lerna command (so, an additional `--ignore` for each of the new packages that have been added to the consolidated `tests` config, minus ssr [see other comment]).

aomarks (changes requested):
Nice! Seems like the right step, thanks for the bite-sized PR!

As I understand it, the big benefit of this change is that we'll be starting up fewer Sauce tunnels, because instead of doing a unique `wtr` invocation for each of these packages (each of which starts up a unique Sauce tunnel), we now just have one invocation, with a single Sauce tunnel shared for all of them.

I just have a couple comments relating to how to account for the Node tests.

package.json
taylor-vann: I'd like to consolidate `--ignore` statements into a single `--ignore` similar to [this setup](https://github.com/lit/lit/blob/38bdf5cb47a1d76bb7a5cb2e5afae5dc43ae282e/package.json#L19)

taylor-vann: take advantage of pattern matching, looks more readable too imho
```
"lerna run test:dev --ignore '{lit,tests,@lit-labs/motion}' --concurrency 1 --stream ",
```

aomarks: Cool. Yeah that is more readable, I agree. (TIL Lerna supports glob-style braces in its flags).

packages/tests/web-test-runner.config.js
taylor-vann: this is similar / related to what i proposed in the original PR
I feel like we shouldn't have to choose here.

we should be able to run all tests locally if we want
and run all dev tests locally when dev tests fail
and run all prod tests locally when prod fails fail
and run all tests remotely
and run all dev tests remotely
and run all prod tests remotely

We need the ability to run all tests, and also isolate where we run tests to resolve blocks


I'd prefer the option to run all `dev` tests and `prod` tests

taylor-vann: Also SSR totally runs and passes so not sure what I'd be 'fixing' there

aomarks: > Also SSR totally runs and passes so not sure what I'd be 'fixing' there

Ah interesting. I think the reason it apparently passes must be that there are actually no files inside `labs/ssr/development`, so `wtr` is just not matching any test files there. We don't emit anything there; there is no prod vs dev distinction for SSR. 

```
> cd lit
> npm run build
> ls packages/labs/ssr/development
No such file or directory
```

taylor-vann: I think adding an `"integration": "cd packages/tests ..."` is a good start.

I'll be pushing for the ability to run tests based on mode in isolation in the future though. I feel like that's pretty standard and something we need to support _if_ we're gonna treat dev and prod mode as distinct

package.json
justinfagnani: That needs to be `"lerna run test --ignore ..."`, yeah?

packages/tests/web-test-runner.config.js
```diff
@@ -164,8 +164,14 @@ export default {
   rootDir: '../',
   // Note this file list can be overridden by wtr command-line arguments.
   files: [
-    '../lit-html/development/**/*_test.(js|html)',
+    '../labs/observers/development/**/*_test.(js|html)',
```
justinfagnani, line 167: Need `motion` in here too

aomarks: Motion tests have never been tested in CI, and they currently fail: https://github.com/lit/lit/issues/2009

So I'd suggest handling that in a separate PR.

packages/tests/web-test-runner.config.js
justinfagnani: We can add `NODE_OPTIONS=--experimental-vm-modules` to the main wtr script, yes?

justinfagnani (changes requested):
Looks great. Just a couple of comments I sent separate.

packages/tests/web-test-runner.config.js
aomarks: In addition to the `--experimental-vm-modules` flag, SSR also needs this separate middleware loaded: https://github.com/lit/lit/blob/3e181bcb3d969775eda799fd6fcae1ead843225b/packages/labs/ssr/web-test-runner.config.js#L18

I'm not sure what the implication of having that middleware loaded for all of our other packages would be.

So I was thinking for now, safer to just keep SSR separate, and consider consolidating separately?

packages/tests/web-test-runner.config.js
justinfagnani: ahh... god point.

package.json
taylor-vann: added packages tested in `tests/` to the lerna ignore line!
Gonna resolve this and request a new review :)

packages/tests/web-test-runner.config.js
taylor-vann: (for future brian, SSR wasn't even being run)

aomarks: I think we should remove the `ssr` glob from this list, even though we know it matches no files, just because it's a bit misleading.

taylor-vann: resolving because I think we agreed to let SSR test in the saucelabs job but not _in_ saucelabs browsers.

aomarks (changes requested):
Code looks great! Can I just suggest a different PR title, since it goes into the Git history?

```
[tests] Consolidate additional browser tests into tests/ package
```

aomarks (approved):
LGTM :tada:

justinfagnani (approved):
awesome!

PR #2756, opened 2022-04-20
Regenerate package locks

justinfagnani:
This was done with a `npm nuke` rune that *also* removed package locks before bootstrapping. I'm trying this because `npm nuke` alone wasn't picking up new transitive dependency versions that dependabot was trying to update.


aomarks (changes requested):
Build is broken with a Rollup error

PR #2770, opened 2022-04-22
[labs/context] Removes use of WeakRef

benjamind:
@justinfagnani realised this was using WeakRef unnecessarily.

The `ContextRoot` should only handle those requests which are intended to be subscribers to updates. Otherwise the semantics are somewhat lost of a one-time context request. By making ContextRoot only work for those context requests which intend to subscribe to future updates, we can remove WeakRef usage as the connected components will clean themselves up through disconnection. This also broadens browser support!

### DIFF
--- a/packages/labs/context/src/lib/context-root.ts
+++ b/packages/labs/context/src/lib/context-root.ts
@@ -11,16 +11,12 @@ import {ContextProviderEvent} from './controllers/context-provider.js';
 type UnknownContextKey = ContextKey<unknown, unknown>;
 
 /**
- * Makes all object properties in a type into WeakRefs
- */
-type MakeWeak<T> = {[P in keyof T]: T[P] extends object ? WeakRef<T[P]> : T[P]};
-
-/**
  * A context request, with associated source element, with all objects as weak references.
  */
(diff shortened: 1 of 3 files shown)


packages/labs/context/src/lib/context-root.ts
```diff
@@ -83,9 +78,8 @@ export class ContextRoot {
   ) => {
     // store a weakref to this element under the context key
     const request: PendingContextRequest = {
-      element: new WeakRef(ev.target as HTMLElement),
-      callback: new WeakRef(ev.callback),
-      subscribe: ev.subscribe,
+      element: ev.target as HTMLElement,
```
justinfagnani, line 81: don't you need to check `ev.subscribe === true` first?

justinfagnani (changes requested):

packages/labs/context/src/lib/context-root.ts
benjamind: Good point, fixed.

benjamind:
Should be good for another review.

justinfagnani (approved):

PR #2771, opened 2022-04-25
chore: correction in benchmarks readme

tlouisse:
The benchmarks package README instructs to change directory to non existing folder 'lit-html' (a leftover from the past.) Changed it to 'lit'.


augustjk:
Thanks for the PR. Looks like this was already taken care of somewhere else!

PR #2777, opened 2022-04-25
[@tests] run browsers serially in saucelabs

taylor-vann:
run browsers separately

### DIFF
--- a/.github/workflows/sauce.yaml
+++ b/.github/workflows/sauce.yaml
@@ -55,10 +55,10 @@ jobs:
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           BROWSERS: sauce:macOS 10.15/safari@latest
         run: |
-          cd packages/labs/ssr
-          npm run test:integration
+          cd packages/tests
+          npm run test
 
-      - name: Test SSR chromium
+      - name: Test chromium SSR
         env:
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
@@ -67,7 +67,7 @@ jobs:
           cd packages/labs/ssr
           npm run test:integration
 
-      - name: Test SSR firefox
+      - name: Test firefox SSR
         env:
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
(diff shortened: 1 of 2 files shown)


aomarks (changes requested):
Nice! Definitely in favor of trying out serial execution of the browsers, and it looks like this passed which is very promising.

We have these environment variables `CONCURRENT_FRAMES` and `CONCURRENT_BROWSERS` that seem like they would also control concurrency:

https://github.com/lit/lit/blob/0e3c9f8237b526404d67023d4702f705ef37dfa6/packages/tests/web-test-runner.config.js#L177

In my mental model, setting `CONCURRENT_BROWSERS` to 1 should have a very similar effect to this PR.

What do you think about creating a separate PR that just changes that environment variable, to see if that also looks stable?

taylor-vann:
@aomarks (wish i could just comment or reference your comment ugh lol)
I've turned the concurrency down with those variables but they seem to run browsers serially but still reserve all of them via launchers.

I think the difference is:
current: reserve 3 browsers, run 1 browser at a time
pr change: reserve 1 browser, run 1 browser at a time

I couldn't get passes like this with CONCURRENT_FRAMES or CONCURRENT_BROWSERS

taylor-vann:
I made #2779 to better target SSR tests alongside this change

aomarks:
> I think the difference is: current: reserve 3 browsers, run 1 browser at a time pr change: reserve 1 browser, run 1 browser at a time

Oh, fascinating, nice find! That actually sounds like it might be @web/test-runner bug to me. It feels like browsers should be spun up lazily, right? It would be great to start a thread in the `open-wc` channel on our slack about this, I bet we'd get some feedback quickly.

aomarks:
> @aomarks (wish i could just comment or reference your comment ugh lol)

Yeah this is a pain. One trick I've learned is that if you highlight somebody's comment in GitHub, and type "R", then it will jump you to the comment box, with that part of the comment in a `>` quote block above.

taylor-vann:
> Yeah this is a pain. One trick I've learned is that if you highlight somebody's comment in GitHub, and type `r`, then it will jump you to the comment box, with that part of the comment in a `>` quote block above.

nice trick!

taylor-vann:
> That actually sounds like it might be @web/test-runner bug to me. It feels like browsers should be spun up lazily, right? It would be great to start a thread in the `open-wc` channel on our slack about this, I bet we'd get some feedback quickly.

i feel browsers should definitely be spun up lazily.
I think this is also correlated with "running out of browsers". when i split browser runs up like this, we get much much much less 500 errors or straight drops on proxy connections.

aomarks (approved):
Nice!

Of a bunch of experiments we've been doing lately, this seems to have the highest success rate, so I think we should merge this and see how it fares over the next few days.

It's not super fast, but it would be great to get a green baseline, and then we can keep experimenting and optimizing going forward.

taylor-vann:
oh dang! first fail on IE tests.

taylor-vann:
@aomarks this makes me wanna give it another go tomorrow :)

taylor-vann:
closing. This PR hints at a solution but does not provide one.

PR #2779, opened 2022-04-25
[@tests] separate local and saucelabs config

taylor-vann:
This PR is a stylistic change to how we setup our web-test-runner configs.

The goal is to simplify how we generate browser launchers and run tests.

By separating local and saucelabs launchers in different configs, we can better control their environments.


taylor-vann:
closing because the goal in this PR was accomplished by #2837 @justinfagnani

Issue #2783, opened 2022-04-26
[analyzer] Add declarations and exports to modules


justinfagnani:
We produce package and module objects currently. We need to add declarations and export to modules, on our way to analyzing classes and LitElements.

PR #2789, opened 2022-04-27
Add minimal class declaration gathering

justinfagnani:
Part of #2783

### DIFF
--- a/packages/labs/analyzer/src/lib/analyzer.ts
+++ b/packages/labs/analyzer/src/lib/analyzer.ts
@@ -54,7 +54,7 @@ export class Analyzer {
       throw new Error('Compiler errors');
     }
     const rootFileNames = this.program.getRootFileNames();
-    console.log('rootFileNames', rootFileNames);
+
     const modules = [];
     for (const fileName of rootFileNames) {
       modules.push(this.analyzeFile(fileName as AbsolutePath));
--- a/packages/labs/analyzer/src/lib/analyzer.ts
+++ b/packages/labs/analyzer/src/lib/analyzer.ts
@@ -5,13 +5,14 @@
  */
 
 import ts from 'typescript';
-import {Package, Module} from './model.js';
+import {Package, Module, ClassDeclaration} from './model.js';
+import {AbsolutePath, absoluteToPackage} from './paths.js';
 
 /**
  * An analyzer for Lit npm packages
  */
 export class Analyzer {
-  readonly packageRoot: string;
+  readonly packageRoot: AbsolutePath;
   readonly commandLine: ts.ParsedCommandLine;
   readonly program: ts.Program;
   readonly checker: ts.TypeChecker;
@@ -20,7 +21,7 @@ export class Analyzer {
    * @param packageRoot The root directory of the package to analyze. Currently
    * this directory must have a tsconfig.json file.
    */
-  constructor(packageRoot: string) {
+  constructor(packageRoot: AbsolutePath) {
     this.packageRoot = packageRoot;
 
     const configFileName = ts.findConfigFile(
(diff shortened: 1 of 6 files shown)


packages/labs/analyzer/test-files/basic-elements/src/not-lit.ts
```diff
@@ -0,0 +1,3 @@
+class LitElement {}
+
+class NotLit extends LitElement {}
```
rictic, line 3: Getting lint errors here, we should exclude packages/labs/analyzer/test-files from linting

packages/labs/analyzer/src/test/analyzer_test.ts
```diff
@@ -4,36 +4,71 @@
...
+  new URL('../test-files/basic-elements', import.meta.url).href
+) as AbsolutePath;
+
+let analyzer: Analyzer;
+
+basicTest.before(() => {
+  analyzer = new Analyzer(basicElementsPackagePath);
+});
+
+basicTest('Reads project files', () => {
```
rictic, line 29: You can scope data to a suite in uvu by giving types to the context object, like so:

```suggestion
const basicTest = suite<{analyzer: Analyzer, package: AbsolutePath}>('Basic');

basicTest.before((ctx) => {
  ctx.package = const basicElementsPackagePath = fileURLToPath(
  new URL('../test-files/basic-elements', import.meta.url).href
) as AbsolutePath;
  ctx.analyzer = 
  analyzer = new Analyzer(basicElementsPackagePath);
});

basicTest('Reads project files', ({analyzer, package}) => {
```

packages/labs/analyzer/src/test/analyzer_test.ts
```diff
@@ -4,36 +4,71 @@
...
 
 import {Analyzer} from '../lib/analyzer.js';
+import {AbsolutePath} from '../lib/paths.js';
 
-test('Basic elements - isLitElement', () => {
-  const basicElementsPackagePath = fileURLToPath(
-    new URL('../test-files/basic-elements', import.meta.url).href
-  );
-  const analyzer = new Analyzer(basicElementsPackagePath);
+const basicTest = suite('Basic');
```
rictic, line 17: I like the style of naming the object returned from suite `test` by default. It being called `basicTest` instead of `test` everywhere in this file doesn't convey much info, but it does make it look different from most tests

packages/labs/analyzer/src/lib/paths.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A path relative to a package root
+ */
+export type PackagePath = string & {
+  __packagePathBrand: never;
+};
+
+export const absoluteToPackage = (
+  path: AbsolutePath,
+  root: AbsolutePath
```
rictic, line 20: For clarity, prefer naming this `packageRoot` rather than `root`, which has a different meaning

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -14,10 +15,35 @@ export class Package {
...
+export type Declaration = ClassDeclaration;
+
+export interface ClassDeclarationInit {
+  name: string | undefined;
+  node: ts.ClassDeclaration;
+}
+
+export class ClassDeclaration {
+  name: string | undefined;
+  node: ts.ClassDeclaration;
```
rictic, line 43: readonly?

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -50,21 +51,38 @@ export class Analyzer {
     if (diagnostics.length > 0) {
       console.error('Please fix errors first');
       console.error(diagnostics);
-      throw new Error('Copmiler errors');
+      throw new Error('Compiler errors');
     }
     const rootFileNames = this.program.getRootFileNames();
+    console.log('rootFileNames', rootFileNames);
```
rictic, line 57: console

rictic (approved):
Some stylistic / clarity recommendations, otherwise LGTM

packages/labs/analyzer/src/test/analyzer_test.ts
justinfagnani: I'm a little unsure if this delivers a lot, but I'll try it!

packages/labs/analyzer/src/test/analyzer_test.ts
justinfagnani: I don't actually like uvu's style here, and prefer nested suites. I was assuming there'd eventually be more suites per file, but I'll rename this for now.

PR #2796, opened 2022-04-27
[analyzer] Initial support for finding LitElement declarations

justinfagnani:
This only supports getting the tag name from the `customElement` decorator. I added a TODO for finding `customElement.define()` calls. I want to mirror the custom elements manifest approach of allowing these to be separate, where the `customElement.define()` calls are considered a type of "export" so that you could have more than one, but for many cases it'll be very useful to understand that a class is intended to have a single specific tag name.

### DIFF
--- a/packages/labs/analyzer/src/lib/analyzer.ts
+++ b/packages/labs/analyzer/src/lib/analyzer.ts
@@ -78,18 +78,19 @@ export class Analyzer {
 
     for (const statement of sourceFile.statements) {
       if (ts.isClassDeclaration(statement)) {
+        const name = statement.name?.text;
         if (this.isLitElement(statement)) {
           module.declarations.push(
             new LitElementDeclaration({
               tagname: getTagName(statement),
-              name: statement.name?.getText(),
+              name,
               node: statement,
             })
           );
         } else {
           module.declarations.push(
             new ClassDeclaration({
-              name: statement.name?.getText(),
+              name,
               node: statement,
             })
           );
@@ -109,7 +110,7 @@ export class Analyzer {
     return (
       this._isLitElementModule(node.getSourceFile()) &&
       ts.isClassDeclaration(node) &&
-      node.name?.getText() === 'LitElement'
+      node.name?.text === 'LitElement'
     );
   };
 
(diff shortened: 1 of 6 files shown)


packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -116,3 +131,34 @@ export class Analyzer {
...
+const getTagName = (declaration: ts.ClassDeclaration) => {
+  // TODO (justinfagnani): support customElements.define()
+  let tagname: string | undefined = undefined;
+  const customElementDecorator = declaration.decorators?.find(
+    isCustomElementDecorator
+  );
+  if (
+    customElementDecorator !== undefined &&
+    customElementDecorator.expression.arguments.length === 1 &&
+    ts.isStringLiteral(customElementDecorator.expression.arguments[0])
```
rictic, line 144: 🏌️ 

```suggestion
  if (
    customElementDecorator?.expression.arguments.length === 1 &&
    ts.isStringLiteral(customElementDecorator.expression.arguments[0])
```

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -116,3 +131,34 @@ export class Analyzer {
...
+  }
+  return tagname;
+};
+
+const isCustomElementDecorator = (
+  decorator: ts.Decorator
+): decorator is CustomElementDecorator =>
+  ts.isCallExpression(decorator.expression) &&
+  ts.isIdentifier(decorator.expression.expression) &&
+  decorator.expression.expression.getText() === 'customElement';
```
rictic, line 156: Never call getText(), it's present on every node but it doesn't always work, and it can masks type problems (e.g. you might think you've established that you're working with an Identifier but there's another case unhandled)

identifiers have a text field that's fine to read

packages/labs/analyzer/src/test/analyzer_test.ts
```diff
@@ -60,14 +61,28 @@ test('isLitElement returns false for non-LitElement', ({
...
+test('Analyzer finds class declarations', ({analyzer}) => {
+  const result = analyzer.analyzePackage();
+  const elementAModule = result.modules.find(
+    (m) => m.path === 'src/class-a.ts'
+  );
+  assert.equal(elementAModule?.declarations.length, 1);
+  assert.equal(elementAModule?.declarations[0].name, 'ClassA');
+});
+
+test('Analyzer finds LitElement declarations', ({analyzer}) => {
```
rictic, line 73: Let's add a test for a LitElement without a tag name

rictic (changes requested):

packages/labs/analyzer/src/lib/analyzer.ts
justinfagnani: +1

rictic (approved):

PR #2797, opened 2022-04-27
[@lit-labs/eleventy-plugin-lit] Fix worker script to run in ESM

### LINKED ISSUE #2795: [@lit-labs/eleventy-plugin-lit] Worker mode not loading in ESM
# Description

Plugin fails to run in `worker` mode due to `worker.js` file not being loaded as ESM.

Error:
```
> eleventy --config=.eleventy.cjs --serve

(node:62151) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
Unexpected error while rendering lit component in worker thread /Users/augustinekim/work/eleventy-test-2/node_modules/@lit-labs/eleventy-plugin-lit/worker/worker.js:6
import { parentPort } from 'worker_threads';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

# Solution

In the package we ship, the `worker` subdirectory should include its own `package.json` with `"type": "module"` (or we rename the TSC emitted js file to mjs)

augustjk:
Fixes #2795 

The `worker/worker.js` file included in the package needed to be run as ES module but would not by default since it was being loaded from the `index.js` written in commonjs to work with eleventy.

This worked during local testing because there is a `worker/package.json` file with `"type": "module"` but this was not included in the package so the npm distributed package would not work. Fixed by including this in the package. Tested locally by running `npm pack` and installing from tarball in test project.

### Also
Includes changes to `package-lock.json`s from having run `npm run nuke`.

### DIFF
--- a/.changeset/large-hounds-exist.md
+++ b/.changeset/large-hounds-exist.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': patch
+---
+
+Fix worker mode by ensuring worker file is run as ES module.
--- a/packages/labs/cli/package-lock.json
+++ b/packages/labs/cli/package-lock.json
@@ -1,3621 +1,3405 @@
 {
-  "name": "@lit-labs/cli",
-  "version": "0.0.0",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
-      "name": "@lit-labs/cli",
-      "version": "0.0.0",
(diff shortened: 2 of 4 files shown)


aomarks (approved):
Great find!

justinfagnani (approved):

yinonov:
Thanks for handling it so quickly!

Issue #2799, opened 2022-04-28
[labs/react] Properties with custom accessors (get/set) does not work with createComponent


lideen:
# Description

Properties with custom accessors (get/set) does not work with createComponent:

```ts
export class BasicElement extends LitElement {
  set customAccessors(customAccessors: Foo) {
    const oldValue = this._customAccessors;
    this._customAccessors = customAccessors;
    this.requestUpdate('customAccessors', oldValue);
  }
  get customAccessors(): Foo {
    return this._customAccessors;
  }
  private _customAccessors = {
    bar: 'inital',
  };

  static elementName = "basic-element";

  static properties = {
    customAccessors: { type: Object },
  };

  render(): TemplateResult {
    return html`<span>${this.customAccessors.bar}</span>`;
  }
}

const BasicElementComponent = createComponent(
  React,
  BasicElement.elementName,
  BasicElement,
  {}
);

class App extends React.Component<AppProps, AppState> {
  render() {
    return (
      <React.Fragment>
        <BasicElementComponent />
        <BasicElementComponent customAccessors={{ bar: 'updated' }} /> {/* <- Property is not being set */}
      </React.Fragment>
    );
  }
}

```

# Steps to Reproduce

1. See example in the description

## Live Reproduction Link

https://stackblitz.com/edit/react-ts-cpu8mh

## Expected Results

I expect properties with custom accessors work just as normal properties

## Actual Results

See stackblitz example

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

sorvell:
This fails because these are non-enumerable and we're using `for (const i in elementClass.prototype)` to find properties. The PR above has one possible fix and another is to use `Object.getOwnPropertyNames`.

ashleyryan:
Following up from slack with, I think, a related issue.

When defining a property with a getter and setter, the `@property` directive ignores the `type` or `converters.toAttribute` options for boolean attributes/properties.

There's a stackblitz here: https://stackblitz.com/edit/clarity-react-experimental?file=package.json

With a simple property, rendering a component with `<MyElement myprop></MyElement>` renders `<my-element myprop></my-element>`. With a getter/setter,  it renders as `<my-element myprop="true"></my-element>` with a string value instead.

justinfagnani:
Fixed in https://github.com/lit/lit/pull/2800

PR #2800, opened 2022-04-28
[labs/react] Fix for #2799: add support for properties with custom accessors

lideen:
Hi! All new to this repo but tried to fix a bug that I found #2799

Any input appreciated 🎉


justinfagnani:
@lideen thanks for identifying this and submitting a fix and tests!

I made some comments on the implementation in the interest of speed here. I think the implementation can be collapsed into one `getInstancePropertiesOf()` function that recursively calls itself on a class's superclass, merges the superclass's properties into its own, then caches and returns the result. This will avoid materializing the prototype chain and processing prototypes multiple times. It should minimize memory allocations too.

I would like to also avoid materializing the property descriptors if possible. I think the core issue here is that accessors are not enumerable and `for...in` doesn't include non-enumerable properties. `Object.getOwnPropertyNames()` does return non-enumerable names (`Reflect.ownKeys()`) does too, but it's not supported on IE11 and we don't need Symbol-keyed properties in JSX). So I hope we can use that. It would include methods, but that might be ok.

lideen:
@justinfagnani Something like this? 1531750aedc7b4c8cbe4a39ad414fe301a496cc9

justinfagnani:
@lideen that what I was thinking at the time (though with memoizing the function).

But in the meantime I chatted with @josepharhar about it, and I think that we should match the Preact and upcoming React support for properties as much as possible. This would mean not pre-calculating the instance keys at all, but just using an `in` check on each given prop key. See https://github.com/facebook/react/blob/2c8a1452b82b9ec5ebfa3f370b31fda19610ae92/packages/react-dom/src/client/DOMPropertyOperations.js#L187-L194

lideen:
@justinfagnani I gave it a shot [79e6a4f](https://github.com/lit/lit/commit/79e6a4f793bd9cbb1dea58ca387498abe7f56844). I removed the warning about reserved props as I don't think it fits well now when the check is done on-demand 🤔 . Let me know if I should re-add it.

lideen:
@justinfagnani sorry to ping you, but is there anything I can do to move this along?

packages/labs/react/src/create-component.ts
```diff
@@ -257,7 +236,12 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
       for (const [k, v] of Object.entries(this.props)) {
         if (k === '__forwardedRef') continue;
 
-        if (elementClassProps.has(k)) {
+        if (
+          eventProps.has(k) ||
+          (!reservedReactProperties.has(k) &&
+            !(k in HTMLElement.prototype) &&
+            k in elementClass.prototype)
```
justinfagnani, line 243: We still need to verify this logic (specifically excluding `k in HTMLElement.prototype`) with @josepharhar to see if it matches React 19. I'm good with going in as-is now though.

packages/labs/react/src/create-component.ts
```diff
@@ -157,28 +158,6 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
...
-  const elementClassProps = new Set(Object.keys(events ?? {}));
-  for (const p in elementClass.prototype) {
-    if (!(p in HTMLElement.prototype)) {
-      if (reservedReactProperties.has(p)) {
-        // Note, this effectively warns only for `ref` since the other
-        // reserved props are on HTMLElement.prototype. To address this
-        // would require crawling down the prototype, which doesn't feel worth
-        // it since implementing these properties on an element is extremely
-        // rare.
-        console.warn(
```
justinfagnani, line 171: I think we should keep this warning. We can move this into a dev build in the future.

justinfagnani (changes requested):

justinfagnani:
@lideen this has increased in urgency for us, so I'm going to try to see if I can get this in asap. Did you enable changes to your PR branch? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

justinfagnani (approved):
Thanks @lideen !

I'll follow up on the comments I left with issues.

PR #2803, opened 2022-04-28
[labs/cli] Add stub command for gen react

kevinpschaaf:


packages/labs/cli/src/lib/commands/generate.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {AbsolutePath} from '@lit-labs/analyzer/lib/paths.js';
+import {Command} from '../command.js';
+
+export const generate: Command = {
+  name: 'gen',
+  description: 'Generate Framework Wrappers',
```
justinfagnani, line 12: I think this can be sentence case - lowercase the other words

packages/labs/cli/src/lib/commands/generate.ts
```diff
@@ -0,0 +1,27 @@
...
+  subcommands: [
+    {
+      name: 'react',
```
justinfagnani, line 15: I would take the frameworks are arguments, not sub commands.

packages/labs/cli/src/lib/commands/generate.ts
```diff
@@ -0,0 +1,27 @@
...
+      description: 'Generate React framework wrapper',
+      options: [{name: 'packageRoot', defaultValue: './'}],
+      async run({packageRoot}: {packageRoot: string}, console: Console) {
+        const generateReactWrappers = await import('../generate/react.js');
+        await generateReactWrappers.run(packageRoot as AbsolutePath, console);
+      },
+    },
+  ],
+  async run() {
+    throw new Error('must use extract or build');
```
justinfagnani, line 25: Update. It'd be nice to generate the help for the command in this case.

justinfagnani (changes requested):

packages/labs/cli/src/lib/commands/generate.ts
kevinpschaaf: Done

packages/labs/cli/src/lib/commands/generate.ts
kevinpschaaf: Done

packages/labs/cli/src/lib/commands/labs.ts
```diff
@@ -0,0 +1,42 @@
...
+export const labs: Command = {
+  name: 'labs',
+  description: 'Experimental commands',
+  subcommands: [
+    {
+      name: 'gen',
+      description: 'Generate framework wrappers',
+      options: [
+        {name: 'packageRoot', defaultValue: './'},
+        {name: 'frameworks', multiple: true, defaultValue: ['react']},
```
justinfagnani, line 19: I wouldn't make a default

packages/labs/cli/src/lib/commands/labs.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const labs: Command = {
+  name: 'labs',
+  description: 'Experimental commands',
+  subcommands: [
+    {
+      name: 'gen',
+      description: 'Generate framework wrappers',
+      options: [
+        {name: 'packageRoot', defaultValue: './'},
```
justinfagnani, line 18: Descriptions would be nice. Even just: "The package to generate wrappers for."

justinfagnani, line 18: I think this option could be `multiple: true`, and to make that just a little shorter, it could be named `package`: `lit labs gen react angular --package ./packages/el-one --package ./packages/el-two`

packages/labs/cli/src/lib/generate/generate.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type FrameworkName = keyof typeof frameworkGenerators;
+
+export const run = async (
+  {packageRoot, frameworks}: {packageRoot: string; frameworks: string[]},
+  console: Console
+) => {
+  const analyzer = new Analyzer(packageRoot as AbsolutePath);
+  const analysis = analyzer.analyzePackage();
+  await Promise.all(
```
justinfagnani, line 22: I think you want [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) here.

packages/labs/cli/src/lib/generate/generate.ts
```diff
@@ -0,0 +1,36 @@
...
+    (frameworks as FrameworkName[])
+      .map((framework) => {
+        const importer = frameworkGenerators[framework];
+        if (importer === undefined) {
+          throw new Error(`No generator exists for framework '${framework}'`);
```
justinfagnani, line 27: Do you want a synchronous error here?

packages/labs/cli/src/lib/generate/react.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {Package} from '@lit-labs/analyzer/lib/model.js';
+
+export const run = async (analysis: Package, console: Console) => {
```
justinfagnani, line 9: Ok for now, but I think we'll want a `@lit-labs/gen-wrapper-react` package very soon.

justinfagnani (approved):
Small comments - most could be TODOs.

packages/labs/cli/src/lib/commands/labs.ts
kevinpschaaf: Done.

packages/labs/cli/src/lib/commands/labs.ts
kevinpschaaf: Done

kevinpschaaf: Done

packages/labs/cli/src/lib/generate/generate.ts
kevinpschaaf: Done

packages/labs/cli/src/lib/generate/generate.ts
kevinpschaaf: Pulled validation out to be sync.

justinfagnani: Oh, it already was synchronous since it was building the array for Promise.all(), but I didn't know the intention, since it was next to an async map, which would have async errors.

Issue #2784, opened 2022-04-26
[analyzer] Analyze ReactiveElement/LitElement class properties and events


justinfagnani:
We need properties and events for the framework wrappers.

PR #2804, opened 2022-04-28
Read property options from decorated properties

justinfagnani:
Part of #2784


packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -80,11 +78,7 @@ export class Analyzer {
       if (ts.isClassDeclaration(statement)) {
         if (isLitElement(statement, this.checker)) {
```
rictic, line 79: Rather than repeating logic between a predicate and the getter, it may be simpler for `getLitElementDeclaration` to return undefined if it's not a LitElement

packages/labs/analyzer/src/lib/lit-element/decorators.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * TODO (justinfagnani): change to looking up decorators by known declarations.
+ */
+const isNamedDecoratorFactory = (
+  decorator: ts.Decorator,
+  name: string
+): decorator is CustomElementDecorator =>
+  ts.isCallExpression(decorator.expression) &&
+  ts.isIdentifier(decorator.expression.expression) &&
+  decorator.expression.expression.getText() === name;
```
rictic, line 27: [inherited from root PR] avoid getText

packages/labs/analyzer/src/lib/lit-element/decorators.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Gets the property options object from a `@property()` decorator callsite.
+ *
+ * Only works with an object literal passed as the first argument.
+ */
+export const getPropertyOptions = (decorator: PropertyDecorator) => {
+  const options = decorator.expression.arguments[0];
+  // console.log('getPropertyOptions', decorator.expression.arguments[0]?.getText());
```
rictic, line 65: debug code

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -11,6 +11,60 @@
...
+    ts.isPropertyDeclaration(m)
+  ) as unknown as ts.NodeArray<ts.PropertyDeclaration>;
+  for (const prop of propertyDeclarations) {
+    const name = prop.name.getText();
+    const type = checker.getTypeAtLocation(prop);
+
+    const propertyDecorator = getPropertyDecorator(prop);
+    if (propertyDecorator !== undefined) {
+      const options = getPropertyOptions(propertyDecorator);
+      // console.log('propertOptions', name, options !== undefined);
```
rictic, line 47: debug code

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -11,6 +11,60 @@
...
+  node: LitClassDeclaration,
+  checker: ts.TypeChecker
+): LitElementDeclaration => {
+  const reactiveProperties = new Map<string, ReactiveProperty>();
+
+  const propertyDeclarations = node.members.filter((m) =>
+    ts.isPropertyDeclaration(m)
+  ) as unknown as ts.NodeArray<ts.PropertyDeclaration>;
+  for (const prop of propertyDeclarations) {
+    const name = prop.name.getText();
```
rictic, line 41: avoid getText()

packages/labs/analyzer/test-files/decorators-properties/src/element-a.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitElement, html, css} from 'lit';
```
rictic, line 7: lint warnings

rictic (changes requested):

packages/labs/analyzer/src/lib/analyzer.ts
justinfagnani: I kind of like how this reads, and it doesn't actually do any duplicate work right now. I could be convinced though. Mind if we defer this? I'll need to change this a little bit to support ReactiveElement, then add a predicate to check for mixins, so this isn't nearly a final shape.

rictic: Yeah, it's a marginal suggestion, fine to defer

PR #2807, opened 2022-04-28
[@tests] roll concurrency down to 1

taylor-vann:
run 1 browser at a time through config

as opposed to run 1 browser at a time through commands: #2777 

@aomarks 
@justinfagnani

### DIFF
--- a/.github/workflows/sauce-ie11.yaml
+++ b/.github/workflows/sauce-ie11.yaml
@@ -14,7 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BROWSERS: preset:sauce-ie11
-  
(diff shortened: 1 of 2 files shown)


aomarks:
Perfect! Let's see if this passes :crossed_fingers:

taylor-vann:
there's always the possibility of a weak pass, so let's be a little cautious
from my experience last few sprints, this doesn't work long term (more than once)

aomarks:
> there's always the possibility of a weak pass, so let's be a little cautious
> from my experience last few sprints, this doesn't work long term (more than once)

Yeah totally. We can manually re-run this 3 or 4 or more times to get more confidence.

PR #2812, opened 2022-04-29
Read events from class JSDoc @fires tags

justinfagnani:
Part of #2784


packages/labs/analyzer/src/lib/lit-element/properties.ts
```diff
@@ -11,6 +11,44 @@
  */
 
 import ts from 'typescript';
+import {LitClassDeclaration} from '../lit-element.js';
+import {ReactiveProperty} from '../model.js';
+import {getPropertyDecorator, getPropertyOptions} from './decorators.js';
+
+export const getProperties = (
```
justinfagnani, line 18: This is just moved from lit-element.ts

packages/labs/analyzer/src/lib/lit-element/events.ts
```diff
@@ -0,0 +1,73 @@
...
+          const result = parseFiresTagComment(comment);
+          if (result === undefined) {
+            // TODO(justinfagnani): report syntax error?
+            continue;
+          }
+          const {name, type, description} = result;
+          // TODO(justinfagnani): how do we dereference the event type name
+          // into a TypeScript type? TypeScript will automatically do this for
+          // jsdoc tags it understands, but unfortunately @fires is not one of
+          // them.
```
kevinpschaaf, line 37: Can you (temporarily) create & insert an identifier with the type name into the AST where it would be in scope (presumably in the module), and then do a `checker.getTypeAtLocation(identifier);`?

packages/labs/analyzer/test-files/events/src/element-a.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitElement} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
```
kevinpschaaf, line 8: Might as well fix this (unused import)

kevinpschaaf (approved):
LGTM, small test file nit

packages/labs/analyzer/src/lib/lit-element/events.ts
justinfagnani: I'm sure something like that could work, though you might have to kick the checker to re-check the module. I hope there are some direct ways of evaluating a type expression within a scope. The built-in tags do this somehow.

PR #2815, opened 2022-04-29
[infra] Convert lit-html and internal-scripts to Wireit

aomarks:
Converts the `lit-html` and `internal-scripts` packages to [Wireit](https://github.com/google/wireit), and enables GitHub Actions caching.

### Notes

- I deleted the `dev.sh` script, because you can now write `npm test watch` instead. (Plus I think nobody was using `dev.sh` anyway).

- I removed the `test:watch` script, which ran the dev tests in watch mode, because you can now write `npm run test:dev watch` or `npm run test:prod watch` instead.

- We don't get nice colorized output for `wtr` because it doesn't detect TTY 😔. I'm thinking of submitting a PR to `@web/test-runner` to add a flag to force the pretty output even when TTY is not detected (similar to `--pretty` for `tsc`).
  - There are also some options we could think about to tweak the Wireit behavior here. https://github.com/google/wireit/issues/56 is one somewhat advanced option that gives us the most power, but an option that doesn't capture `stdout`/`stderr` for replay and instead uses the [`pipe`](https://nodejs.org/api/child_process.html#optionsstdio) spawn setting would also work. Need to think this through a little more.

- Noticed an npm install warning because of our top-level `engines` not being right, threw in a fix for that.

- We're not going to get caching of test results from this PR, only build output. Incremental testing for browser-based tests is going to require some additional refactoring, because we run all of those tests in a single giant `wtr` invocation in the `tests` package. We do that to share more Sauce Labs tunnels, so getting incremental browser testing may be blocked on finding a way to set up https://github.com/saucelabs/sauce-connect-action.


packages/lit-html/package.json
```diff
@@ -155,6 +152,127 @@
...
+      "command": "tsc --build --pretty",
+      "clean": "if-file-deleted",
+      "files": [
+        "src/**/*.ts",
+        "tsconfig.json"
+      ],
+      "output": [
+        "development",
+        "tsconfig.tsbuildinfo",
+        "tsconfig.polyfill-support.tsbuildinfo"
```
justinfagnani, line 173: Is this right? Doesn't it need to go with the command that uses `tsconfig.polyfill-support.json`? Which command is that, btw? This one because it's used as a project reference?

packages/lit-html/package.json
```diff
@@ -155,6 +152,127 @@
...
+      "files": [
+        "rollup.config.js",
+        "../../rollup-common.js"
+      ],
+      "output": []
+    },
+    "check-version": {
+      "command": "node scripts/check-version-tracker.js",
+      "dependencies": [
+        "build:ts"
```
justinfagnani, line 237: This script checks the source, so it doesn't need to depend on `build:ts`

packages/lit-html/package.json
```diff
@@ -155,6 +152,127 @@
...
+        "test:prod"
+      ]
+    },
+    "test:dev": {
+      "command": "cd ../tests && MODE=dev npx wtr \"../lit-html/development/**/*_test.(js|html)\"",
+      "dependencies": [
+        "build:ts"
+      ],
+      "files": [
+        "../tests/web-test-runner.config.js",
```
justinfagnani, line 258: Could a Wireit script hand only files so you could use it as a dependency to get this effect and be robust to the WTR config being split into more files?

packages/lit-html/package.json
```diff
@@ -155,6 +152,127 @@
...
+        "../tests/wtr-config.js"
+      ],
+      "output": []
+    },
+    "test:prod": {
+      "command": "cd ../tests && MODE=prod npx wtr \"../lit-html/development/**/*_test.(js|html)\" \"../lit-html/src/test/version-stability_test.js\"",
+      "dependencies": [
+        "build:rollup",
```
justinfagnani, line 266: tbh, the implicit use of dependencies files messes with me a bit when reviewing. `build:rollup` doesn't build the test sources directly, but it depends on `build:ts` which does. I think I'd like to see a direct dependency here on `build:ts` to make that more clear.

packages/lit-html/package.json
aomarks: Yes. Added a `"#comment"`.

I considered breaking this out into a separate build step, but they're sort of tangled up because they read and write to the same places.

packages/lit-html/package.json
aomarks: Good call, done.

packages/lit-html/package.json
aomarks: Yes, actually in my prototype I did this :) https://github.com/aomarks/wireit-lit-fork/blob/1782c67d2a123106a5f73bbc46c962221e151121/packages/tests/package.json#L41

However, I forgot about this and implemented a validation check that requires all scripts to have either a `command` or `dependencies`: https://github.com/google/wireit/blob/7cf075e186de9bc06e5f7e0da3e6d39fdd50daba/src/analyzer.ts#L215

Added a `#comment`.

(Also realized I forgot to include `rollup-resolve-remap.js` there too).

packages/lit-html/package.json
aomarks: Hmm, I guess I can see that, but OTOH it's genuinely testing the direct output of `build:rollup`, and not at all testing the direct output of `build:ts` -- so couldn't you make the case that the more minimal style is making that fact a little more apparent?

aomarks: Oh, actually I just re-read your comment and realized you were talking about the *test* sources. That's a good argument, done.

justinfagnani (approved):
🎉

packages/internal-scripts/package.json
```diff
@@ -15,9 +15,26 @@
...
-    "clean": "rm -rf lib/"
+    "build": "wireit",
+    "build:ts": "wireit"
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:ts"
+      ]
+    },
```
augustjk, line 26: Would this do anything different than just running `build:ts`? And is `npm run clean` no longer necessary since `"if-file-delete"` should cover all cases we'd actually need to clean, since otherwise outputs would get overwritten?

packages/lit-html/package.json
```diff
@@ -127,20 +127,17 @@
...
   "scripts": {
-    "build": "npm run clean && npm run build:ts && npm run build:version-stability-test && rollup -c",
-    "build:watch": "rollup -c --watch",
-    "build:ts": "tsc --build && treemirror development . '**/*.d.ts{,.map}'",
-    "build:ts:watch": "tsc --build --watch",
-    "build:version-stability-test": "rollup -c rollup-version-stability-test.config.js",
-    "check-version": "node scripts/check-version-tracker.js",
-    "checksize": "rollup -c --environment=CHECKSIZE",
-    "clean": "rm -rf {async-directive,directive-helpers,directive,experimental-hydrate,lit-html,polyfill-support,private-ssr-support,static}.{js,js.map,d.ts} test/ directives/ development/ version-stability-test/ *.tsbuildinfo",
-    "dev": "scripts/dev.sh",
```
augustjk, line 138: Would this now be equivalent to `npm test watch`?

augustjk (approved):
Nice! Just a couple questions for my curiosity's sake.

packages/internal-scripts/package.json
rictic: It's the same as running build:ts directly, yes

And yes, `if-file-delete` combined with `tsc --build` should behave the same as always running clean before building

packages/lit-html/package.json
```diff
@@ -155,6 +152,129 @@
     "!/development/test/",
     "/directives/"
   ],
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:rollup",
```
rictic, line 158: Should this include "build:ts"? It's transitively included, but it seems better to be explicit about it

rictic (approved):

packages/lit-html/package.json
aomarks: Yep!

packages/lit-html/package.json
aomarks: Done.

PR #2820, opened 2022-05-02
[@tests] add tunnel identifier

taylor-vann:
add a `tunnel id`
the goal is to spin up a high velocity / capacity tunnel at saucelabs

### DIFF
--- a/.github/workflows/sauce-chrome.yaml
+++ b/.github/workflows/sauce-chrome.yaml
@@ -13,10 +13,8 @@ jobs:
 
     runs-on: ubuntu-latest
     env:
-      BROWSERS: preset:sauce
-      CONCURRENT_BROWSERS: 3
+      CONCURRENT_BROWSERS: 1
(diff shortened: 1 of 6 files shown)


justinfagnani:
> the goal is to spin up a high velocity / capacity tunnel at saucelabs

Our understanding of WTR is that it creates a new Sauce tunnel for each wtr invocation, yes? How does this help put us in high-available mode then? WTR will still be creating new tunnels until https://github.com/modernweb-dev/web/issues/1932 is addressed, no?

taylor-vann:
@justinfagnani laptop just died, will change sync time for this

taylor-vann:
after the sync, was looking over the documentation. Will be making a separate PR to explore the saucelabs github action.
This PR targets tunnels based on browsers now.
The hope is that a single vm can share a tunnel for multiple tests of the same browser, regardless of commit, branch, or pr

PR #2821, opened 2022-05-02
[@labs/task] add render return value

taylor-vann:
Return value is requested to be typed

The types could be inferred from a `StatusRender`
```
  render<SR extends StatusRenderer<R>>(renderer: SR) {
    switch (this.status) {
      case TaskStatus.INITIAL:
        return renderer.initial?.();
      case TaskStatus.PENDING:
        return renderer.pending?.();
      case TaskStatus.COMPLETE:
        return renderer.complete?.(this.value!);
      case TaskStatus.ERROR:
      default:
        // exhaustiveness check
        this.status as void;
    }
  }
```

Related to: #2301


packages/labs/task/src/test/task_test.ts
```diff
@@ -67,11 +67,11 @@ suite('Task', () => {
       override update(changedProperties: PropertyValues): void {
         super.update(changedProperties);
         this.taskValue = this.task.value ?? this.task.error;
-        this.task.render({
+        this.task.render<string, string>({
```
justinfagnani, line 70: Ideally this could all be inferred from both the task definition and the callbacks. What's preventing that now?

taylor-vann: just talked with steve for a bit! and if i understand correctly, i agree :)

uploaded a change that infers types from the `StatusRenderer` provided by the user

taylor-vann:
The user can explicitly declare typings for return type and error type:
Screen Shot 2022-05-02 at 2 29 46 PM

Those are combined with Data Return types
Screen Shot 2022-05-02 at 2 30 09 PM

taylor-vann:
if a user doesn't provide types:
Screen Shot 2022-05-02 at 2 43 53 PM
TS will do its best to infer return types from the provided `StatusRender`
Screen Shot 2022-05-02 at 2 44 12 PM

taylor-vann:
this doesn't quite cover all use cases in #2301
I want to see if i can infer the return type based on `status`

taylor-vann:
The current code is a little gross. Took some liberties to get a functioning concept.
Screen Shot 2022-05-04 at 11 33 08 AM

taylor-vann:
related to: #2301

packages/labs/task/src/task.ts
```diff
@@ -233,7 +234,9 @@ export class Task<T extends [...unknown[]] = any, R = any> {
     return this._error;
   }
 
-  render(renderer: StatusRenderer<R>) {
+  render<SR extends StatusRenderer<R, E>>(
+    renderer: Partial<SR>
+  ): ReturnType<SR[this['status']]> | undefined {
```
mohe2015, line 239: I think this is suboptimal but I think I didn't manage better with type assertions. If I understand correctly this means that `undefined` is always a possible return value even when all 4 functions are defined and don't return `undefined`

mohe2015: Don't understand though why in your screenshot above this is not the case.

sorvell: Why do we need the `| undefined`?

packages/labs/task/src/task.ts
```diff
@@ -33,11 +33,12 @@ export const initialState = Symbol();
 
 export type TaskStatus = typeof TaskStatus[keyof typeof TaskStatus];
 
-export type StatusRenderer<R> = {
-  initial?: () => unknown;
-  pending?: () => unknown;
-  complete?: (value: R) => unknown;
-  error?: (error: unknown) => unknown;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type StatusRenderer<R, E, I = any, P = any, C = any, ER = any> = {
```
sorvell, line 37: Why do we need types for I, P, C, ER. Isn't the idea that we're inferring them?

packages/labs/task/src/task.ts
taylor-vann: A partial record of callbacks will always return `| undefined` as a possible type. If it were a partial record of numbers it'd be the same: `number | undefined`

packages/labs/task/src/task.ts
taylor-vann: The `i, P, C, ER` is to hint to the TS compiler that these values should be infered.
The default value of `any` is so a type inference is not made at first, they could be `any` type.
When return values for `StatusRenderer` have a value of `unknown` then an `unknown` type is cast and anything that extends `StatusRenderer` will also have an `unknown` return type

taylor-vann: Heres a small demo of that behavior in action.

In the demo below, the function can't infer a type from values set as `unknown`. However in can infer types on values with a default of `any`
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKogNYgPYHcQBC2AHsgN4BQy1yAFhADYPYBcyArpjvgNxU208AEwCebTljwg+AXwoVQkWIhQB1RgwDS3QiXL9q9Jq2QAiI81MAaA3WHs254SOsU5FGJwRhg2EMmYAcyJiAB51JgARCBhQCCEQ5AhiSBAhAGc0LikQgD4ACgAjEjYIhmjYkHiQgEp9GmQoCDB2KH9i4gA6QVxRN3kEP3SwZA6ARjZhqFBA5ABeAOxgknyyCxNzDWxrO16xMx7RUxka+UVoeCRkAEEQEQAVEQAHapJQ+-nkODurZAAlT7fES5eoCLZse58BqHfZ-WRncAXFTIMraHJ6SjQ8EHLauaH2Rww1zuTwgby+fxBW4iELhDQVOIJPTJVIZG53R4vJnEAodUr0mKM2qg6hNFptUYkbpbeEUQYgYaS4gAJkmYGmIFmCypdxCq3WhNxvxhhOcxxqfHlio6AGY1RqtYtAtS9WtsZtjDsTQczScgA

taylor-vann: Here's an example of the slight difference between any and unknown when type inferencing.
Extending an interface with an `unknown` type actually casts a type of `unknown` and it needs to be reduced later somehow.

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKogNYgPYHcQBC2AHgDwAqyAvMgK6Y74A0yAStXQ3iAHzIDeAKGQjkACwgAbSdgBcyABQBKan3IBuYaLF4AJgE95y1W00BfQYNCRYiFAHUpkgNKNCJAVpETpcxSqo+ACIfGSCmL3E9Q38TEOjwwQtBGHoEMGBsEGQZAHMiMkdpABEIGFAIXQLkCGJIEF0AZzQufAKeBQAjEnkiyVLykEqCpXlWCDBaKBByfQAHCFI+gYqqkgBtEKdsIIBdPiFRZCgJqezu4gA6HVwDZXNLBCzGsGQLgEYOPIKFflC-YyBZBbXzhKK3GKA4I3AxBMxKSzWaDwJDIACCIH0swWazIlBocExLHYBMxB0i-yMATUmiOMMh1NMSUR4GRdmQAEkQDBoMMPIdtNsqXF-ok6dF5PEIYlkqkQOlMtk8hj9AUlk4VkNcTU6hAGs0Vdi+cQOhdehqyqsRmNTtMjeqSpatQVNqL9p4jidJtM3iRrtt7szBE8QC9fcQAEzyF5QUC5L7YXIqn5-IWxIEgsIsenCjP0uFKTQhsMXADM0bAsZA8ZoysxKcp6eh2zBOabwPz8KAA

taylor-vann: @justinfagnani maybe you might know of a more ergonomic way to pull types out of an interface supplied by a client?

packages/labs/task/src/task.ts
taylor-vann: And there we go, we don't need it. Updated.

packages/labs/task/src/task.ts
```diff
@@ -19,10 +19,10 @@ export {ArgsFunction as DepsFunction};
  * States for task status
  */
 export const TaskStatus = {
-  INITIAL: 0,
-  PENDING: 1,
-  COMPLETE: 2,
-  ERROR: 3,
+  INITIAL: 'initial',
```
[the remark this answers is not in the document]
taylor-vann: sounds good! done!

taylor-vann:
removed use of type inference with `any` in favor of casting from the client interface.

packages/labs/task/src/task.ts
taylor-vann: opted to use casting over any for type inference.

packages/labs/task/src/task.ts
```diff
@@ -99,14 +96,14 @@ export interface TaskConfig<T extends unknown[], R> {
  * }
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export class Task<T extends [...unknown[]] = any, R = any> {
+export class Task<T extends unknown[] = any, R = any, E = any> {
```
e111077, line 99: Since you're casting all instances of `E` to `as E`, do you even need `E` on `Task`? can't `_error` simply just be `Error` and `StatusRenderer<R, E>` be `StatusRenderer<R, E extends Error = Error>`?

e111077 (changes requested):

justinfagnani:
Superseded by https://github.com/lit/lit/pull/4008

PR #2824, opened 2022-05-03
[@lit-labs/context]: fix(#2839): lower target version

### LINKED ISSUE #2839: [@lit-labs/context]: unexpected token
# Description

When running `@lib-labs/context` in an environment that does not support the optional-chaining operator, we get this error: 

```
Error parsing code: Unexpected token (51:41)
```

in the file `decorators/context-provider.js`.

# Potential solutions

- Lower the target version that the lib is transpiled to. Most of the other `lit` library target `es2019`, while  `@lib-labs/context` targets `es2021`.

belzile:
Fixes #2839

I tried using the context API.

One problem I met is that there are optional chaining operators in the code (?.) which fails at runtime in environments that do not support it.

Since lit is transpiled to es2019, I lowered `@lit-labs/context` to the same version. This ensures tha the dependency won't require additional transpiling step.

### DIFF
--- a/.changeset/famous-dots-dance.md
+++ b/.changeset/famous-dots-dance.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/context': patch
+---
+
+lower target js version
--- a/packages/labs/context/tsconfig.json
+++ b/packages/labs/context/tsconfig.json
@@ -3,7 +3,7 @@
     "composite": true,
     "target": "es2019",
     "module": "es2015",
-    "lib": ["es2020", "DOM", "DOM.Iterable"],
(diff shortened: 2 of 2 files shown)


.changeset/famous-dots-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/context': patch
+---
+
+lower target js version
```
justinfagnani, line 5: These show up in the generated changelogs, so usually we want a little nicer presentation with capitalization and some details that make it understandable in that context like "TypeScript" and "JavaScript version". I'll take care of this post merging though!

justinfagnani (approved):
Thanks for the fix @belzile !!

PR #2827, opened 2022-05-03
chore: logic clearing and improve efficency

metasong:
if just force, no need to check value change.

### DIFF
--- a/packages/labs/context/src/lib/value-notifier.ts
+++ b/packages/labs/context/src/lib/value-notifier.ts
@@ -31,12 +31,9 @@ export class ValueNotifier<T> {
   }
 
   public setValue(v: T, force = false) {
-    let changed = false;
-    if (!Object.is(v, this._value)) {
-      changed = true;
-    }
     this._value = v;
-    if (changed || force) {
+
+    if(force || !Object.is(v, this._value)) {
       this.updateObservers();
     }
   }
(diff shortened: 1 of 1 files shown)


rictic (approved):
Looks equivalent and simpler, thanks for the change!

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -31,12 +31,9 @@ export class ValueNotifier<T> {
...
 
   public setValue(v: T, force = false) {
-    let changed = false;
-    if (!Object.is(v, this._value)) {
-      changed = true;
-    }
     this._value = v;
-    if (changed || force) {
+
+    if(force || !Object.is(v, this._value)) {
```
justinfagnani, line 36: `!Object.is(v, this._value)` will always return `true` here (except when `v` is `NaN`) because of the previous line.

I think the mostly simplification we could get is:
```ts
    const update = force || !Object.is(v, this._value);
    this._value = v;
    if (update) {
```

justinfagnani (changes requested):

metasong (commented):
oh, thanks for pointing out the problem

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {ContextCallback} from './context-request-event.js';
+import { ContextCallback } from './context-request-event.js';
```
justinfagnani, line 7: This should be failing our linter, but it's not...

PR #2828, opened 2022-05-03
Remove private Lit 2 migration helpers

AndrewJakubowicz:
### Context

While migrating internal to Lit 2, we had these additional options to change how the `render` function behaved. However these are no longer required as they are not referenced internally.

### Testing

Tested internally with a TGP using the same change, and checking for all references of `INTERNAL` and `clearContainerForLit2MigrationOnly`. Manually checked that copy.bara wouldn't need changes.

I also removed a unit test that was testing the `INTERNAL` and `clearContainerForLit2MigrationOnly` option behavior.

### DIFF
--- a/packages/lit-html/src/test/lit-html_test.ts
+++ b/packages/lit-html/src/test/lit-html_test.ts
@@ -49,10 +49,6 @@ const isIe = ua.indexOf('Trident/') > 0;
 // We don't have direct access to DEV_MODE, but this is a good enough
 // proxy.
 const DEV_MODE = render.setSanitizer != null;
-/**
- * litHtmlLib.INTERNAL is not exported in prod mode
- */
-const INTERNAL = litHtmlLib.INTERNAL === true;
 
 class FireEventDirective extends Directive {
   render() {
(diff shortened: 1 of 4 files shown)


rictic (approved):
Awesome work!

packages/lit-html/src/test/lit-html_test.ts
```diff
@@ -3195,9 +3190,7 @@ suite('lit-html', () => {
       } as RenderOptions);
       assert.equal(
         stripExpressionComments(container.innerHTML),
-        INTERNAL
-          ? `<p>HELLO</p>${remainingHtml}`
-          : `${clearedHtml}<p>HELLO</p>${remainingHtml}`
+        `${clearedHtml}<p>HELLO</p>${remainingHtml}`
```
justinfagnani, line 3193: You can remove this whole test now.

justinfagnani (changes requested):

packages/lit-html/src/test/lit-html_test.ts
AndrewJakubowicz: Done! Good point!

justinfagnani (approved):

PR #2835, opened 2022-05-05
[@task] remove overloaded constructor

taylor-vann:
Currently, we overload the constructor of `task`.

However, Typescript is smart enough and the overloads are similar enough that typescript can do the work of the three overloads itself.

As in this constructor:
```ts
  constructor(
    host: ReactiveControllerHost,
    task: TaskFunction<T, R> | TaskConfig<T, R>,
    args?: ArgsFunction<T>
  ) {
```
Is categorically the same as these three constructors:
```ts
  constructor(
    host: ReactiveControllerHost,
    task: TaskFunction<T, R>,
    args?: ArgsFunction<T>
  );
  constructor(host: ReactiveControllerHost, task: TaskConfig<T, R>);
  constructor(
    host: ReactiveControllerHost,
    task: TaskFunction<T, R> | TaskConfig<T, R>,
    args?: ArgsFunction<T>
  ) {
```

The argument types between overloads are not different enough for overloading a constructor. But in the future, if completely different arg names / types are needed we can add more constructors! <3


packages/labs/task/src/task.ts
```diff
@@ -125,12 +125,6 @@ export class Task<T extends [...unknown[]] = any, R = any> {
   private _resolveTaskComplete!: (value: R) => void;
   private _rejectTaskComplete!: (e: unknown) => void;
 
-  constructor(
```
sorvell, line 128: Is the idea that the overrides don't actually do anything so we're deleting them? Maybe update the PR comment to make that more clear.

taylor-vann: yes exactly. The overloads are similar enough that typescript can do the work of the three overloads itself.

But in the future if there was a more different constructor, like different args param names, we'd need an other constructor.

also sounds good!

taylor-vann: done!

taylor-vann (commented):
addressed review changes :)

.changeset/wild-ligers-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': patch
```
sorvell, line 2: If this is actually a no-op change, it shouldn't need a release, right?

taylor-vann: yeup, you're correct. sounds good!

.changeset/wild-ligers-cross.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+remove constructor overloads
```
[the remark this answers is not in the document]
taylor-vann: er ... empty changesets are empty. my bad, will be more explicit in the future though!

taylor-vann (commented):
feedback addressed!

.changeset/wild-ligers-cross.md
taylor-vann: i think we commented in the same cache frame lol

taylor-vann:
@AndrewJakubowicz @sorvell 
Added a more explicit patch changeset :)
Should be good to go, ready for another review

justinfagnani (changes requested):
This removes information from the constructor types that points to the valid usage: that it's not valid to pass both a TaskConfig and an args parameter.

taylor-vann:
@justinfagnani If you could clarify a little bit, i don't understand. This change is categorically the same as the overloaded constructor

taylor-vann:
Ah if i understand correctly then you want an overloaded constructor, despite the types being the same it's the use case that's different. If that's the case I'll close this PR and move on.

justinfagnani:
Yeah, one of the things with TypeScript overloads is that the non-overload signature _must_ be a superset of all of the overloads, but it may not express the constraints between the parameters that an overload can.

Class example is:
```ts
function foo(a: X, b: Y): void;
function foo(a: Z, b: W): void;
function foo(a: X | Z, b: Y | W): void {...}
```

The type `function foo(a: X | Z, b: Y | W): void` doesn't express that when `a` is an `X`, `b` must be a `Y`.

PR #2836, opened 2022-05-05
[@task] allow initial value

taylor-vann:
Currently the `initial value` of `Task` is assumed to be `undefined`.
This is the default behavior. When a Task is "re-run", `value` is set as `undefined`.

This PR adds an `initialValue` to the `TaskConfig` interface.
When no `intialValue` is set, the behavior described above is preserved.

When `initialValue` is set, the value is only updated when the task completes. As in, the initial value will not be set in future `PENDING` states.


taylor-vann:
related to: #2367

packages/labs/task/src/task.ts
```diff
@@ -105,6 +106,7 @@ export class Task<T extends [...unknown[]] = any, R = any> {
   private _getArgs?: ArgsFunction<T>;
   private _callId = 0;
   private _host: ReactiveControllerHost;
+  private _intialValue?: R;
```
mohe2015, line 109: spelling mistake

packages/labs/task/src/task.ts
```diff
@@ -34,16 +34,17 @@ export const initialState = Symbol();
 export type TaskStatus = typeof TaskStatus[keyof typeof TaskStatus];
 
 export type StatusRenderer<R> = {
-  initial?: () => unknown;
-  pending?: () => unknown;
-  complete?: (value: R) => unknown;
+  initial?: (value?: R) => unknown;
+  pending?: (value?: R) => unknown;
+  complete?: (value?: R) => unknown;
```
mohe2015, line 39: I don't really like that the `complete` value is now optional.

mohe2015:
What is a concrete use case for this? I don't fully understand that yet.

packages/labs/task/src/task.ts
```diff
@@ -193,8 +197,8 @@ export class Task<T extends [...unknown[]] = any, R = any> {
       });
     }
     this.status = TaskStatus.PENDING;
+    this._value = this._intialValue;
```
sorvell, line 200: I thought we decided to the value stick around and not reset it at all?

packages/labs/task/src/task.ts
sorvell: Is making `value` in `complete` optional related to allowing an initial value? If not, let's not change that.

sorvell (changes requested):

packages/labs/task/src/task.ts
taylor-vann: thank you! <3

packages/labs/task/src/task.ts
taylor-vann: updated to not reset at all

packages/labs/task/src/task.ts
taylor-vann: updated

taylor-vann:
@mohe2015 one potential use-case is lazy loading / lazy execution of a task, providing data for the initial render but before a remote request. In a compsci sense this would provide a kind of `Null Object` for tasks that correlate to an initial render.

As in, user loads page. Component needs a visual "stencil" up until the asynchronous task is requested / performed / completed.

sorvell (changes requested):
Please add a test where `initialValue` is set.

taylor-vann:
@sorvell added a test for initial values :) ready for another review

AndrewJakubowicz (changes requested):
Please remove all the `.wireit` directory files.

They should have been ignored by `.gitignore`?

taylor-vann:
thanks @AndrewJakubowicz, theyve been removed.

sorvell (changes requested):
Almost there, just a couple of points to address, see comments. Let's also make sure to update the PR description to say that the value is only updated when the task completes (currently, it incorrectly says that it's reset to the initialValue).

.changeset/fifty-games-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': patch
+---
+
+allow initial value
```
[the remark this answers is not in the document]
taylor-vann: yes! sorry, @AndrewJakubowicz reminded me and i forgot to change this PR as well, thank youu!

.changeset/fifty-games-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': patch
```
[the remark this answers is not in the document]
taylor-vann: the change felt small but you're right, this is a feature addition. sounds good to me!

taylor-vann:
Sweet! thanks yall
Andrew:
- got rid of the `.wireit` files
- left a more descriptive changelog
Steve:
- updated the description of the PR to correctly reflect behavior
- added a test to test the described behavior: only set `initialValue` on first `PENDING` state, not after.

sorvell:
> `When a Task is "re-run", value is set as undefined.`

This is not true is it?

.changeset/lucky-apples-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lit-labs/task': minor
+---
+
+Added `initialValue` parameter to `TaskConfig`.
+Task can start with an initial value from TaskConfig.
```
sorvell, line 6: We are changing the behavior of how value is treated when the task is re-run. This should be called out specifically here.

packages/labs/task/src/task.ts
```diff
@@ -194,7 +196,6 @@ export class Task<T extends [...unknown[]] = any, R = any> {
     }
     this.status = TaskStatus.PENDING;
     this._error = undefined;
-    this._value = undefined;
```
sorvell, line 197: This is the significant change that I want to make sure we're documenting and testing.

packages/labs/task/src/test/task_test.ts
```diff
@@ -113,6 +113,17 @@ suite('Task', () => {
     assert.equal(el.taskValue, undefined);
   });
 
+  test('tasks start with initialValues in the first PENDING state', async () => {
```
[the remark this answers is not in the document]
sorvell: We want specifically a test that validates that the value is *not* reset when the task is re-run (before it completes). This has nothing to do with initial value, but the behavior is being changed with this PR.

sorvell (changes requested):

packages/labs/task/src/task.ts
justinfagnani: First I think we need to decide that this is the behavior we want. Is it required to add an initial value? Should it be controlled with an option? Let's chat.

e111077 (approved):
Technically this looks fine to me. Code LGTM

I'm also in favor of this behavior – Task has the tendency to flicker. Would be cool if `value` is the way to make it not

justinfagnani:
Superseded by https://github.com/lit/lit/pull/4001

PR #2837, opened 2022-05-05
Choose open WTR ports better

justinfagnani:
This took a while to figure out, but I think it's race-free now.


packages/tests/run-web-tests.js
```diff
@@ -0,0 +1,93 @@
+/**
```
rictic, line 1: Lint warnings in this file for every use of `process`

packages/tests/run-web-tests.js
```diff
@@ -0,0 +1,93 @@
...
+let currentPortIndex = 0;
+let currentRunner;
+
+async function startWithNextAvailablePort() {
+  if (currentPortIndex === SAUCE_PORTS.length) {
+    throw new Error(
+      `No available ports. Ports tried: ${JSON.stringify(SAUCE_PORTS)}`
+    );
+  }
+  const port = SAUCE_PORTS[currentPortIndex++];
```
rictic, line 78: Inline `x++` is harder to reason about, mutate as its own statement.

packages/tests/run-web-tests.js
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+import {startTestRunner} from '@web/test-runner';
+import config from './web-test-runner.config.js';
+
+// eslint-disable-next-line no-undef
+process.on('uncaughtException', (error) => {
+  console.log('error.code', error.code);
+  if (error.code === 'EADDRINUSE') {
+    if (currentPortIndex === SAUCE_PORTS.length) {
```
rictic, line 20: We also check this inside startWithNextAvailablePort

packages/tests/run-web-tests.js
```diff
@@ -0,0 +1,93 @@
...
+  8000, 8001, 8003, 8031, 8080, 8081, 8443, 8765, 8777, 8888,
+  9000, 9001, 9031, 9080, 9081, 9090, 9191, 9876, 9877, 9999,
+  49221, 55001
+];
+
+let currentPortIndex = 0;
+let currentRunner;
+
+async function startWithNextAvailablePort() {
+  if (currentPortIndex === SAUCE_PORTS.length) {
```
rictic, line 73: Shouldn't we check this after incrementing currentPortIndex?

rictic (changes requested):
Looks like there's an issue where Safari can't import lit-element_styling_apply_test.js?

https://github.com/lit/lit/runs/6312589269?check_suite_focus=true

```
../lit-element/development/test/lit-element_styling_apply_test.js:

 ❌ Could not import your test module. Check the browser logs or open the browser in debug mode for more information.  (failed on macOS 10.15 Safari latest)
```

packages/tests/run-web-tests.js
AndrewJakubowicz: Oh yeah. This can be removed now that there is an `if` check on `EADDRINUSE` that won't match on the error thrown in the function. Nice comment!

justinfagnani:
@rictic Safari is definitely flaky around importing modules. I don't think it's related to this PR since the effect we would see from using a bad port for Sauce is that the browser can't load the tests at all. We've seen the module flakiness before, usually on `lit-element-styles_test.js` though.

After replacing all calls to wtr with the new script, we just passed 3/4 of the Sauce jobs, including both IE jobs. I can rerun this a few times. It might already be an improvement.

rictic (approved):

packages/tests/run-web-tests.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
AndrewJakubowicz, line 3: nit: 2022

AndrewJakubowicz (approved):
Tested locally!

aomarks (commented):
Quite a (re)discovery, this one!

We should upstream this to wtr sometime.

Issue #2833, opened 2022-05-05
[@lit-labs/router] API for a fallback route


AndrewJakubowicz:
Currently dynamically adding a route while preserving the final fallback route is cumbersome. See the [test for an example](https://github.com/lit/lit/blob/main/packages/labs/router/src/test/router_test_code.ts#L41-L44).

Instead of:

```ts
routes.splice(routes.length - 1, 0, {
  path: '/server-route',
  render: () => html`<h2>Server</h2>`,
});
```

it would be preferable to write:

```ts
routes.push({
  path: '/server-route',
  render: () => html`<h2>Server</h2>`,
});
```

This issue tracks adding a way to define a fallback route such that dynamically pushing routes can be done with `routes.push(...)`. Or an alternative.

PR #2838, opened 2022-05-05
[@lit-labs/router] API for a fallback route

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/2833

### Context

For dynamically adding routes, we previously needed to push the new route prior to the last one (if the last one is a fallback route).
To make pushing routes dynamically easier, we can instead pass and use a `fallbackRoute` option.

### Testing

Tested by slightly expanding the existing tests. I mostly added assertions to ensure that the dynamic route is added correctly, and that the fallback route is reachable.

### DIFF
--- a/.changeset/fluffy-zebras-sniff.md
+++ b/.changeset/fluffy-zebras-sniff.md
@@ -2,4 +2,6 @@
 '@lit-labs/router': patch
 ---
 
-Add an optional `fallbackRoute` param to the Routes and Router class.
+Add `fallback` route option to the Routes and Router class. The fallback route
+will always be matched if none of the `routes` match, and implicitly matches to
+the path `/*`.
--- a/.changeset/fluffy-zebras-sniff.md
+++ b/.changeset/fluffy-zebras-sniff.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/router': patch
+---
+
+Add an optional `fallbackRoute` param to the Routes and Router class.
--- a/packages/labs/router/src/routes.ts
+++ b/packages/labs/router/src/routes.ts
@@ -125,11 +125,11 @@ export class Routes implements ReactiveController {
   constructor(
     host: ReactiveControllerHost & HTMLElement,
     routes: Array<RouteConfig>,
-    fallback?: BaseRouteConfig
+    options?: {fallback?: BaseRouteConfig}
   ) {
     (this._host = host).addController(this);
     this.routes = [...routes];
-    this.fallback = fallback;
+    this.fallback = options?.fallback;
   }
 
   /**
--- a/packages/labs/router/src/routes.ts
+++ b/packages/labs/router/src/routes.ts
@@ -88,7 +88,7 @@ export class Routes implements ReactiveController {
   /**
    * A fallback route that is always matched after {@link routes}.
    */
-  readonly fallbackRoute?: RouteConfig;
+  private readonly fallbackRoute?: RouteConfig;
 
   /*
    * The current set of child Routes controllers. These are connected via
--- a/packages/labs/router/src/routes.ts
+++ b/packages/labs/router/src/routes.ts
@@ -85,6 +85,11 @@ export class Routes implements ReactiveController {
    */
   routes: Array<RouteConfig> = [];
 
+  /**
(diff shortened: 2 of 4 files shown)


packages/labs/router/src/routes.ts
```diff
@@ -85,6 +85,11 @@ export class Routes implements ReactiveController {
    */
   routes: Array<RouteConfig> = [];
 
+  /**
+   * A fallback route that is always matched after {@link routes}.
+   */
+  private readonly fallbackRoute?: RouteConfig;
```
justinfagnani, line 91: I think this should have the same visibility as `routes`

packages/labs/router/src/routes.ts
```diff
@@ -118,10 +123,12 @@ export class Routes implements ReactiveController {
 
   constructor(
     host: ReactiveControllerHost & HTMLElement,
-    routes: Array<RouteConfig>
+    routes: Array<RouteConfig>,
+    fallbackRoute?: RouteConfig
```
justinfagnani, line 127: Maybe just `fallback` for brevity? Let's ask some of the team.

packages/labs/router/src/routes.ts
```diff
@@ -215,8 +223,17 @@ export class Routes implements ReactiveController {
...
-  private _getRoute(pathname: string) {
-    return this.routes.find((r) => getPattern(r).test({pathname: pathname}));
+  private _getRoute(pathname: string): RouteConfig | undefined {
+    const matchedRoute = this.routes.find((r) =>
+      getPattern(r).test({pathname: pathname})
+    );
+    if (matchedRoute || this.fallbackRoute === undefined) {
+      return matchedRoute;
+    }
+    if (getPattern(this.fallbackRoute).test({pathname: pathname})) {
```
justinfagnani, line 233: I think the fallback shouldn't be able to have a pattern.

justinfagnani (changes requested):

packages/labs/router/src/routes.ts
```diff
@@ -118,10 +124,12 @@ export class Routes implements ReactiveController {
 
   constructor(
     host: ReactiveControllerHost & HTMLElement,
-    routes: Array<RouteConfig>
+    routes: Array<RouteConfig>,
+    fallback?: BaseRouteConfig
```
justinfagnani, line 128: I think we should put `fallback` in an options bag. If we were to add another param, users would have to pass `undefined` for fallback, which is ungainly.

justinfagnani (changes requested):

packages/labs/router/src/routes.ts
AndrewJakubowicz: Clarifying - API like so:

```ts
constructor(
    host: ReactiveControllerHost & HTMLElement,
    routes: Array<RouteConfig>,
    options?: {fallback?: BaseRouteConfig}
)
```

justinfagnani: Yep!

packages/labs/router/src/routes.ts
AndrewJakubowicz: Done!

packages/labs/router/src/routes.ts
AndrewJakubowicz: Done! Thanks!

packages/labs/router/src/routes.ts
AndrewJakubowicz: Done!

.changeset/fluffy-zebras-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/router': patch
+---
+
+Add an optional `fallbackRoute` param to the Routes and Router class.
```
justinfagnani, line 5: Update to describe the new name and options bag.

justinfagnani (approved):

packages/labs/router/src/routes.ts
AndrewJakubowicz: `fallback` seems good.

.changeset/fluffy-zebras-sniff.md
AndrewJakubowicz: Done

PR #2840, opened 2022-05-06
[labs/react] Convert labs/react scripts to wireit

justinfagnani:
Ok to review, but not merge. There are some problems highlighted here:

1. I needed to pass the WTR config in the test command. I have no idea why I didn't need this with #2837 
2. I had to make test:dev depend on reactive-element:build otherwise on a clean repo and run of only `test:dev`, the test server couldn't find reactive-element's _prod_ build files. afaiu, those files shouldn't be requested because we use the `devResolveRemapConfig`. I did confirm that the config is using dev builds, but somehow `devResolveRemapConfig` is not sufficient.


packages/labs/react/package.json
```diff
@@ -30,18 +30,113 @@
...
+      "dependencies": [
+        "test:dev",
+        "test:prod"
+      ]
+    },
+    "test:dev": {
+      "command": "MODE=dev node ../../tests/run-web-tests.js \"development/**/*_test.js\" --config ../../tests/web-test-runner.config.js",
+      "dependencies": [
+        "build:ts",
+        "../../reactive-element:build"
```
justinfagnani, line 116: This should _not_ be necessary, right?

Without, and on a clean repo, I get these errors:

```
Error while handling server request.
[Error: ENOENT: no such file or directory, stat '/Users/justinfagnani/Projects/Lit/lit-html-2/packages/reactive-element/reactive-element.js'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/Users/justinfagnani/Projects/Lit/lit-html-2/packages/reactive-element/reactive-element.js',
  expose: true,
  statusCode: 404,
  status: 404
}
```

justinfagnani: @aomarks any ideas?

justinfagnani:
I made the WTR fix for all packages in https://github.com/lit/lit/pull/2841

packages/labs/react/package.json
aomarks: I saw the same thing in one case in https://github.com/lit/lit/pull/2857. I just added the dependency for now, and put a TODO. Something seems to be broken with the import remapping...

packages/labs/react/package.json
```diff
@@ -30,18 +30,113 @@
...
+      "output": [
+        "*.d.ts{,.map}",
+        "test/**/*.d.ts{,.map}"
+      ]
+    },
+    "build:rollup": {
+      "command": "rollup -c",
+      "dependencies": [
+        "build:ts",
+        "../../reactive-element:build:rollup"
```
aomarks, line 82: This dependency shouldn't be required. Rollups generally build in isolation from their dependencies (except for the polyfill support modules in core).

aomarks (approved):
One comment

packages/labs/react/package.json
```diff
@@ -30,18 +30,113 @@
...
+      ],
+      "files": [
+        "rollup.config.js",
+        "../../rollup-common.js"
```
AndrewJakubowicz, line 86: This is in labs and needs an extra relative `../`. This looks for a file in `/packages/rollup-common.js` that doesn't exist.

I.e. This should be: `"../../../rollup-common.js"`

packages/labs/react/package.json
```diff
@@ -30,18 +30,113 @@
...
+      ]
+    },
+    "checksize": {
+      "command": "rollup -c --environment=CHECKSIZE",
+      "dependencies": [
+        "build:ts"
+      ],
+      "files": [
+        "rollup.config.js",
+        "../../rollup-common.js"
```
AndrewJakubowicz, line 102: One more parent directory up.

AndrewJakubowicz (changes requested):

AndrewJakubowicz (approved):
Pre-approving to not block.

packages/labs/react/package.json
justinfagnani: It seems to be required per my other comment

justinfagnani: actually, this is different. Removed.

AndrewJakubowicz (approved):

aomarks (approved):

PR #2842, opened 2022-05-06
[infra] Set preventAssignment for rollup-replace

justinfagnani:
This removes the error:

```
(!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
```

### DIFF
--- a/rollup-common.js
+++ b/rollup-common.js
@@ -374,11 +374,14 @@ export function litProdConfig({
         // way or another, so it's difficult to define a default in the source code
         // itself.
         replace({
-          'const DEV_MODE = true': 'const DEV_MODE = false',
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
LGTM! It doesn't look like `preventAssignment` will impact any of our replacements.

aomarks (commented):
Here are the docs: https://github.com/rollup/plugins/tree/master/packages/replace#preventassignment

Seems like it should be fine, yeah. Would be nice to double check by running before/after and doing a diff.

justinfagnani:
I checked. But also, every replacement we have ends in a boolean literal that would be a syntax error if it were followed by an assignment.

aomarks (approved):

PR #2848, opened 2022-05-09
Fix typo in API doc.

arthurevans:
Fix is API doc only, error originally reported in https://github.com/lit/lit.dev/pull/769.


AndrewJakubowicz (approved):
Nice! Please add patch changeset.

PR #2849, opened 2022-05-09
[lit-html] Expand documentation for `TemplateResult` and `render`.

AndrewJakubowicz:
Part of https://github.com/lit/lit/issues/1883

This is a documentation only PR which expands the jsdocs for `render` and `TemplateResult`.

Impact:
 - `TemplateResult` is used and referenced very often and has very minimal documentation. This documentation aims to explain what a `TemplateResult` is, and how to render it to DOM via the lit-html `render` function or the LitElement.render lifecycle method.
 - lit-html's `render` function is brilliant and lost some documentation between Lit 1 -> Lit 2.

To document these I referenced:
 - Our wiki page: https://github.com/lit/lit/wiki/How-it-Works#javascript-tagged-template-literals
 - Our Lit 1 documentation for `render`: https://lit.dev/docs/v1/lit-html/rendering-templates/
 - Our Lit 1 generated docs for `render`: https://lit.dev/docs/v1/api/lit-html/templates/#render
 - And the upgrade docs for `render` which document the container appending behavior.

### Screenshots

For example `render` hover over:
Screen Shot 2022-06-08 at 11 05 41 AM


packages/lit-html/src/lit-html.ts
```diff
@@ -443,7 +443,17 @@ const ELEMENT_PART = 6;
 const COMMENT_PART = 7;
 
 /**
- * The return type of the template tag functions.
+ * The return type of the template tag functions, {@linkcode html} and
+ * {@linkcode svg}.
+ *
+ * The `TemplateResult` object captures the arguments passed into the tag
+ * function. A special array of strings that retain their identity between
```
justinfagnani, line 450: I think this might be a little much inside detail for the docs on TemplateResult. It's good for a "How it Works" section though.

I'd try to simplify to something like:

A `TemplateResult` object holds all the information about a template expression required to render it: the template strings, expression values, and type of template (html or svg).

`TemplateResult` objects do not create any DOM on their own. To create or update DOM you need to render the `TemplateResult`. See [Rendering](https://lit.dev/docs/components/rendering) for more information.

I think I'd also include a link to render() and LitElement.render() so users navigating in an IDE don't have to jump to the website to see what can render these.

packages/lit-html/src/lit-html.ts
```diff
@@ -609,9 +619,24 @@ export interface RenderOptions {
 
 /**
  * Renders a value, usually a lit-html TemplateResult, to the container.
- * @param value
- * @param container
- * @param options
+ *
+ * For example, to render "Hello, Zoe!" to the container `document.body`:
```
justinfagnani, line 623: This example doesn't differentiate the string "Hello, Zoe!" from the HTML content "<p>Hello, Zoe</p>". Maybe there's a way to phrase it so that it does? It also doesn't mention that there's a variable in the example - which might be good or might be noise, I'm not sure.

packages/lit-html/src/lit-html.ts
```diff
@@ -609,9 +619,24 @@ export interface RenderOptions {
...
+ *
+ * ```js
+ * import {html, render} from 'lit';
+ *
+ * const name = "Zoe";
+ * render(html`<p>Hello, ${name}!</p>`, document.body);
+ * ```
+ *
+ * @param value Any renderable value, typically a {@linkcode TemplateResult}
```
justinfagnani, line 632: I think this is a good place to document renderable values, or at least include a link to the docs.

justinfagnani (changes requested):

packages/lit-html/src/lit-html.ts
AndrewJakubowicz: Would the documentation link be: https://lit.dev/docs/templates/expressions/#child-expressions

packages/lit-html/src/lit-html.ts
AndrewJakubowicz: Removed variable, but unsure how to address the differentiating between HTML content.

I added a link to the relevant lit.dev in a `@see` tag at the end.

augustjk: I think keeping the variable and expression is valuable, especially if our default example uses `html`.

How about
```
For example, to render "Hello, Zoe!" to the container `document.body` where the name "Zoe" is from a variable:
```

augustjk (approved):
LGTM

packages/lit-html/src/lit-html.ts
```diff
@@ -609,9 +619,28 @@ export interface RenderOptions {
 
 /**
  * Renders a value, usually a lit-html TemplateResult, to the container.
- * @param value
- * @param container
- * @param options
+ *
+ * For example, to render "Hello, Zoe!" to the container `document.body` where
+ * the name "Zoe" is from a variable:
```
arthurevans, line 624: Maybe to @justinfagnani 's point, this should say something like, 

This example renders the text "Hello, Zoe!" inside a paragraph tag, appending it to the container `document.body`.

packages/lit-html/src/lit-html.ts
arthurevans: I think so.

arthurevans (approved):
One small suggestion, otherwise LGTM.

packages/lit-html/src/lit-html.ts
AndrewJakubowicz: Thank you! Adding verbatim.

justinfagnani (approved):

PR #2858, opened 2022-05-09
[context] Prevent context-request event to listen to itself.

### LINKED ISSUE #2843: [labs/context] prevent context-request event to listen to itself
Thanks a lot for the work around lit-lab/context. I was not aware of the [context proposal](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md) and recently discovered (with pleasure) it was implemented. 

Since nothing prevents an element to be a provider AND a consumer of the same context, I am wondering if we should not prevent the `context-request` event to listen to itself (https://github.com/lit/lit/blob/f1b26bfe706ad93a3a312932b0eca5b0261023e0/packages/labs/context/src/lib/controllers/context-provider.ts#L59).

My use-case where this would happen : the DOM tree loosely mapping a json database with some elements holding references to the database hierarchy. One top-level element (context provider) holds a reference to a path in database. Some elements hold child reference relative to this top level reference (context consumer), but could also be context provider for further children down the DOM tree.

christophe-g:
Prevent an element to add callback to itself in case it is a consumer and a provider of the same context. 

Should address #2843

### DIFF
--- a/packages/labs/context/src/lib/controllers/context-provider.ts
+++ b/packages/labs/context/src/lib/controllers/context-provider.ts
@@ -56,7 +56,7 @@ export class ContextProvider<T extends ContextKey<unknown, unknown>>
   public onContextRequest = (
     ev: ContextRequestEvent<ContextKey<unknown, unknown>>
   ): void => {
-    if (ev.context !== this.context) {
+    if (ev.context !== this.context || ev.composedPath()[0] === this.host) {
       return;
     }
     ev.stopPropagation();
--- a/packages/labs/context/src/lib/controllers/context-provider.ts
+++ b/packages/labs/context/src/lib/controllers/context-provider.ts
@@ -56,6 +56,12 @@ export class ContextProvider<T extends ContextKey<unknown, unknown>>
   public onContextRequest = (
     ev: ContextRequestEvent<ContextKey<unknown, unknown>>
   ): void => {
+    // Only call the callback if the context matches. 
+    // Also, in case an element is a consumer AND a provider 
+    // of the same context, we want to avoid the element to self-register.
+    // The check on composedPath (as opposed to ev.target) is to cover cases
+    // where the consumer is in the shadowDom of the provider (in which case, 
+    // event.target === this.host because of event retargeting).
     if (ev.context !== this.context || ev.composedPath()[0] === this.host) {
       return;
     }
--- a/packages/labs/context/src/test/provider-and-consumer_test.ts
+++ b/packages/labs/context/src/test/provider-and-consumer_test.ts
@@ -1,0 +1,115 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+ import {LitElement, html, TemplateResult} from 'lit';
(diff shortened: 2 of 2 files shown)


justinfagnani:
@benjamind can you review this too?

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -56,7 +56,7 @@ export class ContextProvider<T extends ContextKey<unknown, unknown>>
   public onContextRequest = (
     ev: ContextRequestEvent<ContextKey<unknown, unknown>>
   ): void => {
-    if (ev.context !== this.context) {
+    if (ev.context !== this.context || ev.composedPath()[0] === this.host) {
```
justinfagnani, line 59: I think we can check `ev.target === this.host`, since we're guarding against the type of dispatch call on line 72. I think it's a little more clear too.

justinfagnani, line 59: Can you add a comment here on why we do this check?

justinfagnani (changes requested):
We definitely need a test to ensure the correct behavior here.

packages/labs/context/src/lib/controllers/context-provider.ts
christophe-g: @justinfagnani - I wish you are correct. 

`ev.target === this.host` was my initial - non working - code. `ev.target === this.host` is alway truthy in my case, even when the event is dispatched from deeper in the shadowDom hierarchy. 

The thing is, I don't understand why it seems the event is retargeted when it should obviously not. 

I'll try to provide a test to ensure the correct behavior

christophe-g: Actrually, I think we need `ev.composedPath()[0] === this.host`. 

Consider the case where the consumer is in the shadow dom of the supplier (for the same context). By the time the event crosses the shadow boundary, it is retargetted and hence `ev.target === this.host` is truly (and the context would not be  supplied to the consumer). Here, on the contrary we want to the consumer to get the context.

christophe-g: resolved by [f85f2d9](https://github.com/lit/lit/pull/2858/commits/f85f2d985300b97b0a531652613e413482679df9)

benjamind: My understanding of retargetting is spotty, but this sounds right to me?

Would calling `composedPath()` be a performance issue @justinfagnani?

justinfagnani: Ah, yes, All composed events are retargeted. This shouldn't be a perf problem, and is limited to host that consume and provide the same context, so it's not called on all events.

justinfagnani:
The change looks ok - thanks for the explanation - but we still need a test for this.

christophe-g:
>  but we still need a test for this.

Test added via 58e62d02574d3a33a10ea07c8e48ca5b628696a1

sorvell (approved):

justinfagnani (approved):
LGTM! Thanks @christophe-g!

PR #2860, opened 2022-05-10
[@tests] increase default concurrent browsers to 3

taylor-vann:
Currently we run 3 browsers with a max default concurrency of 2.
To avoid leaving a browser hanging, all three should run.
During saucelabs runs this becomes more important due to latency.

This PR increases the default concurrency to 3.

### DIFF
--- a/.changeset/orange-snakes-happen.md
+++ b/.changeset/orange-snakes-happen.md
@@ -1,0 +1,5 @@
+---
+'tests': patch
+---
+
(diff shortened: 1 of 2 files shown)


justinfagnani:
This looks like it reduced the total time for the sauce tests, so I say +1

justinfagnani (approved):

taylor-vann:
Awesome! Glad it improves test time a little more :)

Issue #2724, opened 2022-04-12
[infra] Convert Lit monorepo to Wireit


aomarks:
https://github.com/google/wireit

- [x] benchmarks
- [x] internal-scripts
- [x] labs/analyzer
- [x] labs/cli
- [x] labs/context
- [x] labs/eleventy-plugin-lit
- [x] labs/gen-utils
- [x] labs/gen-wrapper-react
- [x] labs/motion
- [x] labs/observers
- [x] labs/react
- [x] labs/router
- [x] labs/scoped-registry-mixin
- [x] labs/ssr
- [x] labs/ssr-client
- [x] labs/task
- [x] labs/test-projects
- [x] lit
- [x] lit-element
- [x] lit-html
- [x] localize
- [x] localize-tools
- [x] reactive-element
- [x] tests
- [x] ts-transformers

PR #2866, opened 2022-05-11
Convert ts-transformers, localize, and localize-tools to Wireit

aomarks:
Part of https://github.com/lit/lit/issues/2724


packages/localize-tools/package.json
```diff
@@ -19,13 +19,104 @@
...
+        "generate-json-schema",
+        "../lit:build",
+        "../labs/ssr-client:build"
+      ]
+    },
+    "build:ts": {
+      "command": "tsc --build --pretty",
+      "clean": "if-file-deleted",
+      "#comment": [
+        "TODO(aomarks) We depend a test utility from ts-transformers.",
```
rictic, line 44:
```suggestion
        "TODO(aomarks) We depend on a test utility from ts-transformers.",
```

rictic (approved):

packages/localize-tools/package.json
```diff
@@ -19,13 +19,104 @@
...
+    "test:update-goldens": "wireit",
+    "test:check-tsc": "wireit"
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:ts",
+        "generate-json-schema",
+        "../lit:build",
+        "../labs/ssr-client:build"
```
justinfagnani, line 37: is this needed for tests?

justinfagnani: also, the tsconfig reference was to ssr, but this is to ssr-client. Was ssr-client needed but just built transitively?

justinfagnani (approved):

packages/localize-tools/package.json
aomarks: Changed to `ssr:build`. It worked this way because SSR doesn't have a rollup step, so `ssr:build` is just `ssr:build:ts` plus `../ssr-client:build`. But we already had `ssr:build:ts` in the `build:ts` deps here, so the missing piece was `../ssr-client:build`. But it makes more sense to just depend on `../ssr:build`.

PR #2876, opened 2022-05-11
[infra] Wireit conversion for motion, task, observers, context, scoped-registry-mixin

aomarks:
Converts the following packages to Wireit:
- motion
- task
- observers
- context
- scoped-registry-mixin

Part of https://github.com/lit/lit/issues/2724


packages/labs/context/package.json
```diff
@@ -31,18 +31,111 @@
...
+      ]
+    },
+    "test:dev": {
+      "command": "MODE=dev node ../../tests/run-web-tests.js \"development/**/*_test.js\" --config ../../tests/web-test-runner.config.js",
+      "dependencies": [
+        "build"
+      ],
+      "#comment": "TODO(aomarks) Factor out a file-only target in tests/ to DRY these config files into a single place",
+      "files": [
+        "../../tests/run-web-tests.js",
```
AndrewJakubowicz, line 120: Nice dependency on the test script itself. Great example.

packages/labs/context/package.json
```diff
@@ -31,18 +31,111 @@
...
+    },
+    "test:prod": {
+      "command": "MODE=prod node ../../tests/run-web-tests.js \"development/**/*_test.js\" --config ../../tests/web-test-runner.config.js",
+      "dependencies": [
+        "build"
+      ],
+      "files": [
+        "../../tests/web-test-runner.config.js",
+        "../../tests/wtr-config.js",
+        "../../tests/rollup-resolve-remap.js"
```
AndrewJakubowicz, line 135: Nit: Missing dep on `"../../tests/run-web-tests.js",`.

No action required in this PR: Depending on a single file group or a single test dependency as a single target would be awesome here, cause we have a bunch of fiddly test files.

packages/labs/scoped-registry-mixin/package.json
```diff
@@ -28,18 +28,118 @@
...
+    "build:ts:types": "wireit",
+    "build:rollup": "wireit",
+    "test": "wireit",
+    "test:dev": "wireit",
+    "test:prod": "wireit",
+    "checksize": "wireit"
+  },
+  "wireit": {
+    "start": {
+      "command": "web-dev-server -- --node-resolve --watch --preserve-sym-links --open demo/index.html",
```
AndrewJakubowicz, line 43: Unrelated to this PR and non blocking. I could not run this command due to `web-dev-server` command not found. Since it's not in the `devDependencies` I assume this needs to be global.

packages/labs/motion/package.json
```diff
@@ -31,19 +31,119 @@
...
+    "test": "wireit",
+    "test:dev": "wireit",
+    "test:prod": "wireit",
+    "checksize": "wireit"
+  },
+  "wireit": {
+    "start": {
+      "command": "web-dev-server -- --node-resolve --watch --preserve-sym-links --open demo/index.html",
+      "dependencies": [
+        "build"
```
justinfagnani, line 48: For a TODO: Can this be made to just depend on :ts? Should be faster to iterate on. The demos already import local files from `development/`, we'd just need export conditions in a WDS config.

justinfagnani (approved):

packages/labs/motion/package.json
aomarks: Added comment

packages/labs/context/package.json
aomarks: Done, and for all other packages I found that were missing it.

packages/labs/scoped-registry-mixin/package.json
aomarks: `scoped-registry-mixin` depends on `@web/test-runner-mocha`, which is the same as what everything else has. It really doesn't work for you after a top-level bootstrap?

aomarks:
Just pushed a change to disable one of the motion tests on Safari, because it still seems flakey cc @sorvell

PR #2878, opened 2022-05-11
Update labs/analyzer and more to wireit

justinfagnani:
Updates:
* labs/analyzer
* labs/gen-utils
* labs/gen-wrapper-react
* labs/test-projects/test-element-a


kevinpschaaf (approved):
LGTM!

packages/labs/gen-utils/package.json
```diff
@@ -10,9 +10,35 @@
...
+        "test",
+        "index.{js,js.map,d.ts,d.ts.map}"
+      ],
+      "clean": "if-file-deleted"
+    },
+    "test": {
+      "command": "uvu test '.*_test\\.js$'",
+      "dependencies": [
+        "build",
+        "../../lit:build"
```
aomarks, line 37: I've been putting dependency `build` inside `build`. So `build` means "everything you need to actually run this thing".

aomarks: That implies factoring out a `build:ts` here.

packages/labs/gen-utils/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es2020",
-    "module": "es2022",
+    "target": "ES2020",
```
aomarks, line 3: why capitalized?

packages/labs/test-projects/test-element-a/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "incremental": true,
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
aomarks, line 4: This is the default

packages/labs/analyzer/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "incremental": true,
     "composite": true,
-    "tsBuildInfoFile": "./lib/.tsbuildinfo",
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
aomarks, line 5: This is the default (though no object to making that explicit).

packages/labs/gen-wrapper-react/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
-    "target": "es2020",
-    "module": "es2022",
+    "incremental": true,
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
aomarks, line 4: This is the default (though no object to making that explicit).

packages/labs/gen-wrapper-react/package.json
```diff
@@ -10,11 +10,61 @@
...
+      ]
+    },
+    "test:output": {
+      "command": "node ../../tests/run-web-tests.js \"test-output/**/*_test.js\" --config ../../tests/web-test-runner.config.js",
+      "dependencies": [
+        "build",
+        "test:gen"
+      ],
+      "files": [
+        "../tests/run-web-tests.js",
```
aomarks, line 61: These should be `../..` I think?

aomarks (changes requested):

packages/labs/gen-wrapper-react/package.json
aomarks: FYI tracking a feature to detect this problem at https://github.com/google/wireit/issues/215 (but it is non-trivial)

packages/labs/gen-utils/package.json
justinfagnani: This is only needed for the tests to pass, and it's not a static dependency of the tests, but of the generated files. The build task does not depend on this package contents in any way.

This would be clearer if the tests were split into a generate task and test task, but I wanted to avoid changing code in this PR.

packages/labs/gen-utils/tsconfig.json
justinfagnani: I was getting errors in vs code without changing this.

packages/labs/gen-wrapper-react/tsconfig.json
justinfagnani: I had a build error in the test files without adding this there... I did it to all configs in the PR.

packages/labs/test-projects/test-element-a/tsconfig.json
justinfagnani: There was an error without this. We may need to do some path massaging in the analyzer.

aomarks (approved):

packages/labs/gen-wrapper-react/package.json
justinfagnani: Fixed

aomarks (approved):

PR #2879, opened 2022-05-11
Fix motion tests

sorvell:
Some non-ideal workarounds for browser quirks:
* Introduces `closeTo` for numbers that are slightly off and
* adds a sleep to let Safari assert ok.

Fixes https://github.com/lit/lit/issues/2009

### DIFF
--- a/.changeset/breezy-days-prove.md
+++ b/.changeset/breezy-days-prove.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/motion': patch
+---
+
+Test fixes.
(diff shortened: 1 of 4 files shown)


taylor-vann:
awesome!

.changeset/breezy-days-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/motion': patch
```
aomarks, line 2: We shouldn't do a release for a test fix, so this should be an empty change set.

aomarks (changes requested):

.changeset/breezy-days-prove.md
sorvell: Done.

aomarks (approved):

PR #2887, opened 2022-05-12
[@tests] run firefox in tests workflow

taylor-vann:
## abstract

This PR creates a job in the `test` workflow to run integration tests for `firefox` and `firefox-esr`. Firefox does not need to be run in saucelabs to meet our testing requirements.

A `firefox` preset is created so devs can run `BROWSERS=preset:firefox` and run both `firefox` and `firefox-esr`

## background

I wanted to explore alternatives to saucelabs for browser testing. One that we have more control over and less latency.

Github actions provides the ability to create custom docker images. I goof around with `podman` and `docker` in my free time. But I wanted to understand the default docker images github uses for their actions before I made custom docker images.

As in what do we get for free? What comes out of the box?

### what's included in github virtual environments

Turns out [quite a lot](https://github.com/actions/virtual-environments) is provided out of the box!

[MacOS](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-12-Readme.md#browsers) images provide:
- safari
- safari driver
- safari with 'allow remote automation' enabled by default
- (also chrome + drivers and firefox + drivers)

[Ubuntu](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers) images provide:
- firefox + geckodriver
- chrome + chromedrivers

### What's not included in github virtual environments

Unfortunately not everything required from our stated baseline support is provided:
- Firefox ESR must be installed manually (solved in this PR)
- Chrome versioning is unavailable
- IE and Edge require windows-based environments
- Our repo does not currently build in windows-based environments

However, I don't consider these blocking challenges.

Chrome itself doesn't provide older versions based on security concerns. What chrome considers stable is the ideal benchmark for the chrome browser.

This PR installs Firefox ESR. The WTR config then points to the Firefox ESR binary. This challenge is resolved.

Lastly, we can make healthy changes to support windows build in our repository.

## Results

This PR tests Firefox and Firefox ESR in the `test` workflow.  We keep our ability to screenshot tests. Devs can run browser tests on their local machines as well.

Saucelabs remains the default option for testing IE, Chrome, and Safari.

## Future work

It is worth pursuing to create a [windows workflow](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md) to test on IE and Edge.

This PR would include it but our repo does not currently build on windows. Future effort to unblock would be required.
A temporary solution is to retain saucelabs services and run tests on IE11 until a windows-based workflow can replace it.

Chrome does not offer `N-2` versions so it will be tested on saucelabs (unless suitable browser and driver versions can be found).


.github/workflows/firefox.yaml
```diff
@@ -0,0 +1,42 @@
+name: Firefox
```
justinfagnani, line 1: Why is this a separate workflow? Can't it be run with the other local tests?

packages/tests/src/web-test-runner.config.ts
```diff
@@ -147,6 +156,35 @@ See https://wiki.saucelabs.com/display/DOCS/Platform+Configurator for all option
...
+  if (browser.startsWith(SELENIUM)) {
+    let browserName = browser.substring(SELENIUM.length);
+    if (!seleniumBrowsers.has(browserName)) {
+      throw new Error(`
+Invalid Selenium browser string.
+Expected format: selenium:browser
+Available options:
+  - firefox
+  - firefox-esr
+  - safari
```
justinfagnani, line 168: is `safari` really a valid option?

taylor-vann: I'll remove!

taylor-vann: done :)

.github/workflows/sauce.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     runs-on: ubuntu-latest
     env:
-      BROWSERS: preset:sauce
+      BROWSERS: 'sauce:Windows 10/Chrome@latest-2,sauce:macOS 10.15/Safari@latest'
```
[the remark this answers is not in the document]
taylor-vann: done!

.github/workflows/firefox.yaml
taylor-vann: done!

taylor-vann:
sweet! made those changes, thanks for the review!

taylor-vann:
closing because nobody likes selenium lol

PR #2888, opened 2022-05-12
[infra] Create a build script for test configs and depend on it on package tests

justinfagnani:
This removes the need for each test to repeat the test configs and runner script in their files.

This might need to go after all packages are converted to wireit so that their test build the test scripts?


.changeset/nine-apples-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'tests': patch
+---
+
+Convert test scripts and config to TypeScript
```
aomarks, line 5: Should be an empty changeset. Changesets are for releases and communicating to users, neither of which we want for this.

packages/tests/package.json
```diff
@@ -35,9 +34,31 @@
...
+    "test": {
+      "dependencies": [
+        "test:dev",
+        "test:prod"
+      ]
+    },
+    "test:dev": {
+      "command": "node ./run-web-tests.js",
+      "dependencies": [
+        "build"
```
aomarks, line 54: I think `test:dev` and `test:prod` should depend on all of the packages that the monolithic test config tests. And then if we add `files:[], output:[]` we will be able to cache all browser tests as long as no browser packages changed.

aomarks (changes requested):

.changeset/nine-apples-change.md
justinfagnani: Done

packages/tests/package.json
justinfagnani: How about I do this in a follow-up? I think we should include the node tests here too, so given the way we might want to separate tests on runners, we might want `test`', `test:browser`, `test:node`, `test:browser:dev`, etc. Seem worth a little discussion.

aomarks: Follow up sounds good. Surprised that we would want to run node tests in here, though. I think of this package as "the monolithic test package that only exists because it's too slow/flakey to run browser tests in isolation".

justinfagnani: I guess we want to remove running all tests here and move that to the top-level package.json. We need the WTR --no-tunnel flag for that though, so I can do what you recommended here in the meantime. I'll still do a follow-up though.

.changeset/thick-grapes-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+'tests': patch
+---
+
+Update test scripts to depend on tests:build
```
aomarks, line 5: Still got a changeset here.

aomarks (changes requested):

packages/tests/package.json
```diff
@@ -43,19 +41,7 @@
...
-    },
-    "test:dev": {
-      "command": "node ./run-web-tests.js",
-      "dependencies": [
-        "build"
-      ]
-    },
-    "test:prod": {
-      "command": "MODE=prod node ./run-web-tests.js",
+      "command": "node ./run-web-tests.js && MODE=prod node ./run-web-tests.js",
```
aomarks, line 44: We could also set `WIREIT_PARALLEL=1` after the build step, but before the test step.

aomarks (approved):

packages/tests/package.json
justinfagnani: Merging w/o this for now. It would mean we could get sauce tunnel errors if we ran this locally. I think we'll be refactoring this still...

PR #2891, opened 2022-05-13
Convert more packages to depend on tests:build

justinfagnani:
I missed some packages that got converted while the past PR was in flight. This fixes some warnings I saw while building too.


packages/labs/context/rollup.config.js
```diff
@@ -10,5 +10,10 @@ import {createRequire} from 'module';
 export default litProdConfig({
   packageName: createRequire(import.meta.url)('./package.json').name,
   entryPoints: ['index'],
-  external: ['@lit/reactive-element', 'lit', 'lit/directive.js'],
+  external: [
+    '@lit/reactive-element',
```
aomarks, line 14: Can we make all bare modules external? Bundled is the exception, right? Just polyfill-support?

aomarks (approved):

packages/labs/context/rollup.config.js
justinfagnani: I would have thought that the reactive element entry would have taken care of this. I'd like to look into it more.

PR #2894, opened 2022-05-13
[gen-wrapper-react] Add runtime tests for generated react package

kevinpschaaf:
Fixes: #2859

See "Testing generated packages" in [go/lit-cli-generate](http://go/lit-cli-generate) for details.


packages/labs/gen-utils/src/test/package-utils_test.ts
```diff
@@ -4,85 +4,93 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
```
kevinpschaaf, line 6: Changes here are just switching the test file to use FilesystemTestRig (missed that change in the PR that added it), and add tests for `packPackage`.

packages/labs/gen-wrapper-react/src/index.ts
```diff
@@ -46,7 +47,7 @@ export const generateReactWrapper = async (
   const litModules: LitModule[] = getLitModules(analysis);
   if (litModules.length > 0) {
     const reactPkgName = packageNameToReactPackageName(
-      analysis.packageJson.name!
+      path.basename(analysis.rootDir)
```
kevinpschaaf, line 50: This change was to base the folder we generate the package into off the folder name of the analyzed package (rather than the npm package name, since that might have an org in it).

packages/labs/test-projects/test-element-a/package.json
```diff
@@ -1,6 +1,6 @@
 {
```
kevinpschaaf, line 1: Per convo with Justin, we're going to start putting private monorepo packages that won't ever be published in a `@lit-internal` "org", to make it clear what's internal and what's expected to come from the registry. We'll eventually do this for the `tests` package also.

sorvell (approved):

lit-next.code-workspace
```diff
@@ -29,33 +29,57 @@
       "path": "packages/lit-starter-js"
     },
     {
-      "name": "localize",
-      "path": "packages/localize"
+      "name": "labs/ssr",
```
justinfagnani, line 32: Does it really help to have "labs" in the name? Could we make these changes separately?

justinfagnani (approved):

lit-next.code-workspace
kevinpschaaf: I figure I'm the only one using this... 😛

It helps because afaict you have to have the package listed here for its launch config to show up in the UI (otherwise there's not a ton of value listing each package as a workspace folder, you can just open the monorepo folder and browse the file tree). But if they're gonna be listed, I find not having them roughly match the file tree disorienting.

Maybe this argues we should just have one big launch config in the root and not per package. 🤷 

I can move it out of the PR if you want to debate it I guess.

justinfagnani: Oh, I use it. It used to at least make TypeScript behave a lot better because VS Code could understand what was a project and how to resolve references. tsconfig project references may have made that better, but we've removed those recently.

I guess one approach is to have a checked-in base workspace file and if people want to customize they can make their own. We could add an entry to .gitignore to put your personal fork in a known location.

PR #2904, opened 2022-05-14
[labs] Prevent animate to start() when no frame

### LINKED ISSUE #2903: [@lit-labs/motion] onStart should not be called when there is no frame
The `onStart` function is always called whenever an animation starts. 

However, the `onComplete` is called only when animation frames exist. 

This is a problem for instance when you want to hide some elements (e.g. `opacity = 0` ) during the animation (for instance because the applied transform makes them appear weird) and display them again when the animation completes. In some cases, this ends up with only the `onStart` function being called (hiding elements), and the hidden elements not appear again because `onComplete` is not called.

christophe-g:
This should fix #2903

### DIFF
--- a/packages/labs/motion/src/animate.ts
+++ b/packages/labs/motion/src/animate.ts
@@ -534,7 +534,11 @@ export class Animate extends AsyncDirective {
     frames: Keyframe[] | undefined,
     options = this.options.keyframeOptions
   ) {
-    this.start();
+    // Checking existence of frames to prevent situation 
+    // where options.onStart is called and options.onComplete is not. 
+    if (frames !== undefined) {
+      this.start();
+    }
     this.frames = frames;
     let didAnimate = false;
     if (!this.isAnimating() && !this.isDisabled()) {


packages/labs/motion/src/animate.ts
```diff
@@ -534,7 +534,11 @@ export class Animate extends AsyncDirective {
     frames: Keyframe[] | undefined,
     options = this.options.keyframeOptions
   ) {
-    this.start();
+    // Checking existence of frames to prevent situation 
+    // where options.onStart is called and options.onComplete is not. 
+    if (frames !== undefined) {
```
sorvell, line 539: I think it probably makes sense to move this into the condition below line 544 with the same check.

The design and order of the callbacks still needs some refinement but that's probably out of scope here. In particular the frames could become undefined via `onFrames` and you'd still get a start without a complete. To fix that, we'd probably want to change the callback ordering so `onFrames` came first and maybe would be called `onPrepare`. That's a breaking change so would need more consideration.

I think here either punting on that case or changing the `didAnimate` flag to `didStart` and setting that if `onStart` runs would both be ok.

sorvell (changes requested):
We also need a test that covers this case and would break without the change. Thanks!

PR #2913, opened 2022-05-16
[@tests] reduce calls to wtr in IE workflow

taylor-vann:
Reduce calls to web test runner in IE github workflow


packages/tests/package.json
```diff
@@ -45,6 +47,18 @@
       "dependencies": [
         "build"
       ]
```
aomarks, line 49:
```suggestion
      "dependencies": [
        "test:dev",
        "test:prod"
      ]
```

(And remove "command", I couldn't put that in the suggestion)

.changeset/olive-crews-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'tests': patch
+---
+
+expose dev and prod test commands
```
aomarks, line 5: Remove changeset, not user facing

packages/tests/package.json
aomarks: (this is optional, just a neater way to write the same thing)

aomarks (approved):

.changeset/olive-crews-complain.md
taylor-vann: sounds good!

packages/tests/package.json
taylor-vann: will do!

.changeset/olive-crews-complain.md
taylor-vann: done :)

packages/tests/package.json
taylor-vann: done!

aomarks: Make sure to also remove the command, or else it will run everything twice.

aomarks: I still see the command

.github/workflows/sauce-ie11.yaml
```diff
@@ -23,6 +23,12 @@ jobs:
           node-version: 16
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
+      
+      - uses: saucelabs/sauce-connect-action@v2
```
aomarks, line 27: This shouldn't be here, right?

aomarks (changes requested):
Seeing a new change that doesn't seem to belong

.github/workflows/sauce-ie11.yaml
taylor-vann: yes but that's from a totally different commit

taylor-vann: good catch but i have no idea why that's there

taylor-vann: done

taylor-vann:
kewl, removed artifacts.
also removed the added wireit commands since this doesn't need them anymore

aomarks (approved):

PR #2914, opened 2022-05-16
Minor fixups

rictic:
* the type of Object.assign will get stricter in TS 4.7
* specify tsbuildinfo file name in ts-transformers so that the tsbuildinfo file is placed in the right place and doesn't clash with other packages' location.
* ignore .wireit directories in vscode


lit-next.code-workspace
```diff
@@ -62,7 +62,10 @@
     "debug.javascript.warnOnLongPrediction": false,
     // Regardless of the user's usual formatter, files
     // in the Lit repo must be formatted using prettier.
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "files.exclude": {
```
aomarks, line 66: Doesn't it already ignore everything in gitignore?

rictic: Not completely, e.g. wireit files fill the cmd+P menu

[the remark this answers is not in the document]
rictic: [dupe comment]

packages/ts-transformers/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
 
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
aomarks, line 21: I think this is the default but explicit good

aomarks (approved):

packages/ts-transformers/tsconfig.json
rictic: It was reading/writing from the wrong location on my machine, and this fixed it. 🤷

PR #2918, opened 2022-05-17
[chore] Update lockfiles

augustjk:
Update package-lock files with version bumps that happened in the previous release.
These changes were made by running `npm run nuke` from the latest commit.

Perhaps we should consider doing something like this https://github.com/changesets/changesets/issues/421#issuecomment-1029940897 to automate this.


justinfagnani:
I don't think updating lock files needs to or really should be associated with releases. I like the idea of regularly and periodically updating dependencies. Dependabot should theoretically be doing that, but we probably need our test stabilized before automatically merging its PRs.

justinfagnani (approved):
thanks!

augustjk:
> I don't think updating lock files needs to or really should be associated with releases. I like the idea of regularly and periodically updating dependencies. Dependabot should theoretically be doing that, but we probably need our test stabilized before automatically merging its PRs.

In this case we're not actually updating any dependencies. The diffs here are purely the version number of the actual package themselves.

For instance, changeset updates the version in `lit`'s `package.json` from 2.2.3 to 2.2.4 but does not update the top level `"version"` property in the `package-lock.json`, which stays at 2.2.3, and therefore out of sync with the its corresponding `package.json`.

justinfagnani:
ahhh. In that case, changesets should fix and until then we definitely could :)

PR #2922, opened 2022-05-18
[starter-kits] Update starter kits

augustjk:
Many dependencies in the starter kits were out of date and had vulnerabilities.

This update takes us from

```
34 vulnerabilities (10 moderate, 23 high, 1 critical)
```

to

```
3 high severity vulnerabilities
```

The remaining vulnerabilities requires bumping dependency in `lit-analyzer` which I've decided to punt.

There were also several unused dependencies that I've removed from the starter kit repos:
- @esm-bundle/chai
- @open-wc/testing-karma
- @types/mocha
- @web/dev-server-rollup
- @web/test-runner-mocha
- deepmerge
- mocha

Also removed `my-element.bundled.js` from git history. It was already gitignored in the js starter. Added it to ts starter as well.

Confirmed npm scripts in the repos still work.


packages/lit-starter-ts/package.json
```diff
@@ -41,37 +41,30 @@
...
+    "lit-analyzer": "^1.2.1",
+    "prettier": "^2.6.2",
     "rimraf": "^3.0.2",
-    "rollup": "^2.70.2",
-    "rollup-plugin-summary": "^1.2.3",
+    "rollup": "^2.73.0",
+    "rollup-plugin-summary": "^1.4.3",
     "rollup-plugin-terser": "^7.0.2",
-    "typescript": "^4.3.5"
+    "typescript": "^4.6.4"
```
justinfagnani, line 67: Let's change this to a `~` because TypeScript makes breaking changes on minor releases.

packages/lit-starter-js/package.json
```diff
@@ -39,34 +39,27 @@
...
-    "@rollup/plugin-node-resolve": "^13.0.4",
-    "@rollup/plugin-replace": "^3.0.0",
-    "@types/mocha": "^9.0.0",
-    "@web/dev-server": "^0.1.22",
-    "@web/dev-server-legacy": "^0.1.4",
-    "@web/dev-server-rollup": "^0.3.9",
-    "@web/test-runner": "^0.13.16",
-    "@web/test-runner-mocha": "^0.7.4",
-    "@web/test-runner-playwright": "^0.8.4",
+    "@open-wc/testing": "^3.1.5",
```
justinfagnani, line 49: So I guess the Chai globals are coming in via this package?

packages/lit-starter-ts/package.json
augustjk: Done!

packages/lit-starter-js/package.json
augustjk: Yes, `@open-wc/testing` has `@esm-bundle/chai` as a dependency and re-exports so we don't need to depend on it ourselves.

justinfagnani (approved):

PR #2932, opened 2022-05-20
[cli] Lazily install and locally version the react generator

rictic:
This extends the existing Command protocol to include a GeneratorCommand, which follows our generate interface while having the lazy installation and package-level loading and versioning properties of ReferenceToCommand.

Recommend reviewing with &w=1


packages/labs/cli/src/test/gen/react_test.ts
```diff
@@ -47,13 +48,14 @@ test('basic wrapper generation', async ({rig, testConsole}) => {
...
     {
+      cwd: rig.rootDir,
       console: testConsole,
     }
   );
   testConsole.alsoLogToGlobalConsole = true;
   await cli.run();
 
-  assert.equal(testConsole.errorStream.buffer.length, 0);
+  assert.equal(testConsole.errorStream.buffer.join(''), '');
```
kevinpschaaf, line 58: Minor/for my understanding: Unclear how these would be different? I guess the writeable could get a 0-char chunk? Should we just make a `toString()` on `BufferedWriteable` that does the join?

I guess I'm unclear on whether there is semantic meaning to chunk boundaries such that anyone would want the array, or if chunk boundaries could be arbitrary and thus the `buffer` array should be private and callers should only ever get the `toSring()` result?

packages/labs/gen-wrapper-react/src/index.ts
```diff
@@ -14,27 +14,36 @@ import {
...
-    };
-  } else {
-    throw new Error('No Lit components were found in this package.');
-  }
+export const getCommand = () => {
+  return {
+    name: 'react',
+    description: 'Generate React wrapper components from Lit elements',
+    kind: 'resolved',
+    async generate(options: {analysis: Package}): Promise<FileTree> {
```
kevinpschaaf, line 22: Could we keep a top-level export of `generate` for standalone usage, while also exporting a `getCommand()` for use in the CLI?

The idea is that other tools could call the generate function directly, so it's sort of weird if it has a very distinct Lit CLI shape. That would also address the need to make a generate-specific command type (comment above), since the command in `getCommand()` can just be called `run`.

justinfagnani (approved):

rictic:
Fixed https://github.com/lit/lit/issues/2869

PR #2933, opened 2022-05-20
[@tests] update @web/test-runner-saucelabs version

taylor-vann:
This PR updates dependencies in the `tests` package:
- `@web/test-runner-saucelabs` is updated to 0.8.0
- `npm update` was run to update 6 vulnerabilities: 1 critical and 4 high, all resolved


packages/tests/src/web-test-runner.config.ts
```diff
@@ -68,10 +68,14 @@ function makeSauceLauncherOnce() {
           ' and SAUCE_ACCESS_KEY environment variables.'
       );
     }
-    sauceLauncher = createSauceLabsLauncher({
-      user,
-      key,
-    });
+    sauceLauncher = createSauceLabsLauncher(
```
justinfagnani, line 71: Can we only update the `@web/test-runner-saucelabs` version in this PR?

taylor-vann: of course! only added it because previous fail, reverting now

PR #2934, opened 2022-05-21
[workflows] reduce workflows used in PRs

taylor-vann:
This PR consolidates workflows to conserve resources used in github actions. I wanted to use `if` logic in a github action to accomplish the same tasks in one workflow.

This combines `sauce`, `sauce-ie11`, `lint`, `changeset`, and `local-tests` into one github workflow. They all consume the same resources.

One effect is if a `changeset` or `lint` fails, no tests will run. Likewise, `Sauce` tests will only be run if local tests pass and `IE` tests will run only if `sauce` is successful. 

This reduces our saucelabs bandwidth and effectively only runs remote tests when local tests pass. This is also the exact use-case for saucelabs: to test browser behavior past what we can do locally.


taylor-vann:
Screen Shot 2022-05-23 at 11 04 55 AM
Here's an example of what it looks like when all tests pass, notice the `Remote Test Failure Check` does not run on a successful pass

As opposed to remote tests failing, where the failure check will output what remote tests failed:
Screen Shot 2022-05-23 at 11 14 13 AM

justinfagnani:
> This combines sauce, sauce-ie11, lint, changeset, and local-tests into one github workflow. They all consume the same resources.

I'm not sure we should be combining this many things. I thought we talked about combining lint + changeset, and all the sauce browsers once we stabilized IE.

I think the criteria for grouping should be around what dependencies the jobs have (ie, lint and changeset don't require an npm install), how fast the results can come back (lint and changeset can both return in < 1 min, local tests < 10 min, sauce tests maybe a bit longer, benchmarks ~30min), and whether independent results will be useful to us, and how much we'd get out of parallelism. Running Node tests, then local browser tests, then Sauce tests then IE tests in sequence is just going to take a long time and we won't see job results until the whole thing is done. And before IE is stabilized we won't have a separate signal that that failed that we can consider before merging.

taylor-vann:
sounds good. i'll make a separate PR for the lint + changeset and keep this one as a draft.
I thought it was quite successful and greatly reduced our workflows and bandwidth.

taylor-vann:
This is something I'd like to move towards though in terms of workflows and testing. Most of our workflows run on ubuntu. And they all build the same resources.

An approach like this one would provide a gated and measured way of progressing to expensive resources.
- lack of changesets and lints would block any local tests ($ early bail, cheap)
- lack of local tests passing would block any remote tests ($$ costs time)
- sauce labs tests only runs when local tests run ($$$ expensive, actual browsers, saucelabs bandwidth)

I feel this is more reflective of staging and progressing from a `dev` like test environment to a `prod` like test environment.

Something to aim for I feel

Issue #2925, opened 2022-05-18
[@lit-labs/router] Protected properties renamed breaking subclassing.


AndrewJakubowicz:
Expect the following to work:

```ts
class Foo extends Routes {
  log() {
    console.log(this._currentRoute);
  }
}
```

What actually happens: `this_currentRoute` doesn't exist because it has been renamed by our [rollup renaming rule](https://github.com/lit/lit/blob/main/rollup-common.js#L239-L241) which renames properties prefixed with `_`.

Any publicly accessible API should not be renamed. Audit the router API.

Members that need resolving on the `Routes` class:

 - `_parentRoutes`
 - `_currentPathname`
 - `_currentRoute`
 - `_currentParams`
 - `_host`

PR #2937, opened 2022-05-23
[@lit-labs/router] Make _ prefixed class members private fixing renaming.

AndrewJakubowicz:
Addresses: https://github.com/lit/lit/issues/2925

### Context

Currently we have a [rollup renaming rule](https://github.com/lit/lit/blob/main/rollup-common.js#L239-L241) that renames properties prefixed with `_`.

The renaming is incompatible with subclassing.

### Fix

Make the underscored properties private, and expose getters for public/protected endpoints.

For example, `_currentParams` has a public API: `params`.

We can add a public API as we better understand what is required.

### DIFF
--- a/.changeset/tasty-eyes-end.md
+++ b/.changeset/tasty-eyes-end.md
@@ -1,0 +1,7 @@
+---
+'@lit-labs/router': minor
+---
+
+**[BREAKING]** Router properties prefixed with an underscore have been made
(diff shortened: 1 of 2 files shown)


justinfagnani (approved):
We might also have some work to do here to use a prefix for renaming to avoid accidental name collisions, or prevent subclassing.

PR #2940, opened 2022-05-23
[ssr] Add option to defer hydration on top level custom element

### LINKED ISSUE #2938: [ssr] Add option to defer hydration for top level custom elements

augustjk:
Resolve #2938

### DIFF
--- a/packages/labs/ssr/src/lib/render-lit-html.ts
+++ b/packages/labs/ssr/src/lib/render-lit-html.ts
@@ -522,15 +522,16 @@ export type RenderInfo = {
   customElementRendered?: (tagName: string) => void;
 
   /**
-   * An optional flag to defer hydration of top level custom element
+   * Flag to defer hydration of top level custom element. Defaults to false.
    */
-  deferHydration?: boolean;
+  deferHydration: boolean;
 };
 
 const defaultRenderInfo = {
   elementRenderers: [LitElementRenderer],
   customElementInstanceStack: [],
   customElementHostStack: [],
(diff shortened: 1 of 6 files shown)


packages/labs/ssr/src/lib/render-lit-html.ts
```diff
@@ -520,6 +520,11 @@ export type RenderInfo = {
    * template, even in the case of conditional templates.
    */
   customElementRendered?: (tagName: string) => void;
+
+  /**
+   * An optional flag to defer hydration of top level custom element
```
justinfagnani, line 525: Say what the default is?

justinfagnani (approved):
🎉

aomarks (approved):

kevinpschaaf (approved):
It might also be nice to add an integration test for this that ensures `defer-hydration` is actually respected on the top-level element; today we only really test `defer-hydration` at all transitively in the `LitElement` tests at the bottom of `basic.ts`, but not directly.

Testing it in the integration tests would require adding something like an optional `serverRenderOptions` field to `SSRTestDescription` and passing it into [here](https://github.com/lit/lit/blob/main/packages/labs/ssr/src/test/integration/server/server.ts#L52). Then you could add a test to `basic.ts` with `deferHydration: true` that registers an element and `check()`s that e.g. an event handler doesn't actually work yet, and then remove the attribute and see that it does?

That could be in a follow-up PR if we don't want to block this, either way.

kevinpschaaf (approved):
Looks great!

Issue #2948, opened 2022-05-25
[@lit-labs/react] use-controller.js not exported from package.json


ajmchambers:
# Description

@lit-labs/react 'use-controller.js' is not currently exported from the package.json so when you go to import it, you get an error.

```ts
import { useController } from '@lit-labs/react/use-controller.js';
```

```
Missing "./use-controller.js" export in "@lit-labs/react" package
```

PR #2949, opened 2022-05-25
[@lit-labs/react] use-controller.js exported from package

ajmchambers:
fixes https://github.com/lit/lit/issues/2948


ajmchambers:
sorted in https://github.com/lit/lit/pull/2960

PR #2957, opened 2022-05-26
[labs/testing] testing package with SSR testing utilities

### LINKED ISSUE #2867: [labs/ssr] Initial setup of SSR fixture package
Including:
* Initial implementation of `fixture` API
* Initial implementation of WTR plugin/middleware

augustjk:
Resolves #2867 

This PR adds a `@lit-labs/testing` package.

The `litSsrPlugin` plugin for Web Test Runner registers the `lit-ssr-render` server command that takes the template and module paths, and renders them within a worker thread in the same manner as eleventy-plugin-lit.

The server command is utilized by `ssrFixture` to be called in the test file supplying lit-html template to render and component definition module path(s).

The README has example usage.

Still to do are to add tests, and a fixture function with client side rendering with the same signature for ease of writing tests across the different rendering methods.

This is still an early draft and I'm looking for feedback on
- The structure of the package and exports.
- Possibly a better way to handle the relative paths of the modules.
- How to test -- I'm currently thinking doing an integration test with a whole sample component, running WTR with Chrome, FF, Safari, with the plugin registered and running tests utilizing all the fixture function. I'm unsure how to unit test this, if necessary.


packages/labs/testing/README.md
```diff
@@ -0,0 +1,95 @@
...
+    const el = await ssrFixture(html`<my-element></my-element>`, {
+      modules: ['./my-element.js'], // path to component definition relative to project root
+      hydrate: false, // whether to hydrate the component after loading to document (default: true)
+    });
+    assert.equal(el.shadowRoot.querySelector('p').textContent, 'Hello, World!');
+  });
+});
+```
+
+#### `ssrNonHydratedFixture` and `ssrHydratedFixture`
```
kevinpschaaf, line 58: I forget what we concluded, but I was also expecting a CSR version with the same signature that used the client-side `render()`?

packages/labs/testing/src/web-test-runner/worker/worker.ts
```diff
@@ -0,0 +1,63 @@
...
+
+if (parentPort === null) {
+  throw new Error('worker.js must only be run in a worker thread');
+}
+
+let initialized = false;
+
+parentPort.on('message', async (message: RequestMessage) => {
+  switch (message.type) {
+    case 'initialize-request': {
```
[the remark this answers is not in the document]
augustjk: Good point. This structure was necessary in the eleventy plugin because it needed multiple render requests with a single initialize. Here, we start a new worker thread for each command invocation so this could be a lot more simplified.

packages/labs/testing/README.md
augustjk: Yes, we did want that. It's on the todo list.

kevinpschaaf:
> Possibly a better way to handle the relative paths of the modules.

It seems like we'll need to require the `fixture` caller to pass their `import.meta.url` in order to allow them to use relative paths, something like:
```ts
    const el = await ssrFixture(html`<my-element></my-element>`, {
      base: import.meta.url, // path to resolve relative modules against
      modules: ['./my-element.js'], // path to component definition, relative to base (or cwd if not specified)
      hydrate: false, // whether to hydrate the component after loading to document (default: true)
    });
```

On the client, it would just `import(new URL(module[i], base).href)`, and on the server it should be able to work out where to import it from relative to where the server is serving from?

kevinpschaaf:
> How to test -- I'm currently thinking doing an integration test with a whole sample component, running WTR with Chrome, FF, Safari, with the plugin registered and running tests utilizing all the fixture function. I'm unsure how to unit test this, if necessary

Yeah, doing this as an integration test seems like the most bang-for-buck. Seems like you'll want to test a variety of component scenarios, obv a happy case where it all passes, but also some cases where we expect the tests to fail e.g. to ensure we catch when users do SSR-incompatible things. I'd expect main focus to be on the latter, since the `labs/ssr` integration tests themselves focus on all the happy cases.

^^ points out that we'll want to nicely pipe errors raised from `render-lit-html:render()` on the server back through to the client via a rejection on the async fixture; that should allow writing try/catch blocks in the integration test to ensure they fail when expected.

(hmmm, now that I think about it)
> I'd expect main focus to be on the latter, since the SSR tests themselves focus on all the happy cases.

Seems like we might just be able to switch all the `labs/ssr` integration tests to use this and use a normal WTR setup; it would just require all the element definitions in the `LitElement` tests to be moved out to their own modules. But that can happen separately after we get some basic tests for this landed.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
+/**
```
aomarks, line 1: I think we generally use `kebab-case` instead of `camelCase` for our filenames. (I'm also seeing `snake_case` in some packages, but I think `kebab-case` is more common, e.g. see https://github.com/lit/lit/tree/main/packages/lit/src).

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+import {render} from 'lit';
+import type {LitElement, TemplateResult} from 'lit';
+import type {FixtureOption} from './fixtureOption.js';
+
+/**
+ * Renders the provided lit-html template with a Lit element client-side.
+ *
+ * @param {TemplateResult} template - lit-html template. Must contain a single
```
aomarks, line 14: I don't think we want the `{types}` in the jsdoc, since the types should get correctly extracted by vscode and other tools directly from the typescript syntax, and that's a better source of truth.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+ * top level custom element.
+ * @param {string[]} option.modules - Path to custom element definition modules
```
aomarks, line 16: In VSCode this annotation style doesn't quite seem to work -- only the `Path to custom element definition modules` part renders when you are typing the parameter. I think if we annotate the `FixtureOption` properties directly on that type signature, it will work better.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
aomarks, line 14: (Same question as above -- does it really need to be a top-level custom element?)

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+ * needed to render template, relative to the test file.
+ * @param {string} [option.base] - Optional. Base path for the modules.
```
aomarks, line 18: Base path for resolving relative paths in `modules`?

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Generally should be `import.meta.url`. Will guess from stack trace if not
+ * provided.
+ */
+export async function csrFixture<T extends LitElement>(
+  template: TemplateResult,
+  {modules, base}: FixtureOption
+): Promise<T> {
+  if (base === undefined) {
+    // Find the test file url from the call stack
```
aomarks, line 27: This is quite clever! But maybe kind of surprising? I wouldn't realy expect a function to behave differently depending on where it was called from. Is there some precedent for this, like something `@open-wc/testing` also does?

Another idea could be to just always require the base url be manually specified.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+  // TODO(augustinekim) Clean up the container from the document
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+
+  await Promise.all(
+    modules.map((module) => import(new URL(module, base).href))
+  );
+
+  // Webkit does not seem to render the contents of the custom element
+  // synchronously. Awaiting for the next microtask tick seems to work.
```
aomarks, line 62: Isn't this the same on Chrome and Firefox too? Doesn't LitElement always schedule first render on a microtask?

packages/labs/testing/src/lib/fixtures/fixtureOption.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export interface FixtureOption {
```
aomarks, line 7:
```suggestion
export interface FixtureOptions {
```

packages/labs/testing/src/lib/fixtures/fixtureOption.ts
```diff
@@ -0,0 +1,14 @@
...
+  modules: string[];
```
aomarks, line 8: Descriptions

packages/labs/testing/src/lib/fixtures/ssrFixture.ts
```diff
@@ -0,0 +1,152 @@
...
+      template,
+      modules: modules.map((module) => new URL(module, base).pathname),
+    }
+  );
+  // TODO(augustinekim) Clean up the container from the document
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+
+  if (HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
+    // Browser natively supports declarative shadowroot but must use DOMParser
```
aomarks, line 89: expand on why we must use domparser

packages/labs/testing/src/lib/litSsrPlugin.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface Payload {
+  template: TemplateResult;
+  modules: string[];
+}
+
+export function litSsrPlugin(): TestRunnerPlugin<Payload> {
+  return {
+    name: 'lit-ssr-plugin',
+    async executeCommand({command, payload}) {
+      if (command === litSsrPluginCommand) {
```
aomarks, line 23: Suggest early return here

```suggestion
      if (command !== litSsrPluginCommand) {
        return undefined;
      }
```

aomarks (changes requested):
This is looking really great! Pretty minor comments.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
augustjk: I had originally made the base required but I added this purely as a convenience for users. This should do the same thing as the user providing `import.meta.url` as the base which I felt like was going to be the vast majority of use cases.

I have not seen any precedent for this and this need specifically arose for this use case because we need be able to resolve filepaths provided as a string to dynamically import from a different module/environment. Our `renderModule` from ssr requires `import.meta.url` to be passed in for the same reason.

packages/labs/testing/src/lib/fixtures/csrFixture.ts
augustjk: I do not know the answer to this but when I ran the test without the `await` it passed for chromium and firefox, just not webkit. Happy to make the comment just genetically mention awaiting for next microtask to ensure contents are rendered.

packages/labs/testing/src/lib/fixtures/ssrFixture.ts
```diff
@@ -0,0 +1,152 @@
...
+    container.replaceChildren(...Array.from(fragment.body.childNodes));
+  } else {
+    // Utilize ponyfill
+    container.innerHTML = rendered;
+    hydrateShadowRoots(container);
+  }
+
+  const el = container.firstElementChild as T;
+  if (hydrate) {
+    // TODO(augustinekim) Consider handling cases where el is not a LitElement
```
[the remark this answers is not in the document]
augustjk: We'd need to bring in `hydrate` from `lit/experimental-hydrate.js` and pass in the template and container. I am unsure if that alone would be enough to hydrate any nested lit elements. If not, we'd need to walk down and remove the first `defer-hydration` we see as well.

packages/labs/testing/src/lib/worker.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const {template, modules} = workerData as Payload;
+
+await Promise.all(modules.map((module) => import(module)));
+
+// Dangerously spoof TemplateStringsArray
+const strings = template.strings;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(strings as any).raw = strings;
```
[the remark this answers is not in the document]
augustjk: This won't work because `raw` property is readonly. I did try typing out a `SpoofableTemplateStringsArray` which removes the readonly.

Something like
```ts
type SpoofableTemplateStringsArray = {
  -readonly [P in keyof TemplateStringsArray]: TemplateStringsArray[P];
};

// or

interface SpoofableTemplateStringsArray extends TemplateStringsArray {
  raw: readonly string[];
}
```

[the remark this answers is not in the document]
aomarks: How about:

```ts
(strings as unknown as {raw: TemplateStringsArray['raw']})
```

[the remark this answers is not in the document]
augustjk: I like that. Typescript seems to also be okay without the `as unknown`.

packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
```diff
@@ -0,0 +1,145 @@
...
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+
+  if (HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
+    // Browser natively supports declarative shadowroot.
+    // Shadowroots are only parsed and attached during initial HTML parsing.
+    // innerHTML will not work and must use DOMParser.
+    // See https://web.dev/declarative-shadow-dom/#parser-only
+    const fragment = new DOMParser().parseFromString(
+      // DOMParser removes the opening lit-part comment node
```
aomarks, line 85: When I do this:

```ts
new DOMParser().parseFromString("<!--before--><p>hello</p><!--after-->", "text/html")
```

The structure I see is:

```html
<!--before-->
<html>
  <head></head>
  <body>
    <p>hello</p>
    <!--after-->
  </body>
</html>
```

So I think it's not exactly that the comment is ignored, it's that the parser is trying to create a well formed HTML document. It probably keeps the leading comment at the top, because it's allowed to be there, but moves the `<p>` into the `<body>`, because that's the expected well-formed location for it. (The specific element actually matters too -- e.g. a `<p>` will get automatically placed in the `<body>`, while a `<meta>` will get placed in the `<head>`).

So I think your solution is fine, but we could consider doing `<body>${rendered}</body>` and then `fragment.body.childNodes`, to avoid the `querySelector`, and be slightly simpler.

aomarks (approved):
Looks great! Just some minor comments, LGTM otherwise.

packages/labs/testing/src/lib/fixtures/csr-fixture.ts
```diff
@@ -0,0 +1,59 @@
...
+    // Webkit:
+    // @http://localhost:8000/node_modules/@lit-labs/testing/lib/fixtures/ssrFixture.js:29:26
+    // asyncFunctionResume@[native code]
+    // asyncFunctionResume@[native code]
+    // @http://localhost:8000/test/my-element_test.js?wtr-session-id=aKWON-wBOBGyzb2CwIvmK:65:37
+    const match = new Error().stack?.match(
+      /http:\/\/localhost.+(?=\?wtr-session-id)/
+    );
+    if (!match) {
+      throw new Error('Could not find call site for csrFixture');
```
kevinpschaaf, line 40: In what cases do we expect this to fail? It's unclear in the readme if users are expected to pass `import.meta.url` or if they can reasonably expect this to work without having to pass that. Maybe clarify?

packages/labs/testing/src/lib/fixtures/csr-fixture.ts
```diff
@@ -0,0 +1,59 @@
...
+  document.body.appendChild(container);
+
+  await Promise.all(
+    modules.map((module) => import(new URL(module, base).href))
+  );
+
+  render(template, container);
+
+  // Awaiting for the next microtask to ensure contents are rendered.
+  await undefined;
```
kevinpschaaf, line 56: Not sure what the open-wc fixture does, but it might be good to wait a `requestAnimationFrame` instead, to ensure the entire tree of components (if there are multiple levels) are done rendering...?

packages/labs/testing/src/lib/worker.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Dangerously spoof TemplateStringsArray by adding back the `raw` property
+// property which gets stripped during serialization of the TemplateResult.
+// This is needed to get through the check here
+// https://github.com/lit/lit/blob/1fbd2b7a1e6da09912f5c681d2b6eaf1c4920bb4/packages/lit-html/src/lit-html.ts#L867
+const strings = template.strings;
+(strings as {raw: TemplateStringsArray['raw']}).raw = strings;
+
+let rendered = '';
+for (const str of render({...template, strings}, {deferHydration: true})) {
```
kevinpschaaf, line 28: Not familiar with how workers work... since this is all executed in module scope, if it throws, then we'll get a worker.on('error') callback?

packages/labs/testing/src/lib/lit-ssr-plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export interface Payload {
+  template: TemplateResult;
+  modules: string[];
+}
+
+export function litSsrPlugin(): TestRunnerPlugin<Payload> {
+  return {
+    name: 'lit-ssr-plugin',
+    async executeCommand({command, payload}) {
```
kevinpschaaf, line 22: Not blocking for this PR, but would love to think about how this could be factored a little differently so that it could be called from e.g. a koa/express middleware as well, so we're not tied directly to web-test-runner. Should be straightforward, I think?

kevinpschaaf (approved):
Just a few minor suggestions and clarifications; feel free to consider/resolve, but I'm good merging this based on your judgement on those.

packages/labs/testing/src/lib/fixtures/csr-fixture.ts
augustjk: https://github.com/open-wc/open-wc/blob/master/packages/testing-helpers/src/elementUpdated.js

open-wc looks for specific promises to await on and falls back to `requestAnimationFrame` if they don't exist. Perhaps we could do something similar and look for `updateComplete` and await that.

packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  const rendered = await executeServerCommand<string, Payload>(
+    litSsrPluginCommand,
+    {
+      template,
+      modules: modules.map((module) => new URL(module, base).pathname),
+    }
+  );
+
+  // TODO(augustinekim) Clean up the container from the document
```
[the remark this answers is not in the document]
augustjk: Added a `cleanupFixtures` for now.

I haven't seen cases of users creating multiple fixtures in a single test but don't want to discount that it might happen. I don't feel like trying to automatically clean up is important enough to potentially break that particular use case.

packages/labs/testing/src/lib/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export const nextFrame = () =>
+  new Promise((resolve) => requestAnimationFrame(resolve));
+
+/**
+ * Strips Lit expression comments from provided html string.
+ */
+export const stripExpressionComments = (html: string) =>
```
aomarks, line 13: I think we should have some tests for `stripExpressionComments` and `stripExpressionMarkers` probably. Can be an issue/follow up PR though.

packages/labs/testing/src/lib/worker.ts
augustjk: Yes that's correct. Currently the wtr server command simply fails for those cases so the fixture function call will then throw. We see that happen in the bad-element test.

packages/labs/testing/src/lib/utils.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export const nextFrame = () =>
```
aomarks, line 7: Do we want to vend this? It's very basic and not specific to Lit.

packages/labs/testing/README.md
```diff
@@ -31,7 +31,7 @@ Add the plugin to your config file.
 
 ```js
 // web-test-runner.config.js
-import {litSsrPlugin} from '@lit-labs/testing';
+import {litSsrPlugin} from '@lit-labs/testing/web-test-runner.js';
```
aomarks, line 34: I wonder if this should be more specific a module? Like `@lit-labs/testing/web-test-runner-ssr-plugin.js`? If we ever had another web test runner plugin, it might be nice to keep the module graph distinct.

aomarks (approved):
LGTM pending 3 comments!

packages/labs/testing/src/lib/utils.ts
augustjk: Are you asking if we should or should not vend it? Currently only the strip functions are exported directly from the index. The `nextFrame` would be accessible if the user digs into `lib/utils.js` though.

packages/labs/testing/README.md
augustjk: Hmm the name is kind of getting long but I see your point about wanted to keep it distinct. `@lit-labs/testing/wtr-ssr-plugin.js`? I don't know if `wtr` is a common enough acronym that's as recognizable.

aomarks: Either is OK with me.

packages/labs/testing/src/lib/utils.ts
aomarks: Oh I see, that's fine.

packages/labs/testing/README.md
augustjk: Went with the longer more descriptive name.

packages/labs/testing/src/lib/utils.ts
augustjk: Filed https://github.com/lit/lit/issues/3102

packages/labs/testing/src/lib/fixtures/csr-fixture.ts
augustjk: I don't anticipate us to ever reach this unless how the stack is generated gets changed (or the regex somehow fails for some browser besides chrome, firefox, webkit) or if web test runner changes the query to something other than `wtr-session-id`.

I've updated the readme to expand more on the `base` option. Leaving it empty should have the same behavior as passing in `import.meta.url`.

PR #2960, opened 2022-05-26
[labs/react] Fix use-controller.js not being exported from package.json

AndrewJakubowicz:
Fixes https://github.com/lit/lit/issues/2948


taylor-vann:
Very coo, very clean PRl! I don't have any issue submitting. Just need to confirm this part of the package is to be made public at the moment

justinfagnani (approved):

PR #2972, opened 2022-05-31
[localize] Fix multiple runtime msg call with same template but different expressions

### LINKED ISSUE #2971: [localize] msg function substitutes last used parameter
# Description

`msg` function call substitutes last used parameter.

# Steps to Reproduce

1. Write this code

```js
import {html, render} from 'lit';
import {msg} from '@lit/localize';

render(html`
<ol>
<li>${renderItem('ABC')}</li>
<li>${renderItem('CDE')}</li>
<li>${renderItem('123')}</li>
<li>${renderItem('456')}</li>
<ol>
`, document.body);

function renderItem(param) {
  return [
    html`Some text with param <b>${param}</b>`,
    ' ',
    msg(html`Some text with param <b>${param}</b>`),
  ];
}
```

2. See this output...

## Live Reproduction Link

https://stackblitz.com/edit/web-platform-bxitxm

## Expected Results

```
1. Some text with param ABC. Du texte avec paramètre ABC.
2. Some text with param CDE. Du texte avec paramètre CDE.
3. Some text with param 123. Du texte avec paramètre 123.
4. Some text with param 456. Du texte avec paramètre 456.
```

## Actual Results

```
1. Some text with param ABC. Du texte avec paramètre 456.
2. Some text with param CDE. Du texte avec paramètre 456.
3. Some text with param 123. Du texte avec paramètre 456.
3. Some text with param 456. Du texte avec paramètre 456.
```

## Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

augustjk:
Fixes #2971 

We were previously reusing the localized template result object mapped by id for every runtime `msg` call and mutating the values property on it. When multiple `msg` call with template strings that result in the same id were present, it would have the same, latest assigned, value. Changed to return a new object, with the value replaced for each invocation of `msg`.

### DIFF
--- a/packages/localize/src/tests/runtime.test.ts
+++ b/packages/localize/src/tests/runtime.test.ts
@@ -128,6 +128,24 @@ suite('runtime localization configuration', () => {
       assert.equal(container.textContent, 'Hola friend');
     });
 
+    test('renders multiple parameterized template with different expressions in Spanish', async () => {
+      const loaded = setLocale('es-419');
+      lastLoadLocaleResponse.resolve(spanishModule);
+      await loaded;
+
+      const name = 'friend';
+      const name2 = 'buddy';
+      render(
+        html`${msg(html`Hello ${name}`, {
+          id: 'parameterized',
+        })}${' '}${msg(html`Hello ${name2}`, {
+          id: 'parameterized',
+        })}`,
+        container
+      );
+      assert.equal(container.textContent, 'Hola friend Hola buddy');
+    });
+
     test('renders parameterized string template where expression order changed', async () => {
       const loaded = setLocale('es-419');
       lastLoadLocaleResponse.resolve(spanishModule);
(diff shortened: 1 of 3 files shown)


packages/localize/src/tests/runtime.test.ts
```diff
@@ -128,6 +128,24 @@ suite('runtime localization configuration', () => {
       assert.equal(container.textContent, 'Hola friend');
     });
 
+    test('renders multiple parameterized template with different expressions in Spanish', async () => {
```
aomarks, line 131: Mind also adding a similar test, but for a parameterized string without the `html` tag? The implementation is very different so this bug shouldn't manifest there, but would be nice to have a test anyway.

aomarks (approved):
Awesome quick diagnose and fix!

LGTM, just a request for one extra test.

packages/localize/src/tests/runtime.test.ts
augustjk: Thanks! I added more parameterized string template test cases that mirror the html test cases.

aomarks (approved):
Lgtm!

PR #2976, opened 2022-06-01
[labs/analyzer] Add Type and Reference to model

### LINKED ISSUE #3000: [labs/analyzer] Add support for types in analysis

kevinpschaaf:
Adds a `Type` class to the analysis model, which can be used by various other model classes (e.g. `Event`, `ReactiveProperty`, etc.) as well as a `Reference` class, which can point to symbols referenced in `Type`s (amongst others).

To generate references, we turn a `ts.Type` into a `ts.TypeNode` AST node and visit all `TypeReference` and `ImportType` symbols and determines the correct import for them. All of that is wrapped in a new `ProgramContext` utility class that is passed to the various model creators with utility functions for generating a `Type` model based on a `ts.Node`, etc.

In order to test this, also adds analysis of `VariableDeclaration` and a suite of type tests against a module with variously typed variable declarations.

A couple other notes:
* Adds a `DiagnosticsError` class for reporting errors associated with user source code.
* In order to create useful references, we need to know the name of the package we are analyzing, so a name in `package.json` is now required for all analysis
* Implements a gambit for generating `ts.Type` nodes from jsdoc `{Type}` strings (e.g. for `@fires`) by generating dummy variable declarations with those types, pulling off the type node, and the re-associating it back with the jsdoc tag.

Fixes #3000


packages/labs/analyzer/src/test/types_test.ts
```diff
@@ -0,0 +1,231 @@
...
+  assert.equal(type.references[3].name, 'ImportedClass');
+  assert.equal(type.references[3].package, '@lit-internal/test-types');
+  assert.equal(type.references[3].module, 'external.js');
+  assert.equal(type.references[3].isGlobal, false);
+});
+
+test('getImportStringForReferences', ({module}) => {
+  const type = typeForVariable(module, 'complexType');
+  assert.equal(
+    type.getImportStatementsForReferences(),
```
sorvell, line 223: This is great. As noted offline, we probably want to collapse and coalesce the import statements in the wrappers. That seems like something we can do as a future improvement.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -113,26 +121,51 @@ export class Analyzer {
...
     for (const statement of sourceFile.statements) {
       if (ts.isClassDeclaration(statement)) {
-        if (isLitElement(statement, this.checker)) {
+        if (isLitElement(statement, this.programContext)) {
           module.declarations.push(
-            getLitElementDeclaration(statement, this.checker)
+            getLitElementDeclaration(statement, this.programContext)
           );
-        } else {
+        } else if (isExported(statement)) {
```
sorvell, line 139: Why is this check here?

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -113,26 +121,51 @@ export class Analyzer {
...
           module.declarations.push(
             new ClassDeclaration({
               name: statement.name?.text,
               node: statement,
             })
           );
         }
+      } else if (ts.isVariableStatement(statement) && isExported(statement)) {
```
sorvell, line 147: What is this adding exactly and why?

packages/labs/analyzer/src/lib/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+    return process.cwd();
+  },
+  getNewLine() {
+    return '\n';
+  },
+};
+
+export class DiagnosticsError extends Error {
+  diagnostics: ts.Diagnostic[];
+  constructor(diagnostics: readonly ts.Diagnostic[], message?: string);
```
sorvell, line 23: Do we need all these overloads? It doesn't seem like they are used.

packages/labs/analyzer/src/lib/lit-element/events.ts
```diff
@@ -27,22 +31,26 @@ export const getEvents = (node: LitClassDeclaration) => {
...
           }
           const {name, type, description} = result;
-          // TODO(justinfagnani): how do we dereference the event type name
-          // into a TypeScript type? TypeScript will automatically do this for
-          // jsdoc tags it understands, but unfortunately @fires is not one of
-          // them.
           events.set(name, {
             name,
-            typeString: type,
+            type: type
```
sorvell, line 43: Per offline chat and to check understanding: this doesn't support generics like `CustomEvent<Foo>`. Will that fail here? And if not, can we make it fail with a diagnostic?

Let's make an issue for how to move forward on supporting this better. As discussed, we can either try to regex out a generic for the common case of `CustomEvent<Foo>` and assume that's in scope; or we can generate some TS with this code in scope and properly analyze it.

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -176,3 +195,271 @@ export const getLitModules = (analysis: Package) => {
...
+      node
+    );
+  }
+
+  /**
+   * Returns the module specifier for a declaration if it was imported,
+   * or `undefined` if the declaration was not imported.
+   */
+  getImportModuleSpecifier(declaration: ts.Node): string | undefined {
+    // TODO(kschaaf) support the various import syntaxes
```
sorvell, line 341: What isn't supported here?

sorvell (changes requested):
Good job! Some questions in the comments...

packages/labs/analyzer/src/test/types_test.ts
kevinpschaaf: +1, will leave a TODO; we should probably do that as part of actually integrating it into the fw wrappers.

packages/labs/analyzer/src/lib/analyzer.ts
kevinpschaaf: Under my (maybe wrong, unclear) assumption that the only things we should expose in the model are exported declarations (or registered custom elements, which technically don't need to be exported since they're globally registered. As Justin pointed out today, you could subclass a private, non-exported class and need a reference to it, so might need to revisit this when we get to subclassing and generating the CEM. Can add a TODO.

packages/labs/analyzer/src/lib/analyzer.ts
kevinpschaaf: > In order to test this, also adds analysis of VariableDeclaration and a suite of type tests against a module with variously typed variable declarations.

`VariableDeclaration` is one of a few more we ultimately need to add to the analyzer to fully generate the CEM: https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts#L162

packages/labs/analyzer/src/lib/errors.ts
kevinpschaaf: There's only two ways to call it (with an array of `diagnostics` or with a node and an error message)... most are the latter (where we're making a new one), but I added the former for this case, to format diagnostics coming from the initial compilation: https://github.com/lit/lit/pull/2976/files#diff-d0ce248fdc93fd9bbffed64c12cf21d16650a57ad613cbd739be0f03646c4059R86

I figured it would be nice to have them all as one error type with a `diagnostics` field so you could e.g. `try { ... } catch (e) { if (e instanceof DiagnosticError) { /* do something with `e.diagnostics` */ }`

packages/labs/analyzer/src/lib/lit-element/events.ts
kevinpschaaf: Yes, `getTypeForConstructorName` will fail with on a generic with a diagnostic error because there won't be a symbol name matching that text. But I'm working on the latter now to make this just work.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: `import {foo as bar} from 'baz'`, etc.

packages/labs/analyzer/src/test/types_test.ts
kevinpschaaf: Went ahead and added a deduping/coalescing function.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Done

packages/labs/analyzer/src/lib/lit-element/events.ts
```diff
@@ -27,22 +32,24 @@ export const getEvents = (node: LitClassDeclaration) => {
...
           }
           const {name, type, description} = result;
-          // TODO(justinfagnani): how do we dereference the event type name
-          // into a TypeScript type? TypeScript will automatically do this for
-          // jsdoc tags it understands, but unfortunately @fires is not one of
-          // them.
           events.set(name, {
             name,
-            typeString: type,
+            type: type ? programContext.getTypeForJSDocTag(tag) : undefined,
```
justinfagnani, line 44: Why pass the whole tag here and not just the type?

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -176,3 +192,66 @@ export const getLitModules = (analysis: Package) => {
...
+    this.text = text;
+    this.references = references;
+  }
+}
+
+/**
+ * Returns a deduped / coalesced string of import statements required to load
+ * the given references.
+ */
+export const getImportsStringForReferences = (references: Reference[]) => {
```
justinfagnani, line 238: I wonder if you want to return an array of some Import object for further de-duping by the consumer

packages/labs/analyzer/src/test/lit-element/events_test.ts
```diff
@@ -73,24 +78,55 @@ test('Event with type', ({element}) => {
...
+  assert.equal(event.type?.text, 'ExternalCustomEvent');
+  assert.equal(event.type?.references[0].package, '@lit-internal/test-events');
+  assert.equal(event.type?.references[0].module, 'custom-event.js');
+  assert.equal(event.type?.references[0].name, 'ExternalCustomEvent');
+});
+
+test('Event with generic custom event', ({element}) => {
+  const event = element.events.get('generic-custom-event');
+  assert.ok(event);
+  assert.equal(event.type?.text, 'CustomEvent<ExternalClass>');
```
justinfagnani, line 124: Do you want to test a more complex type? I could see an inline detail type being not-uncommon, like:

```ts
CustomEvent<{name: string, age?: number}>
```

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+    getScriptSnapshot: (fileName) => {
+      let file = cache[fileName];
+      if (!file) {
+        if (!fs.existsSync(fileName)) {
+          return undefined;
+        }
+        file = {
+          version: 0,
+          content: ts.ScriptSnapshot.fromString(
+            fs.readFileSync(fileName).toString()
```
justinfagnani, line 37: if you give an encoding the return value is a String already, I think requiring `utf8` is fine.

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+ */
+
+import ts from 'typescript';
+import {DiagnosticsError, createDiagnostic} from './errors.js';
+import * as typedoc from 'typedoc';
+import path from 'path';
+import fs from 'fs';
+import {Module, PackageJson, Type, Reference} from './model.js';
+
+type FileCache = ts.MapLike<{version: 0; content?: ts.IScriptSnapshot}>;
```
justinfagnani, line 14: Why not use a real Map?

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+const createServiceHost = (
+  commandLine: ts.ParsedCommandLine,
+  cache: FileCache
+): ts.LanguageServiceHost => {
+  return {
+    getScriptFileNames: () => commandLine.fileNames,
+    getScriptVersion: (fileName) => cache[fileName]?.version.toString(),
+    getScriptSnapshot: (fileName) => {
+      let file = cache[fileName];
+      if (!file) {
```
justinfagnani, line 30: `file !== undefined`

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+        if (!fs.existsSync(fileName)) {
+          return undefined;
+        }
+        file = {
```
justinfagnani, line 34: Do you need to write this into the cache here? If not, I'd say add a comment on when the cache is written to.

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+  private _typedocContext!: typedoc.Context;
+  private _typedocConverter!: typedoc.Converter;
+
+  currentModule: Module | undefined = undefined;
+
+  fileCache: FileCache = {};
+
+  constructor(commandLine: ts.ParsedCommandLine, packageJson: PackageJson) {
+    const serviceHost = createServiceHost(commandLine, this.fileCache);
+    const services = ts.createLanguageService(
```
justinfagnani, line 72: s/services/service

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+      throw new Error(`Internal error: expected currentModule to be set`);
+    }
+    const declaration = symbol?.declarations?.[0];
+    if (declaration === undefined) {
+      throw new DiagnosticsError(
+        location,
+        `Could not find declaration for symbol '${name}'`
+      );
+    }
+    if (declaration.getSourceFile() !== location.getSourceFile()) {
```
justinfagnani, line 319: If the symbol is imported, will the declaration still be in the location's source file? will it be an import declaration or something? Might make a good comment.

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+      // If the reference declaration doesn't exist in this module, it must have
+      // been a global (whose declaration is in an ambient .d.ts file)
+      return new Reference({
+        name,
+        isGlobal: true,
+      });
+    } else {
+      const module = this.getImportModuleSpecifier(declaration);
+      if (module) {
```
justinfagnani, line 328: module !== undefined

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+            module: path.join(path.dirname(this.currentModule.jsPath), module),
+          });
+        } else {
+          // External import: extract the npm package (taking care to respect
+          // npm orgs) and module specifier (if any)
+          const info = module.match(
+            /^(?<package>(@\w+\/\w+)|\w+)\/?(?<module>.*)$/
+          );
+          if (!info || !info.groups) {
+            throw new DiagnosticsError(
```
justinfagnani, line 345: It think the absolute path case (`import '/...'`) will fall through to here and it might be a confusing error message.

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: So, I was seeing the declaration for an imported symbol being a `ImportSpecifier` (which is a `ts.Declaration` node) in the current file, not the declaration for wherever it was originally declared. However last night I was looking back at [your code here](https://github.com/lit/lit/blob/f8dad56b8e60b5e51510e22434031300b8e77f5a/packages/labs/analyzer/src/lib/lit-element/lit-element.ts#L41-L47), where the declaration for a symbol in the base type is coming from a different file, which made me slightly question that. 

I don't know whether it's just because `getBaseTypes` might return types with symbols resolved to their original declaration? The difference may be that this symbol is coming from `getSymbolsInScope`, which maybe has a `symbol.declaration[0]` that's the most local one?

packages/labs/analyzer/src/lib/program-context.ts
```diff
@@ -0,0 +1,478 @@
...
+        location,
+        `Could not find declaration for symbol '${name}'`
+      );
+    }
+    if (declaration.getSourceFile() !== location.getSourceFile()) {
+      // If the reference declaration doesn't exist in this module, it must have
+      // been a global (whose declaration is in an ambient .d.ts file)
+      return new Reference({
+        name,
+        isGlobal: true,
```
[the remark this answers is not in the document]
kevinpschaaf: Yeah, I guess we could detect if the .d.ts file is a well-known lib one, and assume it's global in that case?

packages/labs/analyzer/src/lib/program-context.ts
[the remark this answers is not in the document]
kevinpschaaf: Yeah, I was thinking about this this morning, for the language service case we probably want to do this. The hard part is figuring out the appropriate spots to catch and continue... I'll take a crack.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -21,9 +28,8 @@ export {PackageJson};
 export class Analyzer {
   readonly packageRoot: AbsolutePath;
   readonly commandLine: ts.ParsedCommandLine;
-  readonly program: ts.Program;
-  readonly checker: ts.TypeChecker;
   readonly packageJson: PackageJson;
+  readonly programContext: ProgramContext;
```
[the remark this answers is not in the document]
kevinpschaaf: It's mostly FUD about avoiding circular imports in our internal toolchain (`analyzer.ts` would `import {getEvents} from './events.js'`, and `events.ts` would `import {Analyzer} from './analyzer.js'` (to type the `analyzer` argument to `getEvents`). Checking with @rictic whether this is actually an issue to avoid in g3 or whether I just dreamed it. Maybe it's not an issue if the circular dep is just the type? Or if they're all in the same ts_library? 

Otherwise yeah, this could all just be in the analyzer. FWIW I also want to factor `Module` and the various `Declaration` features out to their own module, started that here (will wait till this one lands): https://github.com/lit/lit/pull/2994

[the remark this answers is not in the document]
kevinpschaaf: Peter thinks circular refs are likely going to be stressful. For now, added a TODO and leaving this as-is. Totally up for refactoring it later.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -131,8 +135,29 @@ export class Analyzer {
             })
           );
         }
+        // TODO(kschaaf) should we only analyze exported things?
+      } else if (ts.isVariableStatement(statement) && isExported(statement)) {
+        module.declarations.push(
+          ...statement.declarationList.declarations
+            .filter((dec) => ts.isIdentifier(dec.name))
```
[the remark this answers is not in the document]
kevinpschaaf: Ok, so they're `ObjectBindingPattern | ArrayBindingPattern` (for object/array destructuring). Added support for that including tests, as well as for declaring non-exported variables and exporting them separately (actually differentiating the exported vs. non-exported things for e.g. the manifest will come later).

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -131,8 +135,29 @@ export class Analyzer {
...
       }
     }
+    this.programContext.currentModule = undefined;
     return module;
   }
 }
+
+const isExported = (node: ts.Statement): boolean => {
+  return (
+    node.modifiers?.some((m) => m.kind === ts.SyntaxKind.ExportKeyword) ?? false
```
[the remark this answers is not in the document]
kevinpschaaf: +1, that ended up being related to the point above so went ahead and just added a test for that (`separatelyExportedClass` test).

packages/labs/analyzer/src/lib/lit-element/events.ts
kevinpschaaf: The WeakMap holding the pre-processed types for jsdoc tags are keyed off the tag.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Added a TODO. Let's cover that when we integrate it into the generators.

packages/labs/analyzer/src/test/lit-element/events_test.ts
kevinpschaaf: So I added this test, but for some reason the `typedoc` visitor never gets to type references inside the interface. Commented out some asserts in the test and added TODO w/issue for replacing the Type visitor: https://github.com/lit/lit/issues/3001

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: +1 Done

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Heh, this was all just based on code cribbed from the TS wiki. Changed to Map.

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Done

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Yeah it should be, good catch. Fixed.

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Done

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Ok, so looks like you get multiple declarations for:
* Classes (the interface + the constructor)
* Namespace merging (I think)
* ... but not automatically for the original declaration

It looks like you have to explicitly call `getAliasedSymbol()` to get the original declaration. So I think the current logic works, for now.

Added a good comment. 😉

packages/labs/analyzer/src/lib/program-context.ts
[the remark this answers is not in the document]
kevinpschaaf: I'll add a TODO. Will likely come up when fleshing out the manifest generator.

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Done

packages/labs/analyzer/src/lib/program-context.ts
kevinpschaaf: Added support for absolute and href specifiers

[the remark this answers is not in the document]
kevinpschaaf: Actually, will just file an issue. https://github.com/lit/lit/issues/3009

justinfagnani (approved):

PR #2979, opened 2022-06-01
[labs/context] Fixes #2964 and #2965

### LINKED ISSUE #2964: [labs/context] `contextProvider` decorator only provides to last instance
When multiple instances of an element that uses the `contextProvider` decorator are created, only the last created element actually updates the context value properly.

Here's a [repro of the issue](https://lit.dev/playground/#gist=0823cae244a37fb46691d06313f3a231). 
* Expected: typing in an input should update the provided value and render it in the receiving element
* Actual: only the bottom receiving element is updated.

**Explanation**
The decorator stores a reference to a ContextProvider controller and uses it in the property setter. However, the controller is cached [in the decorator finisher](https://github.com/lit/lit/blob/main/packages/labs/context/src/lib/decorators/context-provider.ts#L57) and this runs only once when the class is setup. This single cached value is set for each instance created and therefore the cached controller for all instances always becomes the controller for the last instance created.

One way to fix this would be to find the controller for the element by using a WeakMap keyed on the element.

### LINKED ISSUE #2965: [labs/context] Readme incorrectly references `contextRequest` instead of `contextProvided`
The README section on  consuming a context value, references a non-existent decorator named [@contextRequest](https://github.com/lit/lit/blob/main/packages/labs/context/README.md#my-elementts). The actual decorator name is [@contextProvided](https://github.com/lit/lit/blob/main/packages/labs/context/src/index.ts#L19).

sorvell:
* Fixes #2964: @contextProvider now uses a WeakMap to find the controller associated with the given element. This was previously not looked up by element and therefore all setters would find the controller for the last instantiated instance
* Fixes #2965: Changes incorrect references to `@contextRequest` to the proper name `@contextProvided`.

### DIFF
--- a/packages/labs/context/src/test/context-provider_test.ts
+++ b/packages/labs/context/src/test/context-provider_test.ts
@@ -40,14 +40,14 @@ class ContextProviderElement extends LitElement {
 }
 customElements.define('context-provider', ContextProviderElement);
 
-suite('@contextRequest', () => {
+suite('@contextProvided', () => {
   let consumer: ContextConsumerElement;
   let provider: ContextProviderElement;
   let container: HTMLElement;
   setup(async () => {
     container = document.createElement('div');
     container.innerHTML = `
-        <context-provider value="1000">            
+        <context-provider value="1000">
             <context-consumer></context-consumer>
         </context-provider>
     `;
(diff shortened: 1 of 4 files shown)


kevinpschaaf (approved):
LGTM

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -82,3 +82,57 @@ suite('@contextRequest', () => {
...
+    ) as ContextProviderElement[];
+
+    consumers = Array.from(
+      container.querySelectorAll('context-consumer')
+    ) as ContextConsumerElement[];
+
+    await Promise.all(providers.map((el) => el.updateComplete));
+    await Promise.all(consumers.map((el) => el.updateComplete));
+
+    consumers.forEach((c) => assert.isDefined(c));
```
justinfagnani, line 115: Seems weird to have an assert in setup

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -82,3 +82,57 @@ suite('@contextRequest', () => {
...
+
+    providers = Array.from(
+      container.querySelectorAll('context-provider')
+    ) as ContextProviderElement[];
+
+    consumers = Array.from(
+      container.querySelectorAll('context-consumer')
+    ) as ContextConsumerElement[];
+
+    await Promise.all(providers.map((el) => el.updateComplete));
```
justinfagnani, line 112:
```suggestion
    await Promise.all([...providers, ...consumers].map((el) => el.updateComplete));
```

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -82,3 +82,57 @@ suite('@contextRequest', () => {
...
 });
+
+suite('@contextProvided: multiple instances', () => {
+  let consumers: ContextConsumerElement[];
+  let providers: ContextProviderElement[];
+  let container: HTMLElement;
+  const count = 3;
+  setup(async () => {
+    container = document.createElement('div');
+    container.innerHTML = new Array(count)
```
justinfagnani, line 93: A neat alternate pattern:

```ts
Array.from({length: count}, (_, i) => `...`)
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#using_arrow_functions_and_array.from

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -82,3 +82,57 @@ suite('@contextRequest', () => {
...
+      .map(
+        (_v, i) => `
+        <context-provider value="${1000 + i}">
+            <context-consumer></context-consumer>
+        </context-provider>`
+      )
+      .join('/n');
+    document.body.appendChild(container);
+
+    providers = Array.from(
```
justinfagnani, line 104: you don't need the Array.from if you spread below. `querySelectorAll` is generic too, so you can do `querySelectorAll<ContextProviderElement>('context-provider')`

justinfagnani (approved):

packages/labs/context/src/test/context-provider_test.ts
sorvell: Removed.

packages/labs/context/src/test/context-provider_test.ts
sorvell: Neat

packages/labs/context/src/test/context-provider_test.ts
sorvell: Done.

PR #2962, opened 2022-05-27
@lit-labs/virtualizer package migration from PolymerLabs/uni-virtualizer

usergenic:
This PR brings in the `@lit-labs/virtualizer` package from the https://github.com/PolymerLabs/uni-virtualizer repo.  All other files from uni-virtualizer repo that were imported were not kept.

I prepared a migration script so I could refine this until it worked exactly the way I expected with commit history and everything in its place in the lit monorepo.  The `migrate.sh` script and its output can be found in this gist if interested:

https://gist.github.com/usergenic/226e876af976905ba557f12c8f4c886f

Verifying commit history is present, here is snippet from output of git blame on the package's README:

```shell
$ git blame packages/labs/virtualizer/README.md
e03b71c1e packages/lit-virtualizer/README.md  (Gray Norton     2021-05-07 15:00:26 -0700   1) # @lit-labs/virtualizer
9db07ef51 packages/lit-virtualizer/README.md  (straversi       2019-06-05 14:22:52 -0700   2) 
6261458f8 packages/lit-virtualizer/README.md  (Gray Norton     2022-03-22 13:31:49 -0700   3) `@lit-labs/virtualizer` provides viewport-based virtualization (including virtual scrolling) for [Lit](https://lit.dev).
9db07ef51 packages/lit-virtualizer/README.md  (straversi       2019-06-05 14:22:52 -0700   4) 
2754a144f packages/lit-virtualizer/README.md  (Gray Norton     2022-03-22 13:08:55 -0700   5) ⚠️ `@lit-labs/virtualizer` is in late prerelease. Its API is intended to remain quite stable going forward, but you should expect (increasingly minor) changes before 1.0. Some of these changes may be technically breaking, but we anticipate that they will be mechanical and straightforward to make.
ff7f2ca07 packages/lit-virtualizer/README.md  (straversi       2019-06-25 15:42:39 -0700   6) 
0af8bb7e8 packages/lit-virtualizer/README.md  (straversi       2019-06-05 15:07:31 -0700   7) ## Getting Started
9db07ef51 packages/lit-virtualizer/README.md  (straversi       2019-06-05 14:22:52 -0700   8) 
0af8bb7e8 packages/lit-virtualizer/README.md  (straversi       2019-06-05 15:07:31 -0700   9) Get this package:
9db07ef51 packages/lit-virtualizer/README.md  (straversi       2019-06-05 14:22:52 -0700  10) 
[...]

PR #2982, opened 2022-06-01
[infra/analyzer] Fix analyzer tests

augustjk:
### Changes
- Add `test-files/` to wireit `files` config for the `test` script.
- Add try catch block for all `test.before` callbacks that'll exit with status 1 if an error is caught.
- Add path normalization in a couple of the test files for Windows.

### Context
I noticed here https://github.com/lit/lit/pull/2962#pullrequestreview-991297057 that the analyzer tests were passing from cached results even though the `test-files` directory was erroneously deleted.
During testing this, I also noticed that errors in uvu's `before` callback don't actually cause the test script to fail due to an open issue with uvu. I copied the try/catch check from wireit's test code.
Above change surfaced that `events_test.ts` and `properties_test.ts` have always been failing in Windows due to the path check with a hardcoded `/`. Replaced it with `path.normalize` as it was being done in other test files.


kevinpschaaf (approved):
LGTM, Thank you, this is great! The uvu error swallowing is pretty insidious.

Issue #2734, opened 2022-04-15
[@labs/react] Support adding children to multiple slots


taylor-vann:
## Description

Currently, the react wrapper supports adding children to a default slot.

In order to support multiple slots from react -> lit wrapper, the react dev might have to wrap their components in an html element as an attribute.

```
() => (
  <div slot="slotA">
    <MyReactComponent />
  </div>
  <div slot="slotB">
    <MyReactComponent />
  </div>
  </div slot="slotC">
    <MyReactComponent />
  </div>
)
```

It'd be nice to support multiple slots in a shallow manner with react

```
export const MyElementComponent = createComponent(
  React,
  'my-element',
  MyElement,
  {
    onChange: 'input',
  },
  {
    propName: 'slot_name',
    icons: 'icons',
    suffix: 'suffix',
  }
);
```

However, rendering might be an issue between react and lit. The major hurdle is setting the `slot` attribute on an html element rendered in react.

## Acceptance criteria

1) Decide if this is a feature the labs react wrapper should support

2) If we go forward, we need to agree on how to extend our API

3) Provably use slotted react components in multiple web component slots

bugwheels94:
You can use cloneElement in order to avoid wrapping in another component. Though, primitives like number, string etc. needs to be wrapped

import React, { ReactNode, ReactElement, HTMLAttributes } from 'react';
import { createComponent } from '../../utils/createComponent';
import { FkMenuOption } from './lit';
const LitReactComponent = createComponent(React, 'et-menu-option', FkMenuOption, {});

export const Option = ({
	icon,
	children,
	...props
}: {
	icon?: ReactElement;
	children: ReactNode;
} & React.ComponentProps<typeof S>) => {
	return (
		<LitReactComponent {...props}>
			{React.isValidElement(icon) ? React.cloneElement(icon, { slot: 'icon' } as HTMLAttributes<unknown>) : icon}
			{children}
		</S>
	);
};

taylor-vann:
Thanks for the suggestion @bugwheels94 !
I think for now, the path with the least amount of friction is to regard the `slot` as apart of the components API.

In web component land, elements project themselves onto a component's slots. In react land, children are another prop. So there isn't a great 1:1 relationship (as opposed to events or attributes).

I think the best way moving forward here is the following example:
```
import {MyWebComponent} from "./my-web-component";

const myReactComponent = () => <div slot="named-slot"></div>;
const myFrameComponent = () => (
  <MyWebComponent>
[...]

PR #2985, opened 2022-06-02
[@labs/react] support multiple children with multiple slots

taylor-vann:
Related to https://github.com/lit/lit/issues/2734

This PR supports multiple children by assigning them to multiple slots by wrapping them in a `<div styles="display: contents">`. 

A client can provide a `children` map similar to how we account for `events`.
The syntax is as follows:
```ts
const BasicElementComponent = createComponent(
  window.React,
  elementName,
  BasicElement,
  basicElementEvents,
  displayName,
  { <react_children_property_name>: <slot_name> },
);
```

Which results in a `children` config like:
```ts
const BasicElementComponent = createComponent(
  window.React,
  elementName,
  BasicElement,
  basicElementEvents,
  displayName,
  {
    headIcons: 'head-icons',
    content: 'content',
    tailIcons: 'tail-icons',
  },
);
```

Tests are added to verify rendered elements are rendered with the correct `slot` names.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -108,11 +108,11 @@ type Slots = Record<string, string>;
 type SlotMap<C> = Record<keyof C, React.ReactNode>;
 
 const slotReactNode = (
-  React: typeof ReactModule,
+  createElement: typeof ReactModule.createElement,
   slot: string,
   children: React.ReactNode
 ) => {
-  return React.createElement(
+  return createElement(
     'div',
     {
       style: {display: 'contents'},
@@ -295,7 +295,9 @@ export const createComponent = <
 
         const slot = slots?.[k];
         if (slot) {
-          children.push(slotReactNode(React, slot, v as React.ReactNode));
+          children.push(
+            slotReactNode(createElement, slot, v as React.ReactNode)
+          );
           continue;
         }
 


packages/labs/react/src/create-component.ts
```diff
@@ -127,25 +145,32 @@ type EventProps<R extends Events> = {
...
+  E extends Events,
+  C extends Record<string, string>
+>(
   React: typeof ReactModule,
   tagName: string,
   elementClass: Constructor<I>,
   events?: E,
-  displayName?: string
+  displayName?: string,
+  children?: C
```
sorvell, line 158: I think it makes more sense to call this argument `slots`.

packages/labs/react/src/create-component.ts
```diff
@@ -127,25 +145,32 @@ type EventProps<R extends Events> = {
...
 ) => {
   const Component = React.Component;
   const createElement = React.createElement;
+  const reactChildren = children;
```
sorvell, line 162: Why rename this? Let's just choose a better name as suggested above.

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -73,11 +73,18 @@ suite('createComponent', () => {
     onBar: 'bar',
   };
 
+  const basicElementChildren = {
```
sorvell, line 76: Again, recommend calling this `basicElementSlots`

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -333,6 +340,19 @@ suite('createComponent', () => {
     assert.equal(el.firstElementChild!.localName, 'div');
   });
 
+  test('can set multiple children to multiple slots', async () => {
+    const children = window.React.createElement('div');
+    const foos = window.React.createElement('div');
```
sorvell, line 345: We need to test that you can specify a a set of children for a slot. Ideally we'd be testing with JSX here as well. I think in that case it would be a JSX fragment and without JSX I guess you can make that via a `React.createElement(React.Fragment, ...`

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -333,6 +340,19 @@ suite('createComponent', () => {
...
+    const bars = window.React.createElement('div');
+
+    await renderReactComponent({children, foos, bars});
+
+    assert.equal(el.childNodes.length, 3);
+    assert.equal(el.children?.[0].getAttribute('slot'), null);
+    assert.equal(el.children?.[1].getAttribute('slot'), 'slot-a');
```
sorvell, line 352: These should be `display: contents` right? Let's test for that.

packages/labs/react/src/create-component.ts
```diff
@@ -106,6 +106,24 @@ type EventProps<R extends Events> = {
...
+  children: React.ReactNode
+) => {
+  return React.createElement(
+    'div',
+    {
+      style: {display: 'contents'},
+      key: slot,
+      slot,
+    },
+    children
```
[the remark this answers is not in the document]
sorvell: This is arguable and we should sync up on this before changing the code, but I think it's preferable to set slot attributes where possible rather than to create a wrapper DOM node.

However, if a wrapper node is needed, probably we should just make one and not set slots. So, this would involve iterating the children until you find a non-element node (text nodes or component references): if so making the wrapper; if not, cloneElement and set slots.

sorvell (changes requested):
See comments; we also need to add documentation to the readme for this.

packages/labs/react/src/create-component.ts
taylor-vann: Sounds good!

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: Yah let's! Sounds good

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: Nice! Yah I'll add that test case

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
taylor-vann: I believe React.ReactNode can be any value that react will render: string array ref node element. Children are passed through a single property name in react / jsx (or as "children"). So when a `child` is assed to `createSlottedChild` it wraps the entire prop to a slot via one wrap.

I figured this was less operations for the same behavior, choosing to wrap the entire children prop in a single function than search and create multiple. If the CSS `display contents` let's us avoid styling I figured this would be the most direct way to slot quickly and cheaply

A clarifying question: Are we expect a user to pass in some value than a React.ReactNode?

taylor-vann:
Ugh dumb thumbs, how do I cancel this review ;_;

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
sorvell: Creating the wrapper DOM node itself is likely the most expensive part. If we can avoid it and just set slot attributes, we probably should. I wanted to explore what was possible via their API and so made [this experiment](https://stackblitz.com/edit/react-ts-pfwb7c?file=slots.ts,App.tsx) which might be useful.

khanh-le-otsv:
Could I ask for why this is not being merged? Thank you.

PR #2987, opened 2022-06-03
[labs/react] useController: handle strict mode

### LINKED ISSUE #2950: [@lit-labs/react] use-controller.js hook state not updating
# Description

I've created a [stackblitz example](https://stackblitz.com/edit/vitejs-vite-sem6tx?file=src%2FApp.tsx,src%2Flit%2Flib%2Fuse-controller.ts,src%2Flit%2Fcomponents%2Flit-controller-example.ts,src%2Fcomponents%2FLitControllerExample.ts,src%2Fcomponents%2FReactHookExample.tsx&terminal=dev).

NOTE: use-controller was recreated locally in '/src/lit/lib/use-controller.ts' due to https://github.com/lit/lit/issues/2948

There are two controllers: mouse-controller and clock-controller.

First example is a lit element that uses the two controllers (wrapped into react component).
Second example is a react component using the two hooks (wrapped using use-controller).

In the second example using the wrapped hooks, the state is not updating for some reason.

image

sorvell:
Fixes #2950. The `useController` hook uses `useState` to create a controller. React expects `useState` not to have side effects and this means that a controller's `hostConnected`, which typically produces side effects, must be specially handled. The controller lifecycle `connected + update` should run before the `useController` returns to ensure the controller's value is as expected in component's render call. However, to simulate render being pure, a `hostDisconnected` is scheduled to occur. If the component's effects are processed, this disconnection is cancelled via `useLayoutEffect`. In addition, this `useLayoutEffect` handles disconnect and connect when the component is unmounted/remounted. 

Also adds a test that uses React.StrictMode to verify this behavior.


packages/labs/react/src/use-controller.ts
```diff
@@ -131,23 +136,46 @@ export const useController = <C extends ReactiveController>(
...
+  // on the controller.
+  useLayoutEffect(() => {
+    host._connected();
+    host._update();
+    return () => host._disconnected();
+  }, []);
+
+  // We use useLayoutEffect because we need updated() called synchronously
+  // after rendering.
+  useLayoutEffect(() => host._updated());
```
kevinpschaaf, line 171: Why two `useLayoutEffect` calls?

packages/labs/react/README.md
```diff
@@ -127,6 +127,12 @@ Controller timings are implemented as follows:
 | requestUpdate    | useState setter                     |
 | updateComplete   | useLayoutEffect                     |
 
+### Limitations
+
+Because the controller's lifecycle is run via a `useLayoutEffect` callback, the initial value a controller produces for the React component may
+be incorrect. This is the case only if the controller sets up its initial value via `hostConnected` or `hostUpdate`. This limitation can be worked
```
justinfagnani, line 133: I wouldn't phrase this as "incorrect" here, but say that in React `hostConnected` is run after `hostUpdate` and to be careful about dependencies between the two. Ideally values produced by the controller are computed on demand so the order doesn't matter.

packages/labs/react/src/test/use-controller_test.ts
```diff
@@ -69,33 +69,63 @@ suite('useController', () => {
...
-    const render = (props: any) => {
-      ReactDOM.render(React.createElement(TestComponent, props), container);
-    };
-
-    render({x: 1});
-    assert.equal(ctorCallCount, 1);
-    assert.equal(container.innerHTML, `<div class="foo">x:1, a:a</div>`);
-    assert.deepEqual(testController.log, ['connected', 'update', 'updated']);
-    const firstTestController = testController;
+  const lifeCycleTest = ({strict}: {strict: boolean}) => {
```
justinfagnani, line 72: Why don't we just only test under strict mode?

packages/labs/react/src/use-controller.ts
```diff
@@ -131,23 +136,46 @@ export const useController = <C extends ReactiveController>(
...
 
-  // TODO (justinfagnani): don't call in SSR
-  host._update();
+  // Only call update if connected; it's called explicitly above when
+  // connecting.
+  if (host._isConnected) {
+    // TODO (justinfagnani): don't call in SSR
+    host._update();
```
kevinpschaaf, line 177: Why is this here and in `useLayoutEffect`?

packages/labs/react/src/use-controller.ts
```diff
@@ -131,23 +136,46 @@ export const useController = <C extends ReactiveController>(
...
+  // since the React component sees this value before the controller has run
+  // connected+update.
+  //
+  // We could add an option to do a `requestUpdate` instead of an `update`
+  // here, but this would trigger a React re-render so it should only be done
+  // if required. Users can work around this initial value limitation by using
+  // the controller value in a `useLayoutEffect` and/or calling `requestUpdate`
+  // on the controller.
+  useLayoutEffect(() => {
+    host._connected();
```
kevinpschaaf, line 164: This needs to be guarded on `_isConnected` right?

packages/labs/react/src/test/use-controller_test.ts
```diff
@@ -69,33 +69,63 @@ suite('useController', () => {
...
+      let testController!: TestController;
+
+      let componentRenderLog: string[] = [];
+      let componentLayoutEffectLog: string[] = [];
+
+      const TestComponent = ({x}: {x: number}) => {
+        testController = useTest('a');
+        componentRenderLog = [...testController.log];
+        React.useLayoutEffect(() => {
+          componentLayoutEffectLog = [...testController.log];
```
justinfagnani, line 83: why are we testing this?

packages/labs/react/src/use-controller.ts
kevinpschaaf: When dep array is empty, only called once, got it.

packages/labs/react/src/use-controller.ts
```diff
@@ -131,23 +136,46 @@ export const useController = <C extends ReactiveController>(
...
+    host._update();
```
justinfagnani, line 165: I think it may be problematic that `_update()` is called after a render here.

packages/labs/react/src/use-controller.ts
kevinpschaaf: Got it, the guard prevents update from being called the first time.

packages/labs/react/src/use-controller.ts
kevinpschaaf: Gah, the difference in deps args makes a difference? Please add a comment ala `// This runs every render` and `// This runs only once`

sorvell: Updated with alternate gambit PTAL

packages/labs/react/README.md
sorvell: Removed as this isn't the case with the current gambit.

packages/labs/react/src/test/use-controller_test.ts
sorvell: Seems important to test both since the behavior is different.

packages/labs/react/src/test/use-controller_test.ts
sorvell: To validate the state of the controller if the user accesses it in a `useLayoutEffect`.

packages/labs/react/src/use-controller.ts
sorvell: Updated with alternate gambit PTAL

kevinpschaaf (approved):
Seems ok?!

packages/labs/react/src/test/use-controller_test.ts
justinfagnani: Can you add a comment on what this is doing? Reading it it seems like it may be testing that useLayoutEffect is called in the React component. The key part is that it's checking the log state?

justinfagnani (approved):

PR #2988, opened 2022-06-04
[labs/react] introduce a options object

taylor-vann:
This PR introduces a `params` object for the @labs/react package.

### Why

API functions that rely on 3+ arguments are typically more difficult to extend as a library matures.

Mainly, param objects eliminate ordering as a concern as well as prevent passing `undefined` as an argument.
ie:
```TS
const MyComponent = createComponent(
  React,
  'tag-name',
  MyElement,
  undefined,
  undefined,
  newFeatureArg,
);
```

Since `createComponent` is called once per component in an application, creating an `params` object once per component is not very expensive.

### How

The params object looks like this:
```TS
interface Options<I extends HTMLElement, E extends Events = {}> {
  React: typeof window.React;
  tagName: string;
  elementClass: Constructor<I>;
  events?: E;
  displayName?: string;
}
```
But it could use some review, probably better names.


packages/labs/react/src/create-component.ts
```diff
@@ -127,13 +135,13 @@ type EventProps<R extends Events> = {
...
  * registered via `customElements.define`.
  */
-export const createComponent = <I extends HTMLElement, E extends Events>(
-  React: typeof ReactModule,
-  tagName: string,
-  elementClass: Constructor<I>,
-  events?: E,
-  displayName?: string
-) => {
+export const createComponent = <I extends HTMLElement, E extends Events>({
```
justinfagnani, line 138: Let's make this new signature an additive and non-breaking change by duck-typing. Keep the old signature as an override and add the new signature as an override. Make the base function signature take a `typeof ReactModule | CreateComponentParams<I, E>` as the first argument called `optionsOrReact`, and `tagName?`, etc as the remaining argument. If `tagName` is a string, assume the first argument is a React module, if `tagName` is undefined, the first argument is a options object.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
taylor-vann: ill give it a try! thanks @justinfagnani

justinfagnani:
Isn't the return type just `Component<ComponentProps>`? (ie, ForwardRef shouldn't change the external type). Can we just extract this into a helper type?:

```ts
  // Props the user is allowed to use, includes standard attributes, children,
  // ref, as well as special event and element properties.
  type ReactProps = Omit<React.HTMLAttributes<I>, keyof E>;
  type ElementWithoutPropsOrEvents = Omit<I, keyof E | keyof ReactProps>;
  type UserProps = Partial<
    ReactProps & ElementWithoutPropsOrEvents & EventProps<E>
  >;

  // Props used by this component wrapper. This is the UserProps and the
  // special `__forwardedRef` property. Note, this ref is special because
  // it's both needed in this component to get access to the rendered element
  // and must fulfill any ref passed by the user.
  type ComponentProps = UserProps & {
    __forwardedRef?: React.Ref<unknown>;
  };
```

taylor-vann:
@justinfagnani will def try! Dunno why I got stuck on a 1:1 typing. You're correct that should work!

packages/labs/react/src/create-component.ts
```diff
@@ -55,6 +55,14 @@ export type ReactWebComponent<
   React.PropsWithoutRef<WebComponentProps<I, E>> & React.RefAttributes<I>
 >;
 
+interface Options<I extends HTMLElement, E extends EventNames = {}> {
+  tagName: string;
+  elementClass: Constructor<I>;
+  React: typeof window.React;
```
justinfagnani, line 61: `React` -> `react`

justinfagnani (approved):

packages/labs/react/src/create-component.ts
taylor-vann: done :)

kevinpschaaf (approved):

PR #2994, opened 2022-06-06
[labs/analyzer] Factor Module, Classes, Variables into separate modules

kevinpschaaf:


packages/labs/analyzer/src/lib/javascript/variables.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @fileoverview
+ *
+ * Utilities for working with classes
```
sorvell, line 10:
```suggestion
 * Utilities for working with variables
```

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
sorvell, line 1: Naming the containing folder "javascript" seems a little obtuse. Can you describe the contents more?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+/**
+ * @fileoverview
+ *
+ * Utilities for working with classes
+ */
+
+import ts from 'typescript';
+import {ClassDeclaration} from '../model.js';
```
sorvell, line 14: It seems a little strange that you're factoring out a factory for getting a `ClassDeclaration`, but the `ClassDeclaration` class itself is still defined in the monolithic "model.js" module. Should that be broken up as well and perhaps go here?

packages/labs/analyzer/src/lib/javascript/variables.ts
sorvell, line 10:
```suggestion
 * Utilities for working with variables
```

sorvell (approved):
Seems minor but, just a few comments, consider optional. I'm also not super familiar with the code but no red flags.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -42,14 +30,15 @@ export class Analyzer {
...
     // to analyzePackage() or move it to an async factory function that
     // passes these to the constructor as arguments.
+    let packageJson;
     try {
-      this.packageJson = JSON.parse(
+      packageJson = JSON.parse(
         fs.readFileSync(path.join(packageRoot, 'package.json'), 'utf8')
       );
     } catch (e) {
       throw new Error(`package.json not found in ${packageRoot}`);
```
justinfagnani, line 39: btw, there's two potential errors here: no file and malformed JSON.

packages/labs/analyzer/src/lib/javascript/classes.ts
justinfagnani: I might name it "typescript" - but the context is analysis utilities for TS, as opposed to possibly future utilities for custom-elements-manifest, CSS, or HTML.

packages/labs/analyzer/src/lib/javascript/classes.ts
justinfagnani: I like a separation between the model and the analysis functions that produce them. We'll have other production code paths when we add custom-element-manifest into the mix soon.

justinfagnani (approved):
I'd name the folder "typescript" rather than "javascript". Otherwise, LGTM.

packages/labs/analyzer/src/lib/analyzer.ts
kevinpschaaf: Fixed

PR #2995, opened 2022-06-06
[analyzer] Add support for superClass and mixins

kevinpschaaf:


packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,194 @@
...
+    throw new DiagnosticsError(
+      extendsClause,
+      'Did not find a "superClass" argument used in the extends clause of mixin class.'
+    );
+  }
+
+  return new MixinDeclaration({
+    node: fn,
+    name: name.getText(),
+    superClassArgIdx,
```
justinfagnani, line 171: I'd prefer `superClassArgIndex`

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,194 @@
...
+  fn: ts.FunctionLikeDeclaration,
+  name: ts.Identifier,
+  analyzer: Analyzer,
+  hasMixinHint = nodeHasMixinHint(fn)
+) => {
+  if (!fn.parameters || fn.parameters.length < 1) {
+    return throwIfMixin(
+      hasMixinHint,
+      fn,
+      `Expected mixin to have a superClass parameter.`
```
justinfagnani, line 95: `superClass` spelled and cased that way sounds like it's a specific variable/parameter name. Maybe "have a parameter for the superclass"?

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  return new MixinDeclaration({
+    node: fn,
+    name: name.getText(),
+    superClassArgIdx,
+    classDeclaration: getClassDeclaration(classDeclaration, analyzer),
+  });
+};
+
+const findSuperClassArgIndexFromHeritage = (
```
justinfagnani, line 176: Comments on what's support here would be great too. I'm wondering about cases like:

```ts
const M = (A, B) => class extends A(B) { ... }
```

Seems like that will return `1` because `B` is the argument? I guess that's about right, though we do you return `A` as another applied mixin anywhere? I think it's fine if this is case we don't support, btw. It'd be nice if we could say that this first has to be inferred as a mixin by typescript.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,194 @@
...
+      fn,
+      `Expected mixin to have a block function body.`
+    );
+  }
+  if (!ts.isBlock(functionBody)) {
+    return throwIfMixin(
+      hasMixinHint,
+      functionBody,
+      `Expected mixin to have a block function body; arrow-function class ` +
+        `expression syntax is not supported.`
```
justinfagnani, line 114: So no:

```ts
const M = (S) => class extends S {}
```

for now?

PR #3006, opened 2022-06-07
fix: outputPath is false when permalinks disabled

p-ob:
See https://github.com/11ty/eleventy/issues/653 for more info

From our Netlify deploy, which is using serverless functions:

```
Jun 7, 01:18:32 PM: 9fb35805 ERROR  [11ty] Problem writing Eleventy templates: (more in DEBUG output)
Jun 7, 01:18:32 PM: 9fb35805 ERROR  [11ty] 1. Having trouble writing to "false" from "./src/secure.njk" (via EleventyTemplateError)
Jun 7, 01:18:32 PM: 9fb35805 ERROR  [11ty] 2. Transform `render-lit` encountered an error when transforming ./src/secure.njk. (via EleventyTransformError)
Jun 7, 01:18:32 PM: 9fb35805 ERROR  [11ty] 3. outputPath.endsWith is not a function (via TypeError)
Jun 7, 01:18:32 PM: 9fb35805 ERROR  [11ty] 
```


p-ob:
Fixes #3074

augustjk (changes requested):
@p-ob Thank you very much for the PR and reporting the issue. Would you mind adding a test for this as well as a changeset? Instruction for changeset can be found in the second comment. A patch for `@lit-labs/eleventy-plugin-lit` probably makes sense here.

p-ob:
@augustjk I'm definitely open to try to add tests, but I think these will significantly alter from all other tests given the existing tests mostly just work on output files from eleventy; for this to function, I'll need to actually have a server and request for the content and not just test the file system.

augustjk:
@p-ob Thanks for adding the changeset!

> I'll need to actually have a server and request for the content and not just test the file system.

That's valid for your reproduction but I think we can also recreate the error at build time by creating a markdown file with
```
---
permalink: false
---
```
and assert the `eleventy` command to execute successfully. No output file reading necessary since there would be none.

The only thing is that it seems the eleventy process just hangs on a failed transform so the test might need a timeout for the failure case otherwise it would just hang there.

.changeset/tame-stingrays-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': patch
+---
+
+Fix using Lit SSR with Eleventy's Serverless Plugin #3074
```
augustjk, line 5: Let's make the changelog a bit more descriptive of what's beging fixed. The PR number will be automatically added by the changeset utility.
```suggestion
Fix transform breakage in situations where `outputPath` is false (e.g. setting `permalink: false` or using the serverless plugin).
```

augustjk (changes requested):
The test is great! Sorry I should've suggested this when you first added the changeset, but let's make it more descriptive and this looks good to go!

.changeset/tame-stingrays-sit.md
p-ob: Perfect! Thanks for the feedback!

augustjk (approved):
Thank you very much for the contribution!

Issue #2503, opened 2022-02-07
[meta] Integrate release notes generator into release process


aomarks:
- Update https://github.com/PolymerLabs/release-image to work with multiple package CHANGELOGs
- Integrate it into the release GitHub action
  - Simplest way is to define a GitHub artifact (https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts)
  - Even better would be to post a comment with the image embedded
- Consider moving the changelog generator into the monorepo under `scripts/` or something. It has never been released to NPM, and we probably don't want to release it.

AndrewJakubowicz:
Marking as done. Ergonomics still present a bit to be desired, however the release-image CLI has been improved to handle:
 - Multiple packages (and versions). Including multiple versions of the same package.
 - Rendering an arbitrary markdown file.
 - A github action that renders a release image for any "Version Packages" PRs.

Improving the visuals of the release-image tracked in https://github.com/lit/lit/issues/3145

PR #3007, opened 2022-06-07
[meta] Migrate release-image script into internal-scripts.

AndrewJakubowicz:
Related issue: https://github.com/lit/lit/issues/2503

After this PR, we can set up the release image generation as part of the release Github action.

### Context

As part of the release process, we want to generate release images. This PR imports a release-image script which can generate images from our CHANGELOG.md files.

This was imported from https://github.com/PolymerLabs/release-image in the first commit. The second commit adds a help message to document usage and changes the image name based on the package and version.

### Testing

This was tested manually.

Passing no arguments results in the message:

```
USAGE
    release-image CHANGELOG_PATH [--version VERSION]

EXAMPLES
    To generate the release image for reactive-element:

        release-image packages/reactive-element/CHANGELOG.md
```

Also tested from the `internal-scripts` working directory, executing `./bin/release-image.js ../reactive-element/CHANGELOG.md --version 1.3.1` generated the following file:

`reactive-element-1.3.1.png`
reactive-element-1 3 1


packages/internal-scripts/src/release-image.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const optionDefinitions = [
+  {name: 'file', type: String, defaultOption: true},
+  {name: 'version', type: String},
+];
+
+export const run = async () => {
+  const options = commandLineArgs(optionDefinitions);
+
+  if (!options.filename && !options.version) {
```
augustjk, line 30: This check always fails. It should check `options.file`, if that is meant to be required. It does mean the defaulting to `CHANGELOG.md` in line 44 no longer really makes sense, which I think is fine since our top level changelog file isn't really being updated anyway so I'd say it's okay to require the file input.

Also, `options.version` doesn't need to be checked since it'll default to latest version if not specified.

augustjk (changes requested):
Yay, thanks for bringing this in!

Is the usage of this meant to be something like
```sh
node packages/internal-scripts/bin/release-image.js packages/lit/changelog.md
```
from the root of the monorepo?

Maybe it'll be nicer to add an npm script (maybe using wireit to depend on the `packages/internal-tools:build`) so it's a bit easier to use? Oh, can't use wireit yet since the extra argument passing isn't released yet.

Eh.. on second thought maybe the script isn't necessary yet if this is an intermediary step to eventually automate the image generation anyway.

One change requested regarding the option check though!

packages/internal-scripts/src/release-image.ts
```diff
@@ -0,0 +1,159 @@
...
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
```
augustjk, line 13: I know this was carried over from the PolymerLabs org repo but when it lives here perhaps it's fine and appropriate to add the usual license header?

packages/internal-scripts/src/release-image.ts
AndrewJakubowicz: Whoops! Great catch!

packages/internal-scripts/src/release-image.ts
AndrewJakubowicz: Done! Changed to the shorter license - maintaining the correct year.

augustjk (approved):
🚀 📸 🤖

AndrewJakubowicz:
Thank you for the review! `node` is not required due to the node shebang line at the top of `release-image.js`: `#!/usr/bin/env node`. Therefore the following command is valid:

`packages/internal-scripts/bin/release-image.js packages/lit/CHANGELOG.md` from the root of the repo.


Additionally, any script depending on the `internal-scripts` package will be able to directly call `release-image` as a command that is registered via the `bin` key in package.json. Thus `release-image` can be used in the same way `treemirror` is used in the repo.

justinfagnani (approved):
Thanks! 🎉

PR #3011, opened 2022-06-09
[infra] Pin node version for windows-tools test

augustjk:
The latest patch version of node apparently ships with a version of npm that causes our windows-tools test to fail. (It logs deprecation warnings with every npm script run that messes up our cli test assertions). The latest npm version fixes that. https://github.com/npm/cli/issues/4980

However, updating npm for Windows isn't as simple as `npm i -g npm@latest` https://docs.npmjs.com/try-the-latest-stable-version-of-npm#upgrading-on-windows

It seems easier to just pin the version of node for the windows-tools test to one that shipped with working npm and unpin later once the latest v16 of node includes that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -75,7 +75,9 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: 16
+          # Pin to avoid version with problematic npm. See https://github.com/npm/cli/issues/4980
+          # TODO(augustinekim) Unpin when latest node installed by action includes fixed npm
+          node-version: 16.15.0
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
 


AndrewJakubowicz (approved):
Nice investigation and fix.

augustjk:
@rictic Do you think there might be a better way? Here's the error that's happening. https://github.com/lit/lit/runs/6822126656?check_suite_focus=true#step:8:1919

The bug is that any npm script run will always spit out
```
`npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.`
```
which the test is not expecting to see.

Speaking with @justinfagnani offline, we were wondering if it's possible to pass like a `--silent` flag to the `npm` call so that the warning above gets suppressed. Or should the test assertion be written to be able to ignore other outputs?

This is also hopefully just a temporary thing so maybe it'll be fine to pin for now and unpin soon, though at the risk of any future unexpected output resurfacing this again.

justinfagnani (approved):
Great fix! I would just file an issue to remember to unpin when we can.

PR #3024, opened 2022-06-10
Rename internal packages to @lit-internal namespace

aomarks:
This PR renames all of our internal local-only packages, like `internal-scripts` and `tests` to be within the `@lit-internal` npm namespace, which I have just registered.

The `internal-scripts` package was recently discovered to be malware, see the notice at https://www.npmjs.com/package/internal-scripts.

This would never have affected users installing our packages in the normal way, since these are only `devDependencies`. It would also not affect our own normal development workflows, because `lerna bootstrap` knows to make a symlink for this local package, instead of installing from npm. However, it could have affected somebody who accidentally did `cd packages/lit; npm install` instead of lerna bootstrap.


kevinpschaaf (approved):
LGTM, thanks for doing that!

PR #3032, opened 2022-06-14
Change lint vs changeset order

aomarks:
Currently we don't see actual linting results if the changeset test fails. This seems like the better order.


augustjk (approved):

justinfagnani (approved):
Can these run in a non-fail-fast mode?

PR #3039, opened 2022-06-15
Bump Wireit version to 0.6.1

aomarks:
Changes:

- https://github.com/google/wireit/pull/300
  - Output files are now verified before assuming a script is fresh, which should fix a number of issues, such as the switching branches with out-of-sync `.gitignore` files one.
- https://github.com/google/wireit/pull/274
  - **BREAKING** To watch a script, we should now type `npm run build --watch` instead of `npm build watch`.
  - It's now possible to add extra arguments to a script by writing `npm run build -- --extra`.


augustjk (approved):
Wonderful!

PR #3044, opened 2022-06-16
[infra] Fix bootstrap and wireit configs

augustjk:
This PR includes a few infra fix/updates:

#### Split top level lerna bootstrap command into 2 steps, without and with the starter kits.

I've been seeing a lot of bootstrap failures due to playwright failing to get exclusive locks during install both running locally and within the pipeline. Example of failing workflow https://github.com/lit/lit/runs/6889941907?check_suite_focus=true

Splitting up bootstrap like this for now seems to not cause errors when trying locally.

Extra note: Currently the `@lit-internal/tests`, `@lit/localize`, and each of the starter kits have explicit dependencies on `@web/test-runner-playwright@0.8.8` which brings in `playwright`. At some point we may want to consolidate the `@lit-internal/tests` and `@lit/localize` to share the same version.

#### Wireit config for reactive-elements

Since the wireit update, I've noticed that `build:ts` was logging that outputs were modified without manual intervention. Since `build:babel` also outputs to `development` directory, the output check detects mismatch on repeat calls. Excluding the babel output from the ts output fixes this.

#### Wireit config for labs/ssr-client

Similarly noticed output modified messages. `build:ts` incorrectly had `directives` as output, which was being modified by `build:ts:types`.
Also `checksize` did not have a wireit config for it (helpfully noticed by the Wireit VS Code plugin!)


packages/labs/ssr-client/package.json
```diff
@@ -60,7 +60,6 @@
       ],
       "output": [
         "development",
-        "directives",
```
AndrewJakubowicz, line 63: Subtle. Really great attention to detail.

AndrewJakubowicz (approved):
Great!

aomarks (approved):

Issue #2633, opened 2022-03-13
[ssr] throws when using a number as property that gets reflected to an attribute


daKmoR:
# Description

Trying to ssr a number value that gets reflected as an attribute of a LitElement throws.

# Steps to Reproduce

1. Write this code

```ts
import { render } from "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { html, LitElement, css } from 'lit';

class MyElement extends LitElement {
  static properties = {
    activeIndex: { type: Number, reflect: true, attribute: 'active-index' },
  }

  constructor() {
    super();
    this.activeIndex = 0;
  }

  render() {
    return html`
      <h2>MyElement ${this.activeIndex}</h2>
    `;
  }
}

customElements.define('my-element', MyElement);

const template = () => html`
  <my-element></my-element>
`;

const ssrResult = render(template());

let ssrString = '';
for (const s of ssrResult) {
  ssrString += s;
}

console.log(ssrString);
```

2. execute `node file.js`

## Expected Results

no error thrown

## Actual Results

```
node bug-number-property.js
file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/util/escape-html.js:19
export const escapeHtml = (str) => str.replace(/[&<>"']/g, (char) => replacements[char]);
                                       ^

TypeError: str.replace is not a function
    at escapeHtml (file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/util/escape-html.js:19:40)
    at LitElementRenderer.renderAttributes (file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/element-renderer.js:88:40)
    at renderAttributes.next (<anonymous>)
    at renderTemplateResult (file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:467:33)
    at renderTemplateResult.next (<anonymous>)
    at renderValue (file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:354:16)
    at renderValue.next (<anonymous>)
    at render (file:///html/test-ssr/node_modules/@lit-labs/ssr/lib/render-lit-html.js:336:12)
    at render.next (<anonymous>)
    at file:///html/test-ssr/bug-number-property.js:30:12

Node.js v17.6.0
```

## Suggested changed

Convert value always to string - or check explicitly for a number?

in here https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/element-renderer.ts#L147

```diff
[...]

PR #3045, opened 2022-06-16
[ssr] Cast setAttribute values to string in SSR DOM shim

aomarks:
Previously, when our SSR DOM shim for `setAttribute` received a value, it just assumed it was a string and assigned it to its internal attribute map as-is. Later, when trying to serialize an attribute to HTML, we pass it through an escape function, which crashes if the value is not a string.

In browsers, the value passed to `setAttribute` is always cast to a string. So `42` becomes `"42"`. We now mirror this behavior.

This specifically came up in https://github.com/lit/lit/issues/2633 because in `ReactiveElement`, when a property has `{type: Number}`, we do not cast the value to a string ourselves, since we assume the browser will do that automatically (in fact, we don't handle `Number` at all, since the default behavior is correct https://github.com/lit/lit/blob/21ff05104727d088ab577f7178cbc15eed14841a/packages/reactive-element/src/reactive-element.ts#L321).

Fixes https://github.com/lit/lit/issues/2633

cc @daKmoR

### DIFF
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -56,8 +56,8 @@ export const getWindow = ({
       value: string | null
     ): void;
     setAttribute(name: string, value: unknown) {
-      // The browser will silently cast all values to strings. E.g. `42` becomes
-      // `"42"` and `{}` becomes `"[object Object]""`.
+      // Emulate browser behavior that silently casts all values to string. E.g.
+      // `42` becomes `"42"` and `{}` becomes `"[object Object]""`.
       attributesForElement(this).set(name, String(value));
     }
     removeAttribute(name: string) {
--- a/packages/labs/ssr/src/lib/dom-shim.ts
+++ b/packages/labs/ssr/src/lib/dom-shim.ts
@@ -55,8 +55,10 @@ export const getWindow = ({
       old: string | null,
       value: string | null
     ): void;
-    setAttribute(name: string, value: string) {
-      attributesForElement(this).set(name, value);
+    setAttribute(name: string, value: unknown) {
+      // The browser will silently cast all values to strings. E.g. `42` becomes
+      // `"42"` and `{}` becomes `"[object Object]""`.
+      attributesForElement(this).set(name, String(value));
     }
     removeAttribute(name: string) {
       attributesForElement(this).delete(name);
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -4698,6 +4698,178 @@ export const tests: {[name: string]: SSRTest} = {
     };
   },
 
+  'LitElement: Reflected number attribute': () => {
+    return {
+      registerElements() {
+        class LEReflectedNumberAttribute extends LitElement {
+          @property({type: Number, reflect: true})
+          num = 42;
+        }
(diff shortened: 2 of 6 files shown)


aomarks:
@justinfagnani @augustjk this is ready for review

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -55,8 +55,10 @@ export const getWindow = ({
       old: string | null,
       value: string | null
     ): void;
-    setAttribute(name: string, value: string) {
-      attributesForElement(this).set(name, value);
+    setAttribute(name: string, value: unknown) {
+      // The browser will silently cast all values to strings. E.g. `42` becomes
+      // `"42"` and `{}` becomes `"[object Object]""`.
+      attributesForElement(this).set(name, String(value));
```
justinfagnani, line 61: btw, we found that `typeof value ==='string' ? value : String(value)` was measurably faster in lit-html.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -4698,6 +4698,178 @@ export const tests: {[name: string]: SSRTest} = {
...
   },
 
+  'LitElement: Reflected number attribute': () => {
+    return {
+      registerElements() {
+        class LEReflectedNumberAttribute extends LitElement {
+          @property({type: Number, reflect: true})
+          num = 42;
+        }
+        customElements.define(
```
justinfagnani, line 4708: why not use the decorator?

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -4698,6 +4698,178 @@ export const tests: {[name: string]: SSRTest} = {
...
+        `;
+      },
+      expectations: [
+        {
+          args: [],
+          async check(assert: Chai.Assert, dom: HTMLElement) {
+            const el = dom.querySelector(
+              'le-reflected-number-attribute'
+            )! as LitElement;
+            await el.updateComplete;
```
justinfagnani, line 4725: might be nice to eventually reduce some of the boilerplate here by treating the render() as a fixture and auto-hook-up finding an element and waiting for updateComplete, etc.

justinfagnani (approved):

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -55,8 +55,10 @@ export const getWindow = ({
       old: string | null,
       value: string | null
     ): void;
-    setAttribute(name: string, value: string) {
-      attributesForElement(this).set(name, value);
+    setAttribute(name: string, value: unknown) {
+      // The browser will silently cast all values to strings. E.g. `42` becomes
```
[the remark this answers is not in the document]
aomarks: Done

packages/labs/ssr/src/test/integration/tests/basic.ts
aomarks: Because none of the existing test cases here do

packages/labs/ssr/src/lib/dom-shim.ts
aomarks: There are a whole load of spots where we do this, not sure I want to update them all? Are we really sure it's faster?  I don't actually see this pattern in https://github.com/lit/lit/blob/main/packages/lit-html/src/lit-html.ts

PR #3048, opened 2022-06-16
[labs/react] Fix style attribute conflicts with the type definition of DOM and ReactModule

### LINKED ISSUE #3021: [labs/react] createComponent React component style property typescript problem
# Description
Overriding the style property on a wrapped component (using createComponent) results in a typescript error.
The error expects all style properties to be set.

# Steps to Reproduce

WrappedWebComponent.ts

```ts
import { Slider } from "@material/mwc-slider/slider.js";
import React from "react";

const Component = createComponent(React, "mwc-slider", Slider, {
  oninput: "input",
});

export default Component;

```

Using component in App.tsx

```ts
<WrappedWebComponent
  style={{
    background: "red",
  }}
/>
```

Typescript Error:

```
(property) style?: (CSSStyleDeclaration & React.CSSProperties) | undefined
Type '{ background: string; }' is not assignable to type 'CSSStyleDeclaration & CSSProperties'.
  Type '{ background: string; }' is missing the following properties from type 'CSSStyleDeclaration': accentColor, alignContent, alignItems, alignSelf, and 450 more.ts(2322)
  ```

## Expected Results

Allow overriding of style property

## Actual Results

Typescript error, expecting all css properties to be set

##
- Vscode
- Create React App

### PR DESCRIPTION
**Pull request by indigostar-kr**
* Style that conflicts with children, ref, and key, which are reserved words in React, is omitted from the properties of the web component.
Fixes #3021

* Added a default value for the type because if no Events were dispatched I got the following error:
```typescript
'{ children: string; "data-testid": string; appearance: any; }' 형식은 'Partial<EventProps<Events>>' 형식에 할당할 수 없습니다.
  'children' 속성이 인덱스 시그니처와 호환되지 않습니다.
    'string' 형식은 '(e: Event) => void' 형식에 할당할 수 없습니다.ts(2322)
```

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -127,7 +127,7 @@ type EventProps<R extends Events> = {
  * messages. Default value is inferred from the name of custom element class
  * registered via `customElements.define`.
  */
-export const createComponent = <I extends HTMLElement, E extends Events>(
+export const createComponent = <I extends HTMLElement, E extends Events = {}>(
   React: typeof ReactModule,
   tagName: string,
   elementClass: Constructor<I>,
@@ -143,7 +143,7 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
   // 'children', but 'children' is special to JSX, so we must at least do that.
   type UserProps = React.PropsWithChildren<
     React.PropsWithRef<
-      Partial<Omit<I, 'children'>> &
+      Partial<Omit<I, 'children' | 'style' | 'ref' | 'key'>> &
         Partial<EventProps<E>> &
         Omit<React.HTMLAttributes<HTMLElement>, keyof E>
     >
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -143,7 +143,9 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
   // 'children', but 'children' is special to JSX, so we must at least do that.
   type UserProps = React.PropsWithChildren<
     React.PropsWithRef<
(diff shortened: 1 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -143,7 +143,7 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
   // 'children', but 'children' is special to JSX, so we must at least do that.
   type UserProps = React.PropsWithChildren<
     React.PropsWithRef<
-      Partial<Omit<I, 'children'>> &
+      Partial<Omit<I, 'children' | 'style' | 'ref' | 'key'>> &
```
justinfagnani, line 146: Do we need to add `className` here too?

justinfagnani (changes requested):
Thanks @indigostar-kr !

We'll need some tests here that show we get a build error without the fix. Can you make a test that contains JSX with a style binding? We have one in the tests already with a cast to `any`, so hopefully that can be removed.

packages/labs/react/src/create-component.ts
indigostar-kr: It seems that both class and className should be included.

indigostar-kr:
> Thanks @indigostar-kr !
> 
> We'll need some tests here that show we get a build error without the fix. Can you make a test that contains JSX with a style binding? We have one in the tests already with a cast to `any`, so hopefully that can be removed.

I'll check the test code and see if I can write it.

taylor-vann:
This PR might be resolved with #3072 

:) curious if that can be confirmed

taylor-vann:
oops thought this was an issue my bad, reopened

Issue #3031, opened 2022-06-14
[labs/router] Compression mangling error


diervo:
# Description

There is a bug in the way package `@lib-labs/router` is bundled, specifically in the way files are being mangled during compression.

In production mode (the default export resolution per package.json), the bundled code at runtime is not equivalent (and incorrect in this case) to its original.

## Cause
The cause (intuitively) seems to be that the mangling during compression is done on a per-file basis:

```js
// Routes.js defines t, which represents _childRoutes) as an empty array.
class Routes {
  this.t = []; // this._childRoutes = [];
  ...
}
// Router.js redefines t as the onclick overriding incorrectly the inherited field
class Router extends Routes {
  this.t=t=>{...} // this was this._onClick = (e) => {
  ...
}
```

Since `Router extends Routes` the code is no longer equivalent after compression.

## Conclusion 

Without looking at the specific processor used to compress the files (`wireit` I believe), seems that doing this process by file will create in this (and many other cases) unintended mangling overrides due to the lack of dependencies context.

### Unwanted observation: 
IMHO I found it odd that the package is being shipped "pre-compressed" as the default, since I believe that should be the responsibility of the consumer, but that's up to the team to decide the rationale to do so :)

diervo:
FYI @justinfagnani this might be affecting other packages if the same build process is being used.

AndrewJakubowicz:
Thank you for filing an issue! Class properties were not being correctly identified in the mangling process. We also were not testing our minified outputs for labs packages. Both resolved in https://github.com/lit/lit/pull/3050

@justinfagnani can speak more to the observation about pre-compressing packages.

diervo:
Thanks for the quick turnaround! Really appreciate the good work 🙏

Oh I know what @justinfagnani will say, just wanted to poke the bear a little 😹

justinfagnani:
Thanks for the report @diervo!

Yeah... I think you know already :) but for posterity I'll get into it a little:

We've been applying minification to our core libraries, even though I agree with you that minification should normally be an app concern, because we could get smaller output by using potentially unsafe settings that we know are safe and tested in our codebase, and in order to make using private API more cumbersome.

We probably don't have to apply this to all of our packages, like router. Andrew's fix is just the easiest, imcremental thing to do without re-litigating our strategy here. I'm open to changing though. Ultimately the right answer is private fields, which can be freely and safely renamed so we won't even have to do it - Terser could by default as used in apps.

PR #3050, opened 2022-06-16
[infra] Test labs production builds and fix class property name mangling collisions

AndrewJakubowicz:
Fixes issue https://github.com/lit/lit/issues/3031

### Context

Our labs packages have `test:prod` commands, however they do not have a remapping to the production bundle. This results in the `test:prod` command being a duplicate of the dev mode test.

After fixing the `test:prod` commands, it reproduced the bug in the linked issue, as well as a property mangling failure in labs/React.
An investigation found that the name cache system was tree shaking out the imports, which then meant terser didn't have any context for what property names were in the bundle. Without the name cache terser would rename different subclass properties to the same name.

### Fix

By adding `console.log` statements for each import in the name cache file, it prevents tree shaking and dead code elimination of the classes. This correctly populates the terser name cache preventing naming collisions.

Was verified as the initial commits of this PR are red - with two naming collisions detected in labs/Router and labs/React. After adding the console.log, the tests pass, and the properties are correctly renamed.

### Tests

This PR adds additional testing since it enables correct production testing in our labs packages.

### Risks

This change does modify some of the minified properties - which creates some risk of name collisions.
Spot checked minified changes in go/diff#key=G7taLmvP2t45
 - lit-element and reactive-element are using their module class prefixes.
 - No difference in lit-html, minified output.

Tested checksize outputs. They were pretty much the same between main and this change: go/diff#key=vG22tcDUQTOH

### DIFF
--- a/.changeset/eleven-zebras-crash.md
+++ b/.changeset/eleven-zebras-crash.md
@@ -1,0 +1,6 @@
+---
+'@lit-labs/router': patch
+'@lit-labs/react': patch
+---
+
+Fix a property collision in the minified production build.
--- a/packages/labs/router/rollup.config.js
+++ b/packages/labs/router/rollup.config.js
@@ -9,6 +9,6 @@ import {createRequire} from 'module';
 
 export default litProdConfig({
   packageName: createRequire(import.meta.url)('./package.json').name,
-  entryPoints: ['index', 'router'],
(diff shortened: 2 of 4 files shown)


augustjk (approved):
Great find, fix, and explanation!

justinfagnani (approved):
Awesome investigative work!

PR #2978, opened 2022-06-01
Remove css-tag stylesheet caching by css text

justinfagnani:
We didn't intend to share StyleSheets between components, but we were by having a cache keyed by style text. This removes that cache and stores the StyleSheet with the CSSResult.

PR #3060, opened 2022-06-17
[reactive-element] Adds `preserveExisting` option to `adoptStyles`

### LINKED ISSUE #3010: [reactive-element] Normalize `adoptStyles` so that it works x-browser when called to modify instance styles
ReactiveElement supports static styles that are not modified per instance. This was done so that support for Web Components polyfilled styling via ShadyCSS (which does not support instance styling) would work. The `adoptStyles` function was designed for this use case.

However, now that a significant portion of users do not need Web Components polyfills, it's reasonable to start introducing better instance based styling. Currently `adoptStyles` works inconsistently when the underlying `adoptedStylesheets` is supported or not. This behavior should be normalized so that it works the same x-browser when native Shadow DOM is used.

Separately, [users have asked](https://github.com/lit/lit/issues/2984) for a way to add styles rather than re-apply them. This could be done as a separate method or an alternative signature.

sorvell:
Fixes #3010. Adds ability to add to an existing set of styles via `adoptStyles`.

Size delta:
main: 12.75 kB │ 6.49 kB │ 5.57 kB
preserve: 13.29 kB │ 6.73 kB │ 5.8 kB


packages/reactive-element/src/css-tag.ts
```diff
@@ -54,17 +53,15 @@ export class CSSResult {
...
-    // Note, if `supportsAdoptingStyleSheets` is true then we assume
-    // CSSStyleSheet is constructable.
-    let styleSheet = styleSheetCache.get(this.cssText);
-    if (supportsAdoptingStyleSheets && styleSheet === undefined) {
-      styleSheetCache.set(this.cssText, (styleSheet = new CSSStyleSheet()));
-      styleSheet.replaceSync(this.cssText);
+    // If `supportsAdoptingStyleSheets` is true then we assume CSSStyleSheet is
+    // constructable.
+    if (supportsAdoptingStyleSheets && this._styleSheet === undefined) {
+      (this._styleSheet = new CSSStyleSheet()).replaceSync(this.cssText);
```
justinfagnani, line 62: This is going to cause stylesheet duplication and css text re-parsing in some maybe still relatively common scenarios:

```ts
class BaseElement extends LitElement {
  static get styles() { return css`...`; }
}

class SubElementA extends BaseElement {
  static get styles() { return [BaseElement.styles, css`...`]; }
}
```

justinfagnani: It will however give a speedup on codebases that don't use a getter for `styles`, like all TS codebases, due to skipping a hash of a potentially long string.

I wonder if we can split the difference by caching based on the TemplateStringsArray, but only when the length is `1`, so there are no interpolations to worry about.

sorvell: This PR was based on https://github.com/lit/lit/pull/2978 and that change is made there, so let's discuss this change there?

justinfagnani: Oh yeah... I paused that one to consider what to do about getters.

sorvell:
We've decided not to land this. See https://github.com/lit/lit/issues/3010#issuecomment-1264113426.

PR #3061, opened 2022-06-17
[reactive-element] Adds support for adopting styling elements and adds `getAdoptedStyles`

sorvell:
As a follow-on to #3060 and also to help address #3010:
* `adoptStyles` now allows adopting style and link elements directly: this change facilitates more easily using external stylesheets as well as authoring styles via css, which some users prefer. This has been possible by embedding styles manually, but adding support via `adoptStyles` means that ReactiveElement's `static styles` property can leverage elements.
* `getAdoptedStyles` added. Returns an array of adopted stylesheets and elements. This is useful to capture the adopted set of styles before modifying them.


sorvell:
We've elected not to land this, since it does not make sense without https://github.com/lit/lit/pull/3060 (which was closed based on https://github.com/lit/lit/issues/3010#issuecomment-1264113426),

PR #3062, opened 2022-06-17
[lit-html] Mark TemplateResult brand as internal

AndrewJakubowicz:
Very small documentation change that fixes https://github.com/lit/lit.dev/issues/865

### DIFF
--- a/.changeset/calm-olives-teach.md
+++ b/.changeset/calm-olives-teach.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Mark TemplateResult brand as internal cleaning up generated documentation.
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz:
Ah! Solution does not work.

End up with an error:

```
../lit-html/lit-html.d.ts:349:29 - error TS2339: Property '_$litType$' does not exist on type 'TemplateResult<ResultType>'.
349     constructor({ strings, ['_$litType$']: type }: TemplateResult, options?: RenderOptions);
                            ~~~~~~~~~~~~
```

AndrewJakubowicz:
Not sure this is the right approach. Will re-open once I have a solution.

PR #3072, opened 2022-06-21
[labs/react] flatter faster types for UserProps

taylor-vann:
Currently Typescript struggles to react to changes in a promptly manner.

This is probably due to a couple of nested structures bogging down the type checker.

Replacing the following:

```ts
  type UserProps = React.PropsWithChildren<
    React.PropsWithRef<
      Partial<Omit<I, 'children'>> &
        Partial<EventProps<E>> &
        Omit<React.HTMLAttributes<HTMLElement>, keyof E>
    >
```

With the following set of types
```ts
  type ReactProps = React.PropsWithChildren<React.HTMLAttributes<I>>;
  type ElementWithoutChildrenOrEvents = Omit<I, keyof E | keyof ReactProps>;
  type UserProps = Partial<
    ElementWithoutChildrenOrEvents & ReactProps & EventProps<E>
  >;
```

Helps typescript inference types much faster.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -141,12 +141,11 @@ export const createComponent = <I extends HTMLElement, E extends Events>(
   // ref, as well as special event and element properties.
   // TODO: we might need to omit more properties from HTMLElement than just
   // 'children', but 'children' is special to JSX, so we must at least do that.
-  type UserProps = React.PropsWithChildren<
-    React.PropsWithRef<
-      Partial<Omit<I, 'children'>> &
-        Partial<EventProps<E>> &
-        Omit<React.HTMLAttributes<HTMLElement>, keyof E>
-    >
+
+  type ReactProps = React.PropsWithChildren<React.HTMLAttributes<I>>;
+  type ElementWithoutChildrenOrEvents = Omit<I, keyof E | keyof ReactProps>;
+  type UserProps = Partial<
+    ElementWithoutChildrenOrEvents & ReactProps & EventProps<E>
   >;
 
   // Props used by this component wrapper. This is the UserProps and the


taylor-vann:
This might help resolve: https://github.com/lit/lit/pull/3048

taylor-vann:
This is a better to union and intersect React Props, Events, and Element properties.

a) get react props without event keys
b) get element without react props or event keys
c) combine (a) & (b) & Events

sorvell (dismissed):
Need a better explanation of the change here. For example, can you explicate the various statements in the old v. new code, explain why they are equivalent, and then explain why the change is better. Thanks.

taylor-vann:
No problem @sorvell :)

These changes improve our typing in a couple of ways.

before:
```ts
  type UserProps = React.PropsWithChildren<
    React.PropsWithRef<
      Partial<Omit<I, 'children'>> &
        Partial<EventProps<E>> &
        Omit<React.HTMLAttributes<HTMLElement>, keyof E>
    >
```

after:
```ts
  type ReactProps = Omit<React.HTMLAttributes<I>, keyof E>;
  type ElementWithoutPropsOrEvents = Omit<I, keyof E | keyof ReactProps>;
  type UserProps = Partial<
    ElementWithoutChildrenOrEvents & ReactProps & EventProps<E>
  >;
```

a) Removes the need to explicitly maintain react specific properties like 'style' or 'children' or 'className'. Mainly because react provides these types out of the box through `React.HTMLAttributes`.

```ts
    interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
        ...
        className?: string | undefined;
        ...
        style?: CSSProperties | undefined;
        ...
```

b) Likewise, these changes remove redundant or duplicate `key:value` statements. For instance, `'children'` is provided by both `React.PropsWithChildren` and `React.HTMLAttributes`. So the following two types are the same:

```ts
type DemoComponent = React.HTMLAttributes<I>;
type DemoComponent2 = React.PropsWithChildren<React.HTMLAttributes<I>>;
```

c) Unions on multiple partials is the same as partial-ing a union. It also reads much better. As in the following two types are the same:

```ts
type UnionPartial = Partial<I> & Partial<E>;
type UnionPartial = Partial<I & E>;
```

d) Types are built around `Event` properties. Currently we use nested React types and a sprinkle of ommited properties to get the types we need. However I think taking a more categorical approach not only improves vscode type hint speed but also makes what we're exposing to clients through `props` much more clear.

Currently the logic reads in one big chunk:
```ts
inside ReactPropsWithChildren
  inside react props with ref ->
    Union HTMLElement without children prop
      with EventProps
      with union React.HTMLElements without EventProps keys


  type UserProps = React.PropsWithChildren<
    React.PropsWithRef<
      Partial<Omit<I, 'children'>> &
        Partial<EventProps<E>> &
        Omit<React.HTMLAttributes<HTMLElement>, keyof E>
    >
```

These changes create types more correctly using intersections and unions. It also incrementally explains what types are needed rather than nesting:

```ts
ReactProps from HTMLElement without EventProps keys (missing from above)
Element attributes without EventProps keys or ReactProps keys
Union (ReactProps without EventProp keys) and (Element without ReactProp or EventProps keys) with EventProps.

  type ReactProps = Omit<React.HTMLAttributes<I>, keyof E>;
  type ElementWithoutPropsOrEvents = Omit<I, keyof E | keyof ReactProps>;
  type UserProps = Partial<
    ElementWithoutChildrenOrEvents & ReactProps & EventProps<E>
  >;
```

e) there is a notice-able change in linting speed in vscode from these changes, however the actual build-speed changes are negligible.

.changeset/calm-bottles-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/react': major
```
justinfagnani, line 2: This is not a major change, yeah?

justinfagnani (approved):

taylor-vann:
Gonna confirm ref before merge :)

.changeset/calm-bottles-laugh.md
taylor-vann: no. This is not a breaking change but it could affect typings downstream.

I wanna say this is a patch since it doesn't break anything.

PR #3077, opened 2022-06-23
[labs/router] Fix README.md code sample

AndrewJakubowicz:
Tiny documentation change. `outlet` must be called and should be accessed via `this`.


augustjk (approved):
Good catch!
There's an argument that could be made for doing a changeset version bump since it could change what shows up here https://www.npmjs.com/package/@lit-labs/router but it's such a tiny change that it probably doesn't need to happen.

augustjk (approved):

augustjk:
> TIL! Will this get updated on `npm` with the other Router changes that are pending in release?

Ah yes, since there already is a pending release for the `@lit-labs/router` package as you've pointed out, this README update would indeed be reflected in the next release. So it would just be a matter of do we need an extra changelog line that mentions this update, I'd say no.

justinfagnani (approved):
good catch!

PR #3078, opened 2022-06-23
[infra] Add new bug, feature request, and alternatives issue configs

aomarks:
Using [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) to try and guide users towards filing more useful issues, and to redirect them to better locations in some cases.

Not sure how to make a preview of this, but I've got the same exact config at https://github.com/aomarks/issue-template-test/issues/new/choose where you can try it out.


AndrewJakubowicz (commented):
I really like this. Tested making an issue through the example flow https://github.com/aomarks/issue-template-test/issues/1 and I like the information that the form encourages. And it's not too prescriptive because you can easily type anything to bypass a required input if unsure of what to place.

My only question is if the TypeScript/JavaScript example in the playground should already have a bug? I spent a moment wondering where the bug was and what was safe to delete.

aomarks:
> My only question is if the TypeScript/JavaScript example in the playground should already have a bug? I spent a moment wondering where the bug was and what was safe to delete.

Yeah, they have a bug. The button doesn't work because `this.numClicks` should be `this._numClicks`, and there's a comment "// A typo for bug demonstration purposes. The leading _ is missing.". Open to better bug examples, though. It's sort of weird because it's obviously a user error, but we can't really create an example of an actual bug, since we'd fix it and then the repro wouldn't work.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+body:
+  - type: dropdown
+    id: affected
+    validations:
+      required: true
+    attributes:
+      label: Which package(s) are affected?
+      multiple: true
+      description: Note you can select more than one.
```
augustjk, line 12: Maybe it's obvious but should we say something like "If other, mention in description below"? Eh, maybe it'll just be extra noise.

augustjk (approved):
Nice! It would be nice if the package selection could also dynamically add some headings to the issue title but that's just wishful thinking.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - Context (@lit-labs/context)
+        - Eleventy Plugin (@lit-labs/eleventy-plugin-lit)
+        - Localize (@lit/localize)
+        - Motion (@lit-labs/motion)
+        - Observers (@lit-labs/observers)
+        - React (@lit-labs/react)
+        - SSR (@lit-labs/ssr)
+        - Starter Kits (@lit/lit-starter-(js|ts))
+        - Task (@lit-labs/task)
+        - Other/unknown
```
graynorton, line 24: Should add Virtualizer (@lit-labs/virtualizer)

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - type: textarea
+    id: reproduction
+    validations:
+      required: true
+    attributes:
+      label: Reproduction
+      description: |
+        Please link to a working reproduction that will help us verify and understand the bug.
+
+        For bugs that show up in the browser, please provide a link to the lit.dev playground: [TypeScript example](https://lit.dev/playground/#gist=b90fc1b4d2ee57120763d2e7db934511), [JavaScript example](https://lit.dev/playground/#gist=2d20e91832d3f98332bf23ced55bea04). Click the "Share" button to generate a URL.
```
graynorton, line 60: I wonder whether we want to say anything along the lines of "If you are unable to create a reproduction in the lit.dev playground but can provide a link to a working reproduction hosted elsewhere, please do so."

On the one hand, I really want to encourage people to use the playground, but on the other, any working repro is much better than none.

graynorton (changes requested):
This looks awesome, @aomarks! Just made one request and one potential suggestion.

.github/ISSUE_TEMPLATE/bug.yaml
aomarks: Done

.github/ISSUE_TEMPLATE/bug.yaml
aomarks: Done

.github/ISSUE_TEMPLATE/bug.yaml
aomarks: Done

AndrewJakubowicz (approved):

graynorton (approved):
LGTM—thanks for doing this!

AndrewJakubowicz (approved):

Issue #3091, opened 2022-06-25
[infra] Something broken for testing in recent @web/dev-server


aomarks:
When upgrading from `@web/dev-server` `0.1.30` to `0.1.31`, something breaks in our tests. The tests never start. The HTTP requests for JS modules served by the test server timeout, but there are no error messages.

The main change from this release are (https://github.com/modernweb-dev/web/blob/master/packages/dev-server/CHANGELOG.md#0131):

```
- e10b680d: Support node entry points (export map) containing stars.
- Updated dependencies [e10b680d]
  -@web/dev-server-rollup@0.3.16
```

So the issue may be related to export maps.

PR #3092, opened 2022-06-25
[infra] Upgrade package locks, and pin some broken deps

aomarks:
This upgrades all package locks. `npm run upgrade` should be stable as of this PR.

There were a few issues resulting from new dependencies that needed these fixes:

- Pinned TypeScript versions in a couple of packages. We should be using `~` in general for TypeScript anyway, because TypeScript doesn't follow semver.

- Pinned `ts-clone-node` version because of an unknown breaking change.

- Changed import style for `rollup-plugin-summary` due to a presumed change in its export style.

- Pinned `@web/dev-server` and other `@web/test-*` libraries in the tests package, because something has broken in a recent version. Filed https://github.com/lit/lit/issues/3091 to track.


packages/tests/package.json
```diff
@@ -12,9 +12,10 @@
     "@types/prettier": "^2.0.1",
     "@types/diff": "^5.0.0",
     "@types/fs-extra": "^9.0.1",
+    "@web/dev-server": "=0.1.30",
     "@web/dev-server-legacy": "^0.1.7",
-    "@web/dev-server-rollup": "^0.3.13",
-    "@web/test-runner": "^0.13.22",
+    "@web/dev-server-rollup": "=0.3.15",
```
AndrewJakubowicz, line 17: I wasn't able to find docs around this syntax. Is this a syntax to pin that exact version? Or a version of "save-exact"?

aomarks: It means exactly equal: https://github.com/npm/node-semver#ranges

PR #3096, opened 2022-06-26
Support space-separated classInfo properties in classMap

### LINKED ISSUE #3095: Support for whitespace-separated CSS classnames in classMap directive.

### Description

Sometimes an element's `classList` needs to update reactively in such a way that multiple class names are added or removed based on the same boolean condition. This might apply to other CSS methodologies, but in my recent experience, this has come up when using atomic classes (from TailwindCSS).

In other words, when using the `classMap` directive, it would be desirable to be able to do the following:

```javascript
return html`
  <div
    class=${classMap({
      "foo bar": this.booleanProperty,
      "baz qux": !this.booleanProperty
    })}
  ></div>
`;
```

This currently produces an error, as `classMap` assumes that the keys in the `classInfo` argument to be valid DOM tokens, rather than potentially being space-separated strings. I have created a minimal example [on CodePen](https://codepen.io/nicholas-wilcox/pen/VwXZbpE) to demonstrate the use case.

**Note:** This pattern does not produce an error until classes are added/removed _after_ the first render. If this feature should not be added, then it might be best to explicitly discourage the pattern in the documentation, or in a console warning.

Developers can expand the `classInfo` object so that every class name in is a separate object property, which would duplicate the code used for their shared boolean value.

Alternatively, there is the `classnames` package [on NPM](https://www.npmjs.com/package/classnames), which supports this pattern, while (to my knowledge) completely avoiding the `DOMTokenList` API.

nicholas-wilcox:
Fixes #3095

The main theme of this PR is to insert a small layer of abstraction between `name` and `classInfo`, which is accomplished by iterating over each `key` in the `classInfo` object, and then iterating over each `name` obtained from:
```javascript
key.split(/\s/).filter((s) => s !== '')
```

### DIFF
--- a/packages/lit-html/src/directives/class-map.ts
+++ b/packages/lit-html/src/directives/class-map.ts
@@ -65,41 +65,53 @@ class ClassMapDirective extends Directive {
             .filter((s) => s !== '')
         );
       }
-      for (const name in classInfo) {
-        if (classInfo[name] && !this._staticClasses?.has(name)) {
-          this._previousClasses.add(name);
+      for (const key in classInfo) {
+        // Account for object keys containing multiple class names separated by whitespace.
+        const names = key.split(/\s/).filter((s) => s !== '');
+        for (const name of names) {
+          if (classInfo[key] && !this._staticClasses?.has(name)) {
+            this._previousClasses.add(name);
+          }
         }
       }
       return this.render(classInfo);
     }
 
     const classList = part.element.classList;
+    const classInfoKeys = Object.keys(classInfo);
 
     // Remove old classes that no longer apply
     // We use forEach() instead of for-of so that we don't require down-level
     // iteration.
     this._previousClasses.forEach((name) => {
-      if (!(name in classInfo)) {
+      // Check for strict equality or if a object key contains the name as a substring.
+      if (
+        !(name in classInfo) &&
+        !classInfoKeys.some((key) => key.includes(name))
+      ) {
         classList.remove(name);
         this._previousClasses!.delete(name);
       }
     });
 
     // Add or remove classes based on their classMap value
-    for (const name in classInfo) {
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/directives/class-map.ts
```diff
@@ -65,41 +65,53 @@ class ClassMapDirective extends Directive {
...
-          this._previousClasses.delete(name);
+      const value = !!classInfo[key];
+      const names = key.split(/\s/).filter((s) => s !== '');
+      for (const name of names) {
+        if (
+          value !== this._previousClasses.has(name) &&
+          !this._staticClasses?.has(name)
+        ) {
+          if (value) {
+            classList.add(name);
```
nicholas-wilcox, line 109: `DOMTokenList.add()` (on `part.element.classList`) accepts rest parameters, but `Set.add()` (on `this._previousClasses`) does not, so I didn't bother to rewrite to reduce the number of times `classList.add()` is called.

packages/lit-html/src/directives/class-map.ts
```diff
@@ -65,41 +65,53 @@ class ClassMapDirective extends Directive {
             .filter((s) => s !== '')
         );
       }
-      for (const name in classInfo) {
-        if (classInfo[name] && !this._staticClasses?.has(name)) {
-          this._previousClasses.add(name);
+      for (const key in classInfo) {
+        // Account for object keys containing multiple class names separated by whitespace.
+        const names = key.split(/\s/).filter((s) => s !== '');
```
nicholas-wilcox, line 70: I used the same `split` and `filter` arguments that are used above when initializing `this._previousClasses`. Perhaps this code should be deduplicated.

I also considered using a greedy quantifier `/\s+/` in the RegExp, but I opted to use the code that was already here.

Issue #3093, opened 2022-06-26
[infra] Convert Lit monorepo to npm workspaces


aomarks:
Replace Lerna with npm workspaces.

PR #3098, opened 2022-06-28
[infra] Replace lerna run with top-level wireit scripts

aomarks:
Replaces all `lerna run` scripts with `wireit` scripts.

### Benefits

- Increased parallelism during build/test.

- We'll now only be using Lerna for bootstrapping, so we can replace it with npm workspaces in a followup PR.

- It's now easier to figure out which packages failed, because the log statements will say e.g. `[packages/lit-html:build] Failed ...` instead of `[build] Failed`.

### Enumerated packages

This does include some long enumerations of packages. Once we have https://github.com/google/wireit/issues/23, we could replace this with something like:

```json
  "wireit": {
    "test": {
      "dependencies": [
        {
          "script": "test",
          "packages": "workspaces"
        }
      ]
    }
  }
```

However, we currently exclude some packages, e.g. we don't run `lit-html`/`lit-element` package tests, because they are subsumed by the `tests` package. Previously we excluded them with Lerna `--ignore` flags. Now they are just not in the list.

Maybe once we have the above Wireit feature, we could add `test:ci` scripts, which would not be present in the case of packages whose tests we don't want to run in CI.

### Also

- Fixes an incorrect virtualizer output path.

- Fixes an output issue where the lit-html `build:version-stability-test` config was emitting `polyfill-support.js`, which collided with the same file emitted by `build:rollup`.

- Bumps the wtr tests finished timeout. Running both the SSR and main browser tests in parallel makes tests take longer. I guess because of resource starvation.

- Added some additional dependencies to the new `testing` package.

Fixes https://github.com/lit/lit/issues/2724
Part of https://github.com/lit/lit/issues/3093


packages/tests/src/web-test-runner.config.ts
```diff
@@ -226,7 +226,7 @@ const config: TestRunnerConfig = {
   // enough so that blocked tests have time to wait for all previous test files
   // to run to completion.
   testsStartTimeout: 60000 * 10, // default 120000
-  testsFinishTimeout: 120000, // default 20000
+  testsFinishTimeout: 180000, // default 20000
```
justinfagnani, line 229: why would this change require a longer timeout?

aomarks:
Also added these items:

- Fixes an output issue where the lit-html `build:version-stability-test` config was emitting `polyfill-support.js`, which collided with the same file emitted by `build:rollup`.

- Temporarily `skip` a flaky virtualizer test (https://github.com/lit/lit/issues/3099).

packages/tests/src/web-test-runner.config.ts
aomarks: See PR description. I think it's because the two are now running in parallel, so there must be some resource starvation that causes the test to take longer. It seemed to particularly affect Firefox.

aomarks: I bumped it even higher, to make it pass reliably. I think this is just a contention thing. But the good news is that overall test time seems lower: latest full run of the test step was 7m28s, compared to 9m58s on the most recent commit on main.

augustjk (approved):
Thanks for fixing the labs/testing wireit configs!

justinfagnani (approved):

Issue #3099, opened 2022-06-28
[labs/virtualizer] flaky indices test


aomarks:
This test is frequently failing in CI:

```
 ❌ smoke test > virtualize > visible indices > emits visibilityChanged events with the proper indices
AssertionError: expected 0 to equal 1
+ expected- actual
-0
+1
at o.<anonymous> (src/test/scenarios/smoke.test.ts:125:31)
```

usergenic:
Oh!  Will check it out right away.

PR #3100, opened 2022-06-28
[labs/virtualizer] Removed flaky wait() test helper and replaced with until(fn)

usergenic:
Fixes flaky test reported in issue #3099

I was using a function I thought would reliably wait for browser events to cascade to the point I could check state, but this is apparently not reliable in all browsers and at all times.  I have switched this out for a new function which returns a promise that resolves when a condition returns true and rejects after a timeout of 1 second (default).

This seems to be a more reliable and explicit means to achieve the same result.

cc @aomarks


aomarks (approved):
TY for quick fix!

PR #3109, opened 2022-06-30
[labs/react] more generic interface with react module

taylor-vann:
This PR adds a more generic type with a smaller scope in lieu of `typeof React`.

The entire react "module" is passed as the first argument to `createComponent`.

However we don't need to use the entire API of the react module. It is more portable to only require what we use:
```
Component
createComponent
forwardRef
```

All other instances of react are for types, so the three properties above would have to fit into the expected types from `react` anyways


taylor-vann:
this isn't as beneficial as I thought

taylor-vann:
actually might be beneficial to support preact / react

augustjk:
I tried another take of this here https://github.com/lit/lit/pull/4143 too but ultimately don't feel this is worth pursuing at the moment. It's not going to change anything in terms of types for preact users using `preact/compat`, and it would make changing any React API usage be breaking.

PR #2125, opened 2021-09-01
Class map benchmarks

AndrewJakubowicz:
### Context

Reproduce benchmark raised in issue: https://github.com/lit/lit/issues/2010

[...]

PR #3110, opened 2022-06-30
Initial ClassMap benchmarks

AndrewJakubowicz:
Repurpose most of the work done in https://github.com/lit/lit/pull/2125 to track our classMap directive performance.

### DIFF
--- a/packages/benchmarks/lit-element/class-map/index.ts
+++ b/packages/benchmarks/lit-element/class-map/index.ts
@@ -146,27 +146,6 @@ function generateData(number: number): IData[] {
     };
     await nopupdate();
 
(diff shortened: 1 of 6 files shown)


packages/benchmarks/lit-element/class-map/index.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
```
AndrewJakubowicz, line 3: Note: This license date is from when this code was initially written. I did modify and simplify it, and maybe it makes sense to set this date to the commit date (2022).

gdcmarinho (approved):

PR #3111, opened 2022-07-01
[labs/react] remove StringValued type

taylor-vann:
This PR removes the `StringValued` type.

It can be replaced by the `Events` type.

### DIFF
--- a/.changeset/thin-bees-build.md
+++ b/.changeset/thin-bees-build.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
+Reduced number of types in createComponent
--- a/.changeset/thin-bees-build.md
+++ b/.changeset/thin-bees-build.md
@@ -2,4 +2,4 @@
 '@lit-labs/react': patch
 ---
 
-Reduced number of types in createComponent
+Removed the unexposed and unnecessary `StringValued` type used to correlate property names with event listener names.
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -55,14 +55,14 @@ const addOrUpdateEventListener = (
  * Sets properties and events on custom elements. These properties and events
  * have been pre-filtered so we know they should apply to the custom element.
  */
-const setProperty = <E extends Element, T>(
+const setProperty = <E extends Element>(
   node: E,
   name: string,
   value: unknown,
   old: unknown,
-  events?: StringValued<T>
+  events?: Events
 ) => {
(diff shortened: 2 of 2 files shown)


.changeset/thin-bees-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
+Reduced number of types in createComponent
```
augustjk, line 5: I think I would like this changelog to be a bit more specific. When you say reduced number of types in `createComponent`, is it the file or the function? I know that it's the former by looking at the PR changes which means it doesn't actually affect end users at all, but just by looking at the changelog I wouldn't know that.

Just a thought in general regarding changelogs though, do we want to somehow mark changelogs as typescript only if changes do not affect runtime code? Before I started using TS, I appreciated those in changelogs of other packages since it let me just ignore them.

taylor-vann: Yah this one is interesting because the change makes no difference to the user, only the dev

taylor-vann: Personally I've never found changelings to help or improve my understanding of a library

taylor-vann: If changes need to be logged, I'd propose we only mark changes under two conditions:
1) a PR modifies the promised API of a package
2) there is no issue associated with a PR

The fact that we changeset everything for a mono repo escapes me

augustjk: If it's a code change we want to publish to npm, then there needs to be a changeset describing it along with the version bump.

I think having a clean readable changelog is valuable. It's something we publish along with our release tweets, and is a guide for users looking to update.

taylor-vann: For sure, I'll update the changelog when I get home! But I'm still concerned we put way too much of a value judgement on changesets and other processes than actual code.

taylor-vann: If anything, I'd use a README to understand a library, not changesets

taylor-vann: Changelog also have the trouble that their relevance is inversely proportional to their age. As in changesets from previous major and minor versions become increasingly irrelevant and unhelpful the farther they are from `current`.

My other two cents is we should be publishing something a little more interesting than a changelog. I'd rather see screenshots, demos, or something more visually engaging to tell users what we've improved. A screenshot of a changelog is soooo boring

taylor-vann: done

justinfagnani (approved):

PR #3112, opened 2022-07-01
[infra] Use npm workspaces instead of lerna

aomarks:
This PR removes Lerna entirely. Now we just use npm workspaces + wireit.

### Benefits

- No more `npm run bootstrap`, just `npm ci`.
- To upgrade or add a dependency, edit the `package.json` and `npm i` from the root.
- Cuts around 30s off install time in CI.
- Only one `package-lock.json` file for the whole monorepo.

### Tricky parts

- npm workspaces hoists packages to the top-level `node_modules/` folder whenever possible. This is different to Lerna, which always put all the dependencies in each package's own `node_modules/` folder. Only if there is a conflicting version would there now be a `node_modules/` folder inside a workspace package.

  - This required updating various paths to reach further up the file system.
 
  - In a couple of cases, this caused issues with conflicting `@types` packages for TypeScript, because while previously only the local version was available, now multiple versions could be. This is because TypeScript by default ambiently loads all `node_modules/@types/*` packages from _all recursive parent directories_. The solution to these cases are to specify that only the immediate `node_modules/` folder should be consulted through the `typeRoots` setting.

  - This required changing the tachometer server root directory as well, which ended up being more complicated than it should be, because of the way tachometer was treating paths in config files as relative to the current working directory, instead of the config file path. In https://github.com/Polymer/tachometer/pull/237 I changed that behavior, so this also bumps the tachometer version and updates the configs accordingly.

Fixes https://github.com/lit/lit/issues/3093


package.json
```diff
@@ -17,15 +22,15 @@
...
+    "nuke": "rm -rf node_modules && npm exec -ws -c 'rm -rf node_modules' && npm ci",
+    "upgrade": "rm -rf node_modules package-lock.json && npm exec -ws -c 'rm -rf node_modules' && npm i",
     "test": "wireit",
     "test:lit-starter-js": "wireit",
     "test:lit-starter-ts": "wireit",
     "test:windows": "wireit",
     "prepare": "husky install",
     "changeset": "changeset",
-    "version": "npm run changeset version && npm run update-version-vars && lerna bootstrap -- --package-lock-only",
+    "version": "npm run changeset version && npm run update-version-vars && npm i",
```
augustjk, line 33: I added the the `lerna bootstrap` step in an effort to generate updated versions of package-lock files for the updated packages (which ended up not working). Since we no longer have sub package package-locks, this step is no longer necessary.
```suggestion
    "version": "npm run changeset version && npm run update-version-vars",
```

augustjk (commented):
Haven't looked at everything yet but wanted to leave this comment first.

.github/workflows/tests.yml
```diff
@@ -17,9 +17,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 16
-          # Note: we do not cache here because we're not running lerna
-          # bootstrap, so the ~/.npm dir won't have all dependencies, and if
-          # this job finishes first then we'll only cache root dependencies.
+          cache-dependency-path: '**/package-lock.json'
```
justinfagnani, line 20: Add a comment on why this isn't just the top-level package-lock?

package.json
```diff
@@ -17,15 +22,15 @@
     "format:prettier": "prettier \"**/*.{cjs,html,js,json,md,ts}\" --write",
     "ignore-sync": "ignore-sync .",
     "lint": "eslint \"**/*.{js,ts}\"",
-    "nuke": "rm -rf node_modules && npm install && lerna exec 'rm -rf node_modules' && lerna bootstrap && npm run clean",
-    "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap && npm run clean",
+    "nuke": "rm -rf node_modules && npm exec -ws -c 'rm -rf node_modules' && npm ci",
```
justinfagnani, line 25: This is difference from the previous nuke, which ran a regular `npm i` via `lerna bootstrap`. I also think it should be basically unnecessary as `npm ci` should put all node_modules in the state described by the package-lock. The only case that's different is if you modified files in node_modules/

package.json
```diff
@@ -17,15 +22,15 @@
...
+    "upgrade": "rm -rf node_modules package-lock.json && npm exec -ws -c 'rm -rf node_modules' && npm i",
```
justinfagnani, line 26: `npm upgrade` should work here now.

packages/labs/gen-wrapper-react/src/test-gen/generate_test.ts
```diff
@@ -50,6 +50,15 @@ test('basic wrapper generation', async () => {
...
   });
 
+  // The version of @types/react might conflict with the one installed to the
+  // top-level of our monorepo. By default, TypeScript will load all
+  // node_modules/@types/* packages for all parent directories. By setting
+  // typeRoots here, we ensure it only loads the immediate ones.
+  const tsConfigPath = path.join(outputPackage, 'tsconfig.json');
+  const tsConfig = JSON.parse(fs.readFileSync(tsConfigPath, 'utf8'));
+  tsConfig.compilerOptions.typeRoots = ['./node_modules/@types'];
+  fs.writeFileSync(tsConfigPath, JSON.stringify(tsConfig), 'utf8');
```
justinfagnani, line 60: 👍

packages/tests/src/web-test-runner.config.ts
```diff
@@ -169,7 +169,9 @@ const require = createRequire(import.meta.url);
 
 // https://modern-web.dev/docs/test-runner/cli-and-configuration/
 const config: TestRunnerConfig = {
-  rootDir: '../',
+  // Serve from the root of the monorepo, because most dependencies are hoisted
+  // into its node_modules folder.
+  rootDir: '../../',
```
justinfagnani, line 174: I thought WTR could serve files outside of the root with a special endpoint?

justinfagnani (approved):
🎉

package.json
aomarks: Deleted the script

package.json
aomarks: Deleted the script

packages/tests/src/web-test-runner.config.ts
aomarks: It does, but for our HTML tests we use hard-coded paths to node_modules. Added a comment.

.github/workflows/tests.yml
aomarks: Actually it can just use the top-level one now. Updated.

augustjk (approved):
There are some lingering references to `npm run bootstrap` in the tachometer json files under dependency's `setupCommands` but otherwise LGTM!

aomarks:
> There are some lingering references to `npm run bootstrap` in the tachometer json files under dependency's `setupCommands` but otherwise LGTM!

Yep, to get benchmarks to work I actually restored the `bootstrap` script and just made a no-op. The reason is that, for a while, for benchmarks, we'll need to keep running `npm run bootstrap` in the tachometer setup steps. That's because tachometer checks the current PR against the most recent tagged release, and so until we next release the core libraries, those commits will still require the old bootstrap command.

PR #3114, opened 2022-07-01
[infra] Configure eslint import/no-extraneous-dependencies rule

aomarks:
Configures a new eslint rule which enforces that, when we import a module, that module is referenced in the "dependencies" of the immediate "package.json" file.

Docs: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

This is especially important now that we're using npm workspaces, because npm workspaces hoists as many dependencies as possible to the root `node_modules/` folder, which can make it easy to forget to include a dependency.

Exceptions are made for internal files like tests, which are free to import dependencies that are only declared in the top-level "package.json".

### Violations

- `@lit-labs/cli` was missing a dependency on `chalk`. Added the missing dependency.
- `@lit-labs/cli` depended on `@lit-labs/analyzer` as `devDependency`. Switched to regular `dependency`.
- `@lit-labs/react` imports `react`, which is in `devDependencies`. However, this is intentional, because it's a type-only import. Added a special-case to allow it to be imported from `devDependencies`.

### Also

- Found a few more spots that were referencing `npm run bootstrap`.


augustjk (approved):

.eslintrc.json
```diff
@@ -61,6 +76,29 @@
...
         "@typescript-eslint/no-explicit-any": "off"
       }
+    },
+    {
+      // Files listed here don't need to declare their imports in their
+      // immediate package.json "dependencies". This should include all
+      // internal-only files, like tests, which are allowed to import from
+      // "devDependencies", and also from packages declared only in the root
+      // monorepo "package.json".
+      "files": [
```
justinfagnani, line 86: I think we should exclude all starter kits from this exception

PR #3120, opened 2022-07-07
Bind this to custom attribute converter methods

### LINKED ISSUE #3108: [reactive-element] `ComplexAttributeConverter`, `this` is `undefined` in `fromAttribute` method

### Description

I try to implement a `ComplexAttributeConverter` whereas I want to pass a default value to it that should be used as a fallback in the `fromAttribute` method. Unfortunately `this` is undefined in this method. I debugged why this is the case and it seems the function is not called via [apply()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) which results in `this` being undefined.

I wonder why is this the case. Can we fix this?

### Reproduction

- Implement converter that allows passing a default value via constructor

``` ts
import { ComplexAttributeConverter } from "lit-element";

export class IntegerAttributeConverter implements ComplexAttributeConverter<Number> {
  private _defaultValue: Number;

  constructor(defaultValue: Number) {
    this._defaultValue = defaultValue;
  }

  toAttribute(value: Number, _type?: unknown): unknown {
    return `${value}`;
  }

  fromAttribute(value: string|null, _type?: unknown): Number {
    if (value) {
      const result = Number.parseInt(value, 10);

      if (isNaN(result))
        return this._defaultValue;
      else
        return result;
    }
    else
      return this._defaultValue;
  }
}
```

- Use converter in a property decoration

``` ts
const defaultPageSize: number = 10;

@property({ type: Number, converter: new IntegerAttributeConverter(defaultPageSize) })
  pageSize: number = defaultPageSize;
``` 

### Workaround

Implement the converter as a callback function

``` ts
const defaultPageSize: number = 10;

@property({
  type: Number,
  converter: (attributeValue: string | null, _type?: unknown): Number => {
    // same code as IntegerAttributeConverter.fromAttribute with access to defaultPageSize
  })
pageSize: number = defaultPageSize;
```

### Is this a regression?

### Affected versions

2.0.2

Doesn't matter

nebarf:
Fixes #3108

### DIFF
--- a/.changeset/old-tables-own.md
+++ b/.changeset/old-tables-own.md
@@ -1,5 +1,6 @@
 ---
 '@lit/reactive-element': patch
+'lit': patch
 ---
 
 Bind `this` to custom attribute converter methods
--- a/.changeset/old-tables-own.md
+++ b/.changeset/old-tables-own.md
@@ -2,4 +2,4 @@
 '@lit/reactive-element': patch
 ---
 
-Bind this to custom attribute converter methods
+Bind `this` to custom attribute converter methods
--- a/.changeset/old-tables-own.md
+++ b/.changeset/old-tables-own.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Bind this to custom attribute converter methods
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1127,10 +1127,10 @@ export abstract class ReactiveElement
           : defaultConverter;
       // mark state reflecting
       this.__reflectingProperty = propName;
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       this[propName as keyof this] = converter.fromAttribute!(
         value,
         options.type
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ) as any;
       // mark state not reflecting
       this.__reflectingProperty = null;
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1069,12 +1069,12 @@ export abstract class ReactiveElement
       this.constructor as typeof ReactiveElement
     ).__attributeNameForProperty(name, options);
     if (attr !== undefined && options.reflect === true) {
-      const converter = options.converter;
-      const toAttribute =
-        (converter as ComplexAttributeConverter)?.toAttribute?.bind(
-          converter
-        ) ?? defaultConverter.toAttribute;
-      const attrValue = toAttribute!(value, options.type);
+      const converter =
+        (options.converter as ComplexAttributeConverter)?.toAttribute !==
+        undefined
+          ? (options.converter as ComplexAttributeConverter)
+          : defaultConverter;
+      const attrValue = converter.toAttribute!(value, options.type);
       if (
         DEV_MODE &&
         (this.constructor as typeof ReactiveElement).enabledWarnings!.indexOf(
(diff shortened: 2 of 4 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -1069,9 +1069,11 @@ export abstract class ReactiveElement
       this.constructor as typeof ReactiveElement
     ).__attributeNameForProperty(name, options);
     if (attr !== undefined && options.reflect === true) {
+      const converter = options.converter;
       const toAttribute =
-        (options.converter as ComplexAttributeConverter)?.toAttribute ??
-        defaultConverter.toAttribute;
+        (converter as ComplexAttributeConverter)?.toAttribute?.bind(
+          converter
+        ) ?? defaultConverter.toAttribute;
```
justinfagnani, line 1076: If it wouldn't be a breaking change, it'd be nice and simple to just do:

```ts
const attrValue = (converter ?? defaultConverter).toAttribute?.(value, options.type);
```

but this would change the behavior of the case where there's a converter but with no `toAttribute`. cc @sorvell

justinfagnani (changes requested):
Thanks for the fix @nebarf !

I made some comments around creating fewer short-lived function objects due to `Function.bind()` when setting properties/attributes.

A bigger change that we should consider outside of this fix (I think) would be to do the `toAttribute`/`fromAttribute` determination in `createProperty()` so it's done once per class, not on every property set. cc @sorvell

packages/reactive-element/src/reactive-element.ts
nebarf: It would be a breaking change in case a converter is provided as function. Quoting from docs:

> If converter is a **function**, it is used in place of **fromAttribute**

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1069,9 +1069,11 @@ export abstract class ReactiveElement
       this.constructor as typeof ReactiveElement
     ).__attributeNameForProperty(name, options);
     if (attr !== undefined && options.reflect === true) {
+      const converter = options.converter;
       const toAttribute =
-        (options.converter as ComplexAttributeConverter)?.toAttribute ??
-        defaultConverter.toAttribute;
+        (converter as ComplexAttributeConverter)?.toAttribute?.bind(
```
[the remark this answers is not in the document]
nebarf: Good hint! Thanks

.changeset/old-tables-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Bind this to custom attribute converter methods
```
augustjk, line 5: We'll want to bump `lit` along with `reactive-element` and backticks around `this` so it's more apparent that it's referring to the keyword.
```suggestion
---
'@lit/reactive-element': patch
'lit': patch
---

Bind `this` to custom attribute converter methods
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1117,17 +1119,19 @@ export abstract class ReactiveElement
...
+          ? options.converter
+          : defaultConverter;
       // mark state reflecting
       this.__reflectingProperty = propName;
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      this[propName as keyof this] = fromAttribute!(value, options.type) as any;
+      this[propName as keyof this] = converter.fromAttribute!(
+        value,
+        options.type
+      ) as any;
```
augustjk, line 1134: The eslint comment should be moved to be above this line

justinfagnani:
Thanks for the additional review @augustjk 

@nebarf with those comments addressed this looks great!

augustjk (approved):
Awesome, thank you! I just went ahead and combined the changeset files together which is our preferred way.

justinfagnani (approved):
LGTM!

Issue #2844, opened 2022-05-09
[infra] Remove custom import remapping


justinfagnani:
Currently we use a [custom Rollup plugin](https://github.com/lit/lit/blob/main/packages/tests/rollup-resolve-remap.js) to rewrite import specifiers in tests so that we can run our tests against both development and production builds.

This should be unnecessary as tools add support for Node package exports, import conditions, and package self-references. We should be about there now:

- `plugin-node-resolve` supports
  - An [`exportConditions` option](https://github.com/rollup/plugins/tree/master/packages/node-resolve/#exportconditions)
  - Package self-references, which are resolved via the import conditions. This allows a test to import `lit-html` and get the either development or production build based on settings.
- Web Test Runner and Web Dev Server forward that option to `plugin-node-resolve`
- [TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#package-json-exports-imports-and-self-referencing) supports
  - A `"types"` condition to tell it where to look for type declarations independent of any other condition.
  - Package self-references, by which a package can import it's own modules and still resolve types

To remove `rollup-resolve-remap` we'll need to:
- [ ] Upgrade to TypeScript 4.7
- [ ] Add `"types"` condition to all exports
- [ ] Change our tsconfig's `"moduleResolution"` to`"Node16"`
- [ ] Change all imports from tests into source to use self-references.
- [ ] Delete `tests/rollup-resolve-remap.js` and `tests/wtr-config.js`
- [ ] Add `nodeResolve.exportConditions` to `tests/web-test-runner.config.js` with a value derived from `process.env.MODE`
- [ ] (TBD) Use downlevel-dts to not break older TypeScript versions with newer .d.ts features.

PR #3132, opened 2022-07-14
[infra] Replace custom test module remapping with package exports

aomarks:
Switches from a custom web test runner plugin to the now built-in export conditions configuration, in order to switch between dev and prod builds in our browser tests.

### Background

Some of our packages contain 2 separate builds: dev and prod. Dev is the raw output from TypeScript and includes extra debugging features. Prod is the minified build that users will get in production, and excludes those debugging features.

The two modes are described in our `package.json` files using [export conditions](https://nodejs.org/api/packages.html#conditional-exports). By default, users get the prod build. If they set the `development` export condition in tools like rollup, webpack, node, they will get the dev build.

We run all of our tests against both builds to get full coverage. Previously, we did this with a custom plugin for web test runner which rewrote `development/` paths to production paths when needed. We did this because web test runner did not, at the time, have support for export conditions.

Now, web test runner does have support for export conditions.

### Changes

- Switches all tests to import their respective libraries using *package self-references*, instead of relative paths. This allows export conditions to take effect *even from within the same package*.

- Switches to the `NodeNext` mode for the `module` and `moduleResolution` settings in our `tsconfig.json` files. This is needed for package self-references.

- Adds `types` entries to the export conditions in our `package.json` files. This is needed when using `NodeNext`, and is how TypeScript discovers the typings for each specific package export. (Side note, in the future, this feature will allow us to remove the `.d.ts` files that we currently duplicate into the top-level of the package, but we can't do that until our users are also using `NodeNext`).

- Added a new web test runner middleware that enforces that we only import dev sources in dev mode, and prod sources in prod mode. It serves a 403 forbidden when this is violated, which fails the test. This should prevent us from accidentally using relative imports in tests in the future.

- Deleted the old path-rewriting web test runner plugin, which is no longer needed.

- Upgraded all `@web/test-runner` and `@web/dev-server` packages, and consolidated them where possible into the top-level `package.json` for easier upgrades.

Fixes https://github.com/lit/lit/issues/2844
Fixes https://github.com/lit/lit/issues/3091


packages/lit-element/src/test/polyfill-support/lit-element-ce-sd-noPatch_test.html
```diff
@@ -32,7 +32,7 @@
         mocha,
         sessionFinished,
         sessionFailed,
-      } from '../../../../tests/node_modules/@web/test-runner-mocha/dist/standalone.js';
+      } from '../../../../../node_modules/@web/test-runner-mocha/dist/standalone.js';
       // import { mocha, sessionFinished, sessionFailed } from '@web/test-runner-mocha';
```
augustjk, line 36: I tried switching to the non-relative import and got the following logged in the dev mode run
```
 🚧 Browser logs on Webkit:
      An error was thrown in a Promise outside a test. Did you forget to await a function or assertion?
      Error: test error
        at render (../lit-element/src/test/lit-element_test.ts:257:26)
        at update (../lit-element/src/lit-element.ts:160:29)
        at performUpdate (../reactive-element/src/reactive-element.ts:1303:20)
        at ../reactive-element/src/reactive-element.ts:1207:38
```
But the tests passed. 🤷 Probably not worth trying to investigate for now.

augustjk (approved):
Awesome!

packages/lit-element/src/test/polyfill-support/lit-element-ce-sd-noPatch_test.html
aomarks: On this line in the HTML test? This is a little bit of a different context to the relative/self-reference change in this PR. To test that the dev/prod check is working, you'll want to change an import in one of the `src/tests/*.ts` files that is importing a library from `../something.js`

augustjk: Oh right. This is definitely unrelated to changes in this PR since it's not a self reference, nor is this a ts file. 🤦 I just saw the comment above saying node resolution didn't seem to work there and thought I'd give it a whirl. Please disregard!

justinfagnani (approved):
Awesome work!

PR #3136, opened 2022-07-15
[infra] Upgrade (almost) all dependencies

aomarks:
- Upgrades almost all dependencies, both dev and prod.

- There were only 2 deps I didn't tackle in this PR, because it seems like they'll need a little more work:
    - parse5 6 -> 7 (cc @43081j, seems like the types API is substantially different now, so I wasn't able to easily do this upgrade).
    - react 17 -> 18 (cc @justinfagnani @taylor-vann I wasn't sure what the implications/requirements for upgrading to react 18 will be, so I didn't even try).

- Consolidated some very common devDeps (e.g. uvu, mocha) into the top-level package.json, for easier future upgrade.

- Removes some unnecessary dependencies I found.

- Updates a few spots where typings changed.

- Removes 350MB (1.6GB -> 1.25GB) and 999 packages (3122 -> 2123), from our install.

- FYI these external PRs were also needed to get things fully upgraded:
  - https://github.com/kenchris/urlpattern-polyfill/pull/101
  - https://github.com/Polymer/tachometer/pull/239


43081j:
i would also hold off parse5 until some of the issues have been ironed out.

while trying to finish off the successor to dom5, i've noticed a couple of annoyances/problems in the new version. would be good to get those fixed before upgrading here (one major one is inikulin/parse5#619)

packages/lit/src/test/lit-element-bundle_test.html
```diff
@@ -1,6 +1,6 @@
 <html>
   <head>
-    <link rel="stylesheet" href="../node_modules/mocha/mocha.css" />
+    <link rel="stylesheet" href="../../../node_modules/mocha/mocha.css" />
```
augustjk, line 3: Does this need an extra `../`?

.changeset/lemon-pens-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/localize-tools': patch
+---
+
+Updated xmldom dependency. Minor change to XML attribute formatting can be expected.
```
augustjk, line 5: Sorry I haven't tested this, but do you happen to know if older versions of `.xlf` files would be read correctly? It seems now `>` is also being escaped to `&gt;` so wondering if this might be a breaking change for users. Since it's only for the text nodes so I imagine it probably shouldn't be breaking.

augustjk (approved):
Amazing!

packages/labs/virtualizer/package.json
```diff
@@ -47,7 +47,7 @@
       ]
     },
     "build:copy-polyfill": {
-      "command": "copyfiles -u 1 src/polyfills/resize-observer-polyfill/ResizeObserver.js .",
+      "command": "mkdir -p polyfills/resize-observer-polyfill && cp src/polyfills/resize-observer-polyfill/ResizeObserver.js polyfills/resize-observer-polyfill/ResizeObserver.js",
```
AndrewJakubowicz, line 50: This change is to remove the dep on `copyfiles`?

packages/labs/virtualizer/rollup.config.js
```diff
@@ -22,9 +22,7 @@ export default [
         warnings: true,
         module: true,
       }),
-      filesize({
-        showBrotliSize: true,
-      }),
+      summary(),
```
AndrewJakubowicz, line 25: Manually tested this and it looks great! The summary is missing the size of the compressed `brotli` values.

Worth adding back the `showBrotliSize` option?  https://yousifalraheem.github.io/rollup-plugin-summary/#:~:text=300000-,showBrotliSize,-%3F

AndrewJakubowicz: Actually, maybe we don't need this since it's not in the rollup-common config.

AndrewJakubowicz (approved):

packages/lit/src/test/lit-element-bundle_test.html
aomarks: Yes, I think that's right. Done.

packages/labs/virtualizer/package.json
aomarks: Yep

packages/labs/virtualizer/rollup.config.js
aomarks: It's true by default in `rollup-plugin-summary`

.changeset/lemon-pens-move.md
aomarks: Yeah, both are valid XML serializations (only the < technically needs to be escaped). I  manually tested that the older serialization is parsed in the same way, and there should already be some coverage in our integration tests also, because I only updated the goldens, not the inputs (which includes existing xlf files).

packages/labs/virtualizer/rollup.config.js
AndrewJakubowicz: Manually running `npm run checksize` only shows the size uncompressed, but I do see the `default true` documentation.

It looks like in the implementation the default is false (from a cursory look): https://github.com/yousifalraheem/rollup-plugin-summary/blob/main/src/index.ts#L20-L22

aomarks: Good catch, you're right! I added `showBrotliSize` and `showGzippedSize` to all the invocations.

aomarks: Filed https://github.com/yousifalraheem/rollup-plugin-summary/issues/52

AndrewJakubowicz (approved):

PR #3138, opened 2022-07-15
[labs/ssr] Inline source in source maps

### LINKED ISSUE #2645: [infra] Incorrect sourcemaps causing warning in Astro, Vite
# Description

`dom-shim.js.map` sourcemap points to incorrect source file, causing a Vite warning in Vite, Astro, etc.

```
Sourcemap for "/home/projects/owaiwjnnp.github/node_modules/@lit-labs/ssr/lib/dom-shim.js" 
points to missing source files
```

The issue appears to be this line of the sourcemap: https://unpkg.com/browse/@lit-labs/ssr@2.0.4/lib/dom-shim.js.map

```
"sources":["../src/lib/dom-shim.ts"]
```

Where `@lit-labs/ssr/src/lib/dom-shim.ts` does not exist in package. 

The line is most likely meant to point to `../lib/dom-shim.ts` or even `./dom-shim.ts`.

This warning has probably existed for a while in Vite, but is new in Astro as we stopped swallowing/hiding Vite warnings from the user.

# Steps to Reproduce

1. Start a new Lit + Astro project: https://astro.new/framework-lit?on=stackblitz
2. See the warning in your console on startup. Also occurs on build.

## Live Reproduction Link

https://astro.new/framework-lit?on=stackblitz

## Expected Results

No warnings.

## Actual Results

```
Sourcemap for "/home/projects/owaiwjnnp.github/node_modules/@lit-labs/ssr/lib/dom-shim.js" 
points to missing source files
```

augustjk:
Fixes #2645 

tsconfig was missing `inlineSources` option for labs/ssr package.

~~Alternative fix would be to actually include the source ts files in the npm package. Since this package is meant for server/node use only, would that potentially be preferable?~~
Will stick with inline sourcemap to be consistent with the rest of our packages.

### DIFF
--- a/.changeset/fifty-shirts-flash.md
+++ b/.changeset/fifty-shirts-flash.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+JS source maps now include the TS source inline
--- a/packages/labs/ssr/tsconfig.json
+++ b/packages/labs/ssr/tsconfig.json
@@ -8,6 +8,7 @@
     "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
+    "inlineSources": true,
     "outDir": ".",
     "rootDir": "./src",
     // For module loader tests:


aomarks (approved):
Nice!

augustjk:
@justinfagnani I want to double check if inlining source here is the preferable way. You've mentioned here https://github.com/lit/lit.dev/issues/872#issuecomment-1168046720 that we choose to not publish ts files. Is that also the case for packages like labs/ssr that are meant to be used in the server/node, and potentially in deno as well?

augustjk:
Per offline discussion, we'll go with inlining sources for now as strict improvement for the current issue and re-evaluate whether we want to switch to publishing ts source files down the line.

PR #3139, opened 2022-07-16
Update release-image such that it can accept multiple packages and versions

AndrewJakubowicz:
Part of https://github.com/lit/lit/issues/2503

### Context

Previously the release-image tool could only take one single CHANGELOG and an optional version. This PR extends the tool to concatenate multiple CHANGELOG files and versions making the tool more flexible. This will allow generating a single image for a release.

### Features

 - CLI can now take multiple files and versions with the following pattern: `release-image (-f CHANGELOG_PATH [-v VERSION])...`. If the same changelog is provided multiple times with different versions, the image contains those versions.
 - The packge/release notes are added to the image in the order of the CLI arguments.

For example: `release-image lit-html/CHANGELOG.md -v 2.2.2 -f lit-html/CHANGELOG.md -v 2.2.1` generates the following image:
release

If the versions are omitted, the most recent version is returned for each package. For example: `release-image ../lit-html/CHANGELOG.md ../reactive-element/CHANGELOG.md ../lit/CHANGELOG.md` currently generates:
release

This was tested manually and the risk is minimal.

### DIFF
--- a/.changeset/small-doors-develop.md
+++ b/.changeset/small-doors-develop.md
@@ -1,5 +1,2 @@
 ---
-'@lit-internal/scripts': minor
 ---
-
-Add multiple changelogs and version support to the release-image tool.
--- a/.changeset/small-doors-develop.md
+++ b/.changeset/small-doors-develop.md
@@ -1,0 +1,5 @@
+---
+'@lit-internal/scripts': minor
+---
+
+Add multiple changelogs and version support to the release-image tool.
--- a/packages/internal-scripts/src/release-image.ts
+++ b/packages/internal-scripts/src/release-image.ts
@@ -12,47 +12,90 @@ import {readFile} from 'fs/promises';
 import path from 'path';
 
 const optionDefinitions = [
-  {name: 'file', type: String, defaultOption: true},
-  {name: 'version', type: String},
+  {
+    name: 'files',
+    alias: 'f',
+    type: String,
+    defaultOption: true,
+    multiple: true,
+  },
+  {name: 'versions', alias: 'v', type: String, multiple: true},
 ];
+
+/**
+ * A cache of the parsed changelogs such that a package may be referenced
+ * multiple times to render multiple versions.
+ */
+const CHANGELOG_CACHE = new Map<string, Changelog>();
 
 export const run = async () => {
   const options = commandLineArgs(optionDefinitions);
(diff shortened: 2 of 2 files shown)


abdonrd:
As a recommendation...
Would it make sense to delete the PR numbers and commit hashes?
In a Twitter image I suppose it does not add much value.

.changeset/small-doors-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-internal/scripts': minor
```
augustjk, line 2: Do we care to have changesets for private/internal packages? I can see it potentially having value for tracking changes on this without looking at the git history. It won't be published either way.

packages/internal-scripts/src/release-image.ts
```diff
@@ -12,47 +12,90 @@ import {readFile} from 'fs/promises';
...
+    let changelog = CHANGELOG_CACHE.get(filename);
+    if (!changelog) {
+      const packageJson = JSON.parse(
+        await readFile(path.join(path.dirname(filename), 'package.json'), {
+          encoding: 'utf-8',
+        })
+      );
+      const packageName = packageJson.name as string;
+
+      console.log(`Reading ${packageName} release ${version} from ${filename}`);
```
augustjk, line 80: A minor, totally unnecessary, suggestion. I found it jarring to see "release undefined" get logged when this runs without a version.
```suggestion
      console.log(`Reading ${packageName} release ${version ?? 'latest'} from ${filename}`);
```

augustjk (approved):
This is great!
While I agree the PR and commit hash serves little purpose in an image, I personally don't mind seeing them there. It kind of makes them look changeloggy. 😅 Totally on board if you'd like to remove them though.

.changeset/small-doors-develop.md
AndrewJakubowicz: Great catch! I blindly followed Changeset but 100% agree that in the release we prefer to not have this.

AndrewJakubowicz:
Instead of making changes to commit and PR numbers here, opened an issue to revisit the visuals of the release image: https://github.com/lit/lit/issues/3145

Scoping this change to just handling the multiple packages.

aomarks (approved):
Nice!

PR #3146, opened 2022-07-18
[labs/react] Property and event setter updates

taylor-vann:
This PR focus on logic clarity during property updates.

Changes include:
- Bailing early on matching properties
- Removing nested `if` statements to update logic


packages/labs/react/src/create-component.ts
```diff
@@ -25,27 +25,24 @@ const listenedEvents: WeakMap<
...
-      // Otherwise just update the listener with new value
-    } else {
-      handler.handleEvent = listener;
-    }
-    // Remove listener if one exists and value is undefined
-  } else if (handler !== undefined) {
-    events.delete(event);
-    node.removeEventListener(event, handler);
+    events.set(eventName, listener);
+    node.addEventListener(eventName, listener);
```
justinfagnani, line 43: This will thrash `removeEventListener`/`addEventListener` in cases where the event handler is an inline closure, like:

```tsx
<MyElement onFoo={(e) => console.log(e)}></MyElement>
```

This has a measurable perf overhead in some browsers and is why we only call `removeEventListener()` when the listener goes from defined to undefined and `addEventListener()` when the listener goes from undefined to defined.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
```diff
@@ -60,16 +57,16 @@ const setProperty = <E extends Element>(
   old: unknown,
   events?: Events
 ) => {
+  // bail early if values have not changed
+  if (value === old) return;
```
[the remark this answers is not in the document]
taylor-vann: Oh dang that's disappointing but understandable. I figured it'd be more expensive to update the DOM than not, was trying to match the top -> down or bail behavior vdom has

packages/labs/react/src/create-component.ts
taylor-vann: I see! That makes sense.

I thrashing is the default behavior of react anyways no? LIke `add/removeEventListnener` with an inline function would thrash every render.

So we're opting out of react behavior to provide a more performant option?

taylor-vann: So we're using an `EventListenerObject` to provide a context to attach and swap an `EventListener` without relying on the `removeEventListener` when handlers change but are undefined.

justinfagnani: AFAIU, React uses event delegation and won't thrash add/removeEventListnener. Since we're handling events outside of React we need to take precautions ourselves. We do the same thing in lit-html's `@event` binding support.

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
justinfagnani: Setting properties doesn't necessarily update the DOM... that's dependent on what the element does with that property. For us it's often a setter that does it's own change detection and maybe triggers a batched update, which then might only set other properties...

packages/labs/react/src/create-component.ts
taylor-vann: You're correct! React does the event hub thing and it'd be cheaper than add/remove :) my bad was more clarifying for my understanding. also cool! did not know that about event binding

taylor-vann:
closing because this PR became very aesthetic and that's not what i was aiming for. The logic for updates is fine :) just wanted it to read a little easier. But the changes increase code size and doesn't improve readability as much as I'd like.

PR #3148, opened 2022-07-19
[internal/scripts] Add markdownFile option to release-image script

AndrewJakubowicz:
Required for https://github.com/lit/lit/issues/2503

Update the release-image script so it can be passed a markdown file. This is required for the following PR which uses the Github PR description to create a release-image.

This may also become useful if a custom image is ever desired with arbitrary markdown contents.

### Features

 - Extend release-image so it can take in and render a markdown file via `--markdownFile` or `-m`.

### Testing

Tested by creating a release-image of this description:

release


scripts/fetch-release-changelog.js
```diff
@@ -0,0 +1,53 @@
...
+import * as https from 'https';
+
+/**
+ * Take the `pullRequestNumber` output of the `changesets/action` GitHub action
+ * on standard input and return the CHANGELOG from the PR description in stdout.
+ *
+ * https://github.com/changesets/action#outputs
+ */
+
+const pullRequestNumber = fs.readFileSync(process.stdin.fd, 'utf-8')?.trim();
```
augustjk, line 18: Could this be provided as `env` in the run step and read from `process.env`?
Might also want to throw if this is undefined.

packages/internal-scripts/src/release-image.ts
```diff
@@ -166,8 +161,41 @@ EXAMPLES
...
+  multiple versions of the same package:
+
+      release-image -f reactive-element/CHANGELOG.md -v 3.2.0 \\
+                    -f lit-html/CHANGELOG.md -v 2.0.1 \\
+                    -f lit-html/CHANGELOG.md -v 2.0.0
+
+  To pass arbitrary contents into the image <body>, use the --releaseContent
+  option (or -r):
+
+      release-image -r "<p>Arbitrary Release Contents</p>"
```
augustjk, line 193: Perhaps worth noting the `releaseContent` will be passed through a markdown parser as it is embedded to the body so can be markdown string?

.github/workflows/release.yaml
```diff
@@ -64,6 +64,22 @@ jobs:
...
+        if: steps.cs.outputs.published == 'false'
+        working-directory: lit/packages/internal-scripts
+        run: npm run build
+
+      - name: Generate release image artifact
+        if: steps.cs.outputs.published == 'false'
+        with:
+          cwd: lit
+        run: |
+          $(node scripts/fetch-release-changelog.js <<< ${{ steps.cs.outputs.pullRequestNumber }})
```
augustjk, line 77: It's unfortunate that we have to fetch this separately when the versions action above already has it. This does work but I wonder if we might either make it a feature request to output the pr body or make a fork? https://github.com/changesets/action/blob/6a46d2c28377497e64763ba5f1faeabbd3d88c33/src/run.ts#L367-L369

This is probably more of a niche use case though.. it may be more effort than we want.

packages/internal-scripts/src/release-image.ts
AndrewJakubowicz: ^ Great comment!
Additional async suggestion from @bicknellr - Change this flag to accept a markdown file which is inserted instead of raw content.

scripts/fetch-release-changelog.js
AndrewJakubowicz: +1 added verification and throwing. Since we don't need the pr number anywhere else, I've opted to keep as a passed in value.

packages/internal-scripts/src/release-image.ts
AndrewJakubowicz: Addressed by changing this to `markdownFile`, which also communicates that markdown is parsed.

.github/workflows/release.yaml
AndrewJakubowicz: Agree. Once this approach is tested and I have more confidence in it I can file an issue with the feature request.

.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,38 @@
+name: Generate Release Image
+
+on: [pull_request]
+
+jobs:
+  release-image:
+    # Only generate the release image from the PR contents if the PR was in the
+    # monorepo and opened by lit-robot.
+    if: github.repository == 'lit/lit' && github.actor == 'lit-robot'
```
augustjk, line 9: No action needed.

We use the same `github.actor` check for not running the lint & changeset workflow step for the "Version Packages" PR. I've noticed that when another changeset workflow is triggered while a "Version Packages" PR is open, the workflow updates the existing PR but "github-actions" becomes the actor (even though the commit is authored by lit-robot) due to the credentials used to force push and update the PR.

So I believe this workflow will only run on the initial PR creation. Updates must happen by closing that PR and re-opening. We currently already do this because workflows sometimes don't run when the PR is updated anyway so this will work for us as is.

scripts/fetch-release-changelog.js
```diff
@@ -0,0 +1,57 @@
...
+ * on standard input and return the CHANGELOG from the PR description in stdout.
+ *
+ * https://github.com/changesets/action#outputs
+ */
+
+const pullRequestNumber = fs.readFileSync(process.stdin.fd, 'utf-8').trim();
+if (!pullRequestNumber || isNaN(parseInt(pullRequestNumber, 10))) {
+  console.error(`No pullRequestNumber from changeset action.`);
+  process.exit(1);
+}
```
augustjk, line 22: In the very unlikely event that we enter this block, the release-image would be called with a path with content. I think that would generate an empty image?

I can't think of it ever happening if this is used properly only ever in this workflow so I don't know if we need to really guard for that.

scripts/fetch-release-changelog.js
```diff
@@ -0,0 +1,57 @@
...
+      process.exit(1);
+    });
+});
+
+const {body} = response;
+
+// Strip the preamble from the PR description, and only capture the changelog
+// contents.
+const headingIdx = body.indexOf('# Releases');
+console.log(body.slice(headingIdx === -1 ? 0 : headingIdx));
```
augustjk, line 57: TIL about [process substitution](http://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html#Process-Substitution)

Is there a benefit to doing this and providing to `release-image` with it as opposed to just writing to a fixed filename and providing that filename to `release-image`?

.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: NPM install
+        run: npm ci
+
+      - name: Build release-image script
+        working-directory: packages/internal-scripts
+        run: npm run build
+
+      - name: Create release image
+        run: node_modules/.bin/release-image -m <(node scripts/fetch-release-changelog.js <<< ${{ github.event.pull_request.number }})
```
augustjk, line 32: Just curious.. does this not work without `<<<` and reading from `process.argv[2]` within the script?

scripts/fetch-release-changelog.js
AndrewJakubowicz: Discussed offline. The main benefit is that process substitution handles the temporary intermediate file without needing to maintain that reference in a temporary variable. Because the scope of the script is small, this is probably fine.

If the scope increased and it could become necessary to add additional console logs, then this decision should be revisited.

.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,38 @@
+name: Generate Release Image
+
+on: [pull_request]
```
augustjk, line 3: I'm having second thoughts about whether having this as a separate workflow like this is preferable. It adds an extra line in the list of checks for all opened PR, though this one would be skipped unless it was made by lit-robot.

I think in our discussion offline, we were considering if we want to edit the PR description and re-run it, though I think this wouldn't run in that case because if we're the one re-triggering, is the `github.actor` value still lit-robot?

AndrewJakubowicz: The primary reason I moved it to a separate workflow is so the image artifact is local and attached to the release PR. Otherwise the image artifact ends up on the "released" commit, instead of the pull request. This makes the DX of finding the release image a little harder.

Another alternative is generating the image in the release action, and then adding a "download artifact" step somewhere on the "Version Packages" PR. However that would probably still need an additional action.

augustjk: I see.. The trigger is run on the latest commit in main that we approve. I suppose it might be possible to have it run there, but add a link to the artifact as a comment to the Version Packages PR? Though apparently there isn't an easy way to do that https://github.com/actions/upload-artifact/issues/50. We'd likely need another step/script that uses github api to fetch the artifact id to create a link to it since the artifact upload workflow doesn't provide that, and another api call to post the comment to the correct PR.

It shouldn't be a huge deal but I'd prefer if we don't have to see the "Generate Release Image" workflow on every single PR. But maybe the effort to achieve that is not worth it when functionally this works.

.github/workflows/release-image.yaml
AndrewJakubowicz: Done! Great suggestion!

augustjk (approved):
Nice enhancement to the release-image script!

.github/workflows/release-image.yaml
AndrewJakubowicz: This comment thread was resolved in https://github.com/lit/lit/pull/3176 by using the Github Actions `paths` filter.  Thank you!

aomarks (approved):

.github/workflows/release-image.yaml
AndrewJakubowicz: Noting that this was a brilliant comment - and resulted in a change in approach that is better. Thank you!

PR #3150, opened 2022-07-19
[infra] Misc wireit tweaks

aomarks:
Makes a few more package scripts cacheable by adding input/output files, and fixes a syntax error in one case.

### DIFF
--- a/packages/labs/gen-wrapper-vue/test-output/package.json
+++ b/packages/labs/gen-wrapper-vue/test-output/package.json
@@ -42,7 +42,8 @@
         "installSelf"
       ],
       "files": [
-        "src/**/*.ts, src/**/*.vue",
+        "src/**/*.ts",
+        "src/**/*.vue",
         "tsconfig.json"
       ],
       "command": "vue-tsc --noEmit && vite build",
--- a/packages/labs/test-projects/test-element-a/package.json
+++ b/packages/labs/test-projects/test-element-a/package.json
@@ -36,6 +36,10 @@
         "build"
       ],
       "command": "npm pack",
+      "files": [
+        "package.json",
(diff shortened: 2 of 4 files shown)


packages/labs/test-projects/test-element-a/package.json
```diff
@@ -36,6 +36,10 @@
         "build"
       ],
       "command": "npm pack",
+      "files": [
+        "package.json",
```
justinfagnani, line 40: .gitignore also? I believe `npm pack` etc reads that.

justinfagnani (approved):

packages/labs/gen-wrapper-vue/test-output/package.json
```diff
@@ -42,7 +42,8 @@
         "installSelf"
       ],
       "files": [
-        "src/**/*.ts, src/**/*.vue",
+        "src/**/*.ts",
+        "src/**/*.vue",
```
AndrewJakubowicz, line 46: Tried to investigate the test-local error that points to a test from this package.json. Locally was unable to repro.

AndrewJakubowicz (approved):

packages/labs/gen-wrapper-vue/test-output/package.json
augustjk: Actually, for failure in this particular pipeline, it might be that we have an erroneous cached build artifact. Unsure if there's a way to cache bust the build?

packages/labs/test-projects/test-element-a/package.json
aomarks: It doesn't read `.gitignore`. Otherwise our `.js` artifacts wouldn't get included. It does read `.npmignore`, but we don't use those -- we consistently use the allowlist `files` approach instead.

PR #3155, opened 2022-07-20
[labs/virtualizer] ResizeObserver loop limit exceeded error while scrolling

joekukish:
This PR is based on https://github.com/lit/lit/pull/3133

* Added rAF to _childrenSizeChanged.
* Removed workaround in tests to deal with this error.


graynorton (commented):
Thanks for this pull request.

It's possible that a change along these lines might eventually allow us to avoid the ResizeObserver loop limit errors, but unfortunately in this simple form the fix seems to introduce end-user-observable issues in some cases—specifically, it can cause a virtualizer to be painted before item measurements have been processed by the layout, resulting in a "flash of incorrectly laid out content". For example, the colored items below are supposed to be flush with one another, but here they are initially painted with large gaps in between them:

image

We would like to do further investigation (and hope to find a way to avoid the error), but this work is not currently at the top of our priority list because a) we expect it to be tricky, nuanced work; b) as far as we know, the error has no end-user impact; and c) there is a relatively straightforward, if annoying workaround (to ignore the benign error messages in automated tests).

For now, I think we should close this PR, while leaving the associated issue open.

Issue #3079, opened 2022-06-23
[labs/ssr] Make Lit components less explody in Node


aomarks:

augustjk:
Related https://github.com/lit/lit/issues/2524

abdonrd:
Probably related https://github.com/lit/lit/discussions/2259

PR #3156, opened 2022-07-20
[core/ssr] Add Node builds that don't explode

aomarks:
### Background

Before this PR, importing Lit with Node causes errors such as `window is not defined`. The only way to load Lit was to first ensure that the `dom-shim.js` module provided by the `@lit-labs/ssr` package was loaded, which defines `window` plus APIs like `HTMLElement` and `customElements`.

This meant that using Lit in frameworks such as Next did not work out-of-the-box _even when only client-side rendering is needed_.

### This change

After this PR, Lit can be imported from Node without errors, and custom elements can be defined as no-ops.

This doesn't remove the need for loading the DOM shim shim when SSR is actually needed, but it does give us compatibility at least for client-side rendering in frameworks like Next, without the user needing to do anything special.

This works by adding a Node-specific Rollup build and a `node` export condition to `lit-html` and `@lit/reactive-element`. Neither `lit-element` nor `lit` need a Node build, because only the underlying 2 libraries currently need Node-specific behavior.

This only defines `customElements` as new globals. It does not define `window`, `HTMLElement`, or any other APIs. This way, we won't affect the behavior of libraries that detect whether they are in Node vs the browser by checking for a `window` global.

For testing, I've added a `node-imports.ts` file to each package, and added a new `node:test` script which executes that module with Node. This confirms that Node doesn't crash on import.

Fixes https://github.com/lit/lit/issues/3079

### Followup

In a follow up, I will propose moving the actual functional dom-shim implementations of `HTMLElement` and `customElements` from `@lit-labs/ssr` into `reactive-element` (in this PR they are just no-ops) so that ultimately the user never has to load the dom-shim manually. 

However, in order to do this, we will need to update the Lit SSR `ModuleLoader` to understand export conditions -- because currently it does not support them, so it will not find the Node build. We depend on the `resolve` package to do module resolution, which unfortunately does not yet support export conditions, so this work will be a little bit non-trivial.

Another followup will be to add an `isSsr` const or function. Rather than doing runtime detection, this can simply be `true` in the node build and `false` in the browser build.

### Size

This actually slightly drops the size of the main `lit-html.js` and `reactive-element.js` production build files:

| File                | Old size (raw/gzip/brotli)      | New size (raw/gzip/brotli) |
| ------------------- | -------------- | -------------- |
| lit-html.js         | 8.68 / 3.45 / 3.13 | 8.02 / 3.38 / 3.07 |
| reactive-element.js | 5.97 / 2.17 / 1.93 | 5.95 / 2.17 / 1.92 |


rollup-common.js
```diff
@@ -423,6 +425,58 @@ export function litProdConfig({
...
+              // 1. It should be very slightly faster, even in Node where bytes
+              //    are not as important as in the browser.
+              //
+              // 2. It means we don't need a Node build for lit-element. There
+              //    is no Node-specific logic needed in lit-element. However,
+              //    lit-element and reactive-element must be consistently
+              //    minified or unminified together, because lit-element
+              //    references properties from reactive-element which will
+              //    otherwise have different names. The default export that
+              //    lit-element will use is minified.
```
augustjk, line 468: Interesting... so the alternative is to also create a node build for `lit-element`, if we wish all node builds to not be minified. I'm onboard with this based on your reasonings.

augustjk (approved):
Exciting!

packages/lit-html/src/lit-html.ts
```diff
@@ -10,6 +10,8 @@ import type {Directive, DirectiveResult, PartInfo} from './directive.js';
 const DEV_MODE = true;
 const ENABLE_EXTRA_SECURITY_HOOKS = true;
 const ENABLE_SHADYDOM_NOPATCH = true;
+const NODE_MODE = false;
+const global = NODE_MODE ? globalThis : window;
```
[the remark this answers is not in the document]
aomarks: Yep, it's for IE11 compat.

rollup-common.js
aomarks: Yeah. The other reason not listed was just to change fewer things at once. Since currently node users get the default build, which is minified, I figured we should just keep doing that, all else equal.

packages/lit-html/src/lit-html.ts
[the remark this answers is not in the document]
kevinpschaaf: Wanna throw a comment in there?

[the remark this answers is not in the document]
aomarks: Added comment

packages/reactive-element/src/reactive-element.ts
```diff
@@ -27,6 +27,16 @@ export type {
   ReactiveControllerHost,
 } from './reactive-controller.js';
 
+const NODE_MODE = false;
+const global = NODE_MODE ? globalThis : window;
+
+if (NODE_MODE) {
+  global.HTMLElement ??= class HTMLElement {} as unknown as typeof HTMLElement;
+  global.customElements ??= {
```
kevinpschaaf, line 35: All things equal, if a user didn't explicitly load the dom-shim, it'd be nice to keep as much as possible off the global. I think we could pull HTMLElement into a local variable (and shim it there) rather than put it on the global, right?

`customElements` is a little harder; we could do the same in the decorator, and that would just leave JS users to needing to do `customElements?.define(...)` in their own elements. But maybe it's worth keeping `customElements` on the global to help smooth that over?

justinfagnani: Agreed. I think if this approach is good it's exactly because it 1) puts all the cost on the node import, so 2) can go to extra lengths to isolate any shims to just this module.

I think we should be able to do something like:

```ts
const HTMLElementShim = class HTMLElement { ... }
export class ReactiveElement extends (NODE ? HTMLElementShim : HTMLElement) {
```

Regarding `window.customElements`, shouldn't that be shimmed, if at all, in the decorator definition module? And there we just need the decorator to not access it, yeah?

packages/lit-element/src/test/node-imports.ts
```diff
@@ -0,0 +1,32 @@
...
+import 'lit-element/decorators/query.js';
+import 'lit-element/decorators/query-all.js';
+import 'lit-element/decorators/query-assigned-elements.js';
+import 'lit-element/decorators/query-assigned-nodes.js';
+import 'lit-element/decorators/query-async.js';
+
+import {LitElement, html} from 'lit-element';
+import {customElement} from 'lit-element/decorators.js';
+
+@customElement('my-element')
```
sorvell, line 27: Make sense to include a version that doesn't use the decorator as well?

packages/lit-html/src/lit-html.ts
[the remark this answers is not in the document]
sorvell: What about just defining `globalThis` if it doesn't exist and using that? IE11 is not long for this world.

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -16,11 +16,14 @@ import {decorateProperty} from './base.js';
...
 
+const NODE_MODE = false;
+const global = NODE_MODE ? globalThis : window;
+
 /**
  * A tiny module scoped polyfill for HTMLSlotElement.assignedElements.
  */
 const slotAssignedElements =
-  window.HTMLSlotElement?.prototype.assignedElements != null
+  global.HTMLSlotElement?.prototype.assignedElements != null
```
sorvell, line 26: Is `HTMLSlotElement` going to be defined in Node? Is that in the dom-shim? And what about all the other element classes? Users probably won't use `?.` for these as is done here, right?

packages/reactive-element/src/test/node-imports.ts
```diff
@@ -0,0 +1,29 @@
...
+import '@lit/reactive-element/decorators/query.js';
+import '@lit/reactive-element/decorators/query-all.js';
+import '@lit/reactive-element/decorators/query-assigned-elements.js';
+import '@lit/reactive-element/decorators/query-assigned-nodes.js';
+import '@lit/reactive-element/decorators/query-async.js';
+
+import {customElement} from '@lit/reactive-element/decorators.js';
+import {ReactiveElement} from '@lit/reactive-element';
+
+@customElement('my-element')
```
sorvell, line 28: Add a version that doesn't use the decorator?

packages/reactive-element/src/reactive-element.ts
sorvell: Let's sync on this offline. It seems like Al is moving towards not having a DOM shim separate from 'lit is loaded in Node'? If that' the case, we do want to populate the global in the Node import condition, yeah?

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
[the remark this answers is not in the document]
aomarks: Defining `globalThis` adds extra bytes to the browser build. Seems like this approach gives us the best of both worlds.

I suppose the web components polyfills could define `globalThis`? But I feel a little uncertain about where that would go, requiring users to upgrade, and any potential new weirdness that could cause?

rollup-common.js
```diff
@@ -255,6 +255,7 @@ export function litProdConfig({
   packageName,
   outputDir = './',
   copyHtmlTests = true,
+  nodeBuild = false,
```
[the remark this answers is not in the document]
aomarks: Done

packages/reactive-element/src/test/node-imports.ts
aomarks: Done

packages/reactive-element/src/reactive-element.ts
aomarks: As per meeting this morning, `HTMLElement` is no longer globally shimmed, it is instead just declared inline. The reason we don't need `HTMLElement` globally is that we're only concerned with users who are extending `ReactiveElement` or `LitElement`.

I have kept the global `customElements` definitions so that it will be possible to import components that aren't defined using the decorator.

packages/reactive-element/src/decorators/query-assigned-elements.ts
aomarks: No, and the broader DOM shim doesn't define `HTMLSlotElement` either. Somebody could be doing e.g. `instanceof HTMLSlotElement` somewhere, but I think we don't need to worry about it for now.

packages/reactive-element/src/reactive-element.ts
aomarks: Just pushed another change that shims `HTMLElement` in a slightly more sneaky way. It now temporarily defines `globalThis.HTMLElement`, and then removes it after `ReactiveElement` is defined, instead of changing the extends clause at all.

The reason for this is that I found that the following code ... :

```ts
export abstract class ReactiveElement
  extends (NODE_MODE && global.HTMLElement === undefined
    ? (class HTMLElement {} as unknown as typeof HTMLElement)
    : HTMLElement)
```

... resulted in in this `d.ts` output:

```ts
declare const ReactiveElement_base: {
    new (): HTMLElement;
    prototype: HTMLElement;
};
/**
 * Base element class which manages element properties and attributes. When
 * properties change, the `update` method is asynchronously called. This method
 * should be supplied by subclassers to render updates as desired.
 * @noInheritDoc
 */
export declare abstract class ReactiveElement extends ReactiveElement_base implements ReactiveControllerHost {
```

Which, for some reason, causes ts-transformers to fail with this error:

```
        --··"../reactive-element/reactive-element.d.ts(181,55):·error·TS2749:·'ReactiveElement_base'·refers·to·a·value,·but·is·being·used·as·a·type·here.·Did·you·mean·'typeof·ReactiveElement_base'?\n",
```

I also have a suspicion that the ternary `extends` could be upsetting to Closure compiler.

Open to other suggestions, but this approach of defining and then undefining it should be the least confusing to type systems.

packages/lit-element/src/test/node-imports.ts
aomarks: Done

packages/reactive-element/src/reactive-element.ts
augustjk: I can't think of any better way. I tried playing around with it a bit but TS also did not like an exported class extending something that's module scoped.

aomarks:
Just added a fix to the `update-version-variables.js` script to account for the fact that the code is sometimes now on `global` instead of `globalThis`.

scripts/update-version-variables.js
```diff
@@ -24,12 +24,12 @@ const updateVersionVariable = async (packageDir, sourcePath, variableName) => {
 
   // Replace version number
   const versionVarRegex = new RegExp(
-    `\\(globalThis\\.${variableName} \\?\\?= \\[\\]\\)\\.push\\('\\d+\\.\\d+\\.\\d+'\\)`
+    `(\\(global(?:This)?\\.${variableName} \\?\\?= \\[\\]\\)\\.push\\(')\\d+\\.\\d+\\.\\d+[^']*('\\))`
   );
   let replaced = false;
-  const newSource = fileSource.replace(versionVarRegex, () => {
+  const newSource = fileSource.replace(versionVarRegex, (_, pre, post) => {
```
augustjk, line 30: Ooh very nice utilization of capture groups.

augustjk (approved):
Confirmed by testing pre-release that this does indeed create node builds that don't explode.

kevinpschaaf (approved):
Looks good.

Since we have to commit to this approach long-term, I'm mildly interested in how much code it would add if we just did the shimming in the normal codepath (and not have a separate node export condition and all the config file overhead that requires).

But since as you note the node export condition option will also give us a nice place to put the `isSSR` flag, seems like that having a NODE_MODE will end up being useful for a number of reasons, so I think I'm good with this direction.

aomarks:
> Looks good.
> 
> Since we have to commit to this approach long-term, I'm mildly interested in how much code it would add if we just did the shimming in the normal codepath (and not have a separate node export condition and all the config file overhead that requires).

Below is a comparison of the sizes of the two main affected files. "Prod" is the minified browser build (`NODE_MODE=false` with unreachable code removed by terser). "Combined" is if we were to just have a single build (`NODE_MODE = typeof window === 'undefined'`):

| Build | File name | Size | Gzipped | Brotli |
|-|-|-|-|-|
| Prod | lit-html.js                         | 8.02 kB  | 3.38 kB | 3.07 kB  |
| Combined | lit-html.js                         | 8.09 kB  | 3.41 kB | 3.1 kB  |
| Prod | reactive-element.js                   | 5.95 kB  | 2.17 kB | 1.92 kB |
| Combined | reactive-element.js                   | 6.16 kB  | 2.26 kB | 2 kB    |

So, a little more substantial on the `reactive-element.js` size.

Also I believe this difference will increase even further, because I plan to extend the Node `customElement.define` implementation in `reactive-element.js` to keep track of calls instead of being a no-op, so that they can be later replayed if the full fidelity shim is loaded later.

kevinpschaaf:
Cool thanks. The 90-100b on RE is definitely worth it, plus what you said about possibly expanding it a bit, so I think I'm sold.

sorvell (approved):

justinfagnani (approved):

PR #3157, opened 2022-07-21
[infra] Fix flaky ts-starter-kit tests

aomarks:
I think the flakes we have been seeing from the starter kit test were just timeouts. Since things are running in parallel, it sometimes takes longer for browser tests to start. The starter kits now run separately, after the other tests, so that they have less resource contention. We could also bump the `browserStartTimeout`, but since it's a user template repo, that seems weird.

Also:

- Upgrades dependencies
- Removes some unnecessary devDeps from virtualizer
- Adds a changelog entry from #3132 that I forgot (unrelated, just throwing it in)


AndrewJakubowicz (approved):
You legend!

augustjk (approved):
🎉

augustjk:
While I am totally on board with this to make our pipeline smoother and have excitedly approved, a tiny reservation about this is that the starter kits are also meant as templates for users to potentially replicate for their own projects. Setting a config purely based on our workflow seems less ideal.

Though a longer timeout setting is pretty innocuous, and we're also considering revamping these hopefully soon so I don't think we need to worry so much about it.

Alternative way to deflake would be to run the starter kit tests independently, either in serial or a separate job, with the similar idea as we have separate build steps for those already in our monorepo.

aomarks:
> While I am totally on board with this to make our pipeline smoother and have excitedly approved, a tiny reservation about this is that the starter kits are also meant as templates for users to potentially replicate for their own projects. Setting a config purely based on our workflow seems less ideal.
> 
> Though a longer timeout setting is pretty innocuous, and we're also considering revamping these hopefully soon so I don't think we need to worry so much about it.
> 
> Alternative way to deflake would be to run the starter kit tests independently, either in serial or a separate job, with the similar idea as we have separate build steps for those already in our monorepo.

Good point. Done.

augustjk (approved):
Thank you!!!

PR #3159, opened 2022-07-21
[infra] Fix gen-wrapper-react wireit build output

augustjk:
We were seeing pipeline failures like this https://github.com/lit/lit/runs/7457270083?check_suite_focus=true#step:8:138

Since the test files from the build step weren't being cached and build step was thinking it had restored it.

This wireit config change fixes this.


aomarks (approved):

aomarks:
Oops, good fix!

AndrewJakubowicz (approved):
Subtle!

justinfagnani (commented):
Nice find!

Issue #2951, opened 2022-05-25
[labs/react] Issues using lit-labs wrappers in Next.js


WickyNilliams:
# Description

I have some react wrappers which i have generated from my lit components. These work great in a pure client-side react app. However when using in a next.js app i hit a number of a problems. 

Elliot on slack asked me to raise an issue here. Even though i think some of it is an issue on the next.js side, it might help in tracking your next integration story. 

# Steps to Reproduce

1. Write this code

```ts
import React from "react";
import dynamic from "next/dynamic";

// @org/react contains lit-labs/react wrappers...

// (1) ts error
const MyInput = dynamic(() => import("@org/react").then(({ Input }) => Input), {
  ssr: false,
});

export const Page = () => {
  const ref = React.useRef();

  // (2) ref issue
  const focusInput = () => ref.current.focus();

  return (
    <>
      <MyInput ref={ref} />
      <button onClick={focusInput}>Click me</button>
    </>
  );
};
```

2. See this output...

(1) TS complains about the type of the react wrapper component. To fix this I had to cast to `any` and then back to my actual type. I created a helper function like:

```ts
type ComponentModule = typeof import("@org/react");

function createDynamic<
  K extends keyof ComponentsModule,
  T extends ComponentsModule[K]
>(component: K) {
  return dynamic(
    () => import("@org/react").then((module) => module[component] as any),
    { ssr: false }
  ) as T;
}

const MyInput = createDynamic("Input");
```

(2) when clicking the button you will get an error because the `ref` is not pointing to the underlying custom-element as you would expect, but rather (i think) the return type of `dynamic`.

I was able to work around this by adding some indirection and threading through a ref myself:

```ts
type ComponentsModule = typeof import("@org/react")

function createDynamic<
  K extends keyof ComponentsModule, 
  T extends ComponentsModule[K]
>(component: K) {
  const Wrapped = dynamic(
    () =>
      import("@org/react").then(module => {
        const Component = module[component] as any

        // @ts-expect-error innerRef is unknown
        const Wrapped = ({ innerRef, ...props }) => <Component ref={innerRef} {...props} />
        Wrapped.displayName = `Wrapped${component}`

        return Wrapped as any
      }),
    { ssr: false }
  )

  // @ts-expect-error innerRef is unknown
  const Forwarded = React.forwardRef((props, ref) => <Wrapped {...props} innerRef={ref} />)
  Forwarded.displayName = `Forwarded${component}`

  return Forwarded as T
}

[...]

PR #3163, opened 2022-07-22
[labs/react] Provide explicit return type from createComponent

taylor-vann:
Related to:
https://github.com/lit/lit/issues/3188

After landing, revisit:
https://github.com/lit/lit/issues/3021
https://github.com/lit/lit/issues/2951

This PR is a type-only change (except for one location described below).

This PR exports an explicit return type `ReactWebComponent` for Typescript for downstream users. This allows users to prepare for Types without calling the `createComponent` function. An explicit return type also enables overrides for `createComponent` arguments.

A successor PR https://github.com/lit/lit/pull/2988 will create `params` object override.

++ The only change outside of types is a `null` used to diff check `refs`. However even this change is made to satisfy typing constraints on `refs`. So it's _almost_ a type change.

```ts
246      const userRef = this.props.__forwardedRef ?? null;
```


sorvell (changes requested):
Please change the description of this PR to first clearly state the problem with links to the provoking issue(s). Then discuss the changes. We also need to ensure there are tests to validate that the changes fix the referenced issues. Seems doable at least for https://github.com/lit/lit/issues/3021.

Thanks.

taylor-vann:
@sorvell updated the title to fit the PR better!

As far as providing tests to fix related issues, I feel that's a little in depth and out of scope for this PR.

I don't think this one typing change would fix an issue like https://github.com/lit/lit/issues/2951. But it would eliminate some of the ambiguity described in https://github.com/lit/lit/issues/2951.

The purpose of this PR is to harden types in the react package so we can eliminate the wrapper as a source of type ambiguity.

taylor-vann:
I've updated the description and reduced the PR to the most minimal change.
This is ready for a review review.

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,47 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+// Typecast that curries an Event type through a string to match
```
justinfagnani, line 7: Can we remove the "curries" part here? It's not consistent with my understanding of the definition of the term.

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,47 @@
...
+// a prop name to a typed event callback.
+export type EventName<T extends Event = Event> = string & {
+  __event_type: T;
+};
+
+type Events = Record<string, EventName | string>;
```
justinfagnani, line 13: For our own understanding I'd like to comment this and `EventProps`. I keep forgetting why we have both.

I'd also like to consider another name for `Events`. Is it an `EventMap` maybe?

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,47 @@
...
+    ? (e: R[K]['__event_type']) => void
+    : (e: Event) => void;
+};
+
+type ReactProps<I, E> = Omit<React.HTMLAttributes<I>, keyof E>;
+type ElementWithoutPropsOrEvents<I, E> = Omit<
+  I,
+  keyof E | keyof ReactProps<I, E>
+>;
+// Props the user is allowed to use, includes standard attributes, children,
```
justinfagnani, line 26: style: add newline

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,47 @@
...
+// ref, as well as special event and element properties.
+export type ElementProps<I, E extends Events = {}> = Partial<
+  ReactProps<I, E> & ElementWithoutPropsOrEvents<I, E> & EventProps<E>
+>;
+// Props used by this component wrapper. This is the ElementProps and the
```
justinfagnani, line 31: style: add newline

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,47 @@
...
+// special `__forwardedRef` property. Note, this ref is special because
+// it's both needed in this component to get access to the rendered element
+// and must fulfill any ref passed by the user.
+type ComponentProps<I, E extends Events = {}> = ElementProps<I, E> & {
+  __forwardedRef?: React.Ref<I>;
+};
+
+export type WrappedWebComponent<
```
justinfagnani, line 39: How about `ReactWebComponent`?

packages/labs/react/src/create-component.ts
```diff
@@ -255,17 +264,19 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
...
   }
 
-  const ForwardedComponent = React.forwardRef(
-    (props?: UserProps, ref?: React.Ref<unknown>) =>
-      createElement(
-        ReactComponent,
-        {...props, __forwardedRef: ref} as ComponentProps,
-        props?.children
-      )
+  const ForwardedComponent: WrappedWebComponent<I, E> = React.forwardRef<
```
justinfagnani, line 271: Can you add the return type to `createComponent` on line 153?

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
taylor-vann: How about `EventNameMap` since we also use `EventName` as that type?

packages/labs/react/src/create-component.ts
taylor-vann: sounds good, curry is farther than where i started, makes sense

packages/labs/react/src/create-component.ts
taylor-vann: `EventName` -> `EventNames` -> `EventNameListeners` ?

packages/labs/react/src/create-component.ts

packages/labs/react/src/create-component.ts
taylor-vann: sounds good to me!

packages/labs/react/src/create-component.ts
taylor-vann: done

packages/labs/react/src/create-component.ts
taylor-vann: done

packages/labs/react/src/create-component.ts

taylor-vann:
Tests are passing.

This PR is no longer blocked and ready for submission.

packages/labs/gen-wrapper-react/package.json
```diff
@@ -60,11 +60,13 @@
...
   },
   "dependencies": {
+    "@lit-internal/tests": "0.0.0",
     "@lit-labs/analyzer": "^0.2.0",
     "@lit-labs/gen-utils": "^0.1.0"
   },
   "devDependencies": {
-    "@lit-internal/tests": "^0.0.0"
+    "@lit-internal/tests": "^0.0.0",
+    "@types/react": "^18.0.17"
```
kevinpschaaf, line 69: I understand that this fixed the type compilation issue, but I don't understand why. The `gen-wrapper-react` source does not actually do anything with react (note, it doesn't have react as a dependency), it only generates and tests code that uses react.

The generated react wrapper has a peer dependency on `@types/react` (see [generator](https://github.com/lit/lit/blob/main/packages/labs/gen-wrapper-react/src/index.ts#L97) and [golden](https://github.com/lit/lit/blob/main/packages/labs/gen-wrapper-react/goldens/test-element-a/package.json#L15)), and the test folder into which the generated wrapper is installed depends on `@types/react` [[source](https://github.com/lit/lit/blob/main/packages/labs/gen-wrapper-react/test-output/package.json#L10)].

So it's unclear in my mind what is resolving `@types/react` down into `gen-wrapper-react`'s node_modules.  I want to understand that, otherwise it's not clear we're missing something that will hit actual users, since the direct user of the generate wrapper having a dep on `@types/react` apparently isn't sufficient.

sorvell (changes requested):
I'd like to see a test that leverages and concretely demonstrates the value of this change described in the PR description.

packages/labs/gen-wrapper-react/package.json
taylor-vann: Oh nice, i think you just unlocked something for me!

in the lit-labs/react pacakge.json there's a `types` property
```
{
...
    "types": ["react", "react-dom", "mocha"]
}
```

and when i add that to the `lit-labs/gen-wrapper-react/test-output/package.json`
```
{
...
    "types": ["react", "react-dom", "mocha"]
}
```

We get passing tests for `gen-wrapper-react` without adding `@types/react` to the gen-wrapper-react package.

justinfagnani: That confuses me. The `"types"` field is supposed to limit what `@types/` packages are included, not expand them: https://www.typescriptlang.org/tsconfig#types

taylor-vann: even more confusing it fails windows tests but passes others

taylor-vann: without the `types` array, all local tests fail on unix and windows

taylor-vann: I'm trying to discern whether we are operating within the domain of our npm workspace or if the test-output is completely separate from the workspace.

Adding the @types resolves this but AFAIU @kevinpschaaf we don't want the @types associated with `gen-wrapper-react` only the `test-output`.

This feels in the domain of ownership and npm workspaces. It makes me think that `gen-wrapper-react` is the domain and `test-output` is operating within that domain.

The `test-output` is independent and sort of an integration test for the wrapper no? But do we really know how isolated `test-output` from `gen-wrapper-react`? That's currently where I'm at trying to rationalize why adding `@types` to `gen-wrapper-react` works here.

It would make me wonder if `test-output` was entirely outside the Lit package (a real world use case) would resolve this.

taylor-vann: is it possible that we need to make `test-output` a workspace?

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -131,6 +131,18 @@ suite('createComponent', () => {
...
+  /*
+    The following test is a type-only test.
+  */
+  test('renders element with expected type', async () => {
+    type TypedComponent = ReactWebComponent<BasicElement>;
+
+    let TypedBasicElement!: TypedComponent;
+
+    // @ts-expect-error
+    <TypedBasicElement bool={"string"}></TypedBasicElement>
```
sorvell, line 143: Add a comment explaining why this should be a type error.

packages/labs/react/src/create-component.ts
```diff
@@ -132,31 +170,39 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
...
-  // Props used by this component wrapper. This is the UserProps and the
-  // special `__forwardedRef` property. Note, this ref is special because
-  // it's both needed in this component to get access to the rendered element
-  // and must fulfill any ref passed by the user.
-  type ComponentProps = UserProps & {
-    __forwardedRef?: React.Ref<unknown>;
-  };
+  // Set of properties/events which should be specially handled by the wrapper
+  // and not handled directly by React.
+  const elementClassProps = new Set(Object.keys(events ?? {}));
```
sorvell, line 177: How is this changed related to the PR?

sorvell (dismissed):

packages/labs/react/src/create-component.ts
taylor-vann: i'll revert

taylor-vann: Oh there's nothing to revert to. This is coming across as a change but it's the same as `main`.

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: done!

packages/labs/gen-wrapper-react/package.json
taylor-vann: resolved with https://github.com/lit/lit/pull/3310

packages/labs/react/src/create-component.ts
justinfagnani: I don't see this in the original file in this diff though, so it's confusing. Looks like new code. Same with the for loop below.

Do you need to merge in `main`?

taylor-vann: im not trying to change this part of the code

taylor-vann: well this PR has been sitting in limbo and had other submissions, not surprised there's side effects here

taylor-vann: removed

packages/labs/react/src/create-component.ts
```diff
@@ -121,7 +156,10 @@ type EventProps<R extends Events> = {
  * messages. Default value is inferred from the name of custom element class
  * registered via `customElements.define`.
  */
-export const createComponent = <I extends HTMLElement, E extends Events = {}>(
+export const createComponent = <
+  I extends HTMLElement,
+  E extends EventNames = {}
+>(
```
justinfagnani, line 162: We should be using `ReactWebComponent<I, E>` as the return type here now.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
taylor-vann: yessir!

taylor-vann: lol did everything but add a return type xD

taylor-vann: done!

taylor-vann (commented):
changes reviewd

justinfagnani (approved):
LGTM!

taylor-vann (commented):
changes addressed

PR #3166, opened 2022-07-23
Fix example with `hasNativeDeclarativeShadowRoots`

hsablonniere:
Hey, I just tried the examples and the ponyfill is broken because `hasNativeDeclarativeShadowRoots` is actually a function.

https://github.com/webcomponents/template-shadowroot/blob/62ca5560c757a26e3425dd4c97dc48acd2591e01/src/_implementation/feature_detect.ts#L18


AndrewJakubowicz (approved):
Thank you!

augustjk (approved):
Thank you for the fix!

I think this is good for now, though we might want to update this to be a bit more efficient like shown here https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#example-bootup-strategy

justinfagnani (approved):

PR #3175, opened 2022-07-25
[labs/testing] Pass module file urls to worker for Windows compatibility

### LINKED ISSUE #3165: [labs/testing] WTR plugin for SSR does not work in Windows

### Description

For `@lit-labs/testing`, when utilizing `ssrFixture`, the server command registered by the WTR plugin resolves the provided relative module paths before sending to the worker thread to be imported before rendering.

https://github.com/lit/lit/blob/1ac58058fbdf58633b2bb8d10649d0e11c409256/packages/labs/testing/src/lib/lit-ssr-plugin.ts#L32-L34

In windows, this creates module paths that begin with "drive:" like `c:`, causing an error like
```
Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:' 
```

We should fix it so it's always a proper file URL for all platforms.

### Reproduction

TODO

### Workaround

Manually patch the plugin code.

### Is this a regression?

### Affected versions

latest

OS: Windows

augustjk:
Fixes #3165 

In Mac/Linux, `process.cwd()` returns an absolute file path starting with `/`.
But in Windows, it returns a path starting with `C:\` which Node's import thinks is specifying some invalid protocol. Converting the absolute file paths to file URLs fixes this as it'll always begin with `file://`.

I also added the test for this package to `windows-tools` test job.

### DIFF
--- a/.changeset/four-rocks-pretend.md
+++ b/.changeset/four-rocks-pretend.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/testing': patch
+---
+
+Make resolved paths sent to worker be file urls. Fixes incompatibility with Windows filepaths.
--- a/package.json
+++ b/package.json
@@ -170,7 +170,8 @@
         "./packages/labs/gen-utils:test",
         "./packages/labs/gen-wrapper-angular:test",
         "./packages/labs/gen-wrapper-react:test",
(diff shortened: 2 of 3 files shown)


aomarks (approved):
Awesome, nice fix!

PR #3176, opened 2022-07-26
[infra] Integrate release-image into a Github Action

AndrewJakubowicz:
Last part of https://github.com/lit/lit/issues/2503

This PR adds a new Github Action "Generate Release Image" which runs on the pull_request action if the author modified a CHANGELOG.md file in the monorepo. This should only ever be the "Version Packages" PR. In the case that it isn't - you'll receive an image of your pull request description.

### Features

 - If the PR updates a CHANGELOG.md file, the action will use the PR body to render an image. The image is then uploaded as an artifact.
 - Doesn't appear in the Github Actions list on pull requests that do not touch CHANGELOG.md files.
 - `sed` is used to trim the beginning of the "Version Packages" body before the `# Releases` heading. This was tested manually using this PR description.

### Testing

Tested manually by pushing a commit that edited a CHANGELOG.md file - and then reverted that change. The action triggered and appeared in the Github Actions list for the PR.  After the CHANGELOG.md file was reverted the action disappeared again.

Also tested by copy pasting some content to the top of this PR containing the `# Releases` heading that is generated in a "Version Packages" PR and ensuring that a screenshot is correctly generated:

release 6

Risk of this change is very minimal and is scoped to an action.

### Notes

 - By using the `paths` filter, the action doesn't appear unless a changelog has been modified.


.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,47 @@
+name: Generate Release Image
+
+on:
+  pull_request:
+    paths:
+      - 'CHANGELOG.md'
```
augustjk, line 6: I think this is redundant with `'**/CHANGELOG.md'` that's already in the line below.

.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Create release image
+        run: |
+          # Store the PR body contents containing the changelog in 'release.md'
+          cat <<'EOF' > release.md
+          ${{ github.event.pull_request.body }}
+          EOF
+          # Only render the pull request content including and after the "#
+          # Releases" heading.
+          node_modules/.bin/release-image -m <(sed -n '/# Releases/,$p' release.md)
```
augustjk, line 41: Neat!

augustjk (approved):
This is awesome!

.github/workflows/release-image.yaml
AndrewJakubowicz: Great point! Thank you!

aomarks (approved):
Neat!

PR #3182, opened 2022-07-26
[labs/cli infra] Extend cli test timeout

augustjk:
When running `npm test` from the monorepo root with no cached results, all the tests running in parallel makes the cli test below take longer than the original 10s timeout. https://github.com/lit/lit/blob/41c85d4c18124f066fd46d0d50edec2028644642/packages/labs/cli/src/test/gen/react_test.ts#L33

Increasing it to 30s seems to make it pass on my machine. (I tried 20s and it still timed out)


justinfagnani:
Is there an alternate fix where we instead limit the parallelism for tests?

aomarks (approved):

augustjk:
> Is there an alternate fix where we instead limit the parallelism for tests?

We can manually set the `WIREIT_PARALLEL` env var for our test script. Default is 4 times cpu count. https://github.com/google/wireit#parallelism

Wireit actually does 4 * `os.cpus().length` which returns the number of available threads. So my quad core intel cpu returns 8 for that, so wireit by default will run 32 scripts in parallel.

aomarks:
> Wireit actually does 4 * `os.cpus().length` which returns the number of available threads. So my quad core intel cpu returns 8 for that, so wireit by default will run 32 scripts in parallel.

We should maybe bring down the wireit default. Maybe `2 *` instead of `4 *` CPUs is a better default.

AndrewJakubowicz (approved):
Nice!

augustjk:
When I run the test on my machine with `WIREIT_PARALLEL=16` (which would be the new default for wireit), that particular cli test does seem to pass with the original timeout of 10s. Though now I'm seeing some other virtualizer tests timing out 😞.

With the change to wireit to reduce the default parallel count here https://github.com/google/wireit/pull/363, should we not change the timeout here and instead update wireit when that's released? @aomarks @justinfagnani ?

augustjk:
Ah, nevermind, tried it again on my machine with `WIREIT_PARALLEL=16` and the cli test failed. I think a bit longer timeout here won't hurt.

justinfagnani (approved):

PR #3185, opened 2022-07-27
[labs/gen-wrapper-angular] Prompt to install Angular CLI

justinfagnani:
This is a rough cut of re-using the CLI installation prompt to install the Angular CLI.

I don't love making `generate()` take so many dependencies and prompt the user - just generating a file tree made it seem possible to run in non-CLI environments. That's clearly not true if the generator itself depends on something like the Angular CLI though, so it's not actually a loss there.

There's a question of where the `@angular/cli` package should be installed to that I think also somewhat applies to Lit CLI commands. Right now I made it global, but I think it possibly should be configurable. For the Lit CLI I think we should install commands near the CLI itself (global or local).


packages/labs/cli/src/lib/generate/generate.ts
```diff
@@ -14,28 +14,44 @@ import {Package} from '@lit-labs/analyzer/lib/model.js';
...
+  description: 'Generate React wrappers for LitElements',
   kind: 'reference',
   installFrom: '@lit-labs/gen-wrapper-react',
   importSpecifier: '@lit-labs/gen-wrapper-react/index.js',
 };
 
 const vueCommand: Command = {
   name: 'vue',
-  description: 'Generate Vue wrapper for a LitElement',
+  description: 'Generate Vue wrappers for a LitElements',
```
rictic, line 25: a LitElements -> a LitElement

packages/labs/cli/src/lib/generate/generate.ts
```diff
@@ -49,6 +65,7 @@ export const run = async (
   }: {packages: string[]; frameworks: string[]; outDir: string; cli: LitCli},
   console: Console
 ) => {
+  console.log('packages', packages);
```
rictic, line 68: console log

packages/labs/cli/src/lib/generate/generate.ts
```diff
@@ -93,12 +110,18 @@ export const run = async (
...
     const results = await Promise.allSettled(
       generators.map(async (generator) => {
-        // TODO(kschaaf): Add try/catches around each of these operations and
-        // throw more contextural errors
-        await writeFileTree(out, await generator.generate(options, console));
+        try {
+          await writeFileTree(out, await generator.generate(options));
+        } catch (e) {
+          console.error('Wrapper generation failed', e);
+          return;
```
rictic, line 123: Do we want to return here? That ensures that the error handling below will never run

packages/labs/cli/src/lib/install.ts
```diff
@@ -0,0 +1,89 @@
+import {spawn} from 'child_process';
```
rictic, line 1: missing license comment

packages/labs/cli/src/lib/lit-cli.ts
```diff
@@ -252,10 +252,19 @@ export class LitCli {
     const reference = maybeReference;
     let resolvedPackageLocation = this.resolveImportForReference(reference);
     if (resolvedPackageLocation === undefined) {
-      const installed = await this.installDepWithPermission(reference);
+      const installed = await installDepWithPermission({
+        description: `The command ${reference.name}`,
```
rictic, line 256: This is clearer with the command name quoted, recommend JSON.stringify(reference.name)

packages/labs/gen-wrapper-angular/src/index.ts
```diff
@@ -7,47 +7,144 @@
...
+    if (err.stderr.includes('command not found')) {
+      await installDepWithPermission({
+        description: 'The Angular CLI',
+        npmPackage: '@angular/cli',
+        global: true,
+        cwd: options.cwd,
+        stdin: options.stdin,
+        console: options.console,
+      });
+    }
```
rictic, line 83: Should we check `ng --version` again afterwards, so we can give a clear error in case the install failed?

rictic (changes requested):

rictic (changes requested):
Needs a merge / rebase

PR #3186, opened 2022-07-28
Export StaticValue from lit-html/static

dfreedm:
Class properties initialized with a literal are implicitly typed as
`any` because `StaticValue` is not exported.

### DIFF
--- a/.changeset/hip-eels-ring.md
+++ b/.changeset/hip-eels-ring.md
@@ -1,5 +1,4 @@
 ---
-'lit': patch
 'lit-html': patch
 ---
 
--- a/.changeset/hip-eels-ring.md
+++ b/.changeset/hip-eels-ring.md
@@ -1,0 +1,6 @@
+---
+'lit': patch
+'lit-html': patch
+---
+
+export StaticValue type
--- a/.changeset/hip-eels-ring.md
+++ b/.changeset/hip-eels-ring.md
@@ -3,4 +3,4 @@
 'lit-html': patch
 ---
 
-export StaticValue type
+`StaticValue` interface type is now exported.
--- a/packages/lit/src/index.all.ts
+++ b/packages/lit/src/index.all.ts
@@ -36,7 +36,6 @@ export {
   svg as staticSvg,
   unsafeStatic,
   withStatic,
-  StaticValue,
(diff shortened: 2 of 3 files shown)


packages/lit/src/index.all.ts
```diff
@@ -36,6 +36,7 @@ export {
   svg as staticSvg,
   unsafeStatic,
   withStatic,
+  StaticValue,
```
augustjk, line 39: Despite what the comment says, I wonder if `StaticValue` actually needs to be added here since this file is used just to create the JS bundle. The TS compilation step would strip it away anyway so it probably doesn't matter either way.

.changeset/hip-eels-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit': patch
+'lit-html': patch
+---
+
+export StaticValue type
```
AndrewJakubowicz, line 6: Nit: Could this be turned into a full sentence as it'll appear in the changelog.md

augustjk, line 6: Nit. This becomes a line in the changelog so it would be nice if it's a bit more verbose like
```suggestion
`StaticValue` interface type is now exported.
```

packages/lit/src/index.all.ts
AndrewJakubowicz, line 39: If I understand correctly this file is for the prebuilt [JS bundle](https://lit.dev/docs/getting-started/#use-bundles) of Lit. Not sure if exporting an interface here is required which I assume gets removed from the JS bundle, although not a blocker.

augustjk (approved):

AndrewJakubowicz (approved):

packages/lit/src/index.all.ts
kevinpschaaf, line 39: This is just for a JS bundle, so I don't think adding types here is critical, but probably doesn't hurt.

kevinpschaaf (approved):

packages/lit/src/index.all.ts
dfreedm: If it's not doing anything useful, I'll just remove it and update the changelist.

PR #3187, opened 2022-07-28
[labs/ssr] Add URL and URLSearchParams to VM module global

### LINKED ISSUE #3168: [labs/ssr] URL is missing in the dom-shim

SSR (@lit-labs/ssr)

### Description

The global browser `URL` is not available in the `vm` module when using the `renderModule` technique.

### Reproduction

Any component that used the `URL` global class from browsers.

### Workaround

I tried adding it in [`dom-shim`](renderModule) using the global `URL` from Node.js and it works!!

It would be great if it were enabled by default.

### Is this a regression?

### Affected versions

"@lit-labs/ssr": "^2.2.2"

Node version: 18.0.0
npm version: 8.6.0

augustjk:
Fixes #3168

Adds `URL` and `URLSearchParams` class to the global that's provided to the VM context when instantiating `ModuleLoader`.

### DIFF
--- a/.changeset/mean-toes-cry.md
+++ b/.changeset/mean-toes-cry.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+When using `renderModule`, `URL` and `URLSearchParams` are now available in global of the VM module context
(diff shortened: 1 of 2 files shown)


aomarks (approved):

AndrewJakubowicz (approved):
Nice!

PR #3198, opened 2022-08-04
[localize & labs/testing] Inline TS source in JS source map

### LINKED ISSUE #3107: [infra] Sources not inlined in sourcemap

Localize (@lit/localize)

### Description

The `tsconfig.json` of serveral lit packages do not use the `inlineSources: true` configuration. This is inconsistent with the other packages and make `lit-localize` original source not available for debugging.

Here is the list of packages:
- localize-tools;
- localize;
- tests.

I prepared a commit with the fix locally but I don't have the Google CLA signed yet.
It will certainly be faster if someone with accepted CLA creates the pull request.

### Reproduction

I detected the issue while using the parceljs v2 bundler.

### Workaround

None

### Is this a regression?

### Affected versions

Failing in 0.11.3

This is independent of the browser.

augustjk:
Fixes #3107 

Both `localize` and `labs/testing` packages were missing `inlineSources: true` in their tsconfig. This PR fixes that.

Also added `declarationMap: true` for `localize`.


aomarks (approved):
Nice!

justinfagnani (approved):

PR #3199, opened 2022-08-05
In DEV_MODE, render a warning instead of rendering a template's host in the template.

rictic:
Most commonly this would happen when rendering `${this}` in a LitElement's template, which has the counterintuitive behavior of removing the element from the DOM, because when rendering the element's template we attach it into its own shadow root, which removes it from the DOM, causing it simply disappear. This is especially problematic with a fast HMR system.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1460,6 +1460,14 @@ class ChildPart implements Disconnectable {
           `[probable mistake: rendered a template's host in itself ` +
             `(commonly caused by writing \${this} in a template]`
         );
+        console.warn(
+          `Attempted to render the template host`,
+          value,
+          `inside itself. This is almost always a mistake, and in dev mode `,
+          `we render some warning text. In production however, we'll `,
+          `render it, which will usually result in an error, and sometimes `,
+          `in the element disappearing from the DOM.`
+        );
         return;
       }
       this._commitNode(value as Node);
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1455,6 +1455,13 @@ class ChildPart implements Disconnectable {
     } else if ((value as TemplateResult)['_$litType$'] !== undefined) {
       this._commitTemplateResult(value as TemplateResult);
     } else if ((value as Node).nodeType !== undefined) {
+      if (DEV_MODE && this.options?.host === value) {
+        this._commitText(
+          `[probable mistake: rendered a template's host in itself ` +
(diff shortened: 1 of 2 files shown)


justinfagnani:
Whoa, this is an unexpected feature to need, but with HMR I get it!

But doesn't this potentially apply to any element? What if you wanted to render the id of a child, like `${someEl.id}`?

Maybe the solution should be to not render any element that already has a parent?

I'm also a little surprised that this doesn't just cause an error, since you shouldn't be able to add an element to a descendent as that would cause a cycle in the DOM.

rictic:
I think it's legit and not that uncommon to reparent a node in general with Lit, but it's almost definitely a mistake to render the host assuming that the host contains the template, as is typical.

justinfagnani:
> I think it's legit and not that uncommon to reparent a node in general with Lit

I don't think I've ever seen that. I've definitely seen new, non-parented node be inserted with Lit, but not nodes with a parent already.

The main case I could imagine is if the parent is a DocumentFragment like from cloning a template, but since we have the `templateContent` directive, I haven't seen that either.

rictic:
One use case:

```
render() {
  if (this.noChrome) {
    return html`${this.contentEl}`;
  }
  return html`
    <header>...</header>
    <main>${this.contentEl}</main>
    <footer>...</footer>
}
```

Yeah it's just moving the element around inside the same shadow root, but that's legit! Or for a slightly harder to detect example, you could imagine passing contentEl to one of two kinds of rendering element.

It's somewhat weird, but there are cases where this is a pretty good design, and if we block it here we're effectively banning it.

justinfagnani:
otoh, a pattern I _have_ seen not-uncommonly (especially in server-rendered content) is light-DOM-as-data:

```ts
render() {
  const title = this.querySelector('x-title');
  return html`<h1>${title.textContent}`</h1>`;
}
```

So how do we keep from destructively updating the DOM when the expression is just `${title}` while typing?

I'm wondering if we need an HMR mode with options for these things...

packages/lit-html/src/lit-html.ts
```diff
@@ -1455,6 +1455,13 @@ class ChildPart implements Disconnectable {
     } else if ((value as TemplateResult)['_$litType$'] !== undefined) {
       this._commitTemplateResult(value as TemplateResult);
     } else if ((value as Node).nodeType !== undefined) {
+      if (DEV_MODE && this.options?.host === value) {
+        this._commitText(
```
justinfagnani, line 1459: Log a warning as well?

rictic: Done

justinfagnani (approved):

Issue #3202, opened 2022-08-06
[labs/ssr] ModuleLoader is not able to handle file urls in Windows


augustjk:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

When calling `renderModule()` in Windows, an error like the following shows up.

```
Error: ENOCENT: no such file or directory, open 'C:\C:\path\to\module.js'
```

I believe the culprit is here
https://github.com/lit/lit/blob/0162fbad61826ba0ff4188135ca4ab778762c4d7/packages/labs/ssr/src/lib/module-loader.ts#L134

the `.pathname` of a windows file url is something like `/C:/path/to/module.js`. This is then passed to `fs.readFile()` which adds an extra `C:\` as it tries to read the file.

### Reproduction

Running the `demo:vm-modules` script in the SSR package and trying to load `localhost:8080` in the browser in Windows reproduces this issue.

### Workaround

Manually edit source to remove the extra `/` from the `pathname` of the file url.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

latest

### Browser/OS/Node environment

OS: Windows 11
Node: 16.16.0

PR #3203, opened 2022-08-06
[labs/eleventy-plugin-lit] Use file URLs for resolved module paths

### LINKED ISSUE #3059: [labs/eleventy-plugin-lit] in worker mode fails to resolve component modules on Windows
# Description

When building an 11ty site with lit-ssr and the eleventy-plugin-lit plugin configured in worker mode, the eleventy commands fail with the error `Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'`.

I can open a PR to resolve; my intent would be to do the following (tested changes locally and continues to function):
1. Import the 'os' built-in library
2. Check if the platform is win32, and prepend `file://` before each `resolvedComponentModules` entry at https://github.com/lit/lit/blob/main/packages/labs/eleventy-plugin-lit/src/index.ts#L211, e.g.:
    ```ts
    const resolvedComponentModules = componentModules.map((module) => {
      let result = path.resolve(process.cwd(), module);
      if (os.platform() === 'win32') {
        result = `file://${result}`;
      }
      return result;
    });
    ```

# Steps to Reproduce

1. Write this code

.eleventy.js:
```js
...
eleventyConfig.addPlugin(litPlugin, {
    mode: "worker",
    componentModules: ["src/components/app.js"],
});
...
```

4. See this output...

```
Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:' (via Error)
[11ty]
[11ty] Original error stack trace: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
[11ty]     at new NodeError (node:internal/errors:377:5)
[11ty]     at throwIfUnsupportedURLScheme (node:internal/modules/esm/resolve:1076:11)
[11ty]     at defaultResolve (node:internal/modules/esm/resolve:1156:3)
[11ty]     at ESMLoader.resolve (node:internal/modules/esm/loader:605:30)
[11ty]     at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
[11ty]     at ESMLoader.import (node:internal/modules/esm/loader:404:22)
[11ty]     at importModuleDynamically (node:internal/modules/esm/translators:106:35)
[11ty]     at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14)
[11ty]     at file:///C:/{omittedpath}/node_modules/@lit-labs/eleventy-plugin-lit/worker/worker.js:18:59
[11ty]     at Array.map (<anonymous>)
```

## Live Reproduction Link

None; if a reproduction project would be helpful let me know.

## Expected Results

Eleventy continues to function.

## Actual Results

Eleventy site fails to build/serve/etc.

## Browsers Affected

Not applicable, this is a NodeJS issue.

Using NodeJS v18.2.0.

# Workaround
Use WSL to develop an 11ty site with this plugin on Windows.

augustjk:
Fixes #3059 

Due to module file paths having `C:\` for Windows, the imports fail due to unrecognized protocol. Converting them to file URLs fixes that.

Note:
This makes eleventy-plugin-lit now work in `worker` (default) mode for Windows users.
In `vm` mode, it still breaks due to an issue with `ModuleLoader` noted here https://github.com/lit/lit/issues/3202

I was also unable run the eleventy-plugin-lit package test suite in Windows as the testing rig has issues somewhere with the filesystem symlinking, writing, reading.

I did, however, manually test it by running the example eleventy project on a Windows machine and confirmed `worker` mode runs.

### DIFF
--- a/.changeset/shy-wasps-exercise.md
+++ b/.changeset/shy-wasps-exercise.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': patch
+---
+
+Use file urls for resolved module paths for Windows compatibility
--- a/packages/labs/eleventy-plugin-lit/src/index.ts
+++ b/packages/labs/eleventy-plugin-lit/src/index.ts
@@ -5,6 +5,7 @@
  */
 
 import * as path from 'path';
+import {pathToFileURL} from 'url';
 import {Worker} from 'worker_threads';
 import type {Message} from './worker/types.js';
 
@@ -208,8 +209,8 @@ module.exports = {
       return;
     }
 
-    const resolvedComponentModules = componentModules.map((module) =>
-      path.resolve(process.cwd(), module)
+    const resolvedComponentModules = componentModules.map(
+      (module) => pathToFileURL(path.resolve(process.cwd(), module)).href
     );
 
     switch (mode) {


aomarks (approved):
Nice! Maybe file an issue about testing in windows if there isn't one already?

augustjk:
> Maybe file an issue about testing in windows if there isn't one already?

Thanks! Filed https://github.com/lit/lit/issues/3205

PR #3214, opened 2022-08-12
Also re-run release image generator on pushes

aomarks:
This way it will run not only when a release PR is opened, but also when it's updated.

### DIFF
--- a/.github/workflows/release-image.yaml
+++ b/.github/workflows/release-image.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths:
       - '**/CHANGELOG.md'
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


augustjk (approved):

AndrewJakubowicz (approved):
Nice thanks!

AndrewJakubowicz:
Sorry - completely slipped my mind. This action uses `github.event.pull_request.body` so I don't think it will update on push - and maybe will fail?

aomarks:
> Sorry - completely slipped my mind. This action uses `github.event.pull_request.body` so I don't think it will update on push - and maybe will fail?

I guess we'll find out!

PR #3217, opened 2022-08-15
[labs/react] Filter __forwardedRef in prod build

### LINKED ISSUE #3211: [labs/react] __forwardedRef is being minified in prod builds

React (@lit-labs/react)

### Description

`__forwardedRef` matches our our Terser property name mangling config, so it's being minified. This is noticeable because the prod tests trigger a React warning that `_$Gl` isn't a valid attribute name, which is caused by us adding `__forwardedRef` to props here: https://github.com/lit/lit/blob/main/packages/labs/react/src/create-component.ts#L250

This should mean that user refs are broken, but we have a ref test, so we need to check our coverage here.

We can prevent the mangling either by adding `'__forwardedRef'` to our global reserved property names, or by setting Terser's `keep_quoted` option to `true`.

### Reproduction

Run the react tests

### Workaround

none

### Is this a regression?

### Affected versions

all

all

justinfagnani:
Fixes #3211

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -145,7 +145,7 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
   // it's both needed in this component to get access to the rendered element
   // and must fulfill any ref passed by the user.
   type ComponentProps = UserProps & {
-    __forwardedRef?: React.Ref<unknown>;
+    __forwardedRef: React.ForwardedRef<unknown>;
   };
 
   class ReactComponent extends Component<ComponentProps> {
@@ -200,10 +200,10 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
      *
      */
     override render() {
-      // Since refs only get fulfilled once, pass a new one if the user's
-      // ref changed. This allows refs to be fulfilled as expected, going from
+      const {__forwardedRef: userRef, ...userProps} = this.props;
+      // Since refs only get fulfilled once, pass a new one if the user's ref
+      // changed. This allows refs to be fulfilled as expected, going from
       // having a value to null.
-      const userRef = this.props.__forwardedRef as React.Ref<unknown>;
       if (this._ref === undefined || this._userRef !== userRef) {
         this._ref = (value: I | null) => {
           if (this._element === null) {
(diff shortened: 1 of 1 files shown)


justinfagnani:
Note that there's no change to tests, but you can manually verify that there's no warning in prod test logs anymore.

packages/labs/react/src/create-component.ts
```diff
@@ -145,7 +145,7 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
   // it's both needed in this component to get access to the rendered element
   // and must fulfill any ref passed by the user.
   type ComponentProps = UserProps & {
-    __forwardedRef?: React.Ref<unknown>;
+    __forwardedRef: React.ForwardedRef<unknown>;
```
taylor-vann, line 148: Nice!

packages/labs/react/src/create-component.ts
```diff
@@ -215,17 +215,15 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
...
+      // Save element props while iterating to avoid the need to iterate again
+      // when setting properties.
+      this._elementProps = {};
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const props: any = {ref: this._ref};
-      // Note, save element props while iterating to avoid the need to
-      // iterate again when setting properties.
-      this._elementProps = {};
-      for (const [k, v] of Object.entries(this.props)) {
-        if (k === '__forwardedRef') continue;
```
taylor-vann, line 227: This might regress a previously solved issue
https://github.com/lit/lit/issues/2593

packages/labs/react/src/create-component.ts
```diff
@@ -243,12 +241,12 @@ export const createComponent = <I extends HTMLElement, E extends Events = {}>(
     }
   }
 
-  const ForwardedComponent = React.forwardRef(
-    (props?: UserProps, ref?: React.Ref<unknown>) =>
+  const ForwardedComponent = React.forwardRef<unknown, UserProps>(
+    (props, __forwardedRef) =>
```
taylor-vann, line 245: nice!

taylor-vann:
I like the type changes too!

One question, does this pass `__forwaredRef` to the element? I believe the main reason we skip forwardedRef was to avoid https://github.com/lit/lit/issues/2593. It would add something like `<custom-element $1>` to an element.

packages/labs/react/src/create-component.ts
taylor-vann: is `__forwardedRef` always defined? A user doesn't necessarily _have_ to pass a reference, no?

packages/labs/react/src/create-component.ts
justinfagnani: It's pulled out in the destructuring above in a way that survives property renaming. There's also a test for that issue that's not failing.

packages/labs/react/src/create-component.ts
justinfagnani: The typings for `forwardRef()` and `ForwardedRef` show that it can be `null`, but not `undefined`.

taylor-vann (approved):
As long as https://github.com/lit/lit/issues/2593 is accounted for in this change, I see no problems!

packages/labs/react/src/create-component.ts
taylor-vann: awesome!

packages/labs/react/src/create-component.ts
taylor-vann: wonderful

sorvell (approved):
Looks good.

taylor-vann:
patched into https://github.com/lit/lit/pull/3409 and merged

PR #3223, opened 2022-08-18
[lit-html] Use existing `document` in Node build

### LINKED ISSUE #3216: createComment is not a function error - lit and lit-html 2.3.0

### Description

using vite and vitest error with latest version of lit
Screenshot 2022-08-15 at 11 11 09 AM

### Reproduction

Please check stackblitz URL 

https://stackblitz.com/edit/vitejs-vite-omm8ny

### Workaround

downgrade to `lit@2.2.8` and `lit-html@2.2.7` version 

### Is this a regression?

### Affected versions

2.3.0

Chrome 103
MacOSX
Node 16

augustjk:
In Node builds, use the existing `document` if found.

This should allow node builds running in an environment with DOM emulation loaded to still function, like in a node test runner.

Fixes #3216

### DIFF
--- a/.changeset/selfish-lobsters-cheat.md
+++ b/.changeset/selfish-lobsters-cheat.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+Use existing `document` in Node build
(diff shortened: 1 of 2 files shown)


aomarks (approved):
Nice!

PR #3225, opened 2022-08-18
[cli] Prepare for CLI 0.1.0 release

aomarks:
- Remove `private` and add a changeset, so that the next deploy should release `0.1.0` of everything (because they are all currently `0.0.1`).
- Add disclaimer to README about this being an early release.
- Document the `labs gen` command and its flags in the README.
- Mention that `vue` is also supported in the `help` command text.


augustjk (approved):
LGTM

justinfagnani (approved):
Thanks!

justinfagnani (approved):
Thanks!

PR #3227, opened 2022-08-18
[cli] Prepare vue-utils for release, remove test prod deps, add publishConfig.access

aomarks:
- I forgot `@lit-labs/vue-utils` in https://github.com/lit/lit/pull/3225, so that's now `private` and has a changeset so that the next publish will release it as `0.1.0`.
- Removed a couple of prod `dependencies` on `@lit-internal/tests`, which should have been `devDependencies`.
- Add missing `publishConfig.access` to new packages `package.json` files.

### DIFF
--- a/.changeset/tidy-lobsters-press.md
+++ b/.changeset/tidy-lobsters-press.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/vue-utils': minor
+---
+
+Initial release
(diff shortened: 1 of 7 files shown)


augustjk:
Since we're fixing things.. I think all the to-be-released packages also need
```json
  "publishConfig": {
    "access": "public"
  }
```
in their `package.json` to be published under the scoped name.

aomarks:
> Since we're fixing things.. I think all the to-be-released packages also need
> 
> ```json
>   "publishConfig": {
>     "access": "public"
>   }
> ```
> 
> in their `package.json` to be published under the scoped name.

Good catch! Done.

augustjk (approved):
🚀

PR #3232, opened 2022-08-19
Typo fix in @labs/cli README.md

thomasverleye:
`localize` got mixed up with `labs` on the paragraph about the `labs gen` usage.


aomarks (approved):
Oops, thank you!

PR #3252, opened 2022-08-29
fix: ts issue with `moduleResolution: NodeNext`

Mister-Hope:


justinfagnani:
Thanks! Looks good, but we need a changeset file. Run `npm run changeset`, choose the lit package, and write a description of the chage.

justinfagnani (commented):
_deleted_

justinfagnani (changes requested):
Please add a changeset

augustjk:
Thanks for original contribution. This PR is no longer needed with #3320 merged, as we needed the types to point to those in the `development` directory instead of the package root.

PR #3255, opened 2022-08-30
[infra] Upgrade changesets and remove Prettier version override

aomarks:
Previously, weirdness would occur on release because of differences in the Prettier formatting for our CHANGELOG files between our auto-formatter and the formatting that Changesets itself tries to do. This was because Changesets had an old version of Prettier, but our repo had a newer version.

We had a workaround for this where we deleted Changesets' version of Prettier, so that ours would be prefered.

Now that https://github.com/changesets/changesets/pull/905 has landed (fixing https://github.com/changesets/changesets/issues/616), Changesets will prefer our version of Prettier. Plus the version they depend on is now up to date. So we can remove the workaround.


augustjk (approved):
Awesome, thanks for taking care of this!

justinfagnani (approved):

PR #3261, opened 2022-09-01
[lit-labs/observers] Enable tests on intersection and performance controller

AndrewJakubowicz:
### Context

This PR is to check if the Safari testing issue has resolved since the code was written months ago, and re-enables testing the Intersection and Performance controller.

This is a test only change that has been validated manually. Hopefully Github Actions Safari tests pass.

### Features

 - Add `(window.IntersectionObserver ? suite : suite.skip)` and `((await canTest()) ? suite : suite.skip)` to enable Intersection and Performance controller tests.
 - The post commit hook forced a format on the `IntersectionController` test file. No logic changes were added except enabling the test.


AndrewJakubowicz:
I don't exactly understand how the caching works. Initially the Safari tests failed in an expected way. After re-running failing tests it looks like the previously failing tests passed - however everything was cached and resolved in a matter of seconds.

I expected https://github.com/lit/lit/runs/8163567096?check_suite_focus=true#step:7:91 to fail.

AndrewJakubowicz:
The Safari tests via Sauce are extremely flakey and unworkable with this minor change.

PR #3266, opened 2022-09-03
chore: remove dom5 copy

43081j:
we now have a `@parse5/tools` package which provides the same functionality dom5 and various other helpers previously had.

this is just a wip, trying to move lit over to using it


43081j:
turns out this depends on inikulin/parse5#666

i guess every consumer of parse5@7 so far just had commonjs support (ran through a bundler).

packages/labs/ssr/src/lib/render-lit-html.ts
```diff
@@ -44,19 +44,15 @@ import {
...
 import {escapeHtml} from './util/escape-html.js';
 
-import {
-  traverse,
-  parseFragment,
-  isCommentNode,
-  isElement,
-} from './util/parse5-utils.js';
+import {parseFragment} from 'parse5';
+import {isElementNode, isCommentNode, traverse} from '@parse5/tools';
```
justinfagnani, line 48: Is there a reason to use these over the TreeAdapter methods?

43081j: It's just for simplicity really. So you can import the full set of helpers from one place. 

The tree adapter doesn't have every type guard function. In the utils package, I re export the ones it has _plus_ the extras it doesn't have. So you get them from a single place rather than two

43081j:
> First-time contributors need a maintainer to approve running workflows

btw @justinfagnani is there any way to make github realise im not a first-time contributor? so the CI runs don't need approval. maybe its just a poorly written message by github and _all_ non-org members require approval (first time or not)?

43081j:
something's up with package exports resolution of `.`. 

i'll try investigate in the next couple of days. if it turns out to be enhanced-resolve's problem, i'll pr over there.

43081j:
Looks like tests pass but we got unlucky with a network error on the last ci run 🙈

If I get a good run, I still need to publish the parse5 tools under a stable version too before we can merge any of this

aomarks:
> Looks like tests pass but we got unlucky with a network error on the last ci run 🙈
> 
> If I get a good run, I still need to publish the parse5 tools under a stable version too before we can merge any of this

Looks good. Just need to add a changeset entry to get the lint test to pass. Not sure why the benchmarks are failing.

aomarks (approved):

justinfagnani (approved):
Thanks @43081j !!

PR #3283, opened 2022-09-09
Task: Do not reset task value or error on pending

### LINKED ISSUE #2945: [labs/task] Explore task taking a task as an input
## Description

Chaining could occur ergonomically if a task could take another task as an argument.
Another possibility is putting tasks inside other tasks, similar to hooks. As in have a task execute inside a task.

Challenges:
- Awaiting a promise for success

e111077:
Fixes #2945

Task will no longer reset its value or error on pending. This allows us to start chaining tasks e.g.

```ts
const a = new Task(this, async ([url]) => await fetch(url), () => [this.url]);
const b = new Task(this, async ([value]) => {/* This is not thrashed */}, () => [a.value]);
```

### DIFF
--- a/.changeset/friendly-squids-perform.md
+++ b/.changeset/friendly-squids-perform.md
@@ -2,4 +2,4 @@
 '@lit-labs/task': major
 ---
 
-Do not reset value on status change
+Do not reset value or error on status change
--- a/.changeset/friendly-squids-perform.md
+++ b/.changeset/friendly-squids-perform.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': major
+---
+
+Do not reset value on status change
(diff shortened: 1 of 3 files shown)


e111077:
IE failures seem unrelated

.changeset/friendly-squids-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': major
+---
+
+Do not reset value on status change
```
justinfagnani, line 5: add mention of `error`

justinfagnani (approved):

.changeset/friendly-squids-perform.md
e111077: Can I edit that file directly or do I need to re-run the changeset command?

justinfagnani: You can just edit the file. The only thing that the changeset command does is generate the file with a nice wizard UI.

PR #3287, opened 2022-09-13
Task add onComplete and onError

e111077:
Adds `onComplete` and `onError` callbacks to Task.

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -44,6 +44,8 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
   task: TaskFunction<T, R>;
   args?: ArgsFunction<T>;
   autoRun?: boolean;
+  onComplete?: (value: R) => unknown;
+  onError?: (error: unknown) => unknown;
 }
 
 // TODO(sorvell): Some issues:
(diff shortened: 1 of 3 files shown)


e111077:
Ping on this one. Would like to get it merged before talk

packages/labs/task/src/task.ts
```diff
@@ -215,9 +221,11 @@ export class Task<
         if (error === undefined) {
           this.status = TaskStatus.COMPLETE;
           this._resolveTaskComplete(result as R);
+          this._onComplete?.(result as R);
```
justinfagnani, line 224: If we ever wanted onComplete to be able to change completion status by returning something, we'd want to run this first.

justinfagnani, line 224: I think this needs to be in a try/catch block to keep from messing with task state if it throws. Same for onError.

justinfagnani (approved):

packages/labs/task/src/task.ts
e111077: what should we do on catch? It's already too late to report onError, no?

e111077: Something like this?

```ts
if (error === undefined) {
  try {
    this._onComplete?.(result as R);
  } catch {
    // Ignore user errors from onComplete.
  }
  this.status = TaskStatus.COMPLETE;
  this._resolveTaskComplete(result as R);
} else {
  try {
    this._onError?.(error);
  } catch {
    // Ignore user errors from onError.
  }
  this.status = TaskStatus.ERROR;
  this._rejectTaskComplete(error);
}
this._value = result as R;
this._error = error;
 ```

justinfagnani: exactly

Issue #3147, opened 2022-07-19
[lit-labs/observers] ResizeController cannot easily observe a target other than host


sorvell:
### Which package(s) are affected?

Observers (@lit-labs/observers)

### Description

ResizeController accepts a target element to observe. It observes the given target when the host is connected and/or when the controller is added to the host. Unfortunately, the target is set *after* this, so if the host is already connected, the target doesn't get observed.

### Reproduction

https://lit.dev/playground/#gist=c79371cc956785c4cb5ed5a99cb93cbf

### Workaround

 The user must manually call observe to workaround this issue. This is demonstrated in the repro.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

all

### Browser/OS/Node environment

All

AndrewJakubowicz:
Fixed in https://github.com/lit/lit/pull/3293

PR #3293, opened 2022-09-21
[labs/observers] Fix controllers not observing changes if initialized after host connected

AndrewJakubowicz:
Issue: https://github.com/lit/lit/issues/3147

### Context

When `host.addController(this)` is called on a ReactiveElement that is already connected, the controllers `hostConnected` method is called synchronously. Because in each exported controller `addController` is called before a `_target` is set, the following `hostConnected` logic is executed with an undefined `_target`:

```
    if (this._target) {
      this.observe(this._target);
    }
```

This problem reproduces if the observer is initialized in `firstUpdated`.

### Fix

Move the `addController` call to the end of the constructor after all the initial variables have been initialized.

### Test plan

Tested by adding a unit test for each controller - which failed prior to the controller being updated.

### DIFF
--- a/.changeset/modern-turtles-judge.md
+++ b/.changeset/modern-turtles-judge.md
@@ -2,4 +2,4 @@
 '@lit-labs/observers': patch
 ---
 
-Fix exported controllers not observing changes if initialized after the host has connected.
+Fix controllers not observing changes to target element if initialized after the host has connected.
--- a/.changeset/modern-turtles-judge.md
+++ b/.changeset/modern-turtles-judge.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/observers': patch
+---
+
+Fix exported controllers not observing changes if initialized after the host has connected.
--- a/packages/labs/observers/src/test/intersection_controller_test.ts
+++ b/packages/labs/observers/src/test/intersection_controller_test.ts
@@ -399,4 +399,38 @@ const canTest = () => {
     await intersectionComplete();
     assert.isTrue(el.observerValue);
   });
+
+  test('can observe changes when initialized after host connected', async () => {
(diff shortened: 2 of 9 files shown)


packages/labs/observers/src/test/intersection_controller_test.ts
```diff
@@ -399,4 +399,38 @@ const canTest = () => {
     await intersectionComplete();
     assert.isTrue(el.observerValue);
   });
+
+  test('can observe changes when initialized after host connected', async () => {
```
sorvell, line 403: This test is fine, but the key test we need to add is that you can pass in a different `target` and verify that it correctly observes that target. (I think that test is needed for the other observers as well.)

sorvell (changes requested):

packages/labs/observers/src/test/intersection_controller_test.ts
AndrewJakubowicz: The `host` test was slightly covering this case, as `host` is assigned to `target` when  no target is passed in. For completeness and future proofing I've added an explicit target test as well to each controller that accepts a `target`.

Thank you!

.changeset/modern-turtles-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/observers': patch
+---
+
+Fix exported controllers not observing changes if initialized after the host has connected.
```
sorvell, line 5:
```suggestion
Fix controllers not observing changes to target element if initialized after the host has connected.
```

sorvell (approved):
Just one micro suggestion for the patch note, but LGTM. Thanks!

AndrewJakubowicz:
Done! Thank you!

Issue #2350, opened 2021-12-09
[@lit-labs/observers] Consider some refactoring


sorvell:
There are a few potential refactoring improvements to consider:
* Refine the types associated with the callback and value. These are currently just marked as unknown but they could be handled with a generic. Working this out with the tests will take a bit of refinement.
* The 4 observer controllers could share a base class. They are similar but each have some differences which need to be carefully considered. There's definitely an opportunity for more code sharing, however.

PR #3294, opened 2022-09-22
[labs/observers] Improve controllers value type

AndrewJakubowicz:
Issue: https://github.com/lit/lit/issues/2350

### Background

The `value` property was typed as unknown.

```
const resizeController = new ResizeController(host, {callback: () => true});
resizeController.value // Type is unknown.
```

After this change:

```
const resizeController = new ResizeController(host, {callback: () => true});
resizeController.value // Type is `boolean | undefined`.
```

Impact is that using `value` will no longer require unsafe type assertions. Instead the type of `value` is now always `T | undefined`, where `T` is the return type of the user provided `callback`.

### Changes

  - The generic type does not need to be provided and is unknown by default (for backwards compatibility).
  - [**BREAKING**] The callback `() => true` is no longer defined by default. Instead the callback is `undefined` unless provided.
  - Type checking between the return type of the callback and the `value` property now works.

### Test plan

Tested with type-only unit tests with no runtime behavior.

### DIFF
--- a/packages/labs/observers/src/intersection_controller.ts
+++ b/packages/labs/observers/src/intersection_controller.ts
@@ -11,14 +11,14 @@ import {
 /**
  * The callback function for a IntersectionController.
  */
-export type IntersectionValueCallback = (
+export type IntersectionValueCallback<T = unknown> = (
   ...args: Parameters<IntersectionObserverCallback>
-) => unknown;
+) => T;
 
 /**
  * The config options for a IntersectionController.
  */
-export interface IntersectionControllerConfig {
+export interface IntersectionControllerConfig<T = unknown> {
   /**
    * Configuration object for the IntersectionObserver.
    */
@@ -35,7 +35,7 @@ export interface IntersectionControllerConfig {
    * The callback used to process detected changes into a value stored
    * in the controller's `value` property.
    */
-  callback?: IntersectionValueCallback;
+  callback?: IntersectionValueCallback<T>;
   /**
    * An IntersectionObserver reports the initial intersection state
    * when observe is called. Setting this flag to true skips processing this
@@ -60,7 +60,7 @@ export interface IntersectionControllerConfig {
  * used to process the result into a value which is stored on the controller.
  * The controller's `value` is usable during the host's update cycle.
  */
-export class IntersectionController implements ReactiveController {
+export class IntersectionController<T = unknown> implements ReactiveController {
   private _host: ReactiveControllerHost;
   private _target: Element | null;
   private _observer!: IntersectionObserver;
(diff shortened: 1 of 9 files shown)


sorvell (commented):
Seems good. Makes sense to do this also for the other controllers, right?

packages/labs/observers/src/intersection_controller.ts
```diff
@@ -77,22 +77,22 @@ export class IntersectionController implements ReactiveController {
...
    * function.
    */
-  value?: unknown;
+  value?: T;
   /**
    * Function that returns a value processed from the observer's changes.
    * The result is stored in the `value` property.
    */
-  callback: IntersectionValueCallback = () => true;
+  callback?: IntersectionValueCallback<T>;
```
sorvell, line 85: Is removing the default necessary? It's dubious that it's needed, but it is a breaking change and maybe we shouldn't do it unless we have to?

packages/labs/observers/src/test/intersection_controller_test.ts
```diff
@@ -468,4 +475,38 @@ const canTest = () => {
     await intersectionComplete();
     assert.isTrue(el.observerValue);
   });
+
+  test.skip('IntersectionController<T> type-checks', async () => {
```
sorvell, line 479: Why are these type tests skipped?

sorvell (changes requested):

packages/labs/observers/src/test/intersection_controller_test.ts
AndrewJakubowicz: I skipped them as they are type only tests which are tested via the TS compiler successfully building this file. However I can run them as well, although there are no runtime assertions.

packages/labs/observers/src/intersection_controller.ts
AndrewJakubowicz: It's only a breaking change in the rare situation where someone inspected the code and relied on the default `true | undefined` value (which explicitly typed as `unknown`).

Rationale for removing it was that  it technically defines the default generic type `T` to `T = true` instead of `T = unknown`.
The default callback also seemed like it was for test convenience and I think it might be surprising to find out that the default `value` has  type `true | undefined`, instead of `undefined` .

I may be missing the original reason for the default callback. Was there a benefit to having `value` become `true` after the first observed change?

AndrewJakubowicz: I think the only way to keep the default callback `() => true` is to cast it to `any` thus making type checker happy. Otherwise it's unhappy with the concrete type `T`.

packages/labs/observers/src/test/intersection_controller_test.ts
AndrewJakubowicz: Fixed

packages/labs/observers/src/intersection_controller.ts
AndrewJakubowicz: Spoke offline - This was not an intended API. Having the callback `undefined` makes more sense as a default and the TS type system has highlighted this.

sorvell (approved):
LGTM, thanks!

Issue #3308, opened 2022-09-26
lit-html package has `@types/trusted-types` in dependencies


jorgecasar:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

lit-html package.json has `@types/trusted-types` in dependencies instead of devDependencies

### Reproduction

https://github.com/lit/lit/blob/main/packages/lit-html/package.json#L350

### Workaround

N/A

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

2.3.1

### Browser/OS/Node environment

N/A

augustjk:
I believe this is intentional to avoid TS users hitting errors like this https://github.com/lit/lit/issues/1780.

TypeScript has yet to add support for trusted types: https://github.com/microsoft/TypeScript/issues/30024.

kevinpschaaf:
Some trusted types types show up in the lit-html public interface, and thus need to be dependencies, not dev dependencies. Closing as WAI.

jogibear9988:
@kevinpschaaf but how should the packages be automatically loaded without building or bundler? I load components in my designer (https://github.com/node-projects/web-component-designer) directly from unpkg and parse there dependencies.
How should I know this is a dependency I can ignore?

justinfagnani:
@jogibear9988 Does it matter much? None of our modules load this file, it's just typings. If you follow dependencies by imports, you'll never reach it.

However if you're looking at our type declarations, you _do_ need it, otherwise the declaration won't check and the types that include trusted types will broaden to `any`.

jogibear9988:
@justinfagnani how should I follow the dependecies? I don't parse the javascript files. I load all dependecies (and sub dependencies and so on from all package.jsons). Then I create a import map for all of them (I use the importmap polyfill). And then I load the scripts.
Maybe it does not matter so much...

PR #3309, opened 2022-09-26
[lit-html] move @types/trusted-types to devDep

jorgecasar:
Fix: #3308


augustjk:
Working as intended. See https://github.com/lit/lit/issues/3308#issuecomment-1275297241

PR #3310, opened 2022-09-27
[@labs/gen-wrapper-react] TestOutput links to monorepo for dependencies

taylor-vann:
This PR links `test-output` to the monorepo for the following dependencies:
- react
- react types
- lit
- lit-element
- lit-html

This avoids version conflicts during tests where we emulate a developer using `gen-wrapper-react`.


justinfagnani:
Failures look like flakiness. I'll re-run

packages/labs/gen-wrapper-react/src/index.ts
```diff
@@ -94,8 +94,8 @@ const packageJsonTemplate = (
       },
       peerDependencies: {
         // TODO(kschaaf): make react version(s) configurable?
-        react: '^17 || ^18',
-        '@types/react': '^17 || ^18',
```
kevinpschaaf, line 98: Can these go back to 17 || 18 now? Or does it break if we do that? Seems best for the wrapper to support wider versions if possible.

kevinpschaaf (approved):
Changes LGTM, just one question about the versions.

packages/labs/gen-wrapper-react/src/index.ts
taylor-vann: let's try it out!

taylor-vann: success!

PR #3318, opened 2022-09-28
[lit-html] Add `isServer` environment checker module

### LINKED ISSUE #3158: [labs/ssr]  A way for client code to check if running in server (isSsr)

Lit Core (lit / lit-html / lit-element / reactive-element), SSR (@lit-labs/ssr)

### Description

It would be useful for `lit` to provide a way to detect if code is running in the server like `isSsr` so it can be utilized when authoring components to intentionally skip, or provide alternatives for, code that utilizes browser DOM APIs or otherwise are incompatible with SSR.

This is a better alternative to adding no-op DOM shims to alleviate errors during SSR.

Idea for implementation would be to utilize conditional exports to provide a module that will only be available/truthy in node.

cc @aomarks 

`if (typeof process !== 'undefined')`, `if(typeof window === 'undefined')` 😬 , optional chaining, or relying on no-op DOM shims.

augustjk:
Fixes #3158 

Adds a new `is-server` module to `lit-html` package, with a boolean const export of `isServer`.

The value will be `true` in the node build, and `false` in the development and default (browser) builds.

This is also re-exported from the `lit` package.


packages/lit-html/src/test/is-server_test.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {isServer} from 'lit-html/is-server.js';
+import {assert} from '@esm-bundle/chai';
+
+suite('is-server', () => {
+  test('isServer is falsey', () => {
+    assert.notOk(isServer);
```
aomarks, line 12: suggest `assert.equal(isServer, true)` so that we verify it's a `bool` as well

packages/lit/src/test/node-imports.ts
```diff
@@ -45,3 +45,7 @@ import 'lit/html.js';
 import 'lit/experimental-hydrate-support.js';
 import 'lit/experimental-hydrate.js';
 import 'lit/static-html.js';
+
+import assert from 'node:assert/strict';
+import {isServer} from 'lit';
+assert.ok(isServer, 'Expected isServer to be truthy');
```
aomarks, line 51: Same comment as above, suggest `assert.equal`

packages/lit/src/test/node-imports.ts
```diff
@@ -45,3 +45,7 @@ import 'lit/html.js';
 import 'lit/experimental-hydrate-support.js';
 import 'lit/experimental-hydrate.js';
 import 'lit/static-html.js';
+
+import assert from 'node:assert/strict';
```
aomarks, line 49: Oo! This must be the new Node testing library.

.changeset/eight-hairs-run.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit': minor
+'lit-html': minor
+---
+
+Add `isServer` environment checker
```
aomarks, line 6: Maybe a little more detail so that it reads nicely in the CHANGELOG and on Twitter? Something like:

Adds an `isServer` variable export to `lit` and `lit-html/is-server.js` which will be `true` on Node and `false` in the browser. This can be used when authoring components to change behavior based on whether or not the component is executing in an SSR context.

aomarks (approved):
So simple, great!

.changeset/eight-hairs-run.md
justinfagnani, line 6: Can you describe the module path in there?

packages/lit-html/src/test/is-server_test.ts
justinfagnani, line 12: Go ahead and do a stricter check. We want this to be actually `false`

packages/lit-html/src/test/node-imports.ts
```diff
@@ -35,3 +35,7 @@ import 'lit-html/static.js';
 import 'lit-html/experimental-hydrate.js';
 import 'lit-html/private-ssr-support.js';
 import 'lit-html/polyfill-support.js';
+
+import assert from 'node:assert/strict';
+import {isServer} from 'lit-html/is-server.js';
+assert.ok(isServer, 'Expected isServer to be truthy');
```
justinfagnani, line 41: Interesting!

packages/lit-html/src/is-server.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const NODE_MODE = false;
+
+export const isServer = NODE_MODE;
```
justinfagnani, line 9: Add JS doc comment to the export and add a @file comment for the file.

justinfagnani (changes requested):

packages/lit-html/src/test/is-server_test.ts
augustjk: Used `assert.strictEqual` instead for the `===` check.

packages/lit/src/test/node-imports.ts
augustjk: Also used `assert.strictEqual` although it appears that importing from `node:assert/strict` makes everything strict, so `assert.equal` will behave as `assert.strictEqual`.

packages/lit/src/test/node-imports.ts
augustjk: Just the [`assert`](https://nodejs.org/api/assert.html#assert) that's available globally in Node. The new [test runner](https://nodejs.org/api/test.html#test-runner) is still considered experimental but looks neat!

.changeset/eight-hairs-run.md
augustjk: Great suggestion. Done!

packages/lit-html/src/test/is-server_test.ts
augustjk: Done!

packages/lit-html/src/is-server.ts
augustjk: Done!

packages/lit-html/src/is-server.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This file exports a boolean const whose value will depend on what environment
+ * the module is being imported from.
+ */
+
+const NODE_MODE = false;
+
+/**
+ * Boolean value that will `true` when imported via the "node" export condition
+ * and `false` otherwise.
```
aomarks, line 18: Maybe add a similar sentence about when you would use it for SSR, from the changeset. Since this is what will show up on lit.dev API docs.

aomarks (approved):

justinfagnani (approved):

augustjk:
Pipeline should be fixed by #3320

PR #3320, opened 2022-09-29
[lit] Add "types" to package exports

augustjk:
### Problem
I noticed build failing in pipeline like this https://github.com/lit/lit/actions/runs/3148045621/jobs/5118162175#step:7:67
Module self reference wasn't working during TypeScript build due to missing "types" export condition for the `lit` package.

### Solution
Add "types" field. This needed to point to `./development/...` for the module self reference to work during TypeScript build as that is the specified output directory, so this PR also adds `development/` directory to the published package, as well as the "development" export condition. This particular package doesn't really need a "development" export condition at the moment but added those for consistency with our other packages and since the package published to npm would now be including those files anyway.

This also has the benefit of allowing TS `moduleResolution` to be `nodenext`.

Supersedes #3252

### DIFF
--- a/packages/lit/package.json
+++ b/packages/lit/package.json
@@ -19,197 +19,158 @@
   "exports": {
     ".": {
       "types": "./development/index.d.ts",
-      "development": "./development/index.js",
       "default": "./index.js"
     },
     "./async-directive.js": {
       "types": "./development/async-directive.d.ts",
-      "development": "./development/async-directive.js",
       "default": "./async-directive.js"
     },
     "./decorators.js": {
       "types": "./development/decorators.d.ts",
-      "development": "./development/decorators.js",
       "default": "./decorators.js"
     },
     "./decorators/custom-element.js": {
       "types": "./development/decorators/custom-element.d.ts",
-      "development": "./development/decorators/custom-element.js",
       "default": "./decorators/custom-element.js"
     },
     "./decorators/event-options.js": {
       "types": "./development/decorators/event-options.d.ts",
-      "development": "./development/decorators/event-options.js",
       "default": "./decorators/event-options.js"
     },
     "./decorators/property.js": {
       "types": "./development/decorators/property.d.ts",
-      "development": "./development/decorators/property.js",
       "default": "./decorators/property.js"
     },
     "./decorators/query-all.js": {
       "types": "./development/decorators/query-all.d.ts",
-      "development": "./development/decorators/query-all.js",
       "default": "./decorators/query-all.js"
     },
     "./decorators/query-assigned-elements.js": {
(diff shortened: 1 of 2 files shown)


packages/lit/package.json
```diff
@@ -18,120 +18,198 @@
...
     },
     "./polyfill-support.js": {
+      "types": "./development/polyfill-support.d.ts",
+      "development": "./development/polyfill-support.js",
       "default": "./polyfill-support.js"
     },
     "./static-html.js": {
+      "types": "./development/static-html.d.ts",
+      "development": "./development/static-html.js",
       "default": "./static-html.js"
```
aomarks, line 213: I don't think we need development for lit, because all the differences are I'm the underlying packages

augustjk: I agree that it's no different than the default build. The types field does need to point to the `./development/*` directory for the self reference to work during build time since that is the specified `outDir` so I did have to add the entire `/development/` directory to the published package. We could possibly just include the `d.ts` and `d.ts.map` files there.

I noticed some of the labs packages also include the (unnecessary) `"development"` export probably just from copy pasting so I opted to just follow that here too.

An alternative, and perhaps arguably more sensible, thing to do might be to change the tsc outDir to emit directly to the package root and not do the treemirror copy for the packages that don't need a development build but that seemed like extra work and I wasn't sure if that would require tweaking the rollup config as well.

justinfagnani: > An alternative, and perhaps arguably more sensible, thing to do might be to change the tsc outDir to emit directly to the package root and not do the treemirror copy for the packages that don't need a development build but that seemed like extra work and I wasn't sure if that would require tweaking the rollup config as well.

+1 to that, the minified files in `lit` only save a couple of spaces and newlines.

But also +1 to not trying to do that first to fix the tests. This change seems reasonable to me given that we currently compile to `/development/`, Rollup to the default location, and treemirror in the types, just like in lit-html, etc.

augustjk: I went ahead and removed the `"development"` export conditions from the `lit` package since there's nothing different in those so no need to even include them in `package.json`. I'm still keeping the existing tsc output directory and rollup since I don't want to mess with that, and therefore `"types"` export will still point to `./development/*.d.ts` and the `/development/` directory output will be included in the npm package for those. I think that's a good compromise for now.

aomarks (approved):

Mister-Hope:
Any release schedule about this? My CI has been fallen because of it since Aug.

augustjk:
> Any release schedule about this? My CI has been fallen because of it since Aug.

We're looking to get it out soon. Hopefully within the week!

augustjk:
@Mister-Hope This is now published in `lit@2.4.0`

Issue #2902, opened 2022-05-14
[labs/observers] IntersectionController.observe should restore the observed target on reconnect


e111077:
# Description

If you init an IntersectionController with a target (`oldTarget`), and then observe another target with `IntersectionController.observe(newTarget)` (meaning to override `oldTarget`), if the host is disconnected then reconnected, then `oldTarget` is the only target that is observed and not `newTarget`.

# Steps to Reproduce

1. Init an `intersectionObserver` with a target
2. call `IntersectionObserver.observe` with a new target
3. disconnect the custom element
4. reconnect the custom element
5. note that the instersectionController is only being called with the old target

## Live Reproduction Link

https://lit.dev/playground/#gist=92c47a7579efd223f8490988e6f47efb

* Scroll up and down to see the log update
* Observe the targets being observed
* click the `Disconnect the CE` button
* click the `Reconnect the CE` button
* scroll around again and observe only `outsideDiv` is being called

## Expected Results

After reconnect, `target: intendedTarget` should be logged.

Personally I think there should be a way to unobserve the constructor's target, but if not, calling `this.observe(A)` should re-observe on element `A` on host reconnect

## Actual Results

After reconnect, only `target: oldTarget` is logged.

Issue #3237, opened 2022-08-20
[labs/observers] ResizeController "observe" is available publicly, but "unobserve" is not.


Westbrook:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Observers (@lit-labs/observers)

### Description

You can currently `ResizeController.observe(...)` new elements across the lifecycle of the controller, but you aren't given API to _unobserve_ those elements. This method should likely be made private so that it is less likely to lead someone to leverage it and some sort of setter for the `target` (hopefully one that is [easier to use](https://github.com/lit/lit/issues/3147) than at current and accepts an array of elements) should be surfaced to ensure that more content can be easily managed by the observer.

This seems like a common patterns in the observer controllers, so each might benefit the same API structure.

Happy to mark up a PR for this if it's acceptable.

### Alternatives and Workarounds

WORKAROUND:
You can force the type today with:

```
(this.resizeController as unknown as {
    _observer: ResizeObserver
})._observer.unobserve(item);
```

ALTERANATIVE 1:
Or the `observe` method could be made `private` to clarify that this isn't functionality that should be taken advantage of outside of the `host` and `target` arguments in the constructor.

justinfagnani:
Since ResizeObserver has an `unobserve()` method, it makes sense for the controller to also, for symmetry. I'd maybe be in favor of removing `observe()`, but that's a different discussion.

This is a great issue for a contributino!

PR #3321, opened 2022-09-30
[labs/observers] observed targets are now re-observed when the host is reconnected

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/2902

### Context

Previously we would only track a single `_target` that was assigned in the constructor. This `_target` would be re-connected on `hostConnected`. Unfortunately this resulted in imperatively added targets via `observe(target)` to not be re-observed.

### Fix

This is a light lift fix where we now store a set of the observed targets. This means any targets that are observed imperatively get re-observed after host is re-connected.

Note this change only impacts: ResizeController, IntersectionController, and MutationController which accept a `target` to observe.

### Future work

This PR makes it very simple to support an "unobserve" public API (https://github.com/lit/lit/issues/3237) which is coming in a followup PR.

### Test plan

Tested via unit tests.

### DIFF
--- a/packages/labs/observers/src/intersection_controller.ts
+++ b/packages/labs/observers/src/intersection_controller.ts
@@ -62,7 +62,7 @@ export interface IntersectionControllerConfig<T = unknown> {
  */
 export class IntersectionController<T = unknown> implements ReactiveController {
   private _host: ReactiveControllerHost;
-  private _target: Element | null;
+  private _targets: Set<Element> = new Set();
   private _observer!: IntersectionObserver;
   private _skipInitial = false;
   /**
@@ -84,13 +84,14 @@ export class IntersectionController<T = unknown> implements ReactiveController {
    */
   callback?: IntersectionValueCallback<T>;
   constructor(
-    host: ReactiveControllerHost,
+    host: ReactiveControllerHost & Element,
     {target, config, callback, skipInitial}: IntersectionControllerConfig<T>
   ) {
     this._host = host;
     // Target defaults to `host` unless explicitly `null`.
-    this._target =
-      target === null ? target : target ?? (this._host as unknown as Element);
+    if (target !== null) {
+      this._targets.add(target ?? host);
+    }
     this._skipInitial = skipInitial ?? this._skipInitial;
     this.callback = callback;
     // Check browser support.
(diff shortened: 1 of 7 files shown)


packages/labs/observers/src/intersection_controller.ts
```diff
@@ -124,8 +125,8 @@ export class IntersectionController<T = unknown> implements ReactiveController {
   }
 
   hostConnected() {
-    if (this._target) {
-      this.observe(this._target);
+    for (const target of this._targets.values()) {
```
sorvell, line 128: Sets are iterable so it's not clear why you need to use `.values()` here.

packages/labs/observers/src/intersection_controller.ts
```diff
@@ -147,6 +148,7 @@ export class IntersectionController<T = unknown> implements ReactiveController {
    * @param target Element to observe
    */
   observe(target: Element) {
+    this._targets.add(target);
```
sorvell, line 151: I think the DOM APIs won't observe the same element multiple times, but can you confirm?

sorvell (approved):
One minor suggestion but LGTM.

packages/labs/observers/src/intersection_controller.ts
AndrewJakubowicz: My mistake. `values` is not needed.

packages/labs/observers/src/intersection_controller.ts
AndrewJakubowicz: From what spec text I can find:

- Resize Observer will ensure only one subscription per element, by unobserving and re-observing existing elements. ([ref](http://w3c.github.io/csswg-drafts/resize-observer-1/#dom-resizeobserver-observe))

- Intersection Observer will early return if an element is already being observed. ([ref](https://w3c.github.io/IntersectionObserver/#observe-target-element))

- Mutation Observer is less clear. It looks like it may allow multiple `observe` calls per element because it allows different config options per `observe` call. In our case we always reuse the same config.([ref](https://dom.spec.whatwg.org/#dom-mutationobserver-observe)).

I can add a test for the mutation observer case.

AndrewJakubowicz: I manually tested the mutation case. The same element with the same config will not be observed multiple times. Resolving.

Issue #2929, opened 2022-05-19
[cli] Consider path library for guaranteeing OS-safe paths via types


kevinpschaaf:
https://github.com/lit/lit/pull/2899 fixes a lot of path handling in the CLI-related packages to work on windows in addition to unix.

An idea for making this more foolproof is to make a path library to use throughout in place of node's `path` library that returns branded `Path` types for all operations. All CLI operations would accept only branded path types rather than strings, and the only way to make one from a string would be to use the custom path lib which would ensure the path is normalized.

This would effectively be a generalization of `AbsolutePath` / `PackagePath` branded types already used in analyzer; the difference is that paths would never be cast; they would always be created via a library.

Something along the lines of:

* `type AbsolutePath`
* `type RelativePath`
* `type Path = AbsolutePath | RelativePath`
* `type PackagePath` extends `RelativePath`
* `abs(absolutePath: string): AbsolutePath` - runs `path.resolve()` + `path.normalize()` and casts to `AbsolutePath`
* `rel(relativePath: string): RelativePath` - runs `path.normalize()` and casts to `RelativePath`
* `absToPackage(path: AbsolutePath, packageRoot: AbsolutePath): RelativePath` - returns `RelativePath` of `path` in `packageRoot`
* `resolve(root: Path, ...parts: RelativePath[]): AbsolutePath` - resolves the path parts to an absolute path
* `join<T extends Path>(root: T, ...parts: RelativePath[]): T` - joins the path parts and returns the type of the root

justinfagnani:
One question is whether we also need a `NativePath` vs `PosixPath` distinction, so that `abs()` and `rel()` would return `AbsolutePath & NativePath` and `RelativePath & NativePath` respectively. Or, would `AbsolutePath` and `RelativePath` _always_ be native? Is there a case for typing use-input as absolute or relative but not yet OS-native?

justinfagnani:
Adding to the list: `contains(root: AbsolutePath, child: AbsolutePath | RelativePath): boolean;`

kevinpschaaf:
> One question is whether we also need a NativePath vs PosixPath distinction

I think the goal is that we never need that distinction? i.e. User paths are immediately turned into native paths, and native is used throughout. But will need to validate against the actual use cases in the code.

PR #3333, opened 2022-10-05
[labs/analyzer] Cache Module models based on dependencies.

kevinpschaaf:
* Adds `dependencies` to `Module` model, based on scanning for `ts.ImportStatement`s during AST walk
* Changes `getModule()` to take a source file path rather than a `ts.SourceFile` object
* Adds a cache to `getModule`, keyed off of the source file path.
  * The cached module is invalidated upon next `getModule` call if its sourceFile has changed or any of its dependenies' sourceFiles have changed (but only if a dependent `Module` has already been cached
* For unit tests, adds an `InMemoryAnalyzer` to test utils


packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -5,77 +5,194 @@
...
+  // The filename appears to come out of the ts API with
+  // unix separators; since sourcePath uses OS separators, normalize this so
+  // that all our model paths are OS-native
+  return analyzer.path.normalize(outputPath) as AbsolutePath;
+};
+
+/**
+ * Resolve a module specifier to an absolute path on disk.
+ */
+const getPathForModuleSpecifier = (
```
kevinpschaaf, line 177: This function was just factored out from the original getModule() code.

packages/labs/analyzer/src/lib/javascript/packages.ts
```diff
@@ -19,13 +19,13 @@ export const getPackageRootForModulePath = (
   const {fs, path} = analyzer;
   let searchDir = modulePath as string;
   const root = path.parse(searchDir).root;
```
kevinpschaaf, line 21: This is a minor fix for finding a package.json right in the filesystem root, i.e. `/package.json` (hit this with the in-memory tests). The loop bailed as soon as we hit the root, before checking in there.

Note, this and the next function are similar to these used internally in TS, but they are marked as `/** @internal */`: https://github.com/microsoft/TypeScript/blob/221cf55a21e448bd3fe2cf26a754c9c0dda3dca3/src/compiler/moduleNameResolver.ts#L1825-L1882

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -66,7 +63,7 @@ export class Analyzer implements AnalyzerInterface {
       ...packageInfo,
       modules: rootFileNames.map((fileName) =>
         getModule(
-          this.program.getSourceFile(this.path.normalize(fileName))!,
+          this.path.normalize(fileName) as AbsolutePath,
```
justinfagnani, line 66: Do we have utilities that validate/compute and return the right path types so we don't have to cast?

packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -5,77 +5,194 @@
...
-import * as path from 'path';
 import {getClassDeclaration} from './classes.js';
 import {getVariableDeclarations} from './variables.js';
 import {AbsolutePath, absoluteToPackage} from '../paths.js';
 import {getPackageInfo} from './packages.js';
+import {DiagnosticsError} from '../errors.js';
+
+// Cache of Module models by path; invalidated when the sourceFile
+// or any of its dependencies change
+const moduleCache = new Map<AbsolutePath, Module>();
```
justinfagnani, line 21: This is where plain functions seem problematic for me. I'd ideally want the state kept with some kind of object so that collisions are contained. What happens if we analyze a program under different compiler settings? Do we have to force this `module.ts` module to be unloaded from the VM somehow?

packages/labs/analyzer/.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
       "request": "launch",
       "name": "Test",
       "skipFiles": ["<node_internals>/**"],
-      "program": "${workspaceFolder}/../../../node_modules/.bin/uvu",
+      "program": "${workspaceFolder}/../../../node_modules/uvu/bin.js",
```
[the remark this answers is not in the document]
kevinpschaaf: Allows the debugger to start tests on Windows. The `uvu` command is bash script that just calls out to `bin.js`.

packages/labs/analyzer/src/lib/analyzer.ts
kevinpschaaf: We have an issue for it: https://github.com/lit/lit/issues/2929

packages/labs/analyzer/src/lib/javascript/modules.ts
kevinpschaaf: As discussed, will move the module cache onto the Analyzer class

kevinpschaaf: Done

.changeset/calm-zoos-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/analyzer': minor
+---
+
+Cache Module models based on dependencies.
```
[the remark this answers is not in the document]
kevinpschaaf: I don't think we should consider analyzer as having users that need to understand every change yet; this is all still highly under development.

kevinpschaaf:
@justinfagnani Added https://github.com/lit/lit/issues/3370 based on IRL review/conversation.

justinfagnani (approved):

Issue #3339, opened 2022-10-06
Support passing a Document to adoptStyles


rictic:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

It's useful to be able to apply the same styles to the document as well as to an element's shadow root. Documents support adoptedStylesheets, and when they don't we can fall back to appending a style element as a child, same as we do for a ShadowRoot

I think this is just a change needed to the types, because the code works fine at runtime. i.e.

```typescript
import {html, css, LitElement, adoptStyles} from 'lit';
import {customElement, property} from 'lit/decorators.js';

const utils = css`.green { color: green; }`;

// This cast shouldn't be needed
adoptStyles(document as unknown as ShadowRoot, [utils]);

@customElement('simple-greeting')
export class SimpleGreeting extends LitElement {
  static styles = [utils, css` /* element-specific styles here */`];

  render() {
    return html`<div class="green">this should be green too, and it's inside a shadow root</div>`;
  }
}
```

https://lit.dev/playground/

### Alternatives and Workarounds

For now it works to just cast the Document as a ShadowRoot, but that's potentially fragile if we don't intend to support passing in a Document and we don't have a test of it, a future change could break that code. Alternatively user code could reproduce the logic in adoptStyles

PR #3340, opened 2022-10-06
Support calling adoptStyles with a Document

rictic:
Fixes https://github.com/lit/lit/issues/3339

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -166,11 +166,11 @@ export const css = (
  * shadowRoot.
  */
 export const adoptStyles = (
-  renderRoot: ShadowRoot,
+  renderRoot: Document | ShadowRoot,
   styles: Array<CSSResultOrNative>
 ) => {
   if (supportsAdoptingStyleSheets) {
-    (renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) =>
+    renderRoot.adoptedStyleSheets = styles.map((s) =>
       s instanceof CSSStyleSheet ? s : s.styleSheet!
     );
   } else {
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -182,7 +182,13 @@ export const adoptStyles = (
         style.setAttribute('nonce', nonce);
       }
       style.textContent = (s as CSSResult).cssText;
-      renderRoot.appendChild(style);
+      let renderInto;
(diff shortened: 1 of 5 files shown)


packages/reactive-element/src/css-tag.ts
```diff
@@ -182,7 +182,13 @@ export const adoptStyles = (
         style.setAttribute('nonce', nonce);
       }
       style.textContent = (s as CSSResult).cssText;
-      renderRoot.appendChild(style);
+      let renderInto;
+      if (renderRoot instanceof Document) {
```
justinfagnani, line 186: Use `document.nodeName` for cross-document support?

packages/reactive-element/src/css-tag.ts
```diff
@@ -182,7 +182,13 @@ export const adoptStyles = (
...
+        renderInto = renderRoot.head;
```
justinfagnani, line 187: Is this necessary? Two thoughts:

1. Putting the check for a document in the initial render pass for every element is unfortunate
2. This changes the behavior vs native adoptedStyleSheets which always come after document.styleSheets in the CSS cascade. Appending to a ShadowRoot is a close approximation when we assume that a lit-like template has been set up first and nothing will be added after the styles... Document may well have styles in the body though. So if we kept a check I think I'd append to document.documentElement. Those styles till apply, but they're outside of `<body>` and likely to stay last in most pages.

justinfagnani: Oh and with documentElement I think we could do: `(renderRoot.documentElement ?? renderRoot).append(style)`

justinfagnani: cc @kevinpschaaf

packages/reactive-element/src/css-tag.ts
```diff
@@ -166,11 +166,11 @@ export const css = (
  * shadowRoot.
  */
 export const adoptStyles = (
-  renderRoot: ShadowRoot,
+  renderRoot: Document | ShadowRoot,
```
justinfagnani, line 169: One thing I worry about here is that this function does no deduping. Calling it standalone like in your example in #3339 is fine, but if someone called this an override of `createRenderRoot()` they could end up appending styles for every instance of that element class.

I wonder if a better approach would be to make a `adoptDocumentStyles()` function that dedupes and calls this with `document.documentElement`.

sorvell (changes requested):
I think this is ok, but consider my comment below.. If it makes sense we could do it here or as a follow on.

PR #3348, opened 2022-10-12
[@lit-labs/router] Add Routes.link method tests

AndrewJakubowicz:
Test only change that doesn't bump the package. Add tests to the `Routes.link(<pathname>)` method based on its current implementation.

Tested:
 - It accepts an absolute path.
 - The returned url includes the parent.
 - Custom relative local path can be provided.


packages/labs/router/src/test/router_test.ts
```diff
@@ -198,6 +198,60 @@ const canTest =
...
+    const {contentWindow, contentDocument} = container;
+
+    // Set the iframe URL before appending the element
+    contentWindow!.history.pushState({}, '', '/child1/def');
+    contentDocument!.body.append(el);
+    await el.updateComplete;
+    const child1 = el.shadowRoot!.querySelector('child-1') as Child1;
+    await child1.updateComplete;
+
+    const LOCAL_PATH = 'local_path';
```
justinfagnani, line 234: I wouldn't usually worry about making a const in a test case as simple as this. Copy/paste is more acceptable in a test, and it's probably easier to scan without the interpolation below.

justinfagnani (approved):

packages/labs/router/src/test/router_test.ts
AndrewJakubowicz: Good point! Done!

Issue #3296, opened 2022-09-22
[labs/context] @contextProvided and @contextProvider decorators pollute base class when sub-classing


christophe-g:
### Which package(s) are affected?

Context (@lit-labs/context)

### Description

Using decorators to declare context in a sub-class adds it to the base class. 

```js
class Base extends LitElement {
	@contextProvided({context: simpleContext, subscribe: true})
	@property({type: Number})
	public value = 0;
}

class Ext extends Base {
	@contextProvided({context: simpleContext, subscribe: true})
	@property({type: Number})
	public valueExt = 1;
}
```

Any instance of Base will have 2 controllers, its own AND the one declared in the exended class. 

Expected behavior: It should only have one controller. 

### Reproduction

See https://lit.dev/playground/?linkId=8223273#gist=4875267ed996afc7bad538e17fe80907

### Workaround

Do not use context decorators when sub-classing, but this is not really a workaround

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

lit: 2.3.1; @lit-labs/context: 0.1.3

### Browser/OS/Node environment

Chrome stable latest, linux mint.

christophe-g:
@justinfagnani - any ETA on this one ?

I am aware we all are very busy, so asking this mostly for planning purpose on my end. 

I can also try to have a go (without fixed promises, given my limited exposure to decorators and current time availability), if any pointer is provided.

christophe-g:
Closing as solved by #3374

justinfagnani:
Thanks for following up here @christophe-g !

PR #3374, opened 2022-10-18
Initializers are copied but separate from superclass initializers

### LINKED ISSUE #3373: [reactive-element] addInitializer adds initializer to superclass

### Description

While looking #3296 a bit more in depth, I realize that the culprit is not @lit-labs/context but `ReactiveElement.addInitializer`, which will add intializers to the superclass if it already contains its own initializers. 

[Documentation states](https://lit.dev/docs/components/lifecycle/#addInitializer): 

> Initializers are stored per-constructor. Adding an initializer to a subclass does not add it to a superclass. Since initializers are run in constructors, initializers will run in order of the class hierarchy, starting with superclasses and progressing to the instance's class.

This is not what happens as `addInitializer` only checks if this._initializer exists on the Class prototype chain. In case a superclass already has one, it will be added to it (the superclass). We should have a `hasOwnProperty` check there. 

```ts
   /*
   * Initializers are stored per-constructor. Adding an initializer to a
   * subclass does not add it to a superclass. Since initializers are run in
   * constructors, initializers will run in order of the class hierarchy,
   * starting with superclasses and progressing to the instance's class.
   *
   * @nocollapse
   */
  static addInitializer(initializer: Initializer) {
    this._initializers ??= [];
    this._initializers.push(initializer);
  }

```

### Reproduction

- go to [this playground](https://lit.dev/playground/#gist=410407cc4c6bb0051849d6633d0f0f5d)
- See that the base class has 2 initializers (while it shoul donly have one)

### Workaround

none

### Is this a regression?

### Affected versions

Failing in 2.4.0

NA

sorvell:
Fixes #3373.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -491,8 +491,7 @@ export abstract class ReactiveElement
    */
   static addInitializer(initializer: Initializer) {
     this.finalize();
-    this._initializers ??= [];
-    this._initializers!.push(initializer);
+    (this._initializers ??= []).push(initializer);
   }
 
   static _initializers?: Initializer[];
@@ -764,8 +763,9 @@ export abstract class ReactiveElement
     const superCtor = Object.getPrototypeOf(this) as typeof ReactiveElement;
     superCtor.finalize();
     // Create own set of initializers for this class if any exist on the
-    // superclass and copy them down.
-    if (superCtor._initializers) {
+    // superclass and copy them down. Note, for a small perf boost, avoid
+    // creating initializers unless needed.
+    if (superCtor._initializers !== undefined) {
       this._initializers = [...superCtor._initializers];
     }
     this.elementProperties = new Map(superCtor.elementProperties);
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -490,8 +490,9 @@ export abstract class ReactiveElement
    * @nocollapse
    */
   static addInitializer(initializer: Initializer) {
+    this.finalize();
     this._initializers ??= [];
-    this._initializers.push(initializer);
+    this._initializers!.push(initializer);
   }
 
   static _initializers?: Initializer[];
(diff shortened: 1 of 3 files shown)


christophe-g:
>  and the lit team shall be in better position to efficiently tackle this (esp. on the good test requirement).

indeed, thanks !

kevinpschaaf (approved):
LGTM, add a changeset

packages/reactive-element/src/reactive-element.ts
```diff
@@ -490,8 +490,9 @@ export abstract class ReactiveElement
    * @nocollapse
    */
   static addInitializer(initializer: Initializer) {
+    this.finalize();
     this._initializers ??= [];
-    this._initializers.push(initializer);
+    this._initializers!.push(initializer);
```
justinfagnani, line 495: 🏌️‍♂️: `(this._initializers ??= []).push(initializer)`

packages/reactive-element/src/reactive-element.ts
```diff
@@ -762,6 +763,11 @@ export abstract class ReactiveElement
     // finalize any superclasses
     const superCtor = Object.getPrototypeOf(this) as typeof ReactiveElement;
     superCtor.finalize();
+    // Create own set of initializers for this class if any exist on the
+    // superclass and copy them down.
+    if (superCtor._initializers) {
+      this._initializers = [...superCtor._initializers];
```
justinfagnani, line 769: 🏌️‍♂️: `this._initializers = [...superCtor._initializers ?? []];`

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts

packages/reactive-element/src/reactive-element.ts
sorvell: Done for small perf benefit. Added comment.

christophe-g:
@sorvell, @justinfagnani - any chance to release a patch version for this ? Potential consequences of the bug can be dire (see https://github.com/lit/lit/issues/3296 - wich cause one of my app to break when using @lit-lab/context decorators)

PR #3377, opened 2022-10-19
Improvements to Vue/React wrappers

### LINKED ISSUE #2850: [labs/gen-wrapper-react] Add type reference imports to to gen-wrapper-react

### LINKED ISSUE #3313: [labs/gen-wrapper-vue] Add support for default values to Vue generator

sorvell:
Vue
* updates wrapper dependencies on vite/vue
* updates property and event types 
* handles default values per the Vue convention of always setting undefined values to the default value.

React
* adds event types

FIxes #3313 
Fixes #2850

### DIFF
--- a/packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
+++ b/packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
@@ -30,9 +30,9 @@ export default {
   build: {
     rollupOptions: {
       // Ensures no deps are bundled with build.
-      // Source paths are expected to start with \`./\` or \`/\`.
-      external: (id: string, parent: string, isResolved: boolean) =>
-        !isResolved && !id.match(/^\\.?\\//),
+      // Source paths are expected to start with \`./\` or \`/\` but may be
+      // \`x:\` on Windows.
+      external: (id: string) => !id.match(/^((\\w:)|(\\.?[\\\\/]))/),
       input: [
         ${Object.keys(sfcFiles)
           .map((path) => `'./${path}'`)
@@ -44,7 +44,7 @@ export default {
         format: 'es',
         entryFileNames: ({ name }) => \`\${name}.js\`,
         dir: './',
-        sourcemap: false
+        sourcemap: true
       }
     },
     outDir: './'
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/vite.config.ts
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/vite.config.ts
@@ -6,9 +6,9 @@ export default {
   build: {
     rollupOptions: {
       // Ensures no deps are bundled with build.
-      // Source paths are expected to start with `./` or `/`.
-      external: (id: string, parent: string, isResolved: boolean) =>
-        !isResolved && !id.match(/^\.?\//),
+      // Source paths are expected to start with `./` or `/` but may be
+      // `x:` on Windows.
+      external: (id: string) => !id.match(/^((\w:)|(\.?[\\/]))/),
       input: ['./src/ElementA.vue'],
       preserveModules: false,
       preserveEntrySignatures: true,
@@ -16,7 +16,7 @@ export default {
         format: 'es',
         entryFileNames: ({name}) => `${name}.js`,
         dir: './',
-        sourcemap: false,
+        sourcemap: true,
       },
     },
     outDir: './',
--- a/packages/labs/gen-wrapper-vue/test-output/vite.config.ts
+++ b/packages/labs/gen-wrapper-vue/test-output/vite.config.ts
@@ -7,10 +7,8 @@ export default {
       entry: './src/tests/test-element-a_test.ts',
       fileName: () => `test-element-a_test.js`,
       formats: ['es'],
-      sourcemaps: false,
     },
     outDir: './tests',
-    sourcemaps: false,
   },
   plugins: [vue()],
 };


packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementA.vue
```diff
@@ -1,27 +1,48 @@
...
-const props = defineProps<{
+export interface Props {
   foo?: string | undefined;
-}>();
+}
+
+const vueProps = defineProps<Props>();
+
+const defaults = reactive({} as Props);
+const vDefaults = {
```
kevinpschaaf, line 13: It might be nice to see if all of this stuff (and the render stuff) could be pulled out to `vue/wrapper-utils.js` via some sort of composition API function thing that gets passed `vueProps`, since it doesn't contain any gencode itself... not sure it's possible.

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -32,67 +36,131 @@ export const wrapperModuleTemplateSFC = (
   ]);
 };
 
-const getEventType = (event: ModelEvent) => event.type?.text || `unknown`;
+//const getEventType = (event: ModelEvent) => event.type?.text || `unknown`;
```
kevinpschaaf, line 39: Remove

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -32,67 +36,131 @@ export const wrapperModuleTemplateSFC = (
...
+const getEventPayloadType = (type: ModelEvent['type']) => {
+  const {text} = type ?? {text: 'unknown'};
+  const {payload} = text.match(/.*<(?<payload>.*)>/)?.groups ?? {
+    payload: text,
+  };
+  return payload;
+};
+
+const defaultEventType = `CustomEvent<unknown>`;
+const isCustomEvent = (type: string) => /^CustomEvent/.test(type);
```
kevinpschaaf, line 50: We should go with the same approach in React & Vue (either pass detail for CustomEvent, or just pass the Event in all cases). Per convo, maybe just start with always passing the event.

packages/labs/gen-wrapper-react/src/index.ts
```diff
@@ -10,7 +10,9 @@ import {
   PackageJson,
   LitElementDeclaration,
   ModuleWithLitElementDeclarations,
+  getImportsStringForReferences,
 } from '@lit-labs/analyzer';
+import {Event as ModelEvent} from '@lit-labs/analyzer/lib/model.js';
```
kevinpschaaf, line 15: Naming (here and in Vue): `EventModel` reads better

packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
```diff
@@ -30,7 +30,7 @@ export default {
   build: {
     rollupOptions: {
       // Ensures no deps are bundled with this build.
-      external: () => true,
+      external: (id: string) => !!id.match(/^(vue|@lit.*|lit)$/),
```
kevinpschaaf, line 33: Seems like this should be any bare module specifier (or anything that's not relative...?) `!id.match(/^\./)`.

packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementA.vue
kevinpschaaf: And somewhere, we need to add comments about the vDefaults gambit, since it's pretty needle-thready. Especially re: first render on client & SSR.

kevinpschaaf (changes requested):

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
sorvell: Done.

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
sorvell: Done.

packages/labs/gen-wrapper-react/src/index.ts
sorvell: Done.

packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
sorvell: Done with slight modification.

packages/labs/gen-wrapper-react/src/index.ts
```diff
@@ -149,45 +151,60 @@ const tsconfigTemplate = () => {
...
+const getTypeImports = (declarations: LitElementDeclaration[]) => {
+  // We only need type imports for events.
+  const refs = declarations.flatMap(({events}) =>
+    Array.from(events.values()).flatMap((e) => e.type?.references ?? [])
+  );
+  return getImportsStringForReferences(refs);
+};
+
+const getElementTypeExportsFromImports = (imports: string) =>
+  imports.replace(/(?:^import|(\s)*import)/gm, '$1export type');
```
kevinpschaaf, line 163: Why the `|(\s)*import` and `$1` replacement? Don't grok what could precede the import. Add comment?

Long-term, I think we'll want to add a `getExportsStringForReferences` function, since in the general case it's not always that you'll want to import the same references that you export...

kevinpschaaf (approved):
One question, otherwise LGTM

packages/labs/gen-wrapper-react/src/index.ts
sorvell: Fixed and added comment. Was unsure of format of source string but it's `import ...\nimport...` which is fine.

PR #3378, opened 2022-10-19
Update package.json

### LINKED ISSUE #3124: [lit-html] lit-html has a dependecy to a @types package (should be a dev dependency)

### Description

In the package.json there seems to be a wrong dependecy:

https://github.com/lit/lit/blob/main/packages/lit-html/package.json#L281

### Reproduction

look at https://github.com/lit/lit/blob/main/packages/lit-html/package.json#L281

### Workaround

it's still usable

### Is this a regression?

### Affected versions

current

not relevant

jogibear9988:
fixes #3124

### DIFF
--- a/packages/lit-html/package.json
+++ b/packages/lit-html/package.json
@@ -354,11 +354,11 @@
       "output": []
     }
   },
-  "dependencies": {
-    "@types/trusted-types": "^2.0.2"
+  "dependencies": {  
   },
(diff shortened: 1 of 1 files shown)


augustjk:
Thank you for the PR! As noted in the related issue, the inclusion of `@types/trusted-types` as a dependency is intentional.

PR #3384, opened 2022-10-21
Wrapper test elements + runtime tests

### LINKED ISSUE #2854: [labs/test-projects] Add various test elements for use in generators
- element that renders SR
- has properties
- fires events
- has slots
- kitchen sink

sorvell:
- Adds tests elements that stress property types, events, and slots.
- Adds wrappers and runtime tests for the new elements for Vue and React

Fixes #2854


packages/labs/gen-wrapper-react/test-output/package.json
```diff
@@ -71,7 +71,7 @@
         "build",
         "../../../tests:build"
       ],
-      "command": "node ../../../tests/run-web-tests.js \"tests/**/*_test.js\" --config ../../../tests/web-test-runner.config.js",
+      "command": "node ../../../tests/run-web-tests.js \"tests/**/(tests|*_test).js\" --config ../../../tests/web-test-runner.config.js",
```
kevinpschaaf, line 74: `tests.js` pulls in all the `_test` files, so doesn't this run them all twice? Why have the `tests.js` file?

kevinpschaaf: It's because we need an entrypoint for rollup? If so, then seems like that's the only thing that should be here?

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -111,6 +111,9 @@ const wrapperTemplate = (
 ) => {
   const {tagname, events, reactiveProperties} = declaration;
   return javascript`
+    <script lang="ts">
+      export * from '${wcPath}';
```
kevinpschaaf, line 115: Let's change this to export the type references that were imported.

kevinpschaaf (approved):
LGTM, just a few cleanup / fixes.

packages/labs/gen-wrapper-react/test-output/package.json
sorvell: Fixed.

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
sorvell: Fixed.

packages/labs/test-projects/test-element-a/src/element-events.ts
```diff
@@ -0,0 +1,100 @@
...
+    'number-custom-event': CustomEvent<number>;
+    'my-detail-custom-event': CustomEvent<MyDetail>;
+  }
+}
+
+/**
+ * My awesome element
+ * @fires string-custom-event {CustomEvent<string>} A custom event with a string payload
+ * @fires number-custom-event {CustomEvent<number>} A custom event with a number payload
+ * @fires my-detail-custom-event {CustomEvent<MyDetail>} A custom event with a MyDetail payload.
```
kevinpschaaf, line 44: Can we add a 
```ts
* @fires template-result-custom-event {CustomEvent<TemplateResult>}
```
and dispatch an event like 
```ts
    fromNode.dispatchEvent(new CustomEvent('template-result-custom-event', {detail: html``}))
```
so that we can exercise type references from an external package that's not otherwise imported into the generated source?

Note that this will point out an obvious catch-22: I expect the `TemplateResult` reference in the `@fires` will cause the analyzer to error because `TemplateResult` symbol isn't imported into the current file, which I believe is the right thing to do, since we can't otherwise know what that identifier was actually referring to. But adding an `import type {TemplateResult} from 'lit';` will fail compilation because it's not actually used in the TS source, only the comment.

We can either silence the unused symbol warning, or (better imo) just require the user to actually use the imported symbol, ala `new CustomEvent('template-result-custom-event', {detail: html`` as TemplateResult})`.

Either way, this will allow us to test that we get the proper `import {TemplateResult} from 'lit'` in the generated source.

kevinpschaaf (approved):
Other than one suggested test addition, LGTM

packages/labs/test-projects/test-element-a/src/element-events.ts
sorvell: Done.

PR #3386, opened 2022-10-21
[lit-labs/observers] add base class

AndrewJakubowicz:
Issue: https://github.com/lit/lit/issues/2350

This PR adds a base class to IntersectionController, ResizeController, and MutationController.

Found a bug that we accept a ResizeController config and then never use it.

TODO: Add size difference before and after.

### DIFF
--- a/packages/labs/observers/src/base_controller.ts
+++ b/packages/labs/observers/src/base_controller.ts
@@ -1,0 +1,112 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import {ReactiveControllerHost} from '@lit/reactive-element/reactive-controller.js';
+
+interface IObserver {
+  observe(element: Element): void;
+  takeRecords?: () => unknown[];
+  disconnect(): void;
+}
+
+export interface BaseControllerConfig {
+  /**
+   * The element to observe. In addition to configuring the target here,
+   * the `observe` method can be called to observe additional targets. When not
+   * specified, the target defaults to the `host`. If set to `null`, no target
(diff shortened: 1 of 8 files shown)


packages/labs/observers/src/base_controller.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * By default the `callback` is called without changes when a target is
+   * observed. This is done to help manage initial state, but this
+   * setup step can be skipped by setting this to true.
+   */
+  skipInitial?: boolean;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class BaseController<Config, Callback extends (...args: any) => any> {
```
sorvell, line 32: Maybe `ObserverController`?

AndrewJakubowicz: +100 Will do!

packages/labs/observers/src/intersection_controller.ts
```diff
@@ -7,7 +7,10 @@ import {
   ReactiveController,
   ReactiveControllerHost,
 } from '@lit/reactive-element/reactive-controller.js';
-
+import {
+  ObserverController,
+  type ObserverControllerConfig,
+} from './observer_controller.js';
```
nickschaap, line 13: nit: I noticed a lot of other lit packages use kebab case instead of snake case for file naming. Is it worth changing the names of these controllers to normalize consumption with the remainder of the lit repo?

shaal:
Lit's `ResizeController` always call `render()` twice, before any size changes.
Simplified example - https://lit.dev/playground/#gist=d26a19ef6637e12160c703079e37c9af

When using `{skipInitial = true}`, does this PR solve that problem?

PR #3388, opened 2022-10-23
vue-wrapper: Adds props directive

sorvell:
Factors out Vue's conventional "defaults for undefined values" behavior into a `vProps` directive supplied via `wrapper-utils` and used by wrappers.
* All element properties are set by the `vProps` directive.
* It records defaults when the element is created.
* When the element is mounted/updated, properties are set, coercing undefined values to defaults.


kevinpschaaf (approved):
One question, otherwise LGTM

packages/labs/vue-utils/src/wrapper-utils.ts
```diff
@@ -1,4 +1,41 @@
...
+};
+
+/**
+ * Set properties to bound element, preserving Vue's convention of
+ * applying default values when the bound value is `undefined`. Default values
+ * are recorded at created time, before any properties are set. Then, whenever
+ * the element is updated (including mounted), properties are set to a given
+ * defined value or the default.
+ */
+export const vProps: PropsDirective = {
```
kevinpschaaf, line 20: We should implement `getSSRProps` to just return the props bag during SSR

kevinpschaaf (changes requested):

packages/labs/vue-utils/src/wrapper-utils.ts
sorvell: Done.

kevinpschaaf (approved):
LGTM

Issue #3389, opened 2022-10-24
[labs/context] Consider new names for context decorators


justinfagnani:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Context (@lit-labs/context)

### Description

The current context decorator names (`@contextProvider()` and `@contextProvided()`) are easy to confuse, differing by only one letter, and are a bit long.

I propose we simplify these to just `@provide()` and `@consume()` to be shorter and more closely match the names of the context controllers.

While those names are somewhat generic, they will be imported from the `@lit-labs/context` package, so the meaning should be easy enough to determine.

### Alternatives and Workarounds

`@provideContext()` and `@consumeContext()`

Issue #3058, opened 2022-06-17
[labs/context]: Invalid example code


oliversalzburg:
# Description

The example given under "Consuming a context" is invalid. The code raises a TypeScript error.

# Steps to Reproduce

The code in the example already produces the error. It also imports `contextRequest` and then uses `contextProvided`.

## Expected Results

The example code should be valid.

## Actual Results

```
Argument of type 'Logger' is not assignable to parameter of type 'ReactiveElement & Record<"logger", Logger>'.
  Type 'Logger' is not assignable to type 'Record<"logger", Logger>'.
    Types of property 'logger' are incompatible.
      Type 'Logger | undefined' is not assignable to type 'Logger'.
        Type 'undefined' is not assignable to type 'Logger'.
```

## Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

oliversalzburg:
Maybe the code in the implementation is actually wrong. If I make this `Record` a `Partial<Record<...>>`, the error goes away: https://github.com/lit/lit/blob/main/packages/labs/context/src/lib/decorators/context-provided.ts#L50

drk-mtr:
Doing this from memory so apologies if I'm sharing bad info. It may be as simple as changing `logger?` -> `logger!` - and obviously updating the import for `contextProvided`.

justinfagnani:
I have a fix for this in https://github.com/lit/lit/pull/3398

PR #3398, opened 2022-10-27
[labs/context] Rename context decorators to consume and provide

justinfagnani:
Fixes https://github.com/lit/lit/issues/3389
Fixes https://github.com/lit/lit/issues/3058

This is just _one_ potential choice of names, so we should make sure we get consensus.

cc @benjamind <- I can't add you as a reviewer right now.


benjamind (approved):
LGTM! I much prefer this naming.

kevinpschaaf (approved):

Issue #3314, opened 2022-09-28
[labs/context] @contextProvided doesn't work well with optional fields


justinfagnani:
### Which package(s) are affected?

Context (@lit-labs/context)

### Description

It's not possible to use optional fields with `@contextProvided` because of the use of `Record` in the decorator types.

This should work, since it's always possible that there is no provider for a context key:
```ts
const fooContext = createContext<Foo>('foo');

class MyElement extends LitElement {
  @contextProvided({context: fooContext})
  foo?: Foo;
}
```

### Reproduction

See above

### Workaround

Mark fields _and_ context keys as `X | undefined` instead of optional.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

all

### Browser/OS/Node environment

all

PR #3399, opened 2022-10-27
[labs/context] Make @consume decorator work with optional fields

justinfagnani:
Fixes https://github.com/lit/lit/issues/3314

This doesn't allow `@provide()` to work with optional fields, since if you provide a value for a non-nullable context you should not use optional fields.

cc @benjamind


packages/labs/context/src/lib/decorators/consume.ts
```diff
@@ -48,7 +48,7 @@ export function consume<ValueType>({
   context: ContextKey<unknown, ValueType>;
   subscribe?: boolean;
 }): <K extends PropertyKey>(
-  protoOrDescriptor: ReactiveElement & Record<K, ValueType>,
+  protoOrDescriptor: ReactiveElement,
```
kevinpschaaf, line 51: Seems like the generics `ValueType` and `K` are no longer used, remove?

kevinpschaaf: I guess this just means we remove what little type safety existed... but I guess it's just possible to express?

Does `Partial<Record<K, ValueType>>` help? That's sorta like _if_ it's there then it has to have the right type?

justinfagnani: Yeah, that's better. I added a test that we get the error too.

packages/labs/context/src/lib/decorators/consume.ts
```diff
@@ -48,7 +48,7 @@ export function consume<ValueType>({
   context: ContextKey<unknown, ValueType>;
   subscribe?: boolean;
 }): <K extends PropertyKey>(
-  protoOrDescriptor: ReactiveElement & Record<K, ValueType>,
+  protoOrDescriptor: ReactiveElement & Partial<Record<K, ValueType>>,
```
kevinpschaaf, line 51:
```suggestion
  // Partial<> allows for providing the value to an optional field
  protoOrDescriptor: ReactiveElement & Partial<Record<K, ValueType>>,
```

kevinpschaaf (approved):
LGTM, let's add a comment so we remember why we did this

PR #3404, opened 2022-10-28
[labs/context] Rename ContextKey to Context

justinfagnani:
This aligns the `Context` interface with the name used in the Context protocol: https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md#the-createcontext-function

The other difference of Context objects being a specific interface in the protocol is being fixed as https://github.com/webcomponents-cg/community-protocols/issues/19

cc @benjamind


kevinpschaaf:
Hmm, the wording that was used in the spec notwithstanding, I personally prefer having `Key` in there over not, and would go so far as to say it should be `createContextKey()` instead of `createContext()`.

Why? `createContext()` sounds like it's carrying a lot of weight, making something that stores context or something, when in fact all it's doing is defining a "handle" / namespace / key to differentiate one context from another. I think that's clarifying.

justinfagnani:
@kevinpschaaf I don't totally disagree. This API name is based on the React `createContext()` API, and I _think_ the rationale is that there isn't much of a difference between a "context" and a "context key" anyway. ie, there's a value provided to a context, or a value provided for a context key.

One reason to prefer just "context" is that it's shorter. We could do `createContext()` -> `contextKey()`, but I'd still want `@consume({context: numberContext})` instead of `@consume({contextKey: numberContextKey})` or even a norm of `@consume({context: numberContextKey})`.

kevinpschaaf (approved):
We could just do `@consume({key: someContextKey})`, but then the word "context" isn't in the API at all, which isn't great. I guess I'm fine with this, thanks for entertaining.

PR #3409, opened 2022-10-31
[@labs/react] Filter __forwardedRef in prod build

taylor-vann:
Fixes https://github.com/lit/lit/issues/3211
Branched from https://github.com/lit/lit/pull/3217


packages/labs/react/src/create-component.ts
```diff
@@ -286,32 +286,29 @@ export function createComponent<
      *
      */
     override render() {
-      // Since refs only get fulfilled once, pass a new one if the user's
-      // ref changed. This allows refs to be fulfilled as expected, going from
+      const {__forwardedRef, ...userProps} = this.props;
```
kevinpschaaf, line 289: This is pretty clever, could use a comment:
```suggestion
      // Extract and remove __forwardedRef from userProps in a rename-safe way
      const {__forwardedRef, ...userProps} = this.props;
```

.changeset/tidy-flowers-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
+Filter \_\_forwaredRef from build.
```
kevinpschaaf, line 5: I think you can remove the `\` in the markdown for the changelog?

taylor-vann: hmm i dont remember writing the `\` but it's gone now!

packages/labs/react/src/create-component.ts
taylor-vann: added the comment

kevinpschaaf (approved):

PR #3410, opened 2022-10-31
[labs/react] Add Node build

### LINKED ISSUE #3397: [labs/react] Import of wrapped component in Node/Next.js errors

React (@lit-labs/react)

### Description

https://github.com/lit/lit/issues/3079 allowed importing Lit components in Node but components wrapped with `@lit-labs/react` breaks with following error

```
error - ReferenceError: HTMLElement is not defined
    at r.render (file:///home/projects/nextjs-mgpt7h/node_modules/@lit-labs/react/create-component.js:19:1017)
```

This happens when trying to import the wrapped component in Next.js.

### Reproduction

https://stackblitz.com/edit/nextjs-mgpt7h?file=pages%2Findex.js

### Workaround

For Next.js, use `next/dynamic` to lazy load. See commented out code in repro above.

### Is this a regression?

### Affected versions

1.0.9

N/A

augustjk:
Resolves #3397 

This adds a Node build for the `@lit-labs/react` package. ~~The Node build version will provide a minimal shim of an empty `HTMLElement` class to the `createComponent` module, using the same approach as `reactive-element` in #3156.~~

The Node build version skips the `in HTMLElement.prototype` if `HTMLElement` is not defined in global. In that case, relying on `in element.prototype` would be enough, since `ReactiveElement` would be extending an empty `HTMLElement` class from our Node build shim.

This approach will conflict with https://github.com/lit/lit/pull/3128, but it should just be a matter of moving over the `NODE_MODE` check to where `HTMLElement` reference got moved to. We can coordinate depending on which one gets merged first. cc: @taylor-vann

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -4,6 +4,14 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+const NODE_MODE = false;
+const global = NODE_MODE ? globalThis : window;
+
+const htmlElementShimNeeded = NODE_MODE && global.HTMLElement === undefined;
+if (htmlElementShimNeeded) {
+  global.HTMLElement = class HTMLElement {} as unknown as typeof HTMLElement;
+}
+
 // Match a prop name to a typed event callback by
 // adding an Event type as an expected property on a string.
 export type EventName<T extends Event = Event> = string & {
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -5,12 +5,6 @@
  */
 
 const NODE_MODE = false;
-const global = NODE_MODE ? globalThis : window;
-
-const htmlElementShimNeeded = NODE_MODE && global.HTMLElement === undefined;
-if (htmlElementShimNeeded) {
-  global.HTMLElement = class HTMLElement {} as unknown as typeof HTMLElement;
-}
 
 // Match a prop name to a typed event callback by
 // adding an Event type as an expected property on a string.
@@ -323,7 +317,7 @@ export function createComponent<
         if (
           eventProps.has(k) ||
           (!reservedReactProperties.has(k) &&
-            !(k in HTMLElement.prototype) &&
+            !(NODE_MODE ? false : k in HTMLElement.prototype) &&
             k in element.prototype)
         ) {
           this._elementProps[k] = v;
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -317,7 +317,11 @@ export function createComponent<
         if (
           eventProps.has(k) ||
           (!reservedReactProperties.has(k) &&
-            !(NODE_MODE ? false : k in HTMLElement.prototype) &&
+            // NODE_MODE check below is to support React SSR where HTMLElement
+            // might not be defined
+            !(NODE_MODE && typeof HTMLElement === 'undefined'
+              ? false
+              : k in HTMLElement.prototype) &&
             k in element.prototype)
         ) {
           this._elementProps[k] = v;
--- a/.changeset/eight-elephants-sparkle.md
+++ b/.changeset/eight-elephants-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/react': minor
+---
+
+Add Node build for server rendering. This will allow wrapped components to be added to SSR React frameworks like Next.js.
--- a/.eslintignore
+++ b/.eslintignore
@@ -241,6 +241,7 @@ packages/labs/observers/performance_controller.*
 packages/labs/observers/resize_controller.*
 packages/labs/observers/intersection_controller.*
 packages/labs/react/development/
+packages/labs/react/node/
 packages/labs/react/test/
 packages/labs/react/node_modules/
 packages/labs/react/index.*
--- a/.prettierignore
+++ b/.prettierignore
@@ -227,6 +227,7 @@ packages/labs/observers/performance_controller.*
 packages/labs/observers/resize_controller.*
 packages/labs/observers/intersection_controller.*
 packages/labs/react/development/
+packages/labs/react/node/
 packages/labs/react/test/
 packages/labs/react/node_modules/
 packages/labs/react/index.*
--- a/packages/labs/react/.gitignore
+++ b/packages/labs/react/.gitignore
@@ -1,4 +1,5 @@
 /development/
+/node/
 /test/
 /node_modules/
 /index.*
--- a/packages/labs/react/package.json
+++ b/packages/labs/react/package.json
@@ -20,17 +20,20 @@
     ".": {
       "types": "./development/index.d.ts",
       "development": "./development/index.js",
+      "node": "./node/index.js",
       "default": "./index.js"
     },
     "./use-controller.js": {
       "types": "./development/use-controller.d.ts",
       "development": "./development/use-controller.js",
+      "node": "./node/use-controller.js",
       "default": "./use-controller.js"
     }
   },
   "files": [
     "/development/",
     "!/development/test/",
+    "/node/",
     "/index.{d.ts,d.ts.map,js,js.map}",
     "/create-component.{d.ts,d.ts.map,js,js.map}",
     "/use-controller.{d.ts,d.ts.map,js,js.map}"
@@ -43,6 +46,7 @@
     "test": "wireit",
     "test:dev": "wireit",
     "test:prod": "wireit",
+    "test:node": "wireit",
     "checksize": "wireit"
   },
   "wireit": {
@@ -111,7 +115,8 @@
     "test": {
       "dependencies": [
         "test:dev",
-        "test:prod"
+        "test:prod",
+        "test:node"
       ]
     },
     "test:dev": {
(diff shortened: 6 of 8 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,14 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+const NODE_MODE = false;
+const global = NODE_MODE ? globalThis : window;
+
+const htmlElementShimNeeded = NODE_MODE && global.HTMLElement === undefined;
+if (htmlElementShimNeeded) {
```
aomarks, line 11: Note in the reactive-element case, we also add the following at the end of the module:

```ts
if (htmlElementShimNeeded) {
  delete (global as Partial<typeof global>).HTMLElement;
}
```

https://github.com/lit/lit/pull/3156/files#diff-305d9543c5dd0eef82d2583a84422cfa2a1786f1f466692485e6dac5eb7d43baR1501

The goal is to avoid polluting the global namespace when we don't need to, and the the reason we used the "add it to global and then remove it later" trick was because TypeScript had trouble understanding ternaries that switched between global `HTMLElement` and a local/inline one.

augustjk: Right, thank you for pointing this out. I'm unsure if that's possible here. `HTMLElement` actually needs to be present during runtime. The error isn't happening strictly at import time but when React tries to server render the component, calling the `render()` method. Perhaps it might be better to change the code within `render()` to remove the reference to `HTMLElement` instead. That'll likely cause a conflict with https://github.com/lit/lit/pull/3128 though.

augustjk: https://github.com/lit/lit/commit/fe450aa4766dbce30dc6abbdfef1b320835f4adf

Here's what that would look like. It does feel cleaner to go about it this way..

aomarks: Yeah, I like that idea!

aomarks (approved):

augustjk:
I modified it a bit, just in case a user happens to bring in a DOM shim that does define `HTMLElement`. @aomarks could you take another look? Thanks!

packages/labs/react/src/create-component.ts
aomarks: > I modified it a bit, just in case a user happens to bring in a DOM shim that does define `HTMLElement`. @aomarks could you take another look? Thanks!

Good call, LGTM!

aomarks (approved):

packages/labs/react/src/create-component.ts
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+const NODE_MODE = false;
```
taylor-vann, line 7: NODE_MODE is always false?

taylor-vann:
This shouldn't be a whatever lands first situation.
React is a package that has around 300k downloads a week, we need to align on this

augustjk:
> This shouldn't be a whatever lands first situation. React is a package that has around 300k downloads a week, we need to align on this

I'm saying the conflict will be easy to resolve whichever gets merged first. This node build is strictly an addition that does not affect existing client side behavior in any way.

Here's how a merge of this into your PR with conflicts resolved would look like https://github.com/lit/lit/commit/adde5b809cfd749502ef7a3fb5c3547307a98d3e#diff-70e2e47659157c4129a63acfad45aa00d53e5314c322b4acc9e677cb1cbaf97a

packages/labs/react/src/create-component.ts
augustjk: `NODE_MODE` flag is flipped to `true` for node builds via rollup.

packages/labs/react/src/create-component.ts
```diff
@@ -315,7 +317,11 @@ export function createComponent<
         if (
           eventProps.has(k) ||
           (!reservedReactProperties.has(k) &&
-            !(k in HTMLElement.prototype) &&
+            // NODE_MODE check below is to support React SSR where HTMLElement
+            // might not be defined
+            !(NODE_MODE && typeof HTMLElement === 'undefined'
```
taylor-vann, line 322: in https://github.com/lit/lit/pull/3410 there is no base prototypes being used in this space, only the provided element.
```
        if (eventProps.has(k) || k in element.prototype) {
          this._elementProps[k] = v;
          continue;
        }
```

This check needs to be done at the `setProperty` function. And that's where the `HTMLElement` or whatever prototype if-ing would be great

also nit but `if (NODE_MODE` is always false than unnecessary?

I would advise rebasing this off of https://github.com/lit/lit/pull/3410

taylor-vann:
We need tests that prove this change works as intended.

packages/labs/react/src/create-component.ts
taylor-vann: so rollup replaces this with `const NODE_MODE = true`? Cool

augustjk: https://github.com/lit/lit/blob/9f802646d955198cbaf6e521283fe137e7f5b7a6/rollup-common.js#L447

packages/labs/react/src/create-component.ts
```diff
@@ -315,7 +317,11 @@ export function createComponent<
...
+              ? false
+              : k in HTMLElement.prototype) &&
```
taylor-vann, line 324: I feel like this ternary should be able to be done at the `module` level.

There should be a way to get the `prototype` or `false` outside of the `ReactComponent.render` function and outside of the `createComponent` function?

```
const SERVER_CHECK = <is node server> ? element : HTMLElement;
```

then later we can use the appropriate prototype instead of both?

packages/labs/react/src/create-component.ts
augustjk: This PR aims to simply add the node build without changing any of the client-side runtime behavior from the current state of the package. Also I believe you mean https://github.com/lit/lit/pull/3128. I've added where I'd move the `NODE_BUILD` check for that branch here https://github.com/lit/lit/pull/3410#issuecomment-1297853312

I don't particularly see a reason why one has to go before the other but I'm okay to wait on this if that's what we'd like.

packages/labs/react/src/create-component.ts
taylor-vann: The ternary is potentially run for every property on every render. It should be only called when necessary.

Also the following in-out check disappears in https://github.com/lit/lit/pull/3128
```
 !(k in HTMLElement.prototype) && ...
```

So this check might not conflict with the current code, but it does not fit in the proposed changes of https://github.com/lit/lit/pull/3128/files

taylor-vann: I understand the need for it to be close to the logic. But it's too much happening in an `if` statement. It makes it more difficult to read and maintain later on. Accounting for a fallback at the module level is more durable. This is technically a nested if statement except it's nested at the condition statement.

We are `if`-ing a prototype and a ternary inside a conditional statement does not read that way to me

taylor-vann: > element and HTMLElement are not the interchangeable part here.

`HTMLElement` is not used in #3128, this is what i mean by `conflicts`. Not conflicts as in git i mean conflicts as in implementation and correctness

taylor-vann (changes requested):
see comments

taylor-vann:
Also, I don't know if making `createComponent` aware of what MODE it is in the best approach.

Have we considered an `HTMLElement` shim of our own at the SSR side of things?

I'd like to leave `createComponent` as agnostic as possible since it can possibly be used on the server-side and client-side

augustjk:
@taylor-vann Appreciate the feedback, and apologies for not looping you in sooner!

Per offline discussion, we'll hold on this PR until #3128 goes in, which moves the `HTMLElement` reference out of `render`. I'll revisit afterwards to see if that's enough to not cause errors during initial server render by React, in which case the node build might not be immediately necessary.

For future SSR considerations, beyond making it not error, we'll have to look at the output by React's server rendering to consider which attributes should be added to the custom element SSR result, and whether we need to suppress React's own hydration warning for these wrapped custom elements as done in @kevinpschaaf 's early React SSR exploration work here https://github.com/lit/lit/blob/lit-react-ssr-old/packages/labs/ssr/src/lib/react/create-element.ts. I anticipate we'll eventually either need a node build or conditional code with `isServer` to handle that.

Putting this into draft for now.

packages/labs/react/src/create-component.ts
e111077: Was lurking and not very plugged into this, but was reading the source on the branch and:

> it's too much happening in an if statement

Agree with Brian on this one. If you're keeping this logic, is there a way to rewrite this without hurting perf? Nested `if` I think would be okay here if only for readability

packages/labs/react/src/create-component.ts
taylor-vann: After sync, I understand there will mostly likely need to be some kind of node check. Marking as resolved! :)

packages/labs/react/src/create-component.ts
augustjk: Yup. Now with the reference to `HTMLElement` moved to a lifecycle that won't be called on the server, this gnarly condition here won't be necessary. To address prop/attribute handling during server-render, I anticipate it'll be a much more straight forward single `if (NODE_MODE)` or `if (isServer)` check instead.

augustjk:
Superseded by https://github.com/lit/lit/pull/3885

Issue #3400, opened 2022-10-27
[labs/virtualizer] Virtualizer in a padded container stubbornly asserts size ignoring padding.


usergenic:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

Gray already figured out the fix for this.  Just adding to an issue to get this added to sprint plan...

![image](https://cdn.discordapp.com/attachments/1029426702599864431/1034997070475571251/unknown.png)

### Reproduction

This demos the issue with an 80px padding for the element containing the virtualizer
https://lit.dev/playground/#gist=3f55b065eb257cbaf21b1168cbb65093

### Workaround

n/a

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

masonry preview and 0.7.2

### Browser/OS/Node environment

n/a

usergenic:
Here's the example with no padding in the container:
<img width="1437" alt="Screen Shot 2022-11-01 at 10 45 20 AM" src="https://user-images.githubusercontent.com/578/199289878-5a7ab8fd-a906-4c80-99d5-44a55db2bb1f.png">

Here's the example with padding added to the container, see how it's all shifted over instead of padded from both sides:
<img width="721" alt="Screen Shot 2022-11-01 at 10 45 46 AM" src="https://user-images.githubusercontent.com/578/199289957-8a6f7558-b55c-4b17-bdc4-5200c5bc5a64.png">

Here's the example with the suggested fix, but at this width it only fits one column:
<img width="1437" alt="Screen Shot 2022-11-01 at 10 46 08 AM" src="https://user-images.githubusercontent.com/578/199290047-8a6d4f2f-c5fe-4c43-b354-2442e5b10b89.png">

But what is more an issue is that virtualizer goes into some kind of update loop which both kills performance and never seems to resolve:
https://user-images.githubusercontent.com/578/199290160-94a81441-68ab-4eb6-923a-159d19be556c.mov

PR #3424, opened 2022-11-03
[labs/virtualizer] Fix width inheritance calculation (issue #3400)

usergenic:
The fix here for #3400 was @graynorton's but it took a while to come up with a reliable test for a number of weird reasons.  In the end I opted for a simple test to show where virtualizer would choose to place an item in grid layout.  This was the easiest test case to reason about and it remains now essentially as a regression test.

### DIFF
--- a/.changeset/moody-singers-doubt.md
+++ b/.changeset/moody-singers-doubt.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/virtualizer': patch
+---
+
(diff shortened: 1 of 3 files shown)


graynorton (approved):
LGTM!

usergenic:
I decided to force push on a rebased-on-main version of branch because the chromedriver fix wasn't getting picked up by github actions for some reason after multiple attempts and I couldn't figure it out.

Issue #3341, opened 2022-10-10
[labs/virtualizer] unable to import RangeChangedEvent and VisibilityChangedEvent from full module file specifier '@lit-labs/virtualizer/lit-virtualizer.js'


usergenic:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

It looks like we don't have `./Virtualizer.js` in the exported modules which prevents importing the RangeChangedEvent and VisibilityChangedEvent from an explicit file specifier.

https://github.com/lit/lit/blob/9e978a0c38eddf5631a1dde9bdb182cccbb33a49/packages/labs/virtualizer/package.json#L9

Would have expected that importing from `@lit-labs/virtualizer/lit-virtualizer.js` would have worked but apparently the transitive dependency is not automatically provided without the explicit module export in package.json for `Virtualizer.js`

### Reproduction

```ts
import {RangeChangedEvent} from '@lit-labs/virtualizer/lit-virtualizer.js';`
```

Returns a `Could not resolve import` error.

### Workaround

If you use just the package specifier `@lit-labs/virtualizer` that works.  But that is not a work-around if your project requires full file specifier.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

0.7.2 and 0.0.0-masonry-preview-20220906220426

### Browser/OS/Node environment

n/a

usergenic:
#3430 fixes this issue and will be available in forthcoming release.

PR #3430, opened 2022-11-04
[labs/virtualizer] Export event classes through a new events.js

usergenic:
Apparently, because Virtualizer.js (which defined RangeChangedEvent and VisibilityChangedEvent) was not exported it in the packages.json it was not possible to import RangeChangedEvent and VisibilityChangedEvent by way of `@lit-labs/virtualizer/lit-virtualizer.js` which re-exports them.  To resolve this issue I have moved the event class definitions into a new exported `events.js` file.

Confirmed that this supports now these three import statements:
```ts
import {RangeChangedEvent, VisibilityChangedEvent} from '@lit-labs/virtualizer';
import {RangeChangedEvent, VisibilityChangedEvent} from '@lit-labs/virtualizer/events.js';
import {RangeChangedEvent, VisibilityChangedEvent} from '@lit-labs/virtualizer/lit-virtualizer.js';
```

(I think the first two are preferred form whereas the third is just the more explicit form of the first.)


usergenic:
Fixes #3341

graynorton (approved):
LGTM

PR #3431, opened 2022-11-07
feat(labs/ssr): Prevent exposure of window object on globalThis

### LINKED ISSUE #3142: [labs/task] Task needs an arguments equality function.

Task (@lit-labs/task)

### Description

The lack of an equality function to compare task arguments can cause some major DX problems.

Using  an object or array literal from the args function will cause the task to continually fire, usually creating an iloop as the task completing causes another update cycle which checks for new arguments:

```ts
  _task = new Task<[Array<string>], Foo>(this, async ([strings]) => { ... }, () => ['a', 'b', 'c']);
```

To avoid this you have to hoist the array if you can, so that the `===` check works:
```ts
const strings = ['a', 'b', 'c'];
  _task = new Task<[Array<string>], Foo>(this, async ([strings]) => { ... }, () => string);
```

It's worse when you have an argument that's a deep object. This can be common if you're using Task to wrap an existing async API that takes such an object as part of its parameters.

Here's a controller that wraps the Google Maps Nearby Places API, which takes a `PlaceSearchRequest` request. We'd like to use `Task` to manage the API calls, so it's natural to ask for a `PlaceSearchRequest` from a user of the controller and pass that as a task arg.

```ts
import {ReactiveController, ReactiveControllerHost} from 'lit';
import {initialState, StatusRenderer, Task} from '@lit-labs/task';

export interface Options {
  getMap: () => google.maps.Map | undefined;
  getRequest: () => Partial<google.maps.places.PlaceSearchRequest> | undefined;
}

export class NearbySearchController implements ReactiveController {
  protected _host: ReactiveControllerHost;

  private _task: Task;

  constructor(host: ReactiveControllerHost, options: Options) {
    (this._host = host).addController(this);
    this._task = new Task<[
      google.maps.Map,
      google.maps.places.PlaceSearchRequest
    ], google.maps.places.PlaceResult[]>(
      host,
      async ([map, request]) => {
        if (request === undefined) {
          return initialState;
        }
        const service = new google.maps.places.PlacesService(map);
        return new Promise((res, rej) =>
          service.nearbySearch(request, (results, status) => {
            if (status === google.maps.places.PlacesServiceStatus.OK) {
              res(results);
            } else {
              rej(status);
            }
          })
        );
      },
      () => {
        const request = options.getRequest();
        return [options.getMap(), request] as [
          google.maps.Map,
          google.maps.places.PlaceSearchRequest
        ];
      }
    );
  }

  render(renderer: StatusRenderer<google.maps.places.PlaceResult>) {
    return this._task.render(renderer);
  }

  hostUpdate() {}
}
```

```ts
_nearbySearchController = new NearbySearchController(this, {
  getMap: () => this.map,
  getRequest: () => ({
    keyword: 'one two'
    location: this.location
  }),
});
```

This construction will cause the Task to run every update as well. The only solution is to destructure the `PlaceSearchRequest` object into literals and many arguments before passing to Task.

I would propose three built-in equality checks:

- Strict (default)
- Shallow
- Deep

Destructuring args into primitives, hoisting const args.

### LINKED ISSUE #3412: [SSR] Add option to shim without window
Many libraries check for the presence of `window` to decide if they are running in the browser or on the server. This can pose a problem when trying to run such a library alongside Lit SSR, because the Lit SSR DOM shim defines `window`, so libraries that would usually be compatible with Node get confused about where they're running and break.

- Example from Sass: https://github.com/lit/lit/issues/2393
- Example from Vue router: https://github.com/lit/lit/blob/main/packages/lit-html/src/is-server.ts

We don't *really* need to define `window` in our DOM shim, because any supported global API (e.g. `customElements.define`) that a Lit component library needs to call for SSR can be referenced without the namespace, or using `globalThis` instead of `window`.

However, it is somewhat common to see e.g. `window.addEventListener` in `connectedCallback`, so removing `window` would break those cases. Ultimately, I think it's a better trade off to not define `window`, and just require those occasional lines to switch to `globalThis`. (We don't actually call `connectedCallback` or shim `addEventListener` currently, but we likely will soon).

In the immediate term, to avoid a breaking change, maybe we should make the non-window defining version opt-in. We could export another function from [dom-shim.js](https://github.com/lit/lit/blob/9f802646d955198cbaf6e521283fe137e7f5b7a6/packages/labs/ssr/src/lib/dom-shim.ts#L201) called e.g. `installDomShimWithoutWindow` -- and then eventually make this the standard behavior.

(Side note: our implementation of `isServer` uses node export conditions instead of checking any globals, which I think would be a great direction for other libraries to move in, since it avoids this problem and also avoids any runtime check: https://github.com/lit/lit/blob/main/packages/lit-html/src/is-server.ts)

steveworkman:
Fixes #3412 

I made an extension to the existing `installWindowOnGlobal` method which defaults to installing `window`. This makes this change non-breaking and you can opt-in to the functionality.

I've tested this with the nuxt3 plugin and it's working well - component authors will need to be exceptionally careful about using `window`, perhaps that analysis can be added to the testing library


packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -23,10 +23,12 @@ import fetch from 'node-fetch';
  *  generally only be true when adding window to a fresh VM context that
  *  starts with nothing.
  * @param props Additional properties to add to the window global
+ * @param exposeWindow Whether the `window` object should be exposed globally
```
augustjk, line 26: The team had a discussion and we are okay making no `window` the only behavior for the DOM shim. It's easy enough for users to add `globalThis.window = globalThis` on their own if they need to so we don't really even need an option.

augustjk (changes requested):
Thank you for the PR! Were you able to get the CLA sorted out? As noted below, we'd like to simply make no `window` the de facto behavior with no need for the option.

Could you also do a `npm run changeset` to note this change? It would be a breaking change so a major bump would make sense. I'd be happy to handle that part too if you'd like.

steveworkman:
> Thank you for the PR! Were you able to get the CLA sorted out? As noted below, we'd like to simply make no `window` the de facto behavior with no need for the option.
> 
> Could you also do a `npm run changeset` to note this change? It would be a breaking change so a major bump would make sense. I'd be happy to handle that part too if you'd like.

Still working on the CLA. I will make the other changes as requested

steveworkman:
PR updated with the new functionality and a changeset. Interestingly, the server and the render-lit tests both required `window` to be exposed as `NODE_MODE` switched to `false` at some points of the test, causing a failure in `reactive-element/css-tag` where `global` would be `undefined` as it was looking for `window`

steveworkman:
I'm still working on the CLA - legal are reviewing it now as there's other people in the org affected too

packages/labs/ssr/src/lib/dom-shim.ts
```diff
@@ -23,6 +23,7 @@ import fetch from 'node-fetch';
...
```
augustjk, line 26: We can drop this annotation too.

augustjk (commented):
@steveworkman Any update on the CLA? If it's looking like more trouble, I can make a separate PR referencing this one for credit if you don't mind. Really appreciate it!

packages/labs/ssr/src/lib/dom-shim.ts
steveworkman: Done

steveworkman:
> @steveworkman Any update on the CLA? If it's looking like more trouble, I can make a separate PR referencing this one for credit if you don't mind. Really appreciate it!

CLA signed, awaiting return from Google 🤞

packages/labs/ssr/src/test/integration/server/server.ts
```diff
@@ -31,8 +31,10 @@ export const ssrMiddleware = () => {
       module = await import(`../tests/${testFile}-ssr.js`);
     } else {
       // mode === 'vm'
+      const window = getWindow({includeJSBuiltIns: true});
+      window.window = window;
```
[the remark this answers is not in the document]
augustjk: With https://github.com/lit/lit/pull/3367 merged, we no longer need the reference to `window` for VM mode test. Let's bring that in and remove these `window.window` assignments.

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -14,11 +14,13 @@ import {RenderInfo} from '../../index.js';
 
 import type * as testModule from '../test-files/render-test-module.js';
 
+const window = getWindow({
+  includeJSBuiltIns: true,
+  props: {require: createRequire(import.meta.url)},
+});
+window.window = window;
```
augustjk, line 21: This can be removed with the latest in main.

augustjk (changes requested):
One last bit of cleanup, https://github.com/lit/lit/pull/3367 just got merged so bringing in the latest from main branch will remove the need for `window` references in the VM module tests.

That'll also fix the breaking tests in the `eleventy-plugin-lit` package's test.

.changeset/loud-radios-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': major
+---
+
+dom shim does not expose the window object
```
aomarks, line 5: Would be nice to add a little more detail here, maybe:

```suggestion
The Lit SSR DOM shim no longer defines a global `window` variable. This was removed to improve compatibility with libraries that detect whether they are running in Node vs the browser by checking for the presence of `window`.

If you have code that runs during SSR which depends on the presence of `window`, you can either replace `window` with `globalThis`, or use `isSsr` to avoid running that code on the server (see https://lit.dev/docs/ssr/authoring/#browser-only-code).
```

steveworkman:
@aomarks @augustjk anything I need to do re: benchmark reporter failure?

augustjk:
> anything I need to do re: benchmark reporter failure?

Nope, all set! Looking at a bit of coordination for the monorepo but I'll work on pressing the merge.

PR #3432, opened 2022-11-07
[labs/ssr] Use renderValue instead of render in LitElementRenderer

### LINKED ISSUE #3417: [labs/ssr] Circular module dependency between `render-lit-html` and `lit-element-renderer`

SSR (@lit-labs/ssr)

### Description

`render-lit-html.js` imports `LitElementRenderer` from `lit-element-renderer.js` for adding to the renderers in `defaultRenderInfo` and `lit-element-renderer.js` imports `render` from `render-lit-html.js` for usage within `LitElementRenderer`'s methods.

This circular dependency causes a reference error when Webpack tries to bundle the code. (This was identified as Next.js uses Webpack to bundle the server code.)

### Reproduction

TODO

### Workaround

Specifically, the `defaultRenderInfo` is declared at the module level which makes it problematic. I think moving that to within the scope of `render()` for which `renderInfo` param is optional, and also exporting and using `renderValue()` for `LitElementRenderer` would decouple those.

### Is this a regression?

### Affected versions

2.2.3

Node: 16

augustjk:
Fixes #3417 

This removes the circular dependency of `render` from `render-lit-html.js` being used in `LitElementRenderer` in `lit-element-renderer.js`, while also depending on `LitElementRenderer` for default param.


augustjk:
Per offline convo with @aomarks, also considering splitting out a third module to remove the circular imports too. Probably would just be the `render` function that gets split out. Also consider using `"exports"` field in packcage.json to make sure `renderValue` is internal only.

packages/labs/ssr/src/lib/render-lit-html.ts
```diff
@@ -527,13 +527,6 @@ export type RenderInfo = {
   deferHydration: boolean;
 };
 
-const defaultRenderInfo = {
```
kevinpschaaf, line 530: Hmm, it still seems like there's a circular dependency between the files; if this change fixed the WebPack issue, it seems likely it's because you moved this module-level variable (whose initializer depended on `LitElementRenderer`) into `render` (rather than depending on `renderValue` vs. `render`?

augustjk: Ah, you're right. Just moving the `defaultRenderInfo` into `render` alone does alleviate the error that webpack was throwing.

If we truly want to remove the circular dependency between files, I believe we do need to splice out `render` into its own module so that `lit-element-renderer` depends on `renderValue`, and `render` depends on both `lit-element-renderer` and `renderValue`.

That would be a breaking change since existing code, including our own example, dig into `lib/render-lit-html.js` for the `render` function. If it's going to be breaking like that, perhaps we might consider calling it `renderToString` or something a bit more descriptive, and also add `"exports"` into package.json to specify what's meant to be pubilc? `render` (`renderToString`?), `renderModule`, and dom shim related things?

packages/labs/ssr/src/lib/render.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
```
aomarks, line 7: jsdoc in the wrong place?

.changeset/funny-bananas-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
```
aomarks, line 2: Should this be minor, since it's breaking?

aomarks (changes requested):

packages/labs/ssr/src/lib/render-lit-html.ts
kevinpschaaf: Seems ok to make the breaking change and update 11ty plugin accordingly.

kevinpschaaf (commented):
LGTM, pending fixes Al's feedback.

packages/labs/ssr/src/lib/render-lit-html.ts
augustjk: Was discussing with Al offline about this. It's really just a file name at this point but `render.js` feels pretty generic. The breaking change feels unnecessary if we just keep `render` function within `render-lit-html.js` and move everything else out to a `render-value.js`. Seems simpler then with no need to update the 11ty plugin either.

.changeset/funny-bananas-deny.md
augustjk: Re-worked such that it is no longer breaking.

aomarks (approved):

justinfagnani:
I'm a little unsure of the module names after this change. `render-value.ts` contains a lit-html rendering function, but it sounds more general than `render-lit-html.ts` which contains a lit-html rendering function. Since it's all Lit, I would propose that `render-lit-html.ts` be renamed to `render.ts`. We can keep `render-lit-html.ts` for a while for back compat.

justinfagnani (approved):

augustjk:
> I'm a little unsure of the module names after this change.

Agreed. My only reason for keeping `render-lit-html` as is was for back compat but it makes more sense to rename. I've done that and kept `render-lit-html` with exports marked as `@deprecated`.

justinfagnani (approved):

Issue #3212, opened 2022-08-12
[labs/react] Add back the reserved React property warning


justinfagnani:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

React (@lit-labs/react)

### Description

https://github.com/lit/lit/pull/2800 removed our warning for reserved React property names because it changed how we detect custom element properties. We should add this back, but only in a dev build, which we'll need to set up for the react package.

### Alternatives and Workarounds

n/a

taylor-vann:
Curious if we can close this issue.

We don't iterate over the element like we didn't with #2800 which doesn't give us the opportunity to warn users about properties except for on component updates.

PR #3449, opened 2022-11-10
[@labs/react] add back dev mode prop warnings

taylor-vann:
Solves https://github.com/lit/lit/issues/3212

In dev mode, warn the user about reserved react properties on a wrapped web component.

Moved util classes and functions into module scope to be available to both dev and prod mode test suites.


packages/labs/react/src/create-component.ts
```diff
@@ -65,6 +65,8 @@ interface Options<I extends HTMLElement, E extends EventNames = {}> {
 
 type Constructor<T> = {new (): T};
 
+const MODE = window?.process?.env?.MODE;
```
augustjk, line 68: We can define `const DEV_MODE = true;` which will get replaced by rollup in the production build.

https://github.com/lit/lit/blob/a6a5591327a90e9c7d922d1d978627794909c422/rollup-common.js#L520

Example of how it's done for `reactive-element`:

https://github.com/lit/lit/blob/a6a5591327a90e9c7d922d1d978627794909c422/packages/reactive-element/src/reactive-element.ts#L39

and use it like
https://github.com/lit/lit/blob/a6a5591327a90e9c7d922d1d978627794909c422/packages/reactive-element/src/reactive-element.ts#L666-L675

So our production build wouldn't even contain code within `if (DEV_MODE) {}` with terser stripping away unused code blocks after replacing the const.

The labs/react package already has export conditions set for development and extends our common rollup plugin so there should be no extra config needed.

packages/labs/react/src/create-component.ts
```diff
@@ -241,6 +243,22 @@ export function createComponent<
     tag = tagName;
   }
 
+  // Warn users when web components use reserved React properties
+  if (MODE === 'dev') {
+    for (const p in element.prototype) {
```
justinfagnani, line 248: Since the `reservedReactProperties size` is much smaller than the set of element properties, lets iterate over that and do an `in` check.

packages/labs/react/src/create-component.ts
```diff
@@ -241,6 +243,22 @@ export function createComponent<
...
+      if (!(p in HTMLElement.prototype) && reservedReactProperties.has(p)) {
+        // Note, this effectively warns only for `ref` since the other
```
justinfagnani, line 250: Good point. I'd be ok just checking for `ref` then and adding a comment to `reservedReactProperties` that says if anything is added there that's _not_ on the HTMLElement.prototype, to add a check for it.

justinfagnani (changes requested):

packages/labs/react/src/create-component.ts
taylor-vann: nice! will do! thanks

taylor-vann: Done!

packages/labs/react/src/create-component.ts
taylor-vann: Done!

packages/labs/react/src/create-component.ts
taylor-vann: I think the warning improvement is something we can do in a future PR :)

packages/labs/react/src/create-component.ts
```diff
@@ -241,6 +243,22 @@ export function createComponent<
     tag = tagName;
   }
 
+  // Warn users when web components use reserved React properties
+  if (DEV_MODE) {
+    for (const p in reservedReactProperties) {
```
augustjk, line 248: Should do a `for ... of ` loop through a set.

```suggestion
    for (const p of reservedReactProperties) {
```

kevinpschaaf (commented):
Looks good, can you add a test, copying the setup used here?

https://github.com/lit/lit/blob/main/packages/reactive-element/src/test/reactive-element_dev_mode_test.ts#L15-L44

packages/labs/react/src/create-component.ts
augustjk: With the condition as I see it now, `if (p in element.prototype && !(p in HTMLElement.prototype))`, I don't think this comment applies anymore.

The comment made sense in its former location where the check against reserved props happened nested after an `in HTMLElement.prototype` check.

Now, the loop is purely for looking up whether any reserved prop is present in the custom element's prototype, sans `HTMLElement`'s prototype, so we are actually going to find things besides `ref` if they are defined as prop on the custom element.

augustjk (changes requested):
Just a minor thing with `of` vs `in`. Looks good to me otherwise.

packages/labs/react/src/create-component.ts
kevinpschaaf: I think practically, everything in `reservedReactProperties` besides `ref` exists on the `HTMLElement.prototype` is the observation, right?

augustjk: That is true. And in the original way the loop and condition was written, we used to loop through `p in elemenet.prototype` and filter out `!(p in HTMLElement.prototype)`, then look for reserved prop such that we would only catch `ref` to `console.warn`.

Now that we loop through the set of reserved props instead, we actually get the opportunity to find other things besides `ref` too. The code view that shows above this thread in the conversation view is out of date.

taylor-vann: Soooo are we good here? I feel like this change is appropriate so we don't have to manually add properties as they come

packages/labs/react/src/create-component.ts
taylor-vann: oh haha yah keys vs values doesn't matter too much in a set huh? will change :)

taylor-vann:
@kevinpschaaf added dev mode tests

packages/labs/react/src/create-component.ts
augustjk: I agree and like the loop through of the list. That makes the comment about only warning for `ref` out of date though so I think the comment can be removed.

taylor-vann: good point!

taylor-vann: Looking more closely, I think keeping the `!(HTMLElement.prototype)` in the check is a benefit.

Some properties like:
- localName
- style
- className
- children

Those will always be on any custom element. Theyre properties on `HTMLElement` as well as a react reserved properties.

Warnings on `HTMLElement` properties might be confusing when all custom elements extend  HTMLElement

kevinpschaaf: Seems fine for now. This is devmode, so we can always change this to a hard-coded list if that ends up being less fussy.

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -55,6 +57,14 @@ class BasicElement extends ReactiveElement {
   @property({type: Boolean})
   disabled = false;
 
+  // override a react reserved property
+  @property({type: Boolean})
+  ref = false;
+
+  // override a react reserved property
+  @property({type: Boolean})
```
kevinpschaaf, line 65: Maybe note in comment here or in the assertion below that we only expect a warning for `ref` and not for `localName` (hence assertion of `length === 1`), since we don't warn on overrides of HTMLElement properties that React treats specially.

kevinpschaaf (approved):

augustjk (approved):

packages/labs/react/src/test/create-component_test.tsx
taylor-vann: done!

.changeset/little-terms-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
+Add console warning for reserved react properites found on a wrapped web component.
```
sorvell, line 5: Mention this is only in dev mode.

sorvell (changes requested):

.changeset/little-terms-develop.md
taylor-vann: done! :)

justinfagnani (approved):

sorvell (approved):

Issue #2765, opened 2022-04-21
[labs/context] ContextRoot will duplicate tracking a context-requesting component which is connected multiple times


benjamind:
## Description
`ContextRoot` will track a context-requesting element (that is not getting a satisfied context-request) multiple times if that element is connected/disconnected within the document multiple times. This isn't a major issue in that it only results in a slight performance degradation when that context is ready to be provided in future (since it will receive the context multiple times). But this could be worth addressing if a context-requesting element is frequently added/removed from the DOM without having a satisfying context provider.

justinfagnani:
I tried fixing this this morning, but it might be impossible to correctly fix without a change to the protocol.

The problem is that it's not possible to distinguish from an element firing multiple context-request events because it connected multiple times from firing multiple events because it has multiple ContextConsumers for the same context.

```ts
class MyElement extends LitElement {
  constructor() {
    new ContextConsumer(this, someContext, (value) => {
        console.log('consumer 1', value);
      }, true);
    new ContextConsumer(this, someContext, (value) => {
        console.log('consumer 12, value);
      }, true);
  }
}
```

For whatever reason this may be done - inheritance, mixins, composed controllers, etc. - we _do_ want to call both callbacks here.

So we can't dedupe by element reference. We could maybe dedupe by both element reference and callback references, but that doesn't work for all cases. Sometimes the callbacks won't be stable:

```ts
class MyElement extends LitElement {
  connectedCallback() {
    super.connectedCallback();
    this.consumer = new ContextConsumer(this, someContext, (value) => {
        console.log('consumer 1', value);
      }, true);
  }
}
```

Here the callback will be new every time the element connects.

So maybe it's ok for those callbacks to fire? This would be similar to how addEventListener() works at least: multiple calls with the same event name and listener are deduped.

kevinpschaaf:
It feels like maybe the protocol should have a concept of `unsubscribe` separate from providing a value. You really want the host to unsubscribe from either the context provider or context root if the element is disconnected. But right now, the only way you get `unsubscribe` is by providing a value, which we wouldn't want the context root to do.

justinfagnani:
@kevinpschaaf yes, that's another issue.

PR #3451, opened 2022-11-11
Make ContextRoot deduplicate context requests by element and callback identity

justinfagnani:
Fixes https://github.com/lit/lit/issues/2765

This deduplicates context requests by element identity and callback identity, so that if an element moves around the DOM, causing it to fire multiple context-request events before a provider is available, its context callback is only called once. This behavior is similar to addEventListener() which will call a callback only once no matter how many times it's added for a single event.

This PR also fixes a potential memory leak by using WeakRefs to hold on to both elements and their context request callbacks. We need WeakRefs because when we receive a context-provider event we need to loop over all element/callback pairs for that context and dispatch events. WeakMaps do not allow iterating over entries.

cc @benjamind

### DIFF
--- a/packages/labs/context/src/lib/context-root.ts
+++ b/packages/labs/context/src/lib/context-root.ts
@@ -17,7 +17,8 @@ export class ContextRoot {
   private pendingContextRequests = new Map<
     Context<unknown, unknown>,
     {
-      // The WeakMap lets us detect if we're seen an element/callback pair yet
+      // The WeakMap lets us detect if we're seen an element/callback pair yet,
+      // without needing to iterate the `requests` array
       callbacks: WeakMap<HTMLElement, WeakSet<ContextCallback<unknown>>>;
 
       // Requests lets us iterate over every element/callback that we need to
--- a/packages/labs/context/src/lib/context-root.ts
+++ b/packages/labs/context/src/lib/context-root.ts
@@ -5,27 +5,23 @@
  */
 
 import {Context} from './create-context.js';
-import {ContextRequest, ContextRequestEvent} from './context-request-event.js';
+import {ContextCallback, ContextRequestEvent} from './context-request-event.js';
 import {ContextProviderEvent} from './controllers/context-provider.js';
 
-type UnknownContextKey = Context<unknown, unknown>;
-
 /**
- * A context request, with associated source element, with all objects as weak references.
- */
-type PendingContextRequest = Omit<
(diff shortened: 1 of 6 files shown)


packages/labs/context/src/lib/controllers/context-consumer.ts
```diff
@@ -48,41 +51,41 @@ export class ContextConsumer<
...
-            // if we don't support subscription, immediately unsubscribe
-            if (!this.subscribe) {
-              this.unsubscribe();
-            }
-          }
+      new ContextRequestEvent(this.context, this._callback, this.subscribe)
+    );
+  }
+
+  // This function must have stable identity to properly dedupe in ContextRoot
```
justinfagnani, line 58: The entire change in this file is moving the context-request event callback to an instance field so that it can be used for the callback identity check in ContextRoot. The callback body is the same.

packages/labs/context/src/test/late-provider_test.ts
```diff
@@ -139,4 +139,108 @@ suite('late context provider', () => {
     // `value` should now be provided
     assert.strictEqual(consumer.value, 1000);
   });
+
+  test('late element with multiple properties', async () => {
```
justinfagnani, line 143: This test verifies that the fix in the PR didn't break elements with multiple providers which have different callback identity but the same element.

kevinpschaaf (commented):
Let's either skip Context tests on IE11 and document that the Context package doesn't work on IE11 / older browsers, or else document needing a WeakRef polyfill

justinfagnani:
@kevinpschaaf PTAL

packages/labs/context/src/lib/context-root.ts
```diff
@@ -5,27 +5,30 @@
...
+ * these requests when new providers which satisfy matching contexts
+ * are available.
  */
 export class ContextRoot {
   private pendingContextRequests = new Map<
-    UnknownContextKey,
-    Set<PendingContextRequest>
+    Context<unknown, unknown>,
+    {
+      // The WeakMap lets us detect if we're seen an element/callback pair yet
```
kevinpschaaf, line 20:
```suggestion
      // The WeakMap lets us detect if we're seen an element/callback pair yet,
      // without needing to iterate the `requests` array
```
The info is in `requests`, it'd just be inefficient to need to loop over the array & deref each time?

kevinpschaaf (approved):
LGTM

PR #3454, opened 2022-11-14
Gitignore build output files from virtualizer

e111077:
there are some build output artifacts that don't seem to be ignored


usergenic (approved):
Oh! Thank you, @e111077 ❤️ LGTM

PR #3455, opened 2022-11-14
[@labs/react] prepare for ssr

taylor-vann:
This PR is about logical real estate in the lit labs react package.

As we incorporate Lit SSR and NextJS, `createComponent` might need to be aware of what environment it is in.

This PR will change. But it'll stay in draft mode until necessary.

Mainly, I want to make sure we associate the `NODE_MODE` with the iteration and update of props entirely. We shouldn't if our way through a bucket iteration, instead we should choose what bucket iteration based on what mode.


augustjk:
Superseded by https://github.com/lit/lit/pull/3885

PR #3457, opened 2022-11-15
Logo dark mode support

e111077:
[Preview](https://github.com/lit/lit/tree/lit-logo-dark-mode) (enable dark mode in GH)

[Main Branch for Comparison](https://github.com/lit/lit)

| Before | After |
| --- | --- |
| image | image |
| image | image |


aomarks (changes requested):
This unfortunately doesn't work when the user sets their preference through the GitHub settings, as opposed to through browser/OS settings. E.g. this is what I get with a light preferred browser, but the GitHub dark theme in my settings:

image

But! There's recently added support for the `<picture>` element, and GitHub handles it in a special way so that it does respect the GitHub settings. Example config (just saw it in @cdata's project :slightly_smiling_face: ): https://github.com/subconsciousnetwork/noosphere/blob/main/README.md?plain=1#L4

aomarks (approved):

Issue #2993, opened 2022-06-06
[analyzer] Remaining features for gen-manifest


kevinpschaaf:
Remaining analyzer features required to generate a `custom-elements.json` manifest:

* [ ] Exports (of Module)
* [ ] FunctionDeclaration
* [ ] MixinDeclaration / CustomElementMixinDeclaration
* [ ] Attributes (of LitElementDeclaration)
* [ ] Slots (for LitElementDeclaration)
* [ ] CSSParts (for LitElementDeclaration)
* [ ] CSSProperties (for LitElementDeclaration)
* [ ] Demos (for LitElementDeclaration)
* [ ] Superclass (for ClassDeclaration)
* [ ] ClassField (for ClassDeclaration) - ReactiveProperty should likely extend this
* [ ] ClassMethod (for ClassDeclaration)
* [ ] Summary and description (from jsdoc) for all model classes
* [ ] Deprecated status for all model classes
* [ ] Source reference for all model classes

justinfagnani:
I added attribute support in #5263

justinfagnani:
Basic mixin support was added in https://github.com/lit/lit/pull/4147

I think that what's left there is cross-package mixin support, as in: https://github.com/lit/lit/issues/4492

PR #3464, opened 2022-11-16
[labs/gen-manifest] Adds `exports` and more metadata to manifest generator

kevinpschaaf:
* **Adds `exports` to manifest generator** - [5d7d3b4](https://github.com/lit/lit/pull/3464/commits/5d7d3b433b2457532e721e2b50494d6cbc67ab56) - Also fixes a few bugs in analyzer logic around symbol renaming on the import side and export side, and adds tests for all of that which was woefully missing.

* **Add `@slot`, `@cssProp`, and `@cssPart` to manifest generator** - b5c3f5a

* **Fix config to make JS program analysis faster.** - dbe8189 - JS program analysis (specifically the call to `getSemanticDiagnostics()` that we do to make sure the program isn't broken before analyzing the AST) was suuuuper slow, as a result of the `allowJs` setting causing TypeScript to automatically include every `.d.ts` file under `node_modules/@types`, which doesn't happen when `allowJs` is false. Seems unlikely a JS project will install types like this, so for now setting `typeRoots: []`, which dramatically speeds up analysis.

* **Add support for parsing description, summary, & deprecated** - d5f49ca

Addresses parts of #2993

### DIFF
--- a/packages/labs/analyzer/src/lib/javascript/variables.ts
+++ b/packages/labs/analyzer/src/lib/javascript/variables.ts
@@ -95,3 +95,37 @@ const getVariableDeclarationInfoList = (
     );
   }
 };
+
+/**
+ * Returns declaration info & factory for a default export assignment.
+ */
+export const getExportAssignmentVariableDeclarationInfo = (
+  exportAssignment: ts.ExportAssignment,
+  analyzer: AnalyzerInterface
+): DeclarationInfo => {
+  return {
+    name: 'default',
+    factory: () =>
+      getExportAssignmentVariableDeclaration(exportAssignment, analyzer),
+    isExport: true,
+  };
+};
+
+/**
+ * Returns an analyzer `VariableDeclaration` model for the given default
+ * ts.ExportAssignment, handling the case of: `export const 'some expression'`;
+ *
+ * Note that even though this technically isn't a VariableDeclaration in
+ * TS, we model it as one since it could unobservably be implemented as
+ * `const varDec = 'some expression'; export {varDec as default} `
+ */
+const getExportAssignmentVariableDeclaration = (
+  exportAssignment: ts.ExportAssignment,
+  analyzer: AnalyzerInterface
+): VariableDeclaration => {
+  return new VariableDeclaration({
+    name: 'default',
+    node: exportAssignment,
+    type: getTypeForNode(exportAssignment.expression, analyzer),
+  });
+};
--- a/packages/labs/analyzer/src/lib/javascript/variables.ts
+++ b/packages/labs/analyzer/src/lib/javascript/variables.ts
@@ -16,7 +16,7 @@ import {
   AnalyzerInterface,
   DeclarationInfo,
 } from '../model.js';
-import {isExport} from '../references.js';
+import {hasExportKeyword} from '../references.js';
 import {DiagnosticsError} from '../errors.js';
 import {getTypeForNode} from '../types.js';
 
@@ -49,8 +49,9 @@ export const getVariableDeclarationInfo = (
   statement: ts.VariableStatement,
   analyzer: AnalyzerInterface
 ): DeclarationInfo[] => {
+  const isExport = hasExportKeyword(statement);
   return statement.declarationList.declarations
-    .map((d) => getVariableDeclarationInfoList(d, d.name, analyzer))
+    .map((d) => getVariableDeclarationInfoList(d, d.name, isExport, analyzer))
     .flat();
 };
 
@@ -62,6 +63,7 @@ export const getVariableDeclarationInfo = (
 const getVariableDeclarationInfoList = (
   dec: ts.VariableDeclaration,
   name: VariableName,
+  isExport: boolean,
   analyzer: AnalyzerInterface
 ): DeclarationInfo[] => {
   if (ts.isIdentifier(name)) {
@@ -69,7 +71,7 @@ const getVariableDeclarationInfoList = (
       {
         name: name.text,
         factory: () => getVariableDeclaration(dec, name, analyzer),
-        isExport: isExport(dec),
+        isExport,
       },
     ];
   } else if (
@@ -82,7 +84,9 @@ const getVariableDeclarationInfoList = (
       ts.isBindingElement(el)
     ) as ts.BindingElement[];
     return els
-      .map((el) => getVariableDeclarationInfoList(dec, el.name, analyzer))
+      .map((el) =>
+        getVariableDeclarationInfoList(dec, el.name, isExport, analyzer)
+      )
       .flat();
   } else {
     throw new DiagnosticsError(
--- a/packages/labs/analyzer/src/lib/lit-element/events.ts
+++ b/packages/labs/analyzer/src/lib/lit-element/events.ts
@@ -16,50 +16,37 @@ import {Event} from '../model.js';
 import {AnalyzerInterface} from '../model.js';
 import {getTypeForJSDocTag} from '../types.js';
 
-import {LitClassDeclaration} from './lit-element.js';
-
 /**
  * Returns an array of analyzer `Event` models for the given
  * ts.ClassDeclaration.
  */
-export const getEvents = (
-  node: LitClassDeclaration,
+export const addEventsToMap = (
+  tag: ts.JSDocTag,
+  events: Map<string, Event>,
   analyzer: AnalyzerInterface
 ) => {
-  const events = new Map<string, Event>();
-  const jsDocTags = ts.getJSDocTags(node);
-  if (jsDocTags !== undefined) {
-    for (const tag of jsDocTags) {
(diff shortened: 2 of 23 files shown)


packages/labs/analyzer/src/lib/javascript/variables.ts
```diff
@@ -49,8 +49,9 @@ export const getVariableDeclarationInfo = (
   statement: ts.VariableStatement,
   analyzer: AnalyzerInterface
 ): DeclarationInfo[] => {
+  const isExport = hasExportKeyword(statement);
```
kevinpschaaf, line 52: We were previously checking for the `export` keyword on the wrong node for variable declarations; it will be on the top-level `VariableDeclarationList` statement, rather than on `VariableDeclaration` (the latter are the individual names in a potential destructuring declaration ala `export const {foo, bar} = obj`.

packages/labs/analyzer/src/lib/references.ts
```diff
@@ -16,10 +16,29 @@ import {AbsolutePath} from './paths.js';
 
 const npmModule = /^(?<package>(@[^/]+\/[^/]+)|[^/]+)\/?(?<module>.*)$/;
 
+/**
+ * Returns a ts.Symbol for a name in scope at a given location in the AST.
+ * TODO(kschaaf): There are ~1748 symbols in scope of a typical hello world,
+ * due to DOM globals. Perf might become an issue here.
+ */
+export const getSymbolForName = (
```
kevinpschaaf, line 24: Just moved this function here from `types.ts`, since it's not type-specific.

packages/labs/analyzer/src/lib/references.ts
```diff
@@ -310,16 +331,14 @@ export const getExportReferences = (
...
             analyzer
           ),
         });
       } else {
         // Get the declaration for this symbol, so we can determine if
         // it was declared locally or not
-        const symbol = analyzer.program
-          .getTypeChecker()
-          .getSymbolAtLocation(el.name);
+        const symbol = getSymbolForName(localName, localNameNode, analyzer);
```
kevinpschaaf, line 341: Given
```ts
import {Foo} from 'foo';
export {Foo as Bar};
```
`localNameNode` will be the `Bar` on the export line, and getting that symbol's declaration will point to `Foo` on the import line, which is what we want (so we can tell it's a re-export).

However, 
```ts
import {Foo} from 'foo';
export {Foo};
```
`localNameNode` will be the `Foo` on the export line, and getting that symbol's declaration will point right back at `Foo` on the _export_ line. Hence we can't get to the actual declaration of it on the import line. I don't know why this is. Using `getSymbolForName` (which searches symbols in scope at this location) lets us get to the declaration on the import line.

packages/labs/analyzer/src/lib/references.ts
```diff
@@ -310,16 +331,17 @@ export const getExportReferences = (
...
         // Get the declaration for this symbol, so we can determine if
-        // it was declared locally or not
-        const symbol = analyzer.program
-          .getTypeChecker()
-          .getSymbolAtLocation(el.name);
+        // it was declared locally or not. Note we use name-based searching
+        // to find the symbol, because `getSymbolAtLocation()` for
+        // `export {Foo}` will annoyingly just point back to the export
+        // line, rather than the location it was actually declared.
+        const symbol = getSymbolForName(localName, localNameNode, analyzer);
```
kevinpschaaf, line 344: Given
```ts
import {Foo} from 'foo';
export {Foo as Bar};
```
`localNameNode` will be the `Bar` on the export line, and getting that symbol's declaration will point to `Foo` on the import line, which is what we want (so we can tell it's a re-export).

However, 
```ts
import {Foo} from 'foo';
export {Foo};
```
`localNameNode` will be the `Foo` on the export line, and getting that symbol's declaration will point right back at `Foo` on the _export_ line. Hence we can't get to the actual declaration of it on the import line. I don't know why this is. Using `getSymbolForName` (which searches symbols in scope at this location) lets us get to the declaration on the import line.

packages/labs/gen-manifest/package.json
```diff
@@ -39,10 +39,11 @@
         "build"
       ],
       "files": [
-        "goldens/**/*"
+        "goldens/**/*",
+        "../test-projects/test-element-a"
```
kevinpschaaf, line 43: Note, the manifest above includes a bunch of new files that were added in #3384. I believe CI didn't flag the goldens as out of date because we weren't including the input files in the `wireit` config, so the github action used the cached result. This should prevent that going forward.

packages/labs/gen-manifest/src/index.ts
```diff
@@ -57,7 +58,10 @@ const convertModule = (module: Module): cem.Module => {
     description: 'TODO', // TODO
     declarations: [...module.declarations.map(convertDeclaration)],
     exports: [
-      // TODO
+      ...module.exportNames.map((name) =>
+        convertJavascriptExport(name, module.getExportReference(name))
+      ),
+      ...module.getCustomElementExports().map(convertCustomElementExport),
```
kevinpschaaf, line 64: See https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts#L83-L156 for details... basically exports contains `js` exports (actual module exports), as well as `custom-element-definition` exports, which are CE's registered in the global registry by tag name.

packages/labs/analyzer/src/lib/lit-element/events.ts
```diff
@@ -16,50 +16,37 @@ import {Event} from '../model.js';
...
+export const addEventsToMap = (
+  tag: ts.JSDocTag,
+  events: Map<string, Event>,
   analyzer: AnalyzerInterface
 ) => {
-  const events = new Map<string, Event>();
-  const jsDocTags = ts.getJSDocTags(node);
-  if (jsDocTags !== undefined) {
-    for (const tag of jsDocTags) {
-      if (tag.tagName.text === 'fires') {
```
kevinpschaaf, line 33: Since we need to parse jsdoc tags for more than events, the tag iteration is moved into `lit-element.ts`, which calls out to `addEventsToMap`

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -176,10 +176,29 @@ export class Module {
...
     const exp = this._exportMap.get(name);
-    if (exp instanceof Reference) {
+    if (exp === undefined) {
+      throw new Error(
+        `Module ${this.sourcePath} did not contain an export named ${name}`
+      );
+    } else if (exp instanceof Reference) {
       return exp;
     } else {
-      return this.getDeclaration(name);
```
kevinpschaaf, line 182: This was a bug; it should have been using `exp` here; when it is a string, it's the name of a local declaration (which may be exported under a different name).

I didn't have good export tests; this PR adds a bunch.

packages/labs/gen-manifest/package.json
```diff
@@ -39,10 +39,11 @@
...
       ],
       "#comment": "The quotes around the file regex must be double quotes on windows!",
-      "command": "uvu test \"_test\\.js$\"",
+      "command": "cross-env NODE_OPTIONS=--enable-source-maps uvu test \"_test\\.js$\"",
```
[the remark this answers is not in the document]
kevinpschaaf: Node added source-map support as a built-in feature now, so you don't have to import `source-map-support/register` in every test module you might debug. `cross-env` makes setting the environment variable on Windows work.

Most of the tests have been updated to do this; ths one was missed.

packages/labs/analyzer/src/lib/references.ts
kevinpschaaf: I think the comment already captures it, albeit more tersely? 
>        // Get the declaration for this symbol, so we can determine if
>        // it was declared locally or not. Note we use name-based searching
>        // to find the symbol, because `getSymbolAtLocation()` for
>        // `export {Foo}` will annoyingly just point back to the export
>        // line, rather than the location it was actually declared.

packages/labs/analyzer/src/test/javascript/exports_test.ts
```diff
@@ -0,0 +1,263 @@
...
+  const test = suite<{
+    analyzer: InMemoryAnalyzer;
+  }>(`Exports tests (${lang})`);
+
+  test.before.each((ctx) => {
+    ctx.analyzer = new InMemoryAnalyzer(lang, {
+      '/package.json': JSON.stringify({name: '@lit-internal/in-memory-test'}),
+    });
+  });
+
```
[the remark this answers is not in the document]
kevinpschaaf: We don't have function analysis yet, so we'll add `export function` when we have that.

Otherwise, added support (and/or tests for):
* export {name as default}
* export default <expression>
* export * from 'module'
* export * as ns from 'module'

sorvell (approved):
Looks good. Thanks for adding the other export cases.

Issue #3461, opened 2022-11-15
git repo build in windows, cannot find module


dbwodlf3:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

When i am building this repo, i met some erros.
I am using windows 10.

https://github.com/lit/lit#contributing-to-lit

I follw the scripts, but i met some error scripts. like this
```
src/commands.ts(23,8): error TS2307: Cannot find module '@lit/localize-tools/lib/config.js' or its corresponding type declarations.
src/commands.ts(24,37): error TS2307: Cannot find module '@lit/localize-tools/lib/modes/transform.js' or its corresponding type declarations.
src/commands.ts(25,35): error TS2307: Cannot find module '@lit/localize-tools/lib/modes/runtime.js' or its corresponding type declarations.
src/commands.ts(26,39): error TS2307: Cannot find module '@lit/localize-tools/lib/error.js' or its corresponding type declarations.
src/commands.ts(27,28): error TS2307: Cannot find module '@lit/localize-tools/lib/index.js' or its corresponding type declarations.
src/commands.ts(28,32): error TS2307: Cannot find module '@lit/localize-tools/lib/typescript.js' or its corresponding type declarations.

...

src/release-image.ts:7:28 - error TS2307: Cannot find module 'changelog-parser' or its corresponding type declarations.
src/release-image.ts:9:22 - error TS2307: Cannot find module 'marked' or its corresponding type declarations.
src/release-image.ts:10:23 - error TS2307: Cannot find module 'puppeteer' or its corresponding type declarations.

```

### Reproduction

```bash
git clone https://github.com/lit/lit.git
cd lit
npm ci
npm run build
```

### Workaround

I have not found a workaround

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

git commit 9b9bdb0e7be218b8d6e4478867361aaf14de49d5

### Browser/OS/Node environment

OS: WIndows10 19044.2130

kevinpschaaf:
Hmmm, our CI on Windows 10 at that commit passed: https://github.com/lit/lit/actions/runs/3466028065/jobs/5789549816

Mind checking the configs in that log and note any differences in your setup?

cc @aomarks in case there's a localize or wireit-specific regression here on Windows?

AndrewJakubowicz:
I have a repro on my Windows machine. There are the errors as indicated in the description, but I believe the root cause command is:

[...]

PR #3466, opened 2022-11-17
[infra] Fix "mkdir -p" breaking builds when on Windows

AndrewJakubowicz:
Fixes https://github.com/lit/lit/issues/3461

### Why?

`mkdir -p` [is not supported on Windows](https://github.com/npm/npm/issues/12774).

The result of this, is that `mkdir` would create a `-p` directory on Windows and then crash when run a second time. Whilst on Linux/Max, `-p` would silence the error if the directory already existed.

There was additionally a syntax error caused on Windows due to lack of quotes around the file paths.

### Fix

Instead of installing another library that is commonly used to solve this issue ([`mkdirp`](https://www.npmjs.com/package/mkdirp)), I opted to use node directly with an inline script which makes the directories.

### Testing

Tested manually on my Windows machine and on Mac. After this change `npm run build` on the monorepo builds successfully.
I also tested running `npm run build:copy-polyfill` multiple times.

### Future work

In the future this can be handled by rollup, and using a [`copy`](https://www.npmjs.com/package/rollup-plugin-copy) plugin. - which doesn't appear to be used in Virtualizer.

This is a quick fix to unblock Windows.


packages/labs/virtualizer/package.json
```diff
@@ -50,7 +50,7 @@
       ]
     },
     "build:copy-polyfill": {
-      "command": "mkdir -p polyfills/resize-observer-polyfill && cp src/polyfills/resize-observer-polyfill/ResizeObserver.js polyfills/resize-observer-polyfill/ResizeObserver.js",
+      "command": "node -e \"require('fs').mkdirSync('polyfills/resize-observer-polyfill', {recursive: true})\" && cp \"src/polyfills/resize-observer-polyfill/ResizeObserver.js\" \"polyfills/resize-observer-polyfill/ResizeObserver.js\"",
```
augustjk, line 53: I don't know how much effort we really want to put in for Windows support but `cp` is not available in Windows when using `cmd.exe`. It is available in PowerShell, and Git Bash though.

It would be more compatible to write a Node script that does it all together.

Probably way extra to bring in additional dependencies for this but there are things like https://www.npmjs.com/package/shx which allows using shell commands inline.

augustjk: Ah, I see you've noted this could be part of rollup. I agree that would be a better long term solution. :)

AndrewJakubowicz: Great point about `cp`. I was using Git Bash on Windows so didn't get impacted by the missing `cp`.

Downgrading this PR back to a draft until this is a more complete solution.

yymm120: 2024, work fine. good.

AndrewJakubowicz:
Closing as this is extremely stale and likely not a problem since nothing much has been commented.

Issue #3219, opened 2022-08-16
[labs/ssr] ElementRenderer async rendering interoperability?


nicholasrice:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

SSR (@lit-labs/ssr)

### Description

Hey! I'm investigating some async SSR scenarios for FAST (which implements its own ElementRenderer and renderer), and at this point I'm unsure how to support async rendering in a way that interoperates with other ElementRenderer / rendering implementations. I saw that @kevinpschaaf took an initial stab at an async rendering implementation here: https://github.com/lit/lit/pull/1690, which adjusts renderShadow to conditionally return promise types. That is what I've been exploring as well, however that implementation circumvents the type system which seems problematic from an interoperability perspective.

Is interoperative async rendering a goal of ElementRenderer and if so, have you considered opening up `ElementRenderer.renderShadow()` and `ElementRenderer.renderLight()` to support promise return types? I understand this would likely degrade the sync rendering scenarios which isn't ideal.

### Alternatives and Workarounds

Are there other ways that ElementRenderer can communicate it should render async, which rendering implementations can *choose* to support? Perhaps a flag that sync renderers could ignore, while async rendering could await until ready? This could keep the ElementRenderer APIs synchronous while offering the rendering process at large the opportunity to be async.

```ts
abstract class ElementRenderer {
	/* ... */
	public readyToRender: true | Promise<true> = true; 
}
```

I'm not sure what the right answer here is, but we've got async rendering scenarios that I'd love to get working in a way that interoperates with other ElementRenderer implementations, so I wanted to start the conversation.

justinfagnani:
I think #1690 is the right direction and we can work on the types to include `string | Promise<string>` on the ElementRenderer interface and have top-level render functions either return `Iterable<string | Promise<string>>` or `Iterable<string>` and throw if they encounter a Promise.

As part of enabling Promises, I think we also want a way of send contextual information to renderers about whether the top-level function can handle Promises. Some renderers might be able to conditionally render sync values, maybe just skipping async work, if the renderer can't handle them.

I don't think adding a Promise to ElementRenderer is sufficient because of the way that the iterables are chained. If a child element is async rendered a parent would have to be too and the parent's promise would depend on its children - otherwise `renderShadow()` would emit Promises when it gets to those async children.

nicholasrice:
[...]

Issue #3118, opened 2022-07-06
[labs/ssr] Demo no longer works after conversion to npm workspaces


augustjk:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

Template shadowroot polyfill and client hydration for demo relied serving static files at the package level. https://github.com/lit/lit/blob/c93a8ee740bcd2f89cecf4213e29949013ad1673/packages/labs/ssr/src/demo/global/server.ts#L18

While the demo file needs to be served from package root, the node_modules has now moved to the root of the monorepo.

### Reproduction

```sh
npm run demo:global
```
and click the "Hydrate" button to see following

<img width="550" alt="image" src="https://user-images.githubusercontent.com/40413829/177655373-9c217210-ab2e-4969-b2a6-418faf9570cb.png">

### Workaround

Set the koa-node-resolve root and the koa-static path to be the monorepo root and provide url relative to monorepo root for app-client.js here https://github.com/lit/lit/blob/c93a8ee740bcd2f89cecf4213e29949013ad1673/packages/labs/ssr/src/demo/global/app-server.ts#L33

Maybe this workaround is enough for demo purposes but might consider creating a separate bundle for the client code without relying on blanket static serving with the node-resolve middleware or we potentially utilizing import maps instead of koa-node-resolve and some targeted serving of packages.

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

N/A

### Browser/OS/Node environment

N/A

augustjk:
Fixed in https://github.com/lit/lit/pull/3522.

We might also consider moving the demos to the `examples` folder after https://github.com/lit/lit/pull/3508 lands.

PR #3467, opened 2022-11-17
Allow SSR renderers to produce asynchronous values

justinfagnani:
This is a breaking change that expands the return types of `render()` and the ElementRenderer methods.

Addresses https://github.com/lit/lit/issues/3219


.changeset/thick-bees-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': major
+---
+
+Allow SSR renderers to produce asynchronous values
```
justinfagnani, line 5: I need to add more information about the breaking change and how to upgrade (use collectResult or collectResultSync)

justinfagnani:
One open question for me I didn't look into yet is how this integrates into a streaming response like Koa. I suspect we'll need a utility that can pipe into a Stream correctly, but we don't seem to have an example of that in the monorepo yet. Adding a test that emulates a Koa environment could be a blocker for this PR.

packages/labs/ssr/src/test/lib/render-result_test.ts
```diff
@@ -0,0 +1,50 @@
...
+});
+
+test('collectResultSync throws for a Promise', () => {
+  assert.throws(
+    () => collectResultSync(['a', Promise.resolve('b'), 'c']),
+    'abc'
+  );
+});
+
+test('collectResult throws for iterables of Promises', async () => {
```
kevinpschaaf, line 27:
```suggestion
test('collectResultSync throws for iterables of Promises', async () => {
```

packages/labs/ssr/src/test/lib/render-result_test.ts
```diff
@@ -0,0 +1,50 @@
...
+  assert.equal(await collectResult(['a', ['b', 'c'], 'd']), 'abcd');
+});
+
+test('collectResult collects strings and Promises', async () => {
+  assert.equal(await collectResult(['a', Promise.resolve('b'), 'c']), 'abc');
+});
+
+test('collectResult collects strings and iterables of Promises', async () => {
+  assert.equal(await collectResult(['a', [Promise.resolve('b')], 'c']), 'abc');
+});
```
kevinpschaaf, line 48: I think we need a test of `['a', Promise.resolve(['b', 'c']), 'd']`, which I think is what a common case of a generator pausing and then resuming would look like?

And e.g. a double-nested one:`['a', Promise.resolve(['b', 'c', Promise.resolve(['d', 'e']), 'f']), 'g']`

kevinpschaaf: Actually, will a promised-value ever be a simple string rather than an iterable? I suppose it doesn't hurt, but I'm not following how we'd get that, since all of our "render" functions generally return iterables, not strings.

kevinpschaaf:
> I suspect we'll need a utility that can pipe into a Stream correctly, but we don't seem to have an example of that in the monorepo yet. 

I think we just need the utility to be an async generator that yields out strings as they resolve that can be fed into `Readable.from()`, ala this thing I did:  https://github.com/lit/lit/pull/1690/files#diff-4e468b31bd6e90e2344bb5e19d58aa208aa697524998f5348bd640f85fa7eca1R11-R23

And you can test that by making a `Readable` and then flushing it out.

justinfagnani:
@kevinpschaaf 

> I think we just need the utility to be an async generator that yields out strings as they resolve

This could be micro-optimization, but because of the overhead we saw with async generators before I would like to not incur the overhead of a promise and microtask for synchronous strings. It just seems like that would diminish or possibly eliminate the benefit we get from using sync generators.

packages/labs/ssr/src/test/lib/render-result_test.ts
justinfagnani: I don't think a `Promise<string>` will not be common, but it's simple enough to support. I could remove it.

justinfagnani: Added tests and fixed the bugs!

justinfagnani:
I added a RenderResultReadable that consumes a RenderResult. It should be about as optimal as we can get:

- It uses a stack instead of recursion
- It'll loop through sync values as much as possible, only returning from _read() when the underlying Readable pushes back with `this.push()` returning `false`
- Almost no async overhead when consuming sync values: it doesn't use async functions, but then()s Promises only when encountered.

This should be usable with server frameworks like Koa like:
```ts
app.use('/', async (context) => {
  const result = render(...);
  context.body = new RenderResultReadable(result);
}
```

(Koa automatically pipes the body to the response when it's a Readable).

packages/labs/ssr/src/lib/render-result.ts
```diff
@@ -0,0 +1,82 @@
...
+      // the number of `await`s by only awaiting Promises and not having
+      // to await recursive calls.
+      value = await collectResult(chunk as RenderResult, value);
+    } else {
+      // Must be a Promise
+      chunk = await chunk;
+      if (typeof chunk === 'string') {
+        value += chunk;
+      } else {
+        value = await collectResult(chunk as RenderResult, value);
```
aomarks, line 49:
```suggestion
        value = await collectResult(chunk, value);
```

This also should have been narrowed by typescript, no?

packages/labs/ssr/src/lib/render-result.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+  return value;
+};
+
+/**
+ * Joins a RenderResult into a string synchronously.
+ *
+ * This function throws if a RenderResult contains a Promise.
+ */
+export const collectResultSync = (result: RenderResult, initialValue = '') => {
```
aomarks, line 61:
```suggestion
export const collectResultSync = (result: RenderResult, initialValue = ''): string => {
```

Again, would like to see the return type explicitly for self-documentation.

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,143 @@
...
+ * A Readable that reads from a RenderResult.
+ */
+export class RenderResultReadable extends Readable {
+  private _result: RenderResult;
+
+  /**
+   * A stack of open iterators.
+   *
+   * We need to keep this as instance state because we can pause and resume
+   * reading values at any time and can't run guarentee to run iterators to
```
aomarks, line 24: Something typo'd in this sentence, but not exactly sure what it should be ("cant run guarantee to run")

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,143 @@
...
+    // - _read() should call `this.push()` as many times as it can until
+    //   `this.push()` returns false, which means the underlying Readable
+    //   does not want any more values.
+    // - `this._read()` should not be called by the underlying Readable until
+    //   after this.push() has returned false, so we can wait on a Promise
+    //   and call _read() when it resolves to continue without a race condition.
+    //   (We try to verify this with the this._waiting field)
+    // - `this.push(null)` ends the stream
+    //
+    // RenderResult is a iterable recursive type that can contain other
```
aomarks, line 63:
```suggestion
    // RenderResult is an iterable recursive type that can contain other
```

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,143 @@
...
+    // RenderResults. To accomodate this we keep a stack of open iterators
```
aomarks, line 64:
```suggestion
    // RenderResults. To accommodate this we keep a stack of open iterators
```

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+          // Restore the outer iterator
+          iterator = this._iterators.pop()!;
+          continue;
+        }
+      }
+      const value = next.value as
+        | string
+        | RenderResult
+        | Promise<string | RenderResult>;
```
aomarks, line 91:
```suggestion
      const value = next.value;
```

Is this cast required? Seems like TypeScript should already know this given the signature of `this._iterators`?

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,143 @@
...
+            }
+            this._waiting = false;
+            if (typeof value === 'string') {
+              const continue_ = this.push(value);
+              if (continue_ !== false) {
+                // Continue reading by recursing
+                this._read(size);
+              }
+            } else if (isIterable(value)) {
+              // Stash the new iterator so we can retreive it when we recurse.
```
aomarks, line 123:
```suggestion
              // Stash the new iterator so we can retrieve it when we recurse.
```

aomarks (changes requested):
Awesome! A bunch of small comments.

justinfagnani:
I added https://github.com/lit/lit.dev/pull/1005 for the corresponding lit.dev update.

packages/labs/ssr/src/lib/element-renderer.ts
```diff
@@ -119,21 +120,19 @@ export abstract class ElementRenderer {
...
-  abstract renderShadow(
-    _renderInfo: RenderInfo
-  ): IterableIterator<string> | undefined;
+  abstract renderShadow(_renderInfo: RenderInfo): RenderResult | undefined;
 
   /**
    * Render an element's light DOM children.
    */
-  abstract renderLight(renderInfo: RenderInfo): IterableIterator<string>;
+  abstract renderLight(renderInfo: RenderInfo): RenderResult | undefined;
```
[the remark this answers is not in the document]
aomarks: Ping

aomarks (approved):

packages/labs/ssr/src/lib/element-renderer.ts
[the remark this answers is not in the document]
justinfagnani: oops, forgot to send my comments

augustjk (approved):
Nice! We should update the koa demo to use the `RenderResultReadable`. There's already some maintenance needed for the demo https://github.com/lit/lit/issues/3118 so could be part of that? Maybe the readme update/cleanup should be a separate task so we don't duplicate work across updating readme and the docs on lit.dev.

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -0,0 +1,140 @@
...
+ * A Readable that reads from a RenderResult.
+ */
+export class RenderResultReadable extends Readable {
+  private _result: RenderResult;
+
+  /**
+   * A stack of open iterators.
+   *
+   * We need to keep this as instance state because we can pause and resume
+   * reading values at any time and can't guarentee to run iterators to
```
augustjk, line 24:
```suggestion
   * reading values at any time and can't guarantee to run iterators to
```

justinfagnani:
@augustjk 

> We should update the koa demo to use the RenderResultReadable

Ah, I forgot this was in the monorepo already. I'll update it. I kind of want things like this to be in the /examples folder I was talking about a few eng meetings ago.

packages/labs/ssr/src/lib/render-result.ts
```diff
@@ -0,0 +1,85 @@
...
+ * cause additional overhead compared to a sync iterator.
+ *
+ * Consumers should check the type of each value emitted by the iterator, and
+ * if it is a Promise await it if possible, or throw an error.
+ *
+ * The utility functions {@link collectRenderResult} and
+ * {@link collectRenderResultSync} do this for you.
+ */
+export type RenderResult = Iterable<
+  string | RenderResult | Promise<string | RenderResult>
```
kevinpschaaf, line 22: I do think this type is simpler as 
```ts
type RenderResult = Iterable<string | Promise<RenderResult>>
```
If we don't know where we'd need the `Iterable<RenderResult>` part, I'd prefer not complicating the type and the runtime checks with it. If it's just a matter of letting someone `yield someIterableFn()` rather than `yield *someIterableFn()`, seems better for the type to require just always `yield *` than require the extra check in the hot loop.

If there is some pattern this unlocks that I'm not considering, I'd agree we should make the change now. Not clear what that is.

justinfagnani: Reading up on generator / async generator perf it does seem that in async generators at least, `yield *` is very costly. It seems possible, even likely, that `yield *` is more clostly than yielding a nestest iterable for sync iterators too, though I definitely have not benchmarked that (and it would have to be compared against the duck-type check for `[Symbol.iterator]`)

justinfagnani:
I simplified RenderResult as requested, and the implementations resulting from that. Worth a new review from everyone.

aomarks (approved):

augustjk (approved):
Oops, changing base branch back to main produced a conflict to be resolved. New changes look good!

kevinpschaaf (approved):
LGTM

PR #3183, opened 2022-07-26
[labs/virtualizer] Revise scroll-to behavior and API

graynorton:

PR #3489, opened 2022-12-02
[labs/virtualizer] add scrollToIndex shim for SemVer backwards compatibility to 0.x

usergenic:
Added an equivalent `LitVirtualizer#scrollToIndex()` method to preserve the API that existed prior to latest merge of #3183 which removed it.


graynorton (approved):
LGTM

PR #3490, opened 2022-12-02
[labs/virtualizer] Added missing ScrollerController file info to build:ts output.

usergenic:
Tests were failing due to failure to serve requested dynamic imports.  This is attempting to fix that by adding missing files to package.json's wireit details.


augustjk (approved):
Nice!

PR #3494, opened 2022-12-05
[labs/virtualizer] Accept assigned slots as parent elements

### LINKED ISSUE #3493: [labs/virtualizer] Doesn't update on scroll when a clipping ancestor is in Shadow DOM

### Description

When slotting places a Virtualizer within a scrolling parent, the current parent gathering workflow does not acquire that parent. This means that when the content is scrolled, the virtualization is not triggered and subsequent pages of content are not rendered.

### Reproduction

https://lit.dev/playground/#gist=d12a3422f91b2b55eee102a491b74e3d

### Workaround

Patch the package...

### Is this a regression?

### Affected versions

All

All

Westbrook:
Adds `assignedSlot` to the `getParentElement` workflow to ensure that shadow root content that has scrolling parents are included in the listening process.

fixes #3493

### DIFF
--- a/packages/labs/virtualizer/src/Virtualizer.ts
+++ b/packages/labs/virtualizer/src/Virtualizer.ts
@@ -872,6 +872,9 @@ function getMarginValue(value: string): number {
 
 // TODO (graynorton): Deal with iframes?
 function getParentElement(el: Element) {
+  if (el.assignedSlot !== null) {
+    return el.assignedSlot;
+  }
   if (el.parentElement !== null) {
     return el.parentElement;
   }


graynorton:
@Westbrook I just created two simpler repro cases, which I linked over in #3493. To get this merged, we'd just want to add automated tests matching these simple cases.

If you want to take a whack on that yourself in this branch, that would be ideal (and the fastest way to land the fix). Otherwise, we can make a branch of our own, incorporating your fix and adding the tests. We could get to it fairly soon, I think. Let us know what you prefer.

Westbrook:
I’ll be traveling this next week, so if you could get to it, that would be great. If not, I can revisit this on the 19.

graynorton:
Closing this in favor of #3525, which takes @Westbrook's fix and adds tests.

PR #3495, opened 2022-12-06
[lit-html, reactive-element] Add development node builds

augustjk:
While working on SSR integration, it was found that dev mode was using the unminified development build of `lit-element` but using the minified Node build of `reactive-element` causing error due to the mismatching method names.

This is solved by adding a development Node build which is unminified and will be used in these cases.

### DIFF
--- a/.changeset/light-dryers-talk.md
+++ b/.changeset/light-dryers-talk.md
@@ -1,0 +1,7 @@
+---
+'lit-html': minor
+'@lit/reactive-element': minor
+'lit': patch
+---
+
+`lit-html` and `reactive-element` now include development Node builds with unminified code and dev warnings.
--- a/rollup-common.js
+++ b/rollup-common.js
@@ -454,7 +454,7 @@ export function litProdConfig({
                 },
               }),
               sourcemaps(),
-              // We want the Node build to be minified because:
+              // We want the production Node build to be minified because:
               //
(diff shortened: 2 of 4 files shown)


.changeset/light-dryers-talk.md
```diff
@@ -0,0 +1,7 @@
+---
+'lit-html': minor
+'@lit/reactive-element': minor
+'lit': patch
```
aomarks, line 4: Shouldn't `lit` be minor too or no?

rollup-common.js
```diff
@@ -474,6 +474,36 @@ export function litProdConfig({
               ...(CHECKSIZE ? [skipBundleOutput] : []),
             ],
           },
+          {
+            // Also create a development Node build that does not minify to be
+            // used during devlopment so it can work along side the unminified
```
aomarks, line 479:
```suggestion
            // used during development so it can work along side the unminified
```

aomarks (approved):

.changeset/light-dryers-talk.md
augustjk: I was a bit torn about it.. It's not changing the `lit` package directly, only its deps, but since we consider these deps core to `lit`, I think minor does make more sense.

aomarks (approved):

PR #3508, opened 2022-12-08
[examples] Add nextjs-lit example

### LINKED ISSUE #3012: [labs/ssr] Create a smoke test repo of NextJS SSR'ing Lit
Building on 
* https://github.com/lit/lit/commit/adbfceb02f7deb9cf69d8f44ff88002d851e7788#diff-c9ffebf7c892cdb0442bf316bbd8a4d4be892cca3cd6d0b1d4d9bdc8a39f4fd6
* https://github.com/kevinpschaaf/next-wc-ssr/blob/main/pages/index.js

augustjk:
Resolves #3012

This PR adds an `examples` folder to the root of the monorepo. An initial `nextjs-lit` demo project is present with a barebones Next.js project that renders a Lit component, as well as the component wrapped with `@lit-labs/react`.

All packages in the `examples` folder will not participate in the workspaces which as several advantages:
- `npm ci` from the monorepo root will not install dependencies for these examples
- The dependencies stay within each example's project's `node_modules` directory and doesn't get hoisted which can be problematic for some framework code like Next.js.
- Packages internal to the monorepo can still be referenced with file path like `"lit": "file:../packages/lit"` (It will need to be built)

Changes made to repo settings
- `lint` script and the command for `lint-stage` only affect files within `packages/**.*.{js,ts}` as the Next.js projected had its own eslintrc.

Note: To also make the eslint extension for VS Code pick up the correct config, I needed to add the following config.
```json
// .vscode/settings.json
{
  "eslint.workingDirectories": [{ "mode": "auto" }]
}
```

The example project was bootstrapped by running `npx create-next-app --use-npm`. I'm not sure what the policy for adding license headers to files that were created by this tool and then modified.

I'd also welcome any suggestions for improvements to the components to show case. In particular, I feel the wrapped component example probably would shine more for something that needs event handler mapping.

If this examples folder pattern looks good, we can move the demos for labs/ssr and labs/eleventy-plugin-lit, and localize out here.


package.json
```diff
@@ -19,10 +19,10 @@
     "build:ts": "wireit",
     "build:lit-starter-ts": "wireit",
     "format": "npm run format:eslint && npm run format:prettier",
-    "format:eslint": "eslint \"**/*.{js,ts}\" --fix",
+    "format:eslint": "eslint \"packages/**/*.{js,ts}\" --fix",
```
justinfagnani, line 22: This is a change in that linting is currently handling top-level config files and files in scripts/.

I would add scripts and file an issue to mode rollup-common.js into there or some other config folder. We also have packages/internal-scripts

justinfagnani (changes requested):
Nice start on examples!

Commenting on just the examples folder setup for now:

I would like to see the description and rationale of the /examples/ layout put into a /examples/README.md file

I'd also like to see some script to install, build, and ideally test, all examples using the monorepo packages. Maybe that doesn't have to land in this PR, but some discussion of how to achieve that and an open issue would be good.

examples/nextjs-lit/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "nextjs-lit",
```
aomarks, line 2:
```suggestion
  "name": "@lit-internal/examples-nextjs",
```

For security we should avoid ever having packages, even marked internal, that we don't own (or own the namespace for, so that nobody else can take it).

aomarks (approved):

package.json
augustjk: I reverted back to the same glob `**/*.{js,ts}` and instead added option `--ignore-pattern=examples/` to the command so this'll still work the same as before for top level config files and scripts/. WDYT?

examples/nextjs-lit/package.json
augustjk: Done!

aomarks (approved):

package.json
justinfagnani: sounds great!

examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+  of dependencies that are also used elsewhere in the monorepo.
+- Packages internal to the monorepo can still be referenced by file path like
+  `"lit": "file:../packages/lit"`
+
+Note: For manually testing changes to any package via the example project, the
+recommended way of installing dependencies within the example project is to do
+it with the [install-links]
+(https://docs.npmjs.com/cli/v9/commands/npm-install#install-links) option set to
+false. This is the default behavior for npm v8. However, if you're using npm v9,
+you can set it by running `npm i --install-links=false` within the example
```
justinfagnani, line 27: can we make an [`.npmrc`](https://docs.npmjs.com/cli/v9/configuring-npm/npmrc) file to make this the default in examples?

examples/nextjs-lit/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@lit-internal/examples-nextjs",
```
justinfagnani, line 2: We might want to grab `@lit-examples` so we can name the packages directly after the folder names, like `@lit-examples/nextjs-lit`

examples/nextjs-lit/README.md
```diff
@@ -0,0 +1,8 @@
+# Lit in Next.js example
```
justinfagnani, line 1: Do we want to name this folder `nextjs-lit` or just `nextjs`? We might end up having `-lit` on a lot of things.

examples/nextjs-lit/src/wrapped-simple-greeter.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+import {createComponent} from '@lit-labs/react';
+import {SimpleGreeter} from './simple-greeter';
+
+export default createComponent({
```
justinfagnani, line 11: "wrapped" is a bit of a generic name for me. I would name the module `simple-greeter-react` and at the import site just import it as `SimpleGreeter` (or maybe `SimpleGreeterReact` if you want to make it more clear the difference between it and `<simple-greeter>`).

examples/nextjs-lit/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
justinfagnani, line 3: Can this be newer in a Next app? We really don't want to compile out classes. And I don't see the custom-elements-es5-shim added anywhere... so is this setting even being used?

examples/README.md
augustjk: good idea, done!

examples/nextjs-lit/package.json
augustjk: done. renamed to `@lit-examples/nextjs` dropping `-lit` as per comment below.

examples/nextjs-lit/README.md
augustjk: done

examples/nextjs-lit/src/wrapped-simple-greeter.ts
augustjk: sounds good, done.

justinfagnani (approved):

PR #3514, opened 2022-12-09
Update existing XLIFF files

### LINKED ISSUE #1879: [localize] XLIFF serializer should update instead of replace
When we write out XLIFF XML files after extraction, we always generate a completely fresh XML document from our internal data model.

However, this means any extra XLIFF nodes/attributes that may have been added to the document by the translation process (e.g. `state="needs-translation"`) but that isn't part of our internal data model will be lost.

So, when there is an existing XLIFF document, we should update it rather than replace it to preserve these external changes.

### LINKED ISSUE #3439: [localize] Preserve target state when running lit-localize extract (XLIFF format)
### Should this be an RFC?

Localize (@lit/localize)

### Description

It would be useful to preserve the state of the target elements in the generated XLIFF files. This makes it easier to keep track of the stuff that really needs a new translation. Over the lifetime of a product many new messages are added, and the translators should immediately see what they need to do in a concrete iteration. While I have not performed a diligent market analyisis, I expect many tools that support the XLIFF file format to support the [`state`](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#state) attribute of a target. Thus it would be helpful if the lit-localize CLI tool could respect it.

It is possible to have a manual workflow for creating a todo list for the translators, but this is error-prone.

enqueue:
`lit-localize` updates existing XLIFF translations files, if available, instead of re-generating these files from scratch.

Remarks about this change:

All `note` elements that result from a `lit-localize` run, will receive a `from="lit-localize"` attribute (see [from](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#from). This is necessary to identify the note we are allowed to update in a subsequent run. Other tools might have added further `note` elements. However, this also means, that any existing XLIFF files will be changed once this version of `lit-localize` is run.

We could invest a lot more time into trying to preserve indentation. Other tools might reformat the whole file using tabs or five spaces or... Guessing the most probable level of indentation for new elements, based on surrounding elements, might be doable, but I hope that such an effort is not called for yet.

I adjusted the existing test case, but please feel free to suggest more scenarios we need to cover.

Fixes #1879
Fixes #3439


aomarks (changes requested):
This is fantastic, thank you for the PR! Code and test updates look great to me.

The only missing thing is a changeset file, which is how we manage releases.

If you run `npx changeset add` from the root of the monorepo, then a wizard should walk you through what to do:
- You'll want to mark `@lit/localize-tools` as having a `patch` bump (would usually be `minor`, but we're still on major version `0`).
- I believe there are two changes, you could write something like this (or exactly this):
  ```
  - Existing XLIFF files will be updated instead of re-generated, so additional data added by other processes (such as `state="translated"` attributes) will be preserved instead of deleted.
  - XLIFF `<note>` elements generated by Lit Localize will now have a `from="lit-localize"` attribute to distinguish them from other notes.
  ```

aomarks (changes requested):
Also looks like there are some test failures:

https://github.com/lit/lit/actions/runs/3655389797/jobs/6227289104

enqueue:
I added the changeset as suggested.

Adjusted the code to make the tests work. I was sure I had run them before committing the change (`npm run test` in the localize-tools packages), but perhaps I was looking at a cached result. Hope this will work fine now.

aomarks (approved):

aomarks (approved):
Great work, thank you!

aomarks:
This PR was just released in `@lit/localize-tools@0.6.6`: https://github.com/lit/lit/pull/3556 Thanks again!

Issue #3518, opened 2022-12-13
[labs/virtualizer] New layoutComplete promise created instead of using existing one


graynorton:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

When `layoutComplete` is accessed multiple times for a given layout cycle, the same promise should be returned. It is not.

### Reproduction

1. Run [this Playground example](https://lit.dev/playground/#gist=4894522864a087ffdafa517cef22a430)

Expected result: Console should log `SAME? true`
Actual result: Logs `SAME? false`

### Workaround

If you can, avoid accessing `layoutComplete` multiple times within a given layout cycle

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

1.0.0

### Browser/OS/Node environment

Not environment-specific

PR #3519, opened 2022-12-13
[labs/virtualizer] Fix #3481 and #3518

### LINKED ISSUE #3481: [labs/virtualizer] Error when immediately re-rendering (before async connect code runs)

### Description

When using the latest `0.0.0-masonry-preview-20221107225026` preview, the virtualizer throws an error on initial render. It still seems to work after that.

The error:
```
Virtualizer.js:168 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'detach')
    at Virtualizer.disconnected (Virtualizer.js:168:1)
    at VirtualizeDirective._makeVirtualizer (virtualize.js:74:1)
    at VirtualizeDirective._updateVirtualizerConfig (virtualize.js:59:1)
    at VirtualizeDirective.update (virtualize.js:48:1)
    at VirtualizeDirective._$resolve (directive.ts:134:1)
    at resolveDirective (lit-html.ts:1075:1)
    at ChildPart._$setValue (lit-html.ts:1353:1)
    at TemplateInstance._update (lit-html.ts:1181:1)
    at ChildPart._commitTemplateResult (lit-html.ts:1525:1)
    at ChildPart._$setValue (lit-html.ts:1375:1)
```

This is with the following markup:

```
        <tbody @visibilityChanged=${this.#onVisibilityChanged}>
          ${virtualize({ items: this.items, renderItem: item => this.renderItem(item) })}
        </tbody>
```

This does not throw any error with `0.7.2`

### Reproduction

-

### Workaround

-

### Is this a regression?

### Affected versions

Failing in 0.0.0-masonry-preview-20221107225026, used to work in 0.7.2

-

graynorton:
Fix #3481 and #3518

### DIFF
--- a/packages/labs/virtualizer/src/Virtualizer.ts
+++ b/packages/labs/virtualizer/src/Virtualizer.ts
@@ -20,7 +20,11 @@ import {
   ScrollToCoordinates,
   BaseLayoutConfig,
 } from './layouts/shared/Layout.js';
-import {RangeChangedEvent, VisibilityChangedEvent, UnpinnedEvent} from './events.js';
+import {
+  RangeChangedEvent,
+  VisibilityChangedEvent,
+  UnpinnedEvent,
+} from './events.js';
 import {ScrollerController} from './ScrollerController.js';
 
 export const virtualizerRef = Symbol('virtualizerRef');
@@ -299,10 +303,10 @@ export class Virtualizer {
     );
     this._scrollEventListeners = [];
     this._clippingAncestors = [];
-    this._scrollerController = this._scrollerController!.detach(this);
-    this._mutationObserver!.disconnect();
-    this._hostElementRO!.disconnect();
-    this._childrenRO!.disconnect();
+    this._scrollerController = this._scrollerController?.detach(this) || null;
+    this._mutationObserver?.disconnect();
+    this._hostElementRO?.disconnect();
+    this._childrenRO?.disconnect();
     this._rejectLayoutCompletePromise('disconnected');
   }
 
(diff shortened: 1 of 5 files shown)


packages/labs/virtualizer/src/Virtualizer.ts
```diff
@@ -764,42 +776,44 @@ export class Virtualizer {
...
+
   private _scheduleLayoutComplete() {
-    if (this._pendingLayoutComplete !== null) {
-      cancelAnimationFrame(this._pendingLayoutComplete);
+    // Don't do anything unless we have a pending promise
+    // And only request a frame if we haven't already done so
+    if (this._layoutCompletePromise && this._pendingLayoutComplete === null) {
+      // Wait one additional frame to be sure the layout is stable
+      this._pendingLayoutComplete = requestAnimationFrame(() =>
+        requestAnimationFrame(() => this._resolveLayoutCompletePromise())
```
usergenic, line 803: Interesting! I like the locality/cleanliness of expressing the extra nesting RAF here.  It appears to mean that resolving the layout complete promise is no longer deferrable past the second RAF, however, since `this._resolveLayoutCompletePromise()`.  I am trying to contrive a scenario where that would pose a risk...

Considering a case where _childrenSizeChanged() is called due to observation and that triggers a subsequent scheduleLayoutComplete but the subsequent one doesn't wait yet-another two animation frames--  I'll tell you what I'll make an issue to add more cases to the layout complete tests to verify expectations around cascading resize effects and the layoutComplete promise as a separate issue and approve this as is.

usergenic (approved):
LGTM

packages/labs/virtualizer/src/Virtualizer.ts
graynorton: Yeah, a full set of `layoutComplete` tests would be great. I *think* any given cycle of child measurements and re-layouts is going to occur within a single frame, so believe this logic is ok—but we should certainly confirm via tests.

PR #3522, opened 2022-12-14
[SSR/Core] Add @lit-labs/ssr-dom-shim package

aomarks:
- Adds a new `@lit-labs/ssr-dom-shim` package which exports `Element`, `HTMLElement`, `CustomElementRegistry`, and a default global `customElements` (but does not register any of them as globals).

- Updates `@lit/reactive-element/reactive-element.js` to import from this package in its node build, instead of providing inline no-op versions. This means the parts of the DOM shim required for most Lit components will always be loaded in Node, so that users don't need to think about loading the DOM shim at all anymore.

- When using this new automatic shimming by importing `reactive-element` from Node, only a single global will now be defined: `customElements`. So in particular, `window` will not be defined, solving many issues we encountered with libraries sniffing for `window` to switch between node vs browser modes.

- `ModuleLoader` now provides a default VM global context object which provides basic globals that are available in both Node and browsers. A context object can also be directly created with a new `makeDefaultContextObject` function.

- Updates `lit-html/experimental-hydrate.js` to not depend on `btoa` when running in Node, instead using Node's built-in `Buffer`.

- Updates the existing `@lit-labs/ssr/lib/dom-shim.js` module to import `HTMLElement` and `customElements` from the new package, so that the same base class and registry are used when the full DOM shim is loaded.

- Makes registering a duplicate element in the DOM shim throw an exception.

- Misc minor improvements to the DOM shim code, including moving APIs from `HTMLElement` to `Element`, which is the correct interface for them.

- Updates to the following packages to remove usage of the global DOM shim in code/tests/READMEs:
  - `@lit-labs/testing`
  - `@lit-labs/eleventy-plugin-lit`
  - `@lit-labs/ssr`
  - `@lit/localize-tools`

See also https://github.com/lit/lit.dev/pull/1026 which updates the lit.dev docs for these changes.

Fixes https://github.com/lit/lit/issues/3523


packages/labs/ssr-dom-shim/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@lit-labs/ssr-dom-shim",
+  "version": "1.0.0",
```
augustjk, line 3: Are we good with starting with 1.0.0? For most labs packages we've been starting with 0.1.0.

packages/labs/ssr-dom-shim/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "type": "git",
+    "url": "https://github.com/lit/lit.git",
+    "directory": "packages/labs/ssr-dom-shim"
+  },
+  "main": "index.js",
+  "typings": "index.d.ts",
+  "type": "module",
+  "exports": {
+    ".": {
+      "default": "./index.js"
```
augustjk, line 21: I believe we need to add types here too.
```suggestion
      "types": "./index.d.ts",
      "default": "./index.js"
```

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
augustjk, line 3: Should the date reflect when the file was made or when the code was written? If latter can ignore.
```suggestion
 * Copyright 2022 Google LLC
```

rollup-common.js
```diff
@@ -453,6 +454,13 @@ export function litProdConfig({
                     'const ENABLE_SHADYDOM_NOPATCH = false',
                 },
               }),
+              inject({
+                HTMLElement: ['@lit-labs/ssr-dom-shim', 'HTMLElement'],
+                customElements: ['@lit-labs/ssr-dom-shim', 'customElements'],
+                include: [
+                  '**/packages/reactive-element/development/reactive-element.js',
+                ],
+              }),
```
augustjk, line 463: Inject plugin is a fantastic solution to providing global only for the node builds! We do need this also for the dev node build below, ~line 507.

augustjk (commented):
This is great! Some comments below.

rollup-common.js
aomarks: Done

packages/labs/ssr-dom-shim/src/index.ts
aomarks: From go/copyright:

```
the year the file was first published in full numeric form (e.g., 2010);
don't change this if you edit, move, or copy the file.
```

I consider this file to have been copied, and then edited, so I think the year stays the same.

packages/labs/ssr-dom-shim/package.json
aomarks: Done

packages/labs/ssr-dom-shim/package.json
aomarks: I think we should just start with `1.0.0`, especially for Labs packages, because the "may change frequently" bit (which is what `0` major kinda means) is already implicit in the fact that this is a Labs package. And that allows us to make future updates with full major/minor/patch distinctions, instead of having to collapse minor/patch as we do when the major is 0.

aomarks (commented):
Comments addressed.

Also addressed two comments from @kevinpschaaf that were made in an in-person conversation:

- `reactive-element.js` in Node now falls back to `globalThis.HTMLElement` if one has been defined, which matches the behavior we had before, and allows users to use alternative DOM shims if they want.

- The existing `getWindow` function from the SSR package now behaves like it did before, in that it returns a `new CustomElementRegistry` for each object, instead of using a global singleton. This is important for VM modules that are using `getWindow`, since otherwise there would be a shared CE registry between different VM render contexts.

packages/labs/ssr-dom-shim/package.json
aomarks: Oops, actually even given my explanation above, we do actually want to start from `0.0.1` or similar, because we'll want the first changeset entry to be a `major` bump, so that `1.0.0` is what actually gets published.

aomarks:
Ah, seeing failures relating to code that looks at the TypeScript AST. Now I remember why we did the weird temporary-global-assignment trick previously. Though that won't work here, because of the shadowing we have due to the import. Will need to investigate another pattern.

```
   FAIL  legacy class field emit  "@localized"
    Expected values to be deeply equal:  (equal)

        ··[
        Actual:
        --··"../reactive-element/reactive-element.d.ts(181,55):·error·TS2749:·'ReactiveElement_base'·refers·to·a·value,·but·is·being·used·as·a·type·here.·Did·you·mean·'typeof·ReactiveElement_base'?\n",
        ··]
```

augustjk (approved):
Amazing! 🎉

kevinpschaaf (approved):

packages/labs/ssr-dom-shim/.gitignore
```diff
@@ -0,0 +1,3 @@
+/node_modules/
```
justinfagnani, line 1: isn't node_modules/ globally ignored?

packages/labs/ssr-dom-shim/package.json
```diff
@@ -0,0 +1,51 @@
+{
```
justinfagnani, line 1: Let's add a readme for this package.

And somewhere (not sure where) can we mention that in the future this package could move to something like `@webcomponents/ssr-dom-shim`?

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const attributes: WeakMap<HTMLElement, Map<string, string>> = new WeakMap();
```
justinfagnani, line 7: What are our Node version requirements? It would be nice to just use real private fields. Otherwise, we use `_shadowRoot` for the shadow root, I don't see why we wouldn't be consistent and use `_attributes_` for attributes.

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+const attributesForElement = (element: HTMLElement) => {
+  let attrs = attributes.get(element);
+  if (!attrs) {
```
justinfagnani, line 10:
```suggestion
  if (attrs === undefined) {
```

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  attachShadow(this: HTMLElement, init: ShadowRootInit) {
+    const shadowRoot = {host: this};
+    if (init && init.mode === 'open') {
+      this._shadowRoot = shadowRoot;
+    }
+    return shadowRoot;
+  }
+  getAttribute(name: string) {
+    const value = attributesForElement(this).get(name);
+    return value === undefined ? null : value;
```
justinfagnani, line 54:
```suggestion
    return value ?? null;
```

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+interface CustomHTMLElement {
```
justinfagnani, line 58:
```suggestion
interface CustomHTMLElementConstructor {
```

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  new (): HTMLElement;
+  observedAttributes?: string[];
+}
+
+class ShadowRoot {}
+
+type CustomElementRegistration = {
+  ctor: {new (): HTMLElement};
+  observedAttributes: string[];
```
justinfagnani, line 67: Should we be correct here and tear off the other callbacks?

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  define(name: string, ctor: CustomHTMLElement) {
+    this.__definitions.set(name, {
+      ctor,
+      observedAttributes: (ctor as CustomHTMLElement).observedAttributes ?? [],
+    });
+  }
+
+  get(name: string) {
+    const definition = this.__definitions.get(name);
+    return definition && definition.ctor;
```
justinfagnani, line 82:
```suggestion
    return definition?.ctor;
```

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+class ShadowRoot {}
+
+type CustomElementRegistration = {
+  ctor: {new (): HTMLElement};
+  observedAttributes: string[];
+};
+
+export class CustomElementRegistry {
+  private __definitions = new Map<string, CustomElementRegistration>();
```
justinfagnani, line 71: Let's name private fields consistently: one underscore or two (for `_shadowRoot` above).

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  define(name: string, ctor: CustomHTMLElement) {
+    this.__definitions.set(name, {
```
justinfagnani, line 74: We should throw if an element is redefined.

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    return Array.from(attributesForElement(this)).map(([name, value]) => ({
+      name,
+      value,
+    }));
+  }
+  private _shadowRoot: null | ShadowRoot = null;
+  get shadowRoot() {
+    return this._shadowRoot;
+  }
+  abstract attributeChangedCallback?(
```
justinfagnani, line 29: Why do we need this defined?

justinfagnani (changes requested):
I left a bunch of comments that are for existing code, but I didn't see before.

aomarks:
I've started adding some tests, and noticed that we can also smooth over support for `btoa` in `lit-html/experimental-hydrate.js`. The latest two commits add automatic shimming for `btoa`, which means [1] btoa does not need to be shimmed when doing global SSR on Node 14 anymore, and [2] btoa does not need to be provided as a built-in when doing VM SSR (which actually means that *no* built-ins need to be provided in VM mode for Lit to work anymore).

packages/labs/ssr-dom-shim/src/index.ts
aomarks: @kevinpschaaf do you remember?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -392,18 +393,21 @@ export type WarningKind = 'change-in-update' | 'migration';
...
 export abstract class ReactiveElement
+  // In the Node build, this `extends` clause will be substituted with
+  // `(globalThis.HTMLElement ?? HTMLElement)`.
+  //
+  // This way, we will first prefer any global `HTMLElement` polyfill that the
+  // user has assigned, and then fall back to the `HTMLElement` shim which has
+  // been imported (see note at the top of this file about how this import is
+  // generated by Rollup). Note that the `HTMLElement` variable has been
+  // shadowed by this import, so it no longer refers to the global.
   extends HTMLElement
```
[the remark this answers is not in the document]
aomarks: I found that this makes TypeScript static analysis unhappy, even though it compiles OK. IIRC it causes packages like ts-transformers and I believe localize to fail to understand subclasses of LitElement. Something to do with how that phrasing changes the signature of the class.

packages/lit-html/src/experimental-hydrate.ts
```diff
@@ -14,6 +14,11 @@ import {
   isTemplateResult,
 } from './directive-helpers.js';
 
+// In the Node build, this import will be injected by Rollup:
+// import {Buffer} from 'node';
```
augustjk, line 18:
```suggestion
// import {Buffer} from 'buffer';
```

packages/labs/ssr-dom-shim/src/index.ts
aomarks: Done.

Note this caused the tests to fail, because I had modified the tests to run both with and without DOM shimming. That meant in global mode, the elements were being registered twice.

The simplest solution to this was just to separate out the test scripts, one for shimmed and one for unshimmed, so that the global state doesn't matter. (Once we remove the DOM shim at some later date, we can just drop the shimmed tests).

aomarks (commented):
I believe all comments are now addressed.

I've also added full test coverage, and made a few other changes based on discussions:

- `ModuleLoader` now provides a default VM global context object which provides basic globals that are available in both Node and browsers. A context object can also be directly created with a new `makeDefaultContextObject` function.

- Moved APIs from `HTMLElement` to `Element`, since that's where they truly belong.

- Fixed the demos (which I think have been broken for a while because of `node_modules/` moving to the root when we switched to npm workspaces), and updated them to not import the global DOM shim.

- `btoa` is now automatically shimmed when running in Node in `lit/experimental-hydrate.js` using `Buffer`, which means even an empty global context object will work for basic Lit use cases when doing VM rendering (though we do now also provide `btoa` by default, but it's only available on Node 16+).

- The types of the exports of the shims are now cast to the real thing, which allows them to be used in more type-aware contexts.

aomarks:
Also just pushed updates to the following packages to remove usage of the global DOM shim in its code/tests/READMEs:

- `@lit-labs/testing` (the actual code)
- `@lit-labs/eleventy-plugin-lit` (the actual code)
- `@lit-labs/ssr` (the README)
- `@lit/localize-tools` (a test)

aomarks:
See also https://github.com/lit/lit.dev/pull/1026 which updates the lit.dev docs to account for this new approach of automatic shimming.

packages/labs/ssr-dom-shim/src/index.ts
justinfagnani: You invoke them with `call()`. The registration keeps a pointer to the element and the callbacks for the `this` reference. Something like:

```ts
function connectElement(el) {
  const definition = definitions.get(el.tagName);
  definition.connectedCallback?.call(el);
}
```

I think we'll likely only need to call attributeChangedCallback and maybe connectedCallback on the server. Seems relatively easy to get more strictly correct though.

packages/labs/ssr-dom-shim/src/index.ts
justinfagnani: I suspect we can drop it

justinfagnani (approved):
awesome!

packages/reactive-element/src/reactive-element.ts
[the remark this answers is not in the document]
kevinpschaaf: Oh right, yeah ok.

packages/labs/ssr-dom-shim/src/index.ts
kevinpschaaf: Yeah don't recall, can go.

packages/labs/testing/src/lib/worker.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import {parentPort, workerData} from 'worker_threads';
-import {render} from '@lit-labs/ssr/lib/render-with-global-dom-shim.js';
+import {render} from '@lit-labs/ssr/lib/render-lit-html.js';
```
augustjk, line 8:
```suggestion
import {render} from '@lit-labs/ssr';
```

packages/localize-tools/CHANGELOG.md
```diff
@@ -127,7 +127,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   ```ts
   import {configureSsrLocalization} from '@lit/localize-tools/lib/ssr.js';
-  import {render} from '@lit-labs/ssr/lib/render-with-global-dom-shim.js';
+  import {render} from '@lit-labs/ssr/lib/render-lit-html.js';
```
augustjk, line 130:
```suggestion
  import {render} from '@lit-labs/ssr';
```

packages/localize-tools/src/tests/ssr.unit.test.ts
```diff
@@ -8,7 +8,7 @@ import {test} from 'uvu';
 // eslint-disable-next-line import/extensions
 import * as assert from 'uvu/assert';
 import {configureSsrLocalization} from '../ssr.js';
-import {render} from '@lit-labs/ssr/lib/render-with-global-dom-shim.js';
+import {render} from '@lit-labs/ssr/lib/render-lit-html.js';
```
augustjk, line 11:
```suggestion
import {render} from '@lit-labs/ssr';
```

augustjk (approved):
This is fantastic! Super excited to see this land.

Minor suggestions for importing `render` from `@lit-labs/ssr` in other packages.

kevinpschaaf (approved):

.changeset/empty-cooks-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': minor
```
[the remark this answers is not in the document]
aomarks: Yeah it should be, but since we're still on `0.x.x` we have to call it `minor` or else changesets will release `1.0.0`.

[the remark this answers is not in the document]
aomarks: Although now that I think of it, let's just release `1.0.0`

packages/labs/ssr-dom-shim/src/index.ts
aomarks: We still support Node 14

packages/labs/ssr-dom-shim/src/index.ts
aomarks: But `get` returns the constructor, so `observedAttributes` and `connectedCallback` could still be switched out after registration, right? Doesn't seem like saving them into the definition map does anything useful unless we have another API that can return them separately from the constructor?

justinfagnani: Sorry, I'm using some pseudo-API in that snippet that's not clear. I was thinking that `definitions` would be the map of CE definitions, so `get()` would return a record that holds the ctor, tagName, `observedAttributes`, and callbacks.

aomarks: But that wouldn't match the actual API, so we can't do that, right? We have to return the constructor.

justinfagnani: I'm not talking about the public CustomElementRegistry interface, but the internals which we would use to drive the lifecycles we call on the server.

aomarks: But the internals are isolated within this module -- so that's why I was saying we would need some kind of new side-channel API, right?

justinfagnani: Yes... we'll need a few of those too. Like we also need direct access to `el.shadowRoot` regardless of whether it's open or closed.

aomarks: Ok, talked offline. Consensus is that yes, we would need a separate API to access the lifecycle methods.

Another important thing I didn't realize is that it's important that we read the `observedAttributes` property, in case it's a getter that has side-effects, as it does in the case of Lit (it triggers finalization of the class).

We currently also save `observedAttributes` into the private CE definitions map, but then we don't provide any way for it to possibly ever be read again. For that we would indeed need some kind of side channel/internal API.

aomarks: (But I'm just punting on this for now)

PR #3525, opened 2022-12-15
[labs/virtualizer] Fix #3493

### LINKED ISSUE #3493: [labs/virtualizer] Doesn't update on scroll when a clipping ancestor is in Shadow DOM

graynorton:
Fix #3493

### DIFF
--- a/.changeset/slimy-bees-appear.md
+++ b/.changeset/slimy-bees-appear.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/virtualizer': patch
+---
+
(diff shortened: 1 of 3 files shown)


usergenic (approved):
LGTM!  Thanks for the test cases @graynorton

PR #3526, opened 2022-12-15
disabled shadydom in node dev build

WickyNilliams:
In a next.js environent this line was throwing because `window` is undefined. This module was imported transitively via a component importing the ref directive, which in turn imports async-directive, which in turn imports the file where the fix was added.

Related #3156 

I notice [there are tests](https://github.com/lit/lit/pull/3156/files#diff-1b9f30d8d53005f52e428cbb9c40a8b467742740f50445af9050f67c11944c43R23) for `lit-html` in that PR which import the ref directive, so i'm not sure how they pass (or conversely how it is not working for me)

### DIFF
--- a/.changeset/friendly-papayas-glow.md
+++ b/.changeset/friendly-papayas-glow.md
@@ -1,0 +1,6 @@
+---
+'lit': patch
+'lit-html': patch
+---
+
+disable shadydom in node dev environments
--- a/.changeset/friendly-papayas-glow.md
+++ b/.changeset/friendly-papayas-glow.md
@@ -3,4 +3,4 @@
 'lit-html': patch
 ---
 
-disable shadydom in node dev environments
+Disable ShadyDOM noPatch in Node dev build. This fixes the issue of throwing due to undefined `window`.
--- a/packages/lit-html/package.json
+++ b/packages/lit-html/package.json
@@ -287,7 +287,8 @@
     "test": "wireit",
     "test:dev": "wireit",
     "test:prod": "wireit",
-    "test:node": "wireit"
+    "test:node": "wireit",
+    "test:node-dev": "wireit"
   },
   "files": [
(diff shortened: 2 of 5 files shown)


augustjk:
Thank you for the PR!

Ah.. this probably only happens when running in dev mode in Next.js and introduced with https://github.com/lit/lit/pull/3495. The tests you reference are passing because they are testing against the production Node build in which `ENABLE_SHADYDOM_NOPATCH` is replaced with `false` during the rollup build.

It makes no sense to keep that `true` in Node builds regardless of prod/dev, so I think the correct move might be to make that `false` for dev Node builds as well.

The lines here
https://github.com/lit/lit/blob/feb2494dee3f6f1c907fd432023955fc5f040e28/rollup-common.js#L452-L453

should also be added below here
https://github.com/lit/lit/blob/feb2494dee3f6f1c907fd432023955fc5f040e28/rollup-common.js#L496

WickyNilliams:
excuse the delay, first day back in the office. i'll update the PR as suggested, and also generate a changeset

WickyNilliams:
ok, updated as per suggestion and dropped previous commit.

re: changesets, should i just pick the `lit` package? i am wondering if this change is more broad since it changes the common rollup config

augustjk:
> re: changesets, should i just pick the `lit` package? i am wondering if this change is more broad since it changes the common rollup config

I think a patch to `lit` and `lit-html` makes sense as the ones being affected by that flag.

Thank you!

WickyNilliams:
added the changeset as recommended. i had to battle with the CLA check a little, but it's all good now.

augustjk (approved):
Thank you very much!
I added a bit more description the changeset and tests for node dev builds.

WickyNilliams:
Great, thanks! Appreciate you taking the time

PR #3529, opened 2022-12-16
[labs/analyzer] More analysis coverage for gen-manifest

kevinpschaaf:
This PR includes the following:

* CLI improvements:
  * Print `generate` command diagnostic errors more gracefully to CLI - https://github.com/lit/lit/pull/3529/commits/8d7b0e976d7eace4b8ff94b61e72994d0bef9c5a
  * Add support for `--exclude` options (e.g. exclude tests): https://github.com/lit/lit/commit/7fac64ce9efad4852bc1c5fd592aa81708bf28b0) - important for excluding test files from e.g. manifest or wrapper generation
* Add more analysis support and CEM emit:
  * TS enum type variables: https://github.com/lit/lit/commit/bb8ba3151e559cc197061aa2da8dc23dd61c2fda
  * description, summary, and deprecated for all models: https://github.com/lit/lit/commit/de76b2e891190c0d187c97388aaebb9a0d97b4c3
  * module-level description & summary: https://github.com/lit/lit/commit/32145ba9ef972c002874c2557cf97f4ce8bf0cf6
  * `ClassField` and `ClassMethod`: https://github.com/lit/lit/commit/3fd322c250f620feccd57c688044811820d8b71d
* Also introduces `...IfNotEmpty()` helpers in `gen-manifest` to trim down the JSON size by only including optional fields when they have with meaningful values (e.g. non-empty string, non-empty array, true boolean, etc.)

This PR successfully builds the following manifest for MWC: [custom-elements.json](https://gist.github.com/kevinpschaaf/b52db0f4fc89e90dc791be49a010b11c), after making the following minor MWC source changes:
* `menu/lib/menu.ts`
  * export `ActionDetail` interface
  * remove unused? `{SelectedDetail}` type from `@fires` annotation

Addresses much of remaining items in #2993, with exception of:
* `FunctionDeclaration`
* `MixinDeclaration` / `CustomElementMixinDeclaration`
* `Attributes` (of LitElementDeclaration)
*  `ClassField`s from accessors


sorvell:
I noticed that some of the `privacy` labels in [custom-elements.json](https://gist.github.com/kevinpschaaf/b52db0f4fc89e90dc791be49a010b11c) are incorrect. For example `firstUpdated` is noted as public. Is that because the source is not properly including `override` or does privacy need to be brought forward from the base class?

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -296,6 +323,78 @@ export class VariableDeclaration extends Declaration {
...
+  static?: boolean | undefined;
+  privacy?: Privacy | undefined;
+  inheritedFrom?: Reference | undefined;
+  source?: SourceReference | undefined;
+  parameters?: Parameter[] | undefined;
+  return?: Return | undefined;
+  constructor(init: ClassMethodInit) {
+    super(init);
+    this.static = init.static;
+    this.privacy = init.privacy;
```
sorvell, line 365: It's pretty important that this is correct and it shouldn't change from the base implementation so could/should this read from the `inheritedFrom` reference if it's not provided? (also for ClassField)

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -4,131 +4,281 @@
...
+  if (comment == undefined) {
+    return undefined;
+  }
+  const nameDescSummary = comment.match(parseNameTypeDescSummaryRE);
+  if (nameDescSummary === null) {
+    throw new DiagnosticsError(tag, 'Unexpected JSDoc format');
+  }
+  const {name, type, description, summary} = nameDescSummary.groups!;
+  const info: NamedTypedJSDocInfo = {name, type};
+  if (summary !== undefined) {
```
sorvell, line 92: Can type be undefined as well?

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -4,131 +4,281 @@
...
  * Parses name, summary, and description from JSDoc tag for things like @slot,
  * @cssPart, and @cssProp.
  *
  * Supports the following patterns following the tag (TS parses the tag for us):
  * * @slot name
- * * @slot name summary
- * * @slot name: summary
- * * @slot name - summary
+ * * @slot name description
+ * * @slot name: description
```
sorvell, line 108: Why is a `:` separator supported here but not in `parseNamedTypedJSDocInfo`?

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -304,33 +403,64 @@ export type ClassHeritage = {
 export interface ClassDeclarationInit extends DeclarationInit {
   node: ts.ClassDeclaration;
   getHeritage: () => ClassHeritage;
+  fieldMap?: Map<string, ClassField> | undefined;
```
sorvell, line 406: Can the key be `PropertyKey` instead of `string`? Do we support, for example, symbols here?

sorvell (changes requested):

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Per discussion in chat, it's legal (though probably mostly unintentional) to widen the visibility, so not sure we should/can do anything safe here.

> or does privacy need to be brought forward from the base class?

I think the conclusion here is "yes", the author should carry the visibility forward when overriding. I will add this to the doc I'm writing re: API doc best practices.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
kevinpschaaf: Yes, if the regex does't find a type string in the `{...}`, it just goes in as `undefined`

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
kevinpschaaf: After consulting with Justin, we removed supporting `:` as a separator altogether. My best guess is I misinterpreted a regex originally implemented by Justin as intending to include `:` and started supporting that everywhere.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Hmm, I guess we should, at least in the analyzer?

We'd need to come up with a syntax for emitting that in CEM though, since there's no schema space to describe a Symbol name. Maybe class member could have `name: string | Reference`, where if it's a reference it points to a symbol? @justinfagnani

kevinpschaaf: Will handle separate from this PR. I added an issue on CEM schema here: 

* https://github.com/webcomponents/custom-elements-manifest/issues/114

and an analyzer issue here: 

* https://github.com/lit/lit/issues/3565

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -15,120 +26,264 @@ import {AnalyzerInterface, NamedJSDocInfo, NodeJSDocInfo} from '../model.js';
...
+export const hasJSDocTag = (node: ts.Node, tag: string) => {
+  return ts.getJSDocTags(node).some((t) => t.tagName.text === tag);
+};
+
+/**
+ * Remove line feeds from JSDoc comments, so they are normalized to
  * unix `\n` line endings.
  */
-const normalizeLineEndings = (s: string) => s.replace(/\r/g, '');
+const normalizeLineEndings = (s: string) => s.replace(/\r/g, '').trim();
```
justinfagnani, line 39: Just "normalize" now that it trims?

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -15,120 +26,264 @@ import {AnalyzerInterface, NamedJSDocInfo, NodeJSDocInfo} from '../model.js';
...
+// default slot has no name)
+const parseNameDashDescRE =
+  /^(?<name>^\S*)?(?:\s+-\s+(?<description>[\s\S]*))?$/m;
+
+const getJSDocTagComment = (tag: ts.JSDocTag) => {
+  let {comment} = tag;
+  if (comment === undefined) {
+    return undefined;
+  }
+  if (Array.isArray(comment)) {
```
justinfagnani, line 59: When is this an array?

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -15,120 +26,264 @@ import {AnalyzerInterface, NamedJSDocInfo, NodeJSDocInfo} from '../model.js';
...
+    comment = comment.map((c) => c.text).join('');
+  }
+  if (typeof comment !== 'string') {
+    throw new DiagnosticsError(tag, `Internal error: unsupported node type`);
```
justinfagnani, line 63: Again unsure if we want to throw here... let's chat about error handling.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -15,120 +26,264 @@ import {AnalyzerInterface, NamedJSDocInfo, NodeJSDocInfo} from '../model.js';
...
+ * Thus, we parse module-level documentation using the following heuristic:
+ * - If the first statement only has one JSDoc block, it is only treated as
+ *   module documentation if it contains a `@module`, `@fileoverview`, or
+ *   `@packageDocumentation` tag. This is required to disambiguate a module
+ *   description (with an undocumented first statement) from documentation
+ *   for the first statement.
+ * - If the first statement has more than one JSDoc block, we collect all
+ *   but the last and use those, regardless of whether they contain one
+ *   of the above module-designating tags (the last one is assumed to belong
+ *   to the first statement).
```
justinfagnani, line 202: nice comment

packages/labs/analyzer/src/lib/model.ts
justinfagnani: Do we have real-world cases where we want to analyze symbol-keyed fields? Just wondering if we can punt for a while.

Since you can't bind to them in lit-html, they'd only be good for imperative usage too, which TypeScript should already handle just fine.

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -304,33 +403,64 @@ export type ClassHeritage = {
...
+    this._fieldMap = init.fieldMap;
+    this._methodMap = init.methodMap;
   }
 
   get heritage(): ClassHeritage {
     return (this._heritage ??= this._getHeritage());
   }
+
+  get fields() {
+    return Array.from(this._fieldMap?.values() ?? []);
```
justinfagnani, line 430: Do you not want to just return the iterator or undefined and let the caller convert to an array if needed? 
```suggestion
    return this._fieldMap?.values();
```

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -304,33 +403,64 @@ export type ClassHeritage = {
...
+
+  getField(name: string): ClassField | undefined {
+    return this._fieldMap?.get(name);
+  }
+
+  getMethod(name: string): ClassMethod | undefined {
+    return this._methodMap?.get(name);
+  }
+
+  getMember(name: string): ClassMethod | ClassField | undefined {
```
justinfagnani, line 445: Can you document if these do prototype lookups or just get own properties?

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -304,33 +403,64 @@ export type ClassHeritage = {
...
+    return this.getField(name) ?? this.getMethod(name);
```
justinfagnani, line 446: This actually implements a little bit of prototype lookup in that fields are own properties of the instance and methods are on the prototype. So you could have a field and method of the same name and get correct resolution here.

ie, this is valid JS:

```js
class A {
  x = 1;
  x() {}
}
```

Worth a comment?

justinfagnani (commented):
I'm about halfway through

packages/labs/analyzer/src/lib/analyze-package.ts
```diff
@@ -4,14 +4,21 @@ import * as path from 'path';
 import {DiagnosticsError} from './errors.js';
 import {Analyzer} from './analyzer.js';
 
+export interface AnalyzerOptions {
+  exclude?: string[];
```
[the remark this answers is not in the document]
kevinpschaaf: Done.

Yeah, globs. Useful for excluding things source like test folders that might otherwise be included in a project's tsconfig.

It might be interesting (later on, not now) to analyze a package.json's `files` or `exports` and filter down the input source to analyze based on that. 

However ^^ highlights an unsolved problem in the analyzer right now (will likely need to be resolved via config) which is that just because we can determine that `src/foo.ts` gets output by tsc to `build/foo.js` doesn't mean that's where the file is ultimately importable from, since additional transforms outside of ts can (and often do) further process that into another location that is ultimately published (i.e. we have `tsc` output to `./development/*.js` and process that further with rollup to `./*.js`). 

This is something we're going to hit pretty quickly when we start testing against real-world packages. It's been on my radar, but went ahead and added https://github.com/lit/lit/issues/3578 to track it.

packages/labs/analyzer/src/lib/analyze-package.ts
```diff
@@ -55,6 +65,7 @@ export const createPackageAnalyzer = (packagePath: AbsolutePath) => {
           moduleResolution: 'node',
         },
         include: ['**/*.js'],
+        exclude: options.exclude ? [...options.exclude] : [],
```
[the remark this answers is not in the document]
kevinpschaaf: Changed to `options.exclude ?? []`. Was probably just being paranoid I guess?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -41,7 +52,54 @@ const getClassDeclaration = (
...
+    if (ts.isMethodDeclaration(node)) {
+      methodMap.set(
+        node.name.getText(),
+        new ClassMethod({
+          ...getMemberInfo(node),
+          ...getFunctionLikeInfo(node, analyzer),
+          ...parseNodeJSDocInfo(node),
+        })
+      );
+    } else if (ts.isPropertyDeclaration(node)) {
```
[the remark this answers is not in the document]
kevinpschaaf: > What about accessors? Are they handled in one of these two cases?

No, not handled yet. I added that to the punch list here: https://github.com/lit/lit/issues/2993

> And are function-valued properties handled as-is, or represented as methods?

Right now they're handed as function-valued fields, which feels right since they're not "methods" in the technical sense (often intended to be overridden at instance time where methods are not, etc).

However, for exported declarations, I am planning (unless you have thoughts otherwise) to handle `export const foo = () => {...}` as a `FunctionDeclaration` rather than a function-valued `VariableDeclaration`, since it is mostly unobservable/stylistic differences between that and `export function foo() {...}`.

packages/labs/analyzer/src/lib/javascript/functions.ts
```diff
@@ -0,0 +1,81 @@
...
+    node,
+    ts.SyntaxKind.JSDocReturnTag
+  )[0];
+  const signature = analyzer.program
+    .getTypeChecker()
+    .getSignatureFromDeclaration(node);
+  if (signature === undefined) {
+    throw new DiagnosticsError(
+      node,
+      `Could not get signature to determine return type`
```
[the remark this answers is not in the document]
kevinpschaaf: Adding error boundaries and accumulate + continue is tracked in https://github.com/lit/lit/issues/3009, and once we get some breathing room, I'd like to add that systematically everywhere.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
kevinpschaaf: The first-order purpose is normalizing windows/unix line endings (i.e. getting `\r` out); the trim was just adding a little extra value-add. So I think keeping the "line endings" in the name makes it clearer what it's doing.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
kevinpschaaf: Same as above, lets handle systemically per https://github.com/lit/lit/issues/3009

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Uh, sure. Done.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Good call, done (for now they only return info on the immediate class, not inherited members). 

I figure we'll add another API or `inherited: boolean` flag later to automatically return the full list, once we have use cases that need it, but didn't want to go crazy adding a bunch of API that we're not using yet.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Done.

In TS that's an error, but it's possible when analyzing JS I guess. Added comment describing same-named fields as taking precedence due to being instance properties.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Yeah, we can definitely punt for a while, seems pretty low-priority.

packages/labs/analyzer/src/lib/analyze-package.ts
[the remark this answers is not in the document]
justinfagnani: I think there are a few things we can do here that won't cover _every_ case, but will cover a lot of them... I'll comment on the issue.

The biggest thing that clicked for me here is that these augment the tsconfig exclude setting.

packages/labs/analyzer/src/lib/javascript/classes.ts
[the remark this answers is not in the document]
justinfagnani: I think that's right. I was asking exactly because a field cannot be overridden with a method.

And for exported function variables as functions, yes also - I don't think the difference is observable at all.

packages/labs/gen-manifest/src/index.ts
```diff
@@ -114,12 +113,18 @@ const convertModule = (module: Module): cem.Module => {
   };
 };
 
+const convertCommonInfo = (info: NodeJSDocInfo) => {
+  return {
+    ...pickIfNotEmpty(info, 'description'),
+    ...pickIfNotEmpty(info, 'summary'),
+    ...pickIfNotEmpty(info, 'deprecated'),
+  };
```
justinfagnani, line 121: `undefined` properties aren't written out with `JSON.stringify()`, so might it be nicer to not use spread to write a single value, and do something like:

```suggestion
  return {
    description: removeEmpty(info.description),
    summary: removeEmpty(info.summary),
    deprecated: removeEmpty(info.deprecated),
  };
```

Or, are the all the properties in NodeJSDocInfo always going to be used? Could you transform them all?

```suggestion
  return removeEmpties(info);
```

packages/labs/gen-manifest/src/index.ts
```diff
@@ -197,14 +202,77 @@ const convertClassDeclaration = (
       superClass ? convertReference(superClass) : undefined
     ),
     // mixins: [], // TODO
-    // members: [
-    //   // TODO: ClassField
-    //   // TODO: ClassMethod
-    // ],
+    ...useIfNotEmpty('members', [
```
justinfagnani, line 205: Similar comment about `undefined` being ok here, so we don't need spread.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -116,19 +148,43 @@ export const parseNodeJSDocInfo = (
...
+ * Parse summary, description, and deprecated information from JSDoc comments on
+ * a given source file.
+ */
+export const parseModuleJSDocInfo = (sourceFile: ts.SourceFile) => {
+  const firstChild = ts.forEachChild(sourceFile, (n) => n);
+  if (firstChild === undefined) {
+    return {};
+  }
+  const jsDocs = firstChild.getChildren().filter(ts.isJSDoc);
+  const moduleJSDocs = jsDocs.slice(0, jsDocs.length > 1 ? -1 : 1);
```
justinfagnani: Might be worth some comments in here

justinfagnani (approved):

packages/labs/gen-manifest/src/index.ts
kevinpschaaf: > undefined properties aren't written out with JSON.stringify()

OMG! 🤦  Yeah I think I knew this but totally forgot about it... definitely better to leave the key and avoid the spread. Will refactor.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
kevinpschaaf: Done

packages/labs/gen-manifest/src/index.ts
kevinpschaaf: Done

packages/labs/gen-manifest/src/index.ts
kevinpschaaf: Refactored into
* `ifNotEmpty(v)
* `transformIfNotEmpty(v, transform)
which always return a meaningful value or undefined, and are assigned directly to keys (not spread)

packages/labs/gen-manifest/src/index.ts
```diff
@@ -15,18 +15,25 @@ import {
...
+  NodeJSDocInfo,
 } from '@lit-labs/analyzer';
 import {FileTree} from '@lit-labs/gen-utils/lib/file-utils.js';
 import type * as cem from 'custom-elements-manifest/schema';
 
+const isNotEmpty = (v: unknown) =>
+  v !== undefined &&
+  (typeof v !== 'boolean' || v === true) &&
+  (typeof v !== 'string' || v.length > 0) &&
+  (!Array.isArray(v) || v.length);
```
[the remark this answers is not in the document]
kevinpschaaf: Done

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -406,11 +485,24 @@ export interface LitElementExport extends LitElementDeclaration {
   tagname: string;
 }
 
-export interface ReactiveProperty {
+export interface PropertyLike extends NodeJSDocInfo {
```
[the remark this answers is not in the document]
kevinpschaaf: Fair, I mean those things can't be inferred from source other than from JSDoc comments, which is how I got to that naming, but I renamed them to be more generic as suggested.

packages/labs/gen-manifest/src/index.ts
```diff
@@ -24,43 +24,38 @@ import {
...
+ * we don't serialize a key with a meaningless value to JSON, to cut down size
+ * of the manifest (note that `JSON.serialize` will not emit keys with
+ * `undefined` values)
+ */
+const ifNotEmpty = <T>(v: T): T | undefined => {
+  if (
+    v === undefined ||
+    (typeof v === 'boolean' && v === false) ||
+    (typeof v === 'string' && v.length === 0) ||
+    (Array.isArray(v) && v.length === 0)
```
justinfagnani, line 39: This can still be simplified, ie with `===` the `typeof v === 'boolean'` isn't necessary, and the undefined check isn't needed:

```suggestion
v === false || ((typeof v === 'string' || Array.isArray(v)) && v.length === 0)
```

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -314,6 +323,78 @@ export class VariableDeclaration extends Declaration {
...
+  type?: Type | undefined;
+  default?: string | undefined;
+  constructor(init: ClassFieldInit) {
+    super(init);
+    this.static = init.static;
+    this.privacy = init.privacy;
+    this.inheritedFrom = init.inheritedFrom;
+    this.source = init.source;
+    this.type = init.type;
+    this.default = init.default;
```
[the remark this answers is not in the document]
sorvell: clever

sorvell (approved):

PR #3530, opened 2022-12-16
[infra] Skip labs/ssr for next release

augustjk:
Adds `@lit-labs/ssr` and packages that depend on it to the ignore list for changeset and use my fork of the changeset action https://github.com/augustjk/changesets-action/tree/test for the next release.

This should be reverted after the next release as packages in the ignore list cannot be selected in the changeset cli to add a new changeset entry.


usergenic (approved):
LGTM brilliant workaround for us to do single-package changeset release -- thank you, @augustjk

PR #3533, opened 2022-12-16
[lit-html] clean dependencies

AliMD:
I love to see `lit-html` without any dependencies 🤗


augustjk:
Thanks for the PR but `@types/trusted-types` is intentionally a dependency, not a dev dependency. See https://github.com/lit/lit/issues/3308#issuecomment-1275297241

PR #3535, opened 2022-12-18
Add filename context to generated xliff files

halkeye:
When I upload xliff files it just says en.xliff but doesn't give them any context. For a design library this can be quite confusing.

Adding the original source filename gives them a hint where the string came from.


packages/localize-tools/src/formatters/xliff.ts
```diff
@@ -192,12 +192,20 @@ export class XliffFormatter implements Formatter {
...
     const existingTranslationsByName = this.makeExistingTranslationsMap(doc);
-    const body = getOneElementByTagNameOrThrow(doc, 'body');
     const names = new Set();
-    for (const {name, contents: sourceContents, desc} of sourceMessages) {
+    for (const {name, contents: sourceContents, desc, file} of sourceMessages) {
+      const body = this.getOrCreateFileBody(
+        doc,
+        targetLocale,
+        this.config.sourceLocale,
+        pathLib.join(
```
halkeye, line 204: there has to be a better way to get a nice path, but couldn't think of any relative path stuff. Maybe loop through all sourceMessages and see what the common root is.

Issue #3521, opened 2022-12-13
[labs/analyzer] extract generic analyses from lit-element logic


43081j:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Other/unknown (please mention in description)

### Description

Currently the analyzer can analyze roughly the following from an element:

* Lit decorators (properties, custom element definition)
* JSDoc tags (events, slots, css properties, etc)
* `customElements.define` calls

Ultimately, this stuff only runs right now if we get a truthy `isLitElementSubclass` result (i.e. something which inherits `LitElement`).

That basically means the analyzer **can only currently analyzer lit elements**.

Most of the logic I mentioned above, though, is _not_ specific to lit elements. I wonder if we could actually split it up into two things:

* Anything which inherits `HTMLElement` (including lit elements)
  * JSDoc tags
  * `customElements.define` calls
* Anything which inherits `LitElement`
  * Lit decorators (properties, customElement)

This'd be fairly easy by doing a few things:

* Create an `isCustomElementSubclass` type guard
* Create a `CustomElementDeclaration` model which actually holds everything the current `LitElementDeclaration` model has _except_ reactive properties
* Update `LitElementDeclaration` to inherit the new `CustomElementDeclaration`
* Split the tag name logic up so the custom element declaration looks for define calls, and the lit declaration does that _in addition to_ checking for `@customElement` decorators

Then shuffle the tests around a bit. Somehow structure it in a way that means we can, for example, run the JSDoc tests against a vanilla element _and_ a lit element with little code duplication.

clearly i could have a stab at this myself but i think i'd get stuck in a rabbit hole with how to nicely rework the tests.

### Alternatives and Workarounds

N/A

justinfagnani:
@43081j I think we could consider this, but to be clear about this intention of the analyze: it's mostly a non-goal to be a generic custom element analyzer, especially one with any kind of plug in system. We want something we can iterate fast on and keep perfectly in sync and changes to Lit syntax or semantics by being in the monorepo and sharing tests.

Generic custom element analysis _might_ be best served by something like Open WC's CEM Analyzer.

That said, I do acknowledge that the model of elements that we build up has only a few Lit specifics parts to it. It seems like it wouldn't be too disruptive to the primary goal of supporting Lit to layer the models so that there's a generic layer and a Lit specialization. From that we might support vanilla custom elements with jsdoc syntax and other WC libraries could consume the analyzer as a library to build their own analyzer on.

[...]

PR #3536, opened 2022-12-18
feat (labs/analyzer): support custom element analysis

43081j:
This separates the current lit-element analysis into two layers:

* custom elements
* lit

Primarily this is so we can detect non-lit elements we may depend on inside a lit project, giving the user full coverage of all the elements they're consuming.

---

This is just a draft/poc of how we could do #3521

if it seems right, i'll see if i can figure the tests out (we need to add new ones before this is ready).

cc @kevinpschaaf this is what i was thinking of. i know you did a fairly big rework in a branch somewhere so , if this is a good approach, maybe we need to redo it off your branch?


kevinpschaaf:
This looks perfect. Just getting back from vacation so let me figure out whether it makes sense to land this or https://github.com/lit/lit/pull/3529 first, but either way this looks good.

43081j:
replaced by #3621

PR #3549, opened 2023-01-04
[infra] Add BROWSERS to cache key for all WTR tests

aomarks:
Our WTR config uses the `BROWSERS` environment variable as a way to specify a subset of browsers to test. E.g. if you just want to test in chromium and firefox, you can do `BROWSERS=chromium,firefox npm test`.

In order to prevent wireit from sharing cached successes across runs with different browsers, we need to tell it about the `BROWSERS` variable, using the `env` feature: https://github.com/google/wireit#indicating-external-environment-variables


augustjk (approved):
LGTM

PR #3561, opened 2023-01-11
[reactive-element] Fix Node build auto-shimming of `HTMLElement` to respect existing global shim

### LINKED ISSUE #3559: ssr-dom-shim in reactive-element is breaking test cases in vitest/happy-dom

### Description

as of v1.6.0 of reactive-element, it imports srr-dom-shim when environment is node.
When testing using dom emulation like happy-dom, it is loading the shim, which is causing test cases to fail.

### Reproduction

 https://stackblitz.com/edit/vitest-dev-vitest-7rrc4s?file=package.json,test%2Fbasic.test.ts

### Workaround

force @lit/reactive-element dependency version to 1.5.0

### Is this a regression?

### Affected versions

Failing with @lit/reactive-element 1.6.0, working with 1.5.0

OS: Windows 11
Node Version: 14.19
npm version: 6.14.17

augustjk:
Fixes #3559 

The replace of `extends HTMLElement` with `extends (globalThis.HTMLElement ?? HTMLElement)` was not happening due to the `includes` option for `@rollup/plugin-replace` not working properly. Updating the version fixes this.

Note: To keep a single version of the plugin in the monorepo, the version was also updated for the starter kits.

Now the outputs look like:

Dev Node build:
```js
class ReactiveElement
 extends (globalThis.HTMLElement ?? HTMLElement) {
```

Prod Node build:
```js
class v extends(globalThis.HTMLElement??i)
```

I added another test in the vein of the node-import tests. It feels like there are a lot of scripts for the node tests now but unsure if there's a better way to do this.

### DIFF
--- a/.changeset/tidy-lamps-rest.md
+++ b/.changeset/tidy-lamps-rest.md
@@ -3,4 +3,4 @@
 '@lit/reactive-element': patch
 ---
 
-Fix built-in shimming of `HTMLElement` for Node build of `reactive-element` to respect existing existing `HTMLElement` in global
+Fix built-in shimming of `HTMLElement` for Node build of `reactive-element` to respect existing `HTMLElement` in global
--- a/.changeset/tidy-lamps-rest.md
+++ b/.changeset/tidy-lamps-rest.md
@@ -1,0 +1,6 @@
+---
+'lit': patch
+'@lit/reactive-element': patch
+---
+
+Fix built-in shimming of `HTMLElement` for Node build of `reactive-element` to respect existing existing `HTMLElement` in global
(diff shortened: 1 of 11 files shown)


.changeset/tidy-lamps-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit': patch
+'@lit/reactive-element': patch
+---
+
+Fix built-in shimming of `HTMLElement` for Node build of `reactive-element` to respect existing existing `HTMLElement` in global
```
aomarks, line 6:
```suggestion
Fix built-in shimming of `HTMLElement` for Node build of `reactive-element` to respect existing `HTMLElement` in global
```

aomarks (approved):
Great!

augustjk:
Filed https://github.com/lit/lit/issues/3562 for better handling of the node tests.

PR #3571, opened 2023-01-17
Accept requestUpdate to be passed in @event binding

vdegenne:
This change will permit to pass `requestUpdate` in a @event binding

```html
<button @click=${this.requestUpdate}></button>
```

(Previously, 
```html
<button @click=${() => this.requestUpdate()}></button>
```
)

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1176,13 +1176,13 @@ export abstract class ReactiveElement
    * @category updates
    */
   requestUpdate(
-    name?: PropertyKey,
+    name?: PropertyKey | Event,
     oldValue?: unknown,
     options?: PropertyDeclaration
   ): void {
     let shouldRequestUpdate = true;
     // If we have a property key, perform property update steps.
-    if (name !== undefined) {
+    if (name !== undefined && !(name instanceof Event)) {
       options =
         options ||
         (this.constructor as typeof ReactiveElement).getPropertyOptions(name);
--- a/packages/reactive-element/src/test/reactive-element_test.ts
+++ b/packages/reactive-element/src/test/reactive-element_test.ts
@@ -3278,4 +3278,11 @@ suite('ReactiveElement', () => {
       // Suppress no-unused-vars warnings
     }
   });
+
+  test('requestUpdate can be used as a template @event binding', () => {
+    class E extends ReactiveElement {}
+    customElements.define(generateElementName(), E);
+    const el = new E();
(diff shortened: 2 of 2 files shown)


augustjk:
Thank you for the PR!

While I do understand the want for a simpler syntax, the intention of `requestUpdate` is not to be used as an event handler, and as such, coding it for the possibility of receiving an event as its argument makes it misleading.

It is valid to want to call `this.requestUpdate()` in response to an event, but that should be explicitly done either with an arrow function as in your "previously" code, or with another named method on the class specifically for handling the event which calls `this.requestUpdate()` within its definition.

vdegenne:
Ok I understand that. Thanks for the review.
I was thinking maybe it would be  acceptable to write this condition when the event listener is registered?

Something like that
```javascript
if (callback.name == 'requestUpdate') {
  callback = () => callback()
}
// create listener
```
Would it be legit? If yes I could open another PR for that, but I don't know what part of the code is responsible for treating `@event` bindings.

augustjk:
Any special logic we add to event binding handling that specifically looks for `requestUpdate` or an event argument seems wrong, potentially an incorrect assumption and a further point of maintenance. It could very well be that `lit-html` can be used for non lit elements where `requestUpdate` is not a special method and was written by the user such that it should get the event object passed to it.

I feel the workaround of doing `@event={() => this.requestUpdate()}` is not so much more difficult that it warrants this special logic, either on the event binding side, or on the `requestUpdate` side.

justinfagnani:
Agreed with @augustjk here. We try to have as little magic and a few exceptional cases as possible. This would be a unique rule that event bindings call the method assed in with the event as an argument _except_ for `requestUpdate`, or that in order to call a non-event-handler function you should pass an arrow function that calls the function, _except_ for `requestUpdate`. That's a kind of documentation smell that we want to avoid to help us keep things simple.

In addition, I'm not sure we'd want to encourage just updating in response to an event. In order for the update to change anything the host would have to have some state dependent on the event... but in this case, what state? Is the host reading directly from the child? That's an unusual pattern, and it's probably best if the event handler modifies reactive state that automatically triggers an update.

vdegenne:
@justinfagnani Ok I understand. Just for clarity I am using direct update in a "search" view. Something like that

```javascript
render() {
  const results = ...; // filtering array
  return html`
  <mwc-textfield @input=${() => this.requestUpdate()}></mwc-textfield>
  <div>${results}</div>
  `
}
```
When the value of the textfield is not really important to store in a class field.

Issue #3569, opened 2023-01-16
@lit/localize-tools/0.6.6 regression with variables 


pmcelhaney:
### Which package(s) are affected?

Localize (@lit/localize)

### Description

The extract function is adding an extra `<x>` tag to the beginning of a string.

```xliff
<?xml version="1.0" encoding="UTF-8"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
<file target-language="es-419" source-language="en" original="lit-localize-inputs" datatype="plaintext">
<body>
<trans-unit id="h71437edc6d0d1c12">
  <!--    v-- should not be here -------------------------------v -->
  <source><x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/>Hello <x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/></source>
</trans-unit>
</body>
</file>
</xliff>
```

### Reproduction

See [this minimal reproduction](https://github.com/pmcelhaney/lit-localize-bug) which uses the "my-greeter" example.

```sh
git clone git@github.com:pmcelhaney/lit-localize-bug.git
npm test
cat ./xliff/es-419.xlf
```

Note line 6 of that file. 

Expected:

```xliff  
<source><x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/>Hello <x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/></source>
```

Actual:

Every time you run `lit-localize extract` (`npm test`), it alternates between these two. 

```xliff  
<source>Hello Hello <x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/></source>
```

```xliff  
<source><x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/>Hello <x id="0" equiv-text="&lt;b&gt;${this.who}&lt;/b&gt;"/></source>
```

### Workaround

I have not found a workaround. 

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

Failing in @lit/localize-tools/0.6.6, worked in 0.6.5

### Browser/OS/Node environment

MacOS 13.1
npm: '8.5.0',
node: '16.14.2',

aomarks:
Thanks for filing, investigating now. Likely related to https://github.com/lit/lit/pull/3514.

aomarks:
[...]

PR #3576, opened 2023-01-17
[localize] Fix XLIFF serialization regression

aomarks:
Fixes a bug introduced in https://github.com/lit/lit/pull/3514 where we were not correctly updating XLIFF messages that contained placeholders, because of an index-invalidation bug in the way we were deleting the contents of existing nodes.

Fixes https://github.com/lit/lit/issues/3569


augustjk (approved):
Nice!

PR #3579, opened 2023-01-18
"It can be installed with no configuration"

vdegenne:
The doc says

https://github.com/lit/lit/blob/023cdeb228124d98df4eb1c236a54b2e3db1d88a/packages/labs/router/README.md?plain=1#L53

But TypeScript will complain about not passing a required argument.


justinfagnani:
Thanks for the change @vdegenne. This looks good to me, but we need a changeset. Can you run `npx changeset` from the root of the repo?

PR #3583, opened 2023-01-18
Better hydration error messages

justinfagnani:
This gives a _little_ more information on two hydration errors where we find either too many or no root parts when hydrating.


aomarks (approved):
Need a changeset

augustjk (approved):

kevinpschaaf (approved):
LGTM

PR #3593, opened 2023-01-20
Add webhook action to add external issues to project board

aomarks:
Adds a webhook-triggered GitHub Action which adds an issue from an external repository to the Lit GitHub project board.

The GitHub API only supports adding issues from a repository in the same org as the project board (see https://github.com/orgs/community/discussions/6212). However, the GitHub UI also supports adding URLs from any repo. So we use a headless browser to work around the lack of API support by automating use of the GitHub UI.

We're using a new `lit-project-board-robot` account that has very limited powers, to limit any risks of needing to store the password and 2FA seed in GitHub secrets. By using a webhook triggered action in this repo, the external repos can then use a PAT to trigger it. This means we don't need to copy/publish this code outside of this repo, and we don't need to share the credentials to any other orgs.

See https://github.com/lit/lit/pull/3593 for an example of calling this webhook from another repo.

Part of https://github.com/lit/lit/issues/3586


augustjk:
This is incredible!
Before we take this step have you seen this mutation? https://docs.github.com/en/graphql/reference/mutations#addprojectv2draftissue
Chances are likely slim but I wonder if adding the issue url to the body of the mutation input would do the same thing?

aomarks:
> This is incredible!
> Before we take this step have you seen this mutation? https://docs.github.com/en/graphql/reference/mutations#addprojectv2draftissue
> Chances are likely slim but I wonder if adding the issue url to the body of the mutation input would do the same thing?

Good idea, will try it out. I suspect it will just create one of those "ghost" issues where the text is the URL, rather than a proper link to the issue -- but let's see!

aomarks:
> Good idea, will try it out. I suspect it will just create one of those "ghost" issues where the text is the URL, rather than a proper link to the issue -- but let's see!

OMG... it worked when I set the URL as the title. Great suggestion! Will close this PR.

PR #3598, opened 2023-01-20
[labs/virtualizer] Disable scrolling tests until issue with latest chromedriver is fixed.

usergenic:
Something in the repo has changed which is breaking the scrolling tests.  Assuming it is chromedriver update, but either way, these don't work and we need to disable them until we fix the issue so we can continue to run our tests.


kevinpschaaf (approved):

augustjk (commented):
not a big deal either way but could we maybe do a `describe.skip()` with a comment/TODO rather than renaming file?

usergenic:
@augustjk that's a fine idea.  Force pushing that change.

augustjk (approved):

kevinpschaaf (approved):
Great suggestion, @augustjk! LGTM

Issue #3589, opened 2023-01-18
[labs/nextjs] A plugin for Next.js to integrate Lit SSR


augustjk:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

SSR (@lit-labs/ssr)

### Description

Part of #2391 and depends on #3588 

With `@lit-labs/ssr-react` package being the core implementation of integrating Lit SSR with React Server DOM, it would be good to have an easy way to configure a Next.js project with a plugin.

We could have a package `@lit-labs/nextjs` which exports function `withLitSsr` which can be used to create/wrap the users's `next.config.js`. This could look like https://github.com/vercel/next.js/blob/canary/packages/next-mdx/index.js

The plugin would add to the webpack config of the Next.js project to add side-effect imports to bundle entrypoints to monkey patch `React.createElement` and also replace any runtime jsx transforms to use `@lit-labs/ssr-react` instead.

Alternatives considered:
- Users could manually configure `jsxImportSource` or `jsxFactory` depending on the runtime mode of the project. The problem with this is that it will not work with components made using `@lit-labs/react`, or other precompiled React components housing Lit components.

### Alternatives and Workarounds

N/A

augustjk:
It would also be nice if we could add the template shadowroot polyfill for non-supported browsers. There may be a way to do this through the webpack config as well. Otherwise, users might need to add it to the bottom of the `<body>` in their `_document.js` file.

justinfagnani:
I don't think we should automatically include the polyfill, except possibly as some addition to an automatic polyfill injection that Next already has.

aomarks:
My thought was that frameworks like Next tend to make things "just work", and frameworks like react don't require thinking about this problem. To match that experience, maybe automatic polyfill injection is something we should actually consider.

Another thought is that to prevent the FOUC problem, we'll also want some easy-to-use way to block rendering until the polyfill has activated, as shown at https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#example-bootup-strategy

augustjk:
As it is, we also get React hydration mismatch error if `<template>` element is still present as a child for browsers that don't support DSD (there might be an option to suppress this warning, though not sure if that's the right thing to do here). If we find a way to do it via webpack config, it could very well be an option for the config generating function whether user wants polyfill added.

steveworkman:
[...]

PR #3605, opened 2023-01-26
[labs/ssr-react] Add package for integrating Lit SSR to React

### LINKED ISSUE #3588: [labs/ssr-react] Lit SSR React integration
### Should this be an RFC?

SSR (@lit-labs/ssr)

### Description

Part of #2391 

As it stands, Lit components can be shallowly server rendered by React Server DOM, as in the host element tag will get server rendered without any of its content.

We want a user-friendly way of integrating server-rendering of `@lit-labs/ssr` to React projects, such that rendering the custom element with React Server DOM will include the contents of the Lit component as a declarative shadow DOM (DSD) child template.

This is a proposal for a new package `@lit-labs/ssr-react`.

The package would contain:
1. A `patchCreateElement` function which takes `React.createElement` function and enhances it with logic that will add the DSD template element as a child to a registered lit component.
2. A side-effect module that will use above function to monkey patch `React.createElement` so users would not have to bring it in and use a separate JSX pragma.
3. Perhaps an already patched `createElement` (with a different name) that can be directly imported and used with a `@jsx` pragma or `jsxFactory` config.
4. Package should also be able to serve as `jsxImportSource` for runtime jsx transform, requiring `jsx-runtime` and `jsx-runtime-dev` modules that export `Fragment`, `jsx`, `jsxs` for production use, and `Fragment`, `jsxDEV` for development.
Related RFC: https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md
Example implementations: https://github.com/system-ui/theme-ui/blob/develop/packages/core/src/jsx-dev-runtime.ts, https://github.com/emotion-js/emotion/blob/main/packages/react/src/jsx-dev-runtime.js

The usage of this package is going to be primarily with build tools like babel/tsc/swc with modified jsx config, or via bundlers to add the side-effect monkey patch and hijack the `react/jsx-runtime` imports. As such, there won't be an opportunity for users to provide a different `renderInfo` for those methods. The `patchCreateElement` function could be written to take that as an option though.

N/A

augustjk:
Resolves #3588 

This package contains tools for integrating Lit SSR with React for deeply server rendering Lit components.

A separate package for integrating these with Next.js is forthcoming(#3589)


packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```js
+import '@lit-labs/ssr-react/enable-lit-ssr.js';
+```
+
+The advantage of this approach is that it does not require modification to any other files containing JSX. It will also work for external components that have been pre-compiled with the classic JSX transform or React components made from Lit components using `@lit-labs/react`.
+
+### Note on Enabling Hydration
+
+Both the Runtime JSX Transform replacement and using `@lit-labs/ssr-react/enable-lit-ssr.js` have conditional exports configured such that when imported for a server environment they enhance React's original element creation to deeply render Lit components, and when imported for the browser will bring in `lit/experimental-hydrate-support.js` and provide pass-throughs for React's original element creation.
```
aomarks, line 62: In theory you might want to use Lit SSR with React but not actually need hydration. In that case always including experimental hydration may not be desirable. Thoughts on that?

packages/labs/ssr-react/package.json
```diff
@@ -0,0 +1,112 @@
...
+        "enable-lit-ssr.{d.ts,d.ts.map,js,js.map}",
+        "tsconfig.tsbuildinfo"
+      ]
+    },
+    "test": {
+      "command": "uvu tests _test.js$",
+      "dependencies": [
+        "build"
+      ]
+    }
```
aomarks, line 94:
```suggestion
    "test": {
      "command": "uvu tests _test.js$",
      "files": [],
      "output": [],
      "dependencies": [
        "build"
      ]
    }
```

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Usage
+
+This package includes several ways of adding Lit SSR integration to React's base element creation to address different JSX transforms as well as pre-compiled components.
+
+### Using the Classic JSX Transform
+
+The classic JSX transform replaces JSX with `React.createElement` function calls.
+
+This package provides the `patchCreateElement` function which takes the `React.createElement` function and returns an enhanced function that adds a template shadowroot element as a child to the host custom element for any registered Lit components it gets.
```
aomarks, line 19: Wondering if this should be called something like `wrapCreateElement` or even `makeLitSsrCreateElement`, since `patch` could imply an in-place/monkey patch (which we support, but not from this module).

aomarks (commented):
Just some preliminary thoughts/question, only looked at the README/package.json so far!

packages/labs/ssr-react/README.md
augustjk: I like that. I prefer `wrapCreateElement` as `makeLitSsrCreateElement` feels too verbose.

packages/labs/ssr-react/README.md
augustjk: I feel like vast majority of use case would want hydration so making that easier sounds better to me. I'm also thinking users who want no hydration for their React app will have a different entry point for their client JS, if any, which can avoid the auto imports.

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```js
+import React from 'react';
+import {patchCreateElement} from `@lit-labs/ssr-react`;
+
+const patchedCreateElement = patchCreateElement(React.creatElement);
```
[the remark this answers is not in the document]
augustjk: I was thinking users would include this for every JSX file, though that would be wasteful. They could do it once, export the patched/wrapped function as you suggest and import that for every JSX file. Or we could vend an already wrapped function in the package so users can just import that instead of creating their own wrapped version.

In which case the wrap function doesn't really need to be user facing, unless users want to use it for wrapping non-React alternative `createElement` functions.

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+#### Or, configuring your compiler
+
+- For Babel: specify the [`pragma`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragma) option for `@babel/plugin-transform-react-jsx`.
+- For TypeScript: specify the [`jsxFactory`](https://www.typescriptlang.org/tsconfig#jsxFactory) option in `tsconfig.json`.
+
+Note: Both of the solutions above require the function `patchedCreateElement` (or whatever specifier you choose for the patched function) already be present in scope of the file with JSX to transform. Each `jsx`/`tsx` file will need to include the code snippet at the top of this section. Therefore, it is recommended to use the `@lit-labs/ssr-react/enable-lit-ssr.js` module mentioned below.
+
+### Using the Runtime JSX Transform
+
+If your project is using the [runtime JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), this package is configured to be able to serve as the JSX import source whose JSX functions have been patched to enable Lit SSR.
```
justinfagnani, line 45:
```suggestion
If your project is using the [runtime JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), this package can serve as the JSX import source.
```

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The advantage of this approach is that it does not require modification to any other files containing JSX. It will also work for external components that have been pre-compiled with the classic JSX transform or React components made from Lit components using `@lit-labs/react`.
+
+### Note on Enabling Hydration
+
+Both the Runtime JSX Transform replacement and using `@lit-labs/ssr-react/enable-lit-ssr.js` have conditional exports configured such that when imported for a server environment they enhance React's original element creation to deeply render Lit components, and when imported for the browser will bring in `lit/experimental-hydrate-support.js` and provide pass-throughs for React's original element creation.
+
+If you are not using either of the above methods, you must manually import `lit/experimental-hydrate-support.js` into your source code before `lit-element` is loaded.
+
+### Other considerations
```
justinfagnani, line 66: Maybe just "Enabling declarative shadow DOM"?

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+For browsers that do not yet support declarative shadow DOM, you must include the [`template-shadowroot`](https://github.com/webcomponents/template-shadowroot) polyfill. Without this, React's own hydration may complain of hydration mismatch due to the lingering `<template>` element. See https://lit.dev/docs/ssr/client-usage/#using-the-template-shadowroot-polyfill for inspiration on how to incorporate this into your application.
```
justinfagnani, line 68: Mention browsers that do support it natively?

justinfagnani (commented):
Just through he readme mostly.

augustjk:
Thanks for the initial feedback! Marking as draft to redo some parts of this before putting it up for review again.

packages/labs/ssr-react/package.json
```diff
@@ -0,0 +1,114 @@
...
+      "types": "./index.d.ts",
+      "node": "./node/index.js",
+      "default": "./index.js"
+    },
+    "./enable-lit-ssr.js": {
+      "types": "./enable-lit-ssr.d.ts",
+      "node": "./node/enable-lit-ssr.js",
+      "default": "./enable-lit-ssr.js"
+    },
+    "./jsx-dev-runtime": {
```
justinfagnani, line 30: Add .js extension

packages/labs/ssr-react/package.json
```diff
@@ -0,0 +1,114 @@
...
+    "uvu": "^0.5.6"
+  },
+  "keywords": [
+    "lit",
+    "lit-html",
+    "lit-element",
+    "LitElement",
+    "ssr",
+    "lit-ssr",
+    "react"
```
justinfagnani, line 112: add "web components" and "next". If there's a limit, I don't know if we need `lit-element` and variations.

packages/labs/ssr-react/src/enable-lit-ssr.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * @fileoverview Side-effect import meant to be loaded in the **browser** before
+ * any user code is loaded. Installs hydration support for `LitElement`.
+ */
```
justinfagnani, line 10: why do we need this file in addition to `lit/experimental-hydrate-support.js`?

packages/labs/ssr-react/src/jsx-dev-runtime.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @fileoverview To serve as JSX import source for runtime JSX transforms in
+ * development mode. For use in browsers.
+ */
+
+import 'lit/experimental-hydrate-support.js';
+
+// eslint-disable-next-line import/extensions
+export {jsxDEV} from 'react/jsx-dev-runtime';
```
justinfagnani, line 15: Why doesn't this have Fragment and jsxs? Should it be an `export *`?

Is there a link to docs on the module exports contract for this module?

packages/labs/ssr-react/src/jsx-runtime.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+/**
+ * @fileoverview To serve as JSX import source for runtime JSX transforms in
+ * production mode. For use in browsers.
+ */
+
+import 'lit/experimental-hydrate-support.js';
+// eslint-disable-next-line import/extensions
+export {Fragment, jsx, jsxs} from 'react/jsx-runtime';
```
justinfagnani, line 14: Should this be `export *`?

packages/labs/ssr-react/src/lib/node/create-element.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import React from 'react';
+import {wrapCreateElement} from './wrap-create-element.js';
+
+/**
+ * React's `createElement` function enhanced to deeply render Lit components in
+ * the server. It is simply a passthrough of `React.createElement` in the
+ * browser.
+ */
+export const createElement = wrapCreateElement(React.createElement);
```
justinfagnani, line 15: Can we not do this with an `if (isServer)` check and have one module?

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
```diff
@@ -0,0 +1,81 @@
...
+  // elementAttributes will be provided to React as props for the host element
+  // for properly rendering reflected attributes
+  const elementAttributes =
+    renderer.element !== undefined
+      ? attributesToProps(renderer.element.attributes)
+      : {};
+
+  const {mode = 'open', delegatesFocus} = renderer.shadowRootOptions;
+  const templateAttributes = {
+    shadowroot: mode,
```
justinfagnani, line 72: Change to `shadowrootmode: mode` for compatibility with Safari and streaming SSR. We need to update our polyfill too.

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  renderer.connectedCallback();
+
+  renderInfo.customElementInstanceStack.push(renderer);
+
+  const shadowContents = renderer.renderShadow(renderInfo);
+
+  // elementAttributes will be provided to React as props for the host element
+  // for properly rendering reflected attributes
+  const elementAttributes =
```
justinfagnani, line 65: We should update the ElementRenderer interface to have a `getAttributes()` API.

justinfagnani (approved):
🎉

packages/labs/ssr-react/package.json
augustjk: We can't do that as the compilers that do the transform don't add the `.js` extension on the import line. See https://www.typescriptlang.org/play?jsx=4#code/PQKhAIAECsGcA8CSBbADgewE4BcDK6BXTAYwFMoAbAS2wFoKBDAI1mFlk1s1IeO3BDAAUAB5kAT1qkKpZKQB2-VJnSoAjAF4A5MtVqt4ANalx24+K0A+IQAsqQgN4BtNQBoATK4DMAXQB0yAyoABRU4BqW4CKwqAzyRiYaDlQAvpbJKSJssfKWAJQposASUjJyipZAA

packages/labs/ssr-react/src/enable-lit-ssr.ts
augustjk: With projects like Next.js that use the same source code for both server and client, I figured it's a convenience for users to just import this once at the top, and it'll handle the monkey patching on the server, and hydration import on the client.

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
augustjk: Filed https://github.com/lit/lit/issues/3630

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
augustjk: Should we keep the existing `shadowroot: mode` as well for compatibility with older Chrome versions? Per Elliott's testing, it seems fine to have both present and will have the streaming behavior for supported browsers. The streaming DSD feature that also renames the attribute is estimated to ship in Chrome 111 https://chromestatus.com/feature/5161240576393216

packages/labs/ssr-react/src/jsx-runtime.ts
augustjk: I like the explicit better and the exports here really shouldn't change.

packages/labs/ssr-react/src/jsx-dev-runtime.ts
augustjk: Good catch for the `Fragment`. Added. I think it's better to be explicit rather than `export *` so we know it exports everything that's needed. It doesn't have a separate `jsxs` because in dev mode the `jsxDEV` serves both purpose and gets a `isStaticChildren` boolean as a fourth argument instead instead.

There was no canonical doc that specifies this contract. Things I referenced were:
- React's RFC: https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md
- Babel's release note: https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154httpsgithubcombabelbabelpull11154
- React's own export: https://github.com/facebook/react/blob/9d111ffdfbcfee4b348a3d49c16f02cb718c896f/packages/react/jsx-dev-runtime.js
- Emotion's implementation: https://github.com/emotion-js/emotion/blob/acb72a45592881d9d1f72003b6db2e488b981599/packages/react/src/jsx-dev-runtime.js
- theme-ui's implementation: https://github.com/system-ui/theme-ui/blob/2d8bfc3ab29e54e925095ba0b5dbdb07bf7209db/packages/core/src/jsx-dev-runtime.ts

packages/labs/ssr-react/package.json
augustjk: Done

packages/labs/ssr-react/src/jsx-dev-runtime.ts
justinfagnani: I ask about `export *` because if this module needs to implement a certain contract that we know that `react/jsx-dev-runtime` does, then the easiest way to guarantee that we meet the contract is to `export *` that module. It fits the spirit of monkey-patching that module, which is kind of what we're emulating.

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
justinfagnani: I think we can launch with just the new attribute considering that we have no existing users. We'd have to deprecate the old attribute right away anyway.

augustjk: I'm hesitant on launching with an attribute that doesn't work on any currently shipping stable browsers. Our core labs/ssr package also does both which means nested custom elements will get both attributes. We should match that behavior and deprecate the old attr at the same time as when we do so on in labs/ssr.

justinfagnani: "launch" is a loose term here. Does this PR put the package in releasable state? Chrome 111 goes beta next well and stable in 4 weeks. Safari TP will be promoted relatively soon.

Either way is ultimately fine, though I do want to remove the old attribute as soon as possib.e.

augustjk: This PR does put the package in a releasable state. I think removing the old attribute in sync with core labs/ssr makes sense.

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
```diff
@@ -0,0 +1,82 @@
...
+    props[attr.name] = attr.value;
+  }
+  return props;
+};
+
+/**
+ * Renders the shadow contents of the provided custom element type with props.
+ * Should only be called in server environments.
+ */
+export const renderShadowContents = (type: string, props: {} | null) => {
```
kevinpschaaf, line 32: There's _almost_ nothing React-specific about this, right? This is generally useful because it wraps up the minimum steps needed to transform a tagName, attrs, and props from a call site into the reflected attrs and SR contents using ElementRenderer (something others have stumbled over as well). e.g. the Astro integration could use the same logic.

Can we move this to `labs/ssr`, with an additional `attrs` argument, and then this file would only add specific logic for separating React props into CE attrs & props?

kevinpschaaf: Naming nit: since this is more than just rendering "shadow contents" (technically it handles props -> reflected attributes as well), `renderCustomElement` might be better? Also `type` -> `tagName` would be clearer.

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      if (shadowContents !== undefined) {
+        const templateShadowRoot = originalCreateElement('template', {
+          ...templateAttributes,
+          dangerouslySetInnerHTML: {
+            __html: [...shadowContents].join(''),
+          },
+        });
+
+        return originalCreateElement(
```
kevinpschaaf, line 43: Should be outside the `if (shadowContents !== undefined)` test? We should still render a CE that doesn't render a SR for whatever reason.

packages/labs/ssr-react/src/node/jsx-runtime.ts
```diff
@@ -0,0 +1,80 @@
...
+      renderShadowContents(type, props);
+
+    if (shadowContents) {
+      const templateShadowRoot = createElement('template', {
+        ...templateAttributes,
+        dangerouslySetInnerHTML: {
+          __html: [...shadowContents].join(''),
+        },
+      });
+
```
kevinpschaaf, line 67: This code is duplicated 4x, is there a way to factor it better? This can't all be just `wrapCreateElement` I guess because there are slight differences in how children, keys, etc. are handled between the different fns, but maybe this inner part could be factored?

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
augustjk: It would fall back to the `return originalCreateElement(type, props, ...children)` outside the `if (isCustomElement(type))` block in that case.

packages/labs/ssr-react/src/node/jsx-runtime.ts
augustjk: The `renderShadowContents()` (or maybe now-to-be-called `renderCustomElement()`) was as much commonality I could pull out at the time. Though I think there might be a better way to deduplicate all of this. I'm thinking of actually creating a custom React element to handle the adding of template shadowroot so it would not need to have bespoke handling of the different transforms. I'll see if this idea pans out and push up a commit with it.

justinfagnani: @augustjk I would be tempted to TODO the deduplication and try a component in a follow-on PR.

augustjk: Yeah, it would just be implementation details and not affect the outward API so seems okay to punt that. @aomarks had the same thought while discussing offline.

aomarks (approved):
Awesome!

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
augustjk: Renamed. I agree that this could very well be part of labs/ssr but want to punt it to make sure we can separately discuss the proper API for making it more general use.

packages/labs/ssr-react/README.md
```diff
@@ -0,0 +1,94 @@
...
+- For Babel: set the [`pragma`](https://babeljs.io/docs/en/babel-preset-react#pragma) option for `@babel/preset-react` to `"createElement"`.
+- For TypeScript: set the [`jsxFactory`](https://www.typescriptlang.org/tsconfig#jsxFactory) option in `tsconfig.json` to `"createElement"`.
+
+Note that the import line must still be present for every file that contains JSX expressions to transform in the classic runtime mode.
+
+This approach only works for server-rendering custom elements added to the project in JSX expressions. It will not affect any pre-compiled JSX expressions or direct calls to `React.createElement()`. You will also need to manually import the `lit/experimental-hydrate-support.js` to your client JS. For those scenarios, use the [monkey patching](#monkey-patching-reactcreateelement-recommended) approach.
+
+### Using the Automatic Runtime JSX Transform
+
+If your project is using the [runtime JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), this package can serve as the JSX import source.
```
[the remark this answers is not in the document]
augustjk: It should be a very unlikely use case that affects external React components pre-compiled with auto runtime transform that directly make use of Lit components without the labs/react wrapper. But added a couple sentences to the bottom of this section regarding that use case.

packages/labs/ssr-react/src/lib/node/render-shadow-contents.ts
augustjk: Filed https://github.com/lit/lit/issues/3656

PR #3607, opened 2023-01-26
[infra] Skip sauce tests on forks

augustjk:
The condition here seems to be working
https://github.com/webcomponents/polyfills/blob/09b97dedf1b078f307259f5baaf8dbe23055a6ff/.github/workflows/tests.yml#L46-L48

### DIFF
--- a/.github/workflows/sauce-ie11.yaml
+++ b/.github/workflows/sauce-ie11.yaml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   tests-sauce-ie11:
     # Don't run on forks. Sauce tests need access to secrets.
-    if: github.repository == 'lit/lit'
+    if: github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
 
     # Sauce tests are still unreliable. Run them, but don't mark a commit as
     # failing if it fails.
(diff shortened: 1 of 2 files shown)


.github/workflows/sauce-ie11.yaml
```diff
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   tests-sauce-ie11:
     # Don't run on forks. Sauce tests need access to secrets.
-    if: github.repository == 'lit/lit'
```
justinfagnani, line 8: Can you explain this one? This is true on forks? IMO, worth a comment on the new version why it isn't the seemingly obvious `.repository == 'lit/lit'`

augustjk: We want to specifically compare the PR's head repo (which could be a fork) and the base repo (lit/lit). PRs whose head repo is a fork would fail this condition and skip.

The existing `github.repository == 'lit/lit'` will always be true for PRs open in this repository.

justinfagnani: Ah... Thanks for the explanation!

justinfagnani (approved):

Issue #3051, opened 2022-06-16
[labs/virtualizer] rangeChanged event stops propagation on itself


najikahalsema:
VisibilityChanged and RangeChanged events bubble and compose differently. The rangeChanged event stops propagation on itself which doesn't allow for its host to interact with the event. 

[...]

Issue #2788, opened 2022-04-26
[labs/virtualizer] Add unit tests covering basic functionality


kevinpschaaf:
See if we can port the unit tests over the monorepo (that steven created). Can we reuse any of the tests and infrastructure? Create any new unit tests needed.

usergenic:
Pre-requisite step completed: Migrated to Web Test Runner
https://github.com/PolymerLabs/uni-virtualizer/pull/137

Addition of new tests are pending #2787 migration to `lit` monorepo, now in-progress.

usergenic:
This issue is dragging across sprints and I think I might need to break it up to track progress better.  Before I do that, I'll enumerate the areas I'm trying to cover:

1. Event triggering:
    - rangeChanged
    - visibilityChanged
2. Layout-specific:
    - flow
    - grid
3. scrollToIndex behaviors
4. virtualize directive vs lit-virtualizer

Am still exploring what aspects of size estimation, reflowing and caching expectations can be tested.  Will definitely make separate issue for those.

usergenic:
Remaining coverage items:
- adding grid specific behavior
- margin collapse specific behavior for flow

PR #3609, opened 2023-01-26
[labs/virtualizer] RangeChangedEvent and VisibilityChangedEvent no longer bubble up

usergenic:
This PR solves the event behavior differences between RangeChangedEvent and VisibilityChangedEvent per #3051 

Both event types are ensured not to bubble up and must be listened for now on the `<lit-virtualizer>` itself or the host element in the case of `virtualize()` directive.

### DIFF
--- a/packages/labs/virtualizer/src/events.ts
+++ b/packages/labs/virtualizer/src/events.ts
@@ -11,7 +11,7 @@ export class RangeChangedEvent extends Event {
   last: number;
 
   constructor(range: Range) {
-    super(RangeChangedEvent.eventName, {bubbles: true});
+    super(RangeChangedEvent.eventName, {bubbles: false});
     this.first = range.first;
     this.last = range.last;
   }
@@ -24,7 +24,7 @@ export class VisibilityChangedEvent extends Event {
   last: number;
 
   constructor(range: Range) {
-    super(VisibilityChangedEvent.eventName, {bubbles: true});
+    super(VisibilityChangedEvent.eventName, {bubbles: false});
     this.first = range.first;
     this.last = range.last;
   }
--- a/packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
@@ -7,15 +7,17 @@
 import {array, ignoreBenignErrors, until} from '../helpers.js';
 import {LitVirtualizer} from '../../lit-virtualizer.js';
 import {virtualize} from '../../virtualize.js';
+import {RangeChangedEvent, VisibilityChangedEvent} from '../../events.js';
 import {css, LitElement} from 'lit';
 import {customElement, property} from 'lit/decorators.js';
 import {expect, html, fixture} from '@open-wc/testing';
 
 abstract class TestElement extends LitElement {
   static styles = css`
-    :host {
-      display: block;
-      height: 200px;
+    .item {
+      height: 50px;
+      margin: 0;
+      padding: 0;
     }
   `;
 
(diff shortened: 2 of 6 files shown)


usergenic:
CC: @westbrook @najikahalsema @graynorton this PR wraps up concerns discussed in issue #3051 about Virtualizer's event propagation

packages/labs/virtualizer/src/events.ts
```diff
@@ -11,7 +11,7 @@ export class RangeChangedEvent extends Event {
   last: number;
 
   constructor(range: Range) {
-    super(RangeChangedEvent.eventName, {bubbles: true});
+    super(RangeChangedEvent.eventName, {bubbles: false});
```
justinfagnani, line 14: Not for this PR, but can we get code comments on the event class and on `first` and `last` to document to users and maintainers what they're for? It can just be what's in the readme here: https://github.com/lit/lit/tree/main/packages/labs/virtualizer#visibilitychanged-event

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -24,29 +26,52 @@ abstract class TestElement extends LitElement {
 
   @property({type: Array, attribute: false})
   public items: Array<number> = [];
+
+  @property({type: Array, attribute: false})
```
justinfagnani, line 30: `type` is only used for attribute converters (it would be more accurately named `typeHint`), so if a property has `attribute: false` it doesn't need a type at all. For test elements, it's academic though and you could just leave all properties off. The converter would only kick in if reflecting to an attribute (which is off) or setting from an attribute, which you're not doing.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -96,45 +129,67 @@ describe('lit-virtualizer and virtualize directive', () => {
     const ulv: UsingLitVirtualizer = example.querySelector(
       'using-lit-virtualizer'
     )!;
+    const uvd: UsingVirtualizeDirective = example.querySelector(
+      'using-virtualize-directive'
```
justinfagnani, line 133: For my edification, do you know why we need to wait for these elements first with the `until()` call above? The elements are defined in the module, and before the `fixture` call, which itself is awaited. I wouldn't think the `until()`s are needed at all. but if there was a need to wait for an upgrade, `customElements.whenDefined('using-virtualize-directive')` should work too.

I'm asking because the test is so long that any reduction in boilerplate will help new people read it.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -96,45 +129,67 @@ describe('lit-virtualizer and virtualize directive', () => {
...
+    expect(ulv.shadowRoot?.textContent).to.include('[1 selected]');
+    expect(uvd.shadowRoot?.textContent).to.include('[1 selected]');
 
-    await until(() => uvd.shadowRoot?.textContent?.includes('2 selected'));
+    expect(ulv.shadowRoot?.textContent).to.include('[3 selected]');
+    expect(uvd.shadowRoot?.textContent).to.include('[3 selected]');
 
-    expect(uvd.shadowRoot?.textContent).to.include('2 selected');
-    expect(uvd.shadowRoot?.textContent).to.include('5 selected');
+    expect(ulv.shadowRoot?.textContent).not.to.include('[2 selected]');
```
justinfagnani, line 164: Maybe not now, but since the test is so long, it'd be great to see some comments break it up and describe the scenario, what the setup is doing, what the interaction is, expectations are checking, etc.

justinfagnani (approved):
LGTM!

The comments are just some musings from not being familiar with the code. This can go in as-is.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
usergenic: Fair points.  I'll add this when I do some overall test sprucing up in #2788

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
usergenic: Yeah this was how I was watching for upgrades from the outside.  I didn't want to select the element until after the upgrade occurred, but I'll see if I can either eliminate the need for those, write a helper, or comment why it's necessary to explicitly await the upgrades for the elements when I do test clean ups in #2788

PR #3613, opened 2023-01-27
[labs/nextjs] Add Next.js plugin for Lit SSR

### LINKED ISSUE #3589: [labs/nextjs] A plugin for Next.js to integrate Lit SSR

augustjk:
To follow #3605 

Resolves #3589 

Lacking tests. It probably makes more sense to test this e2e with the example project. PR that updates/adds usage example will follow.

A known issue is that due to the shimmed custom elements registry throwing an error on repeat registrations, when running `npm run dev`, the re-run after modifying a file can cause errors that's only cleared by restarting the dev server.


packages/labs/nextjs/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```js
+// next.config.js
+const withLitSSR = require('@lit-labs/nextjs')();
+
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  // Add your own config here
+  reactStrictMode: true,
+  swcMinify: true,
```
aomarks, line 21: Are these required?

packages/labs/nextjs/README.md
```diff
@@ -0,0 +1,37 @@
...
+};
+
+module.exports = withLitSSR(nextConfig);
+```
+
+## Considerations
+
+The plugin has been tested with Next.js versions 12 and 13. It currently does not support usage with the experimental `app` directory introduced with version 13.
```
aomarks, line 29: Is there an issue filed for this? Maybe file and mention here?

packages/labs/nextjs/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+The plugin may not work properly if you are providing a custom webpack configuration that modifies `config.externals`. Please file an issue if you find a particular configuration is not working.
+
+The server rendered output contains HTML with declarative shadow DOM which has limited browser support. See [`@lit-labs/ssr-react`](../ssr-react/README.md#other-considerations) for more information.
```
aomarks, line 33: Maybe slightly rephrase to say that some browsers still need a polyfill, since otherwise it sounds like it might not work at all.

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        config.plugins.push(
+          new webpack.NormalModuleReplacementPlugin(
+            /react/,
+            function (resource: {request: string; context: string}) {
+              if (
+                resource.request === 'react/jsx-runtime' &&
+                // Don't replace our own imports
+                !/labs\/ssr-react/.test(resource.context)
```
aomarks, line 61: can this be `lit-labs\/ssr-react` instead of just `labs\/ssr-react`? Same Q for below.

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+              ) {
+                resource.request = '@lit-labs/ssr-react/jsx-dev-runtime';
+              }
+            }
+          )
+        );
+
+        // There's potential for conflict here if existing `webpack` config also
+        // overwrites externals
+        if (typeof nextConfig.webpack === 'function') {
```
aomarks, line 78: What is going on here? The user can pass their own webpack?

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+          // Grab entry points for all pages
+          // TODO(augustjk) This may not work for the new "app" directory
+          // introduced in Next.js v13
+          test: /\/pages\//,
+          // Exclude Next's own distributed files as they're commonjs and won't
+          // play nice with `imports-loader`
+          exclude: /next\/dist\//,
+          loader: 'imports-loader',
+          options: {
+            imports: ['side-effects @lit-labs/ssr-react/enable-lit-ssr.js'],
```
aomarks, line 29: Maybe a few comments just to describe high level what's going on?

I think this line means that we'll load the module that monkey patches React.createElement?

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+                // Returning empty promise makes these requests go through webpack
+                return Promise.resolve();
+              }
+              return nextHandleExternals(opt);
+            },
+          ];
+        }
+
+        config.plugins.push(
+          new webpack.NormalModuleReplacementPlugin(
```
aomarks, line 55: High level comment for this section? This is where we intercept the new runtime modules and substitute our own, correct?

aomarks (changes requested):
Looks good! Minor comments.

packages/labs/nextjs/README.md
augustjk: Not required, but they are the default configs that are present when creating a new Next.js project with `create-next-app`.

packages/labs/nextjs/src/index.ts
augustjk: The `resource.context` contains the file path. For downstream users, this would indeed be `@lit-labs/ssr-react` but within our monorepo it is `packages/labs/ssr-react`. I grabbed the `labs/ssr-react` portion to catch both such that tests run within the monorepo would still work. I could add this as a comment.

packages/labs/nextjs/src/index.ts
augustjk: Since we're wrapping/adding to the user provided next config, users could have a `webpack` as part of their own next config (which needs to be a function). This is the same pattern used by the Next team's own plugin here: https://github.com/vercel/next.js/blob/b2ac41c75493ab759bd8ecfd7ee111dc66c2db5c/packages/next-mdx/index.js#L38-L40

I should also clarify that better in the comment.

aomarks (approved):
Pre-emptive LGTM, assuming my minor comments will be addressed.

packages/labs/nextjs/README.md
augustjk: Done

packages/labs/nextjs/README.md
augustjk: Done. Also updated the link to match the correct heading in the other package.

packages/labs/nextjs/src/index.ts
augustjk: Done.

packages/labs/nextjs/src/index.ts
augustjk: Done.

khanh-le-otsv:
Really excited to see this coming! I'm building a lit design system for nextjs and this could be the solution that's I'm looking for!

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+export = (_pluginOptions: LitSsrPluginOptions = {}) =>
+  (nextConfig: NextConfig = {}) => {
+    return Object.assign({}, nextConfig, {
+      webpack: (config, options) => {
+        const {isServer, nextRuntime, webpack} = options;
+        config.module.rules.unshift({
+          // Grab entry points for all pages
+          // TODO(augustjk) This may not work for the new "app" directory
+          // https://github.com/lit/lit/issues/3657
+          test: /\/pages\//,
```
salman-mazhar, line 23: Would it make sense to test for `ts` and `tsx` files only here. We are using SCSS modules from nextjs which leads to scss files under our pages directory (https://nextjs.org/docs/basic-features/built-in-css-support#sass-support) and the loader seems to break these imports. Restricting the test to ts and tsx files only here fixes the issue for us.

augustjk: Thank you for catching and reporting that! Added extensions js,jsx,ts,tsx to the test.

packages/labs/nextjs/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        });
+
+        if (isServer && nextRuntime === 'nodejs') {
+          // Next.js uses this externals setting to skip webpack bundling
+          // external code for server environments but we specifically want to
+          // catch react/jsx-runtime imports for external packages as well.
+          // This might change in a future update to Next.js.
+          // This is what we're grabbing:
+          // https://github.com/vercel/next.js/blob/412dfc52cc5e378e4f74e44af698dad25031a938/packages/next/src/build/webpack-config.ts#L1378-L1429
+          const nextHandleExternals = config.externals[0];
```
kevinpschaaf, line 43: Since this is/was an array, is it safe to only call `nextHandleExternals`? Will it always be the one-element array referenced in the link above?

kevinpschaaf (approved):
One minor question, otherwise LGTM (to the extend I grok it all).

packages/labs/nextjs/src/index.ts
augustjk: Next could choose to change their default webpack config but for now it's a single element array. If the user provides their own enhancement of the webpack config, or composes with another plugin that does, that could be problematic as I describe in the comment lower in the code. It's brittle but I think hard to anticipate that kind of composition so I think we'll have to address that if we get reports of that.

justinfagnani: Could you file an issue with Next on how to do this in a safe way? Maybe they could provide a hook for overriding/wrapping handleExternals.

packages/labs/nextjs/README.md
```diff
@@ -0,0 +1,39 @@
+# @lit-labs/ssr-react
+
+A package for integrating Lit SSR with Next.js.
```
justinfagnani, line 3: micro nit: Kind of like how you should start a classes jsdoc with "A class that ...", everything on npm is a package, so you can say "Integrates Lit SSR..."

```suggestion
Integrates Lit SSR with Next.js to enable deep server rendering of Lit components.
```

packages/labs/nextjs/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Overview
+
+Lit components, by default, can be imported and added to Next.js projects but will only be shallowly rendered on the server.
```
justinfagnani, line 7:
```suggestion
Lit components can be imported and added to Next.js projects but by default they will only be _shallowly_ rendered on the server. That is, the Lit component's tag and attributes set via JSX will be rendered, but the component's shadow DOM will not be.
```

justinfagnani (approved):

packages/labs/nextjs/src/index.ts
augustjk: Done!

PR #3614, opened 2023-01-27
[examples/nextjs] Show usage of `@lit-labs/nextjs`

augustjk:
To follow #3613 

I'm wondering if it might be better to keep the base `nextjs` example without the plugin to show that it works for shallow rendering and adding a separate example, maybe `nextjs-with-lit-ssr`, that uses the plugin.

Screenshot below showing initially served HTML includes declarative shadow dom.
image


aomarks:
I think this needs a merge or rebase? Seeing more diffs than expected.

augustjk:
Included in #3613

PR #3615, opened 2023-01-27
Don't throw if ChildPart.parentNode is null

### LINKED ISSUE #3611: [lit-html] get ParentNode should handle when parentNode is null

### Description

I recently had an issue with how I was working with a lit component. In attempting to debug my problem, I encountered this. 
In the _$setValue method in lit-html.ts, there is an error thrown when this.parentNode === null 

```
    if (DEV_MODE && this.parentNode === null) {
      throw new Error(
        `This \`ChildPart\` has no \`parentNode\` and therefore cannot accept a value. This likely means the element containing the part was manipulated in an unsupported way outside of Lit's control such that the part's marker nodes were ejected from DOM. For example, setting the element's \`innerHTML\` or \`textContent\` can do this.`
      );
    } 
```

However, I was not seeing this useful error because in get ParentNode() in the same file, the first if statement checks the nodeType of the parentNode which fails because the parentNode is null and I got a less helpful error. It seems that if we can have this.parentNode === null, a null value for the parentNode should be handled in the get ParentNode method.

```
get parentNode(): Node {
    let parentNode: Node = wrap(this._$startNode).parentNode!;
    const parent = this._$parent;
    if (
      parent !== undefined &&
      parentNode.nodeType === 11 /* Node.DOCUMENT_FRAGMENT */
    ) { 
```

### Reproduction

1. Create a lit component that will exit and reenter the document
2. Have that lit component render a second lit component in the template
3. Foolishly change the innerText of the parent lit component to remove the child in the Updated method
4. Debug 

### Workaround

I have not found a workaround

### Is this a regression?

### Affected versions

2.6.1

Browser: Chrome 109.0.5414.87
macOS Ventura 13.1

justinfagnani:
Fixes #3611

This seems like it should be a rare edge-case, but a user hit it. The part parent needs to be non-null and the part's parent node needs to be null to hit the issue.

### DIFF
--- a/.changeset/early-snakes-complain.md
+++ b/.changeset/early-snakes-complain.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Don't throw in ChildPart.parentNode if the parentNode is nullwq
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1318,7 +1318,7 @@ class ChildPart implements Disconnectable {
     const parent = this._$parent;
     if (
       parent !== undefined &&
-      parentNode.nodeType === 11 /* Node.DOCUMENT_FRAGMENT */
+      parentNode?.nodeType === 11 /* Node.DOCUMENT_FRAGMENT */
     ) {
       // If the parentNode is a DocumentFragment, it may be because the DOM is
       // still in the cloned fragment during initial render; if so, get the real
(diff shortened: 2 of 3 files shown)


.changeset/early-snakes-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Don't throw in ChildPart.parentNode if the parentNode is nullwq
```
augustjk, line 5: there was an attempt to exit vim :)
```suggestion
Don't throw in `ChildPart.parentNode` if the `parentNode` is null
```

sorvell (approved):

PR #3624, opened 2023-01-31
[labs/virtualizer] Simplify initialization, more complete fix for #3481

graynorton:
- No longer use events to communicate between layout and virtualizer
- Remove dependency on `EventTarget` polyfill
- No longer auto-load `ResizeObserver` polyfill (necessitating major version bump)
- Init is now entirely sync, except for auto-loading of default layout as needed
- More complete fix for #3481 (handle immediate layout change)


usergenic:
@graynorton I'll look at publishing the ResizeObserver polyfill and removing the included loader and version entirely in a future version, but I like the steps this takes towards that.  And there's ResizeObserver polyfill+native behavior tests to boot! Nice!

I think I understand the move away from CustomEvents being dispatched by the layouts to an explicit Layout Message channel too as it is more intention-revealing.  Makes it clearer where they're coming from and why.

usergenic (approved):
LGTM!

packages/labs/virtualizer/README.md
```diff
@@ -285,6 +285,65 @@ export class MyItems extends LitElement {
...
+The package also includes a simple mechanism for loading the `ResizeObserver` polyfill. This mechanism uses feature detection to determine whether the polyfill is required; if it is, then the polyfill is automatically loaded and provided directly to Virtualizer.
+
+You need to invoke the loader—which is aysnchronous—and await its return before running any code that would cause a virtualizer to be instantiated. The simplest way to do this is to load virtualizer itself with a dynamic `import()` statement. For example:
+
+```js
+import {loadPolyfillIfNeeded} from '@lit-labs/virtualizer/polyfillLoaders/ResizeObserver.js';
+
+async function load() {
+  await loadPolyfillIfNeeded();
+  await import('@lit-labs/virtualizer');
```
justinfagnani, line 303: It might be instructive to show this happening at the app level, like before dynamic importing the main app code in a bootstrap script. You could use top-level await there.

packages/labs/virtualizer/src/Virtualizer.ts
```diff
@@ -223,6 +223,13 @@ export class Virtualizer {
   private _layoutCompleteRejecter: Function | null = null;
   private _pendingLayoutComplete: number | null = null;
 
+  /**
+   * Layout initialization is async because we dynamically load
```
justinfagnani, line 227: How big is the default layout? Could we just statically import it?

packages/labs/virtualizer/README.md
```diff
@@ -285,6 +285,65 @@ export class MyItems extends LitElement {
 }
 ```
 
+## ResizeObserver dependency
+
+Virtualizer depends on the standard [`ResizeObserver`]() API, which is supported in all modern browsers. In case your browser support matrix includes older browsers that don't implement `ResizeObserver`, the Virtualizer package includes a `ResizeObserver` polyfill that is known to be compatible with Virtualizer.
```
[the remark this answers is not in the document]
graynorton: It behaves like native `ResizeObserver` with respect to observing changes within shadow roots. At the time I was investigating, neither of the two widely known / used libraries did so.

packages/labs/virtualizer/README.md
graynorton: I considered that briefly, but was struggling with doing so in a way that didn't seem overly precise / prescriptive or make too many assumptions about typical app structure, so I ended up falling back to this minimal, more conceptual sketch. Feel free to propose (or just go ahead and make) a better example.

packages/labs/virtualizer/src/Virtualizer.ts
graynorton: Can dig into the size numbers and reconsider the tradeoffs here, but the default `flow` layout is non-trivial in size, and previous thinking was that it was definitely worth not forcing folks to pay that cost if they were using an alternative layout.

Another option, which we actually briefly tried out at some point in the `0.x` history, is to always require a layout to be explicitly loaded and provided by the app. The DevEx hit for the 80% case was unfortunate, though, so we reverted to dynamically auto-loading.

At this point, with all of the async polyfill stuff out of the picture and some kinks worked out, I think we've converged on a dynamic loading scheme that isn't too heavy and shouldn't cause much trouble going forward, so my inclination is to keep this as-is for now.

graynorton:
> I think I understand the move away from CustomEvents being dispatched by the layouts to an explicit Layout Message channel too as it is more intention-revealing. Makes it clearer where they're coming from and why.

@usergenic Yeah, I think it is clearer and more direct (and we weren't really benefiting from the indirection), but the main reason I did it as part of this particular PR was to eliminate the use of `EventTarget` and the associated polyfill. In doing this at the same time as we stop auto-loading the `ResizeObserver` polyfill, we get rid of all of the async initialization code except for the default layout loading.

usergenic:
I'm not certain why the tests-local and benchmarks checks are failing.  The failures are in the localize examples `/home/runner/work/lit/lit/packages/localize/examples/transform-js/rollup.config.mjs` is importing `@rollup/plugin-typescript` which "can't be found".

I don't get that error running any builds/tests locally.  As soon as I can get the check passing I'll get this merged in.

packages/labs/virtualizer/README.md
```diff
@@ -285,6 +285,65 @@ export class MyItems extends LitElement {
...
+
+### Using the default loader
+
+The package also includes a simple mechanism for loading the `ResizeObserver` polyfill. This mechanism uses feature detection to determine whether the polyfill is required; if it is, then the polyfill is automatically loaded and provided directly to Virtualizer.
+
+You need to invoke the loader—which is aysnchronous—and await its return before running any code that would cause a virtualizer to be instantiated. The simplest way to do this is to load virtualizer itself with a dynamic `import()` statement. For example:
```
43081j, line 296: small spelling error here, should be `asynchronous`

packages/labs/virtualizer/src/layouts/shared/BaseLayout.ts
```diff
@@ -453,51 +426,49 @@ export abstract class BaseLayout<C extends BaseLayoutConfig> implements Layout {
...
+      for (let idx = this._first; idx <= this._last; idx++) {
+        childPositions.set(idx, this._getItemPosition(idx));
+      }
+    }
+    const message: StateChangedMessage = {
+      type: 'stateChanged',
+      scrollSize: {
+        [this._sizeDim]: this._scrollSize,
+        [this._secondarySizeDim]: null,
+      } as Size,
```
43081j, line 455: tiny nit: it seems the virtualiser actually handles nullish dimensions. is it worth updating `Size` to reflect that and drop the cast here?

in the virtualiser it already checks `size.width !== null` etc even though the type says it can't ever be `null`

near enough a non-issue but maybe still worth me pointing out

graynorton: This is a good observation; however, I think it might be a longer thread than we want to pull on as part of this PR. The `Size` type is used in quite a few places in the Virtualizer code base, and I am quite certain that in some of them the logic doesn't anticipate or guard against `null` values. My inclination is to take care of this independently of this PR, perhaps as part of a more systematic review of internally used types. @usergenic, thoughts?

packages/labs/virtualizer/README.md
graynorton: Fixed, thanks!

packages/labs/virtualizer/src/layouts/shared/BaseLayout.ts
usergenic: @graynorton @43081j I'm all for considering making Size nullable as separate PR.  It will give time to fully vet the change; there's enough going on here already.

43081j: Yup of course. Didn't mean to distract at all. Was mostly an observation 👍

usergenic:
This test failure is persistent now on CI but is not happening on my local machine which indicates a possible issue hidden by environment characteristics.  I'll look to see if test can be adjusted or will add a .skip and comment until such adjustment possible
```
test/scenarios/element-and-directive-parity.test.js:

 ❌ lit-virtualizer and virtualize directive > both render changes based on non-item data changes
      AssertionError: expected 2 to equal 1
      + expected - actual
      
      -2
      +1
      
      at o.<anonymous> (test/scenarios/element-and-directive-parity.test.js:[156](https://github.com/lit/lit/actions/runs/4070763955/jobs/7012246645#step:8:157):50)
```

usergenic:
Have adjusted the expectation in the test from an equal to a greaterThanOrEqual as it was been expecting a number of events that may not be completely deterministic since RangeChangedEvents are emitted in reaction to environmental factors like resize observers etc and timing differences may play a role.  Check is passing now :)

PR #3628, opened 2023-02-01
[context] Add type parameter for key to createContext()

justinfagnani:
Small change to get more accurate types. Should be non breaking as the second type parameter has a default and doesn't have to be specified.

cc @benjamind


sorvell (approved):

justinfagnani:
Thanks @sorvell!

PR #3636, opened 2023-02-03
Rearrange the context doc

vdegenne:
As a new comer, I thought the doc was confusing:

- I moved the "Providing" part before the "Consuming" part, as one would expect to see how to *Define* a context before consumption.
- I hid some "extension code" in details tags.
- I used horizontal separators to help identifying the sections.

### DIFF
--- a/packages/labs/context/README.md
+++ b/packages/labs/context/README.md
@@ -28,6 +28,8 @@ export interface Logger {
 export const loggerContext = createContext<Logger>('logger');
 ```
 
+---
+
 ### Providing (Defining) a Context
 
 We want to be able to provide this context from somewhere higher in the DOM.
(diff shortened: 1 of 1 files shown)


e111077 (approved):
I like this change

augustjk (approved):
I also like this change

.changeset/serious-chairs-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/context': patch
+---
+
+Dusting the doc
```
justinfagnani, line 5: This isn't a good description for a changelog.

Something like "Updated the README" is better.

justinfagnani (changes requested):

.changeset/serious-chairs-jam.md
vdegenne: Yes sorry for the ambiguous description. I changed it.

vdegenne:
@benjamind  Have you seen this review request?

I may request a view from @justinfagnani instead, I resolved the change you requested.

AndrewJakubowicz:
@justinfagnani Comments are addressed, could you re-review to unblock merge. Thank you so much!

PR #3639, opened 2023-02-04
[labs/context] Fix getter for @provide() decorated fields

justinfagnani:
Previously there was no getter, so `@provide()` would always result in reading `undefined` from a decorated field. This adds a getter that delegate to an inner getter or the current value of the ContextProvider controller.

### DIFF
--- a/packages/labs/context/src/test/context-provider_test.ts
+++ b/packages/labs/context/src/test/context-provider_test.ts
@@ -7,10 +7,11 @@
 import {LitElement, html, TemplateResult} from 'lit';
 import {property} from 'lit/decorators/property.js';
 
-import {Context, consume, provide} from '@lit-labs/context';
+import {Context, createContext, consume, provide} from '@lit-labs/context';
 import {assert} from '@esm-bundle/chai';
 
 const simpleContext = 'simple-context' as Context<'simple-context', number>;
+const stringContext = createContext<string>(Symbol);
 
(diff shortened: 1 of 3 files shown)


andrico1234:
Does this solve the [following issue](https://discord.com/channels/1012791295170859069/1064230566360256712/1064442847148920873)?

I flagged in the Discord that the `@provide` decorator doesn't clone getters correctly when cloning the context object?

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -136,3 +152,27 @@ suite('@consume: multiple instances', () => {
     );
   });
 });
+
+suite('@provide', () => {
+  test('can get values of provide decorated fields', async () => {
```
rictic, line 157: get and set

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -136,3 +152,27 @@ suite('@consume: multiple instances', () => {
...
+    const container = document.createElement('div');
+    container.innerHTML = `
```
rictic, line 159: we're already importing Lit, let's use it here, just to avoid issues later if we want to test with trusted types enabled

rictic (approved):

augustjk:
This has been superseded by the conversion to hybrid decorators

PR #3642, opened 2023-02-04
[all] Try limiting action runs

justinfagnani:
This a test to see if we can avoid doubling or test jobs where we run on both `push` and `pull_request`.

This runs tests on `push` to `main`, and `pull_request`. I believe that `pull_request` has it's SHA set to the merge commit of the PR branch and main, which should test on how it would merge, which is ultimately what we care about and gate merging on. If that fails, but the branch was passing locally, the dev can merge in main locally and see what the failure is. Outside of flakiness, I don't think I've seen a PR fail on push, but pass on pull_request or vice versa anyway.


justinfagnani:
So far, I think this seems like a good thing to do to reduce our actions runs. It helps blunt the impact of flakiness a little at least.

augustjk (approved):
I agree. The duplicate runs for push and pull_request on PRs seemed excessive. The only thing to be aware is workflow won't run when there's a merge conflict but those need to be resolved before merging anyway, and the resolution would trigger the workflow.

kevinpschaaf (approved):
LGTM

aomarks (approved):

PR #3648, opened 2023-02-06
[labs/analyzer] Fix support for static class members.

kevinpschaaf:
We were previously storing class members (fields & methods) in a map by name (since field-lookup by name is expected to be a key use case), however this ignored the fact that static and non-static members would conflict.

This PR separates the storage into separate maps, and adds an `isStatic` optional argument to `getField()` and `getMethod()`.

### DIFF
--- a/packages/labs/analyzer/src/lib/model.ts
+++ b/packages/labs/analyzer/src/lib/model.ts
@@ -407,7 +407,9 @@ export interface ClassDeclarationInit extends DeclarationInit {
   node: ts.ClassDeclaration;
   getHeritage: () => ClassHeritage;
   fieldMap?: Map<string, ClassField> | undefined;
+  staticFieldMap?: Map<string, ClassField> | undefined;
   methodMap?: Map<string, ClassMethod> | undefined;
+  staticMethodMap?: Map<string, ClassMethod> | undefined;
 }
 
 export class ClassDeclaration extends Declaration {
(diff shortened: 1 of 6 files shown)


packages/labs/analyzer/src/lib/model.ts
```diff
@@ -438,31 +444,31 @@ export class ClassDeclaration extends Declaration {
...
+    return [...this._fieldMap.values(), ...this._staticFieldMap.values()];
   }
 
   /**
    * Returns iterator of the `ClassMethod`s defined on the immediate class
    * (excluding any inherited members).
    */
-  get methods(): IterableIterator<ClassMethod> {
-    return this._methodMap.values();
+  get methods() {
```
bicknellr, line 454: Does removing this return type cause it to implicitly be `Array<ClassMethod>`? Is that something we care about w.r.t. API stability / flexibility?

bicknellr (commented):
WDYT about splitting the four getters into separate ones for static vs. member? I get the feeling that it's more likely that a user would care about only one group at a time since they represent properties of different objects, which might avoid extra (e.g.) `classDecl.fields.filter(x => !x.isStatic)`.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: It does, and I don't think we care that much about stability at this exact point. It's not possible to make a generator getter directly (where we could just yield both values() out; it would need to call out to one), and didn't figure it was worth the trouble to keep this an iterator.

kevinpschaaf:
> WDYT about splitting the four getters into separate ones for static vs. member? I get the feeling that it's more likely that a user would care about only one group at a time since they represent properties of different objects, which might avoid extra (e.g.) `classDecl.fields.filter(x => !x.isStatic)`.

Yeah, that's a good point. It's likely a documentation generator wants to ensure the static and non-static members are collated, so it'd need to do just that I think. I'll separate them.

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Oh, separating them allows keeping that return type I guess. Might as well.

bicknellr (approved):

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -434,35 +440,51 @@ export class ClassDeclaration extends Declaration {
...
+    return this._staticMethodMap.values();
+  }
+
   /**
    * Returns a `ClassField` model the given name defined on the immediate class
    * (excluding any inherited members).
    */
-  getField(name: string): ClassField | undefined {
-    return this._fieldMap.get(name);
+  getField(name: string, isStatic = false): ClassField | undefined {
```
justinfagnani, line 478: Given that we have `fields` and `staticFields`, I'm not sure what value there is in a parameter to separate the two here. I would default to mirroring the two APIs and making `getStaticField()`, etc.

justinfagnani (approved):

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Sgtm, done

PR #3650, opened 2023-02-07
[infra] Pin polyfill version for tests

augustjk:
IE sauce tests have been failing since the polyfill version was updated. This PR pins the version to the one listed in our WTR config file here: https://github.com/lit/lit/blob/1726e236d4d842b56ec1c26b276a9081961a152c/packages/tests/src/web-test-runner.config.ts#L202

We may want to investigate why the test fails with the newer version though.


aomarks (approved):

augustjk:
Looks like the sauce ie11 tests are still failing with the same error so my theory on the polyfill version might be wrong.

PR #3658, opened 2023-02-08
[labs/analyzer] Don't extract JSDoc types in TS. Port type tests to JS.

bicknellr:
TS doesn't consider types provided in JSDoc comments as having any effect in `.ts` files, even though they're still accessible in the AST. This PR updates the analyzer to match this behavior and also copies/ports the type tests to JS to confirm that JSDoc comments still affect the analysis of `.js` files.

This PR is easier to read with [leading whitespace removed](https://github.com/lit/lit/pull/3658/files?w=1).

### DIFF
--- a/packages/labs/analyzer/src/lib/types.ts
+++ b/packages/labs/analyzer/src/lib/types.ts
@@ -62,11 +62,8 @@ export const getTypeForNode = (
 ): Type => {
   // Since getTypeAtLocation will return `any` for an untyped node, to support
   // jsdoc @type for JS (TBD), we look at the jsdoc type first.
-  const jsdocType = ts.getJSDocType(node);
   return getTypeForType(
-    jsdocType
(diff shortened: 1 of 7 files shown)


packages/labs/analyzer/src/lib/types.ts
```diff
@@ -62,11 +62,8 @@ export const getTypeForNode = (
 ): Type => {
   // Since getTypeAtLocation will return `any` for an untyped node, to support
   // jsdoc @type for JS (TBD), we look at the jsdoc type first.
```
bicknellr, line 64: Needs to be removed.

bicknellr: Done.

justinfagnani (approved):

kevinpschaaf (approved):

PR #3659, opened 2023-02-09
Multihosted Controllers

vdegenne:
This update allows two things:
- writing controllers with multihost capability
- create a controller, add and remove hosts later

E.g.
```ts
/** MyController.ts */

class MyController implements ReactiveController {
  private _hosts: ReactiveControllerHost[];

  addHost(host) {
    host.addController(this);
  }

  hostConnected(host) {
    (this._hosts ??= []).push(host);
  }

  hostDisconnected(host) {
    this._hosts?.splice(this._hosts.indexOf(host) >>> 0, 1);
  }

  updateAll() {
    this._hosts?.forEach((host) => host.requestUpdate());
  }
}

// pre-defined
export const myController = new MyController()
```

And then, in a custom element:

```ts
/** my-element.ts */

import { myController } from './MyController.ts';

@customElement('my-element')
  constructor() {
    super();
    myController.addHost(this)
  }
}
```

### DIFF
--- a/.changeset/brown-rivers-perform.md
+++ b/.changeset/brown-rivers-perform.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+MultiHosted Controller support
--- a/packages/reactive-element/src/reactive-controller.ts
+++ b/packages/reactive-element/src/reactive-controller.ts
@@ -56,7 +56,7 @@ export interface ReactiveController {
    * element hosts, this corresponds to the `connectedCallback()` lifecycle,
    * which is only called when the component is connected to the document.
    */
-  hostConnected?(): void;
+  hostConnected?(host: ReactiveControllerHost): void;
 
   /**
    * Called when the host is disconnected from the component tree. For custom
@@ -64,7 +64,7 @@ export interface ReactiveController {
    * which is called the host or an ancestor component is disconnected from the
    * document.
    */
-  hostDisconnected?(): void;
+  hostDisconnected?(host: ReactiveControllerHost): void;
 
   /**
    * Called during the client-side host update, just before the host calls
@@ -73,12 +73,12 @@ export interface ReactiveController {
    * Code in `update()` can depend on the DOM as it is not called in
    * server-side rendering.
    */
-  hostUpdate?(): void;
+  hostUpdate?(host: ReactiveControllerHost): void;
 
   /**
    * Called after a host update, just before the host calls firstUpdated and
    * updated. It is not called in server-side rendering.
    *
    */
-  hostUpdated?(): void;
+  hostUpdated?(host: ReactiveControllerHost): void;
 }
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -967,7 +967,7 @@ export abstract class ReactiveElement
     // (which is set in connectedCallback) to avoid the need to track a
     // first connected state.
     if (this.renderRoot !== undefined && this.isConnected) {
-      controller.hostConnected?.();
+      controller.hostConnected?.(this);
     }
   }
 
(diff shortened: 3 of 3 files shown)


justinfagnani:
Thanks for the contribution, though we really would like changes like this discussed in issues or RFCs first.

In this case I don't think this is a change we want to make - it makes the contract less clear. How does a user of a controller know if it supports multiple hosts? Controllers are supposed to be single-hosted. If you want to share state across elements each should have it's won controller linking that state tot he particular element.

One thing to note is that making controllers multi-host compatible doesn't close the loop on sharing, because while you could pass the controller around, maybe via data-binding, you still need to know when to call `addController()` and `removeController()`, which you don't just get via regular reactive properties. You'd probably need a custom decorator to manage that, and at that point the property could consume the state object and coordinate with a decorator instance that it creates to hook the lifecycle. This is how the labs/context decorators work - they create controllers as part of their initialization.

vdegenne:
I see some advantages of using multi-host binding controllers over contexts:

- **Syntax:**
  - **Intuitive:** It might be more direct and easier to think "controllers" as the name implies they control (something or many things). While contexts can be an excellent solution to share states across multiple elements, controllers aren't just limited to this scenario. (consider this dummy accordion example using the current controller implementation : https://t.ly/fteKI).
  - **Freedom:** Controllers are more intuitive to use and developers have more freedom over the implementation, typically one would make a module file containing all the logic. Later a host that want to make itself available under a controller can do it at anytime, anywhere it needs. `ReactiveController` is still an interface which lets devs have more control on how they want to make things work in a certain situation, passing the host instance in the reactive methods create even more possibilities (also this change still permits the creation of single-host controllers). On the other hand contexts force a specific and fixed syntax to adopt.
  - **Shareability:** For the reasons above controllers are great candidates for modularity, making the creation and the shareability of new tools easier.
- **Non-Contextual:** Because you can add and remove a host easily, that means you don't really need to overthink how to compose your elements in the tree. In contexts you have to make sure a consumer is a direct/indirect child of the provider. A controller on the other hand, just like a TV remote controller, should be free in space.
- **Interoperability:** It's easier to think controllers generally because they are interoperable, an element may want to activate another element or vice-versa based on a specific logic. In contexts consumers can be reactive to the context change, but only the provider can change the context. Using controllers one should be able to dictate how things interact with each other.

---

**In a nutshell**: Controllers are more primitive, while contexts are used in specific scenarios, a controller is a good starting block when one wants to start creating a new feature. A controller shouldn't limit itself to only controlling "one thing" (In fact we should technically be able to make a "context" feature using a controller as a base module only).  
I personally think more energy should be allocated to this feature in the Lit framework as most of the problems I had building apps using Lit could have been solved using controllers. But controllers lacks two things right now: They have no multi-host support and they have no mean to host reactive properties.


(Sorry for not opening a discussion before this PR, I was too excited to make this change)

yorrd:
I realize this is probably not going to be accepted without a proper RFC, but for the lack of a better place:

I really like the approach and support the notion. We often have data loading controllers which are custom lit controllers that are used in several of the custom elements in the project. Creating a new controller for each component and sharing the data source is unintuitive for new developers and has been a talking point in our team several times now. In a perfect world, we could

```
import { todosController } from './shared-controllers'

...
class El extends LitElement {
    todos = todosController.addHost(this); // possible if addHost returns the controller itself
}
```

this just seems so much cleaner than what we're forced to do at the moment along the lines of the following, where the `importedDataSource` needs to be deduplicated "magically" in the background. 
```
    todos = new DataLoadingController(this, importedDataSource);
```

justinfagnani:
@yorrd you can already do what you want by making a factory:

```ts
export const getTodos = (host) => new DataLoadingController(host, importedDataSource);
```

```ts
import { getTodos } from './todos.js';

class El extends LitElement {
    todos = getTodos(this);
}
```

sorvell:
As Justin mentioned, we would want to consider this type of change carefully via an RFC. As noted, part of the appeal of controllers is their primitive flexibility and we want to be pretty careful to maintain that. 

It's currently possible to implement the desired behavior here essentially with a factory pattern that proxies to some shared "controller" resource. Here's a [really simple example](https://lit.dev/playground/#gist=2f2c72c77c52a4b3ab521ffda6679a11) that should be pretty self explanatory. Hope that helps.

rictic:
Thank you for this PR, but as mentioned above, we'd need to get consensus on a change like this first. Please send an RFC to https://github.com/lit/rfcs if you'd like to continue this proposal

PR #3660, opened 2023-02-09
Fix change-in-update warning from Tasks

### LINKED ISSUE #3566: [labs/task] Task improperly warns user about rendering

Task (@lit-labs/task)

### Description

Using the default `render` from `Task` results in a console warning.

image

### Reproduction

The following class warns user there is an issue, despite just following default examples.

```js
export class WorldPosts extends LitElement {
  posts: Post[] = [];

  private _apiTask = new Task(this, {
    task: async () => {
      console.log('demo time');
      return [
        {
          message: 'hello',
        },
      ];
    },
     // this will autorun, but not run automatically on any subsequent updates since it tracks no args
    args: () => [],
  });

  render() {
    return html`
      ${this._apiTask.render({
        complete: (posts: Post[]) => {
          const list = posts.map(post => html`<li>${post.message}</li>`);

          return html`<ul>
            ${list}
          </ul>`;
        },
        error: console.error,
      })}
    `;
  }
}
```

### Workaround

Possibly overwriting update methods? Not sure

### Is this a regression?

### Affected versions

2.5.0

Browser: Chrome
OS: 13.1 (22C65)

justinfagnani:
Fixes #3566 

The fix is to call the initial `host.requestUpdate()` (which is there to get the element to update and see the PENDING state of the task) in a microtask, so that it goes after the change-in-update check.

### DIFF
--- a/.changeset/fast-apes-cross.md
+++ b/.changeset/fast-apes-cross.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': minor
+---
+
+Tasks now run on hostUpdate() instead of hostUpdated()
--- a/.changeset/fast-apes-cross.md
+++ b/.changeset/fast-apes-cross.md
@@ -2,4 +2,4 @@
 '@lit-labs/task': minor
 ---
 
-Tasks now run on hostUpdate() instead of hostUpdated()
+Fix the change-in-update warning from Tasks by delaying the initial host update
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -31,7 +31,7 @@ export const TaskStatus = {
  */
 export const initialState = Symbol();
 
-export type TaskStatus = typeof TaskStatus[keyof typeof TaskStatus];
+export type TaskStatus = (typeof TaskStatus)[keyof typeof TaskStatus];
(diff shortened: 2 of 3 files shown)


e111077:
What if you need DOM access or willUpdate property resolution?

justinfagnani:
@e111077 that's a great question. We don't have any tests that cover and guarantee that ability, so if we want to preserve it we should add one.

There may be another way to solve #3566 where we just don't call `host.requestUpdate()` synchronously. Let me check on that.

justinfagnani:
@e111077 I fixed this by putting `host.requestUpdate()` in a microtask and added a test that Tasks can see side effects of `update()` to ensure that behavior.

.changeset/fast-apes-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': minor
+---
+
+Tasks now run on hostUpdate() instead of hostUpdated()
```
e111077, line 5: Please update changelog, but looks good!

e111077 (changes requested):
Wow great job!

e111077 (approved):

PR #3661, opened 2023-02-10
[labs/analyzer] Factor setupAnalyzerForTest preamble out of tests

kevinpschaaf:


bicknellr (commented):
I was thinking about what you mentioned yesterday around trying to get the context type to be automatically determined by the `setupAnalyzerForTest` arguments (IIRC) and spent some time playing with the generics in this PR today. There's a way to get rid of the double casting thing inside `setupAnalyzerForTest` but I think I've mostly convinced myself that this can't be written such that the type argument of `suite<T>()` is the same for all test suites but the type of the context passed into each test is also narrowed by the arguments given to `setupAnalyzerForTest(...)`. My guess is that TS doesn't try to track control flow across functions for the purpose of narrowing, so it can't know if there's any relationship between the specific `T` instances that would be passed to the `test.before` and `test` callbacks.

Maybe this would work if `file` is the same for every test in a suite: wrap `suite<T>()`, make `file` a parameter of the wrapper, and then make `T` conditional on the type of `file`?

kevinpschaaf:
For now I just separated `setupAnalyzerForTest()` and `setupAnalyzerForTestWithModule()` into separate helpers with separate context types, rather than getting fancy parameterizing. Seems clearer and more straightforward PTAL.

bicknellr (approved):

PR #3667, opened 2023-02-14
[labs/ssr] Emit lit-node marker before node to support raw text nodes.

### LINKED ISSUE #3663: [labs/ssr] Binding to textarea results comment node errantly added inside textarea

SSR (@lit-labs/ssr)

### Description

Server rendering a `<textarea>` with a binding in property/attribute/element position results in a comment node being added inside the textarea content, which is rendered as plain text rather than being parsed as an HTML comment.

Reported by user `Levi` in Discord thread: https://discord.com/channels/1012791295170859069/1073542653402157076

### Reproduction

When server-rendering the following template using `@lit-labs/ssr`:
```ts
renderInput(id: string) {
  return html`<textarea name="${id}"></textarea>`;
}
```
A comment node shows up inside the textarea:
```html
<textarea id="name"><!--lit-node 0--></textarea>
```

### Workaround

Avoid binding to the textarea node.

### Is this a regression?

### Affected versions

@lit-labs/ssr 3.0.1

Any

kevinpschaaf:
Moves the `<!--lit-node n-->` marker for identifying nodes that hydration needs to stop at (to hydrate attribute/property/event/element parts and remove `defer-hydration` attributes) from after the start tag to before it.

This makes it compatible with bindings to the open tag of "raw text elements" (`style`, `script`, `textarea`, `title`) whose child content is parsed as raw text rather than nodes. This requires a quick lookahead in the opcode parsing to note whether the node in question has bindings, and if so write out the node marker before the tag is opened.

This had the side-effect of simplifying the condition around whether to completely finish writing out the open tag or not, since now that condition is only whether the node is a custom element or not.

Fixes #3663

### DIFF
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -1668,7 +1668,7 @@ export const tests: {[name: string]: SSRTest} = {
         },
       },
     ],
-    stableSelectors: ['input'],
+    stableSelectors: ['textarea'],
   },
 
   /******************************************************
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -1628,6 +1628,49 @@ export const tests: {[name: string]: SSRTest} = {
     stableSelectors: ['input'],
   },
 
+  'AttributePart on raw text element in shadow root': {
+    // Regression test for https://github.com/lit/lit/issues/3663.
+    //
+    // Confirms that attribute bindings to raw text elements now
+    // work as expected.
+    registerElements() {
+      class RawElementHost extends LitElement {
+        @property()
+        text = 'hello';
+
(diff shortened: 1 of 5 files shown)


packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -1628,6 +1628,49 @@ export const tests: {[name: string]: SSRTest} = {
...
+          const textarea = host.shadowRoot?.querySelector('textarea');
+          assert.equal(textarea?.value, 'hello');
+
+          host.text = 'goodbye';
+          await host.updateComplete;
+          assert.equal(textarea?.value, 'goodbye');
+        },
+      },
+    ],
+    stableSelectors: ['input'],
```
augustjk, line 1671: Maybe missed during copying from above?
```suggestion
    stableSelectors: ['textarea'],
```

augustjk (approved):
LGTM with the one minor suggestion.

PR #3670, opened 2023-02-14
[infra] Make IE11 Sauce Tests pass

augustjk:
We've been getting IE11 Sauce test failures like this: https://github.com/lit/lit/actions/runs/4176809294/jobs/7233584194

Unsure exactly when it started happening as we haven't been paying much attention due to flakiness but it's likely from some dependency update. My guess is that there's some bad interaction between the Symbol polyfill and possibly the babel transform done by Web Test Runner, such that using a `Symbol()` as a property key causes an error. It might also be due to the interaction of `core-js` Symbol polyfill added by WTR legacy plugin, and the `get-own-property-symbols` polyfill included with `@webcomponents/webcomponentsjs`. I tried changing some versions, configs, and patches but could not resolve.

With IE11 support being less of a concern, I've opted to just skip over the problematic tests in IE11. The shim for `Symbol.asyncIterator` in some tests were also causing the same errors so those have been removed. They are not needed for our modern browser tests and gets polyfilled separately for legacy browser tests. We can revisit if it becomes an issue in userland.

Besides that, this PR also adds wireit configs for the `BROWSERS` env so that those test scripts don't share the same wireit cache. Before, sauce tests could "pass" by reading the cached result from a regular run.


aomarks (approved):
Awesome, seems like a good compromise. Nice catch on the wireit env config too.

Issue #2426, opened 2022-01-20
[localize] Extracting html within html in messages


augustjk:
# Description

`lit-localize extract` does not properly handle `html` calls within `html` within `msg`.

# Steps to Reproduce

Include the following code
```ts
msg(html`Hello <b>${html`<i>World</i>`}</b>!`, {id: "foo"});
```

And run `lit-localize extract`

## Expected Results

```xml
<trans-unit id="foo">
  <source>Hello <x id="0" equiv-text="<b>${html`<i>World</i>`}</b>"/>!</source>
</trans-unit>
```

## Actual Results

```xml
<trans-unit id="foo">
  <source>Hello <x id="0" equiv-text="&lt;b>"/>html `<x id="1" equiv-text="&lt;i>"/>World<x id="2" equiv-text="&lt;/i>${`}&lt;/b>"/>!</source>
</trans-unit>
```

Above results are for xliff format but similar results are seen for xlb format as well.

Workaround for now is probably not to use `html` within `html` to create a placeholder, something like
```ts
html`${msg('Hello')} <b><i>World</i></b>!`;
```
though that doesn't allow a translator to switch the order if that's necessary. Perhaps #1890 would allow this?

PR #3674, opened 2023-02-16
[localize] Fix handling of nested HTML

aomarks:
Fixes handling of nested html-tagged template expressions in lit localize analysis.

For example, the following example was analyzed incorrectly:

```ts
msg(html`Hello <b>${html`<i>World</i>`}</b>!`);
```

We combine runs of HTML markup and expressions so that we have fewer placeholders during translation, so we pass the template through an HTML parser, nested expressions and all. The problem was that if the nested expressions contained HTML, those would get parsed as part of the outer message, which would corrupt the analyzed template. The fix was to replace nested expressions with placeholders (the index) during HTML parsing (because we don't need the HTML parser to see the contents of expressions at all) and then substitute the original expression back after HTML parsing.

Fixes https://github.com/lit/lit/issues/2426

### DIFF
--- a/.changeset/mighty-hornets-explain.md
+++ b/.changeset/mighty-hornets-explain.md
@@ -1,0 +1,5 @@
+---
+'@lit/localize-tools': patch
+---
+
+Fix incorrect extraction of html embedded within html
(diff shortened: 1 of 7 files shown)


augustjk (approved):
Nice! Much simpler this way.

PR #3677, opened 2023-02-16
[labs/ssr-dom-shim] Add basic support for element internals

### LINKED ISSUE #3676: [ssr-dom-shim]: Add lightweight support for ElementInternals
### Should this be an RFC?

### Description

Currently the `@lit-labs/ssr-dom-shim` package throws when it encounters and element that includes a call to `HTMLElement.prototype.attachInternals`. This feature is shipped in Chrome, Firefox and Safari Technology Preview and should generally be called in an element's constructor which is one of the methods that will be called by Lit's SSR.

Because Lit uses a lightweight DOM shim, the polyfill cannot add a function to the `HTMLElement` prototype.

As the author of the `element-internals-polyfill` I attempted to patch `HTMLElement.prototype.attachShadow` in that file to no avail. Alternatively component authors could check for the presence of attachInternals in their components but that puts undue burden on Lit's consumers.

calebdwilliams:
* Add a rough implementation for ElementShim.prototype.attachInternals Fixes #3676

# To do
- [x] Add tests
- [x] Add changeset


calebdwilliams:
Let me know if you'd like me to revert the package-lock as well.

calebdwilliams:
@justinfagnani it looks like I checked in some build files, is that normal for Lit or should I remove those?

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
augustjk, line 3:
```suggestion
 * Copyright 2023 Google LLC
```

justinfagnani:
@calebdwilliams awesome!

> looks like I checked in some build files, is that normal for Lit or should I remove those?

Those should be added to `.gitignore`. We typically put implementation modules in a folder named `lib/` and ignore the whole folder.

justinfagnani:
And yeah, let's revert the package-lock change, since there's no package.json change.

We've been trying to run on very recent versions of npm to keep the package-lock stable, btw.

augustjk:
> Those should be added to `.gitignore`. We typically put implementation modules in a folder named `lib/` and ignore the whole folder.

Also with regards to this, we need to add to wireit configs. @calebdwilliams I'm happy to handle any of our monorepo consistency related changes myself and push it up if you'd like.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
calebdwilliams: > Also with regards to this, we need to add to wireit configs. @calebdwilliams I'm happy to handle any of our monorepo consistency related changes myself and push it up if you'd like.

Yeah, that would be great. I'll work on removing the unnecessary files and reverting the package-lock

calebdwilliams:
I updated the PR to remove changes to the package-lock and removed the compiled files.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+  get shadowRoot() {
+    return this._host.shadowRoot;
+  }
+  constructor(_host: InternalsHost) {
+    this._host = _host;
+
+    initAom(_host, this);
+  }
+  checkValidity() {
+    return true;
```
justinfagnani, line 142: Do we need to implement this one? I'm not sure, but seems to me like we might want to throw rather than return a wrong answer. If we want to provide a correct answer, that would take some work, yeah?

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -59,6 +59,10 @@ const ElementShim = class Element {
     }
     return shadowRoot;
   }
+  attachInternals(): ElementInternals {
+    const internals = new InternalsShim(this);
```
justinfagnani, line 63: Let's assign to a field like `this.__internals` to make available to other SSR code like ElementRenderers.

justinfagnani (changes requested):

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+  ariaValueText = '';
+  role = '';
+  _host: InternalsHost;
+  get shadowRoot() {
+    return this._host.shadowRoot;
+  }
+  constructor(_host: InternalsHost) {
+    this._host = _host;
+
+    initAom(_host, this);
```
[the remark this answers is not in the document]
calebdwilliams: It doesn't if you use internals. The purpose here is to hydrate that behavior. If we were following the semantics, we should check to see if the attributes are set, if they aren't, append it to the element for SEO/SSR purposes then remove the element upon hydration if it still matches. 

I think having the AOM stuff reflected from internals to the pre-hydrated code would be a huge win for capturing the semantics of an element and for the DX of authoring components using internals/Lit SSR.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
calebdwilliams: Yeah, it would take some work for sure. We could just throw an error on the methods or just noop them … 

I think the bigger question would be for things like `setFormValue` and `setValidity`, those should exist but only be side-effectful upon hydration.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+export interface InternalsHost {
+  hasAttribute(name: string): boolean;
+  shadowRoot: ShadowRoot | null;
+  setAttribute(key: string, value: unknown): void;
+}
+
+/**
+ * @TODO
+ * - This can be typed better with keyof ARIAMixin, but TypeScript's definition
```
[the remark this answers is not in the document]
calebdwilliams: Looks like all of the browsers currently support a different set of features. All of them do reflect the from `AriaMixin`, so would it just be better in the long run to keep this list to `Record<keyof AriaMixin, string>`?

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+  ariaSort: 'aria-sort',
+  ariaValueMax: 'aria-valuemax',
+  ariaValueMin: 'aria-valuemin',
+  ariaValueNow: 'aria-valuenow',
+  ariaValueText: 'aria-valuetext',
+  role: 'role',
+};
+
+/** Force the attributes onto the reference element based on internals properties */
+export const initAom = (ref: InternalsHost, internals: ElementInternals) => {
```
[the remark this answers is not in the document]
calebdwilliams: Good call. Updating this.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+  ariaSetSize = '';
+  ariaSort = '';
+  ariaValueMax = '';
+  ariaValueMin = '';
+  ariaValueNow = '';
+  ariaValueText = '';
+  role = '';
+  _host: InternalsHost;
+  get shadowRoot() {
+    return this._host.shadowRoot;
```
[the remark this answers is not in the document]
calebdwilliams: Sure thing.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
```diff
@@ -0,0 +1,155 @@
...
+  ariaSetSize: 'aria-setsize',
+  ariaSort: 'aria-sort',
+  ariaValueMax: 'aria-valuemax',
+  ariaValueMin: 'aria-valuemin',
+  ariaValueNow: 'aria-valuenow',
+  ariaValueText: 'aria-valuetext',
+  role: 'role',
+};
+
+/** Force the attributes onto the reference element based on internals properties */
```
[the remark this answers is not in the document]
calebdwilliams: Does the new description help this?

packages/labs/ssr-dom-shim/src/element-internals.ts
```diff
@@ -0,0 +1,164 @@
...
+};
+
+/**
+ * Reflect internals AOM attributes back to the DOM prior to hydration
+ * to ensure search bots can accurately parse element semantics prior
+ * to hydration. This is called whenever an instance of ElementInternals
+ * is created on an element to wire up the getters/setters
+ * for the AriaMixin properties
+ *
+ * TODO - Determine the proper way to hydrate any attributes set by the shim
```
calebdwilliams, line 64: @justinfagnani Following up on your question from the other day. I'm happy to reference an issue, but I'm not sure where to put that. The version of TypeScript I'm running in VSCode matches the spec, but neither Chromium or Webkit (or Firefox with the ARIA reflection tag) have the same tags. The ones listed here are in line with Chromium's implementation. 

To complicate that a bit more, the version of TypeScript building Lit's repo is different from the version running in VSCode (somehow) so that it recognizes different properties on `ARIAMixin` which appears to be somewhat volatile in TypeScript. 

Any guidance you could provide on how to add more clarity here would be welcomed.

packages/labs/ssr-dom-shim/src/InternalsShim.ts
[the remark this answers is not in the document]
justinfagnani: That sounds reasonable

packages/labs/ssr-dom-shim/src/element-internals.ts
```diff
@@ -0,0 +1,127 @@
...
+  ariaValueMin: 'aria-valuemin',
+  ariaValueNow: 'aria-valuenow',
+  ariaValueText: 'aria-valuetext',
+  role: 'role',
+};
+
+// Shim the global element internals object
+// Methods should be fine as noops and properties can generally
+// be while on the server.
+export const InternalsShim = class ElementInternals implements ARIAMixin {
```
augustjk, line 60: I'm getting
```
Class 'ElementInternals' incorrectly implements interface 'ARIAMixin'.
  Type 'ElementInternals' is missing the following properties from type 'ARIAMixin': ariaColIndexText, ariaRowIndexText ts(2420)
```
Is this expected?

augustjk: Also, maybe related, is it possible to dedupe the class fields listing below with the enum above?

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -54,11 +60,17 @@ const ElementShim = class Element {
...
+    this.__shadowRootMode = init.mode;
     if (init && init.mode === 'open') {
       this.__shadowRoot = shadowRoot;
     }
     return shadowRoot;
   }
+  attachInternals(): ElementInternals {
+    const internals = new InternalsShim(this as unknown as HTMLElement);
+    this.__internals = internals;
+    return internals as ElementInternals;
```
augustjk, line 72: Also getting TS errors on this assignment and casting on local IDE due to `ariaColIndexText, ariaRowIndexText`.

I wonder if the `InternalsShim` should get the type treatment like the other shims in this file:
```ts
const ElementShimWithRealType = ElementShim as object as typeof Element;
export {ElementShimWithRealType as Element};
```

calebdwilliams: Yeah, they type `ARIAMixin` is apparently really fragile. I'll back out of that and use `string` instead. I think most of the issues you're seeing are related to that.

augustjk: `tsc` wasn't breaking so it turns out the issue was my VS Code was using TS v4.9.5 whereas the workspace's v4.7.4.

It seems `ariaColIndexText` and `ariaRowIndexText` were added to TS's `lib.dom.d.ts` in a later version as they're part of the spec, but in practice not implemented by browsers. MDN has also pulled them from their page for that reason https://github.com/mdn/browser-compat-data/pull/18546.

Since this builds, maybe okay to leave as is, but we could also `Omit<ARIAMixin, 'ariaColIndexText' | 'ariaRowIndexText'>` mentioning lack of support for future proofing.

augustjk: Actually.. `Omit<ARIAMixin, 'ariaColIndexText' | 'ariaRowIndexText'>` will help for `implements` but not for casting to built in `ElementInternals` type. So.. maybe okay for now, we can worry about it when we update TS.

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -54,11 +66,17 @@ const ElementShim = class Element {
...
   attachShadow(init: ShadowRootInit): ShadowRoot {
     const shadowRoot = {host: this} as object as ShadowRoot;
+    this.__shadowRootMode = init.mode;
     if (init && init.mode === 'open') {
       this.__shadowRoot = shadowRoot;
     }
     return shadowRoot;
   }
+  attachInternals(): ElementInternals {
+    const internals = new ElementInternalsShim(this as unknown as HTMLElement);
```
justinfagnani, line 76: This should throw if it's called twice.

packages/labs/ssr-dom-shim/src/lib/element-internals.ts
```diff
@@ -0,0 +1,141 @@
...
+  interface ARIAMixin {
+    ariaBraileLabel: string | null;
+    ariaBraileRoleDescription: string | null;
+    ariaDescription: string | null;
+    ariaInvalid: string | null;
+    role: string | null;
+  }
+}
+
+export const ariaMixinEnum: Record<keyof ARIAMixin, string> = {
```
justinfagnani, line 19: To make sure this covers all properties, I'd make a mapped type:

```ts
type ARIAAttributeMap = {
  [K in keyof ARIAMixin]: string;
}

export const ariaMixinAttributes: ARIAAttributeMap = {
```

justinfagnani, line 19: Add a comment on what this is describing? and/or a more descriptive name.

packages/labs/ssr-dom-shim/src/lib/element-internals.ts
```diff
@@ -0,0 +1,141 @@
...
+    // to ensure that the shadow root is always available
+    // to the internals instance even if the mode is 'closed'
+    return (this.__host as HTMLElement & {__shadowRoot: ShadowRoot})
+      .__shadowRoot;
+  }
+  constructor(_host: HTMLElement) {
+    this.__host = _host;
+  }
+  checkValidity() {
+    return true;
```
justinfagnani, line 122: I still don't know if this should return true. I think we need to walk through the cases where for validity is checked on the server to see if this makes sense. I would log a warning for now, and take up a new issue.

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -29,6 +33,31 @@ export class LitElementRenderer extends ElementRenderer {
...
     super(tagName);
     this.element = new (customElements.get(this.tagName)!)() as LitElement;
+
+    // Reflect internals AOM attributes back to the DOM prior to hydration to
+    // ensure search bots can accurately parse element semantics prior to
+    // hydration. This is called whenever an instance of ElementInternals is
+    // created on an element to wire up the getters/setters for the ARIAMixin
+    // properties.
+    const internals = (
+      this.element as object as {__internals: ElementInternals}
```
justinfagnani, line 43: So this only work when we're using our shim. We need to think about what happens if someone loads jsdom. We may need adapters, like a `getInternals()` callback.

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -29,6 +33,31 @@ export class LitElementRenderer extends ElementRenderer {
...
+    ).__internals;
+    if (internals) {
+      for (const [key, value] of Object.entries(internals)) {
```
justinfagnani, line 46: I would iterate on the enum. The internal object could conceivably have an accessor which wouldn't be returned by Object.entries().

packages/labs/ssr-dom-shim/src/index.ts
augustjk: Done

packages/labs/ssr-dom-shim/src/lib/element-internals.ts
augustjk: Done

augustjk: Done

packages/labs/ssr-dom-shim/src/lib/element-internals.ts
augustjk: Done

packages/labs/ssr/src/lib/lit-element-renderer.ts
augustjk: Done

packages/labs/ssr/src/lib/lit-element-renderer.ts
justinfagnani: We should have an issue somewhere for using another DOM shim, and this can go on the list of things we'll need to consider there. What I mean by `getInternals()` is that since there's no standard DOM API for getting at that from outside a component, we'll need a callback per DOM emulation library to get at it. Future work, just noting it.

justinfagnani (approved):

PR #3678, opened 2023-02-16
[labs/analyzer] Collect and log non-fatal diagnostics.

bicknellr:
This adds a few methods to the analyzer for collecting and logging diagnostics. The analyzer no longer crashes when it sees a `LitElement` extension that has a non-identifier property name and instead adds a warning diagnostic. Also, the CEM generator now logs any diagnostics that were added after it generates the manifest.


packages/labs/analyzer/test-files/ts/functions/src/functions.ts
```diff
@@ -36,3 +36,18 @@ export function function2(a: string, b = false, ...c: number[]) {
...
   return 'default';
 }
+
+export function overloaded(x: string): string;
+export function overloaded(x: number): number;
+/**
+ * This function has an overloaded signature in TS.
+ * @param x Some value, either a string or a number.
+ * @returns Returns either a string or a number.
+ */
```
bicknellr, line 46: There seems to be a bug where the descriptions (etc.) aren't picked up if this comment isn't attached to the main declaration.

packages/labs/analyzer/test-files/ts/properties/src/element-a.ts
```diff
@@ -80,4 +82,22 @@ export class ElementA extends LitElement {
...
+  @property()
+  [unsupportedPropertyName]: string;
+
+  overloaded(x: string): string;
+  overloaded(x: number): number;
+  /**
+   * This function has an overloaded signature in TS.
+   * @param x Some value, either a string or a number.
+   * @returns Returns either a string or a number.
+   */
```
bicknellr, line 95: This also shouldn't be required to be attached to the main declaration.

packages/labs/analyzer/src/test/lit-element/properties_test.ts
```diff
@@ -211,5 +212,53 @@ for (const lang of languages) {
...
+    assert.equal(
+      diagnostics[0].code,
+      DiagnosticCode.UNSUPPORTED_PROPERTY_NAME_TYPE
+    );
+
+    // Fields named with symbols are visible in the `fields` iterator.
+    const field = Array.from(element.fields).find(
+      (x) => x.name === '[unsupportedPropertyName]'
+    );
+    assert.ok(field);
```
bicknellr, line 227: I'm not sure if this is a bug or not. Symbol-named properties make it into `fields` here:

https://github.com/lit/lit/blob/be18292938062a3b5233016fdac1a72ba6f1eacf/packages/labs/analyzer/src/lib/javascript/classes.ts#L89-L100


...which doesn't seem to care what type of name a property is using, but I'm not sure if that was intentional or not. `node.name` is the AST node for however the name of a field is described, so its `getText()` ends up returning the name with brackets around it, which looks reasonable but is really just JS text. (I think `getText()` would return `["abc" + "def"]` for a field like `class { ["abc" + "def"] = 123; }`.)

justinfagnani: hmm...is that a bug? What do we want the name to be? I could see that for the text representation, that something like `[mySymbol]` is about as good as we can get. But to fully support Symbol-named fields we probably want to represent them as a reference in the manifest, so we'd need to keep the name AST node.

I think that symbol-named fields are rare enough that we could punt on a few things, keep issues, then come back with a push to support symbols better, including adding name reference support to ClassField and ClassMethod: https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts#L557

packages/labs/analyzer/test-files/ts/properties/src/element-a.ts
bicknellr: This was removed.

packages/labs/analyzer/test-files/ts/functions/src/functions.ts
bicknellr: This was removed.

packages/labs/analyzer/src/test/lit-element/properties_test.ts
bicknellr: I've updated this to also log a diagnostic rather than analyzing the property.

packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -73,6 +75,18 @@ export class Analyzer implements AnalyzerInterface {
...
     });
   }
+
+  addDiagnostic(options: DiagnosticOptions) {
+    this.diagnostics.push(createDiagnostic(options));
+  }
+
+  *getDiagnostics(): IterableIterator<ts.Diagnostic> {
+    yield* this.diagnostics;
+  }
```
bicknellr, line 85: This feels like something that someone using the analyzer's API would want but it's only used in tests as of now. Should I update this to avoid putting on the public API for now?

rictic:
Superceded by https://github.com/lit/lit/pull/3866

PR #3686, opened 2023-02-21
fix for globalThis on iOS 11

ALeschinsky:
`globalThis` is not defined on older ipads, this PR fixes it

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -54,6 +54,8 @@ export * from 'lit-html';
 import {LitUnstable} from 'lit-html';
 import {ReactiveUnstable} from '@lit/reactive-element';
 
+typeof window === 'undefined' || (window['globalThis'] = window); // fix for iOS 11
+
 /**
  * Contains types that are part of the unstable debug API.
  *


justinfagnani:
I think it's best if we require this to be added as a polyfill, since we have other packages that use `globalThis`.

ALeschinsky:
> I think it's best if we require this to be added as a polyfill

Thank you for the review, I moved the line to `packages/lit-element/src/polyfill-support.ts:84`

packages/lit-element/src/polyfill-support.ts
```diff
@@ -81,6 +81,8 @@ const polyfillSupport = ({LitElement}: {LitElement: PatchableLitElement}) => {
   };
 };
 
+typeof window === 'undefined' || (window['globalThis'] = window); // fix for iOS 11
```
rictic, line 84: This seems reasonable, but let's put it in reactive-element's polyfill-support instead, as that also uses globalThis, and should run first

rictic (changes requested):

packages/lit-element/src/polyfill-support.ts
rictic: Also, let's update the comment to mention that it's for older browsers in general

PR #3688, opened 2023-02-21
Check formatting during lint workflow.

bicknellr:
This PR checks formatting during the lint workflow. I've also changed how the `lint` and `format` scripts are organized so that `lint` always refers to linting with eslint and `format` always refers to formatting with prettier. The existing `format` script actually ran both the linter and formatter so I've added another script called `fix` for that use case.


package.json
```diff
@@ -18,11 +18,14 @@
     "build": "wireit",
     "build:ts": "wireit",
     "build:lit-starter-ts": "wireit",
-    "format": "npm run format:eslint && npm run format:prettier",
-    "format:eslint": "eslint --ignore-pattern=examples/ --fix \"**/*.{js,ts}\"",
-    "format:prettier": "prettier \"**/*.{cjs,html,js,json,md,ts}\" --write",
+    "fix": "npm run lint:fix && npm run format:fix",
+    "format": "npm run format:fix",
```
justinfagnani, line 22: Nice. I wonder if a good next step would be to run these with wireit so we take advantage of caching. Would that even be worth it considering how long these take? @aomarks ?

justinfagnani (approved):

package.json
bicknellr: I think I generally have these working in wireit, but one thing I'm a bit confused about is the `output` files list. My initial guess was that the `:fix` versions of these scripts should use the same glob for `files` and `output` since those scripts both read and potentially write to all of them, but this causes all of the matching files to be deleted. Is `output` only for files that are specifically created by the script, not just any that it might modify?

augustjk: I think it would be overkill to do wireit for these. Wireit probably isn't great for scripts like these where only some files are touched during run but we can't know which ones until running. So instead `files` and `output` here are potentially the entirety of the monorepo. It'll probably end up taking more space in the github cache unnecessarily.

bicknellr: Yeah, those globs match almost everything. I'll roll the wireit part back for now and we can add it later again if we find it's useful.

.changeset/slimy-mails-clean.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@lit-labs/virtualizer": patch
+'@lit-labs/virtualizer': patch
```
augustjk, line 2: i wonder if edits to these markdown files might change the PR and commit that get linked with this update in the changelog. is it possible to exclude `.changeset/` from formatting?

bicknellr: I hope that it would associate the PR that introduced the changeset but I'm not sure. I forgot to add one specifically for this PR so at a minimum these edits are tripping up the "did you add a changeset?" check.

augustjk: ah right. also new files would probably get formatted as they go in so likely not a big deal.

augustjk (approved):

.changeset/slimy-mails-clean.md
bicknellr: Yeah, the workflow update should enforce that they're formatted properly from now on.

bicknellr:
There are a few failing tests in IE: https://github.com/lit/lit/actions/runs/4246514504/jobs/7383424568>
I'm going to merge this since they're consistent with the past couple of commits to `main`, so I don't think they're related:
https://github.com/lit/lit/actions/runs/4246317083/jobs/7382969109
https://github.com/lit/lit/actions/runs/4236749763/jobs/7361913647

PR #3691, opened 2023-02-22
Work around TS compiler bug in context

rictic:


packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -28,6 +28,9 @@ export class ContextProviderEvent<
    */
   public constructor(public readonly context: C) {
     super('context-provider', {bubbles: true, composed: true});
+    // Work around TypeScript compiler bug
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    Object.setPrototypeOf(this, new.target.prototype);
```
justinfagnani, line 33: at least using `new.target.prototype` means subclasses of this class don't have to do the same trick (if someone wanted to subclass for some odd reason)

justinfagnani (approved):

PR #3693, opened 2023-02-24
[reactive-element] Only use symbol keys for properties added as private storage.

### LINKED ISSUE #3690: [reactive-element] Leading double underscore in a property name is ignored

### Description

Properties whose names start with `__` are treated as the same property as the property without these underscores.

Use case:
I have an attribute foo that can be set by outside factors:
```
@property()
foo: string;
```

And an internal state variable __foo that is related, but should not override the externally set value
```
@property({state: true})
__foo: string;
```

when I update `__foo`, `foo` is also immediately updated, which should not be the case.

### Reproduction

https://lit.dev/playground/#project=W3sibmFtZSI6InJlcHJvLnRzIiwiY29udGVudCI6ImltcG9ydCB7aHRtbCwgY3NzLCBMaXRFbGVtZW50fSBmcm9tICdsaXQnO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50LCBwcm9wZXJ0eX0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuXG5AY3VzdG9tRWxlbWVudCgneC1yZXBybycpXG5leHBvcnQgY2xhc3MgWFJlcHJvIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIEBwcm9wZXJ0eSgpXG4gIGNvdW50ZXI6IG51bWJlcjtcblxuICBAcHJvcGVydHkoe3N0YXRlOiB0cnVlfSlcbiAgX19jb3VudGVyID0gMDtcblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgXG4gICAgICA8cD5jb3VudGVyIGFuZCBfX2NvdW50ZXIgYXJlIGRpZmZlcmVudCB2YXJpYWJsZXMgYW5kIHNob3VsZCBhY3QgaW5kZXBlbmRlbnRseTo8L3A-XG4gICAgICA8YnV0dG9uIEBjbGljaz0ke3RoaXMuX29uQ2xpY2tCdXR0b259PkNsaWNrIG1lPC9idXR0b24-XG4gICAgICA8cD5jb3VudGVyID0gJHt0aGlzLmNvdW50ZXJ9PC9wPlxuICAgICAgPHA-X19jb3VudGVyID0gJHt0aGlzLl9fY291bnRlcn08L3A-XG4gICAgYDtcbiAgfVxuICBcbiAgX29uQ2xpY2tCdXR0b24oKSB7XG4gICAgdGhpcy5fX2NvdW50ZXIrKztcbiAgfVxufVxuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IURPQ1RZUEUgaHRtbD5cbjxoZWFkPlxuICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL3JlcHJvLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHgtcmVwcm8gY291bnRlcj01PjwveC1yZXBybz5cbjwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4yLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjEuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl4zLjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMi4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfV0

### Workaround

I have changed my internal variable name to use a single underscore (eg `_foo` instead of `__foo`)

### Is this a regression?

### Affected versions

Using lit 2.6.1

Browser: Version 100.0 (64-bit)
OS: Linux mint 20.3

bicknellr:
Fixes #3690.

### DIFF
--- a/packages/reactive-element/src/decorators/query.ts
+++ b/packages/reactive-element/src/decorators/query.ts
@@ -14,6 +14,8 @@
 import {ReactiveElement} from '../reactive-element.js';
 import {decorateProperty} from './base.js';
 
+const DEV_MODE = true;
+
 /**
  * A property decorator that converts a class property into a getter that
  * executes a querySelector on the element's renderRoot.
@@ -41,7 +43,7 @@ import {decorateProperty} from './base.js';
  */
 export function query(selector: string, cache?: boolean) {
   return decorateProperty({
-    descriptor: (_name: PropertyKey) => {
+    descriptor: (name: PropertyKey) => {
       const descriptor = {
         get(this: ReactiveElement) {
           return this.renderRoot?.querySelector(selector) ?? null;
@@ -50,7 +52,9 @@ export function query(selector: string, cache?: boolean) {
         configurable: true,
       };
       if (cache) {
-        const key = Symbol();
+        const key = DEV_MODE
+          ? Symbol(`${String(name)} (@query() cache)`)
+          : Symbol();
         descriptor.get = function (this: ReactiveElement) {
           if (
             (this as unknown as {[key: symbol]: Element | null})[key] ===
--- a/packages/reactive-element/src/decorators/query.ts
+++ b/packages/reactive-element/src/decorators/query.ts
@@ -41,7 +41,7 @@ import {decorateProperty} from './base.js';
  */
 export function query(selector: string, cache?: boolean) {
   return decorateProperty({
-    descriptor: (name: PropertyKey) => {
+    descriptor: (_name: PropertyKey) => {
       const descriptor = {
         get(this: ReactiveElement) {
           return this.renderRoot?.querySelector(selector) ?? null;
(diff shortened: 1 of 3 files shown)


justinfagnani:
I think if we do this it'll be a breaking change, since we don't call `Symbol()` in the core libraries unless you're making a symbol-keyed reactive property. So this would force a polyfill where we don't now.

I'm actually ok with this as part of dropping IE11 support though, but we need to do that with a new major release in mind, and pull all being changes into Google one-by-one.

justinfagnani:
@bicknellr let's do this as part of 3.0. Can you rebase onto the 3.0 branch? I could take the PR over too.

bicknellr:
SGTM, rebased.

justinfagnani (approved):

justinfagnani:
Thanks @bicknellr !

PR #3695, opened 2023-02-26
chore: add devcontainer configuration

Garbee:
Provide a starting point for devcontainer use to develop lit. This can help new contributors quickly get a dev environment going in the cloud without needing to run anything locally. However, if desired a developer can run devcontainers locally in order to isolate different work environments.

This container does the following:

- Uses the existing Node 24 + TypeScript devcontainer from Microsoft as the base
- Sets up the GH CLI so it is available in case anyone is used to in their workflow
- Installs playwright dependencies on the operating system
- After a container is brought online for a workspace, like a new Codespace, `npm ci` is executed so people can get into working quickly. Then `npx playwright install` is ran so the browsers are installed with the current dependencies of the repository in scope.

It does also suggest 4 cores as a requirement. This is because on Codespaces when running build and test the CPU was maxing on the 2 core machines and triggering warnings repeatedly. At 4 cores they can show up from time-to-time, but overall the runs seem to go smoothly.


rictic (approved):
Very nice, seems very helpful!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  "features": {
+    // Github CLI for anyone accustomed to it in their workflow.
+    "ghcr.io/devcontainers/features/github-cli:1": {
+      "installDirectlyFromGitHubRelease": true,
+      "version": "latest"
+    },
+    // Install browser depenedencies for testing
+    "ghcr.io/devcontainers-contrib/features/bash-command:1": {
+      "command": "npx playwright install --with-deps"
```
justinfagnani, line 22: Do the browsers not get installed with the playwright npm package?

justinfagnani:
@rictic someone needs to start the test workflows. I don't see the button on my phone though.

.devcontainer/devcontainer.json
Garbee: So...

1. Yes, absolutely correct the browsers do get installed.
2. I need to tweak the command to install the dependencies only. I messed this up in a few earlier things I was doing until I realized the command problem.

This will become `npx playwright install-deps` in my next commit here.

Garbee:
need to do some tests again to see where 4 core falls. And update the description after that before this is ready for review again.

PR #3702, opened 2023-03-01
[labs/analyzer] Analyze overloaded functions and methods.

bicknellr:
This PR adds an `overloads` field to `FunctionLikeDeclaration` and fills it with declarations for each overload signature.


packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -76,14 +79,15 @@ export const getClassMembers = (
   const methodMap = new Map<string, ClassMethod>();
   const staticMethodMap = new Map<string, ClassMethod>();
   declaration.members.forEach((node) => {
-    if (ts.isMethodDeclaration(node)) {
+    if (ts.isMethodDeclaration(node) && node.body) {
```
bicknellr, line 82: This needs a comment like the change in `modules.ts`.

packages/labs/analyzer/src/lib/javascript/functions.ts
```diff
@@ -70,20 +91,24 @@ const getFunctionDeclaration = (
...
 export const getFunctionLikeInfo = (
   node: ts.FunctionLikeDeclaration,
+  docNode: ts.FunctionLikeDeclaration,
   analyzer: AnalyzerInterface
 ) => {
   return {
-    parameters: node.parameters.map((p) => getParameter(p, analyzer)),
-    return: getReturn(node, analyzer),
+    parameters: node.parameters.map((p, i) =>
+      getParameter(p, docNode.parameters[i], analyzer)
```
bicknellr, line 99: This is why I need to check if `.parameters.length` can vary between override declarations.

packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -117,7 +117,9 @@ export const getModule = (
   for (const statement of sourceFile.statements) {
     if (ts.isClassDeclaration(statement)) {
       addDeclaration(getClassDeclarationInfo(statement, analyzer));
-    } else if (ts.isFunctionDeclaration(statement)) {
+      // Ignore non-implementation signatures of overloaded functions by
+      // checking for `statement.body`.
```
justinfagnani, line 121: Can you maybe expand on this a bit? Like why do we want to ignore overloads? What do we do with different docs on different overloads? I don't the the CEM format explicitly handles overloads, so if we generated multiple declarations with the same name but different parameters, that would basically work by default - with the exception that tools might expect names to be unique.

I could see us documenting in the CEM that names aren't unique and duplicates indicate overloads (though we might want a way to mark and/or hide the implementation declaration because its signature might not be pretty).

packages/labs/analyzer/test-files/js/functions/functions.js
```diff
@@ -36,3 +36,46 @@ export function function2(a, b = false, ...c) {
 export default function () {
   return 'default';
 }
+
+/**
+ * This function has an overloaded signature in TS.
```
justinfagnani, line 41: What does that mean exactly? Is this file associated with `ts/functions/src/functions.ts` so that those overloads apply here? I would have thought that these were separate modules.

packages/labs/analyzer/src/lib/javascript/functions.ts
```diff
@@ -23,6 +23,26 @@ import {getTypeForNode, getTypeForType} from '../types.js';
...
+  declaration: ts.FunctionLikeDeclaration,
+  analyzer: AnalyzerInterface
+) => {
+  if (ts.getJSDocTags(declaration).length !== 0) {
+    return declaration;
+  }
+
+  // Overloaded functions have mulitple declaration nodes. If there are no
+  // JSDoc tags on the provided declaration, use the first one that does have
+  // JSDoc tags for the purpose of extracting a description.
```
justinfagnani, line 36: TypeScript's behavior seems to be to choose the matched overload if there is one, or the first overload with jsdocs, but, surprisingly, _never_ show the impl docs. The impl basically doesn't exist.

https://www.typescriptlang.org/play?ts=5.1.0-dev.20230227#code/PQKhCgAIUh5A3ApgJwDYHsCGATSBGKEYcAMwFcA7AYwBcBLdCyTACgA8AuSAZxuTooBzAJRd46OtgDc4UBGhwkaLLgBMhYuWr1GzdlwpkAtgCMUoyOMky5hSAEkjAB1QbSlWgyatOPPgMFIAB9IQ1NzSABvKEgqRm50VEQAOgxBdmEZAF9wWWBIAAl0JQBaYpQeAAt0AHduSAAiBBQMHEhVBvBWPEzc2wVm5TaCaE0PHSYTfT9+IQsraXdtL0gp3zCzZHmJRf6YRxc3LU9dNa5eWcCQjYjoyFj4xJS0jOy+-KLS8uQq2vqmpStXB4TpTHoyJYnJhUaYXALbayQiaxaY3LZiHY2MB2A6uUZIlYw3xwoTBULGTbCKIxOIUBJJVLodJsXo5PKFb5lJS-OqhdCQABW3Gw6Co3HAMPB4CAA

packages/labs/analyzer/src/lib/javascript/modules.ts
bicknellr: IIRC, I went down this path because there are few places in the analyzer that currently require unique names (and I just assumed this was a requirement). [Kevin's comment on my other PR](https://github.com/lit/lit/pull/3678#discussion_r1120880191) about `getField()` is related. Another example is here, where declarations in modules are collected:

https://github.com/lit/lit/blob/be18292938062a3b5233016fdac1a72ba6f1eacf/packages/labs/analyzer/src/lib/javascript/modules.ts#L105-L107

I think your suggestion of explicitly documenting that names of potentially overloadable items in collections are not necessarily unique sounds like the best way to go. That way I could remove the weird documentation merging stuff here and we would avoid losing information in the output. Conveniently, the CEM schema already seems prepared for this since most (all?) collections of declarations are arrays of items with `name` properties and not an object / map that requires a unique key per item. This would also resolve the issue you pointed out in your other comment about this PR not matching TS's behavior around docs for overloads.

packages/labs/analyzer/test-files/js/functions/functions.js
bicknellr: Yes, the same tests are run twice (once for each analysis result from files in the `ts` and `js` directories) so the doc comments that those tests check for need to match. For overloaded functions in particular, the comments for the single declaration in JS need to match those from the particular overload that they're extracted from in TS. These could be rewritten to make what's happening here clearer though.

However, it also sounds like I need to update this PR to handle overloads by producing multiple entries rather than trying to merge into a single one and I'm not sure how this is going to affect what's in the JS version of these fixtures yet. I'll try to make these more self-explanatory once I know what's happening with them.

packages/labs/analyzer/src/lib/javascript/classes.ts
bicknellr: This check and all the related `docNode` stuff should be dropped once unique names are no longer required.

packages/labs/analyzer/src/lib/javascript/functions.ts
bicknellr: This was removed.

packages/labs/analyzer/src/lib/javascript/classes.ts
bicknellr: I kept this because we still want this map to be keyed by member names, but the value is now specifically the implementation signature's `FunctionDeclaration`. Overloads are now discovered by `getFunctionLikeInfo` instead, which reaches them from the implementation signature's node.

packages/labs/analyzer/src/lib/javascript/modules.ts
justinfagnani: which would move the responsibility for matching TS's behavior or not to consumers of the manifest. That sounds good to me.

justinfagnani (approved):

.changeset/early-berries-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/analyzer': minor
+---
+
+Documentation for overloaded functions and methods is now extracted from another declaration if the implementation declaration has no JSDoc tags.
```
bicknellr, line 5: This needs to be updated.

packages/labs/analyzer/src/lib/javascript/classes.ts
bicknellr: This has a comment now.

PR #3703, opened 2023-03-02
Ignore `/examples/` when lint-staged runs eslint.

bicknellr:
eslint doesn't apply `--ignore-pattern` to any full paths passed to it but lint-staged *only* passes full paths. This replaces the eslint `--ignore-pattern` option with a more specific lint-staged glob.

### DIFF
--- a/package.json
+++ b/package.json
@@ -226,6 +226,6 @@
   },
   "lint-staged": {
     "**/*.{cjs,html,js,json,md,ts}": "prettier --write",
-    "**/*.{js,ts}": "eslint --ignore-pattern=examples/ --fix"
+    "{*.{js,ts},!(examples)/**/*.{js,ts}}": "eslint --fix"
   }
 }


augustjk (approved):
I'm okay with this. We won't get lint-staged checking updates to examples which seems fine since I don't envision that getting much code changes once added.

bicknellr:
We could also install the examples directory or remove the `"extends"` line there. Would either of those be preferable to this?

kevinpschaaf (approved):
Seems ok.

justinfagnani:
> We could also install the examples directory or remove the "extends" line there. Would either of those be preferable to this?

You mean remove the `"extends": "next/core-web-vitals"`? I don't think we want that if that's a common/recommended eslint plugin for Next. We want to make sure we don't have an obvious violation just from using Lit.

justinfagnani (approved):

PR #3707, opened 2023-03-07
[labs/analyzer] Parse ctor-defined fields as ClassField instead of ReactiveProperty

kevinpschaaf:
The `ReactiveProperty` model holds metadata specific to Lit's reactive properties (either via `@property()` decorator or `static properties` block). From early on, it also held `type` information. However, now that we have a `ClassField` model with `type` information, that is redundant, so this removes `type` from `ReactiveProperty`.

In addition, when we added JavaScript support to the analyzer (prior to adding `ClassField`), in order to fill in the `type` information on `ReactiveProperty`, we added code to find the associated initializer in the `constructor` and analyze the `type` info from that. This PR also moves that logic to the `ClassField` analysis, so that we also create class fields based on constructor-initialized fields.

### DIFF
--- a/packages/labs/analyzer/src/lib/model.ts
+++ b/packages/labs/analyzer/src/lib/model.ts
@@ -615,7 +615,9 @@ export interface Parameter extends PropertyLike {
   rest?: boolean | undefined;
 }
 
-export interface ReactiveProperty extends PropertyLike {
+export interface ReactiveProperty {
+  name: string;
+
   reflect: boolean;
 
   // TODO(justinfagnani): should we convert into attribute name?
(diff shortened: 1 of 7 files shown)


packages/labs/analyzer/src/lib/model.ts
```diff
@@ -615,7 +615,9 @@ export interface Parameter extends PropertyLike {
   rest?: boolean | undefined;
 }
 
-export interface ReactiveProperty extends PropertyLike {
+export interface ReactiveProperty {
+  name: string;
+
```
bicknellr, line 620: Would it be worth adding a reference to the property here?

bicknellr (approved):

justinfagnani (approved):

PR #3708, opened 2023-03-07
[labs/virtualizer] handle uninitialized layout in _childrenSizeChanged

### LINKED ISSUE #3726: [labs/virtualizer]: uninitialized layout Error in _childrenSizeChanged

### Description

 `Error` in `_childrenSizeChanged` when children is resized  before layout is initialized.

### Reproduction

Test case in PR.

### Workaround

None.

### Is this a regression?

### Affected versions

1.0.1

Any

megheaiulian:
Handle uninitialized layout in `_childrenSizeChanged`.

 Because the ResizeObserver starts listening very early it is possible that the layout has not been imported and initialized when a children has updated.

Fixes #3726

### DIFF
--- a/packages/labs/virtualizer/src/test/scenarios/immediate-sizing.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/immediate-sizing.test.ts
@@ -47,7 +47,7 @@ describe("Don't behave badly when sizing element before initialization is comple
   ignoreBenignErrors(beforeEach, afterEach);
 
   it('should render the element without error', async () => {
-    const example: ImmediateSizing = await fixture(
+    await fixture(
       testingHtml`
         <immediate-sizing .items=${['item 1', 'item 2']}></immediate-sizing>
       `
--- a/packages/labs/virtualizer/src/test/scenarios/immediate-sizing.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/immediate-sizing.test.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ignoreBenignErrors} from '../helpers.js';
+import {html, css, LitElement} from 'lit';
+import {ref} from 'lit/directives/ref.js';
+import {customElement, property} from 'lit/decorators.js';
+import {virtualize} from '../../virtualize.js';
+import {html as testingHtml, fixture} from '@open-wc/testing';
+
+@customElement('immediate-sizing')
+class ImmediateSizing extends LitElement {
+  static styles = css`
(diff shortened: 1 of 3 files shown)


usergenic:
Seems totally plausible.  Thanks for the fix, @megheaiulian

justinfagnani:
@megheaiulian it would be great to have a test to prevent a regression here.

usergenic:
The scenario where this would create a regression would be that the logic is relying on a weird unhandled runtime error to abort execution of the main childrenSizeChanged callback.  With or without this fix, the inner condition doesn't execute, but the "new" behavior is that code continues to execute without runtime error and invokes this._scheduleLayoutComplete(); etc--

I'll investigate the observer instantiation in relation to a deferred/dynamic layout assignment and see if there's a reason why this would behavior would be unwanted.

usergenic:
i.e. it is an open question is whether this either creates a regression or fixes an edge case.

graynorton:
Thanks for the PR, @megheaiulian!

A simplified repro case would definitely be helpful; at minimum, could you file an issue to match up with this PR?

@usergenic @justinfagnani There are unreleased changes in virtualizer that seem quite likely to be relevant, as they definitely touch (and aim to simplify) virtualizer initialization. Is there anything holding up a release? In any case, once we have a repro for this, we should make sure that it still holds when run against the code in `main`.

megheaiulian:
@graynorton I will add a test for this.

megheaiulian:
Add a test case to reproduce the error.

megheaiulian:
@usergenic Could you please have a another look at this ? It is a very annoying little bug.

packages/labs/virtualizer/src/test/scenarios/immediate-sizing.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      <div ${ref((el?: Element) => el?.classList.toggle('sized'))}></div>
+    </div>`;
+  }
+}
+
+describe("Don't behave badly when sizing element before initialization is complete ", () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  it('should render the element without error', async () => {
+    const example: ImmediateSizing = await fixture(
```
justinfagnani, line 50: This is causing a lint warning. Just don't declare a variable.

justinfagnani (changes requested):
Looks good after the lint fix!

usergenic (approved):
Thank you for your patience on this and for adding the test case etc.  LGTM

usergenic:
@justinfagnani this PR needs your approval because of prior requested change.

justinfagnani (approved):

usergenic:
@megheaiulian alright this is finally merged.  Thanks again!

Issue #3709, opened 2023-03-07
[reactive-element] PropertyValues should include undefined in return type of get()


justinfagnani:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

`changeProperties` passed to `update()`, etc., can return `undefined` from `changeProperties.get()` even if a reactive property doesn't have a nullable type and has a field initializer, because we the first update has `undefined` for all properties.

The current typing of PropertyValues doesn't include `undefined` by default, leading to potential null-dereferences not being seen by TypeScript.

### Reproduction

[...]

PR #3710, opened 2023-03-07
Add undefined to return type of PropertyValues.get()

justinfagnani:
Fixes https://github.com/lit/lit/issues/3709

This is a breaking change for type check, but has no runtime change. Previously we have not considered these breaking changes.


.changeset/tiny-baboons-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
```
augustjk, line 2: We want to bump the `lit` package as well
```suggestion
'@lit/reactive-element': patch
'lit': patch
```

augustjk (approved):

justinfagnani:
@augustjk I rebased this on 3.0 so we don't push even a type-checking break to 2.x

PR #3720, opened 2023-03-10
[ssr-client] Move experimental hydrate modules

### LINKED ISSUE #3713: [ssr-client] Move experimental hydrate modules to `@lit-labs/ssr-client`
### Should this be an RFC?

SSR (@lit-labs/ssr)

### Description

`lit-html/experimental-hydrate.js` and `lit-element/experimental-hydrate-support.js` should be moved to `@lit-labs/ssr-client` so that changes that affect only SSR & hydration concerns don't need to bump versions of the core `lit-html` and `lit-element` packages, especially when those could be considered breaking.

Recent examples:
https://github.com/lit/lit/pull/3667
https://github.com/lit/lit/pull/3677

When the modules are moved to `@lit-labs/client`, I'm thinking "experimental" can be dropped by virtue of it being in the `@lit-labs` namespace. Maybe..
`lit-html/experimental-hydrate.js` -> `@lit-labs/ssr-client/hydrate-lit-html.js`
`lit-element/experimental-hydrate-support.js` -> `@lit-labs/ssr-client/lit-element-hydrate-support.js`

`lit-html/experimental-hydrate.js` has code that depends on having a Node build so that will need to be added for `@lit-labs/ssr-client`.

We can keep the entrypoints in `lit-html` and `lit-element` (re-exported from `lit`) but mark as deprecated, to eventually remove. Update all docs to point to their new location instead.

N/A

augustjk:
Closes #3713 

This PR can be reviewed commit by commit.

1. [Copy experimental hydrate modules to ssr-client](https://github.com/lit/lit/pull/3720/commits/d06d66eb52e9a8aedc2db66fdacef13cade8d6de)
Copies contents of experimental files to new location
2. [Update to fix type errors after copying](https://github.com/lit/lit/pull/3720/commits/e31dc5c6229a38dc34f12a980c22e3f3ff63ebe2)
As `_$LH` is being pulled from `lit-html` package, `@internal` marked properties became invisible.
Some of the `@internal` markings for classes and types that were never meant for public consumption were removed.
For others, like those in `ChildPart` and `AttributePart`, type assertions were added.
3. [Update ssr-client package](https://github.com/lit/lit/pull/3720/commits/0d04bccf6caa6a5d73c9a975c6d8ec9c150ff39d)
Update `package.json` files, exports, wireit configs, rollup to add new entry and node build.
4. [Add deprecation notes to original](https://github.com/lit/lit/pull/3720/commits/88559ffd39f6e43ae11d9b6cbae2f0c5c5053b1d)
I'm leaving the original contents there and marking deprecated, since trying to re-export from `@lit-labs/ssr-client` introduces a circular dependency as `@lit-labs/ssr-client` itself depends on `lit-html`.
`experimental-hydrate-support.js` is side-effectful so added a `console.warn` for deprecation warning.

To do: maybe in a separate PR, would be to update all our docs, examples, and integrations to use the new location.

Open to discussion: if we need to make any changes to these going forward, do we update both locations until we remove the originals from `lit-html` and `lit-element` (Lit3?) or just update the `@lit-labs/ssr-client`?

### DIFF
--- a/packages/labs/ssr-client/package.json
+++ b/packages/labs/ssr-client/package.json
@@ -106,9 +106,10 @@
     "@lit-internal/scripts": "^1.0.0"
   },
   "dependencies": {
+    "@lit-labs/ssr-dom-shim": "^1.1.0",
+    "@lit/reactive-element": "^1.0.0",
     "lit": "^2.0.0",
-    "lit-html": "^2.0.0",
-    "@lit/reactive-element": "^1.0.0"
+    "lit-html": "^2.0.0"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/labs/ssr-client/package.json
+++ b/packages/labs/ssr-client/package.json
@@ -127,7 +127,6 @@
     "@lit-internal/scripts": "^1.0.0"
   },
   "dependencies": {
-    "@lit-labs/ssr-dom-shim": "^1.1.0",
     "@lit/reactive-element": "^1.0.0",
     "lit": "^2.0.0",
     "lit-html": "^2.0.0"
--- a/packages/labs/ssr-client/package.json
+++ b/packages/labs/ssr-client/package.json
@@ -18,10 +18,25 @@
   },
   "exports": {
     ".": {
+      "types": "./development.index.d.ts",
+      "node": {
+        "development": "./node/development/index.js",
+        "default": "./node/index.js"
+      },
       "development": "./development/index.js",
       "default": "./index.js"
     },
+    "./lit-element-hydrate-support.js": {
+      "types": "./development/lit-element-hydrate-support.d.ts",
+      "node": {
+        "development": "./node/development/lit-element-hydrate-support.js",
+        "default": "./node/lit-element-hydrate-support.js"
+      },
+      "development": "./development/lit-element-hydrate-support.js",
+      "default": "./lit-element-hydrate-support.js"
+    },
     "./directives/render-light.js": {
+      "types": "./development/directives/render-light.d.ts",
       "development": "./development/directives/render-light.js",
       "default": "./directives/render-light.js"
     }
@@ -30,7 +45,9 @@
     "/development/",
     "!/development/test/",
     "/directives/",
-    "/index.{d.ts,d.ts.map,js,js.map}"
+    "/lib/",
+    "/index.{d.ts,d.ts.map,js,js.map}",
+    "/lit-element-hydrate-support.{d.ts,d.ts.map,js,js.map}"
   ],
   "scripts": {
     "build": "wireit",
@@ -72,7 +89,8 @@
       "files": [],
       "output": [
         "*.d.ts{,.map}",
-        "directives/*.d.ts{,.map}"
+        "directives/*.d.ts{,.map}",
+        "lib/*.d.ts{,.map}"
       ]
     },
     "build:rollup": {
@@ -86,7 +104,10 @@
       ],
       "output": [
         "index.js{,.map}",
-        "directives/*.js{,.map}"
+        "lit-element-hydrate-support.js{,.map}",
+        "directives/*.js{,.map}",
+        "lib/*.js{,.map}",
+        "node/"
       ]
     },
     "checksize": {
--- a/packages/labs/ssr-client/package.json
+++ b/packages/labs/ssr-client/package.json
@@ -18,7 +18,7 @@
   },
   "exports": {
     ".": {
-      "types": "./development.index.d.ts",
+      "types": "./development/index.d.ts",
       "node": {
         "development": "./node/development/index.js",
         "default": "./node/index.js"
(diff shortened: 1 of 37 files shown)


justinfagnani:
I worry about leaving the existing files in place, even with deprecation warnings. Will versions always be kept in sync? Even if so, will people get version mismatches between lit-html and ssr, like they already have?

I'd be templated to say that since these files are "experimental", that users just need to upgrade to the new locations. This would be the one outward breaking change in 3.0, and I think we can live with that.

justinfagnani:
@augustjk do you need to set the base branch to 3.0 now?

packages/labs/ssr-client/src/lit-element-hydrate-support.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * LitElement support for hydration of content rendered using lit-ssr.
+ *
+ * @packageDocumentation
+ */
+
+import type {PropertyValues} from '@lit/reactive-element';
+import {render, RenderOptions} from 'lit-html';
+import {hydrate} from 'lit-html/experimental-hydrate.js';
+import {HYDRATE_INTERNALS_ATTR_PREFIX} from '@lit-labs/ssr-dom-shim';
```
justinfagnani, line 16: I wish I had caught this in the SSR-internals code review. I don't think the client packages should be depending on the shim package like this. I think we can define this attribute a bit like a protocol and just define it in multiple places so we don't need this dependency. At least I'd want to make a separate module for this in ssr-dom-shim.

packages/labs/ssr-client/package.json
```diff
@@ -106,9 +106,10 @@
     "@lit-internal/scripts": "^1.0.0"
   },
   "dependencies": {
+    "@lit-labs/ssr-dom-shim": "^1.1.0",
```
justinfagnani, line 130: Should you remove this dependency from lit-element?

packages/labs/ssr-client/package.json
```diff
@@ -18,10 +18,25 @@
   },
   "exports": {
     ".": {
+      "types": "./development.index.d.ts",
```
justinfagnani, line 21: Should this be `"./development/index.d.ts"`?

packages/lit-html/src/experimental-hydrate.ts
```diff
@@ -113,6 +113,9 @@ type ChildPartState =
  * @param rootValue
  * @param container
  * @param userOptions
+ *
+ * @deprecated This has been moved to `@lit-labs/ssr-client` and will be removed
+ * in a future release.
```
justinfagnani, line 118: Add a console.warn here too?

justinfagnani (changes requested):
Nice work!

This PR should contain the updates to refer to the files in their new proper location though:

* Update the package.json, .gitignore, rollup.config.js of `lit`, `lit-html`, and `lit-element` to remove entries for the support files.
* Update lit-element/src/test/node-import.ts to use new files locations
* Update SSR demos to use new files locations
* Update 11ty plugin README and demo
* Update labs/nextjs
* Update testing README

packages/labs/ssr-client/package.json
augustjk: Thanks for the find. Fixed.

packages/lit-html/src/experimental-hydrate.ts
augustjk: Added.

packages/labs/ssr-client/src/lit-element-hydrate-support.ts
augustjk: Makse sense. Filed https://github.com/lit/lit/issues/3769

packages/labs/ssr-client/package.json
augustjk: Will remove when deleting we don't reference the const from it for internals attributes.

augustjk:
@justinfagnani Addressed the concerns.

Could you take a look at this commit in particular https://github.com/lit/lit/pull/3720/commits/85597204085098c023b34996419da4352ee47313 ?

Trying to use the hydration modules from ssr-client package surfaced additional issues caused by it being outside so I had to make some changes.

Namely, I added more properties to `lit-html/private-ssr-support.js` and made `TemplateInstance.prototype._parts` into a stable part and renamed to `_$parts`.

justinfagnani:
> Namely, I added more properties to lit-html/private-ssr-support.js and made TemplateInstance.prototype._parts into a stable part and renamed to _$parts.

That probably had to be done anyway to support multiple hydrating versions of lit-html in a render tree.

packages/labs/ssr-client/src/lit-element-hydrate-support.ts
justinfagnani: Can you just copy the value into here now so that this package doesn't need to depend on the ssr-dom-shim?

justinfagnani (approved):
🎉

LasseRosenow:
After this got merged I have this error now:

`Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'C:\Users\Lasse\Desktop\lit-ssr\node_modules\@lit-labs\ssr-client\node\index.js' imported from C:\Users\Lasse\Desktop\lit-ssr\node_modules\@lit-labs\ssr\lib\render-value.js`

augustjk:
@LasseRosenow thanks for the report. this should be fixed in the latest release of `@lit-labs/ssr-client`.

PR #3736, opened 2023-03-17
[labs/context] Add provideHelper for use with isolated components

jpzwarte:
This provides a `provideHelper` method that you can use in cases where your component is isolated from the parent element that provides the context. Some examples here are unit tests, or in Storybook.

```html
<x-foo @context-request=${provideHelper({ foo: 'Foo', bar: false })}></x-foo>
```

### DIFF
--- a/.changeset/lovely-walls-battle.md
+++ b/.changeset/lovely-walls-battle.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/context': minor
+---
+
+Add provideHelper for isolated components


packages/labs/context/src/lib/helper.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import {ContextRequestEvent} from './context-request-event.js';
+import {Context} from './create-context.js';
+
+/**
+ * A helper method for use in situations where a web component is isolated
+ * from the parent that provides the context.
+ *
+ * This helper provides static values for the specified contexts. It does
+ * not support updates via subscribe.
```
rictic, line 15: This doc is missing some important context I think. What actually is this function? How do you use it?

packages/labs/context/src/lib/helper.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export const provideHelper = (params: {[key: string]: unknown}) => {
```
rictic, line 17: Here and elsewhere, the term "helper" is overly generic, like "util" or "meta." This does look like a helpful function, but most good functions are helpful, so naming it a helper doesn't aid the reader in understanding _how_ it's helpful, or in what situations.

Consider a name that communicates more. If I understand this right, I think `staticProvideHandler` is one candidate. `static` because it provides a constant value, and `handler` because it's used as an event handler.

rictic (changes requested):
This is cool, seems very useful for storybook code and other cases where an element that usually has a custom element ancestor is rendered without one.

Please add a few tests

packages/labs/context/src/lib/helper.ts
justinfagnani: I would consider:
- `providerListener()` because they're event listeners, and this is a factory for making listeners that are providers.
- `contextProvider()` though it's a little generic compared to other ways of providing, it reads ok at the use site.
- `contextListener()` which might seem generic on the provide vs consume side, but only providers actually listen for events, while consumers fire them.

rictic: Ah yeah, +1 to listener instead of handler.

Of those, `contextProvider` is my fave, followed by `providerListener`. I think `provide` is so key to what's happening that it needs to be in the name.

PR #3743, opened 2023-03-21
[infra] Reorder release workflow actions for npm caching

augustjk:
Release workflow was broken by https://github.com/lit/lit/pull/3724 as it added the `cache` option to the setup-node action but the `package-lock.json` file wasn't available because the repo hadn't been checked out yet.

This fixes the ordering and accounts for the checkout happening in a nested `lit/` directory.


usergenic (approved):
LGTM!

PR #3749, opened 2023-03-23
Enter prerelease mode for 3.0

justinfagnani:
This will start off the 3.0 branch (already created, but with no commits) and enable us to do prereleases of packages.


.changeset/late-humans-swim.md
augustjk, line 1: Not a big deal at all but do we actually need this file? I know we do a lot of empty changesets to satisfy the workflow but it should only fail when files within the sub packages have been modified. A lot of empty changesets files is noisy when I want to look through to see what's queued for next release.

augustjk (approved):

.changeset/late-humans-swim.md
justinfagnani: oh, I don't know :) just added it in case

PR #3750, opened 2023-03-23
[lit-html] Use toggleAttribute() to simplify boolean attribute parts

justinfagnani:
A small code reduction that we can do when dropping IE11 support.


justinfagnani:
We're going to need `toggleAttribute()` added to the polyfills for this to pass all tests: https://github.com/webcomponents/polyfills/pull/541

rictic (approved):

PR #3751, opened 2023-03-23
[lit-html] Simplify lit-html attribute handling

justinfagnani:
This removes a little workaround we had for IE not iterating attributes in source order. It's not a big change, and I didn't remove the code where we grab attribute names during template string scanning, since that's needed to get the case-preserving name for properties and case-sensitive attributes (for SVG).


kevinpschaaf (approved):
Had to stare at at for a bit, but LGTM!

rictic (approved):

PR #3753, opened 2023-03-23
[all] Use globalThis

justinfagnani:


packages/reactive-element/src/reactive-element.ts
```diff
@@ -48,8 +50,9 @@ let requestUpdateThenable: (name: string) => {
 
 let issueWarning: (code: string, warning: string) => void;
 
-const trustedTypes = (global as unknown as {trustedTypes?: {emptyScript: ''}})
-  .trustedTypes;
+const trustedTypes = (
+  globalThis as unknown as {trustedTypes?: {emptyScript: ''}}
```
augustjk, line 54: This file already has `const global = globalThis;` up top. Do we need this replaced?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -149,7 +152,7 @@ interface DebugLoggingWindow {
  */
 const debugLogEvent = DEV_MODE
   ? (event: ReactiveUnstable.DebugLog.Entry) => {
-      const shouldEmit = (global as unknown as DebugLoggingWindow)
+      const shouldEmit = (globalThis as unknown as DebugLoggingWindow)
```
augustjk, line 155: Same. We already have `global` declared up top.

augustjk (approved):

PR #3761, opened 2023-03-24
Merge main into 3.0

justinfagnani:


augustjk (approved):

augustjk (changes requested):
actually, would it be cleaner to maybe rebase the 3.0 branch off of current main instead of creating these merges?

augustjk (approved):
eh, it's probably fine either way.

PR #3764, opened 2023-03-26
Fixed bug that boolean was not parsed correct

tonqa:
This bugged me for ours: The boolean values are not parsed correct inside the defaultParser.

I found this, when loading a LitElement asynchronously by Vue.js like this: 
`<loading-blocker :loading="loading"></loading-blocker> `

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -335,7 +335,7 @@ export const defaultConverter: ComplexAttributeConverter = {
     let fromValue: unknown = value;
     switch (type) {
       case Boolean:
-        fromValue = value !== null;
+        fromValue = value === null ? null : Boolean(value);
         break;
       case Number:
         fromValue = value === null ? null : Number(value);


tonqa:
I added my CLA, however this test must be triggered again.

augustjk:
Thank you for the contribution but I don't agree with this change. For the case of booleans, we want to return a boolean and never null. The value coming in to `fromAttribute()` is either string if any attribute exists including empty string, or null if the attribute does not exist. If the attribute does not exist, and the value is null, we need to return false. Anything else, we return true. This is the existing implementation.

This change would assign `null` to the value of a boolean property when the attribute doesn't exist, and incorrectly assign `false` when the attribute exists with the value of empty string.

It seems the issue you're having is more a problem that Vue can't tell that particular attribute on the custom element should behave as a boolean attribute so providing `false` in the attribute binding could result in the property having `true` value. You should instead use `.prop` to bind it as a property instead. See https://github.com/vuejs/docs/issues/1972#issuecomment-1259245520

Please file an issue if you think Lit is handling incorrectly.

augustjk:
I feel the "From attribute to property" table above this does a good job of explaining this behavior of the default converter.
https://lit.dev/docs/components/properties/#conversion-type

justinfagnani:
That's true. I clicked on the ToC to go straight to boolean properties and it wasn't as clear then.

PR #3765, opened 2023-03-27
[all] Remove experimental hydration modules

augustjk:
Removes exprimental hydrate modules from `lit-element`, `lit-hteml`, and `lit`.


rictic (approved):
Nice cleanup!

Issue #3767, opened 2023-03-28
[lit-html] styleMap fails when update contains priority directive


regevbr:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

When using `styleMap` with a priority, e.g. `{color: 'red !important}`, the first render works (as it uses string concatenation) but an update, e.g. `{color: 'blue !important'}`, will not work at all and will maintain the previous value.

### Reproduction

The text color should be updated, but it is not:

https://lit.dev/playground/#gist=82856b2ea7bfaa75f1360ff1cf71862e

### Workaround

I have not found a work around

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

2.7.0

### Browser/OS/Node environment

not relevant

regevbr:
I created a PR to fix the issue - https://github.com/lit/lit/pull/3768

PR #3768, opened 2023-03-28
[lit-html] styleMap fails when update contains priority directive

regevbr:
fix: #3767

Added test to verify that issue existed before, and fixed.

The fix is simple - remove the `!important` directive if exists, and set the `priority` argument in the call to `setProperty`

### DIFF
--- a/.changeset/popular-shoes-carry.md
+++ b/.changeset/popular-shoes-carry.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+[lit-html] styleMap fails when update contains priority directive
--- a/.changeset/popular-shoes-carry.md
+++ b/.changeset/popular-shoes-carry.md
@@ -3,4 +3,4 @@
 'lit': patch
 ---
 
-[lit-html] styleMap fails when update contains priority directive
+Fix styleMap's handling of important flags
--- a/packages/lit-html/src/directives/style-map.ts
+++ b/packages/lit-html/src/directives/style-map.ts
@@ -24,6 +24,9 @@ export interface StyleInfo {
   [name: string]: string | undefined | null;
 }
 
+const IMPORTNAT_PRIORITY = 'important';
+const IMPORTANT_DIRECTIVE = '!important';
+
 class StyleMapDirective extends Directive {
   _previousStyleProperties?: Set<string>;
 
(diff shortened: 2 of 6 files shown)


packages/lit-html/src/directives/style-map.ts
```diff
@@ -95,11 +98,13 @@ class StyleMapDirective extends Directive {
       const value = styleInfo[name];
       if (value != null) {
         this._previousStyleProperties.add(name);
-        if (name.includes('-')) {
-          style.setProperty(name, value);
+        const valueToUse = value.replace(IMPORTANT_DIRECTIVE, '');
+        const priority = valueToUse !== value ? IMPORTNAT_PRIORITY : '';
+        if (name.includes('-') || priority) {
+          style.setProperty(name, valueToUse, priority);
```
justinfagnani, line 104: `endsWith()` will be a more efficient test:

```suggestion
        const isImportant = value.endsWith(importantFlag);
        if (name.includes('-') || isImportant) {
          style.setProperty(name, isImportant ? value.slice(0, flagTrim) : value, important);
```

packages/lit-html/src/directives/style-map.ts
```diff
@@ -95,11 +98,13 @@ class StyleMapDirective extends Directive {
...
         } else {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          (style as any)[name] = value;
+          (style as any)[name] = valueToUse;
```
justinfagnani, line 107: We know the important flag isn't set, so you can just use `value` here.

.changeset/popular-shoes-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+[lit-html] styleMap fails when update contains priority directive
```
justinfagnani, line 6: Instead of writing the description of the bug, write the description of the fix. And don't use a package prefix, that's already in the metadata:

```suggestion
Fix styleMap's handling of important flags
```

justinfagnani (changes requested):
Thanks for the fix!

packages/lit-html/src/directives/style-map.ts
regevbr: fixed - you forgot to only set the important flag, if it is important

packages/lit-html/src/directives/style-map.ts
```diff
@@ -24,6 +24,9 @@ export interface StyleInfo {
   [name: string]: string | undefined | null;
 }
 
+const IMPORTNAT_PRIORITY = 'important';
+const IMPORTANT_DIRECTIVE = '!important';
```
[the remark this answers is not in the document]
regevbr:
```js
const important = 'important';
const importantFlag = ' !' + important;
const flagTrim = 0 - importantFlag.length;
console.log("red !important".slice(0, flagTrim))
```
will print `red` without trailing space

regevbr:
@justinfagnani not sure why, but some CI are failing with strange github errors. Have you seen that before?

https://github.com/lit/lit/actions/runs/4544819440
https://github.com/lit/lit/actions/runs/4544819448

justinfagnani (approved):
Thanks!

PR #3771, opened 2023-03-30
[labs/observers] Change filenames to match project convention

justinfagnani:
I noticed that `@lit-labs/observers` was using a different naming convention that we use for the rest of the project. It's a breaking change to fix, but figure it's better to change it now than later with more users.


packages/labs/observers/tsconfig.json
```diff
@@ -20,7 +20,8 @@
     "moduleResolution": "NodeNext",
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
-    "noImplicitOverride": true
+    "noImplicitOverride": true,
+    "noEmitOnError": false
```
rictic, line 24: intentional?

rictic (approved):

packages/labs/observers/tsconfig.json
justinfagnani: oops, no.

PR #3773, opened 2023-03-30
[labs/observers] WIP: try removing initial update from ResizeController

justinfagnani:
We're getting questions about why ResizeController causes two consecutive host updates. It seems like ResizeObservers already call the observer callback unconditionally, so our code to do that as well made there be two updates. Can we just remove our custom initial call to handleChanges? This would let us remove skipInitial altogether.

wdyt @sorvell ?

### DIFF
--- a/packages/labs/observers/src/resize-controller.ts
+++ b/packages/labs/observers/src/resize-controller.ts
@@ -70,7 +70,7 @@ export class ResizeController<T = unknown> implements ReactiveController {
    * state and is performed async by requesting a host update and calling
    * `handleChanges` once by checking and then resetting this flag.
    */
-  private _unobservedUpdate = false;
+  // private _unobservedUpdate = false;
   /**
    * The result of processing the observer's changes via the `callback`
    * function.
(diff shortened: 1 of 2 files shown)


sorvell:
Yeah, I think this stemmed from the fact that the other observers don't have an initial report. 

Couple observations. These tests appear to be wrong here since the `skipInitial` test *should* be failing. I think it's because we're not waiting until ResizeObserver triggers (fix: maybe just change the await here to be `resizeComplete`: https://github.com/lit/lit/blob/25dddb89d6aafe596825f8b23281a7c60fa7737c/packages/labs/observers/src/test/resize-controller_test.ts#L78).

I'd like to preserve the `skipInitial` setting since we have that on the other observers. To do so we could just do something like in the main observer callback (https://github.com/lit/lit/blob/25dddb89d6aafe596825f8b23281a7c60fa7737c/packages/labs/observers/src/resize-controller.ts#L103):
1.  have a `canObserve` flag that's true if `skipInitial` is true
2. and any processing is gated on that 
3. it becomes true after `await host.updateComplete`

AndrewJakubowicz:
After fixing the tests and applying the `skipInitial` setting in the same way that `IntersectionController` does it, we end up with all tests passing except Safari. This is due to WebKit not calling the callback on `observe` at the moment. They are working on changing the behavior to match Chrome and Firefox. See: https://github.com/webcompat/web-bugs/issues/117611#issuecomment-1412749246

justinfagnani:
@AndrewJakubowicz is there a way to detect the browser not calling the callback on `observe()`? Is that initial call synchronous by any chance?

AndrewJakubowicz:
It is not synchronous. The first observe callback runs some time after `observe` is called.

PR #3775, opened 2023-03-31
Upgrade build targets to ES2021

justinfagnani:
I did this across the board: TypeScript target and lib, and Terser `ecma` setting, for all packages and tools tsconfigs. This might be a bit aggressive, but I want to see if this works.


rictic (approved):

augustjk:
Maybe we should drop this line from `CONTRIBUTING.md` since we actually do downlevel
https://github.com/lit/lit/blob/21e22cccf0bbac4147139041c5247b4a4846b524/CONTRIBUTING.md?plain=1#L64

augustjk (approved):

abdonrd:
Looking at this... Wouldn't it help to have a shared `tsconfig.json` between most packages?

augustjk:
> Looking at this... Wouldn't it help to have a shared `tsconfig.json` between most packages?

We totally do this for the lit/lit.dev repo. Maybe we should consider that for here too.

justinfagnani:
Maybe, but not in this PR. Let's keep this as simple as possible.

PR #3788, opened 2023-04-05
Allow numbers to be used as values in styleMap()

### LINKED ISSUE #3787: [lit-html] styleMap breaks if value is a number

### Description

While adding a check for !important in styleMap (line 104) in version 2.7.1 it is not checked if this value is a **string**. In case of z-index the value will be a number and the check will crash as 'endsWith' naturally expects a string.

In our system we're using an older fixed version of Lit (2.6.x) but lit always grabs the latest version for lit-element or lit-html due to ^ which is not what you would expect. Even while setting a fixed version npm i crashed our applications as it installed the latest versions as dependency. Would be better to fix the version for lit sub-packages to be sure to get always the version you request.

Thanks :)

### Reproduction

Use z-index with a number.

### Workaround

Adding lit-html in version 2.7.0 or lower as dependency

### Is this a regression?

### Affected versions

2.7.1

not relevant

justinfagnani:
Fixes #3787

### DIFF
--- a/.changeset/slimy-clouds-roll.md
+++ b/.changeset/slimy-clouds-roll.md
@@ -1,7 +1,6 @@
 ---
 'lit-html': patch
 'lit': patch
-'lit-element': patch
 ---
 
 Allow numbers to be used as values in styleMap()
--- a/.changeset/slimy-clouds-roll.md
+++ b/.changeset/slimy-clouds-roll.md
@@ -1,0 +1,7 @@
+---
+'lit-html': patch
+'lit': patch
+'lit-element': patch
+---
+
+Allow numbers to be used as values in styleMap()
--- a/packages/lit-html/src/test/directives/style-map_test.ts
+++ b/packages/lit-html/src/test/directives/style-map_test.ts
@@ -89,6 +89,7 @@ suite('styleMap', () => {
     assert.equal(el.style.marginTop, '2px');
     assert.equal(el.style.paddingBottom, '4px');
     assert.equal(el.style.opacity, '0.5');
+    assert.equal(el.style.zIndex, '10');
     renderStyleMap({
       marginTop: '4px',
       paddingBottom: '8px',
--- a/packages/lit-html/src/test/directives/style-map_test.ts
+++ b/packages/lit-html/src/test/directives/style-map_test.ts
@@ -98,6 +98,7 @@ suite('styleMap', () => {
     assert.equal(el.style.marginTop, '4px');
     assert.equal(el.style.paddingBottom, '8px');
     assert.equal(el.style.opacity, '0.55');
+    assert.equal(el.style.zIndex, '1');
   });
 
   test('removes properties', () => {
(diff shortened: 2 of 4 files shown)


.changeset/slimy-clouds-roll.md
```diff
@@ -0,0 +1,7 @@
+---
+'lit-html': patch
+'lit': patch
+'lit-element': patch
```
augustjk, line 4: This isn't touching anything in `lit-element`. I think it's okay not to bump that but keep `lit`?
```suggestion
'lit-html': patch
'lit': patch
```

packages/lit-html/src/test/directives/style-map_test.ts
```diff
@@ -79,12 +79,22 @@ suite('styleMap', () => {
...
-    renderStyleMap({marginTop: '4px', paddingBottom: '8px', opacity: '0.55'});
+    renderStyleMap({
+      marginTop: '4px',
+      paddingBottom: '8px',
+      opacity: '0.55',
+      'z-index': 1,
+    });
     assert.equal(el.style.marginTop, '4px');
     assert.equal(el.style.paddingBottom, '8px');
     assert.equal(el.style.opacity, '0.55');
```
augustjk, line 100:
```suggestion
    assert.equal(el.style.opacity, '0.55');
    assert.equal(el.style.zIndex, '1');
```

augustjk (changes requested):

.changeset/slimy-clouds-roll.md
justinfagnani: I think we need to revisit our convention here and write down the decision and rationale, so it's easy to recall.

packages/tests/src/web-test-runner.config.ts
```diff
@@ -24,8 +24,8 @@ const browserPresets = {
   // Default set of Playwright browsers to test when running locally.
   local: [
     'chromium', // keep browsers on separate lines
-    'firefox', // to make it easier to comment out
-    'webkit', // individual browsers
+    // 'firefox', // to make it easier to comment out
+    // 'webkit', // individual browsers
```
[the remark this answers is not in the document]
justinfagnani: Oops!

packages/lit-html/src/test/directives/style-map_test.ts
```diff
@@ -79,12 +79,22 @@ suite('styleMap', () => {
...
+    renderStyleMap({
+      marginTop: '2px',
+      'padding-bottom': '4px',
+      opacity: '0.5',
+      'z-index': 10,
+    });
     const el = container.firstElementChild as HTMLElement;
     assert.equal(el.style.marginTop, '2px');
     assert.equal(el.style.paddingBottom, '4px');
     assert.equal(el.style.opacity, '0.5');
```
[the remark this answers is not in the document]
justinfagnani: Maybe... the fact that it type-checks and doesn't throw is already a big step up in coverage.

augustjk (approved):

PR #3793, opened 2023-04-06
[labs/ssr-dom-shim] Warn instead of throwing on repeat custom element registration in dev mode

### LINKED ISSUE #3672: [labs/nextjs] Custom element registry can cause error on dev mode rebuild

### Description

While running Next.js's dev mode, making changes to any file that include a custom element registration cause the error:
```
error - Error: Failed to execute 'define' on 'CustomElementRegistry'
```

This isn't necessarily a problem with the `labs/nextjs` package in itself but is due to the `CustomElementRegistry` shim included with node builds of the lit package, and how we're running the SSR code in the global scope.

I anticipate most use case of custom elements in Next.js wouldn't be bothered by this much as they'd be importing a ready-to-use component and not touching its definition to be re-registered. For components that require users to import and register separately, this may be more of a problem.

### Reproduction

.

### Workaround

For now, restart dev server.

Possible fixes for this:

- Make our `CustomElementRegistry` shim not throw on re-registers. This lives in `@lit-labs/ssr-dom-shim` https://github.com/lit/lit/blob/e00f6f52199d5dbc08d4c15f62380422e77cde7f/packages/labs/ssr-dom-shim/src/index.ts#L89-L94
- Patch the `CustomElementRegistry` somewhere in the labs/nextjs plugin to not throw
- When available, incorporate https://github.com/lit/lit/issues/3627

### Is this a regression?

### Affected versions

N/A

N/A

augustjk:
Fixes #3672 

This was also an issue with Astro's dev mode. This addresses that as well.

### DIFF
--- a/packages/labs/ssr-dom-shim/src/index.ts
+++ b/packages/labs/ssr-dom-shim/src/index.ts
@@ -111,10 +111,17 @@ const CustomElementRegistryShim = class CustomElementRegistry {
 
   define(name: string, ctor: CustomHTMLElementConstructor) {
     if (this.__definitions.has(name)) {
-      throw new Error(
-        `Failed to execute 'define' on 'CustomElementRegistry': ` +
-          `the name "${name}" has already been used with this registry`
-      );
+      if (process.env.NODE_ENV === 'development') {
+        console.warn(
+          `'CustomElementRegistry' already has "${name}" defined. ` +
+            `Overriding in development mode.`
+        );
+      } else {
+        throw new Error(
+          `Failed to execute 'define' on 'CustomElementRegistry': ` +
+            `the name "${name}" has already been used with this registry`
+        );
+      }
     }
     this.__definitions.set(name, {
       ctor,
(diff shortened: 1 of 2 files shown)


packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -111,10 +111,17 @@ const CustomElementRegistryShim = class CustomElementRegistry {
...
   define(name: string, ctor: CustomHTMLElementConstructor) {
     if (this.__definitions.has(name)) {
-      throw new Error(
-        `Failed to execute 'define' on 'CustomElementRegistry': ` +
-          `the name "${name}" has already been used with this registry`
-      );
+      if (process.env.NODE_ENV === 'development') {
+        console.warn(
+          `'CustomElementRegistry' already has "${name}" defined. ` +
+            `Overriding in development mode.`
```
justinfagnani, line 117: I don't fully grok the "Overriding in development mode." phrase.

I'd make it as clear as possible what the problem is and what will be different in production mode. So say that this is either caused by HMR and is fine, or must be fixed, ideally how to check it (run same page in prod mode?)

justinfagnani (approved):

packages/labs/ssr-dom-shim/src/index.ts
augustjk: Yeah good call. It seems better to be a bit more verbose here. Updated!

PR #3799, opened 2023-04-11
Coerce toggleAttribute's second arg to boolean

rictic:
Otherwise binding the value `undefined` to a boolean part would mean `toggle` rather than `false`


justinfagnani (approved):
Needs a test, but lgtm!

PR #3800, opened 2023-04-11
Relax the typescript version for localize-tools

### LINKED ISSUE #3739: [localize-tools] Transform mode using rollup is broken with typescript >= ~4.9

Localize (@lit/localize)

### Description

When I run `rollup -c` with the `localeTransformer` configured, it no longer replaces any `msg('...')` calls in the outputted files.

### Reproduction

See https://github.com/jpzwarte/lit-localize-rollup-bug

### Workaround

Not very practical, but if I downgrade the project version of typescript to 4.7.4 (the same as what the Lit tools use), then it works as expected again.

### Is this a regression?

### Affected versions

`@lit/localize-tools@0.6.7`

TypeScript ~4.7: ✅
TypeScript ~4.8: ✅
TypeScript ~4.9: ❌
TypeScript ~5.0: ❌

-

jpzwarte:
Fixes #3739 

By relaxing the typescript version from `~4.7.4` to `^4.7.4`, we allow it to use `4.8` and `4.9` as well. This fixes the problem with localize-tools where it uses typescript 4.7, but the project used 4.8 or 4.9. This breaks because of breaking changes between 4.7 and greater.

Note: so `5.0` is not valid here. I'm not sure whether we want to allow that yet.

### DIFF
--- a/.changeset/angry-rockets-search.md
+++ b/.changeset/angry-rockets-search.md
@@ -1,0 +1,5 @@
+---
+'@lit/localize-tools': patch
+---
+
+Relax the typescript version for compatibility with typescript > 4.7 && < 5.0
--- a/package-lock.json
+++ b/package-lock.json
@@ -25770,7 +25770,7 @@
       }
     },
     "packages/lit": {
-      "version": "2.7.1",
+      "version": "2.7.2",
       "license": "BSD-3-Clause",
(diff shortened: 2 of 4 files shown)


aomarks (approved):

aomarks:
Thank you!

PR #3812, opened 2023-04-17
[labs/analyzer] CSS variable fallbacks

### LINKED ISSUE #3811: [labs/analyzer] css property default (i.e. fallback) values
### Should this be an RFC?

### Description

As discussed at the Open Lit Eng Meeting March 14 2023, we wish to add a syntax to `@cssProp[erty]` JSDoc class tags so component authors can specify css custom property fallback values.

```css
color: var(--my-element-color, #323fff);
```
```js
/**
 * @cssProp --my-element-color
 */
class MyElement extends LitElement {}
```
This already produces 
```json
"cssProperties": [
  {
    "name": "--my-element-color"
  },
]
```
But we'd like to produce
```json
"cssProperties": [
  {
    "name": "--my-element-color",
    "default": "#323fff"
  },
]
```
## Non-goals:
- analyzing css sources
- analyzing lit `css` template tag objects
## Open Questions
do we want to also produce type data?
```json
"syntax": "<color>",
```

Make users manually update CEM

bennypowers:
Closes #3811

```ts
/**
 * @license
 * Copyright 2022 Google LLC
 */

import {customElement} from 'lit/decorators.js';

/**
 * @cssProperty [--default-no-description=#324fff]
 * @cssProperty [--default-with-description=#324fff] Description for --default-with-description
 * with wraparound
 * @cssProperty [--default-with-description-dash=#324fff] - Description for --default-with-description-dash
 */
@customElement('element-a')
export class ElementA extends LitElement {}
```
```json
{
  "cssProperties": [
    { "name": "--default-no-description", "default": "#324fff" },
    { "name": "--default-with-description", "default": "#324fff" },
    {
      "name": "--default-with-description-dash",
      "description": "Description for --default-with-description-dash",
      "default": "#324fff"
    },
  ]
}
```

### DIFF
--- a/packages/labs/analyzer/src/lib/javascript/jsdoc.ts
+++ b/packages/labs/analyzer/src/lib/javascript/jsdoc.ts
@@ -16,7 +16,6 @@ import {
   Described,
   TypedNamedDescribed,
   DeprecatableDescribed,
-  NamedDescribedDefault,
   NamedDescribed,
 } from '../model.js';
 
@@ -43,26 +42,15 @@ const normalizeLineEndings = (s: string) => s.replace(/\r/g, '').trim();
 const parseNameTypeDescRE =
   /^(?<name>\S+)(?:\s+{(?<type>.*)})?(?:\s+-\s+)?(?<description>[\s\S]*)$/m;
 
-// Regex for parsing name and description from JSDoc comments
-const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
-
 // Regex for parsing optional name and description from JSDoc comments, where
 // the dash is required before the description (syntax for `@slot` tag, whose
 // default slot has no name)
-const parseNameDashDescRE = /^(?<name>^\S*)?(?:\s+-\s+(?<description>.+))?$/m;
+const parseNameDashDescRE =
+  /^\[?(?<name>[^[\]\s=]+)(?:=(?<defaultValue>[^\]]+))?\]?\s+-\s+(?<description>[\s\S]*)$/;
 
 // Regex for parsing optional name, default, and description from JSDoc comments
-const parseNameDefaultDescRE =
-  /^\[(?<name>[^[\s=]+)=?(?<defaultValue>[^\]]+)?\](?:\s+-\s+(?<description>.+))?/m;
-
-// Regex for parsing optional name and description from JSDoc comments, where
-// the dash is required before the description (syntax for `@slot` tag, whose
-// default slot has no name)
-const parseNameDefaultDashDescRE =
-  /^\[(?<name>[^[\s=]+)=?(?<defaultValue>[^\]]+)?\](?:\s+-\s+(?<description>.+))?$/m;
-
-// is it an optional jsdoc prop name?
-const isOptionalJSDocNameRegex = /^\[.+(?:=.+)?\]/;
+const parseNameDescRE =
+  /^\[?(?<name>[^[\]\s=]+)(?:=(?<defaultValue>[^\]]+))?\]?(?<description>[\s\S]*)$/;
 
 const getJSDocTagComment = (tag: ts.JSDocTag) => {
   let {comment} = tag;
@@ -118,7 +106,8 @@ function makeDashParseError(tag: ts.JSDocTag, requireDash: boolean) {
     tag,
     `Unexpected JSDoc format.${
       requireDash
-        ? ` Tag must contain a whitespace-separated dash between the name and description, i.e. '@slot header - This is the description'`
+        ? ' Tag must contain a whitespace-separated dash between the name and description, ' +
+          "i.e. '@slot header - This is the description'"
         : ''
     }`
   );
@@ -141,41 +130,25 @@ function makeDashParseError(tag: ts.JSDocTag, requireDash: boolean) {
 export const parseNamedJSDocInfo = (
   tag: ts.JSDocTag,
   requireDash = false
-): NamedDescribedDefault | undefined => {
+): NamedDescribed | undefined => {
   const comment = getJSDocTagComment(tag);
   if (comment == undefined) {
     return undefined;
   }
-  if (isOptionalJSDocNameRegex.test(comment)) {
-    const nameDefaultDesc = comment.match(
-      requireDash ? parseNameDefaultDashDescRE : parseNameDefaultDescRE
-    );
-    if (nameDefaultDesc === null) {
-      throw makeDashParseError(tag, requireDash);
-    }
-    const {name, description, defaultValue} = nameDefaultDesc.groups!;
-    const info: NamedDescribedDefault = {name};
-    if (description?.length > 0) {
-      info.description = normalizeLineEndings(description);
-    }
-    if (defaultValue?.length > 0) {
-      info.default = defaultValue;
-    }
-    return info;
-  } else {
-    const nameDesc = comment.match(
-      requireDash ? parseNameDashDescRE : parseNameDescRE
-    );
-    if (nameDesc === null) {
-      throw makeDashParseError(tag, requireDash);
-    }
-    const {name, description} = nameDesc.groups!;
-    const info: NamedDescribed = {name};
-    if (description?.length > 0) {
-      info.description = normalizeLineEndings(description);
-    }
-    return info;
+  const regex = requireDash ? parseNameDashDescRE : parseNameDescRE;
+  const nameDesc = comment.match(regex);
+  if (nameDesc === null) {
+    throw makeDashParseError(tag, requireDash);
   }
+  const {name, description, defaultValue} = nameDesc.groups!;
+  const info: NamedDescribed = {name};
+  if (description?.length > 0) {
+    info.description = normalizeLineEndings(description);
+  }
+  if (defaultValue?.length > 0) {
+    info.default = defaultValue;
+  }
+  return info;
 };
 
 /**
--- a/packages/labs/analyzer/src/lib/javascript/jsdoc.ts
+++ b/packages/labs/analyzer/src/lib/javascript/jsdoc.ts
@@ -49,18 +49,20 @@ const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
 // Regex for parsing optional name and description from JSDoc comments, where
 // the dash is required before the description (syntax for `@slot` tag, whose
 // default slot has no name)
-const parseNameDashDescRE =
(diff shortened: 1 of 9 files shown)


packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -51,6 +52,16 @@ const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
 const parseNameDashDescRE =
   /^(?<name>^\S*)?(?:\s+-\s+(?<description>[\s\S]*))?$/m;
 
+// Regex for parsing name and description from JSDoc comments
+const parseNameDefaultDescRE =
+  /^\[(?<name>[^[\s=]*)=(?<defaultValue>[^\]]*)\](?:\s+-\s+(?<description>[\s\S]*))?/m;
```
43081j, line 57: do you also want to account for optional css props without defaults? for consistency with `@param` and other built-ins?

```ts
/**
 * @cssProp [--some-optional-css-prop]
 * @cssProp --some-required-css-prop
 * @cssProp [--some-optional-css-prop-with-value=123]
 */
```

the square brackets usually mean "optional". you can only have default values for optional things, so thats why you also only ever see default values in such positions

may be worth keeping the supported syntax aligned with the regular jsdoc syntax

also the `\s\S` is basically behaving as `.+` here right? the description is already optional so you could dumb it down to `(?<description>.+)?`

the dash also looks required here but your post suggests it isn't. hows that working?

bennypowers: I think `\s\S` is a typo
Dash shouldn't be required I'll take a look
I'm not sure what the meaning of "optional but without default" is in css world, but maybe `@property` makes that a runtime thing so yeah i'll add that

Thanks for the review!

43081j: yeah in the css world it doesn't make much sense to have an optional vs required property (they're all optional...).

however i imagine a fair amount of people will assume you can do `cssProp --foo` rather than `cssProp [--foo]`.

`@param` and friends already define this behaviour well enough so we may as well follow how they work, even if it doesn't apply as much in css-world

43081j: ah and also you probably want to use `+` rather than `*` on the name and default value since you can't have an empty one (`[]`)

so `(?<name>[^[\s-]+)=(?<defaultValue>[^\]]+)`

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -109,34 +131,49 @@ export const parseNamedTypedJSDocInfo = (tag: ts.JSDocTag) => {
...
-  );
-  if (nameDesc === null) {
-    throw new DiagnosticsError(
-      tag,
-      `Unexpected JSDoc format.${
-        parseNameDashDescRE
-          ? ` Tag must contain a whitespace-separated dash between the name and description, i.e. '@slot header - This is the description'`
-          : ''
-      }`
+  if (comment.match(/^\[.+=.+\]/)) {
```
43081j, line 147: this should be a `test` rather than a `match` really, e.g. `/^whatever/.test(comment)` and then you can maybe put the regex in a const somewhere too

43081j: im wondering if we should also just make this a special case for css props rather than matching syntactically

we could just check if the tag is `cssProp` and run this alternative logic (which i suppose would also line up with my other suggestion of supporting `cssProp --foo`, since this regex wouldn't allow that right now)

bennypowers: >  supporting cssProp --foo, since this regex wouldn't

Help me understand this, since the test cases include this case.

https://github.com/lit/lit/blob/3e510743d314bc7399aec2a45cfd7a67e2cfa589/packages/labs/analyzer/test-files/ts/vanilla-jsdoc/src/element-a.ts#L27

Do you mean you'd prefer to not branch here? Me too. I initially tried to do it in one big honking regexp but that seemed to be prohibitively finicky, so I backed it off to an `if else`. I'm not sure how we'd support both `cssProp --foo` and `cssProp [--foo=bar]` without branching but also with a reasonably readable regexp (= superhero name up for grabs)

WDYT?

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
bennypowers: PTAL

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -109,34 +133,49 @@ export const parseNamedTypedJSDocInfo = (tag: ts.JSDocTag) => {
...
-  );
-  if (nameDesc === null) {
-    throw new DiagnosticsError(
-      tag,
-      `Unexpected JSDoc format.${
-        parseNameDashDescRE
-          ? ` Tag must contain a whitespace-separated dash between the name and description, i.e. '@slot header - This is the description'`
-          : ''
-      }`
+  if (isOptionalJSDocNameRegex.test(comment)) {
```
justinfagnani, line 149: I think maybe a simpler way to structure this is to choose the regex based on `requireDash`, then follow the same code path regardless of default or not:

ie

```ts
const regex = requireDash ? parseNameDefaultDashDescRE : parseNameDefaultDescRE;
const result = comment.match(regex);
if (result === null) { /* error */ }
const {name, description, defaultValue} = nameDefaultDesc.groups!;
// ...
```

This requires making a regex that can handle both plain names and names inside balanced `[]`s. That'll be a little duplication in the regex, but will end up with fewer regex evaluations.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
43081j: i misunderstood partially, i didn't realise the `else` is what covers `@cssProp --foo`

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -48,8 +49,20 @@ const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
 // Regex for parsing optional name and description from JSDoc comments, where
 // the dash is required before the description (syntax for `@slot` tag, whose
 // default slot has no name)
-const parseNameDashDescRE =
-  /^(?<name>^\S*)?(?:\s+-\s+(?<description>[\s\S]*))?$/m;
+const parseNameDashDescRE = /^(?<name>^\S*)?(?:\s+-\s+(?<description>.+))?$/m;
+
+// Regex for parsing optional name, default, and description from JSDoc comments
+const parseNameDefaultDescRE =
+  /^\[(?<name>[^[\s=]+)=?(?<defaultValue>[^\]]+)?\](?:\s+-\s+(?<description>.+))?/m;
```
43081j, line 56: don't you actually want the subgroup to be optional rather than the individual `=` to be optional?

i.e. instead of `=?(?<defaultValue>whatever)?`, you want `(?:=(?<defaultValue>whatever))?`

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
bennypowers: PTAL, I think I have this for the non-dash case, but the dash case needs a little attention

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -100,6 +113,17 @@ export const parseNamedTypedJSDocInfo = (tag: ts.JSDocTag) => {
   return info;
 };
 
+function makeDashParseError(tag: ts.JSDocTag, requireDash: boolean) {
+  return new DiagnosticsError(
+    tag,
+    `Unexpected JSDoc format.${
+      requireDash
+        ? ` Tag must contain a whitespace-separated dash between the name and description, i.e. '@slot header - This is the description'`
```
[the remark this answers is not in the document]
bennypowers: yes

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -48,8 +49,20 @@ const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
 // Regex for parsing optional name and description from JSDoc comments, where
 // the dash is required before the description (syntax for `@slot` tag, whose
 // default slot has no name)
-const parseNameDashDescRE =
-  /^(?<name>^\S*)?(?:\s+-\s+(?<description>[\s\S]*))?$/m;
+const parseNameDashDescRE = /^(?<name>^\S*)?(?:\s+-\s+(?<description>.+))?$/m;
+
+// Regex for parsing optional name, default, and description from JSDoc comments
+const parseNameDefaultDescRE =
```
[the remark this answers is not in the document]
bennypowers: working on it, PTAL at [effea45](https://github.com/lit/lit/pull/3812/commits/effea45049179d51c0998e8ecdb6b5ca416e20da)

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -48,8 +49,20 @@ const parseNameDescRE = /^(?<name>^\S+)(?:\s?-\s+)?(?<description>[\s\S]*)$/m;
 // Regex for parsing optional name and description from JSDoc comments, where
 // the dash is required before the description (syntax for `@slot` tag, whose
 // default slot has no name)
-const parseNameDashDescRE =
-  /^(?<name>^\S*)?(?:\s+-\s+(?<description>[\s\S]*))?$/m;
+const parseNameDashDescRE = /^(?<name>^\S*)?(?:\s+-\s+(?<description>.+))?$/m;
```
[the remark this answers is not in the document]
bennypowers: replaced these

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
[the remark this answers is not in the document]
bennypowers: ok tests are green now and the RE looks on the surface to not be entirely abhorrent

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
bennypowers: giv'er another look, if you please

43081j (approved):
looks good to me 🥳

justinfagnani (approved):
Nice!

PR #3814, opened 2023-04-18
Update to TypeScript 5.0

rictic:


rictic:
.d.ts changes look chill: https://gist.github.com/rictic/1ddd4afd56b88f008609d05f05e4ce80

package.json
```diff
@@ -206,6 +206,7 @@
     "@web/test-runner-mocha": "^0.7.5",
     "@web/test-runner-playwright": "^0.9.0",
     "@web/test-runner-saucelabs": "^0.8.0",
+    "chai": "^4.3.7",
```
justinfagnani, line 209: why are we adding `chai`? Does it pull along types? There's also `@types/chai`

packages/labs/context/package.json
```diff
@@ -147,6 +148,7 @@
     "lit": "^2.7.0"
   },
   "devDependencies": {
+    "@esm-bundle/chai": "4.3.4",
```
justinfagnani, line 151: this with `"@esm-bundle/chai": "4.3.4-fix.0"` will give us multiple copies. Can we standardize on one version?

packages/localize-tools/src/modes/transform.ts
```diff
@@ -218,17 +221,17 @@ class Transformer {
           '_LIT_LOCALIZE_CONFIGURE_TRANSFORM_LOCALIZATION_'
         )
       ) {
-        return ts.createObjectLiteral(
+        return factory.createObjectLiteralExpression(
```
justinfagnani, line 224: Did this API exist on typescript@4.7? Checking if this is a breaking change for localize-tools.

packages/ts-transformers/package.json
```diff
@@ -57,7 +57,7 @@
   },
   "dependencies": {
     "ts-clone-node": "^1.0.0",
-    "typescript": "~4.7.4"
+    "typescript": "~5.0.0"
```
justinfagnani, line 60: Is this a breaking change? Thinking about lit.dev here.

packages/localize-tools/src/modes/transform.ts
jpzwarte: If it is, we probably need https://github.com/lit/lit/pull/3800, but with `~5.0.0` cc @aomarks

jpzwarte: Also, if it *is* compatible, then we need to add `~5.0.0` (see https://github.com/lit/lit/pull/3800)

rictic: There's breaking changes in the TS APIs, so will do this as a breaking change to localize-tools and ts-transformers

packages/ts-transformers/package.json
rictic: Yeah, marked as such

package.json
rictic: Was trying to fix a lint error but ended up going with another approach, removed

packages/labs/context/package.json
rictic: Done

package.json
rictic: Was trying to fix a lint error but ended up going with another approach, removed

packages/labs/context/package.json
rictic: Done

justinfagnani (approved):

PR #3816, opened 2023-04-19
Use Symbol.for in dev mode

rictic:
This makes it much easier to maintain state for an element during HMR because keys stay the same across different versions of the class.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -643,7 +643,9 @@ export abstract class ReactiveElement
     // still overwrite it
     if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
       const key = DEV_MODE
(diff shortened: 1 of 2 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -643,7 +643,7 @@ export abstract class ReactiveElement
     // still overwrite it
     if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
       const key = DEV_MODE
-        ? Symbol(`${String(name)} (@property() cache)`)
+        ? Symbol.for(`${String(name)} (@property() cache)`)
```
justinfagnani, line 646: Add a comment about why you're using Symbol.for()

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts
rictic: Done!

PR #3819, opened 2023-04-19
[labs/virtualizer] Fix #3815

### LINKED ISSUE #3815: [labs/virtualizer] Masonry: size and range bugs when the last item placed isn't the one extending the furthest

### Description

When calculating virtualizer size and range, the masonry layout currently assumes that the item placed last is the one that extends furthest along the block axis. This is not always the case, which means that the virtualizer is sometimes not sized properly and the range is not always calculated correctly.

### Reproduction

Sizing issue: https://lit.dev/playground/#gist=ff60a155695034225bf49d2d9042e97a
Range issue: https://lit.dev/playground/#gist=d8e9639ad3e68cba0a928783594c1fbb

### Workaround

It's probably possible to engineer a hack to work around the sizing issue, but there is no obvious workaround for the range issue.

### Is this a regression?

### Affected versions

Failing in 2.0.0

N/A

graynorton:


usergenic (approved):
LGTM!

justinfagnani (approved):
lgtm

Trying to understand what's going on though, and... it would be great to start adding more explanatory comments as changes are made. I'd like to understand what layoutChildren is doing from a very high level.

justinfagnani:
@usergenic do you want to merge this?

usergenic:
> @usergenic do you want to merge this?

Yes thanks!

PR #3820, opened 2023-04-20
Fixes #3817

### LINKED ISSUE #3817: [labs/eleventy-plugin-lit] Change eleventy-plugin-lit example HTML to show content if JS is disabled
### Should this be an RFC?

Eleventy Plugin (@lit-labs/eleventy-plugin-lit)

### Description

The current example HTML default layout for @lit-labs/eleventy-plugin-lit has the following code:

```html
  <body dsd-pending>
    <script>
      if (HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
        // This browser has native declarative shadow DOM support, so we can
        // allow painting immediately.
        document.body.removeAttribute('dsd-pending');
      }
    </script>

```

However when JS is disabled this means the body remains hidden. What about changing the example to the following?

```html
  <body>
    <script>
      if (!HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
        // This browser does not have native declarative shadow DOM support, so we hide
        // the body until the template-shadowroot polyfill has executed
        document.body.setAttribute('dsd-pending');
      }
    </script>
```

That way at least some content will be displayed even when JS is disabled.

Can just do it yourself in your own project.

alexgagnon:
Update the example HTML document for [labs/eleventy-plugin-lit] to only hide the body element if the browser doesn't support DSD and needs to run the template-shadowroot polyfill.

### DIFF
--- a/packages/labs/eleventy-plugin-lit/README.md
+++ b/packages/labs/eleventy-plugin-lit/README.md
@@ -349,7 +349,7 @@ The file `_includes/default.html` would then contain the following:
       if (!HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
         // This browser does not have native declarative shadow DOM support, so we hide
         // the body until the template-shadowroot polyfill has executed
-        document.body.setAttribute('dsd-pending');
+        document.body.setAttribute('dsd-pending', '');
       }
     </script>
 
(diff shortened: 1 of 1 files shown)


augustjk:
@alexgagnon Thank you so much for taking the time to make this PR. I won't merge this per the the discussion here https://github.com/lit/lit/issues/3817 which is entirely my fault for not thinking it through more. We really appreciate it and please let us know if you think of any other suggestions!

PR #3825, opened 2023-04-20
[lit-html] Static `html` should not add consumed value to `TemplateResult`

### LINKED ISSUE #2246: [labs/ssr] Cannot SSR a dynamic tag with `unsafeStatic`

augustjk:
Fixes #2246 

SSR of static html was erroring due to mismatch of `partIndex` as described in issue above because we were comparing that to the length of the `values` array in `TemplateResult`.

When a static value was the last value of a template, it ended up adding the static value object to the `TemplateResult`. This wasn't an issue if a non-static value was the last value.

i.e.
```js
const tag = literal`p`;
const template = html`<${tag}>Hello, ${this.name}!</${tag}`;
```
resulted in
```js
{
  strings: ['<p>Hello, ', '!</p>'],
  values: ['World', {_$litStatic$: 'p', r: Symbol()}],
}
```
when it should be
```js
{
  strings: ['<p>Hello, ', '!</p>'],
  values: ['World'],
}
```

In client rendering this wasn't an issue as we just never accessed that value. In SSR, it was causing the error due to the unexpected extra length.

### DIFF
--- a/packages/lit-html/src/static.ts
+++ b/packages/lit-html/src/static.ts
@@ -131,7 +131,10 @@ export const withStatic =
         s += staticValue + strings[++i];
         hasStatics = true;
       }
-      dynamicValues.push(dynamicValue);
+      // If the last value is static, we don't need to push it.
+      if (i !== l) {
+        dynamicValues.push(dynamicValue);
+      }
       staticStrings.push(s);
       i++;
     }
--- a/.changeset/hot-rabbits-promise.md
+++ b/.changeset/hot-rabbits-promise.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+`static-html` no longer adds an item to `TemplateResult`'s value array for the last consumed static value. This fixes an error with server-side rendering of static html.
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -7,6 +7,7 @@
 import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
 
 import {html, noChange, nothing, Part} from 'lit';
+import {html as staticHtml, literal} from 'lit/static-html.js';
 import {
   directive,
   Directive,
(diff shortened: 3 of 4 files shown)


packages/lit-html/src/static.ts
```diff
@@ -131,7 +131,10 @@ export const withStatic =
         s += staticValue + strings[++i];
         hasStatics = true;
       }
-      dynamicValues.push(dynamicValue);
+      // If the last value is static, we don't need to push it.
+      if (i !== l) {
+        dynamicValues.push(dynamicValue);
```
justinfagnani, line 136: Looking back at the original code, I don't love that `dynamicValue` apparently can hold a static value. Nice fix for now, but makes me want to clean this up later.

justinfagnani (approved):
Thanks for the quick turnaround!

augustjk:
Failing tests here due to https://github.com/lit/lit/issues/3827
This PR https://github.com/lit/lit/pull/3828 is a band-aid for this issue if we want to unblock our tests for now.

Issue #3827, opened 2023-04-21
[labs/gen-wrapper-vue] Build of generated component fails due to type check


augustjk:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

The `gen-wrapper-vue` test is failing due to failing build. Getting the following TS error on build.

![image](https://user-images.githubusercontent.com/40413829/233513379-06d14a39-d72a-4f02-b950-9b4da30e471a.png)

This seems to be caused by a bump in `vue-tsc` to version `1.4`.

I don't know enough about how Vue components should be written to know whether this is a regression or a fix on `vue-tsc`'s part.

### Reproduction

`npm test` in `packages/labs/gen-wrapper-vue`

### Workaround

Pin the version of `vue-tsc` in generated package to `1.3`

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

0.2.3

### Browser/OS/Node environment

Node 18

augustjk:
Filed issue with `vue-tsc`
https://github.com/vuejs/language-tools/issues/2638

augustjk:
This was fixed upstream by `vue-tsc`.

PR #3828, opened 2023-04-21
[labs/gen-wrapper-vue] Pin version of `vue-tsc` to ~1.3

augustjk:
See #3827 

Above issue is causing test failures in the pipeline.

It could be that the generated Vue code needs to be updated to alleviate the type error, but I don't know enough about Vue to tell if that's the correct thing to do. At least this is how it used to work so I'm proposing we pin this for now until we know if the type error seen in the latest version is legit.

### DIFF
--- a/.changeset/loud-poems-pay.md
+++ b/.changeset/loud-poems-pay.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/gen-wrapper-vue': patch
+---
+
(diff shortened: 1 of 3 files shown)


justinfagnani (approved):

augustjk:
This is no longer necessary as `vue-tsc` has been updated.

PR #3834, opened 2023-04-22
Fixes ContextRoot for requesters in nested shadowRoots

### LINKED ISSUE #3833: [labs/context] ContextRoot fails for consumers in nested shadowRoots

Context (@lit-labs/context)

### Description

When a ContextRoot caches a `context-request` event, it stores the `event.target`. However, if the event is retargeted, this is not the element that fired the event and therefore needs the context to be fulfilled. 

Likely instead, it should be caching the root target, or `event.composedPath()[0]`.

### Reproduction

https://lit.dev/playground/#gist=83c8ff4813b2edd082b29fb03557a82b

The bottom element's context value in the example above is not provided due to this issue.

### Workaround

### Is this a regression?

### Affected versions

all

All

sorvell:
Fixes #3833.

ContextRoot now stores the root target (`event.composedPath()[0]`) to avoid issues with event re-targeting.

### DIFF
--- a/.changeset/mean-elephants-cough.md
+++ b/.changeset/mean-elephants-cough.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/context': patch
+---
+
+Fixes ContextRoot for requesters in nested shadowRoots, #3833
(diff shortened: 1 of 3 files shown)


justinfagnani (approved):

kevinpschaaf (approved):
Lgtm

PR #3835, opened 2023-04-23
[labs/virtualizer] Update README.md

usergenic:
Changed the old `scrollElementIntoView()` in the README to use `element(x).scrollIntoView()`

### DIFF
--- a/.changeset/silly-gorillas-arrive.md
+++ b/.changeset/silly-gorillas-arrive.md
@@ -1,0 +1,4 @@
+---
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/silly-gorillas-arrive.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+Updated an item in the README.md - No patch level update needed.
```
justinfagnani, line 4: If you want the new readme published to npm you'll need a new version.

justinfagnani (approved):

PR #3766, opened 2023-03-28
Fix styleMap initial render of mixed-case custom props

justinfagnani:
Fixes https://github.com/lit/lit/issues/3738

PR #3837, opened 2023-04-25
[infra] Fix IE11 test for `styleMap`

augustjk:
The boolean for `supportsCSSVariables` introduced in https://github.com/lit/lit/pull/3766 was flipped so it was actually trying to test css properties and failing like: https://github.com/lit/lit/actions/runs/4791482116/jobs/8522453311


packages/lit-html/src/test/directives/style-map_test.ts
```diff
@@ -12,9 +12,9 @@ import {assert} from '@esm-bundle/chai';
 const ua = window.navigator.userAgent;
 const isChrome41 = ua.indexOf('Chrome/41') > 0;
 const isIE = ua.indexOf('Trident/') > 0;
-const supportsCSSVariables = isIE || isChrome41;
+const supportsCSSVariables = !isIE && !isChrome41;
 const testIfSupportsCSSVariables = (test: any) =>
-  supportsCSSVariables ? test.skip : test;
+  supportsCSSVariables ? test : test.skip;
```
justinfagnani, line 17: This seems like it shouldn't be a change because you're inverting the condition _and_ branches. What am I missing here?

Though I agree this is more correct and oops on that condition!!

augustjk: Originally this used to be
```ts
const testIfSupportsCSSVariables = (test: any) =>
  isIE || isChrome41 ? test.skip : test;
```

Then we got
```ts
const supportsCSSVariables = isIE || isChrome41; // this is actually NOT supports

const testIfSupportsCSSVariables = (test: any) =>
  supportsCSSVariables ? test.skip : test; // we actually want to not skip if supports
```
which semantically was wrong, but ended up skipping the right tests. This particular branching does not change in this PR but does make it semantically correct.

The problematic lines ended up being here:
https://github.com/lit/lit/blob/8a11f7ea0964c8d8f055e1a085e3f5b99877dccd/packages/lit-html/src/test/directives/style-map_test.ts#L71-L74
where those assertions were actually not skipping for IE, so the change in this PR makes sure these assertions are skipped.

justinfagnani: ah... thanks for the explanation!

justinfagnani (approved):

justinfagnani (approved):

PR #3838, opened 2023-04-25
Prepare 3.0 prerelease versions

justinfagnani:
The first commit sets the prerelease tag to "pre" so that the versions will be suffixed "pre-0". The second commit is the result of running `npx changeset version`.


justinfagnani:
hmm... npm is failing. Locally too.

augustjk (approved):
I suppose the goldens for wrapper tests would have to be changed again when we do the non-pre release. Not ideal but I can't think of a better solution atm.

rictic (approved):

augustjk:
Closing as this was done with https://github.com/lit/lit/pull/3853

PR #3844, opened 2023-04-26
Clarify that declarative shadow DOM isn't just in Chrome anymore

koddsson:
There was a missing `in [Chrome](...` which I was going to fix but then figured I'd update the language here since most modern browsers are now supporting the Declarative Shadow DOM. I linked to the browser support matrix on MDN, but maybe caniuse.com would be better? Or just not having a link at all is even better?

Let me know what you think :)


justinfagnani (approved):
Thanks!

AndrewJakubowicz:
Thank you!

PR #3847, opened 2023-04-26
Add missing await in task.ts

rictic:
Flagged by a linter, was leaving this off intentional?

### DIFF
--- a/.changeset/friendly-grapes-melt.md
+++ b/.changeset/friendly-grapes-melt.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': patch
+---
+
+Add a missing await in performTask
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz (approved):
It doesn't look like this would change any behavior - as `performTask` is called in `hostUpdated` without awaiting.

justinfagnani:
`performTask()` is public, so this matters. We just don't have any test coverage of calling `performTask()` directly.

@rictic it'd be nice to get a test for this. I think one that calls `await performTask()` and checks that some state was updated in the task function on the next line should suffice.

rictic:
performTask is protected, and only called by hostUpdated, and hostUpdated does not return the promise ,so this would be difficult to observe

AndrewJakubowicz:
This could be a breaking change for any user extending `performTask` and awaiting on `performTask` - although it seems like better behavior to make it possible to `await` the run task.

Agree about adding a test.

PR #3849, opened 2023-04-26
[labs/ssr] Add "module" to condition name to resolve for Module Loader

augustjk:
This was causing errors in lit.dev repo running eleventy plugin in VM mode.
image

This was when trying to load `tslib` which has these package exports
```json
    "exports": {
        ".": {
            "module": "./tslib.es6.js",
            "import": "./modules/index.js",
            "default": "./tslib.js"
        },
```
and we actually want the "module" export, as the "import" one is causing the error above.

### DIFF
--- a/packages/labs/ssr/src/test/lib/module-loader_test.ts
+++ b/packages/labs/ssr/src/test/lib/module-loader_test.ts
@@ -80,4 +80,20 @@ test('resolves a root exported path (.)', async () => {
   assert.ok(loader.cache.has(isServerPath));
 });
 
+test('prefers "module" condition over "import"', async () => {
+  const loader = new ModuleLoader();
+  const result = await loader.importModule(
+    './module-package-import.js',
+    testIndex
+  );
+  const {module, path: modulePath} = result;
+  assert.is(module.namespace.packageValue, 'module');
+  assert.ok(loader.cache.has(modulePath));
+  const packagePath = path.resolve(
+    path.dirname(testIndex),
+    '../../../../test-projects/test-module-package/index.module.js'
+  );
+  assert.ok(loader.cache.has(packagePath));
+});
+
 test.run();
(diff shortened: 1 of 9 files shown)


AndrewJakubowicz (commented):
Is there a way to test this change?

augustjk:
> Brilliant investigation & elegant fix! Is there a way to test this change?

Aha, yes! I have manually went to `node_modules/@lit-labs/ssr/lib/module-loader.js` in the lit.dev repo, updated this line, and confirmed the dev command to run successfully.

AndrewJakubowicz:
Nice! Is it possible to add a unit test as well?

augustjk:
@AndrewJakubowicz thanks for keeping me honest, test added!

packages/labs/ssr/src/test/lib/module-loader_test.ts
```diff
@@ -80,4 +80,20 @@ test('resolves a root exported path (.)', async () => {
...
 });
 
+test('prefers "module" condition over "import"', async () => {
+  const loader = new ModuleLoader();
+  const result = await loader.importModule(
+    './module-package-import.js',
+    testIndex
+  );
+  const {module, path: modulePath} = result;
+  assert.is(module.namespace.packageValue, 'module');
```
AndrewJakubowicz, line 90: This is awesome!

AndrewJakubowicz (approved):

justinfagnani:
Looks like this is caused by https://github.com/microsoft/tslib/issues/161 which appears to be stalled.

justinfagnani (approved):

augustjk:
> Looks like this is caused by [microsoft/tslib#161](https://github.com/microsoft/tslib/issues/161) which appears to be stalled.

@justinfagnani I agree this is non-standard on their part but it seems other tools are including "module" in their default conditions too e.g.
https://vitejs.dev/config/shared-options.html#resolve-mainfields
https://webpack.js.org/guides/package-exports/#conditions
https://github.com/rollup/plugins/tree/master/packages/node-resolve/#exportconditions

PR #3855, opened 2023-04-27
Initial commit of eslint-plugin-lit

justinfagnani:
A start on https://github.com/lit/rfcs/pull/18

This is a clean package with our standard/minimal build and test setup. From here we can follow either the import existing code, or start with a clean typescript-eslint + lit-labs/analyzer base, depending how we decide in the RFC.


packages/labs/eslint-plugin/src/test/index_test.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {suite} from 'uvu';
+// eslint-disable-next-line import/extensions
+import * as assert from 'uvu/assert';
```
43081j, line 9: One for another time - this has always bugged me. The poor package exports uvu has. So you can't import the js file. One of its many problems. 
There's now an esm version of chai (prerelease). So maybe we move the monorepo to that in future?

43081j (approved):
Makes sense to me. Good to see wireit in there too :D

kevinpschaaf (approved):
LGTM

packages/labs/eslint-plugin/src/test/index_test.ts
43081j: Possibly yeah! Since this one is a node only project, that could work well. I've used it a bunch before so could set that up

rictic (approved):
LGTM, though there's an issue with package.json / package lock that's failing CI

PR #3857, opened 2023-04-27
[3.0] Prep for prerelease

augustjk:
- Bring in latest from main
- Reset initial versions in changeset prerelease mode
- Update `ts-clone-node` 1.0.0->2.0.4. It still technically doesn't support TS5 but [release notes](https://github.com/wessberg/ts-clone-node/releases) say it'll support TS4.9 so it's closer. Not sure if this is necessary for whatever use we have of the package but feels better to update than not.
  - Had to force install for update above due to semver requirements of `ts-clone-node` and in the process the package-lock file got converted to v3 which we would've needed to do anyway before pre-release. v3 lockfile is backwards compatible to npm7 which works with our node16 workflows.


augustjk:
CLA workflow is complaining about lit-robot which I think is safe to ignore.

PR #3861, opened 2023-04-29
[labs/ssr-react] Handle children passed in via props

### LINKED ISSUE #3823: [labs/ssr-react] Handling of children in props

### Description

When monkey-patching `React.createElement` we assume proper usage of the function such that `children` are presented as extra argument(s) after the prop, rather than as a property within the props object. https://react.dev/reference/react/createElement#createelement

The original `createElement` behavior is to use `props.children` but override it with the extra arguments if present.
https://github.com/facebook/react/blob/v17.0.2/packages/react/src/ReactElement.js#L386

So in effect, providing `props.children` is actually valid unless there are extra arguments after props. Since during SSR we always create the extra arg for the template shadowroot, we end up swallowing any children prop that's provided.

This has been happening for any `@lit-labs/react` wrapped package, or `styled-components` usage, as these pass through children as `props.children`.

### Reproduction

See repro and description: https://github.com/lit/lit/issues/3418#issuecomment-1515939084

### Workaround

We'll need to check for the presence of props.children and carry it over.

### Is this a regression?

### Affected versions

0.1.1

N/A

augustjk:
Fixes #3823 

The runtime jsx mode were never affected by this as they always get children passed in via the props object.

### DIFF
--- a/packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
+++ b/packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
@@ -23,9 +23,11 @@ export function wrapCreateElement(
   // non-React alternatives like preact?
   originalCreateElement: typeof ReactCreateElement
 ) {
-  return function createElement<P extends {}>(
+  return function createElement<
+    P extends {children?: ReactNode | Array<ReactNode>}
+  >(
     type: ElementType<P>,
-    props: P,
+    props: P | null,
     ...children: ReactNode[]
   ): ReactElement {
     if (isCustomElement(type)) {
@@ -40,11 +42,20 @@ export function wrapCreateElement(
           },
         });
 
+        const newChildren: ReactNode[] = [templateShadowRoot];
+        if (props?.children !== undefined) {
+          if (Array.isArray(props.children)) {
+            newChildren.push(...props.children);
+          } else {
+            newChildren.push(props.children);
+          }
+        }
+        newChildren.push(...children);
+
         return originalCreateElement(
           type,
           {...props, ...elementAttributes},
-          templateShadowRoot,
-          ...children
+          ...newChildren
         );
       }
     }
--- a/packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
+++ b/packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
@@ -42,14 +42,15 @@ export function wrapCreateElement(
         });
 
         const newChildren: ReactNode[] = [templateShadowRoot];
-        if (props?.children !== undefined) {
-          if (Array.isArray(props.children)) {
-            newChildren.push(...props.children);
-          } else {
-            newChildren.push(props.children);
-          }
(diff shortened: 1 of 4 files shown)


packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
```diff
@@ -23,9 +23,11 @@ export function wrapCreateElement(
   // non-React alternatives like preact?
   originalCreateElement: typeof ReactCreateElement
 ) {
-  return function createElement<P extends {}>(
+  return function createElement<
+    P extends {children?: ReactNode | Array<ReactNode>}
```
justinfagnani, line 27: Isn't there a React type for props with children?

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
```diff
@@ -40,11 +42,20 @@ export function wrapCreateElement(
...
 
+        const newChildren: ReactNode[] = [templateShadowRoot];
+        if (props?.children !== undefined) {
+          if (Array.isArray(props.children)) {
+            newChildren.push(...props.children);
+          } else {
+            newChildren.push(props.children);
+          }
+        }
+        newChildren.push(...children);
```
justinfagnani, line 53: Did you say that the native behavior was do override props.children with the children argument if it was present? Should we emulate that?

augustjk: My assumption was that tooling that generates `React.createElement()` calls should always do one or the other so it felt harmless for our code to "combine" them but I suppose emulating React makes sense.

justinfagnani (approved):

raoufswe:
thank youu @augustjk for working on this!

I see you already have released a new version of ssr-react to `1.1.2`, but the next wrapper is still pointing to an older version of it. do you reckon we need to update the next wrapper as well or i am doing something wrong?

I explicitly sat my ssr-react version to 1.1.2 via
```
  "resolutions": {
    "@lit-labs/ssr-react": "0.1.2"
  }
```
and the error seems to be resolved 🥳

augustjk:
@raoufswe because of the semver range a fresh install would normally install the latest version of `@lit-labs/ssr-react` even though `@lit-labs/nextjs` doesn't specify it. existing projects would have to do a manual update for it. for npm, doing `npm update @lit-labs/ssr-react` works even if it is not listed directly in package.json. i don't use yarn but i'm hoping `yarn upgrade` works similarly? there's also the approach of deleting the lockfile and re-installing to grab the latest semver acceptable version of all your dependencies.

`@lit-labs/ssr-react` being a nested dependency, there might be a risk of outdated version checks not picking up the new version. our default release approach is to only bump version when a dependency gets a minor version bump. if we do one for every patch, it could cause a lot of new versions. perhaps that's not such a bad thing but it's how we have it for now.

raoufswe:
I deleted my lockfile and re-run yarn and it picked up the latest version indeed. thank youu <3

PR #3866, opened 2023-05-03
[labs/analyzer] Collect up diagnostics rather than throwing on unexpected (legal) conditions 

rictic:


packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -32,6 +33,7 @@ export class Analyzer implements AnalyzerInterface {
   readonly fs: AnalyzerInterface['fs'];
   readonly path: AnalyzerInterface['path'];
   private _commandLine: ts.ParsedCommandLine | undefined = undefined;
+  readonly diagnostics: ts.Diagnostic[] = [];
```
justinfagnani, line 36: I this being written to anywhere?

packages/labs/analyzer/src/lib/diagnostic-code.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export const DiagnosticCode = {
+  /**
+   * This is the default error code and doesn't indicate any specific error.
+   */
+  UNKNOWN: 548000, // The letters L-I-T are on digits 5-4-8 on a phone keypad.
```
justinfagnani, line 11: Why do we need such large codes?

packages/labs/analyzer/src/lib/errors.ts
```diff
@@ -5,27 +5,52 @@
  */
 
 import ts from 'typescript';
+import {DiagnosticCode} from './diagnostic-code.js';
+import {EOL} from 'os';
```
justinfagnani, line 9: I'd like to think forward to running the analyzer in non-Node environments... how does depending on Node packages here effect that?

And does TypeScript have a base/default FormatDiagnosticsHost that we can use?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -95,6 +96,20 @@ export const getClassMembers = (
         })
       );
     } else if (ts.isPropertyDeclaration(node)) {
+      if (!ts.isIdentifier(node.name)) {
+        analyzer.diagnostics.push(
```
justinfagnani, line 100: This could have a lot of false-positives, for well-known symbols, private fields, or overrides of protected methods (ie, you don't want to see a warning for a method you _had_ to name this way). Could we check for those cases and for the rest have a separate code so it's more easily disabled?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -135,7 +150,7 @@ const getClassDeclarationName = (declaration: ts.ClassDeclaration) => {
   if (name === undefined) {
     throw new DiagnosticsError(
       declaration,
-      'Unexpected class declaration without a name'
+      'Internal error: expected class declaration to either have a name or be a default export'
```
justinfagnani, line 153: I don't think an error should both be a diagnostic and internal. If this is a diagnostic, can we handle it instead of throwing? If we throw, where is this caught and recovered from?

packages/labs/analyzer/src/lib/javascript/functions.ts
```diff
@@ -37,7 +37,7 @@ const getFunctionDeclarationName = (declaration: ts.FunctionDeclaration) => {
   if (name === undefined) {
     throw new DiagnosticsError(
       declaration,
-      'Unexpected function declaration without a name'
+      'Internal Error: expected every function declaration to either have a name or be a default export'
```
justinfagnani, line 40: Ditto diagnostic error vs internal error.

packages/labs/analyzer/src/lib/javascript/functions.ts
```diff
@@ -156,10 +157,13 @@ const getReturn = (
     .getTypeChecker()
     .getSignatureFromDeclaration(node);
   if (signature === undefined) {
-    throw new DiagnosticsError(
-      node,
-      `Could not get signature to determine return type`
+    analyzer.diagnostics.push(
+      createDiagnostic({
+        node,
+        message: `Could not get signature to determine return type`,
```
justinfagnani, line 163: If you know: Is this one actionable in any way? It'd be nice to comment here when this is possible if semi-normal source can produce it, or comment that it's unexpected. Maybe TODO?

packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -135,7 +140,7 @@ export const getModule = (
         if (moduleSpecifier === undefined) {
           throw new DiagnosticsError(
             statement,
-            `Expected a wildcard export to have a module specifier.`
+            `Internal Error: expected a wildcard export to always have a module specifier.`
```
justinfagnani, line 143: Diagnostic vs internal.

If we have these cases where we don't think valid source should be reach here, then I guess we'd expect a semantic error to already have been produced by TypeScript? Can we just assume that and carry on, or do we really want to halt execution?

packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -312,18 +322,22 @@ export const getPathForModuleSpecifier = (
...
   ).resolvedModule?.resolvedFileName;
   if (resolvedPath === undefined) {
-    throw new DiagnosticsError(
-      location,
-      `Could not resolve specifier ${specifier} to filesystem path.`
+    analyzer.diagnostics.push(
+      createDiagnostic({
+        node: location,
+        message: `Could not resolve specifier ${specifier} to filesystem path.`,
+        category: ts.DiagnosticCategory.Error,
```
justinfagnani, line 337: I wonder if this is always an error... I feel like a lot of systems play fast and loose with specifiers and some non-standard ones will trigger this often. What happens with a specifier that's covered by a `declare module` statement too? Does it resolve to that declaration?

packages/labs/analyzer/src/lib/javascript/variables.ts
```diff
@@ -134,10 +136,15 @@ const getVariableDeclarationInfoList = (
       )
       .flat();
   } else {
-    throw new DiagnosticsError(
-      dec,
-      `Expected declaration name to either be an Identifier or a BindingPattern`
+    analyzer.diagnostics.push(
+      createDiagnostic({
+        node: dec,
+        message: `Expected declaration name to either be an Identifier or a BindingPattern`,
```
justinfagnani, line 142: What's a normal-person for a binding pattern? Is that an destructuring assignment statement?

packages/labs/analyzer/src/lib/lit-element/properties.ts
```diff
@@ -34,7 +35,18 @@ export const getProperties = (
 
   for (const prop of propertyDeclarations) {
     if (!ts.isIdentifier(prop.name)) {
-      throw new DiagnosticsError(prop, 'Unsupported property name');
+      analyzer.diagnostics.push(
+        createDiagnostic({
+          node: prop,
+          message:
+            '@lit-labs/analyzer only supports analyzing reactive properties of ' +
+            '`LitElement` extension classes named with plain identifiers. This ' +
```
justinfagnani, line 43:
```suggestion
            '`LitElement` subclasses named with plain identifiers. This ' +
```

Also, it's a little ambiguous if the named things are the properties or the classes, but I couldn't think of a nice way to rephrase. No big deal.

packages/labs/analyzer/src/lib/references.ts
```diff
@@ -217,26 +224,30 @@ export const getImportReference = (
       // External import: extract the npm package (taking care to respect
       // npm orgs) and module specifier (if any)
       const info = specifier.match(npmModule);
-      if (!info || !info.groups) {
-        throw new DiagnosticsError(
-          location,
-          `External npm package could not be parsed from module specifier '${specifier}'.`
+      if (info && info.groups) {
```
justinfagnani, line 227:
```suggestion
      if (info?.groups) {
```

packages/labs/analyzer/src/lib/types.ts
```diff
@@ -34,10 +34,14 @@ export const getTypeForTypeString = (
   if (typeString !== undefined) {
     const typeNode = parseType(typeString);
     if (typeNode == undefined) {
-      throw new DiagnosticsError(
-        location,
-        `Internal error: failed to parse type from JSDoc comment.`
+      analyzer.diagnostics.push(
+        createDiagnostic({
+          node: location,
+          message: `Internal error: failed to parse type from JSDoc comment.`,
```
justinfagnani, line 40: Remove "Internal error" here?

packages/labs/analyzer/src/lib/types.ts
```diff
@@ -85,16 +89,24 @@ export const getTypeForType = (
...
   );
+  let getReferences;
   if (typeNode === undefined) {
-    throw new DiagnosticsError(
-      location,
-      `Internal error: could not convert type to type node`
+    analyzer.diagnostics.push(
+      createDiagnostic({
+        node: location,
+        message: `Internal error: could not convert type to type node`,
```
justinfagnani, line 97: Remove "Internal error"?

packages/labs/analyzer/src/test/lit-element/properties_test.ts
```diff
@@ -268,5 +269,26 @@ for (const lang of languages) {
     }
   });
 
+  test('property with an unsupported name type', ({analyzer, element}) => {
+    // This currently results in two diagnostics: one for the
+    // `LitElement`-specific part of the analysis and one for the vanilla class
+    // analysis.
```
justinfagnani, line 275: I was wondering about this... can we make one take precedence and squash the other?

justinfagnani (commented):
Nice! A few comments, but none a big deal.

packages/labs/analyzer/src/lib/diagnostic-code.ts
justinfagnani: I didn't realize these numbers coexisted with others... maybe add a comment?

justinfagnani (approved):

justinfagnani:
@rictic this looks ready to go, but the conflict has to be resolved. It came from https://github.com/lit/lit/pull/3812 which change some error cases for parsing CSS variables.

PR #3874, opened 2023-05-04
Virtualizer: Fix #3873 (visibilityChanged regression)

### LINKED ISSUE #3873: [labs/virtualizer] visibilityChange event not fired if other state (e.g. range) hasn't also changed

### Description

Virtualizer fails to fire the `visibilityChanged` event if the set of visible items has changed but other state (e.g. range, size) hasn't also changed.

### Reproduction

1. Load this Playground example:

https://lit.dev/playground/#gist=d52dd849778ded1bfb0aad080edfc1cf

2. Scroll down some distance.

The numbers in the upper right corner (indicating the first and last visible indices) will increase.

3. Scroll back up to the top.

Expected behavior: The numbers return to their original values, including `0` for the first visible index
Actual behavior: The numbers stop updating before returning all the way to their original values

### Workaround

No reasonable workaround; you'd have to do some kind of expensive inspection from the outside, or monkey-patch Virtualizer.

### Is this a regression?

### Affected versions

Failing in 2.0.x, worked in 1.0.1

Not environment-specific

graynorton:


usergenic:
Interesting. I guess the visibility change tests we have right now only test the case where item sizes are adjusted to make things become visible/invisible.  Will keep in mind when reworking tests, to make sure the tests cover more of the expected triggers...

LGTM

usergenic (approved):

PR #3880, opened 2023-05-05
fix(labs/gen-utils): js highlight extension link

tanishqmanuja:


justinfagnani:
Can you explain why this change should be made?

tanishqmanuja:
Yeah, there are two tagged functions in question here, namely html`` and javascript`` both are just aliases to the concat. The main reason for these aliases is to provide syntax highlighting. 

But the comment above the javascript function is to the html syntax highlighting vscode extension. Which would highlight only these:

HTML (incl. html quoted and unquoted attrs)
SQL
XML
SVG
CSS
GLSL

And obviously not JAVASCRIPT

This is just a quality of life change, nothing about the actual code.

PS:
Also I hope the team would show some love to the angular component wrapper generator, like react and vue. That is why i was looking and testing this code in the first place, because i wanted to wrap the new material design 3 lit components.

rictic (approved):
Ah, good catch, thanks!

PR #3885, opened 2023-05-09
[labs/react & labs/ssr-react] Coordinate props for server rendering and hydration

### LINKED ISSUE #3418: [labs/react] Prop/attribute handling for server-rendered wrapped components

React (@lit-labs/react)

### Description

With https://github.com/lit/lit/pull/3128 merged, wrapped components can be (shallowly) server-rendered by React. However, all props found in the custom element class's prototype are collected and not passed to React, meaning those will not show up in the server-rendered markup of the host element.

This can cause janky behavior like the initial paint rendering a not-upgraded custom element, then the upgrade happens with no attributes set rendering with default props, then the `componentDidMount` lifecycle sets the props causing another potential update.

### Reproduction

TODO

### Workaround

Potential ways of handling as suggested by @sorvell:

1. Could just pass through all the props when `isServer` is true. We'd get poor behavior for complex props and corner cases but primitive values would work.
2. Do (1) but apply special handling for known element properties: (1) always pass primitive values, (2) always just JSON.serialize objects, (3) pass the value of [pseudo-code] this.getPropertyOptions(propName).toAttribute(reactPropValue)

### Is this a regression?

### Affected versions

1.1.0

N/A

augustjk:
Fixes #3418 

Recommend reviewing commit by commit.

- `labs/react` now has a Node build. In Node mode, it'll check to see if `React.createElement` has been patched by checking its `.name === 'litPatchedCreateElement'`, and if it has element properties that needs to be set, passes them in a special prop bag `_$litProps$` which is a property that should not be minified.
- `labs/ssr-react` during SSR will now set all React props it gets as attributes (breaking change), and only set as properties those that were provided via above special bag. It also adds `defer-hydration` attribute to host element if any properties were set this way.
- `labs/react` wrapper on the client will remove `defer-hydration` attribute after setting props on the element in the wrapper component's `componentDidMount` lifecycle.


packages/labs/react/src/create-component.ts
```diff
@@ -354,6 +356,20 @@ the element.`);
 
         props[k] = v;
       }
+
+      // If component is to be server rendered with `@lit-labs/ssr-react` pass
+      // element properties in a special bag to be set by the server-side
+      // element renderer.
+      if (
+        NODE_MODE &&
+        createElement.name === 'litPatchedCreateElement' &&
```
AndrewJakubowicz, line 366: Could `litPatchedCreateElement` become mangled or minified and break this check?

packages/labs/ssr-react/src/lib/node/render-custom-element.ts
```diff
@@ -37,6 +37,10 @@ export const renderCustomElement = (tagName: string, props: {} | null) => {
     deferHydration: false,
   };
 
+  // elementAttributes will be provided to React as props for the host element
+  // for properly rendering reflected attributes
+  const elementAttributes: {[index: string]: string} = {};
```
AndrewJakubowicz, line 42: Ultra nit - can ignore. What is `index`? Would the name `attributeName` make sense?

packages/labs/ssr-react/src/test/integration/enable-lit-ssr_test.tsx
```diff
@@ -210,4 +214,67 @@ test('nested element', () => {
...
+wrappedCEtest('wrapped element with prop and attribute', () => {
+  assert.equal(
+    ReactDOMServer.renderToString(
+      <ReactTestElement name="React" id="react-test-element" />
+    ),
+    `<test-element id="react-test-element" defer-hydration=""><template shadowroot="open" shadowrootmode="open"><style>
+    p {
+      color: blue;
+    }
+  </style><!--lit-part aHUgh01By8I=--><p>Hello, <!--lit-part-->React<!--/lit-part-->!</p>
```
AndrewJakubowicz, line 260: Are the lit part keys stable? In this case `aHUgh01By8I`? Can these be stripped from test using some equivalent of the `stripExpressionMarkers` helpes?

packages/labs/ssr-react/src/test/object-test-element.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'object-test-element': ObjectTestElement;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace JSX {
```
AndrewJakubowicz, line 26: No action - just curious. What is this for? Is this to add a type for the jsx element?

AndrewJakubowicz (approved):
LGTM from a non domain understanding perspective. I learnt a lot reading through this! An amazingly elegant solution to a thorny issue!

packages/labs/react/src/create-component.ts
augustjk: Since the `labs/ssr-react` package is not minified by us, and is meant to be used server-side, I think in practice this is unlikely to be mangled.

There is another potential failure point of this detection that users might wrap `createElement` with another enhancement on top, which we technically allow but we could add a note to make sure ours is the last wrapper. Another consideration would be to set a global flag rather than doing a function name check like this. My feeling is this is a premature concern but I'm open to ideas.

packages/labs/ssr-react/src/test/integration/enable-lit-ssr_test.tsx
augustjk: That part is a digest built from the static portion of the template so it's stable as long as the template doesn't change. These digests are used client side during hydration to detect any changes. I don't feel the need to strip them as I consider these more snapshot tests of the SSRed result. I do think there's room to improve the test infrastructure to more easily generate/update these assertions but haven't gotten around to that. :/

packages/labs/ssr-react/src/test/object-test-element.ts
augustjk: Yes, it allows usage of the bare lower case custom element tag in JSX without TS errors. Normally only built in HTML elements are allowed to be lower case.

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
+# Server Rendering Lit Components in React
+
+## Background
+
+React server rendering of custom elements will only shallow render components by emitting the host element tag with props stringified to attributes and any light DOM children it has. `@lit-labs/ssr` is able to server render Lit elements deeply by emitting the contents of the component as Declarative Shadow DOM. We want usage of custom elements in React with server rendering to be able to take advantage of this as well.
+
+Client rendering of custom elements in React have their shortfalls such as all props being set as attributes on the element and adding event listeners being clunky. We have the `createComponent` in `@lit-labs/react` that creates a React Component that wraps the custom element which helps with this situation. This package could help bridge the gap in server render as well.
```
AndrewJakubowicz, line 7: Nit: Do we need the sentence: "This package _could_ help bridge the gap [...]"?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+Here we simply rely on React's built in SSR which will render the custom element tag, setting all props as attributes. There's no content until JS with component definition is loaded in the client and the custom element is upgraded.
+
+There is potential for FOUC when the component has light DOM children as these won't have a slot to project to and show up unstyled in the initial document before upgrade, and content shifts as upgraded components fill up spaces.
+
+#### What we want
+
+To mirror the client-side React behavior, all props will be set as attributes on the element as it is server rendered with `@lit-labs/ssr`.
+
+This will work fine for simple elements whose properties are all serializable and the attribute name matches the property name. More complex components may fail to render.
```
AndrewJakubowicz, line 34: What does fail to render mean? Is it behavior worse than the vanilla React behavior? Or the same?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Bare custom elements
+
+#### With vanilla React
+
+Here we simply rely on React's built in SSR which will render the custom element tag, setting all props as attributes. There's no content until JS with component definition is loaded in the client and the custom element is upgraded.
+
+There is potential for FOUC when the component has light DOM children as these won't have a slot to project to and show up unstyled in the initial document before upgrade, and content shifts as upgraded components fill up spaces.
+
+#### What we want
```
AndrewJakubowicz, line 30: Is this what we want our `react ssr` package to enhance? "What we want with `labs/ssr-react`?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
+# Server Rendering Lit Components in React
+
```
AndrewJakubowicz, line 2: Is it worth dating this document?

dev-docs/design/ssr-react.md
AndrewJakubowicz: Worth checking in a test to show what happens in these cases?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+The experience here is mostly the same as bare custom elements. No deep server rendering happens. Only the host element tag is emitted. The wrapper prevents props that exist on the component as properties from being emitted as attributes. Component renders client side when definition is loaded.
+
+The situation with FOUC and content shift exists just like above.
+
+#### What we want
+
+The `createComponent` wrapper already separates props to those that should be set as properties on the element and those that can be passed to React to be set as attributes. In the server, the wrapper should specifically pass those that need to be set as properties on the element in a special bag so the element renderer can set as properties before rendering the template shadowroot.
+
+On the client, Lit element hydration should be deferred until the wrapper has chance to set properties on the element.
```
AndrewJakubowicz, line 48: Nit: Add "a", "has _a_ chance"

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Patching `createElement`
+
+The classic JSX transform compiles JSX to `React.createElement` calls. Our `@lit-labs/react` wrapper also uses the `React` passed in by user to call `createElement` on it.
+
+We can provided an alternate `createElement` function that applies the core strategy above.
+
+For convenience, we can provide a side-effectful import that monkey patches `React.createElement` to be this alternate version.
+
+Using export conditions, we can create a module `@lit-labs/ssr-react/enable-lit-ssr.js` that when imported in Node will monkey patch `React.createElement` and when imported in the browser will load the `@lit-labs/ssr-client/lit-element-hydrate-support.js`.
```
AndrewJakubowicz, line 68: Nice!

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+The runtime JSX transform compiles JSX to various jsx function calls, importing `jsx` or `jsxs` from `<package>/jsx-runtime` or `jsxDEV` from `<package>/jsx-dev-runtime` depending on the environment and whether the `children` array was created statically.
+
+We can provide alternate jsx functions that enhance the base React jsx functions with the core strategy above.
+
+This is mainly for targeting bare custom elements as wrapped components must have `React.createElement` patched to work.
+
+### `@lit-labs/react` package
+
+The wrapper made with `createComponent` from the Node build of this package will check whether `React.createElement` has been patched by checking its `.name` patches ours. If so, it will pass element properties in a special prop named `_$litProps$` to `createElement` to be used for server rendering as described above.
```
AndrewJakubowicz, line 80: Typo: "its `.name` patches ours" -> "if its `.name` _matches_ ours"

AndrewJakubowicz, line 80: And it will set `defer-hydration`?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+On the client, the wrapper on mount will set properties on the element and remove the `defer-hydration` attribute to start Lit element hydration. It will also add `suppressHydrationWarning` to the prop of the host element to suppress React's warning for having extra attributes on the server rendered HTML like `defer-hydration`.
+
+There was consideration of baking in the whole SSR template shadowroot generation into the Node build of `@lit-labs/react` package itself which would remove the need for monkey patching `createElement`, but users would then have to make sure to manually load the `lit-element-hydrate-support` early anyway. It's not out of the realms of possiblity.
```
AndrewJakubowicz, line 84: "It's not out of the realms of possiblity." - does this mean this could still be an additional enhancement after this change, and is compatible with this change?

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+### `@lit-labs/nextjs` package
+
+This is a plugin for Next.js projects to wrap the config in `next.config.js`, adding custom ebpack config that automatically imports `@lit-labs/ssr-react/enable-lit-ssr.js` and replaces the `react/jsx-runtime` imports with `@lit-labs/ssr-react/jsx-runtime`.
```
AndrewJakubowicz, line 88: "ebpack" -> "webpack"

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,96 @@
...
+
+While users are able to specify a `jsxImportSource` for the runtime JSX, this won't work for any external components that were pre-compiled with runtime JSX, nor does it allow composing multiple runtime jsx functions. Replacing the module at the webpack config level addresses these.
```
AndrewJakubowicz, line 90: This paragraph may need some additional clarification. I'm not understanding how it fits in with this doc. The doc has been very detailed regarding the property/attribute handling, as well as differences between bare custom elements vs labs/React wrapped elements. For me this section is missing background.

Why would I specify a `jsxImportSource`, and what is an external component? What is the behavior in nextjs without the labs package?

Maybe remove from this PR, as it could be its own `design/nextjs.md` doc.

A slight meta point. This document is called `ssr-react.md` which suggests to me it's the canonical design for React ssr. However after reading it may be better to focus it on this complex feature of property & attribute handling with React SSR.

This is a great doc! Thanks for writing it up!

dev-docs/design/ssr-react.md
augustjk: It would largely depend on the component's `render()` and the prop it takes. If it's expecting an object property to always exist (without guards or optional chains), it could throw. These cases would've thrown during client rendering with React as well.

Even with guards, this'd mean the rendered result would not be the same as what it would be if props were set as properties. Again, still the same behavior as React client rendering.

I see the argument for a test showcasing what happens, but I don't think adding tests expecting failures for these is as helpful for maintenance.

dev-docs/design/ssr-react.md
augustjk: I treat is as the overall SSR behavior we want, not limited to just a single package as in the case of the wrapped component we require coordination across the `labs/ssr-react` and `labs/react` packages.

dev-docs/design/ssr-react.md
augustjk: I had considered that but figured the git history would serve that purpose, especially as this doc gets updated over time.

dev-docs/design/ssr-react.md
augustjk: I did want to include everything in our ecosystem related to React here and not have a design doc per labs package kind of deal as they end up working together.

The point about `jsxImportSource` maybe is better explained here: https://github.com/lit/lit/tree/main/packages/labs/ssr-react#using-the-automatic-runtime-jsx-transform

External component refers to components you bring in from outside the project like npm or design system library.

Behavior of Next.js without labs package is as described above as it's just React SSR.

Thank you for the feedback! I'll add more context/clarify these points.

packages/labs/react/src/create-component.ts
AndrewJakubowicz: Discussed offline - this sounds great. And global flag can be used down the line if required.

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Scenarios to Handle
+
+<!-- prettier-ignore-start -->
+| | With vanilla React | What we want |
+| --- | --- | --- |
+| Bare custom elements | <p>Here we simply rely on React's built in SSR which will render the custom element tag, setting all props as attributes. There's no content until JS with component definition is loaded in the client and the custom element is upgraded.</p><p>There is potential for FOUC when the component has light DOM children as these won't have a slot to project to and show up unstyled in the initial document before upgrade, and content shifts as upgraded components fill up spaces.</p> | <p>To mirror the client-side React behavior, all props will be set as attributes on the element as it is server rendered with `@lit-labs/ssr`.<p><p>This will work fine for simple elements whose properties are all serializable and the attribute name matches the property name. More complex components may fail to render correctly.</p> |
```
AndrewJakubowicz, line 14: Nit: Add (Flash of unstyled content) in brackets or something to the first time the FOUC acronym is used. May not be an immediately familiar acronym to the reader.

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,72 @@
...
+## Background
+
+React server rendering of custom elements will only shallow render components by emitting the host element tag with props stringified to attributes and any light DOM children it has. `@lit-labs/ssr` is able to server render Lit elements deeply by emitting the contents of the component as Declarative Shadow DOM. We want usage of custom elements in React with server rendering to be able to take advantage of this as well.
+
+Client rendering of custom elements in React have their shortfalls such as all props being set as attributes on the element and adding event listeners being clunky. We have the `createComponent` in `@lit-labs/react` that creates a React Component that wraps the custom element which helps with this situation.
+
+## Scenarios to Handle
+
+<!-- prettier-ignore-start -->
+| | With vanilla React | What we want |
```
AndrewJakubowicz, line 12: Nit: Left side is "With vanilla React". Could right hand side be "With Lit SSR packages in React" or something more specific.

I found myself wondering - what do we want that is different from "With vanilla React".

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,72 @@
...
+| `@lit-labs/react` wrapped components | <p>The experience here is mostly the same as bare custom elements. No deep server rendering happens. Only the host element tag is emitted. The wrapper prevents props that exist on the component as properties from being emitted as attributes. Component renders client side when definition is loaded.</p><p>The situation with FOUC and content shift exists just like bare custom elements.</p> | <p>The `createComponent` wrapper already separates props to those that should be set as properties on the element and those that can be passed to React to be set as attributes. In the server, the wrapper should specifically pass those that need to be set as properties on the element in a special bag so the element renderer can set as properties before rendering the template shadowroot.</p><p>On the client, Lit element hydration should be deferred until the wrapper has a chance to set properties on the element.</p> |
```
AndrewJakubowicz, line 15: Nit: Worth making following explicit?

From: `the wrapper should specifically pass ~those~ that need to be set as properties`
To: `the wrapper should specifically pass **props** that need to be set as properties`

dev-docs/design/ssr-react.md
```diff
@@ -0,0 +1,72 @@
...
+
+The runtime JSX transform compiles JSX to various jsx function calls, importing `jsx` or `jsxs` from `<package>/jsx-runtime` or `jsxDEV` from `<package>/jsx-dev-runtime` depending on the environment and whether the `children` array was created statically.
+
+We can provide alternate jsx functions that enhance the base React jsx functions with the core strategy above.
+
+This is mainly for targeting bare custom elements as wrapped components must have `React.createElement` patched to work.
+
+### `@lit-labs/react` package
+
+The wrapper made with `createComponent` from the Node build of this package will check whether `React.createElement` has been patched by checking its `.name` matches ours. If so, it will pass element properties in a special prop named `_$litProps$` to `createElement` to be used for server rendering as described above.
```
AndrewJakubowicz, line 46: "by checking its", "by checking **if** its"

packages/labs/react/package.json
```diff
@@ -124,25 +136,38 @@
...
         "build:rollup",
+        "../../reactive-element:build",
         "../../tests:build"
       ],
       "env": {
         "BROWSERS": {
           "external": true
-        }
+        },
+        "MODE": "prod"
```
AndrewJakubowicz, line 157: Nice cleanup!

packages/labs/ssr-react/README.md
```diff
@@ -87,6 +87,12 @@ const enhancedCreateElement = wrapCreateElement(React.createElement);
...
 
+### Handling of Props
+
+For bare custom elements, all props provided by React are set as **attributes** on the element during server rendering, as is the default behavior for custom elements rendered by React on the client (as of version 18). This works for simple components whose properties can easily be represented as attributes, i.e. they are easily serialized/deserialized and the attribute name does not differ with the property name.
+
+For Lit elements wrapped with `@lit-labs/react`'s `createComponent()`, properties present on the element will be set as **properties** instead of **attributes** for server rendering. Client side hydration will also be deferred such that it'll wait for elemenet properties to be set before the first update happens. Note: this is only made available by the [monkey patching](#monkey-patching-reactcreateelement-recommended) approach.
```
AndrewJakubowicz, line 94: `elemenet` -> `element`

AndrewJakubowicz (approved):
Great! No comments on the code portion, and only minor nits in the docs portion.

augustjk:
@AndrewJakubowicz Thank you for the excellent reviews. Eagle eyes, you have!

Issue #3486, opened 2022-12-01
[labs/motion] transform can be calculated incorrectly


sorvell:
### Which package(s) are affected?

Motion (@lit-labs/motion)

### Description

From discord:

The transformProps functions are a bit wrong here: https://github.com/lit/lit/blob/0260a15e2087d2b1ea8f5072663a637cedd186d9/packages/labs/motion/src/animate.ts#L80

`calculateKeyframes() `is checking to see if op.transform is undefined
but transformProps has an expression like:

```
return {value, transform: value && `translateX(${value}px)`};
```
 
and when value is 0, this makes it transform: 0, not transform: undefined
This should probably be:

```
return {value, transform: value === undefined || isNaN(value) ? undefined : `translateX(${value}px)`};
```

### Reproduction

No repro.

### Workaround

None.

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

all

### Browser/OS/Node environment

All

PR #3886, opened 2023-05-09
[labs/motion] Fix transforms when scaling to or from zero

justinfagnani:
Fixes https://github.com/lit/lit/issues/3486

When scaling to zero with FLIP you need to scale to a small number instead so that you can calculate a finite inverse transform. Luckily we can do this by passing a small override value in the first keyframe of the animation. In this fix we scale to 1px and when the animation ends it'll pop to 0.

### DIFF
--- a/packages/labs/motion/src/animate.ts
+++ b/packages/labs/motion/src/animate.ts
@@ -79,22 +79,49 @@ const quotientOp = (a: number, b: number) => {
 };
 
 // Computes a transform given a before and after input for given properties.
-export const transformProps = {
+export const transformProps: {
+  [p: string]: (
+    a: number,
+    b: number
+  ) => {value?: number; transform?: string; override?: {[k: string]: string}};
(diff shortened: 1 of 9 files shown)


justinfagnani:
ping @sorvell

packages/labs/motion/src/animate.ts
```diff
@@ -524,6 +551,9 @@ export class Animate extends AsyncDirective {
           fromFrame['transform'] = `${fromFrame['transform'] ?? ''} ${
             op['transform']
           }`;
+          if (op.override !== undefined) {
+            Object.assign(fromFrame, op.override);
```
rictic, line 555: fromFrame and override should probably have more descriptive names here

`fromFrame` is more like, `transformToIntoFromFrame`, and `override` is specifically overriding _that_

rictic (approved):

PR #3888, opened 2023-05-09
Fix a memory leak with the shared TreeWalker

justinfagnani:
We weren't resetting the currentNode of the shared TreeWalker after either template prep or template cloning. This caused the TreeWalker to hold on to the last currentNode of the last tree walk.

This should always be after a clone because I can't think of a situation where we prep a template but don't immediately render it. So for the fix I set currentNode after a template clone. You can't set `TreeWalker.currentNode` to `null`, so I set it to the document.

### DIFF
--- a/.changeset/few-feet-design.md
+++ b/.changeset/few-feet-design.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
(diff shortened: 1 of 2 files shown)


rictic (approved):

rictic:
Nice!

Issue #3883, opened 2023-05-05
[labs/context] Context decorators error on private or protected fields


justinfagnani:
### Discussed in https://github.com/lit/lit/discussions/3875

<div type='discussions-op-text'>

<sup>Originally posted by **NealRame** May  4, 2023</sup>
In my component I want to add a context properties which is also private
```
@provide({ context: someContext })
@state()
private _someData = 0
```

I read in the documentation that it is possible however when doing so, I get that typescript compiler error :
```
error TS2345: Argument of type 'MyComponent' is not assignable to parameter of type 'ReactiveElement & Record<"_someData", number>'.
  Type 'MyComponent' is not assignable to type 'Record<"_someData", number>'.
    Property '_someData' is private in type 'MyComponent' but not in type 'Record<"_someData", number>'.

89     @provide({ context: someContext })
```

Could someone explain me if I am missing something please?

I am using `lit@2.6.1` and `@lit-labs/context@0.3.1`.
    </div>

justinfagnani:
Well, this is difficult to address. It seems that either we have type-checking on public fields to ensure the context type and field type are compatible, or we allow decorating private fields, but not both.

The constraint comes from the fact that a property decorator only receives arguments and type parameters for the public class interface and the property name, and unfortunately in the case of private and protected fields we get the name but that field isn't present in the public interface.

The new standard decorators fix this because the decorators get the value type as a type parameter. This could theoretically be added to old decorators, but the TypeScript team has said they're not going to invest in the old decorators anymore.

I have figured out [a way to let you override the types that the decorator infers](https://github.com/lit/lit/pull/3889) so that you can prevent a decorator from type-checking on a private field. It's not pretty, but it's a workaround:

```ts
  @provide<string, any>({context: stringContext})
  @state()
  private _aString = 'a';
```

The first parameter must match the context type, and the second parameter must be `any` to override the class's type.

So... we're faced with a decision here between:

1. field type-checking, but no decorating private fields
2. decorating private fields with no DX hit, but no field type-checking
3. field type-checking and decorating private fields with some boilerplate

It seems like option 3 is good, but an argument for dropping field/context type-checking is that many, if not most, context fields should be private or protected so we shouldn't negatively impact private fields.

cc @benjamind

augustjk:
I feel like option 2 is better. Upgrading to standard decorators that wouldn't require this workaround would be easier with that too.

rictic:
Fixed by #3733

PR #3889, opened 2023-05-09
Allow the provide decorator to work on private fields

justinfagnani:
Possible fix for https://github.com/lit/lit/issues/3883

This adds a type parameter to `provide()` that allows overriding the inferred class type so that it can be set to `any` and allow for decorating a private or protected field.

See my comments in https://github.com/lit/lit/issues/3883 comparing this against other options.

### DIFF
--- a/packages/labs/context/src/test/provider-and-consumer_test.ts
+++ b/packages/labs/context/src/test/provider-and-consumer_test.ts
@@ -5,12 +5,13 @@
  */
 
 import {LitElement, html, TemplateResult} from 'lit';
-import {property} from 'lit/decorators/property.js';
+import {property, state} from 'lit/decorators.js';
 
 import {Context, consume, provide} from '@lit-labs/context';
 import {assert} from '@esm-bundle/chai';
 
 const simpleContext = 'simple-context' as Context<'simple-context', number>;
+const stringContext = 'string-context' as Context<'string-context', string>;
 
 class ContextConsumerAndProviderElement extends LitElement {
   @consume({context: simpleContext, subscribe: true})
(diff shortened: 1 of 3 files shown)


packages/labs/context/src/test/provider-and-consumer_test.ts
```diff
@@ -21,7 +22,17 @@ class ContextConsumerAndProviderElement extends LitElement {
   @property({type: Number})
   public value = 0;
 
+  @provide<string, any>({context: stringContext})
```
rictic, line 25: Needing to pass in the `any` type is a big code smell, what about instead an alias of the provide function with looser typing for private fields? We could call it `@providePrivate`

rictic (changes requested):

packages/labs/context/src/test/provider-and-consumer_test.ts
augustjk: I commented on the issue too, but my feeling is just making `provide` looser seems better considering the eventual upgrade to standard decorators that wouldn't need this workaround.

rictic: Oh, standard decorators give the decorator function's type access to the type of the decorated field?

augustjk: It is mentioned here: https://github.com/lit/lit/issues/3883#issuecomment-1541007298

Which does appear to be the case if I'm reading this correctly:
https://github.com/microsoft/TypeScript/pull/50820#:~:text=to%20support%20correctly.-,Type%20Checking,-When%20a%20decorator

rictic:
Superceded by https://github.com/lit/lit/pull/3733

PR #3894, opened 2023-05-11
Add explicit types to tsconfigs in Lit v3 unblocking generated API.

AndrewJakubowicz:
Unblock building the Lit 3 branch in Lit.dev repo for generated API. Issue: https://github.com/lit/lit.dev/issues/1111

If the `types` field is not set, then TypeScript will walk all parent directories collecting all the `node_modules/@types` it encounters. Issue: https://github.com/microsoft/TypeScript/issues/30124

This is a problem when generating the documentation because lit.dev nests the Lit monorepo within the Lit.dev repo. This means all the Lit.dev `@types` get loaded and break with version mismatches etc.

Setting `types` fixes the issue by only loading the global `@types` used by the package.

Tested by building and testing this within the lit.dev repo. Kept fixing tsconfigs until `npm run build` worked.


AndrewJakubowicz:
Should this be a PR against `main`? It only impacts version 3 as far as I can tell (from Lit.dev testing).

AndrewJakubowicz:
Benchmarks failing due to chromedriver issue. Everything else is now green.

AndrewJakubowicz:
@augustjk Found a much smaller and nicer solution: https://github.com/lit/lit.dev/issues/1111#issuecomment-1544695508

PR #3895, opened 2023-05-11
Make _initialize and static _initializers private

justinfagnani:
We had reports of problems if users added their own `_initialize()` method.

My initial change was to make the name `__initialize` and make the method private so at TypeScript would error. But I couldn't find any place that actually overrode `_initialize()`, so I inlined it into the constructor. We should run this through google3 as a smoke test.

I also made `static _initializers` private. It was already being renamed in prod.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -474,10 +474,10 @@ export abstract class ReactiveElement
    */
   static addInitializer(initializer: Initializer) {
     this.finalize();
-    (this._initializers ??= []).push(initializer);
+    (this.__initializers ??= []).push(initializer);
   }
 
-  static _initializers?: Initializer[];
+  private static __initializers?: Initializer[];
 
   /*
    * Due to closure compiler ES6 compilation bugs, @nocollapse is required on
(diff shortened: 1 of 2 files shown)


packages/reactive-element/src/reactive-element.ts
```diff
@@ -893,25 +895,11 @@ export abstract class ReactiveElement
...
-    this.__updatePromise = new Promise<boolean>(
-      (res) => (this.enableUpdating = res)
-    );
-    this._$changedProperties = new Map();
     this.__saveInstanceProperties();
     // ensures first update will be caught by an early access of
     // `updateComplete`
     this.requestUpdate();
-    (this.constructor as typeof ReactiveElement)._initializers?.forEach((i) =>
+    (this.constructor as typeof ReactiveElement).__initializers?.forEach((i) =>
```
sorvell, line 902: I think at one point we wanted to have a protected entry point to override that supported doing stuff before/after the super call (which you can't do in the constructor). 

This change seems fine, but is there a specific reason to make the change? Just YAGNI?

sorvell (approved):
LGTM

packages/reactive-element/src/reactive-element.ts
justinfagnani: Talked offline, and this was prompted by a user accidentally overriding `_initializer`.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -750,8 +750,8 @@ export abstract class ReactiveElement
     // Create own set of initializers for this class if any exist on the
     // superclass and copy them down. Note, for a small perf boost, avoid
     // creating initializers unless needed.
-    if (superCtor._initializers !== undefined) {
-      this._initializers = [...superCtor._initializers];
+    if (superCtor.__initializers !== undefined) {
```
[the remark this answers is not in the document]
rictic: This code is still inside ReactiveElement, and it's referencing a private field defined by ReactiveElement

As long as the code doing the referencing is inside the class body, and the field is a private field defined by the class then it's fine for both TypeScript private and standard #private fields

rictic (approved):

rictic:
This looks good, merging!

PR #3897, opened 2023-05-11
[benchmarks] Bump chromedriver to 113.0.0 to get check passing

usergenic:
The old chromedriver is mismatched with latest chrome version on CI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8193,9 +8193,9 @@
       }
     },
     "node_modules/chromedriver": {
-      "version": "111.0.0",
-      "resolved": "https://registry.npmjs.org/chromedriver/-/chromedriver-111.0.0.tgz",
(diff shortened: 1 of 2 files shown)


.changeset/strange-steaks-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-internal/benchmarks': patch
```
augustjk, line 2: nit: it doesn't matter either way but we've not been consistent about doing changeset for internal/private packages. i generally do empty changeset for these since it's not end-user facing, but it doesn't really hurt to keep them. maybe it's good to have a record of changes even for non-published packages in the changelog file.

augustjk (approved):
thank you!

.changeset/strange-steaks-invent.md
usergenic: I'm happy to switch this to empty.  I wasn't sure about internal packages 🤷 I'm on team augustjk for now since I would still be clicking rerun-failed-jobs if you didn't tell me to bump this LOL.

usergenic: lol i tried to edit in place and lint check wasted time complaining about my markdown formatting so i'll just leave it with the patch note.

PR #3901, opened 2023-05-12
[labs/ssr-client] Bump `lit-html` dependency version

### LINKED ISSUE #3899: [labs/ssr-client] Bump dependency for requisite version of `lit-html`
### Should this be an RFC?

### Description

https://github.com/lit/lit/pull/3720 moved SSR hydration modules to `labs/ssr-client` but that also required changes to `lit-html` for the shared private hydration support objects.

We should add a dependency semver with `lit-html@2.7.1` as minimum so users using `labs/ssr-client` will know if they have an incompatible version of `lit-html`.

Many thanks to @AndrewJakubowicz for finding and suggesting this.

Update `lit-html` manually.

augustjk:
Fixes #3899 

This dependency bump ensures users have the needed version of `lit-html` installed for hydration code to work.

### DIFF
--- a/.changeset/clever-kids-mix.md
+++ b/.changeset/clever-kids-mix.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr-client': patch
+---
+
+Bump minimum version of `lit-html` dependency to 2.7.1 which includes needed shared internals.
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz (approved):
Amazing!

PR #3905, opened 2023-05-16
[labs/ssr] Allow rendering of non-array iterables

### LINKED ISSUE #3122: [labs/ssr] Non-array iterables fail to render

SSR (@lit-labs/ssr)

### Description

When template contains iterables besides an array, for instance like a `Set`, SSR does not actually render the values.

For example:

This
```js
  render() {
    return html`<p>new Set([1, 2, 3])</p>`;
  }
```

results in a paragraph tag containing `[object Set]`.

The same is true for other iterables that normally work client-side like `Map` or `Generator`, which means using the `map()` directive results in `[object Generator]`. 

This also breaks during hydration as there's a mismatch of the number of parts, displaying the following error:
```
Error: unexpected longer than expected iterable
```

### Reproduction

https://stackblitz.com/edit/koa-starter-3x2kbs?file=index.js

### Workaround

Use `Array.from()`

### Is this a regression?

### Affected versions

2.2.1

Node version: v16.15.0

augustjk:
Fixes #3122

### DIFF
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -15,7 +15,11 @@ import type {
 
 import {nothing, noChange} from 'lit';
 import {PartType} from 'lit/directive.js';
-import {isTemplateResult, getDirectiveClass} from 'lit/directive-helpers.js';
+import {
+  isPrimitive,
+  isTemplateResult,
+  getDirectiveClass,
+} from 'lit/directive-helpers.js';
 import {_$LH} from 'lit-html/private-ssr-support.js';
 
 const {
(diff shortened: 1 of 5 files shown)


packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -574,7 +579,8 @@ export function* renderValue(
       value === noChange
     ) {
       // yield nothing
-    } else if (Array.isArray(value)) {
+    } else if (!isPrimitive(value) && isIterable(value)) {
+      // Not primitive since strings are iterable
```
AndrewJakubowicz, line 583:
```suggestion
      // Check that value is not a primitive, since strings are iterable
```

AndrewJakubowicz (approved):
Nice! Great implementation and tests!

I assume a `Map` would also technically work since it is iterable - but just would return k + v for each entry. No action required.

AndrewJakubowicz (approved):

PR #3909, opened 2023-05-18
Removing some crufty code in Virtualizer ScrollerController

chrispaterson:
There are some checks in here that are unnecessary and illogical.  There are also some type assertions that do not need to be there.

### DIFF
--- a/.changeset/calm-insects-complain.md
+++ b/.changeset/calm-insects-complain.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/virtualizer': patch
+---
+
+Refactored and refurbished ScrollerController attach/detach code in effort to reduce potential memory leaks due to held instance references.
--- a/packages/labs/virtualizer/src/ScrollerController.ts
+++ b/packages/labs/virtualizer/src/ScrollerController.ts
@@ -75,7 +75,7 @@ export class ScrollerController extends ScrollerShim {
   private _originalScroll:
     | typeof Element.prototype.scroll
     | typeof window.scroll;
-  private _clients: unknown[] = [];
+  private _clients = new Set<unknown>();
   private _retarget: Nullable<retargetScrollCallback> = null;
   private _end: Nullable<endScrollCallback> = null;
   private __destination: Nullable<ScrollToOptions> = null;
@@ -237,8 +237,13 @@ export class ScrollerController extends ScrollerShim {
   }
 
   public detach(client: unknown) {
-    this._clients = this._clients.splice(this._clients.indexOf(client), 1);
-    if (this._clients.length === 0) {
+    this._clients.delete(client);
+
+    /**
+     * If there aren't any more clients, then return the node's default
+     * scrolling methods
+     */
+    if (this._clients.size === 0) {
       this._node.scrollTo = this._originalScrollTo;
       this._node.scrollBy = this._originalScrollBy;
       this._node.scroll = this._originalScroll;
@@ -248,8 +253,13 @@ export class ScrollerController extends ScrollerShim {
   }
 
   private _attach(client: unknown) {
-    this._clients.push(client);
-    if (this._clients.length === 1) {
+    this._clients.add(client);
+
+    /**
+     * The node should only have the methods shimmed when adding the first
+     * client – otherwise it's redundant
+     */
+    if (this._clients.size === 1) {
       this._node.scrollTo = this.scrollTo;
       this._node.scrollBy = this.scrollBy;
       this._node.scroll = this.scrollTo;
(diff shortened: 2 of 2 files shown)


graynorton:
Thanks for the PR, @chrispaterson!

I'm traveling this week and probably don't have time for a thorough review right now, but I did give it a quick look. Most of these changes look great, but we should tread carefully before removing the logic allowing a single ScrollerController to serve multiple clients. This is intended to support the case where multiple (non-scrolling) virtualizers live within a single scrolling host element. We should add some comments and some tests to ensure this is working as intended, but I don't think we can just remove it.

cc @usergenic

chrispaterson:
Hey Gray, I hope you're traveling for leisure!  Thanks for taking a look :-)

> we should tread carefully before removing the logic allowing a single ScrollerController to serve multiple clients. This is intended to support the case where multiple (non-scrolling) virtualizers live within a single scrolling host element.

It was a little unclear to me what the intent of this code was – I was unaware of the use case.  I added the logic back!  We're investigating memory leaks with elements hanging around as Detached HTMLElements and I found that static reference to element map while digging through code.  It doesn't seem to be used, so I thought I'd clean it up.

usergenic:
Thanks @graynorton I was trying to understand the use case for that logic too.  We don't have nested virtualizer cases in tests yet and we've got some new related issues to nesting, so I'll add some for those.  #3910

usergenic:
And thank you @chrispaterson for helping clean up these potentially leaky corners of the code!

chrispaterson:
@graynorton and @usergenic .  I've just pushed an update because the `detach` method is actually not doing what it's supposed to.  In that method `this._clients = this._clients.splice(this._clients.indexOf(client), 1);` which re-assigns `this._clients` to a new array of only 1 client; the removed client.   I think the intention here is to actually use `slice` which would would remove the `client` from the array of clients.  I think the idea here is that once there are no more clients attached, we can restore the `node`'s original methods.

Instead of just changing the method from `splice` to `slice`, I think a `Set` would be more appropriate here to avoid confusing `Array.prototype` methods for one, but also because I don't think we want a client to remain attached after `detach` is called in the case that the number of `detach` calls doesn't match the number of `attach` calls.  Because using an array means that a client may be added at multiple indices in that array and calling `detach` would only remove the first instance in that array.  This could be a source of leaks.

nawazmobusss:
Happy Friday, @usergenic @graynorton ! Can I kindly ask for an update on this ticket? It would be highly appreciated if this could be addressed soon as we are looking for an important public release in 2 weeks.

usergenic:
> Happy Friday, @usergenic @graynorton ! Can I kindly ask for an update on this ticket? It would be highly appreciated if this could be addressed soon as we are looking for an important public release in 2 weeks.

Looking at now

usergenic:
I was attempting to contrive a regression test to demonstrate any hidden issues whereby the same controller would attach to the same client multiple times, but I can not figure out what that scenario even means in real world terms...

I don't see any logic differences to the way the _originalScroll* properties are assigned back and forth so ultimately this feels pretty safe to me.

Also appreciate the extra comments added.

Lint is failing due to a missing changeset file.  Can you run `npx changeset` and identify this PR as requiring a "patch" level update.  If `npx changeset` has trouble running in your git clone, just create a file like this in the .changeset directory with a name like "sparkle-rocket-trousers.md" and contents like:

```
---
'@lit-labs/virtualizer': patch
---

Refactored and refurbished ScrollerController attach/detach code in effort to reduce potential memory leaks due to held instance references.
```

chrispaterson:
I was able to use `npc changeset` in my fork.  Thanks for reviewing @usergenic !

usergenic (approved):
LGTM will merge after all checks green

nawazmobusss:
Really appreciated, @usergenic !

usergenic:
@chrispaterson and @nawazmobusss thanks for your patience :)

nawazmobusss:
Hope you had a great long weekend, @usergenic ! Do you know when we can expect this fix to be available to use and which version?

usergenic:
https://github.com/lit/lit/pull/3930 is passing all checks and looks ready to go. Hopefull we'll see virtualizer 2.0.3 tomorrow.

nawazmobusss:
Thank you, @usergenic ,  for keeping us informed about the status of the release in a timely manner. Your promptness is greatly appreciated.

nawazmobusss:
Hello @usergenic , just wondering if 2.0.3 version is in already?

usergenic:
The current release PR #3932 is in process of approval and publishing will be soon.  It hasn't published yet but it shouldn't be long.

usergenic:
Apparently #3930 was cancelled and the process was renewed with more recent changes. The upshot is 2.0.3 virtualizer will also include the testing framework help for the ResizeObserver loop limit exceeded errors too.

nawazmobusss:
Wonderful!!

PR #3918, opened 2023-05-24
Some code golf on ReactiveElement

justinfagnani:
* Tear off Object methods to allow them to be renamed (not sure if Terser does this though) and to eliminate a dereference at the use site.
* Use logical assignment and chaining in `requestUpdate()` to collapse three property references to one.
* Remove a local variable in `requestUpdate()`
* Replace `notEqual` implementation with `Object.is`. This is the same behavior _except_ with +0 and -0, which are considered not the same with `Object.is`, but are equal under `===`.

I considered using `Object.hasOwn` since we have a lot of `x.hasOwnProperty()` calls, but `Object.hasOwn` is part of ES2022... it is polyfillable though, so maybe we should consider it?

### DIFF
--- a/.changeset/nasty-deers-search.md
+++ b/.changeset/nasty-deers-search.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Some code golf on ReactiveElement
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -30,6 +30,9 @@ export type {
   ReactiveControllerHost,
 } from './reactive-controller.js';
 
+// TODO (justinfagnani): Add `hasOwn` here when we ship ES2022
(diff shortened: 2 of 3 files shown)


rictic:
What's the delta in binary size?

rictic (approved):

justinfagnani:
@rictic did you rebase? And does the size check action output the previous and new sizes anywhere? It'd be nice to see the delta here. I'll run that locally though.

rictic:
@justinfagnani the delta was printed as a comment but now you can find it as part of the PR, see commit https://github.com/lit/lit/pull/3918/commits/1d39495fa34f15fb5e91888d0bb3a4d4650be784

rictic:
And yeah, I did rebase

justinfagnani:
@rictic could we print the size report always, whether it's positive or negative?

PR #3921, opened 2023-05-24
Add thank you to community authored changes.

AndrewJakubowicz:
Update changelog generation to add a thanks with link to the author of a community authored change. Lit team members are filtered out.

For example when I removed `augustine` from the filter, `npm run version` locally generated the following changelog entry:

### Sample entry with thank you

[#3905](https://github.com/lit/lit/pull/3905) [f8d72859](https://github.com/lit/lit/commit/f8d7285976390bbb35c540eb7b516f3748064d19) Thanks [@augustjk](https://github.com/augustjk)! - Handle non-array iterables including map directive in child parts

### Multiple authors

Tested changesets that include multiple authors with a lit core member.

```
author:testone
author:testtwo
author:AndrewJakubowicz
```

This will render:

```
- [#3901](https://github.com/lit/lit/pull/3901) [`82e9f370`](https://github.com/lit/lit/commit/82e9f3708973709e916ccbb6d8a450110dea7755) Thanks [@testone](https://github.com/testone), [@testtwo](https://github.com/testtwo)! - <Description>
```

### Multiple authors with only lit core

```
author:AndrewJakubowicz
```

Does not render any Thanks.

### DIFF
--- a/.changeset/changelog-lit.cjs
+++ b/.changeset/changelog-lit.cjs
@@ -9,10 +9,6 @@ const {
   getInfoFromPullRequest,
 } = require('@changesets/get-github-info');
 
-// Forked from: https://github.com/atlassian/changesets/blob/main/packages/changelog-github/src/index.ts
-// Remove the "Thanks!" message, as it's almost always self-congratulatory to our team
-// TODO: add back "Thanks!" for external contributors
-
 const repo = 'lit/lit';
 
 const changelogFunctions = {
@@ -99,6 +95,22 @@ const changelogFunctions = {
       };
     })();
 
+    // Only congratulate community contributions.
+    usersFromSummary = usersFromSummary.filter((user) => {
+      return ![
+        'AndrewJakubowicz',
+        'augustjk',
+        'bicknellr',
+        'dfreedm',
+        'e111077',
+        'justinfagnani',
+        'kevinpschaaf',
+        'rictic',
+        'sorvell',
+        'usergenic',
+      ].includes(user);
+    });
+
     const users = usersFromSummary.length
       ? usersFromSummary
           .map(
@@ -111,6 +123,7 @@ const changelogFunctions = {
     const prefix = [
       links.pull === null ? '' : ` ${links.pull}`,
       links.commit === null ? '' : ` ${links.commit}`,
+      users === null ? '' : ` Thanks ${users}!`,
     ].join('');
 
     return `\n\n-${prefix ? `${prefix} -` : ''} ${firstLine}\n${futureLines
--- a/.changeset/changelog-lit.cjs
+++ b/.changeset/changelog-lit.cjs
@@ -9,6 +9,8 @@ const {
   getInfoFromPullRequest,
 } = require('@changesets/get-github-info');
 
+// Forked from: https://github.com/atlassian/changesets/blob/main/packages/changelog-github/src/index.ts
+
 const repo = 'lit/lit';
 
 const changelogFunctions = {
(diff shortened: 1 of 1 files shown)


augustjk (approved):

.changeset/changelog-lit.cjs
```diff
@@ -101,16 +99,41 @@ const changelogFunctions = {
...
+        'augustjk',
+        'bicknellr',
+        'dfreedm',
+        'e111077',
+        'justinfagnani',
+        'kevinpschaaf',
+        'rictic',
+        'sorvell',
+        'usergenic',
+      ].some((coreTeamUser) => s.includes(coreTeamUser));
```
usergenic, line 130:
```ts
return new RegExp(`\\b(${[
        'AndrewJakubowicz',
        'augustjk',
        'bicknellr',
        'dfreedm',
        'e111077',
        'justinfagnani',
        'kevinpschaaf',
        'rictic',
        'sorvell',
        'usergenic',
      ].join('|')})\\b`, 'i').test(s);
```

I went down a rabbit-hole and overthought the problem of usernames that might be supersets of our usernames :-P

LGTM tho.

usergenic (approved):

.changeset/changelog-lit.cjs
augustjk: hahaha that's a great point. impersonators don't deserve thanks!

AndrewJakubowicz: Applied and tested this fantastic solution! Thank you! We definitely want to thank `rictickler` or `chadfreeman` if they contribute!

.changeset/changelog-lit.cjs
```diff
@@ -101,16 +99,44 @@ const changelogFunctions = {
 
     const users = usersFromSummary.length
       ? usersFromSummary
+          .filter((u) => !containsLitTeamMemberUsername(u))
```
justinfagnani, line 102: This removes any mention of the user, yeah? I don't think that's what we want. I think we just want to say "Thanks" only to the external contributors.

justinfagnani (changes requested):

.changeset/changelog-lit.cjs
augustjk: line 139 below has the `users` check that conditionally adds the whole "Thanks" phrase with the users' github handle linked.
```ts
    const prefix = [
      links.pull === null ? '' : ` ${links.pull}`,
      links.commit === null ? '' : ` ${links.commit}`,
      users === null ? '' : ` Thanks ${users}!`,
    ].join('');
```
If user is a lit team member, `users` would be null.

Or are you saying you want to just say "Thanks" and not have the user's github handle linked?

AndrewJakubowicz: Ah this is a case that our current changesets are not covering. In this code path that isn't exercised the result of filtering out team members would result in an empty string. An empty string would fail to match with `null` and would print an empty `Thanks`.

I will make the null handling also check for an empty string & locally test this code path.

AndrewJakubowicz: Fixed by also checking for an empty string & manually tested the cases with multiple `author:<author>` written in a changeset file.

AndrewJakubowicz: Understood! Although I misunderstood you, you caught a real bug that could have given unattributed thanks! 😄 

Current behavior is that all changes ignore users & have no attribution or thanks.

This PR: Adds attribution & thanks to external contributors, but core members continue to receive no thanks or attribution.

Your proposal is to always add attribution, but only add the additional prefixed "Thanks <contributor list>!" if it's an external contributor?
Would you like me to change to always include attribution?

augustjk: Ooh I totally misunderstood that. I'm less sure if we need the attribution for team members since clicking on the PR link right before it would show who worked on it. But no strong feelings either way.

AndrewJakubowicz:
This is good to go pending a review from Justin. And we have some community contributions queued up in the release pipeline!

rictic (approved):

PR #3922, opened 2023-05-25
Replace $` with code block to fix rendering of changelog.

AndrewJakubowicz:
Changeset magically expands the following token into a generated markdown header:

```
$`
```

Which generated the malformed changelog entry:

`_$litProps# Change Log object` in https://github.com/lit/lit/pull/3920#discussion_r1204831647

Fix by using `code` block instead.


augustjk (approved):
Thank you! I was debating fixing that vs just manually updating the changelog file in the versions PR.

PR #3929, opened 2023-05-31
[labs/virtualizer] Add support methods to work around ResizeObserver loop limit exceeded errors

usergenic:
I really struggled to find a coherent way to offer conventional methods to work around the obnoxious "ResizeObserver loop limit exceeeded" class of errors and whether or not to even do this as part of Virtualizer itself.  In the end, I think this provides enough value to warrant its addition to Virtualizer until such time as we might discover a change to Virtualizer itself or find an alternative approach that works just as well.

I factored out what we do in our test helpers into functions exported in support modules so they can be imported by end-users and leveraged in their test and/or live environments.

### DIFF
--- a/packages/labs/virtualizer/package.json
+++ b/packages/labs/virtualizer/package.json
@@ -41,8 +41,8 @@
       "default": "./polyfillLoaders/ResizeObserver.js"
     },
     "./polyfills/resize-observer-polyfill/ResizeObserver.js": {
-      "types": "./polyfillLoaders/ResizeObserver.d.ts",
-      "default": "./polyfillLoaders/ResizeObserver.js"
+      "types": "./polyfills/resize-observer-polyfill/ResizeObserver.d.ts",
+      "default": "./polyfills/resize-observer-polyfill/ResizeObserver.js"
     },
     "./support/method-interception.js": {
       "types": "./support/method-interception.d.ts",
--- a/packages/labs/virtualizer/package.json
+++ b/packages/labs/virtualizer/package.json
@@ -9,52 +9,52 @@
   "type": "module",
   "exports": {
     ".": {
-      "default": "./lit-virtualizer.js",
-      "types": "./lit-virtualizer.d.ts"
+      "types": "./lit-virtualizer.d.ts",
+      "default": "./lit-virtualizer.js"
     },
     "./events.js": {
-      "default": "./events.js",
-      "types": "./events.d.ts"
+      "types": "./events.d.ts",
+      "default": "./events.js"
     },
     "./layouts/flexWrap.js": {
-      "default": "./layouts/flexWrap.js",
-      "types": "./layouts/flexWrap.d.ts"
+      "types": "./layouts/flexWrap.d.ts",
+      "default": "./layouts/flexWrap.js"
     },
     "./layouts/flow.js": {
-      "default": "./layouts/flow.js",
-      "types": "./layouts/flow.d.ts"
+      "types": "./layouts/flow.d.ts",
+      "default": "./layouts/flow.js"
     },
     "./layouts/grid.js": {
-      "default": "./layouts/grid.js",
-      "types": "./layouts/grid.d.ts"
+      "types": "./layouts/grid.d.ts",
+      "default": "./layouts/grid.js"
     },
     "./layouts/masonry.js": {
-      "default": "./layouts/masonry.js",
-      "types": "./layouts/masonry.d.ts"
+      "types": "./layouts/masonry.d.ts",
+      "default": "./layouts/masonry.js"
     },
     "./LitVirtualizer.js": {
-      "default": "./LitVirtualizer.js",
-      "types": "./LitVirtualizer.d.ts"
+      "types": "./LitVirtualizer.d.ts",
+      "default": "./LitVirtualizer.js"
     },
     "./polyfillLoaders/ResizeObserver.js": {
-      "default": "./polyfillLoaders/ResizeObserver.js",
-      "types": "./polyfillLoaders/ResizeObserver.d.ts"
+      "types": "./polyfillLoaders/ResizeObserver.d.ts",
+      "default": "./polyfillLoaders/ResizeObserver.js"
     },
     "./polyfills/resize-observer-polyfill/ResizeObserver.js": {
-      "default": "./polyfillLoaders/ResizeObserver.js",
-      "types": "./polyfillLoaders/ResizeObserver.d.ts"
+      "types": "./polyfillLoaders/ResizeObserver.d.ts",
+      "default": "./polyfillLoaders/ResizeObserver.js"
     },
     "./support/method-interception.js": {
-      "default": "./support/method-interception.js",
-      "types": "./support/method-interception.d.ts"
+      "types": "./support/method-interception.d.ts",
+      "default": "./support/method-interception.js"
     },
     "./support/resize-observer-errors.js": {
-      "default": "./support/resize-observer-errors.js",
-      "types": "./support/resize-observer-errors.d.ts"
+      "types": "./support/resize-observer-errors.d.ts",
+      "default": "./support/resize-observer-errors.js"
     },
     "./virtualize.js": {
-      "default": "./virtualize.js",
-      "types": "./virtualize.d.ts"
+      "types": "./virtualize.d.ts",
+      "default": "./virtualize.js"
     }
   },
   "scripts": {
--- a/packages/labs/virtualizer/package.json
+++ b/packages/labs/virtualizer/package.json
@@ -9,40 +9,52 @@
   "type": "module",
   "exports": {
     ".": {
-      "default": "./lit-virtualizer.js"
+      "default": "./lit-virtualizer.js",
+      "types": "./lit-virtualizer.d.ts"
     },
     "./events.js": {
-      "default": "./events.js"
+      "default": "./events.js",
+      "types": "./events.d.ts"
     },
     "./layouts/flexWrap.js": {
-      "default": "./layouts/flexWrap.js"
+      "default": "./layouts/flexWrap.js",
+      "types": "./layouts/flexWrap.d.ts"
     },
     "./layouts/flow.js": {
-      "default": "./layouts/flow.js"
+      "default": "./layouts/flow.js",
+      "types": "./layouts/flow.d.ts"
     },
     "./layouts/grid.js": {
-      "default": "./layouts/grid.js"
+      "default": "./layouts/grid.js",
+      "types": "./layouts/grid.d.ts"
     },
     "./layouts/masonry.js": {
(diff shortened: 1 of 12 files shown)


packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -168,6 +168,10 @@ describe('lit-virtualizer and virtualize directive', () => {
     expect(uvd.shadowRoot?.textContent).not.to.include('[5 selected]');
 
     // Clearing event arrays so we can watch for specific future events.
+    await until(() => ulv.rangeChangedEvents.length > 0);
```
usergenic, line 171: Adding these four lines was an opportunistic fix for a flaky test that was advancing to the event clearing step before ensuring all events were received.

packages/labs/virtualizer/src/support/method-interception.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export type Function = (...args: any[]) => any;
```
rictic, line 8: There's a built in type with this name that also has roughly this type I think, could you use that?

packages/labs/virtualizer/src/support/method-interception.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @param object The method's host object
+ * @param methodName The name of the method to intercept/wrap
+ * @param interceptor The interceptor function that is called when the method is
+ *   called.  It is passed the original method as the first argument, followed by
+ *   the original method's arguments.
+ * @returns a teardown function that can be called to restore the original method
+ *   to the object.
+ */
+export function interceptMethod<
+  T extends {},
```
rictic, line 42: Does `T extends object` work here instead?

The `{}` type is the empty interface type, which is potentially broader than you mean

packages/labs/virtualizer/src/support/method-interception.ts
```diff
@@ -0,0 +1,65 @@
...
+ * An MethodInterceptor function is defined as accepting a function it is intercepting
+ * or "wrapping" calls to.  It has the same signature as the original function, except
+ * that it has an additional first argument that is the "original" or wrapped
+ * function that may be called or not depending on the interceptor's logic.  The
+ * approach is analagous to calling super() in a subclass's method.
+ */
+export type MethodInterceptor<T extends Function> = (
+  originalFunction: T | undefined,
+  ...args: Parameters<T>
+) => ReturnType<T>;
```
rictic, line 21: nice!

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
rictic: ah nice! I'd been noticing some flakiness from the virtualizer tests

rictic (commented):
a few suggestions, but the typescript spot check looks good to me!

packages/labs/virtualizer/src/support/method-interception.ts
usergenic: It apparently doesn't like it for `Parameters<T>` where `T extends Function`

Maybe I'll rename it something that doesn't consume the name or cause confusion.  `AnyFunction`?

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
usergenic: I'm going to try and address them where I find them -- mostly I encounter them during the CI automated checks because the environments are slower than my workstations... if you see any others me a msg and I'll try to fix it.  I definitely don't want Virtualizer flakes to slow down anyone's flow! :)

packages/labs/virtualizer/src/support/method-interception.ts
usergenic: `T extends object` works.  I'll go with that for now.  Thanks!

packages/labs/virtualizer/src/support/method-interception.ts
usergenic: I don't even want to tell you how long it took to resign myself to using an explicit `any` in the function type.  It was a wild ride through all the resources.  I had some really ugly expanded type definitions through that exploration.  Usually a sign one is on the wrong track...  :-p

packages/labs/virtualizer/src/support/method-interception.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * An InterceptorTeardown function reverses the effects of an interceptMethod().
+ */
+export type MethodInterceptorTeardown = () => void;
```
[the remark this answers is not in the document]
usergenic: OMG REALLY?  That's super cool!  I'm totally on board for both the coolness factor and the reliability of not forcing teardown to always be a user responsibility.  I wonder if that would work for the Mocha runs/teardowns case.  When method patches are stacked they have to be torn down in the same order or it throws an error-- OH I should add a test for that...!

packages/labs/virtualizer/src/support/method-interception.ts
usergenic: Renamed to `AnyFunction`. Apparently `Parameters<T>` requires `T` to extend `(...args: any[]) => any`

packages/labs/virtualizer/package.json
```diff
@@ -35,6 +35,12 @@
     "./polyfills/resize-observer-polyfill/ResizeObserver.js": {
       "default": "./polyfillLoaders/ResizeObserver.js"
     },
+    "./support/method-interception.js": {
+      "default": "./support/method-interception.js"
+    },
```
augustjk, line 40: this might be more of a tangent since it looks like it's been like this for a while but is it fine that we don't have `"types"` pointing to the `d.ts` files here like our other packages?

packages/labs/virtualizer/src/support/resize-observer-errors.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+import {
+  interceptMethod,
+  MethodInterceptorTeardown,
+} from './method-interception.js';
+
+/**
+ * In a testing environment with a setup/before and teardown/after callback pattern, this function
+ * can be used to quickly install ignore benign errors that are expected to occur in tests.  For example, if a
```
augustjk, line 14: unfinished comment?

usergenic: Oh goodness! Thanks for catching that.  I had an unsaved file in my editor I totally missed that.  Fixed; including param descriptions for before and after.

packages/labs/virtualizer/package.json
usergenic: Went ahead and added them.

augustjk: I tried testing it and it does seem look like TS is able to find it if the path and file location matches, but I think it's nice to be explicit too.

packages/labs/virtualizer/package.json
```diff
@@ -9,34 +9,52 @@
   "type": "module",
   "exports": {
     ".": {
-      "default": "./lit-virtualizer.js"
+      "default": "./lit-virtualizer.js",
+      "types": "./lit-virtualizer.d.ts"
```
augustjk, line 13: "types" should come first in "exports" per https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing:~:text=The%20%22types%22%20condition%20should%20always%20come%20first%20in%20%22exports%22.

packages/labs/virtualizer/package.json
```diff
@@ -9,34 +9,52 @@
...
     },
     "./polyfillLoaders/ResizeObserver.js": {
-      "default": "./polyfillLoaders/ResizeObserver.js"
+      "default": "./polyfillLoaders/ResizeObserver.js",
+      "types": "./polyfillLoaders/ResizeObserver.d.ts"
     },
     "./polyfills/resize-observer-polyfill/ResizeObserver.js": {
-      "default": "./polyfillLoaders/ResizeObserver.js"
+      "default": "./polyfillLoaders/ResizeObserver.js",
+      "types": "./polyfillLoaders/ResizeObserver.d.ts"
```
augustjk, line 45: is it intentional that these lines don't actually map to the `./polyfills/resize-observer-polyfill/RseizeObserver.js` path? if not, it probably should be fixed. that particular file also doesn't seem to have a `d.ts` file in the same directory. looks like the original source is `js` with a separate `d.ts` file. maybe we need to copy over the `d.ts` file too. sorry this could be done in a separate PR than this one. it just caught my eye.

packages/labs/virtualizer/src/support/resize-observer-errors.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+import {
+  interceptMethod,
+  MethodInterceptorTeardown,
+} from './method-interception.js';
+
+/**
+ * In a testing environment with a setup/before and teardown/after callback pattern, this function
+ * can be used to as a succinct declaration to ignore these errors in tests.
```
augustjk, line 14: is this more grammatically correct?
```suggestion
 * can be used as a succinct declaration to ignore these errors in tests.
```

packages/labs/virtualizer/package.json
usergenic: The `/polyfillLoaders/ResizeObserver.js` is actually the "loader" for the polyfill, not the actual polyfill, if that makes sense?

packages/labs/virtualizer/src/support/resize-observer-errors.ts
usergenic: Fix applied!

packages/labs/virtualizer/package.json
usergenic: Oh my!  Didn't realize ordering mattered in dictionary style objects.   Corrected!  Thanks for pointing it out-- I would have never suspected such!

packages/labs/virtualizer/package.json
augustjk: hard to see just from the selection but the path and the target are different.
https://github.com/lit/lit/blob/5d68be35c192e8c4109911eec727fbb598557f72/packages/labs/virtualizer/package.json#L32-L37

graynorton (approved):
LGTM! Do you want to add a brief section to the README?

usergenic:
Thanks, @graynorton I've updated README to describe usage of the three primary helper methods and added a test to demonstrate using `ignoreWindowResizeObserverLoopErrors` which I had only tested implicitly before.

PR #3934, opened 2023-06-02
Rename NodeTemplatePart to ChildTemplatePart

AndrewJakubowicz:
In Lit 2.0 we renamed `NodePart` to `ChildPart`. This renames the internal template part name to follow the same naming convention. Link: https://github.com/lit/lit/blob/6ba8d9d8a8749c6deff7c6adb0e395f438afac49/packages/lit-html/CHANGELOG.md?plain=1#L228

This change is internal only and for internal naming consistency.


augustjk (approved):
Good catch! For reference, `NodePart` was renamed to `ChildPart` in this PR https://github.com/lit/lit/pull/1480.

Though I find it questionable if this actually needs a version bump since it has 0 effect on end users.

rictic (approved):

AndrewJakubowicz:
I think it's worth the `patch` version bump in case someone is somehow depending on the type. It does end up in the `lit-html.d.ts` file.

PR #3937, opened 2023-06-07
Revert "Make _initialize and static _initializers private (#3895)"

rictic:
This reverts commit 338e1b1ce49f307f37f91876643bed400e687c6d.

An internal subclass of ReactiveElement still needs to be able to delay calling _initialize until hydration, long after construction.

FYI @justinfagnani


rictic:
This won't land in 3.0 but it should be landable in a few months, so whenever we do the next breaking change it'll qualify.

AndrewJakubowicz (approved):

PR #3942, opened 2023-06-08
[labs/ssr] Fix hydration for nested custom elements without attributes

### LINKED ISSUE #3939: Lit SSR defer-hydration not being removed on subcomponent that do not have a node binding

SSR (@lit-labs/ssr)

### Description

When you SSR a component with a subcomponent in its shadowroot, lit-ssr will add `defer-hydration` to the element. The hydration step of Lit will then crawl the lit comment markers and remove that attribute on the component with a node marker.

Though, if there is no binding to a sub-component, and thus it does not have a node part marker, then lit hydration will never remove the `defer-hydration` attribute on the child.

### Reproduction

live repro:
https://stackblitz.com/edit/node-jay1jm?file=index.html

useful commands in that repro are `build` to build the TS and `gen-index` to run lit-ssr to generate `index.html`

- create two lit elements
- place one of the elements in the shadow root of the other with no lit-bindings
- import the definition of the child component in the module of the parent element
- SSR the parent element onto a page
- load the javascript bundle for the parent element (and transitively the child element)
- See that `defer-hydration` has not been removed from the child element

### Workaround

Place a dummy binding on the child element `<my-second-element data-yolo=${swag}>`

### Is this a regression?

### Affected versions

Failing in lit@2.7.5 and @lit-labs/ssr@3.1.3

tested on: node 16+ and Chrome 113

augustjk:
Fixes #3939 

The check for `node.attrs.length > 0` made it so that op code for 'possible-node-marker' was never getting added even for nested custom elements if they had no attributes. Removing that condition check fixes this. There shouldn't be much extra work being done by removing this condition as most of the code in the block are for loops over the derived `attrInfo` array which would be empty if `node.attrs.length` is `0`.

### DIFF
--- a/.changeset/eighty-gifts-sparkle.md
+++ b/.changeset/eighty-gifts-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix adding node marker for hydration for nested custom elements without attributes.
--- a/.changeset/eighty-gifts-sparkle.md
+++ b/.changeset/eighty-gifts-sparkle.md
@@ -2,4 +2,4 @@
 '@lit-labs/ssr': patch
 ---
 
-Fix adding node marker for hydration for nested custom elements without attributes.
+Fix adding node marker for hydration for nested custom elements without attributes. This ensures nested custom elements have their `defer-hydration` attribute removed when parent is hydrated even without any attributes or bindings.
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -394,87 +394,82 @@ const getTemplateOpcodes = (result: TemplateResult) => {
             });
           }
         }
-        if (node.attrs.length > 0) {
-          const attrInfo = [] as Array<
-            [boolean, boolean, (typeof node.attrs)[0]]
-          >;
-          for (const attr of node.attrs) {
-            const isAttrBinding = attr.name.endsWith(boundAttributeSuffix);
-            const isElementBinding = attr.name.startsWith(marker);
-            if (isAttrBinding || isElementBinding) {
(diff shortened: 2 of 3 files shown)


packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -394,87 +394,82 @@ const getTemplateOpcodes = (result: TemplateResult) => {
             });
           }
         }
-        if (node.attrs.length > 0) {
```
AndrewJakubowicz, line 397: Would an alternative fix be lifting the `|| node.isDefinedCustomElement`  check here?

That would only allow through zero length attribute custom element nodes (rather than all nodes). Although I agree with your comment that this filter may be unnecessary as the extra work is on zero length arrays.

.changeset/eighty-gifts-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix adding node marker for hydration for nested custom elements without attributes.
```
AndrewJakubowicz, line 5: Nit: This contains *what* the change is but not the *why*. Is it worth mentioning the why?

I quite like the comment in the test: https://github.com/lit/lit/pull/3942/files#diff-e5374f241814044cc1f766670d048d5cb724a230333deedc677dfb47ac606b50R5267-R5269

augustjk: Good suggestion. Done!

AndrewJakubowicz (approved):

packages/labs/ssr/src/lib/render-value.ts
augustjk: As discussed offline we consider having less branching more readable when empty arrays will just flow through. Thank you for the discussion!

PR #3947, opened 2023-06-10
Make taskComplete a getter so that errors are only thrown when requested

### LINKED ISSUE #3946: Issue with uncaught exceptions thrown when a task function rejects/throws an error

Task (@lit-labs/task)

### Description

See this discord thread for a full description:
https://discord.com/channels/1012791295170859069/1047015641225371718/threads/1075045145596137582
@justinfagnani mentioned towards the bottom that an issue was going to be raised but I couldn't find anything so am raising it now as I have just run into this problem and currently the only solution is a little ugly namely manually adding a do nothing catch to 'taskComplete' to stop the error propogating.

### Reproduction

See https://discord.com/channels/1012791295170859069/1047015641225371718/threads/1075045145596137582

### Workaround

See https://discord.com/channels/1012791295170859069/1047015641225371718/threads/1075045145596137582

### Is this a regression?

### Affected versions

2.1.1

N/A

e111077:
fixes #3946

Task now will not throw errors unless `taskComplete` is specifically called. In that case the user should catch the error on `taskComplete`.

**GOTCHAS:**

The following case is a gotcha but it's just how promises work:

```ts
taskThatThrowsAnError.taskComplete.catch(() => {});

// this will throw an error. User needs to catch here too or await previous line
await taskThatThrowsAnError.taskComplete;
```

see reproduction of native behavior [here](https://lit.dev/playground/#gist=3c498dfa3f2dc6f066527b561a015a32)

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -127,9 +127,9 @@ export class Task<
    * is kept and only resolved when the new run is completed.
    */
   get taskComplete(): Promise<R> {
-    // If a task exists, return the cached promise. This is true in the case
+    // If a task run exists, return the cached promise. This is true in the case
     // where the user has called taskComplete in pending or completed state
-    // before and has not run a new task since.
+    // before and has not started a new task run since.
     if (this._taskComplete) {
       return this._taskComplete;
     }
@@ -142,12 +142,10 @@ export class Task<
         this._rejectTaskComplete = rej;
       });
 
-      // Otherwise we are at a run's completion or this is the first request
-      // and we are not in the middle of a task (i.e. INITIAL).
+      // Otherwise we are at a task run's completion or this is the first
+      // request and we are not in the middle of a task (i.e. INITIAL).
     } else {
-      this._resolveTaskComplete = undefined;
-      this._rejectTaskComplete = undefined;
-      this._taskComplete = Promise.resolve(this._value as R);
+      this._taskComplete = Promise.resolve(this._value!);
     }
 
     return this._taskComplete;
@@ -214,14 +212,10 @@ export class Task<
   async run(args?: T) {
     args ??= this._getArgs?.();
 
-    // Clear the last complete task in initial because it may be a resolved
-    // promise. Also clear if complete or error because the value returned by
+    // Clear the last complete task run in INITIAL because it may be a resolved
+    // promise. Also clear if COMPLETE or ERROR because the value returned by
     // awaiting taskComplete may have changed since last run.
-    if (
-      this.status === TaskStatus.COMPLETE ||
-      this.status === TaskStatus.ERROR ||
-      this.status === TaskStatus.INITIAL
-    ) {
+    if (this.status !== TaskStatus.PENDING) {
       this._taskComplete = undefined;
       this._resolveTaskComplete = undefined;
       this._rejectTaskComplete = undefined;
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -257,8 +257,6 @@ export class Task<
           this._rejectTaskComplete?.(error);
         }
         this._value = result as R;
-        // The run's ended so don't generate a task complete promise and either
-        // return the existing, resolved promise or a new resolved promise.
         this._error = error;
       }
       // Request an update with the final value.
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -127,35 +127,29 @@ export class Task<
    * is kept and only resolved when the new run is completed.
    */
   get taskComplete(): Promise<R> {
-    // True only when called for the first time or after a run is complete.
-    const lastPromiseCompletedOrFirstRequest = !this._resolveTaskComplete;
+    // If a task exists, return the cached promise. This is true in the case
+    // where the user has called taskComplete in pending or completed state
+    // before and has not run a new task since.
+    if (this._taskComplete) {
+      return this._taskComplete;
+    }
 
-    // Generate an in-progress promise if the last promise was completed or this
-    // is the first time the promise is being requested and we are in the middle
-    // of a task.
-    if (
-      lastPromiseCompletedOrFirstRequest &&
-      this.status === TaskStatus.PENDING
-    ) {
+    // Generate an in-progress promise if the the status is pending and has been
+    // cleared by .run().
+    if (this.status === TaskStatus.PENDING) {
       this._taskComplete = new Promise((res, rej) => {
         this._resolveTaskComplete = res;
         this._rejectTaskComplete = rej;
       });
+
       // Otherwise we are at a run's completion or this is the first request
       // and we are not in the middle of a task (i.e. INITIAL).
-    } else if (
-      !this._taskComplete ||
-      this.status === TaskStatus.COMPLETE ||
-      this.status === TaskStatus.ERROR
-    ) {
-      // Always generate a new, resolved promise on COMPLETE and ERROR b/c we
-      // don't know if the last resolved value has changed since last requested.
+    } else {
       this._resolveTaskComplete = undefined;
       this._rejectTaskComplete = undefined;
       this._taskComplete = Promise.resolve(this._value as R);
     }
 
-    // Return the promise or the cached in-progress promise.
     return this._taskComplete;
   }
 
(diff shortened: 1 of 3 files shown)


packages/labs/task/src/task.ts
```diff
@@ -44,6 +44,7 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
   task: TaskFunction<T, R>;
   args?: ArgsFunction<T>;
   autoRun?: boolean;
+  throwErrors?: boolean;
```
justinfagnani, line 47: I would love to find a way to address this without a new config option. I think an expansion of options can be bad DX as users have to decide whether they should say the option or not, and what the default it.

I was thinking we could make taskComplete a getter, so we don't even create a promise that could have an uncaught error. The getter would have to create an already resolved promise if the task is complete.

Your comment about an error renderer is insightful! It's be great if we detected an error renderer and also suppressed uncaught errors. And maybe the same for the onError callback?

e111077: What is the advantage of the getter? Is it just not having to create the promise in the first place? If so, I feel like the bloat of the extra checks around it might outweigh the memory savings if that's the only reason

e111077: I like the onError and render.errror suggestion. Would remove the need for the config option

justinfagnani: The getter is so that we avoid even making the promise until it's been asked for, so by default we don't get any uncaught errors. Then if a user does access the promise, and don't handle the error, they'll get the warning if they don't have the error, but in that case they really should get the warning.

justinfagnani: Actually, I think with the on-demand promise creation, you won't need to look at error handlers, since the warning would only ever be made of the user gets a promise and doesn't handle it correctly.

e111077: issues with the automatic error handling via `shouldCatch = !!this.onError || this.render.error` – it's difficult to determine whether the user has an error handler in the error message. We could do something like:

`this._hasErrorRenderer = this._hasErrorRenderer || !!renderer.error` in `render()`, but that requires:

1. The user to have called `render()` at least once with an `error` option before the task has completed
   - The user could be rendering their error renderer _after_ the task has completed in a conditional template
2. The user to continue rendering the `error` renderer because the logic is optimistic that there is an error handler to handle the following case:
   - ```js
      ${task.render({error: ... })}
      ${task.render({complete: ... })}
      ```

I'd also wager that the user having to handle catch on `taskComplete` is not great because it doesn't make sense that you'd have to catch on `taskComplete` as opposed to the task itself handling it.

WDYT of:

- Do not turn `taskComplete` into a getter
- No `throwErrors` and just catch every time and it's up to the user to use our intended API to handle errors at all time

justinfagnani: I think making `taskComplete` a getter solves this. There's no need to look for an error handler in render. The logic should be simple: if a user called `.taskCompete`, reject that promise on error, if not, there's no promise to reject and no unhandled rejection.

We already do reject the taskComplete promise, so this wouldn't even be a breaking change. If we don't reject we have to at least resolve it, so we don't leave it hanging, but I think that's a bit weird.

e111077: took a while to untangle the logic in my head, but tests should be robust. PTAL

e111077: updated main comment with a gotcha, but it's just how Promises work

packages/labs/task/src/task.ts
```diff
@@ -226,17 +246,19 @@ export class Task<
...
           } catch {
             // Ignore user errors from onError.
           }
           this.status = TaskStatus.ERROR;
-          this._rejectTaskComplete(error);
+          this._rejectTaskComplete?.(error);
         }
         this._value = result as R;
+        // The run's ended so don't generate a task complete promise and either
+        // return the existing, resolved promise or a new resolved promise.
```
justinfagnani, line 261: Is this comment still valid?

I think this is the spot where you can do:

```ts
this._resolveTaskComplete = undefined;
this._rejectTaskComplete = undefined;
this._taskComplete = undefined;
```

justinfagnani:
Nice!

I think the method you have here is fine to go in, but I added a suggestion that might make things simper.

packages/labs/task/src/task.ts
e111077: oops not valid anymore. No longer necessary since I do that inthe taskComplete getter.

packages/labs/task/src/task.ts
```diff
@@ -115,21 +115,53 @@ export class Task<
...
   /**
    * A Promise that resolve when the current task run is complete.
    *
    * If a new task run is started while a previous run is pending, the Promise
    * is kept and only resolved when the new run is completed.
    */
-  taskComplete!: Promise<R>;
+  get taskComplete(): Promise<R> {
+    // True only when called for the first time or after a run is complete.
+    const lastPromiseCompletedOrFirstRequest = !this._resolveTaskComplete;
```
justinfagnani, line 131: I think this could be a _little_ simpler if you clear the promise state in `run()` so that if `this._taskComplete` is defined here you just always return it. Then if it's not defined you have the branches to create it with either a pending promise or a resolved one based on the status.

e111077: you are right, MUCH cleaner! thanks!

e111077:
oh no tests are failing on sauce but not locally!

packages/labs/task/src/task.ts
```diff
@@ -115,21 +115,47 @@ export class Task<
...
+      this._taskComplete = new Promise((res, rej) => {
+        this._resolveTaskComplete = res;
+        this._rejectTaskComplete = rej;
+      });
+
+      // Otherwise we are at a run's completion or this is the first request
+      // and we are not in the middle of a task (i.e. INITIAL).
+    } else {
+      this._resolveTaskComplete = undefined;
+      this._rejectTaskComplete = undefined;
```
justinfagnani, line 149: These should always already be undefined

packages/labs/task/src/task.ts
```diff
@@ -115,21 +115,47 @@ export class Task<
...
    *
    * If a new task run is started while a previous run is pending, the Promise
    * is kept and only resolved when the new run is completed.
    */
-  taskComplete!: Promise<R>;
+  get taskComplete(): Promise<R> {
+    // If a task exists, return the cached promise. This is true in the case
+    // where the user has called taskComplete in pending or completed state
+    // before and has not run a new task since.
+    if (this._taskComplete) {
```
justinfagnani, line 133: Optional, some 🏌️ :

```ts
return this._taskComplete ??= this.status === TaskStatus.PENDING
  ? new Promise((res, rej) => {
      this._resolveTaskComplete = res;
      this._rejectTaskComplete = rej;
    })
  : Promise.resolve(this._value!);
```

(except with your nice comments added back in)

packages/labs/task/src/task.ts
```diff
@@ -191,15 +213,20 @@ export class Task<
...
     args ??= this._getArgs?.();
+
+    // Clear the last complete task in initial because it may be a resolved
+    // promise. Also clear if complete or error because the value returned by
+    // awaiting taskComplete may have changed since last run.
     if (
       this.status === TaskStatus.COMPLETE ||
-      this.status === TaskStatus.ERROR
+      this.status === TaskStatus.ERROR ||
+      this.status === TaskStatus.INITIAL
```
justinfagnani, line 223: This is the same as `this.status !== TaskStatus.PENDING` now, yeah?

Which brings up an interesting question: is `run()` doing the correct thing when called on an already pending task? No need to address that in the PR, but maybe we're missing a test?

justinfagnani (approved):
Awesome!!

packages/labs/task/src/task.ts
e111077: oh yikes, veto!

e111077: Resolving with ACK as I think clarity is more important than the unbundled byte size here

justinfagnani: fwiw, I like the readability with the logical and assignment and ternary. The logical and assignment tells me that this is using the current value if available, or computing a new one. The ternary tells me that we're computing a single value based on a condition. To me that's more apparent than with the if/else block with multiple assignment statements form, since any other logic could be in the if/else blocks.

It also compresses better even after minification because there's only one reference to `this._taskComplete` instead of five. `this.` is something a minifier can't get rid of even if it changes the property name.

packages/labs/task/src/task.ts
```diff
@@ -115,21 +115,45 @@ export class Task<
...
+    // cleared by .run().
+    if (this.status === TaskStatus.PENDING) {
+      this._taskComplete = new Promise((res, rej) => {
+        this._resolveTaskComplete = res;
+        this._rejectTaskComplete = rej;
+      });
+
+      // Otherwise we are at a task run's completion or this is the first
+      // request and we are not in the middle of a task (i.e. INITIAL).
+    } else {
```
rictic, line 147: What about if taskStatus is ERROR?

PR #3948, opened 2023-06-10
Optimize lit 3.0's size, add a regression test

### LINKED ISSUE #3777: [all] Size regression when emitting ES2021
In #3775 we're advancing our output to ES2021, but we're seeing a size regression in our built output.

We're probably going to move forward for now with ES2021 in the pre-releases, just to get early signal about whether that breaks anything, but before we launch we need to reevaluate this because it's probably not worth it to advance the language version at the cost of size.

rictic:
Fixes #3777 

This test may be a bit too aggro, but it also does seem very nice to be able to track with each change how it contributes to some very key top of line numbers.


packages/lit-html/src/lit-html.ts
```diff
@@ -982,13 +982,14 @@ class Template {
...
     // leak, but every time we prepare a template, we immediately render it
     // and re-use the walker in new TemplateInstance._clone().
-    debugLogEvent?.({
-      kind: 'template prep',
-      template: this,
-      clonableTemplate: this.el,
-      parts: this.parts,
-      strings,
-    });
+    debugLogEvent &&
```
justinfagnani, line 985: Can we file a Terser bug on this? If you can eliminate `&&`, in null cases you can eliminate `?.`

rictic:
Interesting, I can't get the `f && f()` form to be dead code eliminated on
the terser playground, even when f is const undefined and the file is a
module. I wonder why this is working! Needs.more investigation.

packages/lit-html/package.json
```diff
@@ -402,7 +403,8 @@
         "test:prod",
         "test:node",
         "test:node-dev",
-        "check-version"
+        "check-version",
+        "test:prodsize"
```
justinfagnani, line 407: In `lit-html` you have this run as part of the tests, but I don't see that with `lit`.

I think it might be best to have these run as part of the benchmark workflow, and instead of being run by Wireit as part of another step, have it in it's own step so if it fails we see it separated in the GH workflow logs.

Being part of the benchmark runner would also let us compare sizes against head and the latest release similarly to the benchmarks, since we already check those out and build them.

packages/lit-html/src/lit-html.ts
rictic: Filed as https://github.com/terser/terser/issues/1400

packages/lit-html/package.json
rictic: Refactored to be part of the benchmark runner and to report as comments on the PR.

I'd maybe have liked it if there was no comment in the case there's no size change, but this way we do know that the size benchmark at least, and it ensures that we can update the comment after an initial comment reporting a problem.

packages/lit/package.json
```diff
@@ -247,7 +247,9 @@
         "test/*_test.html",
         "development/test/*_test.html",
         "test/polyfill-support/*_test.html",
-        "development/test/polyfill-support/*_test.html"
+        "development/test/polyfill-support/*_test.html",
+        "lit-core.min.js{,.map}",
```
sorvell, line 251: Does this mean these will be included in releases? And is this new? If so, is that stressful at all?

scripts/check-size.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+if (litHtmlSize !== expectedLitHtmlSize) {
+  if (litHtmlSize < expectedLitHtmlSize) {
+    reportProblem(`This is a size improvement! It shaves **${
+      expectedLitHtmlSize - litHtmlSize
+    }** bytes off lit-html.js. Congratulations!
+
+Please update \`expectedLitHtmlSize\` to ${litHtmlSize} in \`scripts/check-size.js\` and push a new commit to your PR.`);
```
sorvell, line 63: nit: maybe avoid repeating these messages?

scripts/check-size.js
```diff
@@ -0,0 +1,71 @@
...
+  /end render|template instantiated/.test(litHtmlSrc)
+) {
+  reportProblem(
+    `The \`debugLogEvent\` function isn't being dead code eliminated. This will likely greatly increase binary size. We depend on terser for this, and its dead code elimination can be a bit finicky.`
+  );
+}
+
+if (litCoreSize !== expectedLitCoreSize) {
+  if (litCoreSize < expectedLitCoreSize) {
+    reportProblem(`This is a size improvement! It shaves **${
```
sorvell, line 43: Suggest reporting the baseline number as well to provide context.

... because 200 of 1e6 is nbd, but 200 of 300 is a lot.

sorvell (changes requested):

packages/lit/package.json
rictic: No, this is just communicating that these files are generated by this command to wireit. Should have no impact beyond wireit caching etc

scripts/check-size.js
rictic: All four messages are distinct, they're each combination of larger/smaller and lit-html/lit-core

We could parameterize, but that seems like premature generalization

scripts/check-size.js
rictic: +1, done

sorvell (approved):

Issue #3954, opened 2023-06-14
Memory leak in lit-labs/context


AndrewJakubowicz:
### Which package(s) are affected?

Context (@lit-labs/context)

### Description

Issue captured by: https://discord.com/channels/1012791295170859069/1097902841265393716/1116635815527858237

Text from the comment:

I thought this was fixed but it's clear that context has a memory leak. This is a small repro and and in Chrome you can see how the DOM nodes grow and the garbage collector can not remove the about view because it has subscribe: true. If you use Edge to look at detached elements it's more clear that the detached elements are building up as you navigate between home and about multiple times. This is a link to a repro https://lit-context-example.vercel.app/ and this is the repository if you wish to test it without subscribe set to true. When you do, the garbage collector can remove the about view as expected. https://github.com/marcuslindblom/lit-context-example

Thank you @marcuslindblom for reporting this issue

### Reproduction

Live repro: https://lit-context-example.vercel.app/

Source code repro: https://github.com/marcuslindblom/lit-context-example

Repro:

 - Open https://lit-context-example.vercel.app/
 - In the performance profiler select memory. Garbage collect & then profile.
 - While profiling switch between Home & About page a couple times.
 - Trigger a garbage collection and stop profiling.
 - Notice that the number of nodes has increased & garbage collection did not clean up everything back to where we started.

This leak results in Nodes accumulating over time.

### Workaround

N/A

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

lit-labs/context latest

### Browser/OS/Node environment

N/a

AndrewJakubowicz:
The issue seems related to the `callbacks` map here: https://github.com/lit/lit/blob/fc2fd4c8f4a25b9a85073afcb38614209e079bb9/packages/labs/context/src/lib/value-notifier.ts#L23 maintaining callbacks and growing after every new subscription.

Verified by adding a `console.log('we are adding a callback!', this.callbacks.size);` message. When switching back and forth in the repo, the size of the callbacks grows. Nothing disposes the callbacks.

PR #3956, opened 2023-06-15
[lit-labs/context] Fix context memory leak when subscribe is true.

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/3954

The issue is that the `dispose` function was not being passed when adding the callback.
So if the element was removed before receiving any updates, it would never get access to the dispose function and could not remove the callback from the Map resulting in the unbounded leak.

The reason the element does not leak if it received an update is because `updateObservers` passes the `dispose` function:
https://github.com/lit/lit/blob/4f71df99e6df2e5fb31259c53f01f2e87209e3d5/packages/labs/context/src/lib/value-notifier.ts#L47-L51

### Testing

Tested by adding a memory test (and checking that it fails by a huge amount without fix applied).

### DIFF
--- a/packages/labs/context/src/lib/value-notifier.ts
+++ b/packages/labs/context/src/lib/value-notifier.ts
@@ -57,8 +57,11 @@ export class ValueNotifier<T> {
           this.callbacks.delete(callback);
         });
       }
+      const dispose = this.callbacks.get(callback)!;
+      callback(this.value, dispose);
+    } else {
+      callback(this.value);
     }
-    callback(this.value);
   }
 
   clearCallbacks(): void {
--- a/packages/labs/context/src/lib/value-notifier.ts
+++ b/packages/labs/context/src/lib/value-notifier.ts
@@ -20,7 +20,7 @@ type Disposer = () => void;
  * for a number of use cases.
  */
 export class ValueNotifier<T> {
-  private callbacks: Map<ContextCallback<T>, Disposer> = new Map();
+  private disposers: Map<ContextCallback<T>, Disposer> = new Map();
 
   private _value!: T;
   public get value(): T {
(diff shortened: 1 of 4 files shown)


packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -57,8 +57,11 @@ export class ValueNotifier<T> {
           this.callbacks.delete(callback);
         });
       }
+      const dispose = this.callbacks.get(callback)!;
```
justinfagnani, line 60: maybe while you're here: `callbacks` isn't a great name. I usually like naming a map for the values it contains, not the key. I think this would be better named as `disposers`.

justinfagnani (approved):
nice!

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -57,8 +57,11 @@ export class ValueNotifier<T> {
...
+      callback(this.value, dispose);
+    } else {
+      callback(this.value);
```
augustjk, line 63: Nit: to be more consistent with the specifier used in rest of file.
```suggestion
      const disposer = this.callbacks.get(callback)!;
      callback(this.value, disposer);
    } else {
      callback(this.value);
```

augustjk (approved):
Excellent find and fix!

packages/labs/context/src/test/context-provider_test.ts
```diff
@@ -157,3 +158,63 @@ suite('@consume: multiple instances', () => {
...
+  let container: HTMLElement;
+
+  // Make a big array set on an expando to exaggerate any leaked DOM
+  const big = () => new Array(10000).fill(0);
+
+  setup(async () => {
+    container = document.createElement('div');
+    container.innerHTML = `
+        <context-provider value="1000">
+            <context-consumer p="${big()}"></context-consumer>
```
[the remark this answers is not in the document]
AndrewJakubowicz: Yeah - although I'll remove this one since the useful measurements come from the elements added in the test.

packages/labs/context/src/lib/value-notifier.ts
AndrewJakubowicz: Nice!

packages/labs/context/src/test/context-provider_test.ts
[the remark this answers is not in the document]
AndrewJakubowicz: Removed - test still red without change, and green with fix.

PR #3964, opened 2023-06-21
Update to TypeScript 5.1

rictic:


packages/labs/ssr-client/src/directives/render-light.ts
```diff
@@ -22,6 +22,7 @@ class RenderLightDirective extends Directive {
     if (typeof instance.renderLight === 'function') {
       return instance.renderLight();
     }
+    return;
```
justinfagnani, line 25: Do we need this? I thought TS 5.1 actually allowed you to omit returns for functions which include `undefined` in their return type? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#easier-implicit-returns-for-undefined-returning-functions

packages/labs/task/src/task.ts
```diff
@@ -285,9 +285,14 @@ export class Task<
...
       case TaskStatus.ERROR:
         return renderer.error?.(this.error);
-      default:
-        // exhaustiveness check
-        this.status as void;
+      default: {
+        // exhaustiveness check that can get optimized out
+        if (false as boolean) {
+          const never: never = this.status;
+          console.error(`Unexpected task status: ${never}`);
```
justinfagnani, line 292: Why the `console.error()`? This can never be reached?

packages/labs/task/src/task.ts
```diff
@@ -285,9 +285,14 @@ export class Task<
...
+        }
+        return;
```
justinfagnani, line 294: Will Terser remove the `default` case if this return is present? Can we remove the return?

packages/ts-transformers/src/tests/idiomatic-decorators-test.ts
```diff
@@ -1508,6 +1508,7 @@ const tests = (test: uvu.Test<uvu.Context>, options: ts.CompilerOptions) => {
         console.log('click');
       }
     }
+    export { MyElement };
```
justinfagnani, line 1511: This seems unfortunate... it's not necessarily idiomatic to split the class declaration and export like this. Is there no other option with 5.1?

justinfagnani (approved):

packages/labs/task/src/task.ts
rictic: Good point, rewrote to still fail if the type isn't `never` at this point, but in a lighter weight way

packages/labs/ssr-client/src/directives/render-light.ts
rictic: This started to fail with this change! I guess they loosened the check in some places and tightened it in others?

packages/labs/task/src/task.ts
rictic: Without this return, TypeScript gives a `Not all code paths return a value ts(7030)` error.

packages/ts-transformers/src/tests/idiomatic-decorators-test.ts
rictic: Ah, mentally I was in the world of the analyzers where we're running the same tests with many versions of TSC. TSC 5.1 does the more idiomatic thing here in the code it emits, and so I've just updated the goldens.

packages/labs/ssr-client/src/directives/render-light.ts
augustjk: maybe it would not be needed if `update()` method had return type annotated?

Issue #3965, opened 2023-06-21
[lit-html] ref() directive should accept undefined


justinfagnani:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

Sometimes a ref object may be optional, especially if it's an external property. In these cases it would be very useful to let `ref()` accept `undefined` and just do nothing.

Example:

```ts
class MyElement extends LitElement {
  /**
   * Pass a Ref object to receive a reference to the "foo" div
   */
  @property({attribute: false})
  fooRef?: Ref<HTMLDivElement>;
  
  render() {
    return html`<div ${ref(this.fooRef}}></div>`;
  }
}
```

This example currently gives us a type error at `ref(this.fooRef}`, since `this.fooRef` may be `undefined`.

### Alternatives and Workarounds

We can make an orphan ref:

```ts
html`<div ${ref(this.fooRef ?? createRef()}}></div>`;
```

But that's very ugly.

PR #3968, opened 2023-06-21
[lit-html] Allow undefined to be passed to the ref() directive

justinfagnani:
Fixes https://github.com/lit/lit/issues/3965

`ref()` already handled undefined refs correctly, so this is a one-char change.

### DIFF
--- a/.changeset/grumpy-stingrays-grab.md
+++ b/.changeset/grumpy-stingrays-grab.md
@@ -1,5 +1,6 @@
 ---
 'lit-html': patch
+'lit': patch
 ---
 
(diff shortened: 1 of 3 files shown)


rictic (approved):

.changeset/grumpy-stingrays-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
```
augustjk, line 3:
```suggestion
---
'lit-html': patch
'lit': patch
---
```

augustjk (approved):

PR #3969, opened 2023-06-21
[lit-html] Make RefOrCallback generic

### LINKED ISSUE #3966: [lit-html] RefOrCallback should be generic

### Description

`RefOrCallback` is not generic like `Ref<T>` and so does not forward a ref element type to `Ref`:

```ts
export type RefOrCallback = Ref | ((el: Element | undefined) => void);
```

This should be:

```ts
export type RefOrCallback<T = Element> = Ref<T> | ((el: T | undefined) => void);
```

### Reproduction

n/a

### Workaround

Make your own type.

### Is this a regression?

### Affected versions

2.x

all

justinfagnani:
Make `RefOrCallback` generic like `Ref<T>`.

Fixes #3966

### DIFF
--- a/.changeset/smooth-swans-sin.md
+++ b/.changeset/smooth-swans-sin.md
@@ -1,5 +1,6 @@
 ---
 'lit-html': patch
+'lit': patch
 ---
 
 Make RefOrCallback generic like Ref<T>
--- a/.changeset/smooth-swans-sin.md
+++ b/.changeset/smooth-swans-sin.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Make RefOrCallback generic like Ref<T>
--- a/packages/lit-html/src/directives/ref.ts
+++ b/packages/lit-html/src/directives/ref.ts
@@ -38,7 +38,7 @@ const lastElementForContextAndCallback: WeakMap<
   WeakMap<Function, Element | undefined>
 > = new WeakMap();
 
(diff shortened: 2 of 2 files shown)


packages/lit-html/src/directives/ref.ts
```diff
@@ -38,7 +38,7 @@ const lastElementForContextAndCallback: WeakMap<
   WeakMap<Function, Element | undefined>
 > = new WeakMap();
 
-export type RefOrCallback = Ref | ((el: Element | undefined) => void);
+export type RefOrCallback<T = Element> = Ref<T> | ((el: T | undefined) => void);
```
rictic, line 41: Should T extend Element?

rictic (approved):

packages/lit-html/src/directives/ref.ts
justinfagnani: Ref objects and callbacks could _technically_ be used with any type. `Ref` already doesn't have T extending Element, so this also seems inline with that.

.changeset/smooth-swans-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
```
augustjk, line 3:
```suggestion
---
'lit-html': patch
'lit': patch
---
```

augustjk (approved):

PR #3973, opened 2023-06-22
[labs/virtualizer] Prevent test flakes by emptying an event collector that occasionally captures events on initialize.

### LINKED ISSUE #3970: [labs/virtualizer] Flakey virtualizer test

### Description

Example run:
https://github.com/lit/lit/actions/runs/5336743485/jobs/9671855750#step:8:218

cc @usergenic 

### Reproduction

n/a

### Workaround

n/a

### Is this a regression?

### Affected versions

latest

in CI

usergenic:
This small change should fix #3970


packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -148,6 +148,11 @@ describe('lit-virtualizer and virtualize directive', () => {
     expect(ulv.shadowRoot?.textContent).to.include('[5 selected]');
     expect(uvd.shadowRoot?.textContent).to.include('[5 selected]');
 
+    // There may occasionally be visibility change events from the initial render, so
+    // we clear those out before we start watching for new ones to avoid flaky tess.
+    ulv.visibilityChangedEvents.splice(0);
```
justinfagnani, line 153: Do we expect at most one event here? Should we validate that in addition to clearing to ensure we don't have some other unexpected state creep in?

And when we clear, would setting `length = 0` have a more obvious intention?

augustjk: I'm also seeing we clear these again in lines 182-183 before the assertion happens. Could the flakiness be somewhere in between those?

usergenic: @justinfagnani: 
> And when we clear, would setting length = 0 have a more obvious intention?

So I was today years old when I learned you can set length to a smaller value in order to truncate arrays in JavaScript.  Fixed.

@augustjk:
While there's a non-zero possibility that more events could come between the time that the `await until(...)` finds a greater than 0 length on the event collector, I think the most likely cause for flakes was related to behavior during the settling phase of initial virtualizer render.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -148,6 +148,11 @@ describe('lit-virtualizer and virtualize directive', () => {
     expect(ulv.shadowRoot?.textContent).to.include('[5 selected]');
     expect(uvd.shadowRoot?.textContent).to.include('[5 selected]');
 
+    // There may occasionally be visibility change events from the initial render, so
+    // we clear those out before we start watching for new ones to avoid flaky tess.
+    ulv.visibilityChangedEvents.length = 0;
+    uvd.visibilityChangedEvents.length = 0;
```
augustjk, line 154: Maybe I'm not reading this correctly but I don't see any assertion against `ulv.visibilityChangedEvents.length` between this newly added emptying and the next emptying in line 182 below. So the lack of this couldn't have been the source of the failed assertion on line 206, could it?

usergenic: It's not an assertion exactly, it's this bit on line 178-179

```ts
    await until(() => ulv.visibilityChangedEvents.length > 0);
    await until(() => uvd.visibilityChangedEvents.length > 0);
```

Those lines wait for the events to come in.  By truncating the collectors prior to these lines we ensure we actually get the initia visibility changed events and THEN clear them out to ensure we don't get any others after that ~ line 206

usergenic: The weird thing about the await untils is they almost operate like an assertion in the sense that if they time out the test fails, so it's worth reading them almost like an assertion.

augustjk: Aha! So if there's already stuff in there we'd not actually wait for the thing to happen. Thanks for the clarification!

augustjk (approved):
Thanks for the quick turnaround on this!

PR #3976, opened 2023-06-23
Exclude ancestors above a fixed positioned element

### LINKED ISSUE #3904: Clipping parents include ancestors of `position: fixed` element

### Description

When a Virtualizer that is using the Grid display is within a `position: fixed` element ancestor that has sufficient width to display the grid, but which is in a subsequent ancestor with `overflow: hidden` and a width that is insufficient to display the grid, the list will not be rendered.

Within [this code](https://github.com/lit/lit/blob/main/packages/labs/virtualizer/src/Virtualizer.ts#L932-L946), I think we can update the filter to look something like the following to address this situation.

```
function getClippingAncestors(el: HTMLElement, includeSelf = false) {
  let foundPositionFixed = false;
  return getElementAncestors(el, includeSelf).filter(
    (a) => {
      const styles = getComputedStyle(a);
      foundPositionFixed = foundPositionFixed || styles.position === 'fixed';
      return !foundPositionFixed && styles.overflow !== 'visible';
    }
  );
}
```

### Reproduction

In the `index.html` of the following repro, toggle on and off the `width: 75px` or the `overflow: hidden` to experience the issue.

https://lit.dev/playground/#gist=611a3a8ced4d469acbc8bae840b59a30

### Workaround

- Patch the package.
- Make some hard (and possibly impossible) CSS application decisions

### Is this a regression?

### Affected versions

... >=1.0.0

All.

Westbrook:
When an element leverages `position: fixed` it creates a new stacking context and is no longer clipped by its ancestors in the majority of contexts.

fixes #3904


usergenic (approved):
LGTM

PR #3977, opened 2023-06-23
[labs/ssr] Make attribute handling case-insensitive

### LINKED ISSUE #3972: [labs/ssr] Uppercase in attribute name behaves differently

SSR (@lit-labs/ssr)

### Description

Reported in https://github.com/lit/lit/issues/3971#issuecomment-1601767854

In the `reactive-element` code for matching property to attribute, we lowercase the property name when we convert it to an attribute name, since that's what the DOM does. Looks like our DOM shim for SSR does not lowercase attribute names, and ends up with a mismatch.

### Reproduction

See linked issue comment above.

### Workaround

Supply all attributes as lowercase in template.

### Is this a regression?

### Affected versions

latest

.

augustjk:
Fixes #3972 

`ssr-dom-shim` was updated so all attribute methods lowercase the name as the browser does.

`LitElementRenderer` now also lowercases the attribute name when calling `attributeToProperty` with it.

### DIFF
--- a/.changeset/new-news-count.md
+++ b/.changeset/new-news-count.md
@@ -1,5 +1,4 @@
 ---
-'@lit-labs/ssr-dom-shim': patch
 '@lit-labs/ssr': patch
 ---
 
--- a/.changeset/new-news-count.md
+++ b/.changeset/new-news-count.md
@@ -1,0 +1,6 @@
+---
+'@lit-labs/ssr-dom-shim': patch
+'@lit-labs/ssr': patch
+---
+
+Make attribute handling case-insensitive as it is in the browser.
--- a/.changeset/new-news-count.md
+++ b/.changeset/new-news-count.md
@@ -2,4 +2,4 @@
 '@lit-labs/ssr': patch
 ---
 
-Make attribute handling case-insensitive as it is in the browser.
+Make attribute handling case-insensitive for custom element rendering as it is in the browser for HTML elements.
--- a/packages/labs/ssr/src/lib/lit-element-renderer.ts
+++ b/packages/labs/ssr/src/lib/lit-element-renderer.ts
@@ -81,7 +81,7 @@ export class LitElementRenderer extends ElementRenderer {
     _old: string | null,
     value: string | null
   ) {
-    attributeToProperty(this.element as LitElement, name, value);
+    attributeToProperty(this.element as LitElement, name.toLowerCase(), value);
   }
 
   override *renderShadow(renderInfo: RenderInfo): RenderResult {
--- a/packages/labs/ssr/src/lib/lit-element-renderer.ts
+++ b/packages/labs/ssr/src/lib/lit-element-renderer.ts
@@ -81,7 +81,7 @@ export class LitElementRenderer extends ElementRenderer {
     _old: string | null,
     value: string | null
   ) {
-    attributeToProperty(this.element as LitElement, name.toLowerCase(), value);
+    attributeToProperty(this.element as LitElement, name, value);
   }
 
   override *renderShadow(renderInfo: RenderInfo): RenderResult {
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -4870,7 +4870,7 @@ export const tests: {[name: string]: SSRTest} = {
             assert.strictEqual((el as any).camelProp, 'boundProp2');
           },
           html: {
-            root: `<le-mixed-attr-binding camelProp="boundProp2" static></le-mixed-attr-binding>`,
+            root: `<le-mixed-attr-binding camelprop="boundProp2" static></le-mixed-attr-binding>`,
             'le-mixed-attr-binding': `<div>\n  [boundProp2]\n</div>`,
           },
         },
--- a/packages/labs/ssr/src/test/integration/tests/basic.ts
+++ b/packages/labs/ssr/src/test/integration/tests/basic.ts
@@ -4825,6 +4825,88 @@ export const tests: {[name: string]: SSRTest} = {
     };
   },
 
+  'LitElement: Attribute binding (mixed case)': () => {
+    return {
+      registerElements() {
+        class LEMixedAttrBinding extends LitElement {
+          @property()
+          camelProp = 'default';
+          override render() {
+            return html` <div>[${this.camelProp}]</div> `;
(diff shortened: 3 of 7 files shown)


augustjk:
https://github.com/lit/lit/pull/3978 fixes the breaking test in the pipeline.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -4825,6 +4825,88 @@ export const tests: {[name: string]: SSRTest} = {
...
+          args: ['boundProp2'],
+          async check(assert: Chai.Assert, dom: HTMLElement) {
+            const el = dom.querySelector(
+              'le-mixed-attr-binding'
+            )! as LitElement;
+            await el.updateComplete;
+            assert.strictEqual((el as any).camelProp, 'boundProp2');
+          },
+          html: {
+            root: `<le-mixed-attr-binding camelProp="boundProp2" static></le-mixed-attr-binding>`,
```
AndrewJakubowicz, line 4873: Should `camelProp=` be `camelprop=`?

augustjk: Good eye! Turns out it doesn't matter because our DOM assertion treats those as equivalent, as the browser would lowercase it anyway. Still fixed it for consistency's sake.

augustjk:
> #3978 fixes the breaking test in the pipeline.

Rebased to bring this in and local tests are now passing.

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -81,7 +81,7 @@ export class LitElementRenderer extends ElementRenderer {
     _old: string | null,
     value: string | null
   ) {
-    attributeToProperty(this.element as LitElement, name, value);
+    attributeToProperty(this.element as LitElement, name.toLowerCase(), value);
```
AndrewJakubowicz, line 84: Everything in this change looks great! I'm only leaving a comment on this line because I don't have the full picture of what this change does.

My naive thoughts are that the logic for `attributeToProperty` results in calling `_$attributeToProperty` which may call a user defined `converter.fromAttribute`. Does making the name lower case here make sense in case a specific case sensitive name is expected down the line? Or will attributeToProperty always expect a lower case name because attributes from the DOM end up lower cased?

Thank you!

AndrewJakubowicz (approved):
Great!

packages/labs/ssr/src/lib/lit-element-renderer.ts
augustjk: I confirmed that the behavior in the client is that it always gets lowercase attributes, regardless of the casing of the attribute name in the original template. In this particular method, we're defining the `attributeChangedCallback` lifecycle and the browser turns all HTML attributes to lowercase so it would be consistent for us to do that in the server as well.

AndrewJakubowicz: Brilliant! Is it worth committing some of that excellent comment as a code comment?

AndrewJakubowicz (dismissed):
Thanks for addressing feedback!

packages/labs/ssr/src/lib/lit-element-renderer.ts
augustjk: Although.. along those lines, perhaps `attributeChangedCallback` should already be getting lowercased names instead. Going to consider moving this to the call site of `attributeChangedCallback` instead.

augustjk:
Discussed offline with @AndrewJakubowicz about whether lowercasing all attributes is the right thing to do since SVG elements in the browser are case-sensitive. Marking this as draft while we consider this.

augustjk:
Updated to not modify ssr-dom-shim, just the element renderer. https://github.com/lit/lit/issues/3972#issuecomment-1613816597

.changeset/new-news-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Make attribute handling case-insensitive as it is in the browser.
```
AndrewJakubowicz, line 5: Nit: Worth updating this changeset to similarly specify "for HTML elements" or something to differentiate from the SVG namespace.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -4825,6 +4825,88 @@ export const tests: {[name: string]: SSRTest} = {
     };
   },
 
+  'LitElement: Attribute binding (mixed case)': () => {
```
AndrewJakubowicz, line 4828: Is it possible to add a test case where attribute casing is preserved? E.g. svg element in a render template?

.changeset/new-news-count.md
augustjk: Good suggestion!

packages/labs/ssr/src/test/integration/tests/basic.ts
augustjk: I added it separately as a unit test that makes sure the rendered markup has attribute casing preserved. How the browser handles that is up to the browser.

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -360,6 +360,15 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
   });
 
+  test('element with mixed case attributes', async () => {
+    const {render, templateWithMixedCaseAttrs} = await setup();
+    const result = await render(templateWithMixedCaseAttrs('dynamic'));
+    assert.is(
+      result,
+      `<!--lit-part F/N3wW1gNj8=--><!--lit-node 0--><svg dynamicCamel="dynamic" staticCamel="static"></svg><!--/lit-part-->`
+    );
+  });
```
AndrewJakubowicz, line 370: *chefs kiss*

AndrewJakubowicz (approved):

packages/labs/ssr/src/test/integration/tests/basic.ts
AndrewJakubowicz: What is making you suspect something?

augustjk: I tried some combinations in the playground and found inconsistent results. I'll investigate and document in a separate issue. :)

PR #3978, opened 2023-06-23
[labs/react] Only add `suppressHydrationWarning` prop in the client

augustjk:
https://github.com/lit/lit/pull/3977 causes the `suppressHydrationWarning` prop to actually get emitted as an attribute on the host element for wrapped components being server rendered as shown in this test output: https://github.com/lit/lit/actions/runs/5361038597/jobs/9726603547?pr=3977#step:8:443

The `suppressHydrationWarning` only has effect on React components in the client so there's no reason to add it if in node mode.

### DIFF
--- a/.changeset/calm-deers-hope.md
+++ b/.changeset/calm-deers-hope.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz (approved):
Nice!

PR #3980, opened 2023-06-25
[labs/analyzer] Add separate entrypoint for createPackageAnalyzer

justinfagnani:
Previously `index.ts` imported both the main `Analyzer` class and `createPackageAnalyzer()`, which uses Node APIs.

This PR separates the entrypoints into `index.ts` which only exports the Analyzer class and model interfaces, and `package-analyzer.ts` which exports all of that, plus `createPackageAnalyzer()`.

This is a breaking change.


packages/labs/analyzer/package.json
```diff
@@ -53,7 +53,8 @@
     "!/lib/.tsbuildinfo"
   ],
   "exports": {
-    ".": "./index.js"
+    ".": "./index.js",
+    "./package-analyzer.js": "./package-analyzer.js"
```
rictic, line 57: package-analyzer.js or analyze-package.js?

rictic (changes requested):
Build is failing

kevinpschaaf (approved):
LGTM, modulo the naming issue @rictic raised. Did you want the top-level import to be `package-analyzer.js` to match `analyzer.js`? If so, need to rename the file, otherwise rename the `exports` in package.json.

justinfagnani:
Argh. I don't understand why the build is failing on CI now. Build and tests are passing on my machine even after a git clean. I also don't understand why it was _passing_ on my machine earlier as I had some obviously bad imports.

rictic (approved):

PR #3985, opened 2023-06-29
[labs/virtualizer] Update test method 'until' to support error/Chai exceptions

usergenic:
One of the key elements of the virtualizer tests are waiting asynchronously for certain state to change (black-box testing virtualizer behaviors) and this is achieved with a support method called `until` which is often used to test a specific rendered content condition, for example:

```ts
await until(() => container.text.includes('something'))
```

The "problem" with this approach is that if the condition isn't met, the error output is not descriptive enough to say what the container text actually DOES contain, which means manually adding console.log() to debug etc, instead of having useful test error reporting.

Additionally, these `until()` statements work like defacto expectations of state, which means that we often wind up re-expressing the same expectation with Chai later on, to ensure our test asserts the right things or skipping the assertion, making it less clear what the test is trying to prove.

This PR adds a behavior to the underlying code of `until` (by way of the underlying core method `eventually`) to have it swallow errors and retry the expectation repeatedly until the timeout happens, throwing the error in that case, to enable using expectation syntax to clearly express the test assertions inline.  This means that if the expectation continues to fail after the timeout, the error reporting will show the expectation failure message, which is often much more expressive.  For example the following would report what the container text was if 'something' wasn't found.  Because `until` cares about a truthy result before it returns, we provide a different semantic `pass` that ignores the return value and only cares about a no error case:

```ts
await pass(() => expect(container.text).to.include('something'));
```

I opportunistically converted the uses of `await until` to `await pass` where it could benefit from expectation syntax and this helped remove some duplicate code as well as helped me diagnose and fix a previously flaky test, hopefully once and for all.


packages/labs/virtualizer/src/test/layouts/flow.test.ts
```diff
@@ -54,7 +54,7 @@ describe('flow layout', () => {
     const virtualizer = (await until(() =>
       container.querySelector('lit-virtualizer')
     )) as LitVirtualizer;
-    expect(virtualizer).to.be.instanceof(LitVirtualizer);
+    await until(() => expect(virtualizer).to.be.instanceOf(LitVirtualizer));
```
augustjk, line 57: Is the `until()` use here for the instance check needed? Either the line above would resolve to the found `lit-virtualizer` element or would throw. I'm assuming it's not like we're waiting for a found `lit-virtualizer` element to get upgraded.

Unless you're thinking of combining them like
```ts
await until(() => expect(container.querySelector('lit-virtualizer')).to.be.instanceOf(LitVirtualizer));
```

Also TIL that chai aliases `.instanceof` and `.instanceOf`.

packages/labs/virtualizer/src/test/scenarios/range-changed-events.test.ts
```diff
@@ -38,10 +38,10 @@ describe('RangeChanged event', () => {
...
       ></lit-virtualizer>
     </div>`);
-    const virtualizer = (await until(() =>
-      container.querySelector('lit-virtualizer')
-    )) as LitVirtualizer;
-    expect(virtualizer).to.be.instanceof(LitVirtualizer);
+    const virtualizer = await until(
+      () => container.querySelector('lit-virtualizer') as LitVirtualizer
+    );
+    await until(() => expect(virtualizer).to.be.instanceOf(LitVirtualizer));
```
augustjk, line 44: Same comment as previous about `until()` use here.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -140,70 +140,56 @@ describe('lit-virtualizer and virtualize directive', () => {
     uvd.selected = selected;
 
     // Wait for events from initial render to fire.
-    await until(() => ulv.rangeChangedEvents.length > 0);
-    await until(() => uvd.rangeChangedEvents.length > 0);
-    await until(() => ulv.visibilityChangedEvents.length > 0);
-    await until(() => uvd.visibilityChangedEvents.length > 0);
+    await until(() => expect(ulv.rangeChangedEvents.length).to.equal(3));
```
augustjk, line 143: When a condition function with an `expect` is passed in, we're expecting it to actually throw rather than return a falsy value so we're not going to have the benefit of the stack trace of `until()`. Would it make sense to just use `eventually()` directly instead for all of these then?

If we otherwise anticipate/encourage usage of `expect` with `until`, I think there should be mention of that for in `until()`'s jsdoc.

augustjk (commented):
I really like this utility.

I think an alternative approach to use in conjunction with something like `chai-as-promised` would be to have a modified version of `until` or `eventually` which, instead of taking a condition function, takes a value getter and a test function separately such that it'll get the value, test until test is true and resolve with the value, or reject on timeout with the latest value that didn't pass the test. Then you could write assertions against that promise. But that would result in writing the test logic twice so not great.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
```diff
@@ -140,70 +140,56 @@ describe('lit-virtualizer and virtualize directive', () => {
...
+    await until(() => expect(uvd.rangeChangedEvents.length).to.equal(3));
+    await until(() => expect(ulv.visibilityChangedEvents.length).to.equal(3));
+    await until(() => expect(ulv.visibilityChangedEvents.length).to.equal(3));
```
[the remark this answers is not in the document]
usergenic: OMG good catch!@!!!!

packages/labs/virtualizer/src/test/layouts/flow.test.ts
usergenic: Oh you know what I did leave that on accident.  I started converting those to handle a case that never happens and then I undid them but forgot and left that there.  Removed.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
usergenic: So the `until` stack trace code was actually a hack to get something useful to display, namely the source code line that expressed the condition.  It's less useful than what the expectation error emits.  But you're right that the usage of this here is kind of misleading in the sense that the expectation object itself is a truthy thing, being just object or whatever, but its existence doesn't "mean" anything truthy in the semantic sense of until.

The subtle thing about `eventually` is that you still want to return something truthy so you don't have a timeout condition while it's waiting for something truthy.  That means that for concision you're still likely to wind up using the expectation object directly as the return value like:

```ts
await eventually(() => expect(something).to.be.whatever);
```

For example, I beat myself up trying to figure out why I couldn't bundle expectations like this until I realized I needed to return something true at the end or it would timeout at 1000ms and then if i did it more than 2 times per test the default 2000ms mocha timeout would fail my test:

```ts
await eventually(() => {
   expect(x).to.equal(1);
   expect(y).to.equal(2);
});

In reality what I was trying to do was have a block of code that would run until it didn't cause an exception.  But I didn't make a semantic for that yet.  Could create a new function for that though, perhaps `pass`?   Like:

```ts
await pass(() => expect(x).to.equal(1));

await pass(() => {
  expect(x).to.equal(1)
  expect(y).to.equal(2)
});
```

This semantic basically says it will continue as long as no error is thrown and it will keep retrying until it passes or the timeout is reached.

Either that or I could redefine `eventually` to perform that function, I just don't know if this meaning is implied in that word?

augustjk: Aha, right. I do like `pass` or `passing` for the semantic that only checks for exceptions thrown and not a boolean condition.

This would leave `until`/`eventually` generally to be used for asynchronously grabbing truthy values.

usergenic: On quite-a-bit-more further thought, as written, `eventually` doesn't really support any direct use cases that are different than `until` since it still will timeout until it gets something truthy.  Its name is also vague, since eventually and until kind of mean the same thing.  I think I'll just have `pass` and `until` which are more clearly semantically different, even if not immediately obvious what they are as functions unless one sees them in action or reads their docs.

packages/labs/virtualizer/src/test/scenarios/range-changed-events.test.ts
usergenic: Removed.

packages/labs/virtualizer/src/test/scenarios/element-and-directive-parity.test.ts
usergenic: Okay code pushed and all tests changed over to reflect the new `until` and `pass` reality.

packages/labs/virtualizer/src/test/scenarios/sizing-and-styling.test.ts
```diff
@@ -78,24 +78,23 @@ describe('Properly sizing virtualizer within host element', () => {
...
 
-    await until(
-      () =>
-        root.querySelector(
-          'custom-element-containing-lit-virtualizer'
-        ) instanceof CustomElementContainingLitVirtualizer
+    await until(() =>
+      expect(
+        root.querySelector('custom-element-containing-lit-virtualizer')
+      ).to.be.instanceOf(CustomElementContainingLitVirtualizer)
```
augustjk, line 84:
```suggestion
    await pass(() =>
      expect(
        root.querySelector('custom-element-containing-lit-virtualizer')
      ).to.be.instanceOf(CustomElementContainingLitVirtualizer)
```

packages/labs/virtualizer/src/test/scenarios/sizing-and-styling.test.ts
```diff
@@ -78,24 +78,23 @@ describe('Properly sizing virtualizer within host element', () => {
...
       'custom-element-containing-lit-virtualizer'
     )!;
-    await until(
-      () =>
-        ceclv.shadowRoot?.querySelector('lit-virtualizer') instanceof
-        LitVirtualizer
+    await until(() =>
+      expect(
+        ceclv.shadowRoot?.querySelector('lit-virtualizer')
+      ).to.be.instanceOf(LitVirtualizer)
```
augustjk, line 92:
```suggestion
    await pass(() =>
      expect(
        ceclv.shadowRoot?.querySelector('lit-virtualizer')
      ).to.be.instanceOf(LitVirtualizer)
```

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
```diff
@@ -26,7 +26,9 @@ async function windowError(message: string) {
   setTimeout(() => {
     throw new CountingError(message);
   });
-  await until(() => errors >= expectedErrorCount);
+  await until(() =>
+    expect(errors).to.be.greaterThanOrEqual(expectedErrorCount)
+  );
```
augustjk, line 31: This file needs to import `pass` for the following suggestions.
```suggestion
  await pass(() =>
    expect(errors).to.be.greaterThanOrEqual(expectedErrorCount)
  );
```

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
```diff
@@ -97,16 +99,18 @@ describe('preventResizeObserverLoopErrorEventDefaults', () => {
...
     await windowError('Do not record this error');
-    await until(() => messages.length >= 1);
-    expect(messages).to.deep.equal([
-      'Uncaught Error: Do not record this error',
-    ]);
+    await until(() =>
+      expect(messages).to.deep.equal([
+        'Uncaught Error: Do not record this error',
+      ])
+    );
```
augustjk, line 106:
```suggestion
    await pass(() =>
      expect(messages).to.deep.equal([
        'Uncaught Error: Do not record this error',
      ])
    );
```

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
```diff
@@ -97,16 +99,18 @@ describe('preventResizeObserverLoopErrorEventDefaults', () => {
...
-    expect(messages).to.deep.equal([
-      'Uncaught Error: Do not record this error',
-      'Uncaught Error: ResizeObserver loop limit exceeded',
-    ]);
+    await until(() =>
+      expect(messages).to.deep.equal([
+        'Uncaught Error: Do not record this error',
+        'Uncaught Error: ResizeObserver loop limit exceeded',
+      ])
+    );
```
augustjk, line 114:
```suggestion
    await pass(() =>
      expect(messages).to.deep.equal([
        'Uncaught Error: Do not record this error',
        'Uncaught Error: ResizeObserver loop limit exceeded',
      ])
    );
```

packages/labs/virtualizer/src/test/helpers.ts
```diff
@@ -107,29 +107,67 @@ export function isInViewport(element: Element, viewport?: Element) {
...
+        return [undefined, e as Error];
+      }
+    }
   });
 }
 
 /**
+ * Returns a promise for the result of the code block given.  Errors are swallowed
+ * by the function until the timeout is reached, in which case the error will be
+ * thrown.  This means you can `pass` to wait for expectations like so:
```
augustjk, line 144: Maybe better wording here?
```suggestion
 * Returns a promise for the result of the code block given.  Errors are swallowed
 * by the function until the timeout is reached, after which the error will be
 * thrown.  This means you can use `pass` to wait for expectations like so:
```

augustjk: I also thought about if `pass` should just return `void` if the main use case is for retrying until an exception is not thrown.

Leaving as is does allow usage for callback to retrieve a value instead of just running assertions too. Semantically a bit further from `pass` though.

packages/labs/virtualizer/src/test/helpers.ts
```diff
@@ -107,29 +107,67 @@ export function isInViewport(element: Element, viewport?: Element) {
...
+  await eventually(() => {
+    result = block();
+    return result || true;
+  }, timeout);
+  return result! as T;
+}
+
+/**
+ * Returns a promise which will resolve to the first truthy result of condition
+ * function or the last result of calling it within the given timeout.
```
augustjk, line 165: This line actually no longer applies here because `until` will always throw if the last result is not truthy.

augustjk (changes requested):
Awesome!

I don't know how strictly we want to adhere to usage of `pass` for assertions vs. only those that need them due to callback returning falsy, i.e. multiple assertions in a single callback.

Since both `until` and `pass` use `eventually` it does work but should we differentiate based on semantics?

Some more suggestions on comments as well.

usergenic:
> I don't know how strictly we want to adhere to usage of `pass` for assertions vs. only those that need them due to callback returning falsy, i.e. multiple assertions in a single callback.
> 
> Since both `until` and `pass` use `eventually` it does work but should we differentiate based on semantics?

I think semantically, `pass` makes sense for any of the Chai `expect` use cases, whether single-line or multi-line.  And there's a subtle double-meaning for `pass` in reference to test/assert language, even though that isn't the only intent for the word.  So I'm trying to get them all, but thank you for finding the others!

packages/labs/virtualizer/src/test/helpers.ts
usergenic: Oh you're right-- I copied from `eventually` and then didn't read properly.  Good catch.

packages/labs/virtualizer/src/test/helpers.ts
usergenic: Def better wording.  Applying.

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
usergenic: Fixed

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
usergenic: Fixed

packages/labs/virtualizer/src/test/support/resize-observer-errors.test.ts
usergenic: Fixed

packages/labs/virtualizer/src/test/scenarios/sizing-and-styling.test.ts
usergenic: Fixed

packages/labs/virtualizer/src/test/scenarios/sizing-and-styling.test.ts
usergenic: Fixed

augustjk (approved):
Thank you this is great!

usergenic:
Thanks for the thoughtful attention and keen eyes, @augustjk!

PR #3987, opened 2023-06-29
[lit-html] change CompiledTemplate h field to a TemplateStringsArray

AndrewJakubowicz:
Similar to: https://github.com/lit/lit/pull/2642

Thank you @rictic for proposing this!

**NOTE:** changing technique to a branding tag function.

### DIFF
--- a/.changeset/weak-roses-laugh.md
+++ b/.changeset/weak-roses-laugh.md
@@ -1,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+Add an additional security brand check to CompiledTemplates; Similar to [#2307](https://github.com/lit/lit/pull/2307)
--- a/.changeset/weak-roses-laugh.md
+++ b/.changeset/weak-roses-laugh.md
@@ -3,4 +3,4 @@
 'lit': patch
 ---
 
-Add an additional security brand check to CompiledTemplates; Similar to [#2307](https://github.com/lit/lit/pull/2307)
+Change the `h` field of `CompiledTemplate`s to a `TemplateStringsArray` preventing the spoofing of `CompiledTemplate`s by JSON injection attacks. This should not be a breaking change for most users unless you're using CompiledTemplates. This is a necessary security fix, similar to [#2307](https://github.com/lit/lit/pull/2307).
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -652,6 +654,39 @@ export interface DirectiveParent {
...
+function trustFromTemplateString(
+  tsa: TemplateStringsArray,
+  stringFromTSA: string
+): TrustedHTML {
+  // A security check to prevent spoofing of Lit template results.
+  // In the future, we may be able to replace this with Array.isTemplateObject,
+  // though we might need to make that check inside of the html and svg
+  // functions, because precompiled templates don't come in as
+  // TemplateStringArray objects.
+  if (!Array.isArray(tsa) || !tsa.hasOwnProperty('raw')) {
```
AndrewJakubowicz, line 666: This could be improved to use: https://github.com/google/safevalues/blob/main/src/internals/string_literal.ts

.changeset/weak-roses-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+Add an additional security brand check to CompiledTemplates; Similar to [#2307](https://github.com/lit/lit/pull/2307)
```
rictic: note that this is a security fix that may be breaking, but only to an API that most likely no one is using because it's the output of a compiler that we haven't released yet

packages/lit-html/src/lit-html.ts
rictic: True, but we probably don't want to take the dependency or pay for the bytes and runtime overhead.

rictic (commented):
Mind checking out how many bytes this adds to the 3.0 branch? Just merge it into 3.0 and then do `npm run benchmark:size`

.changeset/weak-roses-laugh.md
AndrewJakubowicz: Please see new changeset message. You were fast on the review!

rictic: Nice!

AndrewJakubowicz:
Tested memory impact of this change:

**This is a size regression**. It adds **42** bytes to lit-core.min.js which is 14,740B at head. We're very sensitive to size increases. If this is intended, please update `expectedLitCoreSize` to 14782 in `scripts/check-size.js` and push a new commit to your PR.

**This is a size regression**. It adds **36** bytes to lit-html.js which is 7,220B at head. We're very sensitive to size increases. If this is intended, please update `expectedLitHtmlSize` to 7256 in `scripts/check-size.js` and push a new commit to your PR.

rictic:
36 bytes well spent!

rictic (approved):
Great work!

augustjk (approved):
Great!

PR #3992, opened 2023-07-06
[infra] fix root package-lock to bump published dependencies

AndrewJakubowicz:
Ran `npm i`, bumping the package versions of the recently released: https://github.com/lit/lit/pull/3990


augustjk (approved):
Thanks! Wish there was a way to make this happen in the Version Packages PR, but this'll do for now.

augustjk:
Possible fix: https://github.com/changesets/action/issues/203#issuecomment-1460115073
Turns out I tried this a while back when we had lerna https://github.com/lit/lit/pull/2921 but it didn't work. https://github.com/lit/lit/pull/3028#issuecomment-1154556547
Maybe now with npm workspaces this would work here?

PR #3984, opened 2023-06-28
Initial commit of lit-labs/compiler

AndrewJakubowicz:
Issue: https://github.com/lit/lit/issues/189
RFC: https://github.com/lit/rfcs/pull/21

# lit-labs/compiler

This commit adds a first (and rough) iteration of the Lit compiler for pre-compiling html TTLs. This change passes almost all lit-html tests.

The skipped tests are:
 - Some dev mode warnings, which currently get compiled out. Should instead become diagnostics.
 - The `cache` directive depends on a template string array & needs handling added for compiled templates.

This initial change was used for benchmarking the possible benefits of the compiler.

No changeset is included because the package is marked private and will not be released.

### Why

lit-html is very fast, but if you're already using a build step there are opportunities to squeeze more performance out. By pre-compiling templates, we can skip the [**prepare** phase of lit-html rendering](https://github.com/lit/lit/blob/main/dev-docs/design/how-lit-html-works.md#2i-prepare).

### Benchmarks

The below benchmarks can be run by checking out this branch, building the monorepo and running individual benchmarks e.g., `npm run benchmark:lit-html:template-heavy-compiled -w @lit-internal/benchmarks`.

**Kitchen sink** average tachometer benchmark results:

- [**6.7% faster**]: `render-compiled`: 7.93ms - 8.10ms, `render-uncompiled`: 8.50ms - 8.69ms
- [roughly same]: `update-compiled`: 17.92ms - 18.56ms, `update-uncompiled`: 18.08ms - 19.00ms
- [roughly same]: `nop-update-compiled`: 5.53ms - 5.66ms, `nop-update-uncompiled`: 5.64ms - 5.80ms

**Template Heavy** average tachometer benchmark results:

 - [**46% faster**]: `render-compiled`: 9.51ms - 9.67ms, `render-uncompiled`: 17.66ms - 18.03ms
 - [**21% faster**]: `update-compiled`: 26.40 - 26.91ms, `update-uncompiled`: 33.62 - 34.51ms

**repeat** average tachometer benchmark results:

 - [**14% faster**]: `render-compiled`: 7.40ms - 7.62ms, `render-uncompiled`: 8.56ms - 8.94ms
 - [**8% faster**]: `update-compiled`: 189.57ms - 197.54ms, `update-uncompiled`: 206.51ms - 218.68ms

TODO:
 - [x] Fix wireit config
 - [ ] cleanup
 - [ ] docs
 - [x] hook up benchmarks
 - [x] Add sourcemap tests
[...]

PR #3993, opened 2023-07-06
[lit-html][labs/ssr-client] Fix compiled template support

AndrewJakubowicz:
Change is an independent part of https://github.com/lit/lit/pull/3984

## Summary

Fix `isTemplateResult` return type to include `CompiledTemplateResult`s, and then fix locations in `lit` that didn't handle `CompiledTemplate`s. Primarily the `cache` directive.

## Changes

 - Fixes the return type of `isTemplateResult` to return the correct type (as per current implementation): a `TemplateResult` or a `CompiledTemplateResult`.
 - Fix cache directive to work with both uncompiled and compiled templates using a helper to access the template strings array.
 - Made it explicit that `hydration` and `experimental-hydration` do not work with compiled templates by making the check more strict. *Maybe could add a dev mode error?*
- Added compiled template cache directive unit test, and `isTemplateResult` type tests.

## Considerations

I marked this as a minor change, however it is a breaking change in terms of TypeScript types. The return type of `isTemplateResult` without a second argument now has an additional consideration and could break users of `isTemplateResult`. However the type now more truthfully conveys what `isTemplateResult` is asserting.

## Alternative considered:
 - Make `isTemplateResult` stricter such that the type is only a `TemplateResult`. A breaking change (although people should realistically not be depending on `CompiledTemplateResult`s yet)
 - Add the following helpers: `isCompiledTemplateResult` to check for CompiledTemplates, and `getTemplateStringsArray` to generically retrieve the template strings array from either a compiled or uncompiled template result.

## Test plan

Unit tests and type only tests added.

### DIFF
--- a/.changeset/chilled-drinks-travel.md
+++ b/.changeset/chilled-drinks-travel.md
@@ -1,0 +1,7 @@
+---
+'@lit-labs/ssr-client': minor
+'lit-html': minor
+'lit': minor
+---
+
+Fix return type of `isTemplateResult` helper to include the `CompiledTemplateResult` and fix the `cache` directive to work correctly with `CompiledTemplateResult`s. Add an explicit error message for ssr hydration to throw on `CompiledTemplate`s. Any usage of `isTemplateResult` without a second argument should also be handling the compiled template.
--- a/.changeset/chilled-drinks-travel.md
+++ b/.changeset/chilled-drinks-travel.md
@@ -1,5 +1,5 @@
 ---
-'@lit-labs/ssr-client': minor
+'@lit-labs/ssr-client': patch
 'lit-html': minor
 'lit': minor
 ---
--- a/.changeset/chilled-drinks-travel.md
+++ b/.changeset/chilled-drinks-travel.md
@@ -1,7 +1,6 @@
 ---
-'@lit-labs/ssr-client': patch
 'lit-html': minor
 'lit': minor
 ---
 
-Fix return type of `isTemplateResult` helper to include the `CompiledTemplateResult` and fix the `cache` directive to work correctly with `CompiledTemplateResult`s. Add an explicit error message for ssr hydration to throw on `CompiledTemplate`s. Any usage of `isTemplateResult` without a second argument should also be handling the compiled template.
+Fix return type of `isTemplateResult` helper to include the `CompiledTemplateResult` and fix the `cache` directive to work correctly with `CompiledTemplateResult`s. Also add an explicit `isCompiledTemplateResult` helper.
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -280,7 +280,7 @@ const openChildPart = (
       const template = (
         ChildPart.prototype as ChildPart & {
           _$getTemplate(
-            result: TemplateResult
+            value: TemplateResult
           ): ConstructorParameters<typeof TemplateInstance>[0];
         }
       )._$getTemplate(value);
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -269,17 +269,22 @@ const openChildPart = (
     // if (marker.data !== 'lit-part') {
     //   throw new Error('Hydration value mismatch: Primitive found where TemplateResult expected');
     // }
-  } else if (isTemplateResult(value)) {
+  } else if (
+    isTemplateResult(value) &&
+    // Ensure this is a TemplateResult and not a CompiledTemplateResult.
+    typeof value['_$litType$'] == 'number'
+  ) {
+    const result = value as TemplateResult;
     // Check for a template result digest
-    const markerWithDigest = `lit-part ${digestForTemplateResult(value)}`;
+    const markerWithDigest = `lit-part ${digestForTemplateResult(result)}`;
     if (marker.data === markerWithDigest) {
       const template = (
         ChildPart.prototype as ChildPart & {
           _$getTemplate(
-            value: TemplateResult
+            result: TemplateResult
           ): ConstructorParameters<typeof TemplateInstance>[0];
         }
-      )._$getTemplate(value);
+      )._$getTemplate(result);
       const instance = new TemplateInstance(template, part);
       stack.push({
         type: 'template-instance',
@@ -287,7 +292,7 @@ const openChildPart = (
         part,
         templatePartIndex: 0,
         instancePartIndex: 0,
-        result: value,
+        result,
       });
       // For TemplateResult values, we set the part value to the
       // generated TemplateInstance
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -269,11 +269,10 @@ const openChildPart = (
     // if (marker.data !== 'lit-part') {
     //   throw new Error('Hydration value mismatch: Primitive found where TemplateResult expected');
     // }
-  } else if (
-    isTemplateResult(value) &&
-    // Ensure this is a TemplateResult and not a CompiledTemplateResult.
-    typeof value['_$litType$'] == 'number'
-  ) {
+  } else if (isTemplateResult(value)) {
+    if (typeof value['_$litType$'] != 'number') {
+      throw new Error('compiled templates are not supported');
+    }
     const result = value as TemplateResult;
     // Check for a template result digest
     const markerWithDigest = `lit-part ${digestForTemplateResult(result)}`;
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -20,6 +20,7 @@ import {
   isPrimitive,
   isSingleExpression,
   isTemplateResult,
+  isCompiledTemplateResult,
 } from 'lit-html/directive-helpers.js';
 
 // In the Node build, this import will be injected by Rollup:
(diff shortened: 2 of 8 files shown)


packages/lit-html/src/directive-helpers.ts
```diff
@@ -42,17 +48,23 @@ export const TemplateResultType = {
...
-  type?: TemplateResultType
-): value is TemplateResult =>
+  type: TemplateResultType | undefined
+) =>
   type === undefined
     ? // This property needs to remain unminified.
       (value as TemplateResult)?.['_$litType$'] !== undefined
-    : (value as TemplateResult)?.['_$litType$'] === type;
+    : (value as TemplateResult)?.['_$litType$'] ===
+      type) as unknown as isTemplateResultType;
```
rictic, line 68: I think we want an isCompiledTemplateResult guard function here, otherwise we're repeating the check in a number of places, requiring casts

rictic (changes requested):
Looks good!

packages/lit-html/src/directive-helpers.ts
```diff
@@ -42,17 +48,23 @@ export const TemplateResultType = {
 export type TemplateResultType =
   (typeof TemplateResultType)[keyof typeof TemplateResultType];
 
+type isTemplateResultType = {
+  (val: unknown): val is TemplateResult | CompiledTemplateResult;
+  (val: unknown, type: TemplateResultType): val is TemplateResult;
```
augustjk, line 53: This just occurred to me as I was looking through, but wasn't part of the original implementation. Maybe we don't need this kind of granularity, so feel free to ignore.
```suggestion
  <T extends TemplateResultType>(val: unknown, type: T): val is TemplateResult<T>;
```

.changeset/chilled-drinks-travel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit-labs/ssr-client': patch
+'lit-html': minor
+'lit': minor
+---
+
+Fix return type of `isTemplateResult` helper to include the `CompiledTemplateResult` and fix the `cache` directive to work correctly with `CompiledTemplateResult`s. Add an explicit error message for ssr hydration to throw on `CompiledTemplate`s. Any usage of `isTemplateResult` without a second argument should also be handling the compiled template.
```
augustjk, line 7: I think it would make sense to make 2 separate changeset file, having different lines for `lit-html` and `lit` vs `@lit-labs/ssr-client`.

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -270,24 +270,28 @@ const openChildPart = (
     //   throw new Error('Hydration value mismatch: Primitive found where TemplateResult expected');
     // }
   } else if (isTemplateResult(value)) {
+    if (typeof value['_$litType$'] != 'number') {
```
augustjk, line 273: Nit:
```suggestion
    if (typeof value['_$litType$'] !== 'number') {
```

augustjk: But also, could use `isCompiledTemplateResult` checker like @rictic suggested.

augustjk (changes requested):
Feel free to push back on any requested changes.

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
AndrewJakubowicz: Added `isCompiledTemplateResult` check.

.changeset/chilled-drinks-travel.md
AndrewJakubowicz: Great suggestion - done

packages/lit-html/src/directive-helpers.ts
AndrewJakubowicz: Great! Done

packages/lit-html/src/directive-helpers.ts
AndrewJakubowicz: Awesome!

packages/lit-html/src/directive-helpers.ts
```diff
@@ -42,17 +48,23 @@ export const TemplateResultType = {
...
-  type?: TemplateResultType
-): value is TemplateResult =>
+  type: TemplateResultType | undefined
+) =>
   type === undefined
     ? // This property needs to remain unminified.
       (value as TemplateResult)?.['_$litType$'] !== undefined
-    : (value as TemplateResult)?.['_$litType$'] === type;
+    : (value as TemplateResult)?.['_$litType$'] ===
+      type) as unknown as isTemplateResultType;
```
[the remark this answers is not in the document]
AndrewJakubowicz: Done

rictic (approved):

augustjk (approved):

AndrewJakubowicz:
I added two more small cleanup commits. 1. Use the `isCompiledTemplate` helper in the cache directive fix to simplify some logic. (No behavior change). 2. Add type only tests to verify the excellent improvement in returning the exact (HTML or SVG) TemplateResult type from `isTemplateResult`.

PTAL

packages/lit-html/src/directive-helpers.ts
```diff
@@ -42,17 +48,35 @@ export const TemplateResultType = {
 export type TemplateResultType =
   (typeof TemplateResultType)[keyof typeof TemplateResultType];
 
+type isTemplateResultType = {
```
justinfagnani, line 51: Capitalize all type names, and I wouldn't include "Type" in the name. I think `IsTemplateResult` is fine.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -42,17 +48,35 @@ export const TemplateResultType = {
...
-  type?: TemplateResultType
-): value is TemplateResult =>
+  type: TemplateResultType | undefined
+) =>
   type === undefined
     ? // This property needs to remain unminified.
       (value as TemplateResult)?.['_$litType$'] !== undefined
-    : (value as TemplateResult)?.['_$litType$'] === type;
+    : (value as TemplateResult)?.['_$litType$'] ===
+      type) as isTemplateResultType;
```
justinfagnani, line 70: Is this `as isTemplateResultType` cast is in the wrong place? It reads to me like it's casting a boolean to a function call interface, but with a body-less arrow function I'm not sure of the precedence here.

To make it more clear I would type the `isTemplateResult` variable, instead of a cast, like:

```ts
export const isTemplateResult: IsTemplateResult = ((
  value: unknown,
  type: TemplateResultType | undefined
) =>
```

packages/lit-html/src/directives/cache.ts
```diff
@@ -21,13 +22,25 @@ import {
   clearPart,
   getCommittedValue,
   insertPart,
+  isCompiledTemplateResult,
   isTemplateResult,
   setCommittedValue,
 } from '../directive-helpers.js';
 
+/**
+ * @returns The template strings array for the provided template.
```
justinfagnani, line 31: I would comment the string contents are not compatible, since compiled template strings have already been prepared. They're only suitable for use as a key.

packages/lit-html/src/directives/cache.ts
```diff
@@ -21,13 +22,25 @@ import {
...
+ */
+function getStringsFromTemplate(
```
justinfagnani, line 33: Use an arrow function

packages/lit-html/src/directives/cache.ts
```diff
@@ -21,13 +22,25 @@ import {
...
+  template: TemplateResult | CompiledTemplateResult
```
justinfagnani, line 34: It's not a template, but a template result. I would name it `result`, and put "result" in the function name too.

packages/lit-html/src/directives/cache.ts
```diff
@@ -45,17 +58,23 @@ class CacheDirective extends Directive {
...
+      let cachedContainerPart = this._templateCache.get(
+        getStringsFromTemplate(this._value)
+      );
       if (cachedContainerPart === undefined) {
         const fragment = document.createDocumentFragment();
         cachedContainerPart = render(nothing, fragment);
         cachedContainerPart.setConnected(false);
-        this._templateCache.set(this._value.strings, cachedContainerPart);
+        this._templateCache.set(
+          getStringsFromTemplate(this._value),
```
justinfagnani, line 75: You call this several times. Can you put the result in a variable?

packages/lit-html/src/directive-helpers.ts
AndrewJakubowicz: Went through this with @rictic and unfortunately the cast is required. If I move the type as you've suggested it fails with: `Type '(value: unknown, type: TemplateResultType | undefined) => boolean' is not assignable to type 'IsTemplateResult'.`

packages/lit-html/src/directives/cache.ts
AndrewJakubowicz: I.e. `getStringsFromTemplateResult`?

packages/lit-html/src/directive-helpers.ts
augustjk: https://github.com/lit/lit/pull/3993#discussion_r1255004607

See the snippet here that moves the type next to the specifier. You need the 2nd param to be optional and also one of the type predicate return annotation to satisfy TS on the implementation code.

packages/lit-html/src/directives/cache.ts
AndrewJakubowicz: Something like?

```

/**
 * The template strings array contents are not compatible between the two
 * template result types as the compiled template contains a prepared string,
 * only use for cache key.
 * @returns The template strings array for the provided template result.
 */
 ```

packages/lit-html/src/directive-helpers.ts
AndrewJakubowicz: Oh - I am so sorry. It did not click that you had removed the assertion!

packages/lit-html/src/directives/cache.ts
AndrewJakubowicz: Done. Lifted the calls to the top.

packages/lit-html/src/directives/cache.ts
augustjk: @justinfagnani For my own curiosity's sake, is the main reason for preferring arrow function here for the brevity of the implicit return?

augustjk (approved):

packages/lit-html/src/directives/cache.ts
justinfagnani: A few reasons:
* Consistency
* As an indicator that this function doesn't need a bindable `this`, or rather `function` should be an indicator that a function _does_ need a bindable `this`. Similar to preferring `const` so that `let` is an indicator that a variable will be reassigned.
* It's a few less characters, and because of our consistent usage it _may_ compress better in gzip (brotli likely as a pre-baked dictionary entry for `function`, but I'd have to look that up).

justinfagnani (approved):

Issue #3997, opened 2023-07-07
[labs/task] Allow tasks to be canceled


justinfagnani:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Task (@lit-labs/task)

### Description

Tasks should be cancelable with an `.abort()` method that aborts the AbortSignal that we not pass to task functions.

### Alternatives and Workarounds

None

PR #3998, opened 2023-07-07
Add a Task.abort() method

justinfagnani:
Fixes https://github.com/lit/lit/issues/3997

This makes task cancellation possible, but manual. It's _maybe_ desirable in some cases to somehow allow passing an AbortSignal into the task and composing it with the signal we automatically create, and but the exact requirements and shape of that API aren't clear yet. Chaining tasks is one of the more obvious use cases, but this might be doable already because aborting a task will reject its `taskComplete` promise, which can be passed to downstream tasks.

### DIFF
--- a/.changeset/flat-pugs-push.md
+++ b/.changeset/flat-pugs-push.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': minor
+---
+
+Allow cancelling a task with a Task.abort() method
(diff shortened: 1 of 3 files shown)


rictic (approved):

rictic:
Oh good catch (he he) with undefined error values!

justinfagnani:
@rictic yeah, that _really_ through me for a loop for a while today. Only found it because I was aborting with `undefined` only in Safari 14 because it didn't support `signal.reason`, and the task wasn't transitioning to ERROR in that case. I have a bruise on my head from banging it 🤕

PR #4000, opened 2023-07-07
[labs/react] Update react versions used for testing and add type as peer dependency

### LINKED ISSUE #3149: [labs/react] Upgrade to React 18
We're on React 17. React 18 is out. We should be on React 18 sometime. Or be compatible with both somehow.

cc @justinfagnani

augustjk:
Fixes #3149 

The version of React we use for testing has been bumped to 18. To make this work:
1. Added a bespoke WTR config that bundles the test files with React.
This was needed as React added package exports preventing module resolution into the actual umd dist files we used to have but still only exports cjs.
This also feels more realistic and the code more idiomatic to write.
2. Rewrote the deprecated `ReactDOM.render()` calls to use `createRoot()` from `react-dom/client`.
We also reuse the render root across multiple renders within the same test.
3. Related to above, `render()` calls were no longer synchronous (likely to support React's new ability to batch multiple setState calls) so to make sure we do DOM assertions after React has finished rendering, use `act()` from `react-dom/test-utils` where needed.
https://legacy.reactjs.org/docs/testing-recipes.html#act
This also required adding a global flag up top as described here: https://react.dev/blog/2022/03/08/react-18-upgrade-guide#configuring-your-testing-environment
4. `useController` test required a bit of tweak as React 18 strict mode also reruns effects.

Additionally, we already had direct usage of typings from `@types/react` in consumer facing code like https://github.com/lit/lit/blob/012e8bc874ac1128f4c719648a3a8c9f29d43b7d/packages/labs/react/src/create-component.ts#L53-L58
But we only had `@types/react` listed in dev dependencies. This was moved as a peer dependency.

I do want to revisit dropping this all-together to use generic types that can accept both React and Preact types.


packages/labs/react/src/create-component.ts
```diff
@@ -381,14 +382,13 @@ the element.`);
...
 
-  const ForwardedComponent: ReactWebComponent<I, E> = React.forwardRef<
+  const ForwardedComponent: ReactWebComponent<I, E> = ReactModule.forwardRef<
     I,
     WebComponentProps<I, E>
   >((props, __forwardedRef) =>
     createElement<Props, ReactComponent, typeof ReactComponent>(
       ReactComponent,
-      {...props, __forwardedRef},
-      props?.children
```
augustjk, line 391: This was removed as it's unnecessary and also causing a type error.

augustjk:
This ended up need a lot of additional changes, particularly this: https://github.com/lit/lit/pull/4000/commits/b4ef3c2b31890ba8a2d4e01133eb9d4af50d4131
As the newer version of `@types/react` got hoisted up to the root of the monorepo, it ended up conflicting with `@types/truested-types`.
Since TS digs into all global declaration files by default, I need to add explicit `"types"` fields to many tsconfig. Hat tip to @AndrewJakubowicz for describing the issue and solution here https://github.com/lit/lit/pull/3894

This arguably could be its PR, but it's really only a problem because of the react types so unsure if it really should be separated out.

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -311,8 +341,8 @@ suite('createComponent', () => {
     assert.equal(el.getAttribute('id'), wrappedEl.getAttribute('id'));
     assert.equal(el.id, wrappedEl.id);
 
-    // @ts-expect-error
-    await renderReactComponent({id: null});
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    await renderReactComponent({id: null as any});
```
augustjk, line 345: I'm not sure that these should explicitly be expecting TS error which implies that we're testing that the typing of the created component should not be allowed to take `null` as the value of these HTML attributes.
My thought is we're only concerned with testing runtime behavior of what happens if `null` is somehow provided.

packages/labs/react/src/create-component.ts
AndrewJakubowicz: For my own understanding, why was it provided originally?

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -113,14 +120,15 @@ const BasicElementComponent = createComponent({
...
 ) => {
-  window.ReactDOM.render(
-    <>
-      <div {...(props as React.HTMLAttributes<HTMLDivElement>)}/>,
-      <x-foo {...props}/>
-      <BasicElementComponent {...props}/>,
-    </>,
-    container
-  );
+  act(() => {
```
AndrewJakubowicz, line 123: No action - but this was interesting to see because sometimes folk want a similar thing for Lit. We provide an `updateComplete` promise to await on, but usage of that throughout the tests ends up littering the tests. I wonder if we could provide a similar helper that essentially does a deep, `updateComplete` waiting?

packages/labs/react/src/create-component.ts
augustjk: The actual `createElement()` signature is that children should be passed in as separate arguments after props. This is how JSX transforms would do it. But in component code, `children` is provided as part of the props object. So understandably users often do the same when calling `createElement()`. The actual implementation will use the `children` provided in props, unless a 3rd argument is provided in which case it overwrites what's provided in props. https://github.com/facebook/react/blob/v18.2.0/packages/react/src/ReactElement.js#L401-L417

Interestingly this was discussed before here https://github.com/lit/lit/pull/1420#discussion_r541337739 and was agreed to remove there but this was a missed spot.

packages/labs/react/src/test/create-component_test.tsx
augustjk: Good idea. It could be a nice util to include to `@lit-labs/testing`.

packages/labs/react/src/create-component.ts
AndrewJakubowicz: Awesome – makes sense! Thank you!

packages/labs/react/web-test-runner.config.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
AndrewJakubowicz, line 3: Nit: 2023?

packages/labs/react/web-test-runner.config.js
```diff
@@ -0,0 +1,53 @@
+/**
```
AndrewJakubowicz, line 1: File level comment/question: So React is now tested separately, hence the removal from the `packages/test` directory?

augustjk: Correct. It needs a separate WTR config than the others in `packages/tests` so is run separately.

packages/labs/react/web-test-runner.config.js
augustjk: Thanks!

packages/labs/react/tsconfig.json
```diff
@@ -22,7 +22,8 @@
     "experimentalDecorators": true,
     "jsx": "react",
     "noImplicitOverride": true,
-    "types": ["mocha", "react", "react-dom"]
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 25: Why was this needed? Is there an issue with the `react`, and `react-dom` types being checked as they've been removed from the types array?

AndrewJakubowicz (approved):
Nice! Thanks for splitting out the other changes. It made this much easier to review!

packages/labs/react/tsconfig.json
augustjk: `react` and `react-dom` could be removed from `types` as we're not relying on them being available as UMD globals.
`skipLibCheck` was needed because some React internal typings were causing errors, possibly incompatible with our other tsconfig options.

PR #4001, opened 2023-07-08
[labs/task] Allow tasks to have an initial value

### LINKED ISSUE #2367: [labs/task] Allow setting initial data

justinfagnani:
Fixes #2367 
Supersedes #2836 

This adds an `initialValue` config option. If set, the task is put into the COMPLETE status and `initialValue` is assigned to the current value of the task. Initial args are read off synchronously so that the task won't auto-run until the args change.

This is useful for when an element has the initial data available synchronously (maybe from hydration?) and wants to make the task look like it has completed a first run.

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -49,6 +49,16 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
   task: TaskFunction<T, R>;
   args?: ArgsFunction<T>;
   autoRun?: boolean;
+
+  /**
+   * If initialValue is provided, the task is initialized to the COMPLETE
+   * status and the value is set to initialData.
+   *
+   * Initial args should be coherent with the initialValue, since they are
+   * assumed to be the args that would produce that value. When autoRun is
+   * `true` the task will not auto-run again until the args change.
+   */
+  initialValue?: R;
   onComplete?: (value: R) => unknown;
   onError?: (error: unknown) => unknown;
 }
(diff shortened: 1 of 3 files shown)


packages/labs/task/src/task.ts
```diff
@@ -185,6 +195,13 @@ export class Task<
     if (taskConfig.autoRun !== undefined) {
       this.autoRun = taskConfig.autoRun;
     }
+    // Providing initialValue puts the task in COMPLETE state and stores the
+    // args immediately so it only runs when they change again.
+    if ('initialValue' in taskConfig) {
+      this._value = taskConfig.initialValue;
+      this.status = TaskStatus.COMPLETE;
+      this._previousArgs = this._getArgs?.();
+    }
```
rictic, line 204: Beautiful!

packages/labs/task/src/test/task_test.ts
```diff
@@ -225,6 +225,36 @@ suite('Task', () => {
...
+    initializing = false;
+    await renderElement(el);
+    assert.equal(el.task.status, TaskStatus.COMPLETE);
+    assert.equal(el.task.value, 'initial');
+    assert.equal(el.taskValue, 'initial');
+
+    // The task is complete, so waiting for it shouldn't change anything
+    await tasksUpdateComplete();
+    assert.equal(el.task.status, TaskStatus.COMPLETE);
+    assert.equal(el.task.value, 'initial');
```
rictic, line 246: Consider asserting that assigning `el.a = 'a';` won't run the task

rictic (approved):

packages/labs/task/src/task.ts
justinfagnani: thanks!

Issue #4003, opened 2023-07-08
[labs/task] Tasks should run in hostUpdate(), not hostUpdate()


justinfagnani:
### Which package(s) are affected?

Task (@lit-labs/task)

### Description

Tasks need to update their host when run, so that the host can render the new pending state of the task.

If tasks are started in `hostUpdated()`, then a `requestUpdate()` call will trigger a double-render and associated warning. To work around the warning Task currently calls `requestUpdate()` in a microtask, but there is still a double-render.

We should instead run tasks in `hostUpdate()` and call `requestUpdate()` normally. The main observable difference with this timing is that host changes in `update()` are not visible to the Task in the current host update pass. State changes that should be visible to a task must be done in `willUpdate` which is run before `hostUpdate()`.

### Reproduction

n/a

### Workaround

n/a

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

all

### Browser/OS/Node environment

all

Issue #1808, opened 2021-04-24
[labs/task] Improve README.md usage example


blackgwe:
# Description

It's hard to understand the example in `packages/labs/task/README.md`, there are some syntax errors (back-ticks, comma …) 

Maybe it's possible to add a working demo.

# Proposal for the README.md file

```ts
import {Task, TaskStatus} from '@lit-labs/task';
// ...

export class MyElement extends LitElement {

  @state()
  private _userId: number = 7;
  
  private _apiTask = new Task(
      this, 
      ([dependencyParam1]) => fetch(`//jsonplaceholder.typicode.com/users/${dependencyParam1}`)
          .then(response => response.json()),
      () => [this._userId]   // dependencies getter -> returns array of dependencyParameters
    );

  render() {
     return html`
         <div>user</div>
         ${this._apiTask.render({ 
            pending: () => 'loading', 
            complete: (user) => `${user.name}` 
         })}`;
  }
```

# Working Example

```ts
import {html, LitElement} from 'lit';
import {customElement, state} from 'lit/decorators.js';
import {Task, TaskStatus} from '@lit-labs/task';

@customElement('simple-greeting')
export class SimpleGreeting extends LitElement {

  @state()
  private _userId: number = 7;
  
  private _apiTask = new Task(
      this, 
      ([dependencyParam1]) => fetch(`//jsonplaceholder.typicode.com/users/${dependencyParam1}`)
          .then(response => response.json()),
      () => [this._userId] // dependencies getter -> returns array of dependencyParameters
    );

  render() {
    const pending = () => html`Loading user info...`;
    const complete = (user) => html`${user.name}`;
    return html`<p>Hello ${this._apiTask.render({ pending, complete })}</p>`;
  }
  
  // correct syntax
  // render() {
  //   return html`
  //       <div>user</div>
  //       ${this._apiTask.render({ 
  //          pending: () => 'loading', 
  //          complete: (user) => `${user.name}` 
  //       })}`;
  }

}
```

jkanche:
I was here just to report the exact same thing and what worked for me!

Issue #1499, opened 2020-12-17
[labs/task] Refine API of Task


sorvell:
## Description
There are some API rough edges in AsyncTask that may change. Noting them here for discussion and so that users are aware.

- [x] Currently the `value` and `error` properties are cleared when a task becomes pending. It might be useful to keep these around until the task completes. Additionally, if a task errors, perhaps the previous value should be retained?
- [ ] It would be nice to be able to abort a task. The platform provides an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for this which could be integrated into `AsyncTask`.
- [x] There's not a good way to be informed when a task completes. There is an update, but if you want code external to the update to know, it must be done manually, for example by vending a promise for the task completion.
- [ ] Tasks are triggered in `updated`. This means that an element must render 2x to see its pending state (meaning a common pattern to set a property that's a dep and await updateComplete will not render a pending state), and it also means there's currently a dev mode warning when using `AsyncTask`.

## Acceptance criteria
Refine API and land changes.

sorvell:
Per suggestion from @kevinpschaaf: to avoid timing issues with changing deps and the element lifecycle, it may make sense to alternatively or additively be able to trigger the task (here via `update`), e.g.

```
render() {
  const dep1 = computeDep1(this.value1, this.value2);
  const dep2 = computeDep2(this.value3, this.value4);
  this.asyncTask.update([dep1, dep2]);
  return html`
    <div>${this.asyncTask.render({
      pending() { return html`please wait...`
      complete(value) { return html`${value}`}
    });
  `
}
```

justinfagnani:
All of these issues have been addressed except whether to run the task from `hostUpdated` or `hostUpdate`.

justinfagnani:
I opened https://github.com/lit/lit/issues/4003 to track timing. Closing this issue now.

PR #4004, opened 2023-07-08
[labs/task] Run tasks in update instead of updated

justinfagnani:
Fixes https://github.com/lit/lit/issues/4003
Fixes https://github.com/lit/lit/issues/1808 (there were some issues in the jsdoc example)

This is a breaking change.

This also adds `'afterUpdate'` as a valid value for `autoRun`, which preserves the existing timing. That's useful for tasks that need to read state changes in update, such as DOM.

### DIFF
--- a/.changeset/chilled-jeans-march.md
+++ b/.changeset/chilled-jeans-march.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': major
+---
+
+Run tasks in update instead of updated
--- a/.changeset/chilled-jeans-march.md
+++ b/.changeset/chilled-jeans-march.md
@@ -2,4 +2,4 @@
 '@lit-labs/task': major
 ---
 
-Adds the `'afterUpdate'` option for `autoRun` to Task, and runs tasks by default in `hostUpdate()` instead of `hostUpdated()`.
+Adds the `'afterUpdate'` option for `autoRun` to Task, and runs tasks by default in `hostUpdate()` instead of `hostUpdated()`. `'afterUpdate'` is needed to run tasks dependent on DOM updates, but will cause multiple renders of the host element.
--- a/.changeset/chilled-jeans-march.md
+++ b/.changeset/chilled-jeans-march.md
@@ -2,4 +2,4 @@
 '@lit-labs/task': major
 ---
 
-Run tasks in update instead of updated
+Adds the `'afterUpdate'` option for `autoRun` to Task, and runs tasks by default in `hostUpdate()` instead of `hostUpdated()`.
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -176,22 +176,7 @@ export class Task<
    * Determines if the task is run automatically when arguments change after a
    * host update. Default to `true`.
    *
-   * If `true`, the task checks arguments during the host update (after
-   * `willUpdate()` and before `update()` in Lit) and runs if they change. For
-   * a task to see argument changes they must be done in `willUpdate()` or
-   * earlier. The host element can see task status changes caused by its own
-   * current update.
-   *
-   * If `'afterUpdate'`, the task checks arguments and runs _after_ the host
-   * update. This means that the task can see host changes done in update, such
-   * as rendered DOM. The host element can not see task status changes caused
-   * by its own update, so the task must trigger a second host update to make
-   * those changes renderable.
-   *
-   * Note: `'afterUpdate'` is unlikely to be SSR compatible in the future.
-   *
-   * If `false`, the task is not run automatically, and must be run with the
-   * {@linkcode run} method.
+   * @see {@link TaskConfig.autoRun} for more information.
    */
   autoRun: boolean | 'afterUpdate';
 
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -66,6 +66,9 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
    * those changes renderable.
    *
    * Note: `'afterUpdate'` is unlikely to be SSR compatible in the future.
+   *
+   * If `false`, the task is not run automatically, and must be run with the
+   * {@linkcode run} method.
    */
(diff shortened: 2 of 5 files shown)


packages/labs/task/src/task.ts
```diff
@@ -63,57 +63,77 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
...
-// warning that the update was triggered in response to an update.
-// 2. And as a result of triggering in `updated`, if the user waits for the
-// `updateComplete` promise they will not see a `pending` state since this
-// will be triggered in another update; they would need to
-// `while (!(await el.updateComplete));`.
-// 3. If this is instead or additionally triggered in `willUpdate`, the
-// warning goes away in the common case that the update itself does not change
-// the deps; however, the `requestUpdate` to render pending state  will not
+// TODO(sorvell / justinfagnani): Some issues:
+// 1. With the task triggered in `update`, there is no ReactiveElement
```
justinfagnani, line 67: I'm leaving this comment partially here so that we investigate it more. These may be non-issues.

e111077:
This is a difficult choice. Updated we can be pretty sure all computed properties should be synced by then. Also we know DOM will always be ready in the task. Can you spin up an example real quick with a task that accesses a @query and see how it looks DX-wise?

justinfagnani:
@e111077 yeah, I can add that as a test. A task will not be able to see rendered updates though. I think we need to find a case where something doesn't work at all, not just that we get a double-render, and let that guide us. If someone wanted to trigger a task based on updated shadow root contents... is that a supported case?

I thought we kind of resolved to do this a while back, based on this comment: https://github.com/lit/lit/issues/1499#issuecomment-1097287550

But... do we need this to be configurable? `runAfterUpdate: true`?

e111077:
Personally don't care about a test case too much just want to investigate the DX. Like does the queried property go in the arguments list now? It won't trigger a re-render when it updates but it will trigger dirty props etc.

justinfagnani:
I mean to add the test case as a reviewed, checked-in, working example of the DX, with some assertions about whether it works. I don't _think_ this will be a good case with the new timing, so I want to get it in and have eyes on it.

justinfagnani:
I added the test/example (didn't push yet) and... I don't know. It might be better to take the double-render hit on pending tasks.

For many cases, and especially SSR compatibility, update timing is better. So I wonder if we should have an option, and what the default should be? Is depending on DOM common? Could tasks using updated timing skip SSR?

e111077:
If we keep it as so we push people towards the data system rather than accessing DOM directly which is arguably "idiomatic Lit".

So someone in updated or firstUpdated can access the DOM, update a reactive property dependency in the Task props array, and then the rerender on update can catch it. The Task may run twice (perhaps noop the first time) but the element in both cases would lifecycle twice. But in the cases where state doesn't rely on DOM (a fully SSR-able component) then there is no double render.

Idk drawbacks either way, but I'd argue the DOM case is not as common but not uncommon

justinfagnani:
What about:

```ts
  autoRun?: boolean | 'afterUpdate'
```

The default timing would be `hostUpdate` and afterUpdate would revert to the current timing. afterUpdate probably wouldn't be SSR compatible.

justinfagnani:
@e111077 I added a `autoRun: 'afterUpdate'` option, and I think it might be a good solution. I pushed the change and tests that show it working with DOM.

(I can't for the life of me figure out why the task tests can't see the `lit` package though)

e111077:
Also wondering why not willUpdate? That would allow fetches to be completed on the server and send pure HTML to the user... Oh I guess reactive controllers don't have hostWillUpdate.

I think letting the user select the lifecycle like this is a decent idea. Would be cool if we could enable server fetches though

justinfagnani:
Yeah, we don't have `hostWillUpdate()` and if we did it would go before `willUpdate()` as the other callbacks do, which I don't think would be very useful.

I think we _can_ enable server fetches - easier - with the new default timing. @kevinpschaaf had a branch that made SSR wait for async controllers, and this timing lets us know sooner that a task will have been started.

packages/labs/task/src/task.ts
```diff
@@ -134,7 +176,7 @@ export class Task<
   /**
    * Controls if they task will run when its arguments change. Defaults to true.
```
rictic, line 177: typo:

if they task -> if the task

rictic, line 177: Mention what `'afterUpdate'` means. Who should use it and when? Is it a temporary migration aid?

justinfagnani: That's in the options description. I'll copy down.

e111077 (approved):
LGTM pending Peter's suggestions

packages/labs/task/src/task.ts
```diff
@@ -132,9 +174,27 @@ export class Task<
...
+   * If `'afterUpdate'`, the task checks arguments and runs _after_ the host
+   * update. This means that the task can see host changes done in update, such
+   * as rendered DOM. The host element can not see task status changes caused
+   * by its own update, so the task must trigger a second host update to make
+   * those changes renderable.
+   *
+   * Note: `'afterUpdate'` is unlikely to be SSR compatible in the future.
+   *
+   * If `false`, the task is not run automatically, and must be run with the
+   * {@linkcode run} method.
```
rictic, line 195: Ah, maybe just suggest looking at TaskConfig['autorun'] instead of duplicating the entire doc, don't know how I missed that before!

.changeset/chilled-jeans-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/task': major
+---
+
+Adds the `'afterUpdate'` option for `autoRun` to Task, and runs tasks by default in `hostUpdate()` instead of `hostUpdated()`.
```
rictic, line 5: Is it the case that users that are broken can run with `afterUpdate` timing to get the old behavior, but that they'll get better performance with the new timing? If so, I think this is a good thing to call out in the changelog

rictic (approved):

PR #4008, opened 2023-07-10
[labs/task] Infer the return type of Task.render()

justinfagnani:
Fixes https://github.com/lit/lit/issues/2301

Supersedes https://github.com/lit/lit/pull/2821

### DIFF
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -327,19 +327,20 @@ export class Task<
     return this._error;
   }
 
-  render(renderer: StatusRenderer<R>) {
+  render<T extends StatusRenderer<R>>(renderer: T) {
     switch (this.status) {
       case TaskStatus.INITIAL:
-        return renderer.initial?.();
+        return renderer.initial?.() as MaybeReturnType<T['initial']>;
       case TaskStatus.PENDING:
-        return renderer.pending?.();
+        return renderer.pending?.() as MaybeReturnType<T['pending']>;
       case TaskStatus.COMPLETE:
-        return renderer.complete?.(this.value!);
+        return renderer.complete?.(this.value!) as MaybeReturnType<
+          T['complete']
+        >;
       case TaskStatus.ERROR:
-        return renderer.error?.(this.error);
+        return renderer.error?.(this.error) as MaybeReturnType<T['error']>;
       default:
-        // exhaustiveness check
-        this.status as void;
+        throw new Error(`Unexpected status: ${this.status}`);
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/labs/task/src/test/task_test.ts
```diff
@@ -942,4 +942,19 @@ suite('Task', () => {
...
+    }
+    customElements.define(generateElementName(), TestElement);
+    const el = new TestElement();
+
+    const acceptNumberOrUndefined = (x?: number) => x;
+
+    acceptNumberOrUndefined(el.task.render({initial: () => 123}));
+    acceptNumberOrUndefined(el.task.render({complete: () => 123}));
+    acceptNumberOrUndefined(el.task.render({pending: () => 123}));
+    acceptNumberOrUndefined(el.task.render({error: () => 123}));
```
rictic, line 958: let's add a test for when two callbacks are passed (to ensure it works when there are multiple), and for when all four are passed

also, when all four are passed, I think we can infer the return type as `number`

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAJwKZgCauQHgCqKoAeU6GAzogN4A2qwUA-ABQCUAXIuANZhwDuYADQoYAcwAWTNpx59BAXwB8zAG6c8ragChEiGMETMAsgEMoEgHTJTmOAFs2Wqrr0pUUEMiSrr4qYyWbIimlGYAngBGqABKHl5geOEADqj4ANoA5Mj+UJkAukoA3K4Krmie3oi+dAyBwaGIEdFxlYkpaXhZtXmFJWXaUB1NplGx8d5JqTgAYkqIALyIM4QkZJTMllumyGLknLbhrAtKMGDA2IgxiIxXiLKY9GeoGCXaEAjkUNWL7o-IzFo9CgnGCJ0QAEYAAwiHKSEFGLTggBMUIUrBKAHpMW5EAA9RjabTEZJwZDfKgKIA

packages/labs/task/src/task.ts
```diff
@@ -351,3 +352,8 @@ export class Task<
       : args !== prev;
   }
 }
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+type MaybeReturnType<T extends undefined | ((...args: any) => any)> =
```
rictic, line 357: More self contained, potentially easier on the compiler, and can avoid using `any` by writing:

```ts
type MaybeReturnType<F> = F extends (...args: unknown[])=>infer R ? R : undefined;
```

justinfagnani: thanks!

packages/labs/task/src/task.ts
```diff
@@ -10,8 +10,7 @@ export interface TaskFunctionOptions {
   signal?: AbortSignal;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type TaskFunction<D extends ReadonlyArray<unknown>, R = any> = (
+export type TaskFunction<D extends ReadonlyArray<unknown>, R = unknown> = (
```
justinfagnani, line 13: I tacked this on to this PR instead of making a 7-char PR. It's technically breaking, but I don't see why we should default to `any` in the cases TS can't infer the task type. `unknown` is much safer.

rictic (approved):

PR #4013, opened 2023-07-11
[labs/task] Add pluggable task args equality functions

justinfagnani:
Fixes: https://github.com/lit/lit/issues/3142

- Adds argsEqual config option
- Removes `shouldRun()` and `performTask()` protected methods
- Adds `shallowArrayEquals`, `deepArrayEquals`, and `deepEquals` functions

The `deepEquals()` is inspired by fast-deep-equals, but rewritten to use modern JS, `Object.is()`, and have a different ordering of cases that should better match expected data (Dates would be more likely than RegExps). I did not yet include Map and Set comparison. I didn't find a nice deep equals package to depend on that fit our needs.

Talking to @rictic I think we may also want to generalize args to just "data" and not assume it's array, but this can be a step towards that with the array-assuming equality functions.


packages/labs/task/src/task.ts
```diff
@@ -400,16 +423,93 @@ export class Task<
...
+    // ms since epoch: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf
+    if (a.valueOf !== objectValueOf) {
+      return a.valueOf() === b.valueOf();
+    }
+
+    // Defer to custom toString implementations. This should handle
+    // TrustedTypes, URLs, and such. This might be a bit risky, but
+    // fast-deep-equals does it.
+    if (a.toString !== objectToString) {
+      return a.toString() === b.toString();
```
rictic, line 487: What about Map, Set?

packages/labs/task/src/task.ts
```diff
@@ -400,16 +423,93 @@ export class Task<
...
+    oldArgs.every((v, i) => deepEquals(v, newArgs[i])));
+
+const objectValueOf = Object.prototype.valueOf;
+const objectToString = Object.prototype.toString;
+const {keys: objectKeys} = Object;
+const {isArray} = Array;
+
+export const deepEquals = (a: unknown, b: unknown): boolean => {
+  if (Object.is(a, b)) {
+    return true;
```
rictic, line 455: Do we want NaN to match NaN?

packages/labs/task/src/test/task_test.ts
```diff
@@ -777,24 +808,24 @@ suite('Task', () => {
...
-      await el.task.performTask();
-      taskComplete = true;
-    })();
-    await tasksUpdateComplete();
-    assert.isFalse(taskComplete);
-    el.resolveTask();
-    await tasksUpdateComplete();
-    assert.isTrue(taskComplete);
-  });
+  // test('performTask waits on the task', async () => {
```
rictic, line 811: Is this intended?

packages/labs/task/src/test/task_test.ts
```diff
@@ -225,6 +226,36 @@ suite('Task', () => {
     assert.equal(el.taskValue, `a1,b1`);
   });
 
+  test('tasks with args run when args change: deepEqual', async () => {
+    const el = getTestElement({
+      args: () => [[el.a], [el.b]],
```
rictic, line 231: Recommend adding a comment to note the inner arrays, I almost missed that

packages/labs/task/src/task.ts
```diff
@@ -400,16 +423,93 @@ export class Task<
...
+  oldArgs === newArgs ||
+  (oldArgs.length === newArgs.length &&
+    oldArgs.every((v, i) => deepEquals(v, newArgs[i])));
+
+const objectValueOf = Object.prototype.valueOf;
+const objectToString = Object.prototype.toString;
+const {keys: objectKeys} = Object;
+const {isArray} = Array;
+
+export const deepEquals = (a: unknown, b: unknown): boolean => {
```
rictic, line 453: Deep equality checks require a number of opinionated pieces. Do we want to maintain this? Do we want to ship these bytes, or are we assuming production builds are tree shaked?

packages/labs/task/src/task.ts
justinfagnani: I think so. I also think matching some built-in semantics like Object.is is good. That's why I inverted the -0/0 check.

packages/labs/task/src/test/task_test.ts
justinfagnani: I'll remove. We should have coverage that various ways to run the task already wait for the task to complete.

packages/labs/task/src/test/task_test.ts
justinfagnani: will do

packages/labs/task/src/task.ts
justinfagnani: Yes they do, but... it's quite hard to find a minimally good deep equals, especially that's shipped as JS modules. I think including some basic one is just good DX.

I do need to document the opinions here:
- Object.is for primitives
- Constructor check
- Arrays
- `valueOf()` and `toString()` checks
- Map and Set (whichever way we go)
- Own property requirement for objects
- Cycles (currently we don't detect them and just iloop, which a lot of libraries do, but some don't)

packages/labs/task/src/task.ts
```diff
@@ -247,34 +253,47 @@ export class Task<
...
-   * host update. A task should run when its arguments change from the
-   * previous run.
+   * host update, and runs the task if it should.
    *
-   * Note: this is not checked when `run` is explicitly called.
+   * A task should run when its arguments change from the previous run, based on
+   * the args equality function.
    *
-   * @param args The task's arguments
+   * This method is side-effectful: it stored the new args as the previous args.
```
justinfagnani, line 284:
```suggestion
   * This method is side-effectful: it stores the new args as the previous args.
```

packages/labs/task/src/task.ts
justinfagnani: I can add those. I don't think it's actually too many bytes.

packages/labs/task/src/test/deep-equals_test.ts
```diff
@@ -0,0 +1,437 @@
+import {assert} from '@esm-bundle/chai';
+import {deepEquals} from '../task.js';
+
```
justinfagnani, line 3: Add copyright notice from fast-deep-equals

packages/labs/task/src/test/task_test.ts
```diff
@@ -225,6 +226,36 @@ suite('Task', () => {
...
+    await renderElement(el);
+    el.resolveTask();
+    await tasksUpdateComplete();
+    assert.equal(el.task.status, TaskStatus.COMPLETE);
+    assert.equal(el.taskValue, `a,b`);
+
+    // Changing task argument runs task
+    el.a = 'a1';
+    // Check task pending.
+    await tasksUpdateComplete();
```
justinfagnani, line 243: TODO: the name of this function is confusing. It _doesn't_ wait for the task to complete. It just waits a rAF. That's why it can be used to check that a task is pending.

packages/labs/task/src/task.ts
rictic: WDYT about putting this in its own module?

packages/labs/task/src/test/task_test.ts
```diff
@@ -107,6 +107,9 @@ suite('Task', () => {
     return new A();
   };
 
+  // TODO: the name of this function is confusing. It doesn't wait for the task
+  // to complete. It just waits a rAF. That's why it can be used to check that
+  // a task is pending.
   const tasksUpdateComplete = nextFrame;
```
rictic, line 113: Good point. I'd consider dropping this in favor of just directly using `nextFrame`

rictic (approved):

justinfagnani:
I'm a little stuck on the tests right now. Safari can't seem to load the new deep-equals test module, I would guess because there's some language feature or eagerly called DOM API that Safari 11 doesn't support yet. But I ran both the dev and prod tests in Safari 11 in a live Sauce session and they all load and pass. The GitHub actions logs are giving nothing more than a "failed to load module" error. No browser logs that say what the specific error is.

Issue #2763, opened 2022-04-21
[labs/context] Investigate late-provider/early-provider issues


benjamind:
## Description
As outlined [in this comment](https://github.com/lit/lit/pull/1955#discussion_r840126503) it is possible for a context provided by a late hydrating component to never be provided to its children because an earlier provider at a location higher in the DOM already provided it.

Not sure if this situation is actually solveable without tracking all context requests ever which has memory usage implications.

justinfagnani:
I don't really see how we can do anything here.

I think that if a middle provider was late-upgraded and needed to provide a value, that it's possible that the initial state should _not_ be that the higher provider provided the value, instead in an ideal world the middle element would have been automatically upgraded to provide the value when it was requested.

That requires knowing what element could provide a value _before_ upgrade, which is impossible without a central coordinator and serializing which context keys the element could provide. This is doable - Google's internal framework Wiz does something similar - but requires a bunch of conventions and constraints.

I think the best thing we do here is say that it should be _possible_ for a central coordinator to plug-in to the page to give this kind of behavior. Possibly it could do so with capturing event handlers that override all providers and delegate to some loading system. Maybe elements have eagerly loaded stub classes that handle the context protocol and load the rest of the class when needed.

As long as these kind of add-on behaviors are possible, I think we're ok with not having every last use case covered in the core package.

benjamind:
Agreed. I'd want to see a bunch of use case examples for this behavior before any real effort was put into this as part of the core package. I'm not yet convinced it's a real need that isn't easily worked around with other patterns.

PR #4014, opened 2023-07-13
[labs/context] Late loaded elements can take over the context subscriptions for their children

rictic:
When context-provider event is dispatched, and `event.canTakeOverSubscriptions` is true, then parent providers of the same context can create or push to the `event.movedSubscriptions` array to indicate callbacks that they will no longer handle, but that the new provider should.

Fixes https://github.com/lit/lit/issues/2763

### DIFF
--- a/packages/labs/context/src/lib/controllers/context-provider.ts
+++ b/packages/labs/context/src/lib/controllers/context-provider.ts
@@ -6,7 +6,7 @@
 
 import {ContextRequestEvent} from '../context-request-event.js';
 import {Context, ContextType} from '../create-context.js';
-import {ValueNotifier} from '../value-notifier.js';
+import {MovedSubscription, ValueNotifier} from '../value-notifier.js';
 import {ReactiveController, ReactiveElement} from 'lit';
 
 declare global {
@@ -29,7 +29,28 @@ export class ContextProviderEvent<
   public constructor(public readonly context: C) {
     super('context-provider', {bubbles: true, composed: true});
   }
+
+  /**
+   * Whether the element that dispatched this event can take over subscriptions
+   * from an ancestor provider.
+   */
+  declare canTakeOverSubscriptions: boolean;
+
+  /**
+   * Ongoing subscriptions that one of our ancestors was handling that are now
+   * our responsibility.
+   */
+  movedSubscriptions: undefined | MovedSubscription<ContextType<C>>[] =
+    undefined;
 }
+// True for all ContextProviderEvents that we fire, but we need to set this
+// so that ancestor providers can tell whether the provider that dispatched this
+// event will actually take over any subscriptions that they move to
+// `event.movedSubscriptions`.
+// This way we fail gracefully in the case where there's version drift, or
+// another implementor that of the context protocol that also dispatches
+// `context-provider` events but doesn't handle movedSubscriptions.
+ContextProviderEvent.prototype.canTakeOverSubscriptions = true;
 
 export interface Options<C extends Context<unknown, unknown>> {
   context: C;
@@ -83,19 +104,52 @@ export class ContextProvider<T extends Context<unknown, unknown>>
     // The check on composedPath (as opposed to ev.target) is to cover cases
     // where the consumer is in the shadowDom of the provider (in which case,
     // event.target === this.host because of event retargeting).
-    if (ev.context !== this.context || ev.composedPath()[0] === this.host) {
+    const consumerHost = ev.composedPath()[0] as Element;
+    if (ev.context !== this.context || consumerHost === this.host) {
       return;
     }
     ev.stopPropagation();
-    this.addCallback(ev.callback, ev.subscribe);
+    this.addCallback(ev.callback, ev.subscribe, consumerHost);
+  };
+
+  public onProviderRequest = (
+    ev: ContextProviderEvent<Context<unknown, unknown>>
+  ): void => {
+    // Ignore events when the context doesn't match.
+    // Also, in case an element is a consumer AND a provider
+    // of the same context, we want to avoid the element to self-register.
+    // The check on composedPath (as opposed to ev.target) is to cover cases
+    // where the consumer is in the shadowDom of the provider (in which case,
+    // event.target === this.host because of event retargeting).
+    const childProviderHost = ev.composedPath()[0] as Element;
+    if (
+      !ev.canTakeOverSubscriptions ||
+      ev.context !== this.context ||
+      childProviderHost === this.host
+    ) {
+      return;
+    }
+    const ourEv = ev as ContextProviderEvent<T>;
+    const movedSubscriptions = this.moveSubscriptionsFor(childProviderHost);
+    if (movedSubscriptions === undefined) {
+      return;
+    }
+    (ourEv.movedSubscriptions ??= []).push(...movedSubscriptions);
   };
 
   private attachListeners() {
     this.host.addEventListener('context-request', this.onContextRequest);
+    this.host.addEventListener('context-provider', this.onProviderRequest);
   }
 
   hostConnected(): void {
     // emit an event to signal a provider is available for this context
-    this.host.dispatchEvent(new ContextProviderEvent(this.context));
+    const providerEvent = new ContextProviderEvent(this.context);
+    this.host.dispatchEvent(providerEvent);
+    if (providerEvent.movedSubscriptions !== undefined) {
+      for (const {callback, consumerHost} of providerEvent.movedSubscriptions) {
+        this.addCallback(callback, true, consumerHost);
+      }
+    }
   }
 }
--- a/packages/labs/context/src/lib/controllers/context-provider.ts
+++ b/packages/labs/context/src/lib/controllers/context-provider.ts
@@ -62,8 +62,7 @@ export class ContextProvider<T extends Context<unknown, unknown>>
     super(
       (contextOrOptions as Options<T>).context !== undefined
         ? (contextOrOptions as Options<T>).initialValue
-        : initialValue,
-      host.id
+        : initialValue
     );
     this.host = host;
     if ((contextOrOptions as Options<T>).context !== undefined) {
--- a/packages/labs/context/src/lib/controllers/context-provider.ts
+++ b/packages/labs/context/src/lib/controllers/context-provider.ts
@@ -91,6 +91,12 @@ export class ContextProvider<T extends Context<unknown, unknown>>
     this.addCallback(ev.callback, ev.subscribe, consumerHost);
   };
 
+  /**
+   * When we get a provider request event, that means a child of this element
+   * has just woken up. If it's a provider of our context, then we may need to
+   * re-parent our subscriptions, because is a more specific provider than us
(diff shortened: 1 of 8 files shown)


packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -29,7 +29,28 @@ export class ContextProviderEvent<
...
+   */
+  movedSubscriptions: undefined | MovedSubscription<ContextType<C>>[] =
+    undefined;
 }
+// True for all ContextProviderEvents that we fire, but we need to set this
+// so that ancestor providers can tell whether the provider that dispatched this
+// event will actually take over any subscriptions that they move to
+// `event.movedSubscriptions`.
+// This way we fail gracefully in the case where there's version drift, or
+// another implementor that of the context protocol that also dispatches
```
AndrewJakubowicz, line 51: Nit: reword `another implementor that of the`. I think there is an erroneous `that`.

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -83,19 +104,52 @@ export class ContextProvider<T extends Context<unknown, unknown>>
...
-    this.addCallback(ev.callback, ev.subscribe);
+    this.addCallback(ev.callback, ev.subscribe, consumerHost);
+  };
+
+  public onProviderRequest = (
+    ev: ContextProviderEvent<Context<unknown, unknown>>
+  ): void => {
+    // Ignore events when the context doesn't match.
+    // Also, in case an element is a consumer AND a provider
+    // of the same context, we want to avoid the element to self-register.
```
AndrewJakubowicz, line 120: Nit: `avoid the element to self-register` -> `avoid the element self-registering`

No action required.

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -83,19 +104,52 @@ export class ContextProvider<T extends Context<unknown, unknown>>
...
     this.host.addEventListener('context-request', this.onContextRequest);
+    this.host.addEventListener('context-provider', this.onProviderRequest);
   }
 
   hostConnected(): void {
     // emit an event to signal a provider is available for this context
-    this.host.dispatchEvent(new ContextProviderEvent(this.context));
+    const providerEvent = new ContextProviderEvent(this.context);
+    this.host.dispatchEvent(providerEvent);
+    if (providerEvent.movedSubscriptions !== undefined) {
```
AndrewJakubowicz, line 149: Nit & no action required: Is it too much to add a comment here that `movedSubscriptions` may be populated by an ancestor provider & they may need to be moved to this provider?

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -45,26 +55,65 @@ export class ValueNotifier<T> {
...
     }
   }
 
   clearCallbacks(): void {
-    this.disposers.clear();
+    this.callbacks.clear();
+  }
+
+  /**
+   * Handle a late registration of an provider in between us and any consumers
```
AndrewJakubowicz, line 89:
```suggestion
   * Handle a late registration of a provider in between us and any consumers
```

AndrewJakubowicz (commented):
Ran out of time to complete review – leaving the couple nits and will pick up in ~ 1 hour.

It looks great!

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -11,6 +11,16 @@ import {ContextCallback} from './context-request-event.js';
  */
 type Disposer = () => void;
 
+interface CallbackInfo {
+  disposer: Disposer;
+  consumerHost?: Element;
+}
+
+export interface MovedSubscription<T> {
```
AndrewJakubowicz, line 19: Nit: Could this exported interface be documented?

packages/labs/context/src/lib/value-notifier.ts
```diff
@@ -45,26 +55,65 @@ export class ValueNotifier<T> {
...
+    for (const [callback, {disposer}] of this.callbacks) {
       callback(this._value, disposer);
     }
   };
 
-  addCallback(callback: ContextCallback<T>, subscribe?: boolean): void {
+  addCallback(
+    callback: ContextCallback<T>,
+    subscribe?: boolean,
+    consumerHost?: Element
```
AndrewJakubowicz, line 66: Looking through this change – won't `consumerHost` always be an Element? Can the type signature be updated to reflect that?

packages/labs/context/src/test/late-provider_test.ts
```diff
@@ -300,4 +301,81 @@ suiteSkipIE('late context provider', () => {
...
+
+    @customElement('late-context-provider-4')
+    class LateContextProvider4Element extends LitElement {
+      provide = new ContextProvider(this, {
+        context,
+        initialValue: 'late provider initial value',
+      });
+    }
+    await directChildConsumer.updateComplete;
+    // bad!
```
AndrewJakubowicz, line 368: I'm confused, why is this bad? (Same as the line beneath). These assertions seem to be the correct behavior that I would expect. The `indirectChildConsumer` starts listening to the `late-context-provider`, and the `directChildConsumer` to the `grandparentProvider`. 

It might be nice to add one more test which is that `setValue` on the grandparent provider will only update the `directChildConsumer`.

packages/labs/context/src/lib/controllers/context-provider.ts
rictic: Removed the code in question

packages/labs/context/src/lib/controllers/context-provider.ts
rictic: Removed the code in question

packages/labs/context/src/lib/value-notifier.ts
rictic: That would make this a breaking change. Maybe we want that though?

packages/labs/context/src/lib/value-notifier.ts
rictic: Removed

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -83,19 +83,50 @@ export class ContextProvider<T extends Context<unknown, unknown>>
...
+    const consumerHost = ev.composedPath()[0] as Element;
+    if (ev.context !== this.context || consumerHost === this.host) {
       return;
     }
     ev.stopPropagation();
-    this.addCallback(ev.callback, ev.subscribe);
+    this.addCallback(ev.callback, ev.subscribe, consumerHost);
+  };
+
+  public onProviderRequest = (
```
AndrewJakubowicz, line 94: really minor nit: do we want explicit `public` keyword?

AndrewJakubowicz: non blocking nit: Worth adding comments on the intent of this handler? Maybe write a little about what you explained to me in person.

packages/labs/context/src/lib/controllers/context-provider.ts
```diff
@@ -83,19 +83,50 @@ export class ContextProvider<T extends Context<unknown, unknown>>
...
+    const providerEvent = new ContextProviderEvent(this.context);
+    this.host.dispatchEvent(providerEvent);
+  }
+
+  private reparentSubscriptions(_childProviderHost: Element) {
+    for (const [callback, {consumerHost}] of this.callbacks) {
+      if (consumerHost === undefined) {
+        continue;
+      }
+      consumerHost.dispatchEvent(
```
AndrewJakubowicz, line 127: Nit: add a comment for the why of this logic. Dispatching this event will ensure that the context will get the nearest provider.

packages/labs/context/src/lib/value-notifier.ts
AndrewJakubowicz: pre-graduation is a good time for some breaking changes, although I guess it can be in a separate change (if at all).

packages/labs/context/src/test/late-provider_test.ts
AndrewJakubowicz: Nit: remove the `// bad!` comments.

AndrewJakubowicz (commented):
Awesome!

packages/labs/context/src/lib/controllers/context-provider.ts
rictic: > really minor nit: do we want explicit public keyword?

It's a big odd, but I'm following local convention. Raised as a topic on the discord.

Good point on docs, added!

packages/labs/context/src/lib/controllers/context-provider.ts
rictic: inlined into onProviderRequest and added a comment

packages/labs/context/src/test/late-provider_test.ts
rictic: Ah, forgot to upload!

packages/labs/context/src/lib/value-notifier.ts
rictic: +1, done

packages/labs/context/src/test/late-provider_test.ts
```diff
@@ -300,4 +301,90 @@ suiteSkipIE('late context provider', () => {
...
+        initialValue: 'late provider initial value',
+      });
+    }
+
+    // The indirect child now gets the late provider's initial value
+    await directChildConsumer.updateComplete;
+    assert.equal(directChildConsumer.value, 'grandparent updated');
+    assert.equal(indirectChildConsumer.value, 'late provider initial value');
+
+    // Updating the middle provider updates its childd, but not its sibling,
```
AndrewJakubowicz, line 373: Typo: `childd`

packages/labs/context/src/test/late-provider_test.ts
```diff
@@ -300,4 +301,90 @@ suiteSkipIE('late context provider', () => {
...
+    // the direct child.
+    const middleProvider = container.querySelector(
+      'late-context-provider-4'
+    ) as LateContextProvider4Element;
+    middleProvider.provide.setValue('late provider updated');
+    await directChildConsumer.updateComplete;
+    // bad!
```
AndrewJakubowicz, line 380: bad?

AndrewJakubowicz (approved):
Excellent

packages/labs/context/src/test/late-provider_test.ts
rictic: Done!

packages/labs/context/src/test/late-provider_test.ts
rictic: Sorry that was kept in from earlier assertions, this is correct behavior. directChildConsumer is directly under grandparent, so it still see grandparent updated, but the indirect consumer sees values from the late provider

packages/labs/context/src/lib/value-notifier.ts
rictic: Good point! Done

PR #4021, opened 2023-07-14
[labs/gen-wrapper-react] Update to emit `createComponent` with option object syntax

augustjk:
It should not emit deprecated syntax.

### DIFF
--- a/packages/labs/gen-wrapper-react/goldens/test-element-a/src/element-slots.ts
+++ b/packages/labs/gen-wrapper-react/goldens/test-element-a/src/element-slots.ts
@@ -3,9 +3,9 @@ import {createComponent} from '@lit-labs/react';
 
 import {ElementSlots as ElementSlotsElement} from '@lit-internal/test-element-a/element-slots.js';
 
-export const ElementSlots = createComponent(
-  React,
-  'element-slots',
-  ElementSlotsElement,
-  {}
-);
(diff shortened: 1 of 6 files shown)


packages/labs/gen-wrapper-react/goldens/test-element-a/src/element-slots.ts
```diff
@@ -3,9 +3,9 @@ import {createComponent} from '@lit-labs/react';
...
-  React,
-  'element-slots',
-  ElementSlotsElement,
-  {}
-);
+export const ElementSlots = createComponent({
+  react: React,
+  tagName: 'element-slots',
+  elementClass: ElementSlotsElement,
+  events: {},
```
AndrewJakubowicz, line 10: It looks like events can be `undefined` or `null` (and in the implementation we do a defensive `?? {}`. Is it worth the complexity to not have this field emitted in the empty case?

Non blocking.

AndrewJakubowicz (approved):
Nice!

packages/labs/gen-wrapper-react/goldens/test-element-a/src/element-slots.ts
augustjk: I'd say it's pretty harmless to emit empty object instead of trying to remove the option entirely.

PR #4025, opened 2023-07-18
[all] Add explicit "types" to tsconfig

augustjk:
Split out from https://github.com/lit/lit/pull/4000#issuecomment-1626406658

The problem with a monorepo using npm workspaces is that all shared node_modules are hoisted to the monorepo root, including any `@types/` packages. Some of these types packages can have global declarations that conflict with another package's, or get included in compiled output unnecessarily.

In the PR linked above, I found that `@types/react` adds `TrustedHTML` interface to the global which was conflicting with `@types/trusted-types` and also ends up adding itself as a reference to the `.d.ts` file where `TrustedHTML` is used, for example, `lit-html.d.ts`.

Setting https://www.typescriptlang.org/tsconfig#types makes it explicit which global types are used and included. This does not affect types of any packages that are imported, only those added implicitly by their global usage.


packages/labs/gen-wrapper-react/test-output/tsconfig.json
```diff
@@ -21,7 +21,9 @@
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
     "jsx": "react",
-    "noImplicitOverride": true
+    "noImplicitOverride": true,
+    "types": ["mocha"],
+    "skipLibCheck": true
```
AndrewJakubowicz, line 26: Why is `skipLibCheck` required?

packages/lit-starter-ts/tsconfig.json
```diff
@@ -26,7 +26,8 @@
         "name": "ts-lit-plugin",
         "strict": true
       }
-    ]
+    ],
+    "types": ["mocha"]
```
AndrewJakubowicz, line 30: Are explicit types required on starter packages?

packages/localize-tools/testdata/build-runtime-xlb/input/tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "noUnusedParameters": true,
     "preserveConstEnums": true,
     "forceConsistentCasingInFileNames": true,
-    "rootDir": "./"
+    "rootDir": "./",
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 9: Same question – what is the `skipLibCheck` needed for?

packages/localize-tools/testdata/build-runtime-xliff-ph/goldens/tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "noUnusedParameters": true,
     "preserveConstEnums": true,
     "forceConsistentCasingInFileNames": true,
-    "rootDir": "./"
+    "rootDir": "./",
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 9: skipLibCheck?

packages/localize-tools/testdata/build-runtime-xliff-ph/input/tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "noUnusedParameters": true,
     "preserveConstEnums": true,
     "forceConsistentCasingInFileNames": true,
-    "rootDir": "./"
+    "rootDir": "./",
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 9: skipLibCheck?

packages/localize-tools/testdata/build-runtime-xliff/input/tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "noUnusedParameters": true,
     "preserveConstEnums": true,
     "forceConsistentCasingInFileNames": true,
-    "rootDir": "./"
+    "rootDir": "./",
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 9: skipLibCheck?

packages/localize-tools/testdata/extract-xlb-fresh/goldens/tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "noUnusedParameters": true,
     "preserveConstEnums": true,
     "forceConsistentCasingInFileNames": true,
-    "rootDir": "./"
+    "rootDir": "./",
+    "skipLibCheck": true,
```
AndrewJakubowicz, line 9: skipLibCheck? Is it because it's in goldens?

AndrewJakubowicz (commented):
Awesome! Just a question on the usage of `skipLibCheck`. Thank you!

AndrewJakubowicz (approved):
nvm – noticed that all `skipLibCheck` was in testdata directories.
I like this change!

packages/lit-starter-ts/tsconfig.json
augustjk: It might not be needed when it's cloned to its own template repo but it's needed for the build and test to run within the monorepo.

augustjk:
The original commit was cherry-picked from when I was testing in the React update branch, adding configs until build and test scripts were passing. But I figured if we're going this route we should add this to every `tsconfig.json` we have and be explicit all the way.

PR #4027, opened 2023-07-18
[labs/react] Remove deprecated syntax and remove extra class component

augustjk:
- Refactored to move implementation directly into render function of `forwardRef` rather than creating a class component. This removes an extra React component of the same name showing up in the component tree.
- Removed deprecated old signature taking separate arguments.
- Simplified typings for `WebComponentProps` and `ReactWebComponent` and added jsdoc.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -4,72 +4,86 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const NODE_MODE = false;
-
 import type React from 'react';
 
-// Match a prop name to a typed event callback by
-// adding an Event type as an expected property on a string.
-export type EventName<T extends Event = Event> = string & {
-  __event_type: T;
-};
+const NODE_MODE = false;
+const DEV_MODE = true;
 
-// A key value map matching React prop names to event names
-type EventNames = Record<string, EventName | string>;
+/**
+ * Creates a type to be used for the props of a web component used directly in
+ * React JSX.
+ *
+ * Example:
+ *
+ * ```ts
+ * declare module "react" {
+ *   namespace JSX {
+ *     interface IntrinsicElements {
+ *       'x-foo': WebComponentProps<XFoo>;
+ *     }
+ *   }
+ * }
+ * ```
+ */
+export type WebComponentProps<I extends HTMLElement> = React.HTMLAttributes<I> &
+  // TODO(augustjk) Consider omitting keyof LitElement to remove "internal"
+  // lifecycle methods or allow user to explicitly set the prop type instead
+  // of grabbing from class
+  Partial<Omit<I, keyof HTMLElement>>;
 
-// A map of expected event listener types based on EventNames
-type EventListeners<R extends EventNames> = {
-  [K in keyof R]: R[K] extends EventName
-    ? (e: R[K]['__event_type']) => void
-    : (e: Event) => void;
-};
-
-type ReactProps<I, E> = Omit<React.HTMLAttributes<I>, keyof E>;
-type ElementWithoutPropsOrEventListeners<I, E> = Omit<
-  I,
-  keyof E | keyof ReactProps<I, E>
->;
-
-// Props the user is allowed to use, includes standard attributes, children,
-// ref, as well as special event and element properties.
-export type WebComponentProps<
-  I extends HTMLElement,
-  E extends EventNames = {}
-> = Partial<
-  ReactProps<I, E> &
-    ElementWithoutPropsOrEventListeners<I, E> &
-    EventListeners<E>
->;
-
-// Props used by this component wrapper. This is the WebComponentProps and the
-// special `__forwardedRef` property. Note, this ref is special because
-// it's both needed in this component to get access to the rendered element
-// and must fulfill any ref passed by the user.
-type ReactComponentProps<
-  I extends HTMLElement,
-  E extends EventNames = {}
-> = WebComponentProps<I, E> & {
-  __forwardedRef: React.Ref<I>;
-};
-
+/**
+ * Type of the React component wrapping the web component. This is the return
+ * type of `createComponent`.
+ */
 export type ReactWebComponent<
   I extends HTMLElement,
   E extends EventNames = {}
 > = React.ForwardRefExoticComponent<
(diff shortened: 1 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -4,73 +4,87 @@
...
  */
 
+import type React from 'react';
+
 const NODE_MODE = false;
+const DEV_MODE = true;
 
-import type React from 'react';
+/**
+ * Creates a type to be used for the props of a web component used directly in
```
AndrewJakubowicz, line 13: does `directly` mean "unwrapped" here?

packages/labs/react/src/create-component.ts
```diff
@@ -128,15 +142,18 @@ const setProperty = <E extends Element>(
...
   const event = events?.[name];
+  // Dirty check event value.
   if (event !== undefined && value !== old) {
-    // Dirty check event value.
     addOrUpdateEventListener(node, event, value as (e?: Event) => void);
     return;
   }
 
+  // But don't dirty check properties; elements are assumed to do this.
+  node[name as keyof E] = value as E[keyof E];
```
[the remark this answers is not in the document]
augustjk: All tests of expected behavior passes. I think it's a bit more correct to set the property to these nullish values then remove the attribute, rather than removing the attribute only and leaving the previous property values as it currently is.

packages/labs/react/src/create-component.ts
```diff
@@ -4,73 +4,87 @@
...
+ *   events: {
+ *     onfoo: 'foo' as EventName<FooEvent>,
+ *   }
+ * });
+ * ```
+ *
+ * `onfoo` prop will have the type `(e: FooEvent) => void`.
+ */
 export type EventName<T extends Event = Event> = string & {
   __event_type: T;
```
justinfagnani, line 65: while you're here, we'd normally write this as `__eventType`, ie the `__` is a prefix, the whole identifier is not snake_case.

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
justinfagnani: Any way we can add a test for what we think is the more correct behavior? I think it'll matter for properties that don't reflect from an attribute, or have a name that differs from the attribute.

Like for props: `{ariaChecked: null}`? The new behavior should be correct where the old was wrong.

packages/labs/react/src/create-component.ts
```diff
@@ -185,215 +189,121 @@ const setRef = (ref: React.Ref<unknown>, value: Element | null) => {
...
-    WebComponentProps<I, E>
-  >((props, __forwardedRef) =>
-    createElement<Props, ReactComponent, typeof ReactComponent>(
-      ReactComponent,
-      {...props, __forwardedRef}
-    )
-  );
+    return React.createElement(tagName, {
+      ...reactProps,
+      ref: (node: I) => {
```
justinfagnani, line 295: This creates a new closure every render. Can we hoist it?

justinfagnani (approved):

AndrewJakubowicz (approved):
I love all the great React labs package improvements!

packages/labs/react/src/create-component.ts
augustjk: Correct. No wrapper and just the custom element tag in JSX.

packages/labs/react/src/create-component.ts
augustjk: It needs to be declared in the scope of the function component so we can assign to `elementRef.current` as well. I think a new closure every render here is unavoidable.

packages/labs/react/src/create-component.ts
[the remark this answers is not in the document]
augustjk: `ariaChecked` is kind of a weird one. React would normally not accept it as a valid prop on native elements and do nothing with it. It expects `aria-checked` as a prop to set that attribute, and if the user does that we'd never enter this logic and let React handle it.

Our wrapped component does accept `ariaChecked` as a prop since we find it in `HTMLElement.prototype` and set it. It's not quite the idiomatic React use case but since we're setting it we should remove. I can add a test for that.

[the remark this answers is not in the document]
augustjk: So turns out `el.ariaChecked = true` doesn't reflect to `aria-checked` in Firefox so users shouldn't really be doing that.
Our types disallow `ariaChecked` to be passed in as a prop and only accepts `aria-checked`. I'm now thinking we shouldn't have to add a test for something users shouldn't do.

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -542,6 +542,27 @@ suite('createComponent', () => {
     );
   });
 
+  test('unsets reflected attributes from property with different specifier on null or undefined prop', async () => {
+    // Our types don't allow `ariaChecked` only `aria-checked` but it's possible
+    // for JS users and wrapper will set the property on the element
```
AndrewJakubowicz, line 547: Nit: `users and wrapper` -> `users, and the wrapper` maybe?

packages/labs/react/src/test/create-component_test.tsx
```diff
@@ -542,6 +542,27 @@ suite('createComponent', () => {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    await renderReactComponent({ariaChecked: true} as any);
+
+    // Assertion below fails for Firefox as setting the property doesn't reflect
+    // assert.equal(wrappedEl.hasAttribute('aria-checked'), true);
```
AndrewJakubowicz, line 552: `'ariaChecked' in document.body` could be used to skip this test on Firefox. (Returns `true` in Chrome & `false` in safari. This would allow testing instead of using a comment.

packages/labs/react/src/test/create-component_test.tsx
AndrewJakubowicz: Nice changes!

AndrewJakubowicz (approved):
Nice updates & nice test!

packages/labs/react/src/test/create-component_test.tsx
augustjk: Excellent suggestion. Thank you!

PR #4029, opened 2023-07-19
[labs/analyzer] Require a TypeScript object to construct an Analyzer

### LINKED ISSUE #4018: [labs/analyzer] Support for BYO TypeScript version
### Should this be an RFC?

### Description

The analyzer currently uses its own version of TypeScript (currently `4.7.4`). This is a problem when users are running the analyzer against a TS project that is being built against a different version of TS, since we currently stop analysis if there are semantic diagnostic errors (e.g. type errors), which often happens when running an older `tsc` against a codebase updated to a newer version of TS.

We should probably consider:
- Requiring that the `ts` module be passed into the `Analyzer` constructor rather than importing it directly. We could then consider having the CLI / `gen-manifest` support finding or pointing to a user-provided TS installation
- Update the analyzer to accept some number of different `ts` versions. e.g. build a union type of multiple versioned `ts` libs, and add the required duck-typing to choose correct API for where they vary. For example `node.decorators` for pre-4.8 vs. `ts.getDecorators(node)` for 4.8+
- Adding multiple builds / tests to the monorepo for building the analyzer and testing it against multiple versions

Not possible today.

justinfagnani:
Fixes #4018

Builds on https://github.com/lit/lit/pull/3980

Implements a BYOTS pattern for the analyzer. This PR does not update TypeScript yet, but I think we should update it to 5.0 as a follow up.

I implemented this fairly mechanically, bottom-up, adding a typescript arg to functions that needed it, and re-using the analyzer argument if there already was one. The end result might be a little inconsistent, but functions generally take only what they need. I think stylistically it might be nice to most the analyzer argument first, as would be done with curry-compatible style (put the varying arguments last).


packages/labs/analyzer/src/lib/analyzer.ts
```diff
@@ -29,15 +32,19 @@ export class Analyzer implements AnalyzerInterface {
...
   constructor(init: AnalyzerInit) {
-    this._getProgram = init.getProgram;
-    this.fs = init.fs;
-    this.path = init.path;
+    ({
+      fs: this.fs,
+      path: this.path,
+      typescript: this.typescript,
+      getProgram: this._getProgram,
+    } = init);
```
rictic, line 47: Is this syntax any better? a sequence of assignments seems more legible and more likely to be optimized by VMs

packages/labs/analyzer/src/lib/types.ts
```diff
@@ -250,43 +259,57 @@ const getRootName = (
...
+};
 
 /**
  * Uses the stub LSH above to parse a type string and return a syntax-only
  * TypeNode (there will be no valid symbol information retrievable directly from
  * these nodes).
  */
-const parseType = (typeString: string): ts.TypeNode | undefined => {
+// TODO (justinfagnani): This would probably be a bit cleaner as an instance
+// method on Analyzer now.
```
rictic, line 294: There will never be an easier time to make breaking changes than now

rictic (approved):

packages/labs/analyzer/src/lib/types.ts
justinfagnani: This wouldn't be a breaking change, and would add public API, so I wanted to be conservative.

There are a lot of potential cleanups I saw while in here, I also needed to limit the changes to just the relevant ones to keep things reviewable.

packages/labs/analyzer/src/lib/analyzer.ts
justinfagnani: In size constrained libraries it removes a dereference per property, and I've started to like the look of the pattern. It's also nice for code-completion when you add a new option to the init interface.

PR #4038, opened 2023-07-24
[infra] Allow benchmarks workflow in fork PRs

augustjk:
Follow up to https://github.com/lit/lit/pull/4041

This renames the workflow name we look for in the reporting workflow to "Benchmarks" instead of "Benchmarks Test" and updates the existing benchmarks workflow to upload report artifacts instead of writing comments directly. This allows the reporting workflow to download and write to comment.

The workflow doesn't work on this PR because the workflow needs to be present in the default branch.

This same code has been merged to a fork and tested in this PR: https://github.com/augustjk/lit/pull/2
See the tachometer benchmark results showing in the comment. This was created with 2 separate "Report Benchmark Results" workflow runs: https://github.com/augustjk/lit/actions
One to create the comment when benchmarks begin, and one to fill in the result after completion.

### DIFF
--- a/.github/workflows/benchmarks-report.yaml
+++ b/.github/workflows/benchmarks-report.yaml
@@ -2,7 +2,7 @@ name: Report Benchmark Results
 
 on:
   workflow_run:
-    workflows: ['Benchmarks Test']
+    workflows: [Benchmarks]
     branches: ['**']
     types:
       - completed
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,13 +1,9 @@
-name: Benchmarks
+name: Benchmarks Test
 
 on: [pull_request]
 
 jobs:
   benchmarks:
-    # Don't run on forks. Benchmarks can't run on forks because secrets aren't
-    # available, which are needed for posting result comments.
-    if: github.event.pull_request.head.repo.id == github.event.pull_request.base.repo.id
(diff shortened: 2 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -34,76 +30,47 @@ jobs:
...
         run: |
           cd packages/benchmarks
           npx tachometer \
             --config reactive-element/list/tachometer.json \
-            --json-file reactive-element/list/results.json
+            --json-file reactive-element-list.json
 
-      - name: Report reactive-element-list
-        uses: andrewiggins/tachometer-reporter-action@v2
+      - name: Upload results
```
AndrewJakubowicz, line 67: Nice! And this is what will be downloaded by the other action & reported as a comment.

AndrewJakubowicz (approved):

.github/workflows/benchmarks-report.yaml
```diff
@@ -2,7 +2,7 @@ name: Report Benchmark Results
 
 on:
   workflow_run:
-    workflows: ['Benchmarks Test']
+    workflows: [Benchmarks]
```
AndrewJakubowicz, line 5: Nice – so this should hook into the modified Benchmarks action `requested` and `completed`.

AndrewJakubowicz (approved):

PR #4039, opened 2023-07-24
[labs/virtualizer] Don't evaluate result of scrollerController.detatch() as it's always null

alanbuzek:


usergenic:
Verified this fix for linter/compiler error internally. LGTM.

usergenic (approved):

PR #4041, opened 2023-07-24
[infra] Add benchmarks report workflow

augustjk:
This is the first step in allowing benchmarks to run for PRs from forks.

Approach is taken from https://github.com/andrewiggins/tachometer-reporter-action#working-with-forks

The [`workflow_run` hook](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) for GitHub Actions only works when the workflow file is present in the default branch. This PR is to add this, but looks for workflow named "Benchmarks Test".

Once merged, I'll update this PR https://github.com/lit/lit/pull/4038 to rename the Benchmarks workflow to Benchmarks Test to see if this report result workflow works.

### DIFF
--- a/.github/workflows/benchmarks-report.yaml
+++ b/.github/workflows/benchmarks-report.yaml
@@ -1,0 +1,48 @@
+name: Report Benchmark Results
+
+on:
+  workflow_run:
+    workflows: ['Benchmarks Test']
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
Nice!

PR #4048, opened 2023-07-27
[lit-labs/compiler] only compress marker comments

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/pull/4047#discussion_r1275557741

### Why

@rictic pointed out an oversight that the naive logic of minifying empty comments could match and miscompile code if these empty comments were used elsewhere.

### Example

The code:

```html
<style>
  div::before {
    content: '<!---->';
  }
  div::after {
    content: '<!--?-->';
  }
</style>
<div>Hello world</div>
```

Renders: `<!---->Hello world<!--?-->`.

However when compiled without this fix, the compiled code would result in `<?>Hello world<?>` rendering.

### Fix

Only minify marker comments by matching on `<!--lit$12345$-->` and converting them to `<?>`. It's then far less likely to result in a collision.

### Test

Added a regression golden test case.


packages/labs/compiler/test_files/raw_text_element_edgecase.js
```diff
@@ -0,0 +1,13 @@
...
+
+render(html`<style>
+  div::before {
+    content: '<!---->';
+  }
+  div::after {
+    content: '<!--?-->';
+  }
+</style>
+<div>Hello world</div>`, document.body);
```
augustjk, line 13: Nit: newline at end of file
```suggestion
<div>Hello world</div>`, document.body);

```

augustjk (approved):
Nice!

rictic (approved):

PR #4050, opened 2023-07-27
[lit-labs/compiler] only compile the valid html tag function imported from lit

AndrewJakubowicz:

### Why

This change exists to restrict compilation down to our `html` tag function. The compiler will only run over a file if an `html` import is detected from `lit` or `lit-html`.

### How

In the pass to locate all the templates, also look for import declarations, and optionally mark the file for compilation if a valid import is found.

### Test plan

Added two goldens.

### Considerations

This change is better than having no handling, but it is still possible to cause a miscompilation if you import `html` from `lit`, and then shadow it with another tag function. There are two options I can think of that remedy this:

 1. Allow the user to customize and override `isLitTemplate`. It would need a different public API.
 2. Invoke TypeScript TypeChecker analysis & check that the declaration came from a valid import. Maybe it's worth the complexity to do this?


packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -139,6 +147,28 @@ class CompiledTemplatePass {
...
+      node.importClause?.namedBindings != null &&
+      ts.isNamedImports(node.importClause.namedBindings)
+    ) {
+      const namedBindings = node.importClause.namedBindings;
+      return namedBindings.elements.some(
+        (imp) => ts.isIdentifier(imp.name) && imp.name.text === 'html'
+      );
+    }
+    return false;
+  }
```
rictic, line 170: Can we resolve the reference for the template tag function instead? I'm concerned this approach would miscompile:

```ts
import {html} from 'lit';
import * as static from 'lit/static.js';

function foo() {
  const html = static.html;
  return html``;
}
```

rictic: Also, if we go with the currently implemented approach, does it properly handle code like this:

```ts
import {svg as html} from 'lit';
```

AndrewJakubowicz: Yes – those cases will miscompile. I can add resolution.

AndrewJakubowicz:
Converted to draft – making the checking much more robust.

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Also handle:
```
import * as lit from 'lit';

lit.html`rar`;
```

AndrewJakubowicz:
From the great feedback, this PR has been replaced with https://github.com/lit/lit/pull/4055

PR #4052, opened 2023-07-27
fix(labs/gen-wrapper-*): fix module path error when generating wrapper code in windows

### LINKED ISSUE #4051: [labs/gen-wrapper-*] In Windows systems, the module path generated by the template wrapper is incorrect

React (@lit-labs/react)

### Description

In Windows systems, the module path generated by the template wrapper is incorrect.

In [labs/analyzer/src/lib/analyzer.ts](https://github.com/lit/lit/blob/main/packages/labs/analyzer/src/lib/analyzer.ts#L73-L82), we use `path.normalize` to adapt file paths to various operating systems:

image

This value will be passed down and eventually used in [packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts](https://github.com/lit/lit/blob/main/packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts#L24-L34)

image

It will cause the generated wrapper code to be incorrect in the Windows system:
image

### Reproduction

execute `lit labs gen --framework=vue` in windows system.

### Workaround

We should replace `\` with `/` of module path when generating the wrapper code. I will solve this problem in #4052

### Is this a regression?

### Affected versions

The lasted version

Node version: 18.12.0

zh-lx:
Fix module path error when generating wrapper code in windows. close #4051

### DIFF
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
@@ -10,15 +10,27 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
 const root = `${__dirname}/..`;
-const files = fs.readdirSync(root);
-for (const file of files) {
-  if (file.endsWith('.vue.d.ts')) {
-    fs.renameSync(
-      `${root}/${file}`,
-      `${root}/${file.replace('.vue.d.ts', '.d.ts')}`,
-      (err) => {
-        console.log(err);
+
+function traverseRename(directoryPath) {
+  const files = fs.readdirSync(directoryPath);
+
+  for (const file of files) {
+    const filePath = path.join(directoryPath, file);
+    const stat = fs.statSync(filePath);
+    if (stat.isFile()) {
+      if (file.endsWith('.vue.d.ts')) {
+        fs.renameSync(
+          filePath,
+          `${filePath.replace('.vue.d.ts', '.d.ts')}`,
+          (err) => {
+            console.log(err);
+          }
+        );
       }
-    );
+    } else if (stat.isDirectory()) {
+      traverseRename(filePath); // 递归遍历子目录
+    }
   }
 }
+
+traverseRename(root);
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
@@ -28,7 +28,7 @@ function traverseRename(directoryPath) {
         );
       }
     } else if (stat.isDirectory()) {
-      traverseRename(filePath); // 递归遍历子目录
+      traverseRename(filePath); // transfer the sub directory and rename file
     }
   }
 }
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
@@ -28,7 +28,7 @@ function traverseRename(directoryPath) {
         );
       }
     } else if (stat.isDirectory()) {
-      traverseRename(filePath); // transfer the sub directory and rename file
+      traverseRename(filePath); // traverse the sub directory and rename files
     }
   }
 }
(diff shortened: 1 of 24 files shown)


augustjk:
Thanks for the PR! If possible we'd like to see a test for this fix so we'll be able to catch this.

packages/labs/gen-wrapper-vue/src/lib/rename-template.ts
```diff
@@ -19,16 +21,28 @@ const __filename = fileURLToPath(import.meta.url);
...
+          filePath,
+          \`\${filePath.replace('.vue.d.ts', '.d.ts')}\`,
+          (err) => {
+            console.log(err);
+          }
+        );
       }
-    )
+    } else if (stat.isDirectory()) {
+      traverseRename(filePath); // 递归遍历子目录
```
augustjk, line 42: Could you update the comment to be in English? It should update in the golden too.

packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
```diff
@@ -35,14 +36,22 @@ export default {
...
-        entryFileNames: ({ name }) => \`\${name}.js\`,
+        // For subpath component, we should output it in subpath.
+        // Otherwise, if there are files with the same file name in the subPath and rootPath,
+        // output to the same directory will conflict.
+        entryFileNames: ({ name, facadeModuleId }) => {
+          const sourceFileDir = path.dirname(facadeModuleId);
+          const relativePath = path.relative(__dirname, sourceFileDir);
+          const targetName = path.join(relativePath, \`\${name}.js\`).replace(/\\\\/g, '/').replace(/^src\\\//, '');
+          return targetName;
+        },
```
augustjk, line 54: I'm trying to think over this logic. I think we might be able to do `preserveModules: true` and `preserveModulesRoot: 'src'` to avoid this. https://rollupjs.org/configuration-options/#output-preservemodulesroot

augustjk (changes requested):
Thanks for adding the sub element to test!
I have a couple requests for changes, but I think overall this is looking good.

We also need a changeset entry as per https://github.com/lit/lit/pull/4052#issuecomment-1653957375 but I'd be happy to do that before merging if you're unsure.

packages/labs/gen-wrapper-vue/src/lib/rename-template.ts
zh-lx: Ok, I have changed it to English.

packages/labs/gen-wrapper-vue/src/lib/vite.config-template.ts
zh-lx: I have used `preserveModules` to replace these code and it does work.

packages/labs/gen-wrapper-vue/goldens/test-element-a/scripts/rename.js
```diff
@@ -10,15 +10,27 @@ const __filename = fileURLToPath(import.meta.url);
...
+          filePath,
+          `${filePath.replace('.vue.d.ts', '.d.ts')}`,
+          (err) => {
+            console.log(err);
+          }
+        );
       }
-    );
+    } else if (stat.isDirectory()) {
+      traverseRename(filePath); // transfer the sub directory and rename file
```
AndrewJakubowicz, line 31: Nit: Change `transfer` to `traverse`

AndrewJakubowicz: Whoops - this is the golden file. Disregard.

packages/labs/gen-wrapper-vue/src/lib/rename-template.ts
```diff
@@ -19,16 +21,28 @@ const __filename = fileURLToPath(import.meta.url);
...
+          filePath,
+          \`\${filePath.replace('.vue.d.ts', '.d.ts')}\`,
+          (err) => {
+            console.log(err);
+          }
+        );
       }
-    )
+    } else if (stat.isDirectory()) {
+      traverseRename(filePath); // transfer the sub directory and rename file
```
AndrewJakubowicz, line 42: Rename `transfer` to `traverse`, and change `file` to `files`.

AndrewJakubowicz (commented):
I took a quick look and this is looking very good. The tests are excellent.

zh-lx:
> 
I'm not sure how to write the changeset, but I would appreciate it if you could to that.

augustjk (approved):
Thank you very much! This is great! I added a changeset file and tweaked the comment wording a bit.

PR #4054, opened 2023-07-28
GC more often in leak tests

rictic:
If we have leaks, doing more GC runs wouldn't help. Seems like Chrome's gc runs don't always compact, and performance.memory.usedJSHeapSize is too fast to be precise.

By doing more GC passes as the test runs we can reduce noise in heap growth.


AndrewJakubowicz (approved):
LGTM! Out of interest how did you choose the various thresholds?

rictic:
They're a bit above the observed heap growth

PR #4055, opened 2023-07-28
[lit-labs/compiler] only compile the valid html tag function imported from lit

AndrewJakubowicz:

PR made to replace https://github.com/lit/lit/pull/4050 (which was not granular or specific enough).

### Why

This change exists to restrict compilation down to our (`lit` or `lit-html`) `html` tag function. This prevents miscompilation by ignoring the `static.js` html tag function (and others).

Note this change doesn't cover all cases, e.g. reassignment is not covered because it became far too complex to include. A test was checked in with TODO.

### How

Add a single file simple TypeScript TypeChecker that provides a much more robust `isLitTemplate` method.

Cases handled in this change:

```ts
import {html} from 'lit';                     // Detects: html``
import {html as h} from 'lit';                // Detects: h``
import {html as svg, svg as html} from 'lit'; // Detects svg``
import * as lit from 'lit';                   // Detects lit.html``
```

Anything else does not get compiled. E.g.:

```ts
import {html} from 'lit-html/static.js"; // html`` will not be detected as a compilable template.
```

### Test plan

Added goldens.


augustjk (approved):
LGTM. I imagine this also won't work for re-exports of `html` from packages not named `lit` or `lit-html`. Unsure if that's a use case we need to support at the moment.

packages/labs/compiler/src/lib/type-checker.ts
```diff
@@ -0,0 +1,213 @@
...
+      return false;
+    }
+    const templateImport = symbol.declarations?.[0];
+    if (!templateImport || !ts.isImportSpecifier(templateImport)) {
+      return false;
+    }
+
+    if (
+      templateImport.propertyName &&
+      templateImport.propertyName?.text !== 'html'
```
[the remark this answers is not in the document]
AndrewJakubowicz: oh yeah! Great catch!

AndrewJakubowicz:
> 

This is accurate. It's something to consider as a feature in the future.

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -98,8 +100,17 @@ class CompiledTemplatePass {
    * HTML.
    */
   private securityBrandIdent: ts.Identifier;
-  private constructor(private readonly context: ts.TransformationContext) {
+  /**
+   * Type checker which can be used to query if a tagged template expression is
+   * a lit template.
+   */
+  private checker: ReturnType<typeof getTypeChecker>;
```
rictic, line 107: Can this be readonly?

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -83,7 +84,8 @@ class CompiledTemplatePass {
    */
   private topLevelStatementToTemplate = new Map<ts.Statement, TemplateInfo[]>();
   /**
-   * Map an `html` tagged template expression to the template info.
+   * Map an `html` tagged template expression to the template info for each
+   * template to compile.
    */
   private expressionToTemplate = new Map<
```
rictic, line 90: Can this be readonly?

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -83,7 +84,8 @@ class CompiledTemplatePass {
    */
   private topLevelStatementToTemplate = new Map<ts.Statement, TemplateInfo[]>();
```
rictic, line 85: readonly?

packages/labs/compiler/src/lib/type-checker.ts
```diff
@@ -0,0 +1,213 @@
...
+    const importDeclaration = namespaceImport.parent.parent;
+    const specifier = importDeclaration.moduleSpecifier;
+    if (!ts.isStringLiteral(specifier)) {
+      return false;
+    }
+    return this.isLitTemplateModuleSpecifier(specifier.text);
+  }
+
+  private isLitTemplateModuleSpecifier(specifier: string): boolean {
+    return specifier === 'lit' || specifier === 'lit-html';
```
rictic, line 163: Or 'lit-element' I think

packages/labs/compiler/test_files/handle_aliased_imports.js
```diff
@@ -0,0 +1,4 @@
+import {svg as html, html as svg} from 'lit';
+
+html`<text>Do not compile me, I am an svg</text>`;
```
rictic, line 3: ha ha, great test!

packages/labs/compiler/test_files/handle_html_identifier_reassignment.js
```diff
@@ -0,0 +1,16 @@
+import {html as a} from 'lit';
+// TODO: Renaming import outside the import declaration is not supported.
+// This test doesn't currently work correctly.
```
rictic, line 3: We miss this optimization, but it totally works!

rictic (approved):

packages/labs/compiler/test_files/handle_html_identifier_reassignment.js
AndrewJakubowicz: Oh I see! It "works" because it's still correct – just not optimized.

Great point – will reword.

packages/labs/compiler/src/lib/type-checker.ts
AndrewJakubowicz: Ah yes – added. Thank you!

PR #4056, opened 2023-07-28
[lit-labs/compiler] add element, attribute, boolean, event, property parts

AndrewJakubowicz:

### Why

Prior to this PR the compiler only handled child parts. This PR adds all the attribute parts, as well as element parts.

### How

The logic is quite similar to what happens in the [`Template constructor` code in `lit-html.ts`](https://github.com/lit/lit/blob/main/packages/lit-html/src/lit-html.ts#L873-L966).

Because this is using parse5 and not TreeWalker, there are small differences in the API, but otherwise it's the same.

The compiler also needs to add imports for the various attribute parts, tracked in a separate data structure. During preparation if an attribute part is encountered (for the first time), a unique identifier is assigned for the attribute part constructor. This signals that the constructor needs to be imported and reassigned to the unique identifier. The unique identifier is used to avoid naming collisions.

While implementing this change, I also noticed that I messed up the `tsconfig` and it wasn't accessing any DOM types – which are needed for parse5 types to work correctly. Hence the modifications to `tsconfig`.

### Test plan

Tested by adding a golden for each individual part. This ensures that a single part only generates a single constructor import. A golden was added containing every part as well.

Note I also manually tested the kitchen sink compiled golden in the lit playground, and checked that each part is correctly handled.

### Missing features

Note: while comparing the preparation code to lit-html, note that the prepare phase is not completed by this change. Still to come: raw text elements handling, bindings in comments.

Thank you!


packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+  if (brandIdx === -1) {
+    throw new Error(
+      `Internal Error: Could not find security brand declaration.`
+    );
+  }
+  // Add parts import immediately before the security brand.
+  const beforeSecurityBrand = sourceFile.statements.slice(0, brandIdx);
+  const afterSecurtyBrand = sourceFile.statements.slice(brandIdx);
+
+  const partConstructors = [
```
justinfagnani, line 113: This can be hoisted outside of the function, and based on `attributePartConstructors`:

```ts
const partConstructors = Object.values(attributePartConstructors);
```

But also, the name is a bit ambiguous here. First, this seems to have only the attribute-part-like constructor names. If the name included "attribute" it'd be very similar to `attributePartConstructors` so my question would be "how is it different?". Then these aren't really constructors, but constructor names.

I tend to just go for long names if they're descriptive and accurate, rather than try too hard to shorten things. I would call this array `attributePartConstructorNames` and the other object possible `attributePartConstructorNameMap`.

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+    'AttributePart',
+    'PropertyPart',
+    'BooleanAttributePart',
+    'EventPart',
+  ] as const;
+  const partsObjectBinding = partConstructors
```
justinfagnani, line 119: I think this section could use a comment

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+    .map((part) => {
+      const identifierAlias = partIdentifiers[part];
+      if (!identifierAlias) {
```
justinfagnani, line 122: If I'm following this, you allow an object to be provided that renames the part constructors that are imported.

Who decides on the mapping, and is this needed? Can't you create a new unique name for the opt-level scope with TS APIs(`factory.createUniqueName`) and always rename the constructors?

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+      return factory.createBindingElement(
+        undefined,
+        factory.createIdentifier(part),
+        identifierAlias,
+        undefined
+      );
+    })
+    .filter((i): i is ts.BindingElement => i !== undefined);
+
+  return factory.updateSourceFile(sourceFile, [
```
justinfagnani, line 134: add comment?

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -111,16 +252,43 @@ export const createCompiledTemplateResult = ({
...
 }) =>
   f.createArrayLiteralExpression(
     parts.map((part) => {
       const partProperties = [
         f.createPropertyAssignment('type', f.createNumericLiteral(part.type)),
         f.createPropertyAssignment('index', f.createNumericLiteral(part.index)),
       ];
+      if (
+        part.type === PartType.ATTRIBUTE &&
+        (part.ctorType === PartType.ATTRIBUTE ||
```
justinfagnani, line 272: isn't this second clause always true? what other `ctorType` could there be?

AndrewJakubowicz: ... Yes it is. I'll be removing it. Great catch!

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+}
+
+/**
+ * Creates the statement:
+ *
+ * ```ts
+ * import * as <namespace> from "<moduleSpecifier>"
+ * ```
+ */
+const importNamespaceDeclaration = (
```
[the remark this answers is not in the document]
AndrewJakubowicz: Agree – done

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+    };
+
+const attributePartConstructors = {
+  [PartType.ATTRIBUTE]: 'AttributePart',
+  [PartType.PROPERTY]: 'PropertyPart',
+  [PartType.BOOLEAN_ATTRIBUTE]: 'BooleanAttributePart',
+  [PartType.EVENT]: 'EventPart',
+} as const;
+
+// These constructors have been renamed to reduce the chance of a naming collision.
```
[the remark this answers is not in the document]
AndrewJakubowicz: This comment was updated. Hopefully it makes more sense now:

```
/**
 * Mapping of part constructors to a unique identifier alias. If undefined, the
 * constructor will not be imported.
 */
```

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: Done

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: You are following correctly. The prep stage is where we decide which parts to import (because we know which parts have been encountered). Therefore if a file contains no attribute parts, we do not need to generate any imports.

The unique names are to provide defense against naming collisions – in the edge case where a file already contains an `AttributePart` identifier.

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: Much better – done.

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,155 @@
...
+ *
+ * ```ts
+ * import * as litHtmlPrivate_1 from "lit-html/private-ssr-support.js";
+ * const { AttributePart: <AttributePart alias>, PropertyPart: <PropertyPart alias>, BooleanAttributePart: <BooleanAttributePart alias>, EventPart: <EventPart alias> } = litHtmlPrivate_1._$LH;
+ * ```
+ */
+export const addPartConstructorImport = ({
+  factory,
+  sourceFile,
+  securityBrand,
```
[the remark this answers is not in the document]
AndrewJakubowicz: Good call. Done

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: Done

packages/labs/compiler/src/lib/ast-fragments.ts
justinfagnani: Can't you just use `factory.createUniqueName` instead of taking a mapping?

AndrewJakubowicz: Currently the order of operations, taken from `CompiledTemplatePass.getTransformer()` are:
 1. Do a traversal of the AST to locate all the templates. `pass.findTemplates`
 2. Rewrite the found templates. `pass.rewriteTemplates`. If an attribute part requires a constructor, the unique name is created here, and added to the part constructor mapping.
 3. Finally, if any part constructors are used, add the part constructor import to the source file.

The benefit of this ordering is that we only add the import if a compiled template needs the constructors. Maybe I've overcomplicated it.

An alternative order could be. 

1. Generate all the unique constructor identifier up front.
2. Do steps 1. and 2. above. While rewriting the templates track which constructors are actually used.
3. Then pass in a set of which ones we want to actually add to the file.

But I think the book keeping is still required.

Another alternative is to not bother doing any book–keeping, and add the import and all the Attribute part constructors regardless. Although that would require a downstream transform to tree shake or dead code eliminate them.

AndrewJakubowicz: Ah, and going back to your top questions.

> Who decides on the mapping, and is this needed?
The `rewriteTemplate` method only adds a `uniqueIdentifier` for an attribute part constructor if it is used. It's not necessarily needed. We could do one of the alternatives in my comment above.

> Can't you create a new unique name for the opt-level scope with TS APIs(factory.createUniqueName) and always rename the constructors.

What is the `opt-level` scope? In the current implementation the constructors are always renamed. They're just instantiated with `factory.createUniqueName` during the `rewriteTemplate` pass; so we both track the unique name and which ones are used in a `CompiledTemplate`.  Then we only import and pull the constructors off `_$LH` that are used.

justinfagnani: Sorry, I meant "top-level" scope - ie, a name unique in that scope. But I mispoke a bit because I'm not sure that the TS API has scope-aware unique name generation, and the name has to be unique in all scopes that have templates, not just the top-level scope. Anyway, `factory.createUniqueName()` is fine.

packages/labs/compiler/tsconfig.json
```diff
@@ -18,10 +18,9 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitAny": true,
     "noImplicitThis": true,
-    "moduleResolution": "NodeNext",
-    "stripInternal": true,
+    "moduleResolution": "Node",
```
augustjk, line 21: Curious as to why this was needed as many of our packages use `"NodeNext"` which I think is preferable as it's needed for resolving package exports.

packages/labs/compiler/test_files/part_attribute.js
```diff
@@ -0,0 +1,2 @@
+import {html} from 'lit-html';
+const attributePart = html`<div data-attr="${'attributeValue'}"></div>`;
```
augustjk, line 2: Unsure how extensive this needs to be but should there also be examples of multiple expressions within the double quotes, and binding single value expression without double quotes?

augustjk (approved):
Looks good to me overall.

packages/labs/compiler/test_files/part_attribute.js
AndrewJakubowicz: That's a great point. I can definitely add those!

AndrewJakubowicz: They don't need to be extremely exhaustive as eventually the lit-html_test suite will be compiled & tested which exercises many of these edge cases.

packages/labs/compiler/tsconfig.json
AndrewJakubowicz: I haven't figured out why this is needed. If I use `NodeNext` I get the errors.
(I get lots, but this is the main error, repeated):
```
../../../node_modules/@parse5/tools/lib/creation.d.ts:2:91 - error TS2307: Cannot find module 'parse5/dist/tree-adapters/default.js' or its corresponding type declarations.

2 import type { CommentNode, TextNode, Element, Template, DocumentFragment, Document } from 'parse5/dist/tree-adapters/default.js';
                                                                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

packages/labs/compiler/test_files/part_attribute.js
AndrewJakubowicz: Test updated

packages/labs/compiler/tsconfig.json
augustjk: I see, thank you for the details. Looks like the types for `@parse5/tools` is digging into the `parse5` package's innards not listed in the `package.json`. This is an upstream issue that seems fixable but not important for now.

augustjk: Filed issue here: https://github.com/parse5/parse5-tools/issues/16

augustjk (approved):
Thanks for adding the extra tests!

packages/labs/compiler/tsconfig.json
AndrewJakubowicz: Thank you so much!

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,175 @@
...
+      strings: Array<string>;
+      ctorType: PartType;
+    };
+
+const attributePartConstructors = {
+  [PartType.ATTRIBUTE]: 'AttributePart',
+  [PartType.PROPERTY]: 'PropertyPart',
+  [PartType.BOOLEAN_ATTRIBUTE]: 'BooleanAttributePart',
+  [PartType.EVENT]: 'EventPart',
+} as const;
```
rictic, line 42: Consider separating out the attribute kind as its own type, because `ctorType` can't be any PartType, it's just one of the attribute kinds. Likewise, maybe we should remove the values that can only be attribute kinds from PartType

```suggestion
export const AttributeKind = {
  ATTRIBUTE: 1,
  PROPERTY: 3,
  BOOLEAN_ATTRIBUTE: 4,
  EVENT: 5,
} as const;

export type AttributeKind = (typeof AttributeKind)[keyof typeof AttributeKind];

export type TemplatePart =
  | {
      type:
        | typeof PartType.CHILD
        | typeof PartType.COMMENT_PART
        | typeof PartType.ELEMENT;
      index: number;
    }
  | {
      type: typeof PartType.ATTRIBUTE;
      index: number;
      name: string;
      strings: Array<string>;
      ctorType: AttributeKind;
    };

const attributePartConstructors = {
  [AttributeKind.ATTRIBUTE]: 'AttributePart',
  [AttributeKind.PROPERTY]: 'PropertyPart',
  [AttributeKind.BOOLEAN_ATTRIBUTE]: 'BooleanAttributePart',
  [AttributeKind.EVENT]: 'EventPart',
} as const;
```

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,175 @@
...
+      factory.createNamespaceImport(namespace)
+    ),
+    moduleSpecifier,
+    undefined
+  );
+
+/**
+ * Add Parts constructors import. To avoid identifier collisions, each part
+ * constructor is mapped to a unique identifier that must be provided. If no
+ * identifier is provided for a part, then the part will not be emitted.
```
rictic, line 87: > If no identifier is provided for a part, then the part will not be emitted.

what does that mean? should we instead throw in that case?

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -105,22 +266,47 @@ export const createCompiledTemplateResult = ({
...
     parts.map((part) => {
       const partProperties = [
         f.createPropertyAssignment('type', f.createNumericLiteral(part.type)),
         f.createPropertyAssignment('index', f.createNumericLiteral(part.index)),
       ];
+      if (part.type === PartType.ATTRIBUTE) {
+        const ctorAlias =
+          attributePartConstructorNameMap[
+            attributePartConstructors[
+              part.ctorType as keyof typeof attributePartConstructors
```
rictic, line 291: The AttributeKind type should remove the need for this cast

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -68,12 +76,29 @@ class CompiledTemplatePass {
...
+        }
         const transformedSourceFile = pass.rewriteTemplates(sourceFile);
         if (!ts.isSourceFile(transformedSourceFile)) {
           throw new Error(
             `Internal Error: Expected source file to be transformed into another source file.`
           );
         }
+        if (pass.addedSecurityBrandVariableStatement === null) {
+          // No compilation occurred.
+          return sourceFile;
```
rictic, line 91: this might lead to a confusing situation if later on we introduce changes do some transformation but don't always add the security brand statement for some reason

I think we should return transformedSourceFile here, it's the same as returning sourceFile isn't it, and this is a rare case anyways, where we found templates but bailed out of compiling them?

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -242,6 +322,39 @@ class CompiledTemplatePass {
...
+              f.createUniqueName('E');
+            break;
+          }
+          case PartType.PROPERTY: {
+            this.attributePartConstructorNames.PropertyPart ??=
+              f.createUniqueName('P');
+            break;
+          }
+          default: {
+            throw new Error(`Internal Error: Unexpected attribute type.`);
```
rictic, line 352: Can add a `never` assignment here for the type system to help assert that we have all cases, once we have a specific type for the ctorType field

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: This means that if a certain part constructor does not have a unique identifier provided, we ignore it and don't pull it off `_$LH`.

This prevents compiled files containing every part constructor. We only import what is used.

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: \*chefs kiss\*
Thank you!

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Agree – done. I removed this and rolled the expectation of the security brand into the next `if` statement. That also simplifies the flow.

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Done! Nice!

packages/labs/compiler/src/lib/ast-fragments.ts
rictic: Oh, I see. I initially read it like "in [this case], [this parameter] is ignored", which set off my warning bells, but it's more like "add imports as requested by the parameters". The only things that are ignored are the values which aren't requested, i.e. that aren't in the attributePartConstructorNameMap

packages/labs/compiler/src/lib/ast-fragments.ts
```diff
@@ -7,14 +7,175 @@
...
+    moduleSpecifier,
+    undefined
+  );
+
+/**
+ * Add Parts constructors import. To avoid identifier collisions, each part
+ * constructor is mapped to a unique identifier that must be provided. If no
+ * identifier is provided for a part, then the part will not be emitted.
+ *
+ * This uses a namespace import for g3 compatibility.
```
rictic, line 96: Consider this rewrite:

```suggestion
/**
 * Add an import for the needed part constructors.
 *
 * By the time this function is called, we've already transformed the templates in the 
 * file, and discovered which part constructors we need, and those transforms have 
 * generated unique identifiers for each, so this function's job is to make sure the 
 * requested constructors are available as those identifiers.
 *
 * This uses a namespace import for g3 compatibility.
```

rictic (approved):

packages/labs/compiler/src/lib/ast-fragments.ts
AndrewJakubowicz: Exactly! I'll reword the comment to make that clearer

AndrewJakubowicz: Reworded in https://github.com/lit/lit/pull/4056/commits/54a40ce01e717c67e23371402d9b0fb0fbded930

PR #4059, opened 2023-07-31
[infra] improve error message for test:common

AndrewJakubowicz:
### Context

There is an import rewriting error that can occasionally create a confusing situation for someone working on a PR. The confusing situation is that locally `npm run test -w <package>` passes, but then it fails in CI.

Most recently this issue blocked: https://github.com/lit/lit/pull/4013#issuecomment-1656505585

### Features

*Previous behavior:*

The only logged error was: `Unexpected dev request in prod mode: /packages/labs/task/development/deep-equals.js`

*New behavior:*

A new log is added with: 

```
❌ There may be a relative import in 'deep-equals_test.js' which is resolving to '/packages/labs/task/development/deep-equals.js'. Ensure the import is a bare module. Reproduce locally with: `MODE=prod npm run test:common -w @lit-internal/tests`
```

### Testing

Low risk change. I tested locally by manually triggering the error.


augustjk (approved):
Nice, thanks for adding this!

PR #4061, opened 2023-08-01
[labs/react]  Fixes `WebComponentProps` type to accept `ref` prop

### LINKED ISSUE #4060: [labs/react] `WebComponentProps` type does not allow adding `ref`

React (@lit-labs/react)

### Description

The package exports a `WebComponentProps` type that is meant to be used for better typing bare custom element tag usage in JSX by adding to `JSX.IntrinsicElements`.

This type, however, does not allow adding `ref` as a prop when using the bare custom element tag.

### Reproduction

https://lit.dev/playground/#gist=b6297b43a2e7cd00ab4e1693a98df189

### Workaround

Manually type the component.

### Is this a regression?

### Affected versions

1.2.1

N/A

augustjk:
Fixes #4060

Using `React.DetailedHTMLProps` includes the optional `ref` prop with a ref that should return the element class itself. This is the same type used for built-in elements in `React.JSX.IntrinsicElements`.

We only need the `React.DetailedHTMLProps` for `WebComponentProps` meant to be used for `IntrinsicElements` whereas the internal type for the component props still just needs the `React.HTMLAttributes` as we did before. So I added a `ElementProps` type that's shared between those.

Test for this could be better but I just added a small build-time check for now.

I want to better refactor the test to really separately test the different parts of `create-component.ts` instead of the single test render harness trying pass props to a div, an x-foo, and the wrapped component like this https://github.com/lit/lit/blob/47b48915b50d3509b262a83af1453b440a5ff595/packages/labs/react/src/test/create-component_test.tsx#L126-L132
but that should be its own PR.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -29,10 +29,7 @@ export type WebComponentProps<I extends HTMLElement> = React.DetailedHTMLProps<
   React.HTMLAttributes<I>,
   I
 > &
-  // TODO(augustjk) Consider omitting keyof LitElement to remove "internal"
-  // lifecycle methods or allow user to explicitly set the prop type instead
-  // of grabbing from class
-  Partial<Omit<I, keyof HTMLElement>>;
+  ElementProps<I>;
 
 /**
  * Type of the React component wrapping the web component. This is the return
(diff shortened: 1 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -25,11 +25,11 @@ const DEV_MODE = true;
...
-export type WebComponentProps<I extends HTMLElement> = React.HTMLAttributes<I> &
-  // TODO(augustjk) Consider omitting keyof LitElement to remove "internal"
-  // lifecycle methods or allow user to explicitly set the prop type instead
-  // of grabbing from class
-  Partial<Omit<I, keyof HTMLElement>>;
+export type WebComponentProps<I extends HTMLElement> = React.DetailedHTMLProps<
+  React.HTMLAttributes<I>,
+  I
+> &
+  ElementProps<I>;
```
AndrewJakubowicz, line 32: Is this a type exported from `React`?

AndrewJakubowicz: nvm - GitHub diff wasn't showing me the declaration. Ignore sorry.

packages/labs/react/src/create-component.ts
```diff
@@ -39,10 +39,18 @@ export type ReactWebComponent<
...
-    React.RefAttributes<I>
+  React.PropsWithoutRef<ComponentProps<I, E>> & React.RefAttributes<I>
 >;
 
+// TODO(augustjk) Consider omitting keyof LitElement to remove "internal"
+// lifecycle methods or allow user to explicitly set the prop type instead
+// of grabbing from class
+type ElementProps<I> = Partial<Omit<I, keyof HTMLElement>>;
+
+type ComponentProps<I, E extends EventNames = {}> = React.HTMLAttributes<I> &
```
AndrewJakubowicz, line 50: Nit: worth adding a comment?

AndrewJakubowicz (approved):
Nice

packages/labs/react/src/create-component.ts
augustjk: Thank you, added!

PR #4063, opened 2023-08-01
[lit-labs/compiler] add raw text element handling

AndrewJakubowicz:

### Why

Raw text elements are tricky, and are special cased in the `lit` prepare phase. This is because comments within raw text elements become raw text.

The following html: `<textarea><!--?lit$1234$--></textarea>`, renders as:

Screenshot 2023-08-01 at 2 10 22 PM

In runtime lit-html we walk the text within a raw text element and split up the text into separate Text nodes. Then we can refer to a specific text node for dynamic content. Unfortunately we cannot represent adjacent text nodes in a prepared HTML string so the compiler cannot currently optimize this structure.

### Solution

Currently, if a raw text element contains any bindings, do not optimize that template.

### Test plan

Added a golden test.


rictic (approved):
simple and elegant!

PR #4064, opened 2023-08-01
[lit-labs/compiler] add comment binding handling

AndrewJakubowicz:

### Why

It's possible to create a binding in a comment node.

For example: html`<!-- ${'hi'} -->`. It also occurs if a portion of a template with bindings is commented out.

### How

The compiler transform handles this in exactly the same way as the runtime lit-html prepare phase. We find the markers and add no-op comment parts for those values.

### Test

Add golden (and also manually check the golden result): https://lit.dev/playground/#gist=21e8224b1195990b09d9a5821b441dfc


packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -343,6 +343,20 @@ class CompiledTemplatePass {
...
+            while ((i = node.data.indexOf(marker, i + 1)) !== -1) {
+              // Comment node has a binding marker inside, make an inactive part
+              // The binding won't work, but subsequent bindings will
+              parts.push({type: PartType.COMMENT_PART, index: nodeIndex});
+              // Move to the end of the match
+              i += marker.length - 1;
+            }
+            // In the compiled result, remove the markers, so compiled
+            // files are deterministic.
+            node.data = node.data.split(marker).join('');
```
augustjk, line 359: nit: what do you think about doing 
```suggestion
            node.data.replaceAll(marker, '');
```

augustjk (approved):

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: I think that's a far superior method.

AndrewJakubowicz: Oh no: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.

AndrewJakubowicz: Fixed

augustjk: Darn. Too new. I suppose. Since we're moving to es2021 for 3.0 maybe that's okay to bump. Otherwise split and join, or we'd need to `.replace()` constructing a RegExp with the marker, escaping the `$`.

rictic (approved):
Nice!

PR #4065, opened 2023-08-02
fix(analyzer): handle circular module reexports

bennypowers:
## What I did 
- add a test for circular module graphs with reexports
- keep track of which modules have been validated so as to skip them when circular graphs emerge


bennypowers:
I've managed to produce a reproduction in the gen-manifest tests. PTAL @kevinpschaaf

43081j:
nice find!

its because `depsAreValid` recursively gets stuck in a loop when there's a circular dependency:

https://github.com/lit/lit/blob/15a3b5c059a92b3204f6001608de92ea9e0763a2/packages/labs/analyzer/src/lib/javascript/modules.ts#L258-L259

`depIsValid` (run per individual direct dependency) will ultimately get to this line of code:

https://github.com/lit/lit/blob/15a3b5c059a92b3204f6001608de92ea9e0763a2/packages/labs/analyzer/src/lib/javascript/modules.ts#L245-L246

which ofc goes back to `depsAreValid` and so on ➿

bennypowers:
ok this is ready for review

packages/labs/analyzer/src/lib/javascript/modules.ts
```diff
@@ -255,18 +256,29 @@ const getAndValidateModuleFromCache = (
...
  */
-const depsAreValid = (module: Module, analyzer: AnalyzerInterface) =>
-  Array.from(module.dependencies).every((path) => depIsValid(path, analyzer));
+const depsAreValid = (
+  module: Module,
+  analyzer: AnalyzerInterface,
+  seen: Set<AbsolutePath>
+): boolean =>
+  Array.from(module.dependencies).every(
+    (path) => seen.has(path) || depIsValid(path, analyzer, seen)
```
justinfagnani, line 265: I'm finding this organization of the logic a little bit hard to reason about, because as used here, `seen.has(path)` implies the dep is valid, but as added in `depIsValid` there is no such assumption. I'm caught wondering if this works correctly, but I don't see it, or if there's a missing test case that shows this doesn't work. Do we need a test with circular imports, but we invalidate one of the modules (but not the first reached in the traversal)?

This may be ok though, but I'd at least like to see a comment about it's safe to short-circuit here, something like "if we've seen this dep and it's not valid we would have already returned false from `depIsValid` for a previous dependent" or similar.

justinfagnani (changes requested):

packages/labs/analyzer/src/lib/javascript/modules.ts
bennypowers: How's that?

justinfagnani (approved):
Thanks!

PR #4071, opened 2023-08-04
[lit-labs/compiler] gracefully handle bindings in tag name position

AndrewJakubowicz:

### Why

In DEV mode, this results in the error: `Bindings in tag names are not supported`.

### How

If a marker exists in the tag name then we know that the user supplied a binding in tag name position, and can stop compiling the template.

### Test plan

Tested by adding a golden.

### Alternative

We could add a diagnostic here – however I've opted for a simpler but correct solution as there is no diagnostic support wired into the compiler, and this is one of the last cases we need to handle.


packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -283,6 +283,13 @@ class CompiledTemplatePass {
       'pre:node': (node): boolean | void => {
         if (isElementNode(node)) {
           const attributesToRemove = new Set<unknown>();
+          if (node.tagName.includes(marker)) {
+            // Do not compile template if a marker was inserted in tag name
+            // position.
+            // TODO: Provide a diagnostic here.
+            shouldCompile = false;
+            return true;
```
augustjk, line 291: Is there value to returning `true` here? There's no need to traverse into the children if we're not compiling, right?

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -283,6 +283,13 @@ class CompiledTemplatePass {
       'pre:node': (node): boolean | void => {
         if (isElementNode(node)) {
           const attributesToRemove = new Set<unknown>();
+          if (node.tagName.includes(marker)) {
+            // Do not compile template if a marker was inserted in tag name
+            // position.
+            // TODO: Provide a diagnostic here.
```
augustjk, line 289: Nit. I think we like to add a username to TODOs so we know who might have context.

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Whoops! This should be false.

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Agree – done

rictic (approved):

PR #4077, opened 2023-08-08
[lit-labs/compiler] rollup example and source map tests

AndrewJakubowicz:

### Why

The initial labs compiler public export is only a TypeScript transform. This makes it usable within multiple build tools that accept TypeScript transformers. E.g., within g3, [`@rollup/plugin-typescript`](https://www.npmjs.com/package/@rollup/plugin-typescript), [`ttypescript`](https://www.npmjs.com/package/ttypescript), and maybe one day natively through TypeScript.

This PR adds an example rollup config for testing source maps. The example rollup config uses `@rollup/plugin-typescript` to compile the sourcemap test file. A new test has been added to ensure the resulting source maps correctly reference the authored example.

### How

- Add [`source-map`](https://www.npmjs.com/package/source-map) which allows easy queries on a source map, and manually setup some source map tests.
- Add a `rollup.source_map_tests.js` config which uses the compiler transform to compile a basic typescript file for source map testing.

### Risk

This change has minimal risks as it doesn't add or change any logic in the compiler, it's a test only and example only change.

> Why not run this transform on `lit-html_tests`?

Because labs/compiler uses TypeScript 5, while the rest of the monorepo uses TypeScript 4. This will also come in a followup, and can use the example rollup config here as inspiration.

### Notes

I noticed in this change that the license does not remain at the top of the generated `basic.js` file. This will be addressed in a followup.


packages/labs/compiler/src/test/source_map_test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const setupTest = async (filename: string): Promise<SourceMapConsumer> => {
+  const resolvedSourcemapPath = path.resolve(TEST_FILES_DIR, filename);
+  const map = await readFile(resolvedSourcemapPath, {encoding: 'utf-8'});
+  const smc = await new SourceMapConsumer(map);
+  return smc;
+};
+
+test('basic.ts', async () => {
+  const smc = await setupTest('basic.js.map');
```
augustjk, line 61: probably not a big deal with a single file test like this but maybe it's worth putting this in a `suiteSetup()` and using `suiteTeardown()` to call [`smc.destroy()`](https://github.com/mozilla/source-map#sourcemapconsumerprototypedestroy:~:text=When%20the%20SourceMapConsumer%20will%20no%20longer%20be%20used%20anymore%2C%20you%20must%20call%20its%20destroy%20method.)?

packages/labs/compiler/src/test/source_map_test.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+  assertSourceMapLocationMapping({
+    smc,
+    generatedPosition: [10, 10],
+    authoredPosition: [7, 13],
+    msg: "'sayHello' identifier declaration maps to authored declaration",
+  });
+  assertSourceMapLocationMapping({
+    smc,
+    generatedPosition: [9, 37],
```
augustjk, line 77: slight concern about maintenance of this should this change. maybe a comment to what the generated code at this position looks like would help?
concern is more so for this one that doesn't have an authored position since the authored code is in git, but the generated code is not.
something like..
```ts
// const lit_template_1 = { h: b_1 `<h1>Hello <?><?></h1>`, parts: [{ type: 2, index: 1 }, { type: 2, index: 2 }] };
//                                     ^
```

augustjk: i guess i also wonder if the .js files should be committed just like the goldens. the sourcemap test really should only be necessary to run when goldens are updated.

packages/labs/compiler/package.json
```diff
@@ -44,10 +45,26 @@
       ],
       "clean": "if-file-deleted"
     },
+    "build:rollup-source-map-tests": {
+      "command": "rollup -c rollup.source_map_tests.js",
+      "dependencies": [
+        "build"
```
augustjk, line 51: should this depend on `build:ts`, and `build` actually have `build:rollup-source-map-tests` as a dependency so that `npm run build` builds everything?

packages/labs/compiler/rollup.source_map_tests.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {compileLitTemplates} from './lib/template-transform.js';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import typescript from '@rollup/plugin-typescript';
```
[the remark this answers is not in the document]
AndrewJakubowicz: Good call – I can put this file in the list.

packages/labs/compiler/src/test/source_map_test.ts
AndrewJakubowicz: Great catch. This is a small test so I think I'm inclined to add the `smc.destroy()` directly to the end of the test.

I tried adding a `test.after.each` which cleans up the smc, but noted that if the destructor throws outside the test, we get a 0/1 tests passed with an exit code of 0 – which makes it look like everything worked. Went with simpler approach which doesn't have this strange issue.

packages/labs/compiler/src/test/source_map_test.ts
AndrewJakubowicz: Agree – will check in the generated js. Should I also check in the `js.map`?

packages/labs/compiler/package.json
AndrewJakubowicz: Thank you! I think I've fixed this.

`build` -> builds everything.
`build:ts` -> only builds typescript.
`build:rollup-source-map-tests` -> depends on build:ts.
`test` -> depends on `build`.

AndrewJakubowicz:
Thank you for the excellent suggestions!

AndrewJakubowicz:
I did some further investigating regarding the license falling below the imports in the generated source_map javascript file. That looks like a rollup/plugin-typescript behavior – as it occurs without the compiler transform.

I'll add a license golden to the test_files to show that the compiler transform doesn't do this by default.

packages/labs/compiler/rollup.source_map_tests.js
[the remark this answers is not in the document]
augustjk: Nice, you found a more appropriate location!

packages/labs/compiler/src/test/source_map_test.ts
augustjk: since all the assertions are synchronous it's probably fine as is, though something about test code execution ordering makes me a bit uncomfortable.

how about something like this?
https://github.com/lit/lit/commit/c5169669f3dfd396f3544a739edf11d6169e7d4e

this seems to work on my end. is there an actual chance that `.destroy()` would fail? and should that then report as a test failure?
if so we can add try catch like this https://github.com/google/wireit/blob/df9c9e9d3dc4bd1c92933ef6d4e1919679390478/src/test/analysis.test.ts#L27-L34

augustjk (approved):
suggestion https://github.com/lit/lit/pull/4077#discussion_r1289400431 is optional.

excellent addition of rollup example and making sure generated source maps are accurate!

rictic (approved):

PR #4084, opened 2023-08-10
[labs/context] Change labs/context bump to minor

augustjk:
Since context is still 0.3.3, a breaking change is fine in a minor bump.


AndrewJakubowicz (approved):
Thank you!

PR #4086, opened 2023-08-11
[lit-labs/compiler] Fix svg tag being accidentally compiled

AndrewJakubowicz:

### Context

I have a test branch which compiles lit-html_tests and runs tests and it caught that `svg` templates were not correctly handled.

`import {svg} from 'lit'` was marking the `svg` tag as a compilable template.

### Fix

The original check handles the complex cases, but misses the simple case. For example an import of `import {html as svg, svg as html} from 'lit'` works prior to this change because we have been assuming there will be an aliased property name.

This change adds a check for direct imports. Note that the aliased golden continues to work (I just fixed punctuation).

### Test plan

Unit tested by adding a new golden.


packages/labs/compiler/src/lib/type-checker.ts
```diff
@@ -93,9 +93,18 @@ class TypeChecker {
...
+    // This check allows aliasing `html` by ensuring propertyName is `html`.
+    // Thus `{html as myHtml}` is a valid template that can be compiled.
+    // Otherwise a compilable template must be a direct import of lit's `html`
+    // tag function.
     if (
-      templateImport.propertyName &&
-      templateImport.propertyName.text !== 'html'
+      (templateImport.propertyName &&
+        templateImport.propertyName.text !== 'html') ||
+      (!templateImport.propertyName && templateImport.name.text !== 'html')
```
augustjk, line 107: does this mean that in the case of no alias, the import specifier is like this?
```
import {<name>} from <moduleSpecifier>;
```
so if there is no `propertyName`, we check that `name` is the right value?

augustjk (approved):
Good catch. Looking good!

packages/labs/compiler/src/lib/type-checker.ts
AndrewJakubowicz: That is exactly right, I added your example without the alias to the source code.

Issue #3907, opened 2023-05-18
Lit SSR doesn't work in Cloudflare Workers


bluwy:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element), SSR (@lit-labs/ssr)

### Description

Lit SSR doesn't work in Cloudflare Workers as they rely on the ["workerd", "worker", "browser" export conditions](https://github.com/cloudflare/workers-sdk/blob/133c0423ccb4c2b35a1dd26157ce9a24c6a743bb/packages/wrangler/src/bundle.ts#L391) for bundling with esbuild. In which case, the `lit-html` and `@lit/reactive-element` packages are bundled using the `browser` condition.

But because those two packages uses `window` in the `browser` condition (which is fair), it doesn't work in Cloudflare. We've found that using the `node` condition works since they don't seem to use actual Node-specific APIs.

Related https://github.com/withastro/astro/pull/6915. We're fixing Astro's Lit + Cloudflare compatibility.

### Reproduction

https://github.com/bluwy/cloudflare-lit-ssr-repro

1. npm install
2. npm run dev

The repo readme has some extra information of how to debug and patch it.

### Workaround

Adding the "worker" condition that's an exact copy of the "node" condition to the aforementioned two packages fixes it.

The repro has a patch script to work around it.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

`@lit-labs/ssr` - `3.1.2`, `lit` - `2.7.4`

### Browser/OS/Node environment

(I don't think these are related, but)

OS: macos 13.3
node: 16.19.1
npm: 6.19.3

bluwy:
Darn it, I think this is a duplicate of https://github.com/lit/lit/issues/1983. Sorry about it I'll make a comment there instead.

augustjk:
There might be some overlap but I think it's valid to keep this a separate issue from #1983 which asks for Service Worker rendering (in the browser) versus handling non-browser worker/edge environments.

My biggest hesitation about adding something like this is we ourselves don't have any testing for whether our Node build actually works properly in these environments. We'd have to dig through our dependencies to make sure they don't require any Node specific things. We probably want to limit built-in APIs to these https://common-min-api.proposal.wintercg.org/.

We do have an import of `buffer` for a Node build here https://github.com/lit/lit/blob/440890ba058e1e3c747c45c22f6526b72cb91e11/packages/lit-html/src/experimental-hydrate.ts#L17-L18
[...]

PR #4087, opened 2023-08-11
[labs/ssr] WinterCG exports

PonomareVlad:
This PR should improve SSR compatibility with runtimes following [WinterCG specs](https://runtime-keys.proposal.wintercg.org)

As part of this PR, I plan to test listed runtimes and include necessary keys for support all targets

Checklist of successful tests:
- [ ] Alibaba Cloud `edge-routine`
- [ ] Cloudflare Workers
- [ ] Deno Land
- [ ] Lagon Runtime
- [ ] Netlify - Edge Functions
- [ ] Bun
- [ ] Vercel Edge Functions

### DIFF
--- a/packages/labs/ssr-client/package.json
+++ b/packages/labs/ssr-client/package.json
@@ -19,6 +19,10 @@
   "exports": {
     ".": {
       "types": "./development/index.d.ts",
+      "edge-light": {
+        "development": "./node/development/index.js",
+        "default": "./node/index.js"
+      },
       "node": {
         "development": "./node/development/index.js",
         "default": "./node/index.js"
@@ -28,6 +32,10 @@
     },
     "./lit-element-hydrate-support.js": {
       "types": "./development/lit-element-hydrate-support.d.ts",
+      "edge-light": {
+        "development": "./node/development/lit-element-hydrate-support.js",
+        "default": "./node/lit-element-hydrate-support.js"
+      },
       "node": {
         "development": "./node/development/lit-element-hydrate-support.js",
         "default": "./node/lit-element-hydrate-support.js"
--- a/packages/labs/react/package.json
+++ b/packages/labs/react/package.json
@@ -19,6 +19,10 @@
   "exports": {
     ".": {
       "types": "./development/index.d.ts",
+      "edge-light": {
+        "development": "./node/development/index.js",
+        "default": "./node/index.js"
+      },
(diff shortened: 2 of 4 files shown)


augustjk:
Nice! Related: https://github.com/lit/lit/issues/3907
I think Lit SSR as it is does work with wintercg common APIs but we likely want running tests to confirm that is the case if we're adding specific export conditions for them, and to make sure we don't add code that breaks support. Perhaps using `vercel dev` for `edge-light` and `wrangler` for `workerd` (cloudflare)?

PonomareVlad:
> and to make sure we don't add code that breaks support

Okay, then I'll prepare a repository with demo projects as a reference

> Perhaps using vercel dev for edge-light and wrangler for workerd (cloudflare)?

Yep, I will add it soon, need som time for prepare environment 🌚

PonomareVlad:
@augustjk Following this https://github.com/solana-labs/solana-web3.js/pull/3137#issuecomment-2313170611, I propose to move the remaining runtimes out of this PR and focus on them later

PonomareVlad:
@Augustjk I added Edge Runtime in the examples for Next.js, does it make sense or should we somehow make `vercel dev` work (the launch of which will require linking a Vercel account) ?

packages/labs/ssr-client/package.json
```diff
@@ -32,6 +40,14 @@
     },
     "./lit-element-hydrate-support.js": {
       "types": "./development/lit-element-hydrate-support.d.ts",
+      "edge-light": {
+        "development": "./node/development/lit-element-hydrate-support.js",
+        "default": "./node/lit-element-hydrate-support.js"
+      },
+      "workerd": {
```
thescientist13, line 47: One thought here is to make sure that something like Workerd can actually support a Node bundle / export condition, since it is not technically a Node runtime (though their compat is improving), and even outside of Node, it doesn't support some web standards like `import.meta.url`, for example.  (and that users themselve should be aware as well for their own first party code building on top of these packages)

So I would be cautious to automatically forward each runtime key to the Node export by default without knowing that the web / Node APIs Lit uses are actually supported there.

The long way to validate that could be to create a Git repo and hook them them against every runtime and test manually, or in some cases, like workerd or Netlify, you may be able to run their CLI locally and see if you can emulate that runtime in a GitHub Action as part of this repo.

PR #4088, opened 2023-08-11
[lit-labs/compiler] do not compile templates containing deprecated octal sequences

AndrewJakubowicz:

### Context

This is part of handling all the edge cases we test in lit-html.

We specifically throw a dev warning if someone uses a deprecated octal escape sequence in their template. Unfortunately the compiler doesn't use this code path, and the prepared template will also skip this code path.

This is important because a TTL containing one of these escape sequences will get an `undefined` entry in the TTL's static strings array in whichever index contained that deprecated escape.

Without this change, the following template:

```
html`<p>Hi!</p>\2`
```

Will render nothing at all with an uncompiled template (and throw a DEV_MODE warning).

However when compiled, it instead renders the text `undefined`.

### Fix

Hand roll detection of these deprecated octal escape sequences, and do not compile those templates. This would then allow the runtime lit DEV_MODE warning to inform the user.

We could also add a warning to the compiler detection.

### Risk

The main risk of this change is a bug in the regex resulting in deoptimizations due to false positives preventing template optimization. This behavior is more correct than what currently happens with these sequences.

### Alternatives

- Maybe do nothing? Is this simply invalid code?
- A more robust detection mechanism. The following code: "`\9`" throws natively. Could we somehow construct that from the ast raw string and catch the error? I couldn't think of any way that doesn't result in `eval` or `Function` exploits.


packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -261,6 +261,46 @@ class CompiledTemplatePass {
...
+
+    for (const staticString of rawTextForOctalCheck) {
+      // Regex to detect if an invalid octal sequence was used. It also detects
+      // the invalid sequences: `\8` and `\9`.
+      // References:
+      //   - eslint no-octal-escape rule:
+      //     https://github.com/eslint/eslint/blob/main/lib/rules/no-octal-escape.js
+      //   - Chromium octal escape sequence scanner:
+      //     https://source.chromium.org/chromium/chromium/src/+/181fa1f62f501c27ed19fcb69206a0e2e1eff513:v8/src/parsing/scanner.cc;l=452-460
+      if (/^([^\\|\s]|\\.)*?\\(0)*[1-9]/gm.test(staticString)) {
```
justinfagnani, line 299: I often prefer to put regexes into named variables to be self-descriptive and have a nice place to put comments like these.

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -261,6 +261,46 @@ class CompiledTemplatePass {
...
+        return {shouldCompile: false};
```
justinfagnani, line 300: If this will cause a warning at runtime, why don't we error or log a warning here?

Thinking ahead, ultimately we'd like to be able to use the compiler to drop the template prep code from the client bundle. We'll need to compile every template for that.

justinfagnani (commented):
Just a couple style nits. Looks great!

justinfagnani (approved):

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: That is a great point. Currently the compiler has been implemented in a conservative manner where any prepare phase DEV_MODE warnings simply don't compile the template – deferring it to runtime.

E.g.:
 - dynamic tag binding
 - bindings in `<template>` | `<textarea>`
 - octal escape

This technique feels like it should be the easiest for adoption, because we are very specific about what to compile. But, I agree that we need a strategy into informing and helping developers fix the issues that deoptimize templates so we can get to that world with complete removal of prepare in certain projects.

Do you think `console.warn` as a reasonable starting goal?

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -261,6 +261,46 @@ class CompiledTemplatePass {
...
+    // This results in an undefined entry in the template strings array. lit
+    // then simply renders the `undefined` directly on the page as text. A worse
+    // result is if the octal is adjacent to other content, e.g.
+    // html`<p>Hi</p>\2`. If compiled naively, this will only render `undefined`
+    // on the page. Thus we do not compile templates containing invalid octal
+    // sequences.
+    //
+    // Reference:
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_octal
+    const rawTextForOctalCheck = ts.isNoSubstitutionTemplateLiteral(
```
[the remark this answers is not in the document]
AndrewJakubowicz: Great style nit! Applied 😄

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Another great style nit! Applied 😄

rictic (approved):
Awesome work, and great comments!

PR #4091, opened 2023-08-12
Render result via `ReadableStream` for web runtimes (edge/workers)

PonomareVlad:
Some runtimes (_like [Vercel Edge](https://vercel.com/docs/concepts/functions/edge-functions)_) [do not support](https://vercel.com/docs/concepts/functions/edge-functions/edge-runtime#unsupported-apis) Node.js APIs like `stream`

This can be solved using the web-compatible [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), which can reproduce the current rendering implementation for Node.js

[Available in Node.js globally](https://nodejs.org/docs/latest-v18.x/api/globals.html#class-readablestream) since version `18.0.0`

<details><summary>How to test this PR</summary>
<p>

Just add `overrides` field in your `package.json`:

```json
{
  "overrides": {
    "@lit-labs/ssr": "npm:@ponomarevlad/lit-labs-ssr@3.1.5-ssr-readable-stream-npm-1",
    "@lit-labs/task": "npm:@ponomarevlad/lit-labs-task@3.0.0-ssr-readable-stream-npm-1",
    "@lit-labs/router": "npm:@ponomarevlad/lit-labs-router@0.1.1-ssr-readable-stream-npm-1",
    "@lit-labs/context": "npm:@ponomarevlad/lit-labs-context@0.3.3-ssr-readable-stream-npm-1",
    "@lit-labs/ssr-client": "npm:@ponomarevlad/lit-labs-ssr-client@1.1.3-ssr-readable-stream-npm-1",
    "@lit-labs/ssr-dom-shim": "npm:@ponomarevlad/lit-labs-ssr-dom-shim@1.1.1-ssr-readable-stream-npm-1",
    "@lit/reactive-element": "npm:@ponomarevlad/reactive-element@1.6.3-ssr-readable-stream-npm-1",
    "lit-element": "npm:@ponomarevlad/lit-element@3.3.3-ssr-readable-stream-npm-1",
    "lit-html": "npm:@ponomarevlad/lit-html@2.8.0-ssr-readable-stream-npm-1",
    "lit": "npm:@ponomarevlad/lit@2.8.0-ssr-readable-stream-npm-1"
  }
}
``` 

And run `npm install` ✨

</p>
</details>

### DIFF
--- a/packages/labs/ssr/src/lib/render-result-readable-stream.ts
+++ b/packages/labs/ssr/src/lib/render-result-readable-stream.ts
@@ -1,0 +1,81 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {RenderResult} from './render-result.js';
+
+type RenderResultIterator = Iterator<string | Promise<RenderResult>>;
+
+/**
+ * A Readable that reads from a RenderResult.
+ */
+export const renderResultReadableStream = (result: RenderResult) => {
+  let closed = false;
+
+  /**
+   * A stack of open iterators.
+   *
+   * We need to keep this as instance state because we can pause and resume
+   * reading values at any time and can't guarantee to run iterators to
+   * completion in any one loop.
+   */
+  const iterators = [result[Symbol.iterator]()];
+
+  const encoder = new TextEncoder();
+
+  return new ReadableStream({
+    cancel: () => void (closed = true),
+    pull: async (controller) => {
+      // This implementation reads values from the RenderResult and pushes them
+      // into the base class's Readable implementation. It tries to be as
+      // efficient as possible, which means:
+      //   1. Avoid microtasks and Promise allocations. Read and write values
+      //      synchronously when possible.
+      //   2. Write as many values to the Readable per call to _read() as
+      //      possible.
+      //
+      // To do this correctly we must adhere to the Readable contract for
+      // _read(), which states that:
+      //
+      // - The size parameter can be safely ignored
+      // - _read() should call `this.push()` as many times as it can until
+      //   `this.push()` returns false, which means the underlying Readable
+      //   does not want any more values.
+      // - `this._read()` should not be called by the underlying Readable until
+      //   after this.push() has returned false, so we can wait on a Promise
+      //   and call _read() when it resolves to continue without a race condition.
+      //   (We try to verify this with the this._waiting field)
+      // - `this.push(null)` ends the stream
+      //
+      // This means that we cannot use for/of loops to iterate on the render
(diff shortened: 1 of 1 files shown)


justinfagnani (changes requested):
Thanks for the PR @PonomareVlad !

We'll need some tests here to make sure that this works and stays working. They can be pretty simple and just make sure that we can read from the stream, like in: https://github.com/lit/lit/blob/main/packages/labs/ssr/src/test/lib/render-result-readable_test.ts

PonomareVlad:
@justinfagnani thank you, I'll try to add it soon

packages/labs/ssr/src/lib/render-result-readable-stream.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+import {RenderResult} from './render-result.js';
+
+type RenderResultIterator = Iterator<string | Promise<RenderResult>>;
+
+/**
+ * A Readable that reads from a RenderResult.
+ */
+export const renderResultReadableStream = (result: RenderResult) => {
```
justinfagnani, line 14: In https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/render-result-readable.ts we use subclassing so that you construct a new `RenderResultReadable`. I think we should make these modules consistent*. So should we use subclassing here? Or should we add a factory function to `render-result-readable.ts`?

*overriding consistency would be being idiomatic - so if users of the different stream APIs have different expectations, we should meet those.

justinfagnani (changes requested):

packages/labs/ssr/src/lib/render-result-readable-stream.ts
PonomareVlad: I have a issues with types when try to do subclassing, some refs:

https://github.com/microsoft/TypeScript/issues/39051
https://github.com/whatwg/streams/issues/778

I can try to organize it in a different way, if there are ideas 🫠

PonomareVlad:
> We'll need some tests here to make sure that this works and stays working.

Tests added

p.s. sorry for accidental `force push` 🙈

justinfagnani (approved):
thanks!!

justinfagnani (changes requested):
oops, I approved a bit too early. We need you to sign the CLA, and add a changeset file by running `npx changeset` in the monorepo root.

PonomareVlad:
> We need you to sign the CLA...

@justinfagnani I have some problems with this... 🫠

Снимок экрана 2023-11-26 в 23 39 59

Снимок экрана 2023-11-26 в 23 34 38

PonomareVlad:
@justinfagnani what else need to finish this PR ?

Issue #4067, opened 2023-08-03
[labs/testing] Call site location by stack trace might not work in webkit


augustjk:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

Seeing failures in this workflow. https://github.com/lit/lit/actions/runs/5744321662/job/15570775282?pr=4062#step:8:506

This is due to webkit stack trace not including the `?wtr-session-id` querystring in the source file url, which we rely on here https://github.com/lit/lit/blob/51c431ee2a4c28a6e1a2fe6b0069ad24c7df0df3/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts#L60-L62

Cause seems to be `playwright` version greater than 1.32. Unsure if it's playwright + webkit issue, or webkit straight up not including the querystring in stack.

Solution might be to find a different way to pinpoint the source file, maybe finding the first file that isn't an internal `@lit-labs/testing` file.

### Reproduction

TODO

### Workaround

Pin version of playwright.

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

0.2.1

### Browser/OS/Node environment

.

justinfagnani:
> or webkit straight up not including the querystring in stack

I don't think it's that the querystring isn't in the stack, I think it's that the stack frame that includes it is pushed out of the 10 frame window that's output in `error.stack`. Somehow there are more lines for the test framework in Webkit not.

augustjk:
The behavior varied depending on the version of playwright. I've seen the mocha stacks, but also the correct lines without the querystrings that won't match the regex.

PR #4092, opened 2023-08-13
Update playwright

justinfagnani:
Firefox playwright tests are crashing for me in the standard decorators test and updating playwright seems to fix that. We're pretty far behind, so it seems like a good thing to do.

Updating playwright causes a few problems though:

1. Chromium outputs inline styles slightly differently, so I needed to make SSR tests be able to a set of expectations.
2. Webkit has changes it's stack traces. This one is trickier because I don't see the stack entry that the SSR fixtures are looking for at all in the Webkit stacks - they're buried under repeated calls to a Mocha auto-run function. I updated the `testing` tests to use the base option to the fixture just to see them pass. I'm not sure if we have any choice except to force users to pass the base URL if we can't reliably get the test URL from the stack.


augustjk (approved):
LGTM.

Filed this issue for labs/testing. https://github.com/lit/lit/issues/4067

PR #4095, opened 2023-08-14
[lit-labs/compiler] Do not compile templates with invalid template or textarea tags

AndrewJakubowicz:

### Context

This is similar in spirit to https://github.com/lit/lit/pull/4088, where there is a runtime DEV_MODE check in lit-html that the compiler needs to handle.

Initially, handling by marking the templates as uncompilable such that the check remains in runtime.
In the future will add diagnostics to these cases.

### Fix

Take the same logic as lit-html https://github.com/lit/lit/blob/lit-html%402.8.0/packages/lit-html/src/lit-html.ts#L893-L911, and use it to mark a template as uncompilable.

### Test plan

Tested by using the lit-html_test tests, and ensuring the invalid goldens do not compile.

Future work is adding diagnostics to the compiler.


justinfagnani (approved):

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -290,6 +290,16 @@ class CompiledTemplatePass {
             shouldCompile = false;
             return false;
           }
+          if (
+            /^(?:textarea|template)$/i!.test(node.tagName) &&
```
rictic, line 294: This regex works, but it's somewhat tricky! tagName is always all caps, so we don't need to worry about that. A regexp can be faster if it doesn't have to ignore case, so consider writing it with all caps. But since this code isn't _that_ performance sensitive, consider just checking membership in a set.

Breaking that part of the conditional out into a local variable also lets us name it, which is helpful for documentation.

Looking at the HTML spec, for elements with a Text content model, I wonder if we should also check for `<title>`, `<style>`, `<rp>`, and an `<option>` with a label attribute but not a value attribute

AndrewJakubowicz: Is this something to improve in lit-html as well (regarding the other elements)?

I can refactor compiler.

augustjk (approved):

PR #4097, opened 2023-08-14
[labs/react] Prefer provided event handler types over React's built-in type

### LINKED ISSUE #4093: [labs/react] 2.0 custom event typing conflicts with common/global react events

React (@lit-labs/react)

### Description

In previous versions of `@lit-labs/react` events defined in the `createComponent` call always took precedence type-wise over React types.

This meant wherever we needed to override a global [React event prop](https://react.dev/reference/react-dom/components/common#common-props) we could and our `CustomEvent` typing would be what the user of the component sees. Something like

```typescript
export const ExampleInput = createComponent({
  tagName: 'my-example-input',
  elementClass: MyExampleInput,
  react: React,
  events: {
    onInput: 'onInput' as EventName<CustomEvent<TestInputEventDetail>>,
  },
});
```

would just work:

```tsx
const test = () => {
  return (
    <ExampleInput
      onInput={(e: CustomEvent<TestInputEventDetail>) => {
        console.log(e);
      }}
    />
  );
};
```

In `@lit-labs/react` these two types are now combined with an intersection which means users trying to handle our custom events are hit with a typescript error no matter what for any of the native bindings that we are attempting to override.

image

Previously this worked because the React wrapper component would set the native event listener for us, which would override the React synthetic event. Is that still happening here and this is just a typing issue, or are both firing somehow now?

### Reproduction

**Stackblitz recreation:** [https://stackblitz.com/edit/withastro-astro-7easvq?file=src%2Fcomponents%2FReactWrapper.tsx](https://stackblitz.com/edit/withastro-astro-7easvq?file=src%2Fcomponents%2FReactWrapper.tsx) (let `npm install` finish to see type error)

### Workaround

I could probably figure out a way to omit or override the provided type where these conflicts happen and instead, only export the CustomEvent handler function type, but it'd probably be nicer if we could tweak the internal typing to do this like it used to.

### Is this a regression?

### Affected versions

Failing in `2.0.0`. Worked in `<= 1.2.0`

Browser: N/A, Chrome
OS: MacOS, Liunux
Node: 18.16.0, 16.20.0
NPM: 9.5.1, 9.4.2

augustjk:
Fixes #4093 

This was a regression introduced the 2.0.0 update to refactor `createComponent` and simplify types.

Added back the removed `Omit` from `React.HTMLAttributes`
https://github.com/lit/lit/pull/4027/files#diff-70e2e47659157c4129a63acfad45aa00d53e5314c322b4acc9e677cb1cbaf97aL27
and added a build-time type test for this.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -48,8 +48,13 @@ export type ReactWebComponent<
 // lifecycle methods or allow user to explicitly provide props.
 type ElementProps<I> = Partial<Omit<I, keyof HTMLElement>>;
 
-// Acceptable props to the React component.
-type ComponentProps<I, E extends EventNames = {}> = React.HTMLAttributes<I> &
+// Acceptable props to the React component. Omit keyof E from HTMLAttributes to
+// prefer provided event handler mapping over React's built-in event handler
+// props.
+type ComponentProps<I, E extends EventNames = {}> = Omit<
+  React.HTMLAttributes<I>,
+  keyof E
+> &
   ElementProps<I> &
   EventListeners<E>;
 
--- a/.changeset/kind-deers-act.md
+++ b/.changeset/kind-deers-act.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/react': patch
+---
+
+Fix type regression to prefer provided event handler prop type over React's built-in handler type.
--- a/packages/labs/react/src/test/create-component_test.tsx
+++ b/packages/labs/react/src/test/create-component_test.tsx
@@ -226,6 +226,25 @@ suite('createComponent', () => {
     <x-foo ref={React.createRef()}></x-foo>;
   });
 
+  // Type only test to be caught at build time.
+  test.skip('Prefer passed in events over built-in React event types', async () => {
+    const TestComponent = createComponent({
+      react: React,
+      tagName: 'event-component',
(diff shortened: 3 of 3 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -48,8 +48,13 @@ export type ReactWebComponent<
 // lifecycle methods or allow user to explicitly provide props.
 type ElementProps<I> = Partial<Omit<I, keyof HTMLElement>>;
 
-// Acceptable props to the React component.
-type ComponentProps<I, E extends EventNames = {}> = React.HTMLAttributes<I> &
+// Acceptable props to the React component. Omit keyof E from HTMLAttributes to
+// prefer provided event handler mapping over React's built-in event handler
+// props.
```
justinfagnani, line 53: Does this actually _prefer_ the provided types or just remove all the default event types? ie, does `onClick` still remain if it's not defined in the createComponent events?

augustjk: It'll only remove the default event type if a user provided an `events` option with a conflicting name like `onClick` which is captured in the `E` type. Otherwise, nothing will be omitted.

justinfagnani: oh right... sorry I was reading the type wrong!

justinfagnani (approved):

AndrewJakubowicz:
Nice!

PR #4109, opened 2023-08-16
Update ts-clone-node to latest

justinfagnani:
ts-clone-node now officially supports TypeScript 5.0, reducing duplication of typescript in our node modules.


augustjk (approved):
Excellent!

PR #4110, opened 2023-08-16
Add missing tests to CI

rictic:


augustjk (dismissed):
It is quite confusing and maybe it should change but, as it is, labs/context, labs/motion, labs/observers, labs/router, labs/scoped-registry-mixin, and labs/task have their test run as part of packages/tests:test
https://github.com/lit/lit/blob/78e9dd1d84564e2cabdf29bb6075f720bbec56ef/packages/tests/src/web-test-runner.config.ts#L165-L175

rictic:
Huh, that doesn't seem good. That leaves us with two ways of running the same tests, and it means our tests are less granular and parallelizable. What's the advantage, we can avoid paying the spinup cost for each test separately? Given that most changes don't touch most of these packages I think the caching win will be larger, save for changes to core.

justinfagnani:
Oh right... that setup was to have a command that launched a WTR process and single sauce connection for all tests. It was pretty critical in getting our flakiness down.

I think if we got the sauce github action working, we might be able to remove that hack.

Mandown745 (approved):

rictic:
Good catch! Updated to only include tests that aren't run as part of ./packages/tests:test

AndrewJakubowicz (approved):
Awesome! GitHub Actions is breaking because gen-manifest golden is out of sync.

Would it be worth merging this into `main` instead of `3.0`?

Goldens were recently updated on `main`, so that may fix the test failure: https://github.com/lit/lit/pull/4177

rictic:
Rebased onto `main`

PR #4111, opened 2023-08-17
[examples] Add examples with Preact project type test

### LINKED ISSUE #3289: [labs/react] Figure out pattern and/or fix types for using a wrapped component in React and Preact
### Should this be an RFC?

React (@lit-labs/react)

### Description

[A bug report on Shoelace](https://github.com/shoelace-style/shoelace/issues/868) indicates that our wrapper typings don't work well in Preact when the wrappers have been made by the component author for React - this is not a case where the Preact API was given to `createComponent()`.

Unknown

augustjk:
### Changes
- Convert top level `examples/` directory to be part of npm workspaces
  - Ultimately this will make it easier to add examples that make use of monorepo packages and run tests on them
  - This required loosening nextjs example dependencies and deduplicating packages
- Add `test-elements-react` package to `labs/test-projects/` that include a subset of components from `test-element-a` wrapped with `@lit-labs/react`
- Add `preact` example project to top level `examples/` directory that import `test-element-react` and has a `tsc --noEmit` test to make sure types are compatible.

Resolves #3289


packages/labs/react/examples/preact/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  EventSubclass,
+  MyDetail,
+  SpecialEvent,
+} from '@lit-internal/test-elements-react/element-events.js';
+import {ElementProps} from '@lit-internal/test-elements-react/element-props.js';
+
+export function App() {
+  return (
+    <>
+      <ElementA foo="foo" onAChanged={() => {}} />
```
AndrewJakubowicz, line 25: Could this example also pass children into the default slot and `footer` slot?

packages/labs/react/examples/preact/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "ESNext",
+    "moduleResolution": "NodeNext",
+    "noEmit": true,
```
AndrewJakubowicz, line 6: Test is already explicitly setting `--noEmit`. Is this needed?

packages/labs/react/examples/test-elements-react/src/element-a.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import {createComponent, EventName} from '@lit-labs/react';
+
+import {ElementA as ElementAElement} from '@lit-internal/test-element-a/element-a.js';
+
```
AndrewJakubowicz, line 11: Nit: document public export

rollup-common.js
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import summary from 'rollup-plugin-summary';
+import {summary} from 'rollup-plugin-summary';
```
AndrewJakubowicz, line 7: Why is this necessary? This seems out of scope for this PR?

AndrewJakubowicz (commented):
This is looking great!

justinfagnani:
@augustjk why `packages/labs/react/examples` instead of the top-level examples folder we already have?

rollup-common.js
augustjk: Adding the example projects (specifically the preact project with vite) nested under `packabes/labs/react` directory created `packages/labs/node_modules` that included rollup v3. The way rollup handles config files in v3 requires this change for the `build:rollup` script for `labs/react` package. This thankfully still works with other packages' rollup v2. I considered updating our whole repo to use rollup v3 but since this works for now, decided not to.

AndrewJakubowicz: Ah! Makes sense! Also yay for backwards compatibility :D

augustjk:
> @augustjk why `packages/labs/react/examples` instead of the top-level examples folder we already have?

My original thinking was these are specifically for testing `labs/react` package and should be included under it.

Top level examples folder does not participate in npm workspaces so we can't reference by internal package name, and we don't run any tests against those. Maybe that should change though.

The original intent of not making top level examples not be part of npm workspaces was we didn't want to bloat the monorepo `npm install` and `npm build` with all these example projects and their dependencies like `next` and `vite`, and to make each example have their own `node_modules` directory for better isolation. I can be convinced to change top level examples to npm workspace and move the examples there since the goal would be to add tests that run in CI for all examples, so they should be installed anyway and maybe the isolation of `node_modules` is less of a concern if npm workspace handles versioning properly.

packages/labs/react/examples/preact/tsconfig.json
augustjk: This is actually default for vite projects as vite doesn't use `tsc` but `esbuild` for transpiling only. I would consider the `--noEmit` in the test script redundant but added for the sake of explicitly signaling we're running `tsc` for type checking only.

AndrewJakubowicz: Excellent!

packages/labs/compiler/package.json
```diff
@@ -106,7 +106,7 @@
     "@parse5/tools": "^0.2.0",
     "lit-html": "^2.7.5",
     "parse5": "^7.1.2",
-    "typescript": "^5.1.6"
+    "typescript": "~5.2.0"
```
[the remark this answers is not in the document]
AndrewJakubowicz: Thank you so much!

examples/preact/index.html
AndrewJakubowicz, line 1: opt nit: Omit optional tags: https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags

```html
<!DOCTYPE html>
<meta charset="UTF-8" />
<meta name="viewport" content="width=device-width, initial-scale=1.0" />
<meta name="color-scheme" content="light dark" />
<title>Preact with @lit-labs/react</title>

<div id="app"></div>
<script type="module" src="/src/index.tsx"></script>
```

examples/preact/src/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+} from '@lit-internal/test-elements-react/element-events.js';
+import {ElementProps} from '@lit-internal/test-elements-react/element-props.js';
+
+export function App() {
+  return (
+    <>
+      <ElementA foo="foo" onAChanged={() => {}}>
+        This goes in default slot
+        {/* TODO(augustjk): Add this back with #4142 */}
+        {/* <div slot="stuff">This goes in stuff slot</div> */}
```
AndrewJakubowicz, line 28: Nice!

packages/labs/test-projects/test-elements-react/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@lit-internal/test-elements-react",
+  "type": "module",
```
AndrewJakubowicz, line 3: Add private: true please.

Is it also worth documenting in a `description` field that this is essentially the output of the react generator checked in for testing purposes?

packages/labs/test-projects/test-elements-react/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "compilerOptions": {
```
AndrewJakubowicz, line 2: No action comment. I wonder if in the future we could consider creating a couple base tsconfigs which are extended. Then we'll have less tsconfig duplication.

AndrewJakubowicz (approved):
Excellent

examples/preact/index.html
augustjk: Ooh thanks for linking the style guide. TIL!

packages/labs/test-projects/test-elements-react/package.json
augustjk: Thanks for catching that! Description vs README... I guess it's simple enough that a description would be fine.

packages/labs/test-projects/test-elements-react/tsconfig.json
augustjk: Agree. That would make it easy for all of our packages to be consistent, and have one point for updating them all. Something to consider in the future.

packages/labs/react/examples/preact/src/index.tsx
augustjk: Added todo item that'll be fixed in #4142

packages/labs/react/examples/test-elements-react/src/element-events.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import * as React from 'react';
+import {createComponent, EventName} from '@lit-labs/react';
+
+import {ElementEvents as ElementEventsElement} from '@lit-internal/test-element-a/element-events.js';
+import {MyDetail} from '@lit-internal/test-element-a/detail-type.js';
+import {EventSubclass} from '@lit-internal/test-element-a/element-events.js';
+import {SpecialEvent} from '@lit-internal/test-element-a/special-event.js';
+export type {MyDetail} from '@lit-internal/test-element-a/detail-type.js';
```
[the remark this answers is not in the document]
augustjk: Done. As the basis for this was the wrapper generator, it just naively did the exports individually.

PR #4113, opened 2023-08-17
attempt to fix typings for React refs

KonnorRogers:
An attempt to fix issues related to refs when using React + Shoelace.

Relevant issue: https://github.com/shoelace-style/shoelace/issues/1509
PR on shoelace: https://github.com/shoelace-style/shoelace/pull/1526

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -39,8 +39,7 @@ export type ForwardComponent<
     ref?: React.ForwardedRef<Element>;
   }
 > &
-  React.FC<ReactComponent>;
-
+  Omit<React.FC<ReactComponent>, 'ref'>;
 /**
  * Type of the React component wrapping the web component. This is the return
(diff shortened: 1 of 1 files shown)


augustjk:
@KonnorRogers thanks for the PR but I don't think this fix here is needed for upstream `@lit-labs/react`. The returned component from `createComponent` as is works with React refs as it's already made with `React.forwardRef` and typed as such. See https://stackblitz.com/edit/vitejs-vite-kwojn7?file=src%2FApp.tsx

Looking at https://github.com/shoelace-style/shoelace/issues/1510 and trying a bit on my own, the issue might be due to the extra wrapping React component that adds `Component.define(tagName)` in the constructor, which actually isn't using `React.forwardRef` to property pass the user's ref to the underlying component it's wrapping.

KonnorRogers:
👍 thanks! Let me remove the wrapper and go back and double check!

PR #4130, opened 2023-08-22
[labs/virtualizer] Fix issue with position: fixed scrollers

### LINKED ISSUE #4125: [labs/virtualizer] Fixed-position virtualizer not refreshing items on scroll

### Description

Works in 2.0.3 but since the virtualizer stops refreshing (i.e. stops adding new content / updating the list) when scrolling in a popup container.

It only seems to be an issue for virtualizer listings in "popups". I.e. only an issue when the container is fixed or absolute positioned or something like that.

2.0.3
Screenshot 2023-08-22 at 12 27 20

2.0.6
Screenshot 2023-08-22 at 12 27 42

### Reproduction

I can't do this easily. It is deep in my application.

### Workaround

I haven't been able to find a workaround.

### Is this a regression?

### Affected versions

Failing in 2.0.4+ and not in 2.0.3

Chrome Version 116.0.5845.97 (Official Build) (64-bit) on Windows.

graynorton:
Fixes #4125, which was a regression caused by #3976 (a fix for #3904).

When connected, a virtualizer walks up the DOM tree, visiting its ancestor elements and determining which of them are "clipping ancestors" (ancestor elements whose scroll positions may affect the visibility of the virtualizer itself). In principle, virtualizer should stop looking for clipping ancestors as soon as it comes to a `position: fixed` ancestor, since fixed positioning is always relative to the document, making any higher-level ancestors irrelevant.

Prior to #3976, we didn't pay attention to whether ancestors were `position: fixed`, causing the issue described in #3904.

However, the logic in #3976 wasn't quite right; it excluded the `position: fixed` ancestor itself, along with its higher-level ancestors. In the case where the `position: fixed` ancestor was itself a scroller, this caused virtualizer not to listen for `scroll` events and therefore not re-render children when scrolling occurred (as described in #4125).

Fixed the logic and added regression tests for both issues.

### DIFF
--- a/packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
@@ -1,0 +1,119 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {array, ignoreBenignErrors} from '../helpers.js';
+import {LitVirtualizer} from '../../lit-virtualizer.js';
+import {grid} from '../../layouts/grid.js';
+import {styleMap} from 'lit/directives/style-map.js';
+import {expect, html, fixture} from '@open-wc/testing';
+
+interface NamedItem {
+  name: string;
+}
+
+const _100Items = array(100).map((_, i) => ({name: `Item ${i}`}));
+
+describe('Virtualizer behaves properly when it has a position: fixed ancestor', () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  // Regression test for https://github.com/lit/lit/issues/3904
+  it('should calculate its viewport correctly', async () => {
+    // Our outer container is overflow: hidden, so would normally be
+    // a "clipping ancestor" of our virtualizer. Since it has a
+    // width of zero, it would cause the virtualizer to be invisible.
+    const containerStyles = {
+      overflow: 'hidden',
+      width: '0',
+    };
+
+    // But since this scrolling element is position: fixed, it will
+    // be positioned relative to the document, essentially "popping out"
+    // of its zero-width parent. This means the virtualizer will be
+    // visible and have some space to work with, after all.
+    const scrollerStyles = {
+      position: 'fixed',
+      width: '200px',
+      height: '200px',
+    };
+
+    // We use the grid layout here because it conveniently will not
+    // render any items if it calculates that it doesn't have sufficient
+    // space, making our testing job easier.
+    const container = await fixture(html`
+      <div id="container" style=${styleMap(containerStyles)}>
+        <div id="scroller" style=${styleMap(scrollerStyles)}>
+          <lit-virtualizer
+            .layout=${grid({
+              itemSize: '100px',
+            })}
+            .items=${_100Items}
+            .renderItem=${({name}: NamedItem) => html`<div>${name}</div>`}
+          ></lit-virtualizer>
+        </div>
+      </div>
+    `);
+
+    const scroller = container.querySelector('#scroller')!;
+    expect(scroller).to.be.instanceOf(HTMLDivElement);
+    const virtualizer = container.querySelector('lit-virtualizer')!;
+    expect(virtualizer).to.be.instanceOf(LitVirtualizer);
+
+    // If virtualizer has properly ignored the zero-width ancestor of our
+    // fixed-position scroller, some children will be rendered; otherwise, not.
+    //
+    // In practice, we'll time out if we fail here because the `layoutComplete`
+    // promise will never be fulfilled.
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 0');
+  });
+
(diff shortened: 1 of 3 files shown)


packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    const virtualizer = container.querySelector('lit-virtualizer')!;
+    expect(virtualizer).to.be.instanceOf(LitVirtualizer);
+
+    // If virtualizer has properly ignored the zero-width ancestor of our
+    // fixed-position scroller, some children will be rendered; otherwise, not.
+    //
+    // In practice, we'll time out if we fail here because the `layoutComplete`
+    // promise will never be fulfilled.
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 0');
```
augustjk, line 70: Considering a failure's a failure so I'm okay either way, but what are your thoughts on using `pass` here since the failure message would be more explicit?

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    // If the position: fixed scroller has properly been recognized as
+    // a clipping ancestor, then virtualizer will re-render as scrolling
+    // occurs; otherwise, not.
+    //
+    // In practice, we'll time out if we fail here because the `layoutComplete`
+    // promise will never be fulfilled.
+    scroller.scrollTo(0, scroller.scrollHeight);
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 99');
```
augustjk, line 117: Interestingly replacing both of the `await virtualizer.layoutComplete` here with `pass()` makes this fail. There must be something about starting a `scroller.scrollTo()` specifically after awaiting `layoutComplete` that looking for `'Item 0'` alone isn't enough.

augustjk (approved):
Excellent fix and test additions!

I had a some questions about whether we might not need the awaiting of `layoutComplete` but seems like it is indeed needed at least when we want to test scrolling.

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
graynorton: I prefer to avoid using `pass` except in cases where we really can't avoid it; instead, I'd prefer to make sure `layoutComplete` is as reliable as possible and use that for cases where we want to check the state of things after a virtualizer renders.

That said, I was also bothered by the fact that this test failed with a timeout rather than something more akin to the failure you'd observe in manual testing, so I did a bit of digging. It turns out that there were two quirks in the `layoutComplete` implementation that resulted in the timeout in this case:

- When you provided a layout explicitly, rather than relying on the default layout to be dynamically loaded, the initial layout pass would complete before you had a chance to request the `layoutComplete` promise.

- If a layout pass resulted in no children being rendered (as in the failure case for this test), `layoutComplete` would never resolve.

I have made tweaks to address both of these quirks and pushed a new commit to this branch. There is a decent chance that making this change will require users to update some existing tests; specifically, if they are rendering a virtualizer and immediately checking the state of its initial render (without any delay or retry logic), they will need to insert an `await virtualizer.layoutComplete`. (I had to do this to two of our existing tests.) So we need to decide whether we want to go ahead (I'm inclined to) and, if so, whether we call this a breaking change for the Virtualizer release (I'm open to discussion).

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
graynorton: Yeah, it is possible in a testing context to initiate a scroll while virtualizer is still in the middle of a layout pass, and this can cause unpredictable results. Waiting for `layoutComplete` avoids this situation, which is another reason why I'd like to lean on it more heavily in our tests.

In my updated version of this test, I added some ad hoc logic to race the `layoutComplete` promise against a short timeout so that we don't fail with a timeout; instead, we fail because the rendered state of the virtualizer is not correct (which aligns nicely with how you'd observe this failure in manual testing).

graynorton:
@augustjk Thanks for your review! I made some changes in response to your comments, though a bit different from the ones you proposed. See my detailed comments above. It's not obvious whether we should proceed with the changes in my last commit, so please take a look and let's discuss.

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
augustjk: I'm torn. I think the change to `layoutComplete` is worth discussing but probably better to be split from this PR just so we can land the `position: fixed` scroller fix first.

Like the comment you added, it feels bad to introduce a delay purely for the sake of a promise whose main use is for testing.

graynorton: Yeah, that makes sense. I'll roll back that commit here and make the `layoutComplete` changes into a separate PR.

PR #4132, opened 2023-08-23
fix(analyzer): jsdoc type order, lowercase tags, css syntax

bennypowers:
- adds tests
- support `@cssproperty`, `@cssprop`, `@csspart` (all lowercase)
- support `@fires {Type} event-name - description`
- support `@cssProp {<syntax>} [--name=default] - description`

```js
/**

 * @fires event-name
 * @fires event-name description
 * @fires event-name - description
 * @fires event-name: description
 * @fires event-name {Type}
 * @fires event-name {Type} description
 * @fires event-name {Type} - description
 * @fires event-name {Type}: description
 * @fires {Type} event-name
 * @fires {Type} event-name description
 * @fires {Type} event-name - description
 * @fires {Type} event-name: description
 * @slot name
 * @slot name description
 * @slot name - description
 * @slot name: description
 * @cssProp [--name=default]
 * @cssProp [--name=default] description
 * @cssProp [--name=default] - description
 * @cssProp [--name=default]: description
 * @cssprop {<color>} [--name=default]
 * @cssprop {<color>} [--name=default] description
 * @cssprop {<color>} [--name=default] - description
 * @cssprop {<color>} [--name=default]: description
 */

### DIFF
--- a/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
+++ b/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
@@ -141,12 +141,13 @@ export const getJSDocData = (
           addNamedJSDocInfoToMap(slots, tag, analyzer);
           break;
         case 'cssProp':
-          addNamedJSDocInfoToMap(cssProperties, tag, analyzer);
-          break;
+        case 'cssprop':
         case 'cssProperty':
+        case 'cssproperty':
           addNamedJSDocInfoToMap(cssProperties, tag, analyzer);
           break;
         case 'cssPart':
+        case 'csspart':
           addNamedJSDocInfoToMap(cssParts, tag, analyzer);
           break;
       }
--- a/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
+++ b/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
@@ -20,7 +20,10 @@ import {
   NamedDescribed,
 } from '../model.js';
 import {addEventsToMap} from './events.js';
-import {parseNodeJSDocInfo, parseNamedJSDocInfo} from '../javascript/jsdoc.js';
+import {
+  parseNodeJSDocInfo,
+  parseNamedTypedJSDocInfo,
+} from '../javascript/jsdoc.js';
 
 const _isCustomElementClassDeclaration = (
   t: ts.BaseType,
@@ -112,7 +115,7 @@ const addNamedJSDocInfoToMap = (
   tag: ts.JSDocTag,
   analyzer: AnalyzerInterface
 ) => {
-  const info = parseNamedJSDocInfo(tag, analyzer);
+  const info = parseNamedTypedJSDocInfo(tag, analyzer);
   if (info !== undefined) {
     map.set(info.name, info);
   }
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: packages/labs/analyzer/src/lib/javascript/jsdoc.ts [hunk 1]
```diff
@@ -151,32 +162,45 @@ function makeDashParseDiagnostic(
...
+
+  const info: NamedDescribed | NamedDescribedSyntax = {name};
   if (description?.length > 0) {
     info.description = normalizeLineEndings(description);
   }
   if (defaultValue?.length > 0) {
     info.default = defaultValue;
   }
+  if (
+    tag.tagName.text.toLowerCase().startsWith('cssprop') &&
```
justinfagnani, line 199: Rather than special-casing cssprop, can we treat the CSS syntax as a type? In the jsdoc format it comes in exactly the same place as type. Then we could reuse parseNamedTypedJSDocInfo for css properties, and just move the type into a syntax field of a CSSPropertyInfo object.

justinfagnani (commented):
This looks great! Amazing tests!

My only question is about the return type of the jsdoc parsing and why we can't use parseNamedTypedJSDocInfo for parsing CSS properties.

packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
```diff
@@ -141,12 +141,13 @@ export const getJSDocData = (
           addNamedJSDocInfoToMap(slots, tag, analyzer);
           break;
         case 'cssProp':
-          addNamedJSDocInfoToMap(cssProperties, tag, analyzer);
-          break;
+        case 'cssprop':
```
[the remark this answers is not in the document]
bennypowers: i initially lowercased the tag, but like ... `@cSsPRop`...

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
```diff
@@ -47,7 +47,7 @@ const parseNameTypeDescRE =
 
 // Regex for parsing type, name, and description from JSDoc comments
 const parseTypeNameDescRE =
-  /^\{(?<type>[^}]+)\}\s+\[?(?<name>[^{}[\]\s=]+)(?:=(?<defaultValue>[^\]]+))?\]?(?:\s+-\s+)?(?<description>[\s\S]*)$/m;
+  /^\{(?<type>.+)\}\s+\[?(?<name>[^{}[\]\s=]+)(?:=(?<defaultValue>[^\]]+))?\]?(?:\s+-\s+)?(?<description>[\s\S]*)$/m;
```
justinfagnani, line 50: I think we're in luck here in that CSS property type syntax doesn't include `}`, but _event_ type syntax definitely can. Do we have tests for that? I think we'll need a way to escape `}`.

No need to fix or change anything now, but let's come back to that.

justinfagnani (approved):
Loogs great!

justinfagnani:
Thanks a lot @bennypowers! I'll merge as soon as the tests pass.

packages/labs/analyzer/src/lib/javascript/jsdoc.ts
bennypowers: Well they're all on the same code path now, so I think the existing tests should cover it, no?

justinfagnani: Yeah, we have that [here](https://github.com/lit/lit/blob/b9c3443185ef2f7bbe728dc61eb4b13b7a0dccb4/packages/labs/analyzer/test-files/js/events/element-a.js#L34) so we're good!

PR #4133, opened 2023-08-23
[infra] update benchmark generator

AndrewJakubowicz:
### Context

For easier and more robust benchmarking, the generator is a fantastic tool for changing the parameters of template-heavy.

This change updates some old dependencies and ensures that the generator works.

### Fix

Some dependency updates, tsconfig update, and added a license to the generated output.

### How to test

Check out the branch. `git checkout origin/update-benchmark-generator`.

Then:

```sh
$ cd packages/benchmarks/generator/
$ ./scripts/generate.sh
Generating lit-html/template-heavy
Generating variant template-heavy
Done.
```

Note that `template-heavy.html` and `template-heavy.js` has been updated.

### Test plan

Also tested manually with small options to generate a tiny file. E.g. by setting width and depth to 1.

```sh
node generator/build/index.js  -r lit-html,lit-html@tot=generator/scripts/package-versions-tot.json,lit-html@release=generator/scripts/package-versions-release.json -n template-heavy -o lit-html/template-heavy -u 10 -q -m render,update -g -w 1 -d 1
```

which generated:

```ts
/**
 * @license
 * Copyright 2023 Google LLC
 */

/*
Benchmark generated via the following invocation:
node generator/build/index.js -r lit-html,lit-html@tot=generator/scripts/package-versions-tot.json,lit-html@release=generator/scripts/package-versions-release.json -n template-heavy -o lit-html/template-heavy -u 10 -q -m render,update -g -w 1 -d 1

Parameters:
  renderers: lit-html
  depth: 1
  width: 1
  attrPerNode: 4
  dynAttrPct: 0.5
  valPerDynAttr: 2
  dynNodePct: 0.5
  updateCount: 10
  updateNodePct: 1
  updateAttrPct: 1
  uniqueTemplates: true
  measure: render,update
*/

  import {html, render as litRender} from 'lit-html';
  
const render = () => html`<div a0_0="0_0" a0_1="0_1" a0_2="${'0_2:0#' + updateId}-${'0_2:1#' + updateId}" a0_3="${'0_3:0#' + updateId}-${'0_3:1#' + updateId}">0</div>`;

const container = document.getElementById('container');
let updateId = 0;
let useTemplateA = true;
window.update = () => {
  litRender(render(), container);
  updateId++;
  useTemplateA = !useTemplateA;
}
performance.mark('initial-render-start');
window.update();
performance.mark('initial-render-end');
for (let i=0; i<10; i++) {
  window.update();
}
performance.mark('updates-end');
performance.measure('render', 'initial-render-start', 'initial-render-end');
performance.measure('update', 'initial-render-end', 'updates-end');
performance.measure('time', 'initial-render-start', 'updates-end');
window.tachometerResult = performance.getEntriesByName('render')[0].duration +
      performance.getEntriesByName('update')[0].duration;
document.title = window.tachometerResult.toFixed(2) + 'ms';
```


packages/benchmarks/generator/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/react": "^16.9.36",
     "@types/react-dom": "^16.9.8",
     "@types/rimraf": "^3.0.0",
+    "@web/dev-server": "^0.3.0",
```
augustjk, line 29: Maybe it's not really needed, but in the code this replaced `es-dev-server`, should the npm scripts here also be replaced with `wds` and `es-dev-server` dependency removed?

augustjk (commented):
I was also thinking whether this should be added to the npm workspace so it doesn't get its own package-lock and we might dedupe some deps, but maybe that's too extra? I don't feel that strongly about it so am okay with just these minimal changes to get it to run.

packages/benchmarks/generator/package.json
```diff
@@ -40,6 +41,7 @@
   },
   "devDependencies": {
     "puppeteer": "^15.4.0",
-    "tachometer": "^0.7.0"
+    "tachometer": "^0.7.0",
+    "typescript": "^4.7.4"
```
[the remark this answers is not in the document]
AndrewJakubowicz: Yes it totally should be!

packages/benchmarks/generator/package.json
AndrewJakubowicz: Oh! I thought I had done that. I must have somehow reverted it. Great eye!

packages/benchmarks/generator/src/deoptigate.ts
```diff
@@ -38,13 +37,16 @@ export const deoptigate = async (
   url: string,
   open = false
 ) => {
-  const config = createConfig({
+  const config = {
```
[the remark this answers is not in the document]
AndrewJakubowicz: Nice – done!

packages/benchmarks/generator/package.json
AndrewJakubowicz: Fixed

AndrewJakubowicz:
> I was also thinking whether this should be added to the npm workspace so it doesn't get its own package-lock and we might dedupe some deps?

I agree with you, and can investigate in a followup.

justinfagnani:
> > I was also thinking whether this should be added to the npm workspace so it doesn't get its own package-lock and we might dedupe some deps?
> 
> I agree with you, and can investigate in a followup.

I think this is a great idea.

packages/benchmarks/generator/package.json
```diff
@@ -14,8 +14,8 @@
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc --watch",
-    "serve": "es-dev-server --node-resolve --dedupe --preserve-symlinks",
-    "serve:open": "es-dev-server --node-resolve --dedupe --preserve-symlinks --open generated/index.html"
+    "serve": "web-dev-server --node-resolve --dedupe --preserve-symlinks",
+    "serve:open": "web-dev-server --node-resolve --dedupe --preserve-symlinks --open generated/index.html"
```
justinfagnani, line 18: did you double-check all the flags? I _think_ they're the same

justinfagnani (approved):

packages/benchmarks/generator/package.json
AndrewJakubowicz: They are the same except `dedupe` has been removed.  I can remove.

PR #4134, opened 2023-08-23
[infra] add `@lit-internal/benchmarks-generator` to monorepo workspace

AndrewJakubowicz:
### Context

Follow up change to address @augustjk 's excellent comment: https://github.com/lit/lit/pull/4133#pullrequestreview-1592214642

### Test plan

Ensured the package.json `npm run` commands continued to work.
Otherwise if GitHub Actions pass this change should be ok.


packages/benchmarks/generator/package.json
```diff
@@ -39,6 +31,14 @@
     "table": "^5.4.6"
   },
   "devDependencies": {
+    "@types/command-line-args": "^5.0.0",
+    "@types/command-line-usage": "^5.0.1",
+    "@types/js-combinatorics": "^0.5.32",
+    "@types/mkdirp": "^1.0.1",
+    "@types/puppeteer": "^3.0.2",
+    "@types/react": "^16.9.36",
+    "@types/react-dom": "^16.9.8",
```
augustjk, line 40: maybe out of scope for this PR but since we're here are these deps needed? this package apparently also has `preact` as a dep but i can't really find its usage.

augustjk (dismissed):
Approving as far as adding to the workspace, but maybe we have an opportunity to clean up more deps here?

packages/benchmarks/generator/package.json
AndrewJakubowicz: That's a great point – agree there is room for cleanup. `preact` is used here (in an indirect way): https://github.com/lit/lit/blob/3f620eb47b37b7b2532984bd92c839649bd6ef35/packages/benchmarks/generator/src/renderers.ts#L378-L386

I'm not sure why these `@types` are needed though.

AndrewJakubowicz: I speculate that they are needed after generating files that use the various renderers.

augustjk: ah, okay. in that case they should stay, but they also hardcode the locations with `../node_modules/` so if the dep gets hoisted this is problematic.

AndrewJakubowicz: Great point. Closing the PR temporarily.

I need to spend a bit more time understanding the scope and usages of this package. It seems to do quite a lot, with generating template-heavy being a small feature.

augustjk: we can't guarantee the dependency location when it's a workspace but we could workaround by doing something like
```js
import {createRequire} from 'node:module';

createRequire(import.meta.url).resolve('preact')
```
to grab the location and piece together to try and dig into the /dist/ directory and all that. but maybe that seems like too much effort for this.

PR #4139, opened 2023-08-24
[lit, lit-element, reactive-element] Prevent standard decorators from being published

augustjk:
As we're considering the standard decorator migration path and potentially creating a hybrid decorator, we don't want to publish them just yet.

We can revert this when we want them published later.

Note: There was a typo in reactive-element's package.json files entry for "/std-decorators/" so this'll need to be fixed if/when we revert this commit.


augustjk:
Oops, tests are breaking because of the change in exports breaks bare module imports.

augustjk:
This is no longer needed. We'll do the next prerelease when hybrid decorators are ready.

PR #4140, opened 2023-08-24
[reactive-element] 3.0 hybrid decorators

justinfagnani:

Make a single set of decorators that work in TypeScript with `experimentalDecorators` set to either true or false, and in Babel with `version` equal to "2023-05" (not working because of [lack of metadata](https://github.com/babel/babel/issues/15889)). This works by duck-typing the decorator arguments.

After this effort we will have one set of decorators that work under the following cases:

1. TypeScript with `experimentalDecorators: true` and no accessor keyword
2. TypeScript with `experimentalDecorators: true` and accessor keywords
3. TypeScript with `experimentalDecorators: false` and accessor keywords
4. Babel with `experimentalDecorators: false` and accessor keywords

The trade-off is that we have to have the same behavior in each configuration:

- We can't skip reflection for initial property values
- We have to wrap accessors as we can't detect the difference between auto-accessors and hand-written accessors with experimental decorators

Tasks:
- [x] Make legacy-decorators work with `experimentalDecorators: false` and pass the std-decorators tests
- [ ] Unify the behavior
  - [ ] Reflect initial property values to attributes
  - [ ] Always wrap accessors
- [ ] Remove std-decorator and legacy-decorator files, but not std-decorator _tests_.
- [ ] Run decorators against std-decorator tests with `experimentalDecorators: true`.
- [ ] Re-enable Babel tests when https://github.com/babel/babel/issues/15889 is fixed.
- [ ] Add a compiler-time flag to choose one or both decorator implementations to enable code elimination
- [ ] Ensure we have tests for `noAccessor: true` in all cases
- [ ] Deprecate and add a dev-mode warning for overrides of `createProperty` and `getPropertyDescriptor`


rictic (approved):

packages/reactive-element/tsconfig.std-decorators-tests.json
```diff
@@ -29,7 +29,6 @@
   "include": [
     "../../node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.d.ts",
     "src/env.d.ts",
```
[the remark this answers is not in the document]
justinfagnani: I'm working on running the std-decorators tests (using `accessor`) with `experimentalDecorators: true`. `experimentalDecorators: true` and `useDefineForClassFields: true` on regular class fields however won't work, due to the field shadowing any accessor on the prototype.

packages/reactive-element/src/legacy-decorators/state.ts
```diff
@@ -35,6 +35,7 @@ export interface InternalPropertyDeclaration<Type = unknown> {
 export function state(options?: InternalPropertyDeclaration) {
   return property({
     ...options,
+    attribute: false,
```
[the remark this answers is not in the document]
justinfagnani: In the std-decorators implementation, the only thing that matters is `attribute: false`. `state: true` exists for DX within the static properties block when not using decorators.

I think that the `state` option should be excluded from the options available to `@property()`. We want people to use `@state()` instead.

packages/reactive-element/src/legacy-decorators/query.ts
```diff
@@ -41,32 +65,75 @@ const DEV_MODE = true;
...
       }
-      return descriptor;
-    },
-  });
+      return;
+    }
+  }) as QueryDecorator;
 }
+
+// export function query(selector: string, cache?: boolean) {
```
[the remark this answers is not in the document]
justinfagnani: removed!

packages/reactive-element/src/legacy-decorators/query-assigned-nodes.ts
```diff
@@ -58,20 +82,52 @@ type TSDecoratorReturnType = void | any;
...
+        },
+        enumerable: true,
+        configurable: true,
+      });
+      return;
+    }
+  }) as QueryAssignedNodesDecorator;
 }
+
+// export function queryAssignedNodes(
```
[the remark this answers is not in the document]
justinfagnani: removed

packages/reactive-element/src/legacy-decorators/state.ts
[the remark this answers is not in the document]
rictic: Is there a reason to pass `state: true` here then?

packages/reactive-element/src/legacy-decorators/query-assigned-nodes.ts
```diff
@@ -26,9 +25,34 @@ export interface QueryAssignedNodesOptions extends AssignedNodesOptions {
...
+  <C extends ReactiveElement, V extends Array<Node>>(
+    value: ClassAccessorDecoratorTarget<C, V>,
+    context: ClassAccessorDecoratorContext<C, V>
+  ): void;
+};
+
 // TypeScript requires the decorator return type to be `void|any`.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type TSDecoratorReturnType = void | any;
+// type TSDecoratorReturnType = void | any;
```
AndrewJakubowicz, line 55: Can this be deleted?

packages/reactive-element/src/legacy-decorators/event-options.ts
```diff
@@ -12,7 +12,32 @@
  */
 
 import {ReactiveElement} from '../reactive-element.js';
-import {decorateProperty} from './base.js';
+
+/**
+ * Generates a public interface type that removes private and protected fields.
+ * This allows accepting otherwise compatible versions of the type (e.g. from
```
AndrewJakubowicz, line 18: `compatible` -> `incompatible` (I think is the intent)?

packages/reactive-element/src/legacy-decorators/query-all.ts
```diff
@@ -38,14 +62,26 @@ import {decorateProperty} from './base.js';
...
+      );
+    } else {
+      Object.defineProperty(protoOrTarget, nameOrContext as PropertyKey, {
+        get(this: ReactiveElement) {
+          return this.renderRoot?.querySelectorAll(selector) ?? [];
+        },
+        enumerable: true,
+        configurable: true,
+      });
+      return;
```
AndrewJakubowicz, line 84: Is an empty explicit return required? Can this be omitted.

packages/reactive-element/src/legacy-decorators/custom-element.ts
```diff
@@ -58,8 +62,14 @@ const standardCustomElement = (
...
-  (classOrDescriptor: CustomElementClass | ClassDescriptor) =>
-    typeof classOrDescriptor === 'function'
-      ? legacyCustomElement(tagName, classOrDescriptor)
-      : standardCustomElement(tagName, classOrDescriptor as ClassDescriptor);
+  (tagName: string): CustomElementDecorator =>
+  (
+    classOrTarget: CustomElementClass | unknown,
+    context?: ClassDecoratorContext<Constructor<HTMLElement>>
+  ) =>
+    (typeof classOrTarget === 'function'
```
AndrewJakubowicz, line 70: Later decorators this check swaps to `typeof classOrTarget === 'object'`, and then the first branch is the standard decorator, followed by the legacy decorator in the `else`.

Is one check preferred over the other. No action, just curious.

packages/reactive-element/src/std-decorators/query-async.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
AndrewJakubowicz, line 3: Nit: 2023?

packages/reactive-element/src/test/std-decorators/queryAsync_test.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
AndrewJakubowicz, line 3: Nit: 2023

packages/reactive-element/src/test/std-decorators/queryAsync_test.ts
```diff
@@ -0,0 +1,77 @@
...
+    container.id = 'test-container';
+    document.body.appendChild(container);
+    el = new C();
+    container.appendChild(el);
+  });
+
+  teardown(() => {
+    if (container !== undefined) {
+      container.parentElement!.removeChild(container);
+      (container as any) = undefined;
```
AndrewJakubowicz, line 59: Nit:

```js
container?.remove();
(container as any) = undefined;

**Review by AndrewJakubowicz — APPROVED** [review]
Nice!!! – all comments are non blocking nits

File: packages/reactive-element/src/legacy-decorators/query-assigned-nodes.ts (hunk 5 above)
Comment (justinfagnani) [reply packages/reactive-element/src/legacy-decorators/query-assigned-nodes.ts:55]: yep!

File: packages/reactive-element/src/legacy-decorators/event-options.ts (hunk 6 above)
Comment (justinfagnani) [reply packages/reactive-element/src/legacy-decorators/event-options.ts]: done, and deduplicated all `Interface<T>` definitions

File: packages/reactive-element/src/legacy-decorators/query-all.ts (hunk 7 above)
Comment (justinfagnani) [reply packages/reactive-element/src/legacy-decorators/query-all.ts:76]: It's required by [noImplicitReturns](https://www.typescriptlang.org/it/tsconfig#noImplicitReturns)

File: packages/reactive-element/src/legacy-decorators/custom-element.ts (hunk 8 above)
Comment (justinfagnani) [reply packages/reactive-element/src/legacy-decorators/custom-element.ts:64]: There's no real difference.

File: packages/reactive-element/src/std-decorators/query-async.ts (hunk 9 above)
Comment (justinfagnani) [reply packages/reactive-element/src/std-decorators/query-async.ts:3]: done

File: packages/reactive-element/src/test/std-decorators/queryAsync_test.ts (hunk 11 above)
Comment (justinfagnani) [reply packages/reactive-element/src/test/std-decorators/queryAsync_test.ts:59]: done

File: packages/reactive-element/src/legacy-decorators/property.ts [hunk 12]
```diff
@@ -10,70 +10,45 @@
...
+
+  // legacy decorator signature
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (protoOrDescriptor: Object, name: PropertyKey): any;
 };
 
 const legacyProperty = (
-  options: PropertyDeclaration,
+  options: PropertyDeclaration | undefined,
   proto: Object,
```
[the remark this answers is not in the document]
justinfagnani: It's ultimately not that helpful of a type, but it's `Object` the because the prototype extends from `Object` the value.

PR #4141, opened 2023-08-25
Update every typescript to 5.2.0

justinfagnani:


augustjk (approved):
🎉

AndrewJakubowicz (changes requested):
Please update `vue-tsc` to at least `1.8.8` (from `0.29.8`). That fixes `npm run test` from within `packages/labs/gen-wrapper-vue`.

justinfagnani:
@AndrewJakubowicz I did update vue-tsc to 1.8.8 and still got the error

packages/labs/gen-wrapper-vue/src/lib/package-json-template.ts
```diff
@@ -42,7 +42,7 @@ export const packageJsonTemplate = (
         '@vitejs/plugin-vue': '^3.1.2',
```
AndrewJakubowicz, line 42: Non blocking: The generator sets TypeScript to `~4.7.4` as default. Should this also be bumped to TS 5.2?

AndrewJakubowicz (approved):
Nice!

packages/labs/gen-wrapper-vue/src/lib/package-json-template.ts
justinfagnani: Yeah, I'll bump that back up. We'll only have test coverage of one version at a time, so it should match what we test.

PR #4143, opened 2023-08-26
[labs/react] Loosen type for react module in option

augustjk:
While working making the top level examples directory to part of the npm workspaces here https://github.com/lit/lit/pull/4111, I noticed the code that wraps the custom element with `createComponent()` was failing type check as the `React` type imported in the nextjs example project was a different copy than the one used by `@lit-labs/react`, and due to things like unique symbol type and function overloading, copies are not compatible. In that PR, I solved the issue by de-duping the `@types/react` package in the monorepo but we may eventually have valid reasons to have multiple copies of `@types/react` in the monorepo (should we add projects/tests with React 17, for instance).

The type change done here allows for the nextjs example project and `@lit-labs/react` to have different React types, and also only specifies APIs we use from React so users could create their own bag of React functions rather than passing in the whole module.

### DIFF
--- a/packages/labs/react/src/create-component.ts
+++ b/packages/labs/react/src/create-component.ts
@@ -102,10 +102,27 @@ type EventListeners<R extends EventNames> = {
     : (e: Event) => void;
 };
 
-// Interface that includes parts of the React module that we use. This allows
-// users to construct an object with just the needed methods and prevents type
-// incompatibilities when there are multiple React types being used.
-interface ReactLike {
+/**
+ * Interface that includes parts of the React module that `createComponent()`
+ * uses. Use this to create an object to pass into the `react` option.
+ *
+ * Example:
+ *
+ * ```ts
+ * const myReactLike = {
+ *   forwardRef,
+ *   useRef,
+ *   useLayoutEffect,
+ *   createElement,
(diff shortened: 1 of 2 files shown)


packages/labs/react/src/create-component.ts
```diff
@@ -90,8 +90,22 @@ type EventListeners<R extends EventNames> = {
     : (e: Event) => void;
 };
 
+// Interface that includes parts of the React module that we use. This allows
+// users to construct an object with just the needed methods and prevents type
+// incompatibilities when there are multiple React types being used.
+interface ReactLike {
```
justinfagnani, line 96: Should we export this? I could see someone using this with `satisfies` to make an object to pass to several createComponent calls. If so, make the comment a doc comment.

justinfagnani (approved):

augustjk:
Upon further exploration and thought, I'm less convinced that this is a valuable change to the package.

My main motivation for this was to allow multiple copies of React to type check for the case of our examples directory having a different version of React from the rest of the monorepo but that's actually a valid error as React breaks when multiple copies are used.

The only value, then, would be that users can provide a subset of the React module, but that also shoehorns us into only using those API and anything additional would be a breaking change.

Issue #4144, opened 2023-08-26
[reactive-element] Hybrid decorators


justinfagnani:
Make a single set of decorators that work in TypeScript with `experimentalDecorators` set to either true or false, and in Babel with `version` equal to "2023-05" by duck-typing the decorator arguments.

After this effort we will have one set of decorators that work under the following cases:

1. TypeScript with `experimentalDecorators: true` and no accessor keyword
2. TypeScript with `experimentalDecorators: true` and accessor keywords
3. TypeScript with `experimentalDecorators: false` and accessor keywords
4. Babel with `experimentalDecorators: false` and accessor keywords

The trade-off is that we have to have the same behavior in each configuration:

- We can't skip reflection for initial property values
- We have to wrap accessors as we can't detect the difference between auto-accessors and hand-written accessors with experimental decorators

Tasks:
- [x] Make legacy-decorators work with `experimentalDecorators: false` and pass the std-decorators tests (https://github.com/lit/lit/pull/4140)
- [x] Unify the behavior
  - [x] Reflect initial property values to attributes (https://github.com/lit/lit/pull/4145)
  - [x] Always wrap accessors (https://github.com/lit/lit/pull/4146)
- [x] Remove std-decorator and legacy-decorator files, but not std-decorator _tests_.
- [x] Run decorators against std-decorator tests with `experimentalDecorators: true`.
- [ ] Re-enable Babel tests when https://github.com/babel/babel/issues/15889 is fixed.
- [ ] Add a compiler-time flag to choose one or both decorator implementations to enable code elimination
- [x] Ensure we have tests for `noAccessor: true` in all cases
- [x] Deprecate and add a dev-mode warning for overrides of `createProperty` and `getPropertyDescriptor`

PR #4145, opened 2023-08-26
[reactive-element] Make standard property decorator reflect initial values

justinfagnani:
Part 2 of https://github.com/lit/lit/issues/4144

In order to get around the fact that the decorator result `init()` callback is called before the auto-accessor private field is installed, we add new parameters to `requestUpdate()` to forward the initial value.


packages/reactive-element/src/std-decorators/property.ts
```diff
@@ -112,14 +111,13 @@ export const property = (
...
         init(this: C, v: V): V {
-          if (v !== undefined) {
-            // We need this cast because these decorators are in a separate
-            // compilation unit from ReactiveElement, so can't see the
-            // @internal _$changedProperties field.
-            (
-              this as unknown as {_$changedProperties: PropertyValues}
-            )._$changedProperties.set(name, undefined);
-          }
+          (this.requestUpdate as InternalRequestUpdate)(
```
AndrewJakubowicz, line 114: Please add documentation somewhere as per Discord chat:

```
we need to call requestUpdate() for initial values, like we do in the legacy decorators, to both populate the changedProperties map and to perform the initial reflection. But standard decorators do that call too early and it would be an error to read the field to get the previous value at that point
```

AndrewJakubowicz (approved):
Nice!

rictic (approved):

packages/reactive-element/src/std-decorators/property.ts
justinfagnani: done!

packages/reactive-element/src/test/std-decorators/property_test.ts
```diff
@@ -80,13 +80,13 @@ suite('@property', () => {
     assert.equal(el.getAttribute('attr'), null);
     assert.equal(el.customAttr, 'customAttr');
     // Make sure that the default value doesn't reflect
```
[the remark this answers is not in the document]
justinfagnani: done

PR #4146, opened 2023-08-26
[reactive-element] Always wrap user accessors

justinfagnani:

This makes all reactive properties - static properties block, experimental decorators, and standard decorators - wrap user written accessors and call `this.requestUpdate()` from the generated setter.


.changeset/hungry-cups-destroy.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/reactive-element': major
+'lit-element': major
+'lit': major
+---
+
+Generated accessor for reactive properties now wrap user accessors and automatically call `this.requestUpdate()` in the setter.
```
bennypowers, line 7:
```suggestion
Generated accessor for reactive properties now wrap user accessors and automatically call `this.requestUpdate()` in the setter. As in previous versions, users can still specify `noAccessor: true`, in which case they should call `this.requestUpdate()` themselves in the setter if they want to trigger a reactive update.
```

packages/reactive-element/src/test/decorators/property_test.ts
```diff
@@ -128,15 +128,13 @@ suite('@property', () => {
       _foo?: number;
       updatedContent?: number;
 
-      @property({reflect: true, type: Number})
```
rictic, line 131: How does this code behave with this change? Is it a breaking change?

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -1160,9 +1160,9 @@ suite('ReactiveElement', () => {
       }
 
       set bar(value) {
-        const old = this.bar;
+        // const old = this.bar;
```
rictic, line 1163: Remove commented out code

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -1213,9 +1213,9 @@ suite('ReactiveElement', () => {
...
```
rictic, line 1216: Remove commented out code

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -1255,9 +1255,9 @@ suite('ReactiveElement', () => {
       }
 
       set foo(value) {
-        const old = this.foo;
+        // const old = this.foo;
```
rictic, line 1258: Remove commented out code

packages/reactive-element/src/test/reactive-element_test.ts
```diff
@@ -1272,9 +1272,9 @@ suite('ReactiveElement', () => {
       }
 
       set bar(value) {
-        const old = this.foo;
+        // const old = this.foo;
```
rictic, line 1275: Remove commented out code

scripts/check-size.js
```diff
@@ -9,7 +9,7 @@ import * as fs from 'fs';
 // it's likely that we'll ask you to investigate ways to reduce the size.
 //
 // In either case, update the size here and push a new commit to your PR.
-const expectedLitCoreSize = 15023;
+const expectedLitCoreSize = 15135;
```
rictic, line 12: Huh, bigger change than I'd have expected from this PR

.changeset/hungry-cups-destroy.md
AndrewJakubowicz: ^ Excellent suggestion.

packages/reactive-element/src/test/decorators/property_test.ts
rictic: Er, I should say _before_ this change

packages/reactive-element/src/reactive-element.ts
```diff
@@ -728,16 +729,24 @@ export abstract class ReactiveElement
     key: string | symbol,
     options: PropertyDeclaration
   ): PropertyDescriptor | undefined {
+    const {get, set} = getOwnPropertyDescriptor(this.prototype, name) ?? {};
     return {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       get(): any {
-        return (this as {[key: string]: unknown})[key as string];
+        return get === undefined
```
justinfagnani, line 736: actually... I want to change this so that the conditional is outside of the accessors and run only once

scripts/check-size.js
justinfagnani: Yeah... there's a bit of change here. I'm going to change the approach slightly, though I don't think it'll be fewer bytes

packages/reactive-element/src/test/reactive-element_test.ts
justinfagnani: done

packages/reactive-element/src/test/reactive-element_test.ts
justinfagnani: done

packages/reactive-element/src/test/decorators/property_test.ts
justinfagnani: It's not a breaking change with `experimentalDecorators: true` because those read the entire property descriptor off the class and it doesn't matter which is decorated.

Standard decorator usage requires that we decorate the setter though, because we're only passed the setter method.

I can test both in legacy tests.

scripts/check-size.js
justinfagnani: my new change made it worse: 15214 bytes, but I think aside from that it's a better implementation.

justinfagnani: Ok, kept it more static and got the size down to 15115

packages/reactive-element/src/test/decorators/property_test.ts
justinfagnani: Added a test for decorating a getter in experimental decorators. I don't think we can detect which accessor was decorated, so I couldn't add a deprecation warning.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -668,12 +674,7 @@ export abstract class ReactiveElement
     // is called before `finalize`, we ensure finalization has been kicked off.
     this.finalize();
     this.elementProperties.set(name, options);
-    // Do not generate an accessor if the prototype already has one, since
-    // it would be lost otherwise and that would never be the user's intention;
-    // Instead, we expect users to call `requestUpdate` themselves from
-    // user-defined accessors. Note that if the super has an accessor we will
-    // still overwrite it
-    if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
+    if (!options.noAccessor) {
```
AndrewJakubowicz, line 677: For my own understanding, this is where the wrapping is happening. `!this.prototype.hasOwnProperty(name)` would previously only wrap class fields and skip accessors or user defined get/set.
Now explicit getters/setters and `accessor` keyword cases pass through this code path (in addition to the class fields).

packages/reactive-element/src/reactive-element.ts
```diff
@@ -728,21 +729,22 @@ export abstract class ReactiveElement
     key: string | symbol,
```
AndrewJakubowicz, line 729: Does the comment on this method need to be updated? Specifically the part: `If no descriptor is returned, the property will not become an accessor.`

packages/reactive-element/src/reactive-element.ts
```diff
@@ -728,21 +729,22 @@ export abstract class ReactiveElement
...
     options: PropertyDeclaration
   ): PropertyDescriptor | undefined {
+    const {get, set} = getOwnPropertyDescriptor(this.prototype, name) ?? {
```
AndrewJakubowicz, line 732: Ah, I got surprised by https://google.github.io/styleguide/tsguide.html#static-this . No action required. Is it worth doing something like `ReactiveElement.prototype`. I guess all over the codebase we already use a static `this`.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -728,21 +729,22 @@ export abstract class ReactiveElement
...
+      set(this: ReactiveElement, v: unknown) {
+        (this as unknown as Record<string | symbol, unknown>)[key] = v;
+      },
+    };
     return {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      get(): any {
-        return (this as {[key: string]: unknown})[key as string];
+      get() {
+        return get!.call(this);
```
AndrewJakubowicz, line 742: Does `get` need the `!`? It looks like it should be definitely assigned. Same with `set` below.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -837,6 +839,22 @@ export abstract class ReactiveElement
...
     }
     this.elementStyles = this.finalizeStyles(this.styles);
+    if (DEV_MODE) {
+      if (this.hasOwnProperty('createProperty')) {
+        issueWarning(
+          'no-override-create-property',
+          'Overriding ReactiveElement.createProperty() is deprecated. ' +
+            'The override will not be called with standard decorators'
+        );
+      }
```
AndrewJakubowicz, line 849: I am missing something fundamental. How does this work?

E.g., I tried to create a minimal repro in a playground:

```ts

class Base {
  static createProperty() {}

  static isItOverridden() {
    if (this.hasOwnProperty('createProperty')) {
      // This incorrectly is true.
      console.log('Yes');
    }
  }
}

Base.isItOverridden(); // Logs yes???
```

packages/reactive-element/src/std-decorators/property.ts
```diff
@@ -139,10 +141,13 @@ export const property = (
         },
       };
     } else if (kind === 'setter') {
-      // We do not wrap the author's setter. They are expected to call
-      // requestUpdate themselves, and will possibly do so conditionally.
+      const {name} = context;
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
augustjk, line 145: This can be removed

packages/reactive-element/src/reactive-element.ts
justinfagnani: That comment is still true because of the ` if (descriptor !== undefined)` check in createProperty

AndrewJakubowicz (approved):
Nice! The tests are really great as well!

packages/reactive-element/src/reactive-element.ts
justinfagnani: It can't be `ReactiveElement.prototype` because we're calling into a potentially overridden method.

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: Can a test be added for this DEV_MODE warning?

packages/reactive-element/src/reactive-element.ts
justinfagnani: my understanding is that `f.call()` should be able the same speed as `f.bind()` and then a call. I think bind() used to actually be slower.

packages/reactive-element/src/reactive-element.ts
justinfagnani:
```ts
class Sub extends Base {}
Sub.isItOverridden(); // Doesn't log yes!

class Sub2 extends Base {
  static createProperty() {}
}
Sub2.isItOverridden(); // Logs yes!
```

packages/reactive-element/src/std-decorators/property.ts
justinfagnani: done!

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: Peter just gave me a great rundown on how all of this works. Yep I understand.

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: Peter ramped me up on what I missed as well. Thank you, it makes sense now!

augustjk (approved):

packages/reactive-element/src/test/std-decorators/property_test.ts
```diff
@@ -161,22 +161,36 @@ suite('@property', () => {
...
     }
     customElements.define(generateElementName(), E);
     const el = new E();
     container.appendChild(el);
     await el.updateComplete;
+
+    // Check initial values
     assert.equal(el._foo, undefined);
-    assert.equal(el.updatedContent, undefined);
+    assert.equal(el.updatedProperties!.get('foo'), undefined);
```
rictic, line 175: Here and in the main decorators tests: should we assert on `el.updatedProperties!.has('foo')` here?

rictic (approved):

packages/reactive-element/src/test/std-decorators/property_test.ts
justinfagnani: Since there's no initializer it'll be false, but I can assert on that instead

PR #4120, opened 2023-08-21
feat (labs/analyzer): add support for mixins

43081j:
Just a draft for now since it still needs tests, and the logic can probably be tidied up a bit (reduce some duplication).

PR #4147, opened 2023-08-26
[labs/analyzer] add mixin support

43081j:
Replaces #4120.

This is actually kevin's branch but caught up from main, etc.

Draft while i get my head around it


43081j:
i've also tried to update all occurrences of `throw new DiagnosticsError` or `throw new Error` to instead create/add a diagnostic and return `undefined`.

seems to be what we do elsewhere in the codebase, but do correct me if im wrong

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -428,7 +431,10 @@ export interface ClassDeclarationInit extends DeclarationInit {
 }
 
 export class ClassDeclaration extends Declaration {
-  readonly node: ts.ClassLikeDeclaration;
+  readonly node:
+    | ts.ClassDeclaration
+    | ts.ClassLikeDeclaration
+    | ts.CallExpression;
```
43081j, line 437: fwiw im not a fan of this since its a `ClassDeclaration` we're in, but im not sure how else we'd do it (to separate `ClassDeclaration` and mixins somehow)

43081j:
fyi i've marked for review but i have some tests to write still

there's also a comment or two i left myself which would be good to resolve first

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,198 @@
...
+        code: DiagnosticCode.UNSUPPORTED,
+        category: analyzer.typescript.DiagnosticCategory.Warning,
+      })
+    );
+    return undefined;
+  }
+
+  // TODO (43081j): deal with empty class declaration names properly, maybe
+  // throw an error?
+  const classDeclarationName = classDeclaration.name?.text ?? 'unknown';
```
43081j, line 165: what do we wanna do here?

maybe if somehow the class declaration has no name, add a diagnostic and return `undefined`?

43081j:
FYI i've also now updated gen-manifest to support mixins.

i still need to update the goldens at some point, though.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,201 @@
...
+ * Utilities for working with mixins
+ */
+
+import ts from 'typescript';
+import {AnalyzerInterface, MixinDeclaration} from '../model.js';
+import {getClassDeclaration} from './classes.js';
+import {createDiagnostic} from '../errors.js';
+import {DiagnosticCode} from '../diagnostic-code.js';
+import {parseNodeJSDocInfo} from './jsdoc.js';
+import {getFunctionLikeInfo} from './functions.js';
```
43081j, line 19: FYI this introduces a circular dependency as `functions.js` imports `mixins.js`

not too sure what you want to do about that

.changeset/cuddly-mugs-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lit-labs/analyzer': minor
```
bennypowers, line 2: should these be patches (== minor in pre-release range)?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -302,34 +306,81 @@ export const getHeritage = (
...
   expression: ts.Expression,
+  foundMixins: Reference[],
   analyzer: AnalyzerInterface
 ): Reference | undefined => {
   // TODO(kschaaf) Could add support for inline class expressions here as well
   if (analyzer.typescript.isIdentifier(expression)) {
     return getReferenceForIdentifier(expression, analyzer);
+  } else if (
+    analyzer.typescript.isCallExpression(expression) &&
+    analyzer.typescript.isIdentifier(expression.expression)
```
bennypowers, line 332: to what degree do we have to worry about weird cases like inline mixins?
```ts
class A {}
export class C extends (s => class extends s {})(A) {}
```

packages/labs/analyzer/src/lib/javascript/mixins.ts
bennypowers: can we dig up to the function name, and fall back to that before giving up?

```ts
export const JMixin = superClass => class extends superClass {}

// { name: 'JMixin' }
```

packages/labs/analyzer/src/test/javascript/mixins_test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  test('complex mixin declaration A', ({getModule}) => {
+    const decl = getModule('mixins').getDeclaration('A');
+    assert.ok(decl.isMixinDeclaration());
+    assert.instance(decl.classDeclaration, ClassDeclaration);
+    assert.instance(decl.classDeclaration, LitElementDeclaration);
+    const classDecl = decl.classDeclaration as LitElementDeclaration;
+    assert.equal(classDecl.name, 'A');
+    if (lang === 'ts') {
+      //TODO(kschaaf): Add detection of constructor-defined fields to ClassField
```
bennypowers, line 53: there was some preliminary work on this in #4135, so you may be able to remove the condition here

.changeset/cuddly-mugs-invite.md
43081j: probably right, i'll update

packages/labs/analyzer/src/lib/javascript/classes.ts
43081j: one for @kevinpschaaf i think - imo we shouldn't worry about it since there's all sorts of crazy syntax you could put there so it may be a very deep rabbit hole

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,201 @@
...
+      analyzer
+    );
+    return undefined;
+  }
+  if (!ts.isBlock(functionBody)) {
+    addDiagnosticIfMixin(
+      fn,
+      hasMixinHint,
+      `Expected mixin to have a block function body; arrow-function class ` +
+        `expression syntax is not supported.`,
```
kevinpschaaf, line 89: FYI, I believe this restriction is just due to the fact that you can't put decorators on class expressions in TS without it complaining (ala [this note](https://lit.dev/docs/composition/mixins/#applying-decorators-in-mixins)); maybe add a TODO here, since this might be overly restrictive for JS users...?

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -113,12 +113,11 @@ export const isLitElementSubclass = (
...
   const type = checker.getTypeAtLocation(node) as ts.InterfaceType;
   const baseTypes = checker.getBaseTypes(type);
-  for (const t of baseTypes) {
-    if (_isLitElementClassDeclaration(t, analyzer)) {
-      return true;
-    }
-  }
-  return false;
+  return baseTypes.some((t) =>
+    t.isIntersection()
```
kevinpschaaf, line 117: Let's leave a comment for why this is here:
```suggestion
    // Mixins will cause the base types to be an intersection that
    // includes `LitElement`
    t.isIntersection()
```

packages/labs/analyzer/src/lib/model.ts
kevinpschaaf: Yeah, the TS `node` field on the different declaration models is really of sort of dubious value; the declaration types in the model are sort of "semantic" declarations from the point of view of the module exports, not 1:1 mappings to AST types. 

Meaning, we're taking the stance that it's more useful to represent `export const MyClassWithSomeMixin = SomeMixin(MyClass)` as a ClassDeclaration than just a VariableDeclaration with a class type, since it's semantically almost identical to `export class MyClassWithSomeMixin extends SomeMixin(MyClass) {}`; consumers of the module should just care that it's a class, not how it was assembled.

The `node`s are only really going to be useful for e.g. maybe getting linkages back to source code locations, and I think it's ok to have this weirdness.

kevinpschaaf (changes requested):

43081j:
ok as per discord, for now i've done two things:

1. drop `isMixinClass` so i can just figure it out from scratch (see note below)
2. only consider call expressions to be applied mixins if the identifier is a mixin

Regarding the `isMixinClass` stuff... it actually had a bug in it before anyway so i've decided to remove it entirely and just solve the problem from scratch.

The bug, if anyone's curious, was that we only set `isMixinClass: true` for unknown class declarations. For custom elements and lit elements, we never passed the flag along. This meant they always had a `superClass`, while regular mixin class declarations didn't.

If anyone wants to help me figure it out, the aim of that flag is to make mixin classes have no `superClass` in their heritage (since it'd be the parameter to the function, nothing useful i guess). To do that, _somehow_ we need to detect we're computing the heritage of a mixin class in `getHeritageFromExpression` and null off the `superClass` if so. We achieved that with the flag before, but I feel like there must be a cleaner way.

If it turns out there's no cleaner way, i'll go back to the flag.

43081j:
ok for now at least i've reintroduced `isMixinClass` and fixed the bug with it

tests should pass now, PTAL if anyone gets chance

kevinpschaaf:
The changes LGTM, and thanks for tracking down the `isMixinClass` bug, the change to thread it through looks right to me.

It looks like we still need to fix the error of the Vue wrapper generator not walking the heritage chain to collect properties (and be robust to an element not having any properties).

So we don't lose the context, here's some notes I put in Discord a while back:

> It looks to me like the type error is because the generated Props interface is an empty object with no keys (so p as keyof Props is always never).

> It looks like Props is empty because it is just using declaration.reactiveProperties from here to generate that interface: https://github.com/lit/lit/blob/main/packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts#L80

> So I'm guessing we (a) need to make/use a helper to collect reactive properties from the entire proto chain (not just the youngest class declaration in the chain), and (b) improve the Vue generation code to deal with the fact that it's actually valid for a component to have no reactive properties (which currently leads to the type error here)
iow, the root cause is that [ElementMixins](https://github.com/lit/lit/pull/4147/files#diff-094f387742ae224684fee0e57bd2927b8da849cbb04f84d397cbaa242b831365R29) itself has no reactive properties, it inherits them from the mixin, and we're not collecting all reactive properties to generate the view Props interface

43081j:
#4293 might fix that. I opened it back when we discussed it, maybe worth merging that first. Will have to remind myself what I did there though

kevinpschaaf:
Awesome, sorry I missed that. LGTM'ed, and once that merges, hopefully we can get this merged.

augustjk:
With https://github.com/lit/lit/pull/4293 merged, perhaps this can be rebased and revisited? I know @e111077 has tried this on MWC and had it work for its doc generation.

e111077:
Can confirm that this implementation worked on the MWC repo with a small fix to the MWC repo's generator which made some silly assumptions

43081j:
👍  i'll rebase later today and see how it goes

43081j:
rebasing was proving difficult which is why i ended up catch up merging last time i think, so i just did the same again

the tests still fail thanks to some typescript import stuff, not too sure what to do about that. all ts imports are the same, in files i've not touched too. so not entirely sure which import exactly is causing problems

43081j:
looks like i got tests passing now 👍

packages/labs/analyzer/test-files/js/mixins/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@lit-internal/test-mixins",
+  "dependencies": {
+    "lit": "^2.0.0"
```
augustjk, line 4:
```suggestion
    "lit": "^3.0.0"
```

packages/labs/analyzer/test-files/ts/mixins/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@lit-internal/test-mixins",
+  "dependencies": {
+    "lit": "^2.0.0"
```
augustjk, line 4:
```suggestion
    "lit": "^3.0.0"
```

packages/labs/analyzer/test-files/js/mixins/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@lit-internal/test-mixins",
```
augustjk, line 2: probably doesn't matter that the name here conflicts if they're not part of our npm workspace but just in case.
```suggestion
  "name": "@lit-internal/test-mixins-js",
```

packages/labs/analyzer/test-files/ts/mixins/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@lit-internal/test-mixins",
```
augustjk, line 2:
```suggestion
  "name": "@lit-internal/test-mixins-ts",
```

augustjk (commented):
I'm totally not the best person to review the actual implementation here but nothing stands out to me as seriously erroneous. Tests all look good to me.

e111077:
BTW @kevinpschaaf has given this PR a verbal LGTM

e111077 (approved):

e111077 (approved):

43081j:
just did the minor changes @augustjk suggested (some package name updates) FYI

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -308,31 +312,62 @@ export const getHeritage = (
...
+  const superClass = getSuperClassAndMixins(expression, mixins, analyzer);
   return {
-    superClass,
+    superClass: isMixinClass ? undefined : superClass,
     mixins,
   };
 };
 
-export const getSuperClass = (
+export const getSuperClassAndMixins = (
```
AndrewJakubowicz, line 326: Nit: Could `getSuperClassAndMixins` return the `mixins` reference array, rather than expecting an empty one and mutating it?

Unless there is a case where a previously populated mixins array should be passed in?

AndrewJakubowicz: I take it back, I like the recursive implementation that this allows.

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -308,31 +312,62 @@ export const getHeritage = (
...
+    // TODO (43081j): consider supporting external mixins properly at some point
+    if (mixinRef === undefined || mixin === undefined) {
+      analyzer.addDiagnostic(
+        createDiagnostic({
+          typescript: analyzer.typescript,
+          node: expression,
+          message:
+            `This is presumed to be a mixin but it could not be found ` +
+            `in the current project. Mixins imported from outside the ` +
+            `project are not yet supported.`,
```
AndrewJakubowicz, line 354: How were mixins handled prior to this change? Were they skipped or did they have a diagnostic. Wondering if the diagnostic should remain, but the analysis should try and continue with a graceful fallback of continuing to traverse the mixin expressions.

E.g. `const myClass = localMixin(externalMixin(BaseClass))` could maybe skip the `externalMixin` but still return `BaseClass` as the super class.

Or would we prefer to error than provide potentially wrong analysis results?

No action required - mostly a question so I can better understand the error handling philosophy of this package.

justinfagnani: If I'm following this correctly, we're going from TS expression -> model reference and bailing if we don't find the model reference. This makes some sense for things where we need analysis info outside of the type system (attributes, events, etc). But we should be able to support analysis of a mixins' basic class contributions even when they're not found in the analyzed model. We have to process the mixin type to find properties, etc. That seems valuable for elements that use mixins that are plain JS and not part of the CEM ecosystem.

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -355,3 +390,33 @@ export const isConstructorFieldInitializer = (
...
+    analyzer.typescript.isCallExpression(expression) &&
+    analyzer.typescript.isIdentifier(expression.expression)
+  ) {
+    const mixinRef = getReferenceForIdentifier(expression.expression, analyzer);
+
+    try {
+      mixinRef?.dereference(MixinDeclaration);
+    } catch (_err) {
+      return undefined;
+    }
```
AndrewJakubowicz, line 409: Can `.dereference` throw? Does the dereference at https://github.com/lit/lit/pull/4147/files#diff-8cba81979748429e02fab6dda2f43ed55e1fa2b634c3faf818d4cf1d0ee3707eR344 also need to handle exceptions?

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
AndrewJakubowicz, line 3: Nit: I think 2024 is when this file will be added to the repo, so this should be updated to 2024.
We'll get this PR in!!!!

packages/labs/analyzer/src/lib/javascript/classes.ts
43081j: you may be right though, that it could technically throw. i guess we should try/catch it just the same

packages/labs/analyzer/src/lib/javascript/classes.ts
43081j: keep in mind this particular error has a `TODO` above it to implement support for external mixins (i.e. those we can't get a reference to, but know exist)

so do we want to do that in this PR or a separate one?

if i followed correctly, i think thats what you're asking for (the same as the todo)

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022 Google LLC
+ * Copyright 2024 Google LLC
```
AndrewJakubowicz, line 3: License only needs to be updated on new files. This should remain 2022.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
+ * If the function is unannotated and does not match the above mixin shape, it
+ * will silently just be analyzed as a simple function and not a mixin. However,
+ * the `@mixin` annotation can be added to produce specific diagnostic errors
+ * when a condition for being analyzed as a mixin is not met.
+ */
+export const maybeGetMixinFromFunctionLike = (
+  fn: ts.FunctionLikeDeclaration,
+  name: string,
+  analyzer: AnalyzerInterface,
+  hasMixinHint = nodeHasMixinHint(fn, analyzer)
```
AndrewJakubowicz, line 60: What is the case where `hasMixinHint` is passed in? Can this be defined as a const in the function block?

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
+      fn,
+      hasMixinHint,
+      `Expected mixin to have a block function body; arrow-function class ` +
+        `expression syntax is not supported.`,
+      analyzer
+    );
+    return undefined;
+  }
+  let classDeclaration!: ts.ClassDeclaration;
+  let returnStatement!: ts.ReturnStatement;
```
AndrewJakubowicz, line 98: Below these are both checked if they have been initialized. Could the type here contain the possible `undefined` value? For less characters these could be initially assigned to `null`?

```
let classDeclaration: ts.ClassDeclaration | null = null;
...
```

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
+      classDeclaration,
+      classDeclarationName,
+      analyzer,
+      undefined,
+      true /* isMixinClass */
+    ),
+  };
+};
+
+const findSuperClassArgIndexFromHeritage = (
```
AndrewJakubowicz, line 185: Documentation on this function would be helpful since it's quite impressive what this function is doing, and it took me a while to understand it.

If I understand correctly, this function is finding which argument is the superClass.

packages/labs/analyzer/src/lib/javascript/classes.ts
43081j: my mistake, i did intend on only updating the new files. not sure how this one slipped through, i'll update

packages/labs/analyzer/src/lib/javascript/mixins.ts
AndrewJakubowicz: I may be completely misunderstanding this function as well, but it may be worth adding an additional test case where a mixin applies some other mixin.

AndrewJakubowicz (approved):
Some minor comments, but overall this is awesome! pre-approving!

Only blocking comments from me are:
 1. The license date that was updated in an old file.
 2. Clarifying that `findSuperClassArgIndexFromHeritage` works correctly.

Thank you so much!

packages/labs/analyzer/src/lib/javascript/mixins.ts
43081j: you're right, not sure why this was a param. maybe left over from an old iteration and just forgot to remove it

packages/labs/analyzer/src/lib/javascript/mixins.ts
43081j: turns out it was because closures don't change the type.

for example:

```ts
let foo: number | null = null;

[].forEach(() => {
  foo = 5;
});

// at this point, `foo` is `null` still

if (!foo) {
  return;
}

// at this point, `foo` is `never`!
```

i think i've fixed it by using a `for...of` rather than `forEach`

packages/labs/analyzer/src/lib/javascript/mixins.ts
43081j: you understood correctly i think

it'll basically recursively resolve your example:

- initially it'll have the expression `someOtherMixin(someArg, superClass)` and some possible names of `['superClass', 'someArg']`
- it'll then check if `someArg` is one of the possible names
- it is so it'll return index `1` (because its `[1]` of the possible names)

so you're right, it'll pick up `someArg` as the class arg rather than `superClass`

not sure what we can do about that though 🤔

packages/labs/analyzer/src/lib/javascript/classes.ts
justinfagnani: yeah, I think for our own reference it might be nice to note these two types of "external". I wouldn't mind seeing an open issue and have the error message link to it.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
```
justinfagnani, line 185: Is this finding the index in the mixin function's parameter list or the heritage call expression's argument list?

Note that in general, the superclass arg could be a named property too. I'm not sure what TypeScript's rules are.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
+  possibleSuperClasses: string[],
+  expression: ts.Expression,
+  analyzer: AnalyzerInterface
+): number => {
+  if (analyzer.typescript.isIdentifier(expression)) {
+    return possibleSuperClasses.indexOf(expression.text);
```
justinfagnani, line 191: Can't we match by declaration rather than a string comparison?

packages/labs/analyzer/test-files/ts/mixins/src/mixins-vanilla.ts
```diff
@@ -0,0 +1,27 @@
...
+type Constructor<T> = new (...args: any[]) => T;
+
+/**
+ * Some mixin
+ * @mixin
+ * @template {Constructor<HTMLElement>} T
+ * @param {T} superClass
+ * @return T
+ */
+export const MixinA = <T extends Constructor<HTMLElement>>(superClass: T) => {
```
justinfagnani, line 17: I think we need a test case for arrow functions with concise bodies, ala:

```ts
export const MixinA = <T extends Constructor<HTMLElement>>(superClass: T) =>
  class MixedElement extends superClass {
```

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -559,6 +563,22 @@ export class ClassDeclaration extends Declaration {
   }
 }
 
+export interface MixinDeclarationInit extends FunctionLikeInit {
+  node: ts.FunctionLikeDeclaration;
```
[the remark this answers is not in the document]
43081j: this one is left over from long ago before we introduced `node` to the base interface iirc

i'll remove it

packages/labs/analyzer/src/lib/javascript/mixins.ts
43081j: it is the index in the function's parameter list, based on the heritage of the class

i think if we want to change this, we should probably do it in a follow-up since it could be a bit of a refactor to allow an object to be used as a parameter

packages/labs/analyzer/src/lib/javascript/mixins.ts
kevinpschaaf: Good catch here, this does seem wrong. I think in the `isCallExpression` branch, rather than recursing into `findSuperClassArgIndexFromHeritage()`, this function likely needs to try to look up / dereference the mixin that's being called and return the index of its own argument passed into the mixin's superclass argument; that will recurse until a mixin extends one of its arguments directly (iow, taking the `isIdentifier` branch) rather than calling another mixin.

So in the example, we'd look up `someOtherMixin`, it would have a `superClassArgIdx` of `1`, and so we'd find the index of the `superClass` argument, and search the current mixin's arg list for that name and return `0`.

As Justin said in the other comment, the name searches could be likely be done with symbol declaration identity, but that could be a separate change.

packages/labs/analyzer/src/lib/javascript/mixins.ts
justinfagnani: I'm not sure we _should_ fix this, especially if TypeScript is strict about recognizing mixins. We can keep an eye on it.

packages/labs/analyzer/src/lib/javascript/mixins.ts
```diff
@@ -0,0 +1,205 @@
...
+  let returnStatement: ts.ReturnStatement | null = null;
+  for (const s of functionBody.statements) {
+    if (analyzer.typescript.isClassDeclaration(s)) {
+      classDeclaration = s;
+    }
+    if (analyzer.typescript.isReturnStatement(s)) {
+      returnStatement = s;
+    }
+  }
+  if (!classDeclaration) {
```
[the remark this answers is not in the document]
43081j: now it works 👍

packages/labs/analyzer/src/lib/javascript/classes.ts
43081j: i've added more info and an issue at least

packages/labs/analyzer/src/lib/javascript/mixins.ts
justinfagnani: I think we might need to add something to the custom elements manifest here to identify which parameter of a mixin is the superclass.

We can do something internally to the analyzer in the mean time by just adding that to the model object.

justinfagnani:
I think we should merge this now and open an issue to address known problems in the mixin superclass lookup.

justinfagnani (approved):

PR #4148, opened 2023-08-27
fix(analyzer): ecma private members

### LINKED ISSUE #4127: [labs/analyzer]: ecma #private methods marked as public

### Description

Given 

```ts
class HasHash {
  #priv = 0;
  #hash() { }
}
```

I get 
```json
{
  "schemaVersion": "1.0.0",
  "modules": [
    {
      "kind": "javascript-module",
      "path": "has-hash.js",
      "declarations": [
        {
          "kind": "class",
          "name": "HasHash",
          "members": [
            {
              "kind": "method",
              "name": "#hash",
              "privacy": "public",
              "return": {
                "type": {
                  "text": "void"
                }
              }
            }
          ]
        }
      ]
    }
  ]
}
```

### Reproduction

have a private method in a class

### Workaround

I will modify the generated manifest post-facto

### Is this a regression?

### Affected versions

    "@lit-labs/analyzer": "^0.9.0",     "@lit-labs/gen-manifest": "^0.2.4",

Node 18.12.1

bennypowers:
Closes #4127

### DIFF
--- a/.changeset/perfect-bags-cheer.md
+++ b/.changeset/perfect-bags-cheer.md
@@ -1,0 +1,5 @@
+---
+"@lit-labs/analyzer": patch
+---
+
+Correctly sets `privacy` field for ECMAScript private methods
--- a/.changeset/perfect-bags-cheer.md
+++ b/.changeset/perfect-bags-cheer.md
@@ -1,5 +1,5 @@
 ---
-"@lit-labs/analyzer": patch
+'@lit-labs/analyzer': patch
 ---
 
 Correctly sets `privacy` field for ECMAScript private methods
--- a/packages/labs/analyzer/src/lib/javascript/classes.ts
+++ b/packages/labs/analyzer/src/lib/javascript/classes.ts
@@ -153,7 +153,10 @@ export const getClassMembers = (
         })
       );
     } else if (typescript.isPropertyDeclaration(node)) {
-      if (!typescript.isIdentifier(node.name)) {
+      if (
+        !typescript.isIdentifier(node.name) &&
+        !typescript.isPrivateIdentifier(node.name)
+      ) {
         analyzer.addDiagnostic(
           createDiagnostic({
             typescript,
@@ -169,6 +172,7 @@ export const getClassMembers = (
       }
 
       const info = getMemberInfo(typescript, node);
+
       (info.static ? staticFieldMap : fieldMap).set(
         node.name.getText(),
         new ClassField({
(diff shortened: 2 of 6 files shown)


.changeset/perfect-bags-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lit-labs/analyzer": patch
```
bennypowers, line 2:
```suggestion
'@lit-labs/analyzer': patch
```

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -153,7 +153,10 @@ export const getClassMembers = (
         })
       );
     } else if (typescript.isPropertyDeclaration(node)) {
-      if (!typescript.isIdentifier(node.name)) {
+      if (
+        !typescript.isIdentifier(node.name) &&
+        !typescript.isPrivateIdentifier(node.name)
+      ) {
```
justinfagnani, line 159: No action needed: thinking out loud. Regarding this diagnostic below, I wonder if it really deserves being a warning. It's not a problem with the code, it's a limitation of the analyzer. I hope we have a way of ignoring errors at the line and project level.

packages/labs/analyzer/src/lib/utils.ts
```diff
@@ -47,6 +47,8 @@ export const hasProtectedModifier = (ts: TypeScript, node: ts.HasModifiers) => {
 const isPrivate = (ts: TypeScript, node: ts.Node) => {
   return (
     (ts.canHaveModifiers(node) && hasPrivateModifier(ts, node)) ||
+    ((ts.isPropertyDeclaration(node) || ts.isMethodDeclaration(node)) &&
+      ts.isPrivateIdentifier(node.name)) ||
```
justinfagnani, line 51: Does `ts.isPrivateIdentifier(()` need to be guarded by the property and method declaration checks?

Just thinking that this condition has a lot of clauses now, and it'd be easier to read if it were just:

```ts
return (
  hasPrivateModifier(ts, node) ||
  ts.isPrivateIdentifier(node.name) ||
  hasJSDocTag(ts, node, 'private')
);
```

If the guards are needed, maybe some variables would help reading:
```ts
const isTsPrivate = ts.canHaveModifiers(node) && hasPrivateModifier(ts, node));
const isMember = ts.isPropertyDeclaration(node) || ts.isMethodDeclaration(node);
const isJsPrivate = isMember && ts.isPrivateIdentifier(node.name);
const isJsDocPrivate = isMember && hasJSDocTag(ts, node, 'private');

return isTsPrivate || isJsPrivate || isJsDocPrivate;
```

If you don't want to refactor this, no action needed though. We have tests :)

justinfagnani (approved):
🎉

packages/labs/analyzer/src/lib/utils.ts
bennypowers: yeah i wasn't super jazzed about that, but at a certain point my desire to fight the compiler just drops off sharply, and if I can fill the gaps in the typescript ast's api with more checks - :shrug: good enough for now

PR #4009, opened 2023-07-11
Add standard decorators to @lit-labs/context

rictic:
Fixes https://github.com/lit/lit/issues/3994

PR #4151, opened 2023-08-28
[reactive-element] Make context decorators work with standard decorators

justinfagnani:

One thing to check on here is how optional fields are translated to auto-accessors and whether or not we have the types correct, and had them correct prior to this PR. The exist tests include a case of an optional field decorated with a non-optional context type. That doesn't exactly work with accessors, which can't be optional, and so have to include `undefined` in their type to get the same effect. This difference is that this setup requires the context include `undefined`, whereas optional fields _don't_. The actual type of optional fields is the same though, so it seems like something was wrong in the original types.


rictic (commented):
This is a small subset of the tests in https://github.com/lit/lit/pull/4009/files

I can land that PR after this one and land all of the tests as part of that

justinfagnani:
I didn't copy over all of the tests because I'm not entirely happy with how the tests are organized right now. Ideally I would have just copied over the decorator tests, but there aren't tests for just the decorators. There are tests for scenarios, and some components of the tests directly use controllers or use decorators.

It's a little tricky because of the two-part nature of the protocol. It's easier to test a consumer if you have a provider and vice-versa. I don't think it's that great to reimplement half of the context protocol with the possibility of mistakes there. It would be great if there were common protocol interop tests. But even with a combined comsumer/provider test, that could use the controllers only and then we could have isolated and smaller tests for decorators.

packages/labs/context/src/lib/decorators/consume.ts
```diff
@@ -47,28 +46,60 @@ export function consume<ValueType>({
...
+      );
+    }
+  }) as ConsumeDecorator<ValueType>;
 }
 
-type ConsumerDecorator<ValueType> = {
-  <K extends PropertyKey, Proto extends ReactiveElement>(
+/**
+ * Generates a public interface type that removes private and protected fields.
+ * This allows accepting otherwise compatible versions of the type (e.g. from
```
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/context/src/lib/decorators/provide.ts
```diff
@@ -48,15 +47,41 @@ export function provide<ValueType>({
...
-      ctor.addInitializer((element: ReactiveElement): void => {
-        controllerMap.set(element, new ContextProvider(element, {context}));
+  return (<C extends ReactiveElement, V extends ValueType>(
+    protoOrTarget: ClassAccessorDecoratorTarget<C, V>,
+    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<C, V>
+  ) => {
+    // Map of instances to controllers
+    const controllerMap = new WeakMap();
+    if (typeof nameOrContext === 'object') {
+      nameOrContext.addInitializer(function (this: C) {
```
rictic, line 57: Here and for consume: let's add a comment saying essentially "this is the standard decorators branch" and "this is the legacy experimental decorators branch"

packages/reactive-element/src/test/decorators/property_test.ts
```diff
@@ -160,10 +160,10 @@ suite('@property', () => {
 
     // Check initial values
     assert.equal(el._foo, undefined);
-    assert.isFalse(el.updatedProperties!.has('foo'));
+    assert.equal(el.updatedProperties!.get('foo'), undefined);
```
rictic, line 163: Looks like a merge error.

rictic (approved):

packages/labs/context/src/lib/decorators/provide.ts
justinfagnani: done

PR #4156, opened 2023-08-30
[reactive-element] Consolidate decorator implementations

justinfagnani:

This gets us most of the way through the implementation side of https://github.com/lit/lit/issues/4144, the rest being additional tests.

There are some new breaking changes:
- `decorators/base.js` is no long a package export. I don't think we should commit to that as a public API. I removed the main helper also.
- `@queryAll()` returns a NodeList always to match the newer standard decorator implementation. Not a big change.
- Removed the inline `slot.assignedElements` polyfill

The PR is large, but I made isolated, single-purpose commits to make it easier to review.


justinfagnani:
Looks like I forgot that we had another decorator in localized-decorator.

AndrewJakubowicz:
Do the changes mentioned in the description need a changeset?

justinfagnani:
@AndrewJakubowicz yeah, I'll add one

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -4,4 +4,59 @@
...
+ * @param tagName The tag name of the custom element to define.
+ */
+export const customElement =
+  (tagName: string): CustomElementDecorator =>
+  (
+    classOrTarget: CustomElementClass | Constructor<HTMLElement>,
+    context?: ClassDecoratorContext<Constructor<HTMLElement>>
+  ) => {
+    if (context !== undefined) {
+      context.addInitializer(() => {
```
rictic, line 53: Why not always immediately call customElements.define? What happens as part of class finalization that it might care about?

I guess another decorator could mess with the prototype chain, add custom element callbacks…

Ok, yeah, I see

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -4,4 +4,110 @@
...
+  return (<C extends Interface<ReactiveElement>, V extends Array<Element>>(
+    protoOrTarget: ClassAccessorDecoratorTarget<C, V>,
+    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<C, V>
+  ) => {
+    const {slot, selector} = options ?? {};
+    const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
+    const doQuery = (el: Interface<ReactiveElement>) => {
+      const slotEl =
+        el.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
+      const elements = slotEl?.assignedElements(options) ?? [];
```
rictic, line 89: Oh nice, yeah we don't need the polyfill anymore: https://caniuse.com/mdn-api_htmlslotelement_assignedelements

packages/reactive-element/src/decorators/query-assigned-nodes.ts
```diff
@@ -4,4 +4,98 @@
...
+    protoOrTarget: ClassAccessorDecoratorTarget<C, V>,
+    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<C, V>
+  ) => {
+    const {slot} = options ?? {};
+    const slotSelector = `slot${slot ? `[name=${slot}]` : ':not([name])'}`;
+    const doQuery = (el: Interface<ReactiveElement>) => {
+      const slotEl =
+        el.renderRoot?.querySelector<HTMLSlotElement>(slotSelector);
+      return (slotEl?.assignedNodes(options) ?? []) as unknown as V;
+    };
```
rictic, line 83: Most of this is shared with queryAssignedElements. Since base is private now, we could use a shared function from there, save a few bytes

rictic (approved):

packages/reactive-element/src/decorators/custom-element.ts
justinfagnani: I tried to add a test for this and failed for now, but addInitializer() should go after static fields are assigned, whereas accessing the class immediately would be before that. I tried to make some test that triggered that, but a static getters (I used it for `observedAttributes`) are installed early. Anyway, I know this is the right timing, we can make a test later?

packages/reactive-element/src/decorators/query-assigned-elements.ts
justinfagnani: yep :)

packages/reactive-element/src/decorators/query-assigned-nodes.ts
justinfagnani: yeah... this closure captures some variables. I'm inclined to do that later.

PR #4159, opened 2023-08-30
[localize] Port localize decorator to hybrid

justinfagnani:

This makes `@localize()` work with standard decorators, adds standard decorator tests, and adds a test case to `@lit/localize-tools` to ensure that the transformer works.

This is needed to make https://github.com/lit/lit/pull/4156 pass.


packages/localize/src/internal/localized-decorator.ts
```diff
@@ -36,28 +43,18 @@ import type {
...
-    finisher(clazz: typeof ReactiveElement) {
-      clazz.addInitializer(updateWhenLocaleChanges);
-    },
+export const localized: Localized =
+  (): LocalizedDecorator =>
+  (
+    clazz: typeof ReactiveElement,
+    _context?: ClassDecoratorContext<typeof ReactiveElement>
+  ) => {
+    (clazz as typeof ReactiveElement).addInitializer(updateWhenLocaleChanges);
```
rictic, line 52: is this cast needed?

packages/localize/src/internal/localized-decorator.ts
```diff
@@ -36,28 +43,18 @@ import type {
...
-  return {
-    kind,
-    elements,
-    finisher(clazz: typeof ReactiveElement) {
-      clazz.addInitializer(updateWhenLocaleChanges);
-    },
+export const localized: Localized =
+  (): LocalizedDecorator =>
+  (
+    clazz: typeof ReactiveElement,
```
rictic, line 49: Should we use the `Interface` trick to strip off non-public fields?

packages/localize/src/internal/localized-decorator.ts
rictic, line 52: Nice that it gets simpler and smaller with this change!

rictic (approved):
Great!

Taking on trust that the testdata files are actually copies

Very weird that github doesn't see them as such, since they should have the exact same content

packages/localize/src/internal/localized-decorator.ts
justinfagnani: nope!

packages/localize/src/internal/localized-decorator.ts
justinfagnani: done

packages/localize/src/internal/localized-decorator.ts
justinfagnani: Yes!

PR #4161, opened 2023-08-30
[labs/task] Disambiguate `TaskFunction` type

### LINKED ISSUE #4155: [labs/task] Type error for task function using abort signal

Task (@lit-labs/task)

### Description

When passing in a task function to the `Task` controller, it fails to infer the type of `signal` when used like:

```ts
#task = new Task(
  this,
  {
    task: async ([], {signal}) => {
    //                ^? Property 'signal' does not exist on type 'TaskFunctionOptions | undefined'.
    },
    args: () => []
  }
)
```

### Reproduction

https://stackblitz.com/edit/vitejs-vite-p3wbf8?file=src%2Fmy-element.ts

### Workaround

We probably need to overload the TaskFunction type rather than making the second param optional.

### Is this a regression?

### Affected versions

3.0.1

N/A

augustjk:
Fixes #4155

The optional notation on 2nd param and the signal option made TypeScript not accept a task function with an explicit second param. Removing the `?` fixes this. Since we always provide the second argument with the abort signal when calling the provided function, it makes sense to type it so.

TypeScript is still happy to accept a function that doesn't have a 2nd param.

### DIFF
--- a/.changeset/gentle-needles-perform.md
+++ b/.changeset/gentle-needles-perform.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/task': patch
+---
+
+Disambiguate `TypeFunction` type to fix type error when providing a task function making use of the abort signal.
--- a/packages/labs/task/src/task.ts
+++ b/packages/labs/task/src/task.ts
@@ -7,12 +7,12 @@ import {notEqual} from '@lit/reactive-element';
 import {ReactiveControllerHost} from '@lit/reactive-element/reactive-controller.js';
 
 export interface TaskFunctionOptions {
-  signal?: AbortSignal;
+  signal: AbortSignal;
 }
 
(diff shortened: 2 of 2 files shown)


rictic (commented):
Please add a type test that fails without this change and passes with it.

We've been writing these as:

```ts
if (false as boolean) {
  // This is good code because [...]
  codeThatShouldCompileCorrectly();

  // We want this to be a type error because [...]
  // @ts-expect-error
  codeThatShouldFailToCompile()
}
```

justinfagnani (approved):

augustjk:
> Please add a type test that fails without this change and passes with it.

@rictic Thanks! I added a test following an existing pattern shown here https://github.com/lit/lit/blob/170e91648472d21ecee3fca9ac7a0a52787b6e98/packages/labs/task/src/test/task_test.ts#L626-L636

rictic (approved):

PR #4164, opened 2023-08-30
Update wireit, use the new quiet logger in CI testing

rictic:
Addressing feedback in https://github.com/google/wireit/pull/850 but I think this is an improvement as is

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,6 +68,7 @@ jobs:
 
       - name: Test
         env:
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz (approved):
Nice!

justinfagnani:
Yeah... looking at the CI logs, this seems better already. There are a lot of nearly blank lines w/ spinners, and progress bars from WTR that would be nice to omit, but the failure in the Sauce runner was easier to get to.

justinfagnani (approved):

PR #4167, opened 2023-08-31
fix(analyzer): attributes

bennypowers:
## What I did
- analyze a ce's the `attributes` field, populated by the `@attr` JSDoc tag
- apply in gen-manifest for `attribute` and `reflects` on custom element class fields
- added a `CustomElementField` class in the analyzer's OOP representation
- format gen-manifest output (2 spaces JSON stringify formatting, and selective key sorting)
- add a `type: json` to snapshot testing helpers which uses uvu deep record equivalence instead of string comparison for jaon snapshots. Used for CEM goldens
-----

The reason for the `CustomElementField` class is to represent the schema, which defines `ClassField` and `CustomElementField extends ClassField`: In the spec, a field can have an attribute and a reflect property but only if it represents a field on a custom element. Whereas a field on a regular class that's not a custom element class doesn't have a corresponding attribute and certainly doesn't reflect. In general we can say that the specification is ambiguous as to how attributes can be represented. You can represent in the `attributes` array only those attributes which don't have a corresponding class field, or you can duplicate them in the `attributes` array while also documenting them in the `members` array. In the past I've followed the first option and only documented attributes in the `attributes` array which don't also exist in the `members` array. 

We could alternately expose a single `ClassField` object for all class properties and then make `attribute` and `reflect` optional on that interface, but that wouldn't accurately represent the specification which has the `CustomElementField` definition extending from the `ClassField` definition, and it also would misrepresent the DOM, since regular non-ce class fields don't have attributes.

This PR introduces some formatting to the gen-manifest output. Unfortunately that makes review more difficult, but I felt it necessary because it also makes testing much easier via visual diff programs


bennypowers:
tests are green, but this isn't quite ready to merge yet: i wasn't able to produce attrs in my real-world project

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -598,6 +616,16 @@ export class LitElementDeclaration extends CustomElementDeclaration {
   constructor(init: LitElementDeclarationInit) {
     super(init);
     this.reactiveProperties = init.reactiveProperties;
+    for (const [name, prop] of this.reactiveProperties) {
+      const field = this.getField(name);
+      if (field instanceof CustomElementField) {
+        field.attribute =
```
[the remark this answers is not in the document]
bennypowers: yeah i know what you mean
tests were green without it, but running it in my project failed to provide attribute and reflect :see_no_evil:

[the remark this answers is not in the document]
bennypowers: we do have to cross-reference with the `reactiveProperties` here, though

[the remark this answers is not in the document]
bennypowers: ok tinkering complete :) PTAL.

justinfagnani:
> In general we can say that the specification is ambiguous as to how attributes can be represented. You can represent in the attributes array only those attributes which don't have a corresponding class field, or you can duplicate them in the attributes array while also documenting them in the members array.

We should add some more comments here to describe the intention: every attribute should be listed in the attributes array, and every field in the members array. Fields have the `attribute` property only to describe the _relationship_ between a specific field and attribute, not to fully describe the attribute. Consider it a reference.

> We could alternately expose a single ClassField object for all class properties and then make attribute and reflect optional on that interface, but that wouldn't accurately represent the specification which has the CustomElementField definition extending from the ClassField definition

This is essentially what `CustomElementField` already is. It adds to optional fields on top of `ClassField`, but has no discriminator itself, so any `CustomElementClass` can just add the `reflects` and `attribute` properties to any field.

packages/labs/gen-manifest/goldens/test-element-a/custom-elements.json
```diff
@@ -59,6 +60,16 @@
           ],
           "tagName": "element-a",
           "customElement": true,
+          "attributes": [
+            {
```
justinfagnani, line 64: This should also have `foo`.

justinfagnani (changes requested):

justinfagnani:
I opened a PR to clarify attributes in the schema: https://github.com/webcomponents/custom-elements-manifest/pull/126

packages/labs/gen-manifest/goldens/test-element-a/custom-elements.json
bennypowers: I updated the golden with attributes and with new formatting. the formatting introduces unrelated whitespace and key-ordering changes, but those changes to help a lot with testing

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -413,6 +418,16 @@ export class ClassField extends Declaration {
   }
 }
 
+export class CustomElementField extends ClassField {
+  attribute?: string | boolean | undefined;
```
43081j, line 422: afaict this is always a string in CEM. presumably you added a boolean to say "this is an attribute, by its field name"?

its probably fine but would be good to understand. i do wonder if we should purposely just duplicate the name if that's the case too, i.e. `{attribute: 'foo', name: 'foo'}`

bennypowers: Yes, I found I had to pass the boolean through so that gen-manifest and friends could correctly calculate the actual attribute down the libe

43081j: my point is maybe we should just always pass the attribute when there is one

right now:

- `name` and `attribute: false` (or `undefined`) means the field has no attribute
- `name` and `attribute: true` means the field has an attribute by the same name
- `name` and `attribute: str` means the field has an attribute named `str`

maybe we should just set it right to avoid later computation etc, as in `{attribute?: string}`.

```ts
{
  name: "my-field",
  attribute: "my-field"
}
```

etc...

bennypowers: ok that sounds good. I made that change. the analyzer's `CustomElementField` instance will always get the computable string attribute when it's constructed

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -175,7 +192,7 @@ export const getClassMembers = (
 
       (info.static ? staticFieldMap : fieldMap).set(
         node.name.getText(),
-        new ClassField({
+        new CustomElementField({
```
43081j, line 195: is this right? all properties of all classes are now "custom element" fields?

or am i missing a condition somewhere that means this only happens for lit properties?

43081j: this chain of logic is what has been itching my mind.

i think what we should be doing (somewhere) is the same as what we do for classes:

```ts
  if (isLitElementSubclass(declaration, analyzer)) {
    return getLitElementDeclaration(declaration, analyzer);
  }
  if (isCustomElementSubclass(declaration, analyzer)) {
    return getCustomElementDeclaration(declaration, analyzer);
  }
```

similar, somewhere around here, we'd do:

```ts
if (isCustomElementSubclass(node, analyzer)) {
  return getCustomElementField(node, analyzer);
}
```

basically, right now you're computing all a class' members. you then _mutate_ those members later on in lit-specific code. which is the bit that irks me.

i feel like you should be computing the right thing the first time around, somehow.

everywhere else, we seem to have achieved that by type-guards and delegations to the associated `get*` code.

bennypowers: good catch, there should be a predicate here

bennypowers: actually this should always be a `ClassField`, because custom elements and lit elements are handled elsewhere

bennypowers: > basically, right now you're computing all a class' members. you then mutate those members later on in lit-specific code. which is the bit that irks me.

i guess the trade-off would be to have lit-specific code in the general `classes` file? we don't have code for attr/prop pairs in `extends HTMLElement` yet

bennypowers: i opted to new up, like your snippet - seemed the most expedient. At the same time I broke up `getClassMembers` - to ease future refactoring

packages/labs/analyzer/src/lib/custom-elements/attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import ts from 'typescript';
+import {parseNamedTypedJSDocInfo} from '../javascript/jsdoc.js';
+import {Attribute} from '../model.js';
+import {AnalyzerInterface} from '../model.js';
+import {getTypeForTypeString} from '../types.js';
+
+/**
+ * Returns an array of analyzer `Event` models for the given
```
justinfagnani, line 14: Update comment?

packages/labs/analyzer/src/lib/custom-elements/attributes.ts
```diff
@@ -0,0 +1,34 @@
+/**
```
justinfagnani, line 1: Add copyright header

packages/labs/analyzer/src/lib/custom-elements/attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+  tag: ts.JSDocTag,
+  attributes: Map<string, Attribute>,
+  analyzer: AnalyzerInterface
+) => {
+  const info = parseNamedTypedJSDocInfo(tag, analyzer);
+  if (info === undefined) {
+    return;
+  }
+  const {name, type, description, summary} = info;
+  attributes.set(name, {
```
justinfagnani, line 27: what about merging, since we potentially have several ways to describe the same attribute? Do we have a good precedence defined if something overrides the other, or should we be careful to merge?

I think it could be a valid pattern to say use jsdoc to add attribute-only docs to an attribute otherwise defined by a reactive property, for instance.

We could drop a warning and todo if the map already contains the attribute for now.

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -43,6 +43,22 @@ import {
...
+}
+
+interface ClassMemberInfo {
+  fieldMap: Map<string, ClassField>;
+  staticFieldMap: Map<string, ClassField>;
+  methodMap: Map<string, ClassMethod>;
+  staticMethodMap: Map<string, ClassMethod>;
+}
+
+interface ClassMemberInfoIntermediate extends ClassMemberInfo {
```
justinfagnani, line 58: jsdoc for what this is for?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -94,64 +110,158 @@ const getIsReadonlyForNode = (
...
+
+export const deriveFieldsFromConstructorAssignments = (
+  node: ts.ConstructorDeclaration,
+  {fieldMap}: ClassMemberInfo,
+  analyzer: AnalyzerInterface
+) => {
+  // Ignore non-implementation signatures of overloaded methods by checking
+  // for `node.body`.
+  if (node.body) {
+    // TODO(bennypowers): We probably want to see if this matches what TypeScript considers a field initialization.
```
justinfagnani, line 139: nit: rewrap comments to 80 col?

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -94,64 +110,158 @@ const getIsReadonlyForNode = (
...
+    // assignment expression so that we get ones inside of if statements, in parenthesized expressions, etc.
+    //
+    // Also, this doesn't cover destructuring assignment.
+    //
+    // This is ok for now because these are rare ways to "declare" a field,
+    // especially in web components where you shouldn't have constructor parameters.
+    node.body.statements.forEach((node) => {
+      if (isConstructorAssignmentStatement(node, analyzer.typescript)) {
+        const name = node.expression.left.name.getText();
+        fieldMap.set(
```
justinfagnani, line 150: what do we do if there's both a field _declaration_ and a constructor assignment? I think we should ignore the constructor assignment then, yeah?

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -33,18 +39,75 @@ export const getLitElementDeclaration = (
...
+  const {
+    privacy,
+    inheritedFrom,
+    source,
+    readonly,
+    type,
+    description,
+    deprecated,
+    summary,
+  } = field;
```
justinfagnani, line 69: why two destructuring statements here? combine?

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -33,18 +39,75 @@ export const getLitElementDeclaration = (
...
+  const attribute =
```
justinfagnani, line 70: I wonder how well the current code handles property metadata merging? ie, you can have `@property()` for the same field on a base class and sub class and the sub class will get merged property options.

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
```diff
@@ -33,18 +39,75 @@ export const getLitElementDeclaration = (
...
+    deprecated,
+    default: field.default,
+    readonly,
+    inheritedFrom,
+    source,
+    type,
+  });
+}
+
+const getLitElementClassMembers = (
```
justinfagnani, line 94: Are the reactive properties not already fields when using decorators? How does this handle that case? Add that in a comment?

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -585,11 +602,42 @@ export class CustomElementDeclaration extends ClassDeclaration {
...
   }
+
+  #addFieldToAttributeMap(
+    field: ClassField,
+    attributes: Map<string, Attribute>
+  ) {
+    if (field instanceof CustomElementField && field.attribute) {
+      const {attribute, name: fieldName, ...rest} = field;
+      const name =
+        typeof attribute === 'boolean' ? fieldName.toLowerCase() : attribute;
```
justinfagnani, line 619: `typeof attribute === 'boolean'` should be the same as `attribute === true`, right? I find the later a tad easier to read.

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -585,11 +602,42 @@ export class CustomElementDeclaration extends ClassDeclaration {
...
     this.slots = init.slots;
     this.cssProperties = init.cssProperties;
     this.cssParts = init.cssParts;
   }
+
+  #addFieldToAttributeMap(
+    field: ClassField,
+    attributes: Map<string, Attribute>
+  ) {
+    if (field instanceof CustomElementField && field.attribute) {
```
justinfagnani, line 616: I know that `attribute: ''` is invalid, but do we want to represent it here for linters to be able to warn on? `&& field.attribute` will prevent that from going through.

justinfagnani (changes requested):

packages/labs/analyzer/src/lib/javascript/classes.ts
bennypowers: Map set handles that, right? So comment that order of operations is important?

packages/labs/analyzer/src/lib/model.ts
43081j: i think you're right and we should let it through. not sure how else we'd get at the original value without some hackery

packages/labs/analyzer/src/lib/model.ts
bennypowers: i think this is a leftover from an earlier iteration. at this point, the `attribute` on the field model should always be a `string | undefined`, never a bool

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
bennypowers: It will narrow the type to CustomElementField, as we'd expect

What about when there's a static reactive prop without a corresponding class field. even if we'd recommend linting that out, nonetheless, the property will act like a field at runtime, right? so we should add it into the fields list

bennypowers: added this logic and a test for it

packages/labs/analyzer/src/lib/custom-elements/attributes.ts
bennypowers: do we really need to warn? if i set up a converter, the attribute might be genuinely different from the field

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
bennypowers: removed the destructuring in favour of property assignment

packages/labs/analyzer/src/lib/javascript/classes.ts
bennypowers: ok I added a guard which should handle cases like this:
```ts
class A {
  /** don't clobber me */
  a = 'a';
  constructor() {
    this.a = 'a';
  }
}
```

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -95,63 +115,168 @@ const getIsReadonlyForNode = (
...
         name,
-        new ClassMethod({
-          ...info,
-          ...getFunctionLikeInfo(node, name, analyzer),
-          ...parseNodeJSDocInfo(node, analyzer),
+        new ClassField({
+          name,
+          type: getTypeForNode(accessor, analyzer),
+          privacy: getPrivacy(analyzer.typescript, accessor),
+          readonly: !!get && !set,
```
e111077, line 244: jsdoc has been missing from accessors: `...parseNodeJSDocInfo(accessor, analyzer)`. Do you mind adding that it so we can obviate my PR at #4198

e111077 (changes requested):

packages/labs/analyzer/src/lib/javascript/classes.ts
bennypowers: [db95dfd](https://github.com/lit/lit/pull/4167/commits/db95dfdd84c04c5fada8b5b51e474bb4b760438f)

e111077 (approved):
LGTM on my specific comments

packages/labs/analyzer/src/lib/lit-element/lit-element.ts
bennypowers: I've lost track over the holidays, last eng meeting i came to we agreed not to do comprehensive inheritance inside the analyzer class, but rather provide tools (which presumably would be responsible for disk io and package resolution)

How does that affect this thread?

packages/labs/analyzer/src/lib/custom-elements/attributes.ts
bennypowers: a contrived example of my previous comment, 

```js
@property({ converter: {...} }) constructors = [];
```
```html
<x-l constructors="a,b ,   c"></x-l>
<script>
console.log(document.currentScript.previousElementSibling.constructors);
// [object A], [object B], [object C]
</script>

**Review by 43081j — APPROVED** [review]

**Review by justinfagnani — APPROVED** [review]

File: packages/labs/analyzer/src/lib/custom-elements/attributes.ts [hunk 17]
```diff
@@ -58,6 +62,17 @@ export const addJSDocAttributeToMap = (
     customElementFieldMap,
     name
   );
+  if (attributes.has(name))
```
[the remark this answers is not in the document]
justinfagnani: oh yeah, linting or formatting should catch this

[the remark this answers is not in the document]
bennypowers: 🤷‍♂️ prettier's better-than-yours opinions were not invoked

bennypowers:
i'm not going to be focusing on this pr for the next while. please take it over

PR #4168, opened 2023-08-31
feat (localize-tools): upgrade parse5 to 7.x

43081j:
This just upgrades the parse5 dependency to 7.x which ships its own typescript types amongst various other improvements.

The SSR package has also been updated to import types from the parse5 root as the deep path isn't exported in parse5's `package.json`.

cc @augustjk

### DIFF
--- a/.changeset/thin-apricots-help.md
+++ b/.changeset/thin-apricots-help.md
@@ -1,0 +1,6 @@
+---
+'@lit/localize-tools': minor
+'@lit-labs/ssr': patch
+---
+
+Upgrade parse5 to 7.x in localize-tools and import from root of parse5 where possible
--- a/.changeset/thin-apricots-help.md
+++ b/.changeset/thin-apricots-help.md
@@ -1,5 +1,5 @@
 ---
-'@lit/localize-tools': minor
+'@lit/localize-tools': patch
 '@lit-labs/ssr': patch
 ---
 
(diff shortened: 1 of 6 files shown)


packages/labs/ssr/package.json
```diff
@@ -195,7 +195,6 @@
     "@types/koa-cors": "*",
     "@types/koa-mount": "^4.0.2",
     "@types/koa-static": "^4.0.1",
-    "@types/parse5": "^6.0.1",
```
AndrewJakubowicz, line 198: Ah – can this be removed because the types come from `parse5` directly?
Very nice!

.changeset/thin-apricots-help.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lit/localize-tools': minor
```
AndrewJakubowicz, line 2: No action required – just curious: Why `minor` and not `patch`?

AndrewJakubowicz (approved):

AndrewJakubowicz:
Thank you!

.changeset/thin-apricots-help.md
augustjk: Yeah I think patch makes sense for this too. `@lit/localize-tools` is at 0.6.9. I wouldn't consider this a breaking change.

```suggestion
'@lit/localize-tools': patch
```

augustjk (approved):
Thank you this is great!

.changeset/thin-apricots-help.md
43081j: wasn't too sure tbh. i figured both should be a patch but wasn't sure what we do when we bump a major version of a dependency.

ill update

PR #4170, opened 2023-09-01
Graduate @lit-labs/task out to @lit/task

rictic:


packages/labs/task/src/task.ts
```diff
@@ -3,441 +3,16 @@
...
-  : undefined;
-
-export const shallowArrayEquals = <T extends ReadonlyArray<unknown>>(
-  oldArgs: T,
-  newArgs: T
-) =>
-  oldArgs === newArgs ||
-  (oldArgs.length === newArgs.length &&
-    oldArgs.every((v, i) => !notEqual(v, newArgs[i])));
+export {
```
augustjk, line 7: should we re-export with `@deprecated`? or we leave as is and just mark the package as deprecated in the npm registry?

rictic: Yeah, good idea!

packages/task/CHANGELOG.md
```diff
@@ -0,0 +1,196 @@
+# Change Log
+
+## 3.0.1
```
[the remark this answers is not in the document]
rictic: Great point!

rollup-common.js
```diff
@@ -23,17 +23,18 @@ const PACKAGE_CLASS_PREFIXES = {
...
-  '@lit-labs/react': '_$G',
-  '@lit-labs/scoped-registry-mixin': '_$H',
-  '@lit-labs/ssr-client': '_$I',
-  '@lit-labs/task': '_$J',
-  '@lit-labs/router': '_$K',
-  '@lit-labs/observers': '_$L',
-  '@lit-labs/context': '_$M',
-  '@lit-labs/vue-utils': '_$N',
-  '@lit-labs/preact-signals': '_$O',
+  '@lit/task': '_$E',
```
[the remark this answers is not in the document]
rictic: Hm, I didn't think of that. Accessing private fields from other versions is fraught, but it should reduce surprising behavior at least if we keep them consistent. Ideally I guess we'd have unique prefixes for each release, but that would bloat binary sizes a bit, and probably not worth the additional engineering cost..

augustjk (approved):
🎓 🎊

packages/task/package.json
```diff
@@ -0,0 +1,165 @@
...
+      "output": []
+    },
+    "test": {
+      "dependencies": [
+        "test:dev",
+        "test:prod"
+      ]
+    },
+    "test:dev": {
+      "command": "MODE=dev node ../tests/run-web-tests.js \"development/**/*_test.js\" --config ../tests/web-test-runner.config.js",
```
AndrewJakubowicz, line 125: Nit: `MODE=dev` and `MODE=prod` (in the next test) could be moved into wireit `env`, but non blocking.

AndrewJakubowicz (approved):
Awesome! LGTM!

lit-next.code-workspace
```diff
@@ -20,6 +20,10 @@
       "name": "ssr",
       "path": "packages/labs/ssr"
     },
+    {
```
justinfagnani, line 23: VS Code has gotten a lot better about working with nested packages that aren't workspace roots, so I don't know if we need to add every package here anymore. I'm also fine with it.

packages/labs/task/README.md
```diff
@@ -2,85 +2,8 @@
 
 A controller for Lit that renders asynchronous tasks.
 
-## Overview
+This package has graduated from labs! It is now available as `@lit/task`. This package will no longer be updated, but it will continue to work and get updates through the ^1.0.0 version range of `@lit/task` because it's now just a re-export of that package. This should reduce duplication of code while everyone migrates their imports and dependencies.
```
justinfagnani, line 5: I think the "it's now just a re-export of that package" part should com earlier, then mention this package is a proxy and won't be updated.

We might want a headline here too.

packages/labs/task/package.json
```diff
@@ -57,12 +56,14 @@
         "build:ts:types",
         "build:rollup",
         "../../lit:build",
-        "../../reactive-element:build"
+        "../../reactive-element:build",
```
justinfagnani, line 59: do we need lit and reactive-element anymore?

packages/labs/task/package.json
```diff
@@ -156,8 +146,9 @@
     "@lit-internal/scripts": "^1.0.1-pre.0",
     "lit": "^3.0.0-pre.0"
   },
+  "#versionComment": "We depend on <2.0.0 of task because we want to be compatible with 0.x.x versions, as well as the 1.x.x versions, as they're intended to all be compatible with the 3.x.x API of lit-labs/task before it graduated out of labs.",
   "dependencies": {
-    "@lit/reactive-element": "^2.0.0-pre.0"
+    "@lit/task": "<2.0.0"
```
justinfagnani, line 151: The only issue with doing it this way is that it's slightly harder to pin to a specific version. In that case though the user can use `@lit/task` directly, or use npm's new-ish resolution overrides.

packages/labs/task/src/deep-equals.ts
```diff
@@ -4,121 +4,19 @@
...
- *    - have the same constructor
- *    - have same set of own property names
- *    - have each own property be deeply equal
- *  - Arrays, Maps, Sets, and RegExps
- *  - Objects with custom valueOf() (ex: Date)
- *  - Objects with custom toString() (ex: URL)
- *
- * Important: Objects must be free of cycles, otherwise this function will
- * run infinitely!
+ * You can import directly from @lit/task/deep-equals.js now.
```
justinfagnani, line 8: Does this deprecate imports of this file by being on the export? Or does that need to be in a `@fileoverview` or `@module` comment?

packages/task/package.json
```diff
@@ -0,0 +1,165 @@
+{
+  "name": "@lit/task",
+  "version": "0.1.0",
```
justinfagnani, line 3: This could be `0.0.0` to make it clear it hasn't been released yet.

packages/task/rollup.config.js
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {litProdConfig} from '../../rollup-common.js';
+import {createRequire} from 'module';
+
+export default litProdConfig({
+  packageName: createRequire(import.meta.url)('./package.json').name,
+  entryPoints: ['deep-equals', 'index', 'task'],
+  external: ['@lit/reactive-element'],
```
justinfagnani, line 13: In the future I think we should change all of these to `external: (id) => !(id.startsWith('.') || id.startsWith('/')),`

justinfagnani (approved):
🎉

packages/labs/task/package.json
rictic: For tests, yes

packages/labs/task/package.json
rictic: Package locks are the solution to this I think

packages/labs/task/src/deep-equals.ts
rictic: I tried those techniques as well as a couple others, and none worked to deprecate the whole module

packages/task/package.json
rictic: +1

packages/task/rollup.config.js
rictic: +1

lit-next.code-workspace
rictic: +1, but not in this PR

packages/labs/task/README.md
rictic: +1, done

packages/tests/package.json
```diff
@@ -33,10 +34,39 @@
...
+      "files": [],
+      "output": [],
+      "env": {
+        "BROWSERS": {
+          "external": true
+        }
+      }
+    },
+    "test:common:prod": {
+      "command": "MODE=prod node ./run-web-tests.js",
```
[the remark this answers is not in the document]
rictic: Good point, done

packages/task/package.json
rictic: Done

Issue #4138, opened 2023-08-24
[labs/react] Typescript errors with react wrapped component in Vite Preact Typescript template


ajmchambers:
### Which package(s) are affected?

React (@lit-labs/react)

### Description

Trying out shoelace in a fresh vite preact-ts template and getting a couple of typescript issues.

(shoelace 2.8.0 which is using @lit-labs/react 2.0.1, NOTE: react wrapper files are generated on build [here](https://github.com/shoelace-style/shoelace/blob/next/scripts/make-react.js#L58-L68))

With the following code I am getting 2 different typescript errors:

```tsx
import { useState } from 'preact/hooks';
import SlButton from '@shoelace-style/shoelace/dist/react/button/index.js';
import SlIcon from '@shoelace-style/shoelace/dist/react/icon/index.js';

export default function Test() {
    const [count, setCount] = useState(0);
    return (
        <SlButton onClick={() => setCount((count) => count + 1)} variant="primary">
            <SlIcon slot="prefix" library="system" name="star-fill" />
            count is {count}
        </SlButton>
    );
}
```

SlButton type error (not an error in stackblitz example but it is locally in vscode):
```
'SlButton' cannot be used as a JSX component.
  Its type 'ReactWebComponent<SlButton, { onSlBlur: EventName<SlBlurEvent>; onSlFocus: EventName<SlFocusEvent>; onSlInvalid: EventName<SlInvalidEvent>; }>' is not a valid JSX element type.
    Type 'ForwardRefExoticComponent<Omit<HTMLAttributes<SlButton>, "onSlBlur" | "onSlFocus" | "onSlInvalid"> & Partial<Omit<SlButton, keyof HTMLElement>> & EventListeners<...> & RefAttributes<...>>' is not assignable to type 'FunctionComponent<any>'.
      Type 'ReactNode' is not assignable to type 'VNode<any> | null'.
        Type 'undefined' is not assignable to type 'VNode<any> | null'.ts(2786)
```

SlIcon has this error:
```
Type 'Element' is not assignable to type 'ReactNode'.
  Property 'children' is missing in type 'VNode<any>' but required in type 'ReactPortal'.ts(2322)
```

If I change SlIcon to a div:
```tsx
<div slot="prefix"></div>
```

The typescript error is just:
```
Type 'Element' is not assignable to type 'ReactNode'.ts(2322)
```

### Reproduction

Here is a stackblitz example (note: the first error does not appear here for me, only in local vscode):
[...]

PR #4172, opened 2023-09-01
[labs/react] Remove custom children type and include our own copy of `PropsWithoutRef` for usage with `preact/compat`

### LINKED ISSUE #4169: [labs/react] Created component type is incompatible in Preact with typescript >=5.1

React (@lit-labs/react)

### Description

May be related to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types

Using wrapped component in Preact yields:
```
'ElementA' cannot be used as a JSX component.
  Its type 'ReactWebComponent<ElementA, { onAChanged: EventName<CustomEvent<unknown>>; }>' is not a valid JSX element type.
    Type 'ForwardRefExoticComponent<Omit<HTMLAttributes<ElementA>, "children" | "onAChanged"> & EventListeners<{ onAChanged: EventName<CustomEvent<unknown>>; }> & { ...; } & Partial<...> & RefAttributes<...>>' is not assignable to type 'FunctionComponent<any>'.
      Type 'ReactNode' is not assignable to type 'VNode<any>'.
        Type 'string' is not assignable to type 'VNode<any>'.ts(2786)
```

We may need to ditch `ForwardRefExoticComponent` and just type the component in a more generic way.

### Reproduction

Bump the version of typescript in the examples/preact in the monorepo.

### Workaround

downgrade TS to ~5.0

### Is this a regression?

### Affected versions

2.0.2

N/A

augustjk:
Fixes #4169, #4138 

Instead of trying to figure out React-Preact compatible types ourselves like the franken-child type, it would be better for us to defer to `preact/compat` typings that provide their versions of React types.

Due to https://github.com/preactjs/preact/issues/4124, we need to use our own copy of `PropsWithoutRef` type for proper type checking to occur with `preact/compat`.

This will also ensure we don't mix React and Preact types for users of TS >= 5.1 where the types become further incompatible.

Tested with our `examples/preact` project configured now to use `preact/compat`. I have also tested this after manually updating the typescript dependency version for the example project.


packages/labs/react/src/create-component.ts
```diff
@@ -31,6 +31,18 @@ export type WebComponentProps<I extends HTMLElement> = React.DetailedHTMLProps<
...
+// included in preact/compat yet.
+// See: https://github.com/preactjs/preact/issues/4124
+/** Ensures that the props do not include ref at all */
+type PropsWithoutRef<P> =
+  // Omit would not be sufficient for this. We'd like to avoid unnecessary mapping and need a distributive conditional to support unions.
+  // see: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types
+  // https://github.com/Microsoft/TypeScript/issues/28339
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  P extends any ? ('ref' extends keyof P ? Omit<P, 'ref'> : P) : P;
+
```
AndrewJakubowicz, line 45: spoke offline – would be nice to isolate this in a separate file and include the license.

AndrewJakubowicz (commented):
Fantastic investigating and solution!

packages/labs/react/src/create-component.ts
augustjk: Thank you, I have split out `PropsWithoutRef` into its own file and included all license and attributions there.

packages/labs/react/package.json
```diff
@@ -80,7 +81,7 @@
       ]
     },
     "build:ts:types": {
-      "command": "treemirror development . \"**/*.d.ts{,.map}\" \"!development/test\"",
+      "command": "treemirror development . \"**/*.d.ts{,.map}\" \"!test\"",
```
AndrewJakubowicz, line 84: Is this change expected?

AndrewJakubowicz (approved):
Nice!

packages/labs/react/package.json
augustjk: Yes intended! It was incorrect when I added this a while back. Since we already set `development` as the source dir in the first argument, the glob starts relative to that. It was no harm that it wasn't working, just unnecessarily copying extra test files.

packages/labs/react/rollup.config.js
```diff
@@ -9,6 +9,11 @@ import {createRequire} from 'module';
 
 export default litProdConfig({
   packageName: createRequire(import.meta.url)('./package.json').name,
-  entryPoints: ['index', 'create-component', 'use-controller'],
+  entryPoints: [
+    'index',
+    'create-component',
+    'use-controller',
+    'props-without-ref',
```
justinfagnani, line 16: Should this be listed as an entrypoint?

justinfagnani (approved):
LGTM, with a question on if the new file needs to be in the rollup config entrypoints.

We're not publishing in exports, so it doesn't matter _that_ much, but we are putting the file in the top-level, so someone not using package exports could import this file.

packages/labs/react/rollup.config.js
rictic: +1. We should keep exports and entrypoints in sync.

AndrewJakubowicz: I reverted the entrypoints so it doesn't contain the new file "props-without-ref"

augustjk:
> LGTM, with a question on if the new file needs to be in the rollup config entrypoints.
> 
> We're not publishing in exports, so it doesn't matter _that_ much, but we are putting the file in the top-level, so someone not using package exports could import this file.

It's included as rollup config to generate the js and js.map files but it is type only. I wasn't sure if the type import would resolve with just the d.ts file present. It's an empty file so i think it's no harm to have it emit and included in the package without any package export mapping.

justinfagnani:
> It's included as rollup config to generate the js and js.map files but it is type only.

Ah. In that case I'd like to use `import type` to note that it will elide the import completely, and then skip the .js files. But...

Can we just inline the type into `create-component.ts`? We can still indicate that we copied it from DefinitelyTyped with a link to the file. It's only ~60 chars or so.

AndrewJakubowicz:
> Can we just inline the type into create-component.ts?

This is my fault since I asked to keep the license with the copied source code. By copying the code & license into a separate file it's easier to delete in the future when the type is included in preact's compatibility types.

+1 to `import type`

justinfagnani:
Since it's a one-liner, can we just inline something?

Here's a version that's not a copy that distributes `Omit`:

```ts
type DistributiveOmit<T, K extends string | number | symbol> = T extends any ? Omit<T, K> : T;

// usage
export type OmitRef<T> = DistributiveOmit<T, 'ref'>;
```

The only difference is that in editors if you hover over the type you'll always see the `OmitRef<..., 'ref'>` even if the original type didn't have ref in it. I don't think this is bad, considering the only place we use the type is in a complex type anyway.

AndrewJakubowicz:
Applied code review feedback. @justinfagnani, I needed to tweak the type slightly more to get the generic distributive omit to type check.

#### Test plan

I manually tested this change against examples/preact with TypeScript 5.2 and the error no longer occurs.
Requesting re-review from @justinfagnani  or @rictic .

justinfagnani (approved):
thanks!

PR #4175, opened 2023-09-02
chore (labs/compiler, labs/ssr, localize-tools): update parse5/tools

43081j:
Bumping `@parse5/tools` up so we can benefit from this easier import:

```ts
// new...

import {Element} from '@parse5/tools';

// previously...

import {DefaultTreeAdapterMap} from 'parse5';
type Element = DefaultTreeAdapterMap['element'];
```

FYI i chose to alias the names just for sake of readability, i.e. a knowing that we're referencing a parse5 `Element` rather than a DOM `Element`

### DIFF
--- a/packages/labs/ssr/src/lib/util/parse5-utils.ts
+++ b/packages/labs/ssr/src/lib/util/parse5-utils.ts
@@ -4,11 +4,16 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {DefaultTreeAdapterMap} from 'parse5';
-import {traverse, replaceWith, isElementNode} from '@parse5/tools';
+import {
+  traverse,
+  replaceWith,
+  isElementNode,
+  Element as p5Element,
+  Node as p5Node,
+} from '@parse5/tools';
 
-export function removeFakeRootElements(node: DefaultTreeAdapterMap['node']) {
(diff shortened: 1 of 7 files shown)


packages/labs/ssr/src/lib/util/parse5-utils.ts
```diff
@@ -4,11 +4,16 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {DefaultTreeAdapterMap} from 'parse5';
-import {traverse, replaceWith, isElementNode} from '@parse5/tools';
+import {
+  traverse,
+  replaceWith,
+  isElementNode,
+  Element as p5Element,
```
justinfagnani, line 11: All type names should start with a capital letter. And I'm personally a fan of longer, more readable, names. I would go for `Parse5Element`, although in these libraries the DOM types aren't loaded and just `Element` and `Node`, etc would be fine with me too.

justinfagnani (changes requested):

packages/labs/ssr/src/lib/util/parse5-utils.ts
43081j: sure no worries, ill just revert to the normal names

AndrewJakubowicz (approved):
Awesome!

justinfagnani (approved):
Thanks!

PR #4177, opened 2023-09-03
chore (labs/gen-manifest): update goldens

43081j:
looks like these weren't updated at some point, so i'm just doing it here to avoid other ongoing branches needing to


bennypowers (approved):

AndrewJakubowicz (approved):
Perfect! Thank you!

PR #4180, opened 2023-09-06
fix(analyzer): heritage

bennypowers:
DO NOT MERGE until #4167 lands and this is rebased

## What I did
- update goldens for gen-manifest to include inherited members, attrs, and events
- added a subclass of element-a in test projects


justinfagnani:
I'm taking a look at #4167 again. Can you say what's fixed here? I assume there isn't an issue for it?

bennypowers:
This is about subclasses listing inherited properties

PR #4183, opened 2023-09-06
[reactive-element] Run modern decorator tests on experimental decorators

justinfagnani:

Changes:
- Moves the `src/test/std-decorators/*` to `src/test/decorators-modern/*` to try to indicate that the tests are "modern" style, but not tied to "standard decorators". Naming is hard.
- Changes the main tsconfig to not exclude the std-decorator tests (which don't exist in src anymore)
- Changes the std-decorators-tests.tsconfig to compile the modern tests and output them to test/std-decorators
- Fixes the errors that these tests flagged, notably:
  - We were calling `defineProperty()` int he decorators, which worked for fields, but for accessors the TS helper was overwriting the property with the original property descriptor. Turns out the solution is to just return a descriptor (we didn't do this before because Babel), and for the properties we use, the descriptor has the same shape as the standard decorator result, so we can fully share the implementations resulting in less decorator code!
  - `@property`'s setter doesn't receive initial values for accessors. The solution is to try to detect that case when we save instance properties and feed initial values back through `requestUpdate()`. This seems to work.


justinfagnani:
This TypeScript issue is causing issues with making experimental decorators behave the same with `accessor` as without, and so behave the same as standard decorators: https://github.com/microsoft/TypeScript/issues/55669

Because experimental decorators can't see the initializer for accessors, we can't reflect it or add it to changed properties. We could read off all initial values, but that would include `undefined` for non-initialized fields which would over-reflect and unnecessarily populated changeProperties.

packages/reactive-element/src/test/decorators-modern/property_test.ts
```diff
@@ -74,7 +74,7 @@ suite('@property', () => {
     await el.updateComplete;
     assert.equal(el.updateCount, 1);
     assert.equal(el.noAttr, 'noAttr');
-    assert.isTrue(changedProperties!.has('noAttr'));
+    assert.isTrue(changedProperties!.has('noAttr'), 'A');
```
AndrewJakubowicz, line 77: What is the `, 'A'`?

packages/reactive-element/src/test/decorators-modern/property_test.ts
```diff
@@ -404,7 +404,7 @@ suite('@property', () => {
     await el.updateComplete;
 
     // Properties should have the pre-upgraded values
-    assert.equal(el.foo, 'hi');
+    assert.equal(el.foo, 'hi', 'B');
```
AndrewJakubowicz, line 407: What is the `, 'B'`?

packages/reactive-element/src/test/decorators-modern/property_test.ts
justinfagnani: debug label, I'll remove

AndrewJakubowicz:
I don't follow the "(we didn't do this before because Babel)" part of the description. Is it because babel legacy decorator support has been removed?

justinfagnani:
> I don't follow the "(we didn't do this before because Babel)" part of the description. Is it because babel legacy decorator support has been removed?

Yes, and for hybrid decorators between Babel and TS that we called defineProperty. But going forward we will only support Babel on standard decorators.

packages/reactive-element/src/decorators/query.ts
```diff
@@ -59,50 +60,63 @@ export type QueryDecorator = {
...
     };
-    if (typeof nameOrContext === 'object') {
+    if (cache) {
+      // Accessors to wrap from either:
+      //   1. The decorator target, in the case of standard decorators
+      //   2. The property descriptor, in the case of experimental decorators
+      //      on auto-accessors.
+      //   3. Functions that access our own cache-key property on the instance,
+      //      in the case of experimental decorators on fields.
+      const {get, set} =
```
AndrewJakubowicz, line 79: Worth renaming `get` and `set` so it's clear that this is getting/setting the cache?

packages/reactive-element/src/decorators/query.ts
```diff
@@ -59,50 +60,63 @@ export type QueryDecorator = {
...
+    descriptor?: PropertyDescriptor
   ) => {
     const doQuery = (el: Interface<ReactiveElement>): V => {
+      // TODO: if we want to allow users to assert that the query will never
+      // return null, we need a new option and to throw here if the result
+      // is null.
       return (el.renderRoot?.querySelector(selector) ?? null) as V;
     };
-    if (typeof nameOrContext === 'object') {
+    if (cache) {
```
rictic, line 72: consider structuring this like:

```ts
if (!cache) {
  // simple case, just return the doQuery getter
}
// harder case, with caching
```

I think it will read a bit better

packages/reactive-element/src/decorators/query.ts
```diff
@@ -59,50 +60,63 @@ export type QueryDecorator = {
...
-      } else {
-        defineProperty(protoOrTarget, nameOrContext as PropertyKey, {
-          get(this: ReactiveElement) {
-            return doQuery(this);
-          },
-        });
-      }
-      return;
+      // This object works as the return type for both standard and
+      // experimental decorators.
```
AndrewJakubowicz, line 114: Unsure if this comment is needed, as the other decorators in this PR don't have a similar comment.

packages/reactive-element/src/test/decorators-modern/README.md
```diff
@@ -0,0 +1,6 @@
+This is a set of tests that use "modern" decorator syntax (`accessor`, etc) that are run with both experimental and standard decorators to ensure identical behavior between the experimental and standard decorator code paths.
+
+The tests compiled with two tsonfigs:
```
AndrewJakubowicz, line 3: Typo:  "The tests __are__ compiled"

Spelling: `tsconfigs`.

packages/reactive-element/src/test/decorators-modern/README.md
```diff
@@ -0,0 +1,6 @@
+This is a set of tests that use "modern" decorator syntax (`accessor`, etc) that are run with both experimental and standard decorators to ensure identical behavior between the experimental and standard decorator code paths.
```
AndrewJakubowicz, line 1: Technically the behavior is not identical because there is a slight difference in initial `undefined` values? But maybe not worth documenting here?

AndrewJakubowicz:
I think ssr tests are failing because `queueMicrotask` is not defined.

AndrewJakubowicz (commented):
Awesome!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1038,19 +1038,32 @@ export abstract class ReactiveElement
...
+            this[p as keyof this] !== undefined &&
+            !this._$changedProperties.has(p)
+          ) {
+            // Trigger initial value reflection and populate the initial
+            // changedProperties map, but only for the case of experimental
+            // decorators on accessors, which will not have already populated
+            // the changedProperties map. We skip `undefined` values assuming
+            // those aren't initial values - a small difference from
+            // experimental decorators on fields.
+            this.requestUpdate(p, undefined, options);
```
rictic, line 1065: Does this trigger a second render, or is this still before the initial render? Do we have a test of that?

rictic (approved):

packages/reactive-element/src/decorators/query-all.ts
```diff
@@ -58,28 +59,13 @@ let fragment: DocumentFragment;
...
-  return (<C extends Interface<ReactiveElement>, V extends NodeList>(
-    protoOrTarget: ClassAccessorDecoratorTarget<C, V>,
-    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<C, V>
-  ) => {
-    const doQuery = (el: Interface<ReactiveElement>) => {
-      const container =
-        el.renderRoot ?? (fragment ??= document.createDocumentFragment());
-      return container.querySelectorAll(selector);
+  return (<C extends Interface<ReactiveElement>>() => {
+    return {
```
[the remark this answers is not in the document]
AndrewJakubowicz: Extremely glad you shared that playground :D

packages/reactive-element/src/decorators/query.ts
justinfagnani: renaming requires more syntax, that might be harder to read:

```ts
const {get: getValue, set: Value} = 
```

justinfagnani:
> I think ssr tests are failing in Github Actions because `queueMicrotask` is not defined. 🤔 Or something is going on there.

queueMicrotask has been in the code, so it shouldn't be that.

packages/reactive-element/src/reactive-element.ts
justinfagnani: There are tests that count the number of updates - even the number of `hasChanged()` calls, which happen every requestUpdate instead of every update. This behaves exactly the same in all the cases we currently test. We may be missing coverage for something like a property that reflects, is initialized to `undefined` _and_ has an attribute converter that converts `undefined` to a value rather than removing the attribute.

packages/reactive-element/src/decorators/query-all.ts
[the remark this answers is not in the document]
justinfagnani: Yeah, it just overwrites the auto-accessor. It's unfortunate that the accessors and private storage will be emitted and then not used. Our compiler will be needed to clean that up.

justinfagnani:
One problem with the workaround to https://github.com/microsoft/TypeScript/issues/55669 is that we have to read instance properties from auto-accessors in a microtask because their private storage will not be set up by the time this code runs in the ReactiveElement base class.

We were hoping to remove the microtask in the constructor as soon as https://github.com/tc39/proposal-decorators/issues/513 was resolved and shipping in TypeScript and Babel. Now we would need to keep it until either we end support for experimental decorators (years!) or https://github.com/microsoft/TypeScript/issues/55669 is addressed.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1048,7 +1048,11 @@ export abstract class ReactiveElement
       }
     }
     if (elementProperties.size > 0) {
-      queueMicrotask(() => {
+      // This has to run in a microtask so that we go after field
+      // initialization in subclasses. This microtask can be removed when both
+      // standard decorator addInitializer() runs after field initialization
+      // and experimental decorators are no longer supporter, or
```
AndrewJakubowicz, line 1054: trailing sentence, but I assume the `or` relates to: https://github.com/microsoft/TypeScript/issues/55669

AndrewJakubowicz (commented):
I'm unclear what the SSR impact is. Is it possible to synchronously SSR hybrid experimental decorators, or does this promise mean it will always be async?

justinfagnani:
@rictic @AndrewJakubowicz @sorvell I moved the initialization logic to the first update. Tests are passing locally now, so PTAL. Check that the comments make sense especially, and we'll have to document the subtleties of this behavior.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1390,6 +1390,35 @@ export abstract class ReactiveElement
...
+      // Trigger initial value reflection and populate the initial
+      // changedProperties map, but only for the case of experimental
+      // decorators on accessors, which will not have already populated the
+      // changedProperties map. We can't know if these accessors had
+      // initializers, so we just set them anyway - a difference from
+      // experimental decorators on fields and standard decorators on
+      // auto-accessors.
+      const elementProperties = (this.constructor as typeof ReactiveElement)
+        .elementProperties;
+      if (elementProperties.size > 0) {
```
rictic, line 1411: Ah, took me a sec to notice that this is still in the `!this.hasUpdated` condition

rictic (approved):

rictic:
Need to bump the size

packages/reactive-element/src/decorators/base.ts
```diff
@@ -17,15 +17,3 @@ export type Constructor<T> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
 };
-
-export const defineProperty = <T>(
```
AndrewJakubowicz, line 21: Since this runtime export was removed, there is no longer a `.js` emit for this file (it's only types).

This causes an import error in the `node-imports.ts` test. Fix is to remove `import '@lit/reactive-element/decorators/base.js';` from `src/test/node-imports.ts`

AndrewJakubowicz (approved):
Awesome!

AndrewJakubowicz (commented):
An optional nit, and something I can look at after this is merged is that it would help to see which environment is failing a particular decorator test.

For example when I tried the failing tests from https://github.com/lit/lit/issues/4191 in this change, it's difficult to tell which environment is failing.

packages/reactive-element/src/decorators/base.ts
justinfagnani: How did you notice this error? Nothing is failing here?

AndrewJakubowicz: I noticed the error because I locally ran these changes, and after a clean install `base.js` is no longer emitted (due to lack of runtime), and the node build fails. Not sure why it wasn't caught in GitHub actions.

PR #4184, opened 2023-09-06
[reactive-element] fix superclass properties being lost if subclass is decorated

AndrewJakubowicz:
# Context

Found this issue when migrating Material Web Components to use standard decorators.

They have a common pattern which is to have an _implementation_ base class, and then wrap it with a class that defines it with the custom element registry.

E.g.

```ts
class Button extends LitElement {
  // Lots of implementation
}

@customElement('md-button')
class MdButton extends Button { }
```

However, this would result in broken reactive properties.

I was able to shrink this to a reproducible test case which is checked into this PR.

# Fix

The fix is to change where `__collectMetadata` is called. Before this change the order of calls looked like this:

```
Finalize BaseClass
Finalize SubClass
CollectMetadata SubClass
CollectMetadata BaseClass
```

After this change, the order is now:

```
FinalizeBaseClass
CollectMetadata BaseClass
FinalizeSubClass
CollectMetadata SubClass
```

Why is this order important? `this.elementProperties = new Map(superCtor.elementProperties);` occurs in `finalize`, and is where the subclass lifts the base classes reactive properties. If collect metadata has not yet run, then there are no properties to lift, and the property is no longer tracked in the superclass.

This results in `MdButton` class having no reactive properties.

With the fixed order, the properties can be applied from the metadata before the subclass finalizes, and thus it collects the reactive properties correctly.

# Risk

There is some risk in this change because `collect metadata` now happens more eagerly, and it can also occur earlier if a finalize is triggered earlier by something like `addInitializer` etc.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -371,6 +371,7 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
  * this hack to bypass any rewriting by the compiler.
  */
 const finalized = 'finalized';
+const collectedMetadata = 'collectedMetadata';
 
 /**
  * A string representing one of the supported dev mode warning categories.
@@ -528,6 +529,7 @@ export abstract class ReactiveElement
    * from decorators.
    */
   protected static [finalized] = true;
+  protected static [collectedMetadata] = true;
 
   /**
    * Memoized list of all element properties, including any superclass properties.
@@ -619,6 +621,7 @@ export abstract class ReactiveElement
   static get observedAttributes() {
     // note: piggy backing on this to ensure we're finalized.
     this.finalize();
+    this.__collectMetadata();
     const attributes: string[] = [];
     for (const [p, v] of this.elementProperties) {
       const attr = this.__attributeNameForProperty(p, v);
@@ -772,6 +775,13 @@ export abstract class ReactiveElement
   // because `finalize` can be called before a class has been fully initialized.
   // This method should only be called once per class, as late as possible.
   private static __collectMetadata() {
+    if (this.hasOwnProperty(collectedMetadata)) {
+      return;
+    }
+    this[collectedMetadata] = true;
+    const superCtor = Object.getPrototypeOf(this) as typeof ReactiveElement;
+    superCtor.__collectMetadata();
+
     const metadata = this[Symbol.metadata];
     if (metadata == null) {
       return;
@@ -779,6 +789,11 @@ export abstract class ReactiveElement
     const properties = litPropertyMetadata.get(metadata);
     if (properties === undefined) {
       return;
+    }
+    // Grab reactive properties from superclass that may have been added by
+    // metadata.
+    for (const [k, v] of superCtor.elementProperties.entries()) {
+      this.elementProperties.set(k, v);
     }
     for (const [p, options] of properties) {
       this.elementProperties.set(p, options);
@@ -816,7 +831,6 @@ export abstract class ReactiveElement
       this._initializers = [...superCtor._initializers];
     }
     this.elementProperties = new Map(superCtor.elementProperties);
-    this.__collectMetadata();
     // initialize Map populated in observedAttributes
     this.__attributeToPropertyMap = new Map();
     // make any properties
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -619,7 +619,6 @@ export abstract class ReactiveElement
   static get observedAttributes() {
     // note: piggy backing on this to ensure we're finalized.
     this.finalize();
-    this.__collectMetadata();
     const attributes: string[] = [];
     for (const [p, v] of this.elementProperties) {
       const attr = this.__attributeNameForProperty(p, v);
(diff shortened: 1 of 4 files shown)


packages/reactive-element/src/test/std-decorators/property_test.ts
```diff
@@ -415,4 +415,23 @@ suite('@property', () => {
...
+      @property() accessor second = 'second';
+    }
+
+    const elName = generateElementName();
+    customElements.define(elName, SubClass);
+
+    container.innerHTML = `<${elName} first="overrideFirst"></${elName}>`;
+
+    const el: SubClass = container.querySelector(elName)!;
+    assert.equal(el.first, 'overrideFirst');
```
rictic, line 435: I love the practice of adding a test, then fixing that test in a PR, makes it so nice to verify behavior!

For future changes, it'd be better to demonstrate in the first commit the problematic behavior. I think this would involve doing:

```ts
    container.innerHTML = `<${elName} first="overrideFirst" second="overrideSecond"></${elName}>`;

    const el: SubClass = container.querySelector(elName)!;
    assert.equal(el.first, 'overrideFirst');
    // bug! this should be "overrideSecond"
    assert.equal(el.second, 'second');
```

rictic (approved):
Great find, and great fix!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -817,6 +816,7 @@ export abstract class ReactiveElement
       this._initializers = [...superCtor._initializers];
     }
     this.elementProperties = new Map(superCtor.elementProperties);
+    this.__collectMetadata();
```
justinfagnani, line 819: I'm curious how this isn't breaking tests. Are we lacking some coverage? `finalize()` can be called from `addInitializer()` and `createProperty()`. If either of those are called before `get observedAttributes` and before all the decorators are processed, then we should only convert a subset of the decorator metadata to elementProperties.

I think we at least need a test that has a decorator application that adds an initializer followed by `@property()` decorators.

justinfagnani (changes requested):
Would this order work:

```
FinalizeBaseClass
FinalizeSubClass
CollectMetadata BaseClass
CollectMetadata SubClass
```

I am worried about how eagerly `finalize()` can be called.

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: Completely agree with you. In the material-web case, the decorator on their super class is actually a `@customElement` decorator, which then results in every single property losing reactivity (since properties are only set in super classes).

Will add more tests.

packages/reactive-element/src/test/std-decorators/property_test.ts
AndrewJakubowicz: Rebased commits so that it follows this excellent pattern.

AndrewJakubowicz:
> Would this order work:
> 
> ```
> FinalizeBaseClass
> FinalizeSubClass
> CollectMetadata BaseClass
> CollectMetadata SubClass
> ```
> 
> I am worried about how eagerly `finalize()` can be called.

I implemented your proposal (and it passes tests). I like it better because it keeps `__collectMetadata` very late. The only _con_ is that it requires a little more lifting of superclass properties, because if `finalize` is called before `__collectMetadata`, then `this.elementProperties = new Map(superCtor.elementProperties);` runs too early. And any properties added to the superclass after get lost.

Thank you!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -781,6 +790,11 @@ export abstract class ReactiveElement
     if (properties === undefined) {
       return;
     }
+    // Grab reactive properties from superclass that may have been added by
+    // metadata.
+    for (const [k, v] of superCtor.elementProperties.entries()) {
+      this.elementProperties.set(k, v);
+    }
```
AndrewJakubowicz, line 797: The timing of this bothers me. If I lift this up to L784 it results in other tests failing. But here or on L789 is the perfect spot – and I don't yet understand why.

sorvell (changes requested):
This issue raises a yellow flag for me on the standard decorator strategy, and I think it's worth zooming out a bit:

* the addition of collect metadata makes the finalization process more complex and harder to maintain.
* standard decorators bypass the `getPropertyDescriptor` API which could be an obvious or subtle breaking change.
* standard decorators require subclasses that want to change a reactive property default to redecorate the property.
* standard decorator initializer requiring a new argument to requestUpdate is, apparently, back?

Based on the tricky differences using typescript with `accessor` and `useDefineForClassFields`, it might perhaps be best to try to just get standard decorators working exactly as the previous version with just these changes. To do this, I think the following would be required:
1. Simulate metadata collection in legacy decorators.
2. Disallow calling finalization early and instead just collect metadata.
3. Integrate metadata processing into finalization (this is very similar to the properties object after all)
4. Make all property decorators, legacy and standard call `createProperty`. (A future version which avoids this and relies on the built in decorator mechanism can be done in the next version as a specific feature change)
5. Verify that (for standard and legacy) a feature can be added which optionally seeds the initial value with a set or just setting the private storage, thus potentially avoiding initial reflection.

packages/reactive-element/src/reactive-element.ts
AndrewJakubowicz: Nvm – it's a bug in this implementation. You can overwrite properties by lifting twice.

I think the simpler solution may be much more robust:

```
FinalizeBaseClass
CollectMetadata BaseClass
FinalizeSubClass
CollectMetadata SubClass
```

But adding a bunch more tests.

AndrewJakubowicz:
This PR has been superseded by https://github.com/lit/lit/pull/4204

PR #4185, opened 2023-09-07
Add mixin to use ViewTransition

KeithHenry:
Using the  (View Transitions API)[https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API]

This mixin extends `performUpdate` to use `document.startViewTransition`

See discussion https://github.com/lit/lit/discussions/4179

### DIFF
--- a/packages/labs/motion/src/ViewTransitionUpdate.ts
+++ b/packages/labs/motion/src/ViewTransitionUpdate.ts
@@ -1,0 +1,42 @@
+import {LitElement} from 'lit';
+
+/** Mixin to extend a LitElement to use view transitions (if available) on DOM updates */
+export function ViewTransitionUpdate<TBase extends LitElement>(Base: TBase) {
+    return class extends LitElement {
+        private lifecycleRenderTransition: ViewTransition;
+  
+        async performUpdate() {
+            if (!(document as DocumentWithTransition).startViewTransition)
+                return await super.performUpdate();
+    
+            if (this.lifecycleRenderTransition)
+                await this.lifecycleRenderTransition.finished;
+    
+            this.lifecycleRenderTransition = (document as DocumentWithTransition).startViewTransition(
+                async () => 
+                    await super.performUpdate());
+        }
+    };
+  }
+
+interface DocumentWithTransition 
+    extends Document {
+    /** The startViewTransition() method of the View Transitions API starts a new view transition and returns a ViewTransition object to represent it.  
+     When startViewTransition() is invoked, a sequence of steps is followed as explained in The view transition process. 
+     @param callback A callback function typically invoked to update the DOM during the view transition process, which returns a Promise. The callback is invoked once the API has taken a screenshot of the current page. When the promise returned by the callback fulfills, the view transition begins in the next frame. If the promise returned by the callback rejects, the transition is abandoned.*/
+    startViewTransition(callback: () => Promise<unknown>): ViewTransition;
+}
+
+interface ViewTransition { 
+    /** A Promise that fulfills once the transition animation is finished, and the new page view is visible and interactive to the user. */
+    finished: Promise<unknown>;
+
+    /** A Promise that fulfills once the pseudo-element tree is created and the transition animation is about to start. */
+    ready: Promise<unknown>;
+
+    /** A Promise that fulfills when the promise returned by the document.startViewTransition()'s callback fulfills. */
+    updateCallbackDone: Promise<unknown>;
+
+    /** Skips the animation part of the view transition, but doesn't skip running the document.startViewTransition() callback that updates the DOM. */
+    skipTransition(): void;
+}


packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
+import {LitElement} from 'lit';
```
justinfagnani, line 1: Add license header like we have in other files, copyright 2023

packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type Constructor<T> = new (...args: any[]) => T;
+
+/** Extend a LitElement to use view transitions (if available) on DOM updates */
+export const ViewTransitionUpdate =
+    <T extends Constructor<LitElement>>(superClass: T) => {
+    class LitElementTransition extends superClass {
+        private lifecycleRenderTransition: ViewTransition;
+  
+        async performUpdate() {
```
justinfagnani, line 11: `performUpdate()` can't be async in Lit 3.0, so we need to await in `scheduleUpdate()` - but this approach here also nicely causes the current updateComplete promise to wait, not just the scheduler to wait to start the next update based on a current transition. I'll have to think about that change in 3.0 a bit.

See also [this example](https://lit.dev/playground/#project=W3sibmFtZSI6ImFwcC50cyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcbmltcG9ydCB7Y3VzdG9tRWxlbWVudCwgcHJvcGVydHl9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJztcbmltcG9ydCB7IHRyYW5zaXRpb25IZWxwZXIgfSBmcm9tICcuL3V0aWxzLmpzJztcblxuQGN1c3RvbUVsZW1lbnQoJ2RlbW8tYXBwJylcbmV4cG9ydCBjbGFzcyBEZW1vQXBwIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHN0YXRpYyBzdHlsZXMgPSBjc3NgXG4gICAgLmNvdW50IHtcbiAgICAgIGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xuICAgICAgdGV4dC1hbGlnbjogY2VudGVyO1xuICAgICAgcG9zaXRpb246IGFic29sdXRlO1xuICAgICAgaW5zZXQ6IDUwJSAwIGF1dG87XG4gICAgICB0cmFuc2Zvcm06IHRyYW5zbGF0ZVkoLTUwJSk7XG4gICAgICBmb250LXNpemU6IDI1dnc7XG4gICAgICB2aWV3LXRyYW5zaXRpb24tbmFtZTogY291bnQ7XG4gICAgICAvKiBUaGlzIHdvbid0IGJlIHJlcXVpcmVkIHNvb24uIEluIGZhY3QsIGl0IGFscmVhZHkgd29ya3Mgd2l0aG91dCB0aGlzIGluIENhbmFyeSAqL1xuICAgICAgY29udGFpbjogbGF5b3V0O1xuICAgIH1cbiAgYDtcbiAgXG4gIGluY3JlbWVudENsaWNrKCkge1xuICAgIHRyYW5zaXRpb25IZWxwZXIoe1xuICAgICAgdXBkYXRlRE9NOiBhc3luYyAoKSA9PiB7XG4gICAgICAgIHRoaXMuY291bnQrKztcbiAgICAgICAgYXdhaXQgdGhpcy51cGRhdGVDb21wbGV0ZTtcbiAgICAgIH1cbiAgICB9KTtcbiAgfVxuXG4gIEBwcm9wZXJ0eSgpXG4gIGNvdW50ID0gMDtcblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgXG4gICAgICA8YnV0dG9uIEBjbGljaz0ke3RoaXMuaW5jcmVtZW50Q2xpY2t9PkluY3JlbWVudDwvYnV0dG9uPlxuICAgICAgPGRpdiBjbGFzcz1cImNvdW50XCI-JHt0aGlzLmNvdW50fTwvZGl2PlxuICAgIGA7XG4gIH1cbn1cbiJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHN0eWxlPlxuICAgIGh0bWwge1xuICAgICAgYmFja2dyb3VuZDogYmxhY2s7XG4gICAgICBjb2xvcjogI2ZmZjtcbiAgICB9XG5cbiAgICAvKiBDdXN0b20gdHJhbnNpdGlvbiAqL1xuICAgIEBrZXlmcmFtZXMgcm90YXRlLW91dCB7XG4gICAgICB0byB7XG4gICAgICAgIHRyYW5zZm9ybTogcm90YXRlKDkwZGVnKTtcbiAgICAgIH1cbiAgICB9XG5cbiAgICBAa2V5ZnJhbWVzIHJvdGF0ZS1pbiB7XG4gICAgICBmcm9tIHtcbiAgICAgICAgdHJhbnNmb3JtOiByb3RhdGUoLTkwZGVnKTtcbiAgICAgIH1cbiAgICB9XG5cbiAgICBodG1sOjp2aWV3LXRyYW5zaXRpb24tb2xkKGNvdW50KSxcbiAgICBodG1sOjp2aWV3LXRyYW5zaXRpb24tbmV3KGNvdW50KSB7XG4gICAgICBhbmltYXRpb24tZHVyYXRpb246IDIwMG1zO1xuICAgICAgYW5pbWF0aW9uLW5hbWU6IC11YS12aWV3LXRyYW5zaXRpb24tZmFkZS1pbiwgcm90YXRlLWluO1xuICAgIH1cblxuICAgIGh0bWw6OnZpZXctdHJhbnNpdGlvbi1vbGQoY291bnQpIHtcbiAgICAgIGFuaW1hdGlvbi1uYW1lOiAtdWEtdmlldy10cmFuc2l0aW9uLWZhZGUtb3V0LCByb3RhdGUtb3V0O1xuICAgIH1cbiAgPC9zdHlsZT5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9hcHAuanNcIj48L3NjcmlwdD5cbjwvaGVhZD5cbjxib2R5PlxuICA8ZGVtby1hcHA-PC9kZW1vLWFwcD5cbjwvYm9keT4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXRcIjogXCJeMi4wLjBcIixcbiAgICBcIkBsaXQvcmVhY3RpdmUtZWxlbWVudFwiOiBcIl4xLjAuMFwiLFxuICAgIFwibGl0LWVsZW1lbnRcIjogXCJeMy4wLjBcIixcbiAgICBcImxpdC1odG1sXCI6IFwiXjIuMC4wXCJcbiAgfVxufSIsImhpZGRlbiI6dHJ1ZX0seyJuYW1lIjoidXRpbHMudHMiLCJjb250ZW50IjoiZXhwb3J0IGZ1bmN0aW9uIHRyYW5zaXRpb25IZWxwZXIoe1xuICBza2lwVHJhbnNpdGlvbiA9IGZhbHNlLFxuICBjbGFzc05hbWVzID0gW10sXG4gIHVwZGF0ZURPTSxcbn0pIHtcbiAgaWYgKHNraXBUcmFuc2l0aW9uIHx8ICFkb2N1bWVudC5zdGFydFZpZXdUcmFuc2l0aW9uKSB7XG4gICAgY29uc3QgdXBkYXRlQ2FsbGJhY2tEb25lID0gUHJvbWlzZS5yZXNvbHZlKHVwZGF0ZURPTSgpKS50aGVuKCgpID0-IHt9KTtcbiAgICBjb25zdCByZWFkeSA9IFByb21pc2UucmVqZWN0KEVycm9yKCdWaWV3IHRyYW5zaXRpb25zIHVuc3VwcG9ydGVkJykpO1xuXG4gICAgLy8gQXZvaWQgc3BhbW1pbmcgdGhlIGNvbnNvbGUgd2l0aCB0aGlzIGVycm9yIHVubGVzcyB0aGUgcHJvbWlzZSBpcyB1c2VkLlxuICAgIHJlYWR5LmNhdGNoKCgpID0-IHt9KTtcblxuICAgIHJldHVybiB7XG4gICAgICByZWFkeSxcbiAgICAgIHVwZGF0ZUNhbGxiYWNrRG9uZSxcbiAgICAgIGZpbmlzaGVkOiB1cGRhdGVDYWxsYmFja0RvbmUsXG4gICAgICBza2lwVHJhbnNpdGlvbjogKCkgPT4ge30sXG4gICAgfTtcbiAgfVxuXG4gIGRvY3VtZW50LmRvY3VtZW50RWxlbWVudC5jbGFzc0xpc3QuYWRkKC4uLmNsYXNzTmFtZXMpO1xuXG4gIGNvbnN0IHRyYW5zaXRpb24gPSBkb2N1bWVudC5zdGFydFZpZXdUcmFuc2l0aW9uKHVwZGF0ZURPTSk7XG5cbiAgdHJhbnNpdGlvbi5maW5pc2hlZC5maW5hbGx5KCgpID0-XG4gICAgZG9jdW1lbnQuZG9jdW1lbnRFbGVtZW50LmNsYXNzTGlzdC5yZW1vdmUoLi4uY2xhc3NOYW1lcylcbiAgKTtcblxuICByZXR1cm4gdHJhbnNpdGlvbjtcbn0ifV0) from https://developer.chrome.com/docs/web-platform/view-transitions/#working-with-frameworks

packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
...
+            if (!(document as DocumentWithTransition).startViewTransition)
+                return await super.performUpdate();
```
justinfagnani, line 13: run `npm run format` from the monorepo root

packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
...
+    
+            if (this.lifecycleRenderTransition)
+                await this.lifecycleRenderTransition.finished;
```
justinfagnani, line 16: Given the discussion thread, we'll need a try/catch and a comment on why this is the right promise to wait instead of say updateCallbackDone.

And it is a good opinion to await finished? It makes sense to me, but maybe we need an option? As this is labs we could go with this and wait for feedback.

packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
...
+    
+            this.lifecycleRenderTransition = (document as DocumentWithTransition).startViewTransition(
+                async () =>
```
justinfagnani, line 19: async not needed. If `super.performUpdate()` were async, it would return a Promise that we just forward.

packages/labs/motion/src/view-transition-update.ts
```diff
@@ -0,0 +1,48 @@
+import {LitElement} from 'lit';
+
+type Constructor<T> = new (...args: any[]) => T;
+
+/** Extend a LitElement to use view transitions (if available) on DOM updates */
```
justinfagnani, line 5: I think we need to extend these docs to explain the choices this makes, especially what timing this operates on (waits for ViewTransition.finished) and how this effects the LitElement lifecycle (it should delay update complete until the transition is finished).

justinfagnani (changes requested):
Even though this is labs, we need at least some kind of test to make sure we don't completely break things.

I'm not sure exactly what to check for here. I don't see that JS can detect that a transition is happening or finished globally. The `transitionend` event doesn't seem to be fired on document. Maybe we can at least check that the DOM updated in both the case that the browser supports view transitions and not.

packages/labs/motion/src/view-transition-update.ts
KeithHenry: Done e62776a9dd534c451d6530eeb4c21a347f842f30

packages/labs/motion/src/view-transition-update.ts
KeithHenry: try-catch and comment added in a1b5fa32a4916a1b704b9289755b7e2f891bdc76

I'm not certain about `await finished` vs `await updateCallbackDone` vs `skipTransition()` either.

packages/labs/motion/src/view-transition-update.ts
KeithHenry: I've added d8f3449121029757fbc52a161fa0073f790b91e9 to removed that `await` and add `return this.updateComplete;` instead, as we need the DOM changes to complete before the promise returns.

packages/labs/motion/src/view-transition-update.ts
KeithHenry: @justinfagnani hmm, am I missing a step? We need to tell the transition once the DOM update has finished before it runs the anim, and `performUpdate` isn't and shouldn't be `async`, but we'd potentially want `updateComplete` to only return once the anim has finished?

Is there a promise before `updateComplete` I can pass to `startViewTransition` instead?

packages/labs/motion/src/view-transition-update.ts
KeithHenry: Done in f1b75c87f7cef8ed6c2db7e6ac44fe3facbb2c04

PR #4196, opened 2023-09-16
Fix SSR truncated response

### LINKED ISSUE #4187: [labs/ssr] Truncated response

SSR (@lit-labs/ssr)

### Description

During the rendering of a component having a list with 1000 elements, the SSR response is truncated - only some part of the items is rendered.

### Reproduction

https://stackblitz.com/edit/lit-ssr-global-wydn3p?file=src%2Fcomponents%2Fsimple-greeter.js

### Workaround

The error happens in `RenderResultReadable`, in a code branch, which is executed when the consumer cannot accept more data (https://github.com/lit/lit/blob/bb4d10704e82cd4ebfea6dede662e6233bb4c086/packages/labs/ssr/src/lib/render-result-readable.ts#L76C11-L76C18). We return from the `_read`, but the iterator may still have next values. On the subsequent `_read` call, we pop the next iterator (https://github.com/lit/lit/blob/bb4d10704e82cd4ebfea6dede662e6233bb4c086/packages/labs/ssr/src/lib/render-result-readable.ts#L60C38-L60C38), ignoring the values from the previous iterator that were not sent because of back pressure. 

Probably the correct approach would be to keep the current iterator as an instance variable so that when the subsequent `_read` is called, we can send the rest of the values from the previous call.

### Is this a regression?

### Affected versions

3.1.2

Browser: 116.0.5845.179
OS: 13.5.1
Node: v18.17.1

robak86:
Fixes #4187

### DIFF
--- a/packages/labs/ssr/src/lib/render-result-readable.ts
+++ b/packages/labs/ssr/src/lib/render-result-readable.ts
@@ -57,10 +57,9 @@ export class RenderResultReadable extends Readable {
     // result, because we must be able to return in the middle of the loop
     // and resume on the next call to _read().
 
-    // Get the current iterator, only if we don't already have one from the previous call to _read()
-    if (!this._currentIterator) {
-      this._currentIterator = this._iterators.pop();
-    }
+    // Get the current iterator, only if we don't already have one from the
+    // previous call to _read()
+    this._currentIterator ??= this._iterators.pop();
 
     while (this._currentIterator !== undefined) {
       const next = this._currentIterator.next();
--- a/packages/labs/ssr/src/lib/render-result-readable.ts
+++ b/packages/labs/ssr/src/lib/render-result-readable.ts
@@ -23,6 +23,7 @@ export class RenderResultReadable extends Readable {
    * completion in any one loop.
    */
   private _iterators: Array<RenderResultIterator>;
+  private _currentIterator?: RenderResultIterator;
 
   constructor(result: RenderResult) {
     super();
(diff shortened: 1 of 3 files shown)


packages/labs/ssr/src/test/lib/render-result-readable_test.ts
```diff
@@ -40,6 +40,27 @@ test('RenderResultReadable collects strings and nested Promises of iterables', a
   assert.equal(s, 'abcd');
 });
 
+test('RenderResultReadable collects all iterables 2', async () => {
```
AndrewJakubowicz, line 43: Nit: Could the test name be updated to be more specific of what is being tested?

AndrewJakubowicz (approved):
Looks great! Thank you so much! Verified the issue is fixed with this change.

Could you please add a changeset?

packages/labs/ssr/src/lib/render-result-readable.ts
```diff
@@ -56,14 +57,16 @@ export class RenderResultReadable extends Readable {
     // result, because we must be able to return in the middle of the loop
     // and resume on the next call to _read().
 
-    // Get the current iterator
-    let iterator = this._iterators.pop();
+    // Get the current iterator, only if we don't already have one from the previous call to _read()
+    if (!this._currentIterator) {
+      this._currentIterator = this._iterators.pop();
+    }
```
justinfagnani, line 63: nits: wrap comments at 80 cols and use nullish-coalescing assignment:

```suggestion
    // Get the current iterator, only if we don't already have one from the
    // previous call to _read()
    this._currentIterator ??= this._iterators.pop();
```

justinfagnani (approved):
Awesome, thanks!!

I have just one small comment

PR #4200, opened 2023-09-18
Support @provide with experimental decorators on a property

rictic:
It might feel like this is working at cross purposes to our standard decorators work, because those all require accessors, but I don't think so. With standard decorators we can use the type system to error out if they're used on a property instead of an accessor. With experimental decorators, they sorta work with properties, in that they do provide values and update them, but reading the value back off the field would always give undefined because we were defining it without a getter.

### DIFF
--- a/packages/labs/context/src/lib/decorators/provide.ts
+++ b/packages/labs/context/src/lib/decorators/provide.ts
@@ -86,17 +86,14 @@ export function provide<ValueType>({
       );
       let newDescriptor;
       if (descriptor === undefined) {
-        const symbol = Symbol();
+        const valueMap = new WeakMap<ReactiveElement, ValueType>();
         newDescriptor = {
-          get: function (this: ReactiveElement & {[symbol]: ValueType}) {
-            return this[symbol];
+          get: function (this: ReactiveElement) {
+            return valueMap.get(this);
           },
-          set: function (
-            this: ReactiveElement & {[symbol]: ValueType},
-            value: ValueType
-          ) {
-            controllerMap.get(this)?.setValue(value);
-            this[symbol] = value;
+          set: function (this: ReactiveElement, value: ValueType) {
+            controllerMap.get(this)!.setValue(value);
+            valueMap.set(this, value);
           },
         };
       } else {
@@ -105,9 +102,7 @@ export function provide<ValueType>({
           ...descriptor,
           set: function (this: ReactiveElement, value: ValueType) {
             controllerMap.get(this)?.setValue(value);
-            if (oldSetter) {
-              oldSetter.call(this, value);
-            }
+            oldSetter?.call(this, value);
           },
         };
       }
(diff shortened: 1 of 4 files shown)


Marbitch:
Someone is hacking my accounts and they are using my google to contact you,
please let me know who I can contact to report this!

packages/labs/context/src/lib/decorators/provide.ts
```diff
@@ -84,16 +84,33 @@ export function provide<ValueType>({
...
+        };
+      } else {
+        const oldSetter = descriptor?.set;
+        newDescriptor = {
+          ...descriptor,
+          set: function (this: ReactiveElement, value: ValueType) {
+            controllerMap.get(this)?.setValue(value);
+            if (oldSetter) {
+              oldSetter.call(this, value);
+            }
```
justinfagnani, line 110:
```suggestion
            oldSetter?.call(this, value);
```

packages/labs/context/src/lib/decorators/provide.ts
```diff
@@ -84,16 +84,33 @@ export function provide<ValueType>({
...
+        const symbol = Symbol();
+        newDescriptor = {
+          get: function (this: ReactiveElement & {[symbol]: ValueType}) {
+            return this[symbol];
+          },
+          set: function (
+            this: ReactiveElement & {[symbol]: ValueType},
+            value: ValueType
+          ) {
+            controllerMap.get(this)?.setValue(value);
```
justinfagnani, line 98: This element will always be in the map by this point, so I don't think we need to optional chaining.

packages/labs/context/src/lib/decorators/provide.ts
```diff
@@ -84,16 +84,33 @@ export function provide<ValueType>({
...
-        set: function (this: ReactiveElement, value: ValueType) {
-          controllerMap.get(this)?.setValue(value);
-          if (oldSetter) {
-            oldSetter.call(this, value);
-          }
-        },
-      };
+      let newDescriptor;
+      if (descriptor === undefined) {
+        const symbol = Symbol();
```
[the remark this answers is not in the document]
rictic: Switched to a WeakMap for consistency in this file

packages/labs/context/src/lib/decorators/provide.ts
rictic: Switched to not-null assertion

packages/labs/context/src/lib/decorators/provide.ts
rictic: Done

packages/labs/context/src/test/provider-consumer_test.ts
```diff
@@ -137,4 +143,49 @@ suite('context-provider', () => {
     assert.strictEqual(consumer.value, 500);
     assert.strictEqual(consumer2.value, 1000); // one-time consumer still has old value
   });
+
```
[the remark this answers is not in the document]
rictic: Yeah, done

justinfagnani (approved):
nice!

Issue #4191, opened 2023-09-12
Hybrid decorators do not lift reactive properties through subclasses


AndrewJakubowicz:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

See https://github.com/lit/lit/pull/4184 which reproduces the issue.

This was found when testing the new Hybrid decorators in the MWC repo. They use a pattern that looks like this:

```
class Button extends LitElement {
  @property() exampleProp = '';
  // Lots of implementation
}

@customElement('md-button')
class MdButton extends Button { }
```

The addition of `@customElement` on `MdButton` causes the reactive property `exampleProp` to not work correctly on `MdButton`.

A possible workaround is replacing `@customElement` with a `customElements.define(...)`, which fixes the issue.

See the failing test cases here: https://github.com/lit/lit/pull/4184/commits/b846db696516a284cc5c5f700a20fb71c620edaa

```ts
class BaseTest extends ReactiveElement {
  @property() first = 'first';
}

class SubClass extends BaseTest {
  @property() second = 'second';
}
```

Now if the `SubClass` is initialized with: `container.innerHTML = '<sub-class first="overrideFirst" second="overrideSecond"></sub-class>'` 

`subclass.first` is `first`, but I expected `overrideFirst`.
`subclass.second` is correctly set to `overrideSecond`.

### Reproduction

na

### Workaround

na

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

Not released

### Browser/OS/Node environment

N/A

AndrewJakubowicz:
Fixed by https://github.com/lit/lit/pull/4204

PR #4201, opened 2023-09-19
[reactive-element] fix standard decorator subclassing

AndrewJakubowicz:
Fixes https://github.com/lit/lit/issues/4191

Currently this PR is failing tests. These tests are copied from `reactive-element_test.ts` and migrated to use standard decorators.
If tests pass then our standard decorators will maintain the behavior of static properties block.


justinfagnani:
Closing in favor of https://github.com/lit/lit/pull/4204

PR #4202, opened 2023-09-19
Put types on Map, Set, etc constructor calls

rictic:
These default to overbroad types like `any` and `object`.

Didn't change all tests, nor cases where we're assigning to an existing thing that has a type.


justinfagnani (approved):

AndrewJakubowicz (approved):
Awesome! Maybe changeset for ssr so it gets pulled into the release.

PR #4203, opened 2023-09-19
Some small fixups for jscompiler

rictic:


AndrewJakubowicz:
`@localized`  test is failing

packages/reactive-element/src/decorators/query-all.ts
```diff
@@ -59,9 +59,9 @@ let fragment: DocumentFragment;
  * @category Decorator
  */
 export function queryAll(selector: string): QueryAllDecorator {
-  return (<C extends Interface<ReactiveElement>>() => {
+  return (() => {
     return {
-      get(this: C) {
+      get(this: ReactiveElement) {
```
AndrewJakubowicz, line 64: Discussed offline if these should be `Interface<ReactiveElement>`?

AndrewJakubowicz (approved):

PR #4204, opened 2023-09-20
3.0 decorator subclass properties

justinfagnani:
Fixes https://github.com/lit/lit/issues/4191

This is _a_ fix, so the problem and most basic solution can be reviewed.

I'm not entirely happy with it since it requires another instance field and increase the size a bit. I'm going to refactor `finalize()` and `__collectMetadata()` to hopefully make finalization make more sense and get the size down some.

I made a new PR from https://github.com/lit/lit/pull/4201 so @AndrewJakubowicz can be added as a reviewer.


justinfagnani:
I refactored `finalize()` and `__collectMetadata()` and got this down to a +59 byte change.

I also cleaned up the shadowed property check so that it uses the `elementProperties` keys instead of a dev-mode-only separate set, which had the same keys.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -779,76 +750,90 @@ export abstract class ReactiveElement
...
-    if (properties === undefined) {
+  /**
+   * Initializes static own properties of the class used in bookkeeping
+   * for element properties, initializers, etc.
+   *
+   * Can be called multiple times by code that needs to ensure these
+   * properties exist before using them.
+   *
+   * This method ensures the superclass is finalized so that inherited
+   * property metadata can be copied down.
```
AndrewJakubowicz, line 761: nit: needs `@nocollapse`

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1418,6 +1398,7 @@ export abstract class ReactiveElement
             !this._$changedProperties.has(p) &&
             this[p as keyof this] !== undefined
           ) {
+            // console.log(this.constructor.name, 'reflect initial', p);
```
AndrewJakubowicz, line 1401: nit: left log

AndrewJakubowicz (commented):
Tiny nits, need to review properly tomorrow.
Amazing!!!!! Great stuff!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -608,35 +609,21 @@ export abstract class ReactiveElement
...
-  /**
-   * The set of properties defined by this class that caused an accessor to be
-   * added during `createProperty`.
-   *
-   * This field is only used in dev mode. It should be eliminated in prod
-   * builds.
-   *
-   * @nocollapse
-   */
-  private static __reactivePropertyKeys?: Set<PropertyKey>;
```
AndrewJakubowicz, line 620: Removing this caused a failing DEV_MODE test case in GitHub Actions.

justinfagnani: yeah... working on it

packages/reactive-element/src/reactive-element.ts
justinfagnani: done

packages/reactive-element/src/reactive-element.ts
justinfagnani: fixed

packages/reactive-element/src/reactive-element.ts
justinfagnani: fixed

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1372,16 +1353,15 @@ export abstract class ReactiveElement
...
-          if (this.hasOwnProperty(p) && !this.__instanceProperties?.has(p)) {
-            shadowedProperties.push(p as string);
-          }
-        });
+        // Produce warning if any reactive properties on the prototype are
+        // shadowed by class fields. Instance fields set before upgrade are
+        // deleted by this point, so any own property is caused by class field
+        // initialization in the constructor.
+        const ctor = this.constructor as typeof ReactiveElement;
+        const shadowedProperties = [...ctor.elementProperties.keys()].filter(
```
sorvell, line 1361: Not exactly sure why this didn't use `elementProperties` previously, but I think it might have been to not warn on `noAccessor` properties, but that doesn't seem important given the prototype in check so this seems ok. Can double check with @bicknellr who I think did the initial warning code.

sorvell (approved):
LGTM

packages/reactive-element/src/test/decorators-modern/property_test.ts
```diff
@@ -415,4 +415,181 @@ suite('@property', () => {
...
+
+    assert.equal(sub.foo, 6);
+    assert.isFalse(sub.hasAttribute('zug'));
+    assert.isFalse(sub.hasAttribute('foo'));
+    assert.equal(sub.bar, 'subbar');
+    assert.equal(sub.getAttribute('bar'), 'subbar');
+    assert.equal(sub.nug, 5);
+  });
+
+  test('subclass contains super classes reactive properties', async () => {
```
rictic, line 568: An apostrophe would help legibility here a bit

```ts
`subclass contains super classes' reactive properties`

File: packages/reactive-element/src/reactive-element.ts [hunk 6]
```diff
@@ -37,6 +37,7 @@ const {
   getOwnPropertyDescriptor,
   getOwnPropertyNames,
   getOwnPropertySymbols,
+  getPrototypeOf,
```
rictic, line 40: Is this smaller? It's writing getPrototypeOf twice instead of once to save an `Object`

justinfagnani: It's only writing it once - the second reference can be renamed.

Because we're already tearing off Object statics, I think even one use pays for itself. This block is rewritten to something like:

```ts
const{is:e,defineProperty:t,getOwnPropertyDescriptor:r,getOwnPropertyNames:o,getOwnPropertySymbols:p,getPrototypeOf:y}=Object;
```

So the total cost is `,getPrototypeOf:y` + `y(this)` instead of `Object.getPrototypeOf(this)`, so it's essentially a minification of just `Object.` to whatever the variable is.

The other benefit is that there isn't an object dereference of `.getPrototypeOf` at the call site.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -779,76 +750,91 @@ export abstract class ReactiveElement
...
+   * Can be called multiple times by code that needs to ensure these
+   * properties exist before using them.
+   *
+   * This method ensures the superclass is finalized so that inherited
+   * property metadata can be copied down.
    * @nocollapse
    */
-  protected static finalize() {
-    if (this.hasOwnProperty(finalized)) {
-      return false;
```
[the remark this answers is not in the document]
justinfagnani: Not that I could see!

AndrewJakubowicz (approved):
Nice!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -967,7 +952,7 @@ export abstract class ReactiveElement
   /**
    * Map with keys of properties that should be reflected when updated.
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/reactive-element/src/test/decorators-modern/property_test.ts
justinfagnani: done

packages/reactive-element/src/reactive-element.ts
```diff
@@ -779,76 +750,91 @@ export abstract class ReactiveElement
...
         ...getOwnPropertyNames(props),
         ...getOwnPropertySymbols(props),
       ];
-      // This for/of is ok because propKeys is an array
       for (const p of propKeys) {
-        // note, use of `any` is due to TypeScript lack of support for symbol in
+        // Use of `any` is due to TypeScript lack of support for symbol in
         // index types
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         this.createProperty(p, (props as any)[p]);
```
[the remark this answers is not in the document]
justinfagnani: ncie! done

packages/reactive-element/src/reactive-element.ts
justinfagnani: I moved a `noAccessor` check to here before removing it in favor of the `in` check. The `in` check should subsume everything.

rictic (approved):

PR #4214, opened 2023-09-21
Add one more jsconformance workaround.

rictic:


packages/reactive-element/src/decorators/property.ts
```diff
@@ -33,7 +33,14 @@ if (DEV_MODE) {
   issueWarning = (code: string, warning: string) => {
     warning += ` See https://lit.dev/msg/${code} for more information.`;
     if (!issuedWarnings.has(warning)) {
-      console.warn(warning);
+      // Work around a jsconformance rule that fails when we might be
+      // logging outside of DEV_MODE, but this log is kosher because it is
+      // only in DEV_MODE.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
justinfagnani, line 39: Can we not copybara in some google3-specific check in addition to or instead of DEV_MODE?

packages/reactive-element/src/decorators/property.ts
```diff
@@ -33,7 +33,14 @@ if (DEV_MODE) {
...
+      const cons = (globalThis as any)['cons' + 'ole'] as {
```
justinfagnani, line 40: Does the conformance check yell for any access to `console` as well as calling any methods on it? This couldn't be `globalThis['console'] as ...` to just replace the type?

rictic:
Yeah, just patching this on import

PR #4215, opened 2023-09-25
[infra] fix reactive-element node test

AndrewJakubowicz:
### Context

`npm run test:node` in the `@lit/reactive-element` workspace was failing with error:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'lit/packages/reactive-element/node/decorators/base.js' imported from lit/packages/reactive-element/development/test/node-imports.js
```

### Fix

`base.js` was recently updated to only contain types and no runtime code. Thus TypeScript no longer emits a `base.js` file causing this test to fail.

### Why was this not detected in Github Actions

I am not sure. I can reproduce locally. Needs further investigation in a separate change.


rictic (approved):

augustjk (approved):

augustjk:
https://github.com/lit/lit/pull/4228 is adding stuff back to `base.ts` so this might not be needed anymore.

AndrewJakubowicz:
> #4228 is adding stuff back to `base.ts` so this might not be needed anymore.

oops! Missed this. I wonder if this is worth reverting, or if it's fine that the imports indirectly import `base.js`.

augustjk:
I think it's fine to keep as is. We really mainly care about entrypoint imports.

PR #4217, opened 2023-09-25
[repro] Next.js dynamic ref is broken

AndrewJakubowicz:
Reproduces https://github.com/lit/lit/issues/2951

The only issue that still reproduces seems to be `ref` behavior with the dynamically imported component.

Repro by running `npm run dev`, and pressing btn to see `ref` fail. It works if the react component is imported normally (not via `dynamic`).

### DIFF
--- a/examples/nextjs/pages/index.tsx
+++ b/examples/nextjs/pages/index.tsx
@@ -1,33 +1,33 @@
 import Head from 'next/head';
 import styles from '../styles/Home.module.css';
 import '../src/simple-greeter';
+import {SimpleGreeter as SimpleGreeterWC} from '../src/simple-greeter';
 import SimpleGreeter from '../src/simple-greeter-react';
 
-export default function Home() {
-  return (
-    <div className={styles.container}>
-      <Head>
-        <title>Lit in Next.js</title>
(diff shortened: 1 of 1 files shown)


justinfagnani:
Thanks @AndrewJakubowicz this is helpful!

AndrewJakubowicz:
@justinfagnani [Workaround from the open nextjs issue](https://github.com/cookpete/react-player/issues/1455#issuecomment-1207154843) fixes the behavior, and explicit TypeScript types also work. See latest commit. It builds without issue, and `npm run dev` works great.

I'm not sure what we can do here on our libraries side. We could maybe create a nextjs "lazy" wrapper that just aliases `ref` under a different name? We could also document this workaround.

justinfagnani:
It would be nice to document this in the Next integration docs, when we get those. @augustjk where do you think those would go? Maybe an integrations page under SSR?

augustjk:
I think it would be good to include that under "Frameworks" section where we currently only have "React". Unsure if Next.js should be it's own section or something under React, but as a React meta-framework that we have a separate labs package for it could warrant its own section and we can discuss known issues like this and the "app router"/React Server Component support.

PR #4224, opened 2023-09-26
Graduate react

### LINKED ISSUE #4208: Graduate labs/react

rictic:
Fixes #4208


packages/labs/react/rollup.config.js
```diff
@@ -9,6 +9,6 @@ import {createRequire} from 'module';
 
 export default litProdConfig({
   packageName: createRequire(import.meta.url)('./package.json').name,
-  entryPoints: ['index', 'create-component', 'use-controller'],
+  entryPoints: ['index', 'use-controller'],
   includeNodeBuild: true,
```
augustjk, line 13: Just a thought, the node build for this re-exporting package will not look different at all. The same would be true for development builds. Would it be worth the trouble to get rid of that and update the package exports to only have the default and let the source package deal with the conditional exports?

augustjk (changes requested):
I think https://github.com/lit/lit/pull/4172 hasn't been brought over to 3.0 branch yet. It would be best to sync that before moving the source files.

rictic:
Merged main into 3.0 and rebased this PR on top!

packages/labs/react/rollup.config.js
rictic: There's likely a lot of simplification we could make here, but I don't know if it's worth the effort. I expect most users to migrate pretty quickly since it's trivial to do

augustjk (approved):
`gen-wrapper-react` tests are failing because `@lit/react` isn't published yet. Unsure if we need to really address that right now. It's not blocking the merge but we could skip the test for now or add a [package.json overrides](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#overrides) to make `@lit/react` resolve to the local copy in the monorepo or latest published labs version. Or ignore it as it should get fixed when everything is published.

Everything else looks good to me.

rictic:
Ah, the gen-wrapper-react failure was legit! Turns out we needed a number of additional changes in gen-wrapper-react

rictic:
Changes were big enough to want another review please!

augustjk (approved):
Aha, forgot we already had links in `gen-wrapper-react` test output, just needed it to point to the new location. Good finds for the rest of the spots to update from labs.

PR #4228, opened 2023-09-26
Some small, subtle fixes to 3.0 decorators

### LINKED ISSUE #4222: Fix issues with decorators discovered syncing into google3

rictic:
Based on fixing internal tests that failed.

Fixes #4222

### DIFF
--- a/packages/reactive-element/src/decorators/base.ts
+++ b/packages/reactive-element/src/decorators/base.ts
@@ -18,9 +18,35 @@ export type Constructor<T> = {
   new (...args: any[]): T;
 };
 
-export const descriptorDefaults = {
-  enumerable: true,
-  configurable: true,
-} as const;
+/**
+ * Wraps up a few best practices when returning a property descriptor from a
+ * decorator.
+ *
+ * Marks the defined property as configurable, and enumerable, and handles
+ * the case where we have a busted Reflect.decorate zombiefill (e.g. in Angular
+ * apps).
+ *
+ * @internal
+ */
+export const desc = (
+  obj: object,
+  name: PropertyKey | ClassAccessorDecoratorContext<unknown, unknown>,
+  descriptor: PropertyDescriptor
+) => {
+  // For backwards compatibility, we keep them configurable and enumerable.
+  descriptor.configurable = true;
+  descriptor.enumerable = true;
+  if (
+    // We check for Reflect.decorate each time, in case the zombiefill
+    // is applied via lazy loading some Angular code.
+    (Reflect as typeof Reflect & {decorate?: unknown}).decorate &&
+    typeof name !== 'object'
+  ) {
+    // If we're called as a legacy decorator, and Reflect.decorate is present
+    // then we have no guarantees that the returned descriptor will be
+    // defined on the class, so we must apply it directly ourselves.
 
-export const extendedReflect: typeof Reflect & {decorate?: unknown} = Reflect;
+    Object.defineProperty(obj, name, descriptor);
+  }
+  return descriptor;
+};
--- a/packages/reactive-element/src/decorators/base.ts
+++ b/packages/reactive-element/src/decorators/base.ts
@@ -17,3 +17,10 @@ export type Constructor<T> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
 };
+
+export const descriptorDefaults = {
+  enumerable: true,
+  configurable: true,
+} as const;
+
+export const extendedReflect: typeof Reflect & {decorate?: unknown} = Reflect;
--- a/packages/reactive-element/src/decorators/property.ts
+++ b/packages/reactive-element/src/decorators/property.ts
@@ -79,8 +79,6 @@ const legacyProperty = (
   // define in createProperty() with the original descriptor. We don't do this
   // for fields, which don't have a descriptor, because this could overwrite
   // descriptor defined by other decorators.
-  // for fields, which don't have a descriptor, because this could overwrite
-  // descriptor defined by other decorators.
   return hasOwnProperty
     ? Object.getOwnPropertyDescriptor(proto, name)
     : undefined;
--- a/packages/reactive-element/src/decorators/property.ts
+++ b/packages/reactive-element/src/decorators/property.ts
@@ -74,14 +74,6 @@ const legacyProperty = (
     name,
     hasOwnProperty ? {...options, wrapped: true} : options
   );
-  // For accessors (which have a descriptor on the prototype) we need to
-  // return a descriptor, otherwise TypeScript overwrites the descriptor we
-  // define in createProperty() with the original descriptor. We don't do this
-  // for fields, which don't have a descriptor, because this could overwrite
-  // descriptor defined by other decorators.
-  return hasOwnProperty
-    ? Object.getOwnPropertyDescriptor(proto, name)
-    : undefined;
 };
 
 // This is duplicated from a similar variable in reactive-element.ts, but
--- a/packages/reactive-element/src/decorators/property.ts
+++ b/packages/reactive-element/src/decorators/property.ts
@@ -74,6 +74,16 @@ const legacyProperty = (
     name,
     hasOwnProperty ? {...options, wrapped: true} : options
   );
+  // For accessors (which have a descriptor on the prototype) we need to
+  // return a descriptor, otherwise TypeScript overwrites the descriptor we
+  // define in createProperty() with the original descriptor. We don't do this
+  // for fields, which don't have a descriptor, because this could overwrite
+  // descriptor defined by other decorators.
+  // for fields, which don't have a descriptor, because this could overwrite
+  // descriptor defined by other decorators.
+  return hasOwnProperty
+    ? Object.getOwnPropertyDescriptor(proto, name)
+    : undefined;
 };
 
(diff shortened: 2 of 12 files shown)


packages/reactive-element/src/decorators/property.ts
```diff
@@ -79,6 +79,8 @@ const legacyProperty = (
   // define in createProperty() with the original descriptor. We don't do this
   // for fields, which don't have a descriptor, because this could overwrite
   // descriptor defined by other decorators.
+  // for fields, which don't have a descriptor, because this could overwrite
+  // descriptor defined by other decorators.
```
augustjk, line 83: these lines got duplicated somehow

rictic: So they did! Fixed

packages/reactive-element/src/decorators/state.ts
```diff
@@ -41,6 +41,7 @@ export type InternalPropertyDeclaration<Type = unknown> =
 export function state(options?: StateDeclaration) {
   return property({
     ...options,
+    state: true,
```
augustjk, line 44: Just for my own curiosity, does this actually have an effect beyond forcing `attribute: false` that's below this line?

packages/reactive-element/src/decorators/query-all.ts
```diff
@@ -59,13 +59,25 @@ let fragment: DocumentFragment;
...
       get(this: ReactiveElement) {
         const container =
           this.renderRoot ?? (fragment ??= document.createDocumentFragment());
         return container.querySelectorAll(selector);
       },
     };
+    if (typeof name !== 'object' && extendedReflect.decorate) {
+      // If we're called as a legacy decorator, and Reflect.decorate is present
+      // then we have no guarantees that the returned descriptor will be
+      // defined on the class, so we must apply it directly ourselves.
```
augustjk, line 77: I don't fully understand why this is the case (I have a vague memory of you talking about it), but I do appreciate the comment stating this behavior.

augustjk (approved):
Nice! I'm less familiar with the new decorator code but the changes here mostly makes sense to me.

packages/reactive-element/src/decorators/state.ts
rictic: Only for code that actually introspects on the property options, which it turns out some does!

packages/reactive-element/src/decorators/query-all.ts
rictic: TypeScript experimental decorators defer to the nonstandard Reflect.decorate if it's present, and some (all?) Angular apps install a very ancient version that ignores the returned value

packages/reactive-element/src/decorators/state.ts
AndrewJakubowicz: Worth adding an inline comment on why both are needed?

packages/reactive-element/src/decorators/query.ts
```diff
@@ -112,11 +114,24 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
...
       // experimental decorators.
-      return {
+      resultDescriptor = {
+        ...descriptorDefaults,
         get(this: ReactiveElement) {
           return doQuery(this);
         },
       };
     }
+    if (typeof nameOrContext !== 'object' && extendedReflect.decorate) {
```
AndrewJakubowicz, line 124: Could a unit test be added? Although I assume this is tested internally via TAP so can punt.

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -93,5 +97,13 @@ export function queryAssignedElements(
         ) as V;
       },
     };
+    if (typeof name !== 'object' && extendedReflect.decorate) {
+      // If we're called as a legacy decorator, and Reflect.decorate is present
+      // then we have no guarantees that the returned descriptor will be
+      // defined on the class, so we must apply it directly ourselves.
+      Object.defineProperty(obj, name, descriptor);
+      return;
+    }
```
AndrewJakubowicz, line 106: Is code size smaller to have this imported once for each decorator from `base.js`? Can ignore.

AndrewJakubowicz (approved):
Very subtle issue. Exceptional find and fix!

packages/reactive-element/src/decorators/base.ts
```diff
@@ -17,3 +17,10 @@ export type Constructor<T> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
 };
+
+export const descriptorDefaults = {
+  enumerable: true,
+  configurable: true,
+} as const;
+
+export const extendedReflect: typeof Reflect & {decorate?: unknown} = Reflect;
```
justinfagnani, line 26: Add comment - why can't you use Reflect directly?

packages/reactive-element/src/decorators/query-all.ts
```diff
@@ -59,13 +59,25 @@ let fragment: DocumentFragment;
  * @category Decorator
  */
 export function queryAll(selector: string): QueryAllDecorator {
-  return (() => {
-    return {
+  return ((
+    obj: object,
+    name: PropertyKey | ClassAccessorDecoratorContext<unknown, unknown>
```
justinfagnani, line 64: I think this:

```suggestion
    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<unknown, unknown>
```

will make the `typeof name !== 'object'` below a lot more clear.

packages/reactive-element/src/decorators/base.ts
justinfagnani, line 26: I only see this accessed as `extendedReflect.decorate` coerced to a boolean, so could you just export that boolean?

```suggestion
export const reflectDecorateExists = 'decorate' in Reflect;
```

packages/reactive-element/src/decorators/base.ts
```diff
@@ -17,3 +17,10 @@ export type Constructor<T> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
 };
+
+export const descriptorDefaults = {
```
justinfagnani, line 21: I don't really want to commit to these exports being in the public API. Can you mark `@internal`? And can these be renamed by Terser?

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -93,5 +97,13 @@ export function queryAssignedElements(
         ) as V;
       },
     };
+    if (typeof name !== 'object' && extendedReflect.decorate) {
+      // If we're called as a legacy decorator, and Reflect.decorate is present
+      // then we have no guarantees that the returned descriptor will be
+      // defined on the class, so we must apply it directly ourselves.
+      Object.defineProperty(obj, name, descriptor);
+      return;
```
justinfagnani, line 105: is it bad to let this fall through and return the descriptor?

packages/reactive-element/src/decorators/query-assigned-elements.ts
```diff
@@ -93,5 +97,13 @@ export function queryAssignedElements(
         ) as V;
       },
     };
+    if (typeof name !== 'object' && extendedReflect.decorate) {
```
justinfagnani, line 100: I think you could even wrap this into a function exported from base, like:

base.ts:
```ts
/**
  * Ensures that the property is defined even if we're called as a legacy decorator, and
  *  Reflect.decorate is present. In that case we have no guarantees that a descriptor 
  * returned from a decorator will be
  * defined on the class, so we must apply it directly ourselves.
  * Also applies defaults...
  */
export const ensureProperty = (obj, nameOrContext, descriptor) => {
  descriptor = {
      enumerable: true,
      configurable: true,
      ...descriptor,
  };
  if (typeof name !== 'object' && extendedReflect.decorate) {
      Object.defineProperty(obj, name, descriptor);
      return;
  }
  return descriptor;
}
```

Here:
```ts
return ensureProperty(obj, name, descriptor);
```

packages/reactive-element/src/decorators/base.ts
rictic: Marked as @internal

Looks like it's not renamed, Terser emits:

```js
const e=(e,t,c)=>(c.configurable=!0,c.enumerable=!0,Reflect.decorate&&"object"!=typeof t&&Object.defineProperty(e,t,c),c);export{e as desc};
```

Gave it a short name since it's not minified, and marked it as internal

packages/reactive-element/src/decorators/query-assigned-elements.ts
rictic: Done. Gave it a short name for size and because it has no easily explicable name, and just mutated the descriptor rather than allocating a new one, since we know all the cases where this is called and they're all ok with mutation

packages/reactive-element/src/decorators/query-assigned-elements.ts
rictic: Let's try it, might have to revert it internally, but we'll see

packages/reactive-element/src/decorators/query-all.ts
rictic: Less necessary now that we're just passing this along

packages/reactive-element/src/decorators/base.ts
rictic: Done

PR #4231, opened 2023-09-27
Update chromedriver.

rictic:


.github/workflows/benchmarks.yml
```diff
@@ -39,7 +39,9 @@ jobs:
           comment_tag: check-size # ensures we create or update one size comment
 
       - name: Build
-        run: npm run build
+        run: |
+          cd packages/benchmarks
+          npm run build
```
AndrewJakubowicz, line 44: Is this change expected when updating chromedriver?

rictic: No, I added this in because I was already touching the benchmark code. Currently the build step takes a couple minutes, and this should drop that down a good bit

AndrewJakubowicz (approved):
Awesome!

PR #4251, opened 2023-09-30
A few minor things for google internal support.

rictic:


AndrewJakubowicz (approved):

AndrewJakubowicz:
Is it new that occasionally GitHub Actions fails with a lot of type errors. I think we saw this yesterday and theorized it was due to a failure in the TypeScript types being built?

PR #4252, opened 2023-09-30
fix (labs/analyzer): use analyzer typescript instead of import

43081j:
There are still some places where we use methods available on our imported `ts` rather than from the analyzer's reference (`analyzer.typescript`). This causes various problems with version mismatches and bundling.

These have now been updated to use the analyzer's typescript.

### Related question

While doing this change, I noticed we generally follow two common signatures for our functions:

- `(node, analyzer)`
- `(ts, node)`

For the latter, it does make a sense from a purist point of view... type guards and what not which operate purely on a given AST node don't need the analyzer in theory, so they accept a typescript lib ref instead.

But it does mean we switch back and forth inconsistently sometimes between these signatures. maybe we want to _always_ pass an analyzer and just pull the typescript from that?

### DIFF
--- a/.changeset/hot-apricots-type.md
+++ b/.changeset/hot-apricots-type.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/analyzer': patch
+---
+
+Always use consumer's typescript rather than analyzer's dependency to avoid version mismatches
--- a/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
+++ b/packages/labs/analyzer/src/lib/custom-elements/custom-elements.ts
@@ -11,7 +11,7 @@
  * subclasses.
  */
 
-import ts from 'typescript';
+import type ts from 'typescript';
 import {getClassMembers, getHeritage} from '../javascript/classes.js';
 import {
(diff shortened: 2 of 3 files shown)


justinfagnani (approved):
Thanks!

justinfagnani:
Regarding the API shape question, I noted when I made the change that it's a bit inconsistent: https://github.com/lit/lit/pull/4029

I think I'd like to see the types narrowed to just what's necessary. There a couple of issues to consider though:
- It may be a bit cumbersome to pass say a typescript lib, checker and/or program, which is largely what an Analyzer object holds. So sometimes it's just nicer to pass an Analyzer.
- If functions are too narrow and then need a new argument in the future (say to call another utility that needs it), it'll be a breaking change to widen.

That said, I think it'd be good to clean up now. I'm starting to add template utilities that are useful without full program analysis, like in https://github.com/lit/lit/pull/4261 and made those just take a typescript and checker (though maybe that should be a Program?).

Another thing is the order of operations. I think the Analyzer or TypeScript lib objects should probably come first in the parameter list.

PR #4254, opened 2023-10-03
Change the type of renderRoot to include Element

justinfagnani:
I found a type mismatch between `renderRoot` and `createRenderRoot` when removing some unsafe casts.


justinfagnani:
Oh... `Element | ShadowRoot` doesn't play nice with lit-html's `render()` which accepts a `HTMLElement | DocumentFragment`.

I'll change `createRenderRoot()` instead to return a `HTMLElement | DocumentFragment`.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1083,7 +1092,7 @@ export abstract class ReactiveElement
    * @return Returns a node into which to render.
    * @category rendering
    */
-  protected createRenderRoot(): Element | ShadowRoot {
+  protected createRenderRoot(): HTMLElement | DocumentFragment {
```
AndrewJakubowicz, line 1095: Why change `ShadowRoot` to `DocumentFragment`, as `ShadowRoot` seems to be the more specific return from `this.attachShadow`?

justinfagnani: Just because lit-html's render can render to a DocumentFragment. It seems best to align all these types.

AndrewJakubowicz (approved):

AndrewJakubowicz:
@rictic We'll need to test this change in google3

rictic (approved):

PR #4255, opened 2023-10-03
[infra] sync 3.0 package-lock after second pre-release

AndrewJakubowicz:
After second pre-release, need to sync package-lock as changeset doesn't do it for us.

Note: prior to this change, Node 20 install fails. I used Node 18 to update package-lock. Then verified that it works with Node 20.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,16 +58,16 @@
     },
     "examples/nextjs": {
       "name": "@lit-examples/nextjs",
-      "version": "0.1.0",
+      "version": "0.1.1-pre.0",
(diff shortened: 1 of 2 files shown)


augustjk (approved):
Thanks!

AndrewJakubowicz:
> **This is a size regression**. It adds **22** bytes to lit-core.min.js which is 15,403B at head. We're very sensitive to size increases. If this is intended, please update `expectedLitCoreSize` to 15425 in `scripts/check-size.js` and push a new commit to your PR.

WAT

PR #4257, opened 2023-10-04
elementProperties is a static property that's dynamic similarly to finalized

rictic:
To get correct semantics out of it, we need to hide it from closure's instinct to collapse it by using dynamic property access syntax.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -380,6 +380,7 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
  * this hack to bypass any rewriting by the compiler.
  */
 const finalized = 'finalized';
+const elementProperties = 'elementProperties';
 
 /**
  * A string representing one of the supported dev mode warning categories.
@@ -553,7 +554,7 @@ export abstract class ReactiveElement
    * @nocollapse
    * @category properties
    */
-  static elementProperties: PropertyDeclarationMap = new Map();
+  static [elementProperties]: PropertyDeclarationMap = new Map();
 
   /**
    * User-supplied object that maps property names to `PropertyDeclaration`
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz (approved):
Thank you!

Also... a 145 byte improvement! That's awesome! 🎉

packages/reactive-element/src/reactive-element.ts
```diff
@@ -553,7 +554,7 @@ export abstract class ReactiveElement
    * @nocollapse
    * @category properties
    */
-  static elementProperties: PropertyDeclarationMap = new Map();
+  static [elementProperties]: PropertyDeclarationMap = new Map();
```
justinfagnani, line 557: Why is this suddenly needed, when this property has existed since the first lit-element version? There are other static properties that don't have this need... what's different about them?

justinfagnani:
> A 145 byte improvement! That's awesome! 🎉

Computed property access is slower though. I think ideally we want to keep property access in the update path as simple as possible.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -791,9 +792,7 @@ export abstract class ReactiveElement
    * @nocollapse
    */
   private static __prepare() {
-    if (
-      this.hasOwnProperty(JSCompiler_renameProperty('elementProperties', this))
-    ) {
+    if (this.hasOwnProperty(elementProperties)) {
```
justinfagnani, line 795: I don't understand why whatever issue is effecting `elementProperties` isn't also effecting if `this.hasOwnProperty(JSCompiler_renameProperty('properties', this)` on line 832. They're both static properties access with `hasOwnProperty()` and both were passed through `JSCompiler_renameProperty` before this change.

The previous change that's failing google3 seems to only be significant in using `hasOwnProperty('elementProperties')`, is that what's triggering https://github.com/google/closure-compiler/issues/3177 ?

rictic:
Closure compiler will convert the dynamic property access to a dotted property access, terser won't. We can just handle this with a copybara change.

packages/reactive-element/src/reactive-element.ts
rictic: Tracked it down and got a better solution in [b3daf46](https://github.com/lit/lit/pull/4257/commits/b3daf46ff6b038d46cb70ec8b5a5166a1d93ec2f) that also works for `finalized`

justinfagnani (approved):

Issue #4237, opened 2023-09-28
[reactive-element] Avoid `@query` caching `null` results before the first render


MaxArt2501:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

If you access `@query`d property before the first render, if will yield `null` but will stay `null` is you set the cache flag to `true`. The documentation should probably warn the users that this might happen (might open a separate issue on lit.dev repo).

I think it could be beneficial if `@query` caches the result only if it actually has a chance to return a non-null result, i.e. only if `hasUpdated` is `true`. Something like this:

```ts
descriptor.get = function (this: ReactiveElement) {
  // No caching here
  if (!this.hasUpdated) return null;
  // Rest of the getter code...
};
```

I'm available to open a PR for this.

### Alternatives and Workarounds

No alternative that would be as simple as using `@query` straight away.

One could redefine their own `@query` selector, but of course it'd need to be kept in line with Lit's decorator.

Alternatively, one could check if `hasUpdated` is `true` if they need to access the queried property. A secondary getter could be handy, but adds overhead:

```ts
class MyElement extends LitElement {
  @query('button', true) private _button!: HTMLButtonElement | null;

  get button() {
    return this.hasUpdated ? this._button : null;
  }
  // ...
}
```

justinfagnani:
I _think_ this sounds like a good idea. I wonder in what cases you're accessing the query field before first render though?

justinfagnani:
@rictic thoughts?

rictic:
I feel like I've seen code kinda like:

```
@query('button', true) button: HTMLButtonElement|undefined;

clickButton() {
  this.button?.click();
}
```

Is that the best? No. Should it cache `null`? That does seem worse

rictic:
It'd add few bytes, little overhead, and is otherwise a solid improvement, so I'm in favor.

Even better with a dev mode warning for accessing a query before first updated.

MaxArt2501:
@rictic That's exactly the main use case that came to my mind.

The other case is when queried properties are accessed in `willUpdate`.

PR #4259, opened 2023-10-04
Make @query not cache null results before first render ( #4237)

MaxArt2501:
This change makes the `@query` decorator not cache the result if it's `null` and before the first render.
The result might be not null in case of a Declarative Shadow DOM.

Might follow a PR for lit.dev repo to state the new behavior.

### DIFF
--- a/packages/reactive-element/src/decorators/query.ts
+++ b/packages/reactive-element/src/decorators/query.ts
@@ -98,15 +98,14 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
             })();
       return desc(protoOrTarget, nameOrContext, {
         get(this: ReactiveElement): V {
-          if (cache) {
-            let result: V = get!.call(this);
-            if (result === undefined) {
-              result = doQuery(this);
+          let result: V = get!.call(this);
+          if (result === undefined) {
+            result = doQuery(this);
+            if (result || this.hasUpdated) {
               set!.call(this, result);
             }
-            return result;
           }
-          return doQuery(this);
+          return result;
         },
       });
     } else {
--- a/packages/reactive-element/src/test/decorators-modern/query_test.ts
+++ b/packages/reactive-element/src/test/decorators-modern/query_test.ts
@@ -95,11 +95,11 @@ import {assert} from '@esm-bundle/chai';
     assert.notEqual(el.span, el.renderRoot.querySelector('span'));
   });
 
-  test('returns cached value when accessed before first update', async () => {
+  test('does not cache null values when accessed before first update', async () => {
     const notYetUpdatedEl = new C();
     assert.equal(notYetUpdatedEl.divCached, null);
     container.appendChild(notYetUpdatedEl);
     await notYetUpdatedEl.updateComplete;
-    assert.equal(notYetUpdatedEl.divCached, null);
+    assert.instanceOf(notYetUpdatedEl.divCached, HTMLDivElement);
   });
 });
--- a/packages/reactive-element/src/test/decorators/query_test.ts
+++ b/packages/reactive-element/src/test/decorators/query_test.ts
@@ -90,12 +90,12 @@ import {assert} from '@esm-bundle/chai';
     assert.notEqual(el.span, el.renderRoot.querySelector('span'));
   });
 
-  test('returns cached value when accessed before first update', async () => {
+  test('does not cache null values when accessed before first update', async () => {
     const notYetUpdatedEl = new C();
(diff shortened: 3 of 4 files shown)


packages/reactive-element/src/decorators/query.ts
```diff
@@ -98,15 +98,14 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
             })();
       return desc(protoOrTarget, nameOrContext, {
         get(this: ReactiveElement): V {
-          if (cache) {
-            let result: V = get!.call(this);
-            if (result === undefined) {
-              result = doQuery(this);
+          let result: V = get!.call(this);
```
MaxArt2501, line 101: I've removed the `if (cache)` statement as it should always be true at this point. Possibly a copy+paste mistake?

justinfagnani: good catch!

packages/reactive-element/src/decorators/query.ts
```diff
@@ -98,15 +98,14 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
...
+          if (result === undefined) {
+            result = doQuery(this);
+            if (result || this.hasUpdated) {
```
justinfagnani, line 104: We try to always do an explicit comparison to avoid ToBoolean conversions:

```suggestion
            if (result !== null || this.hasUpdated) {
```

justinfagnani (commented):
A couple of quick comments, though I want to take discussion of the change in behavior to https://github.com/lit/lit/issues/4237

packages/reactive-element/src/decorators/query.ts
```diff
@@ -98,15 +98,14 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
             })();
```
rictic, line 98: Consider add in our DEV_MODE only warning issuer code to this file:

```ts
let issueWarning: undefined | (code: string, warning: string) => void;
if (DEV_MODE) {
  // Ensure warnings are issued only 1x, even if multiple versions of Lit
  // are loaded.
  const issuedWarnings: Set<string | undefined> = (global.litIssuedWarnings ??=
    new Set());

  // Issue a warning, if we haven't already.
    if (!issuedWarnings.has(warning)) {
      console.warn(warning);
      issuedWarnings.add(warning);
    }
  };
}
```

And issuing a warning in doQuery if DEV_MODE is true and !el.hasUpdated, letting the user know that their query field is being accessed before the element has updated.

MaxArt2501: Should the warning be for *any* access before the first update, whether the result is `null` or not, and whether the query is cached or not?
Should I create a `code` of my choice or there's a guideline for this? I was thinking of `'early-cached-query-access'`.

MaxArt2501: Added the warning with a tentative code and message. Let me know if it's ok.

justinfagnani:
One note: we need to update the output of the ts-transformers package so that the behavior of the emitted query getter matches this change.

AndrewJakubowicz:
> One note: we need to update the output of the ts-transformers package so that the behavior of the emitted query getter matches this change.

I just verified that this is not needed.
The idiomatic decorator transform already has this behavior.

The transform does the following:

```
@query('#mySpan', true)
span?: HTMLSpanElement

// after idiomatic transform
get span() {
  return this.__span ??= this.renderRoot?.querySelector('#mySpan') ?? null;
}
```

Because of the nullish coalescing operator, the idiomatic decorator transform won't cache the `null` value.

MaxArt2501:
I'm not sure I need to do something to make this PR approved. I see unsuccessful checks - is theere something I should/can do?
Will it make it in time for this Tuesday's release?

rictic:
Oh that's messed up, deleting the 3.0 branch closed this PR? In the past in similar cases it just changed the target branch of the PR

AndrewJakubowicz:
Sorry that this PR closed. We're unable to reopen it on our end (probably because it's from a fork).

Would you be able to re-create this change, and base it off `main` branch (ideally using the email with the CLA)?

We had a discussion and decided this isn't blocking 3.0, and can be added as a minor patch to 3.0.

The rationale is that, in the rare case that this breaks someone, it probably should already be an error. The error case would hopefully also be noticed, so a cached `null` value isn't depended on.

MaxArt2501:
Sure, no problem! I'll do that as soon as possible.

Thank you

PR #4261, opened 2023-10-04
[labs/analyzer] Add template utils to analyzer and use them in the compiler

justinfagnani:
Builds on https://github.com/lit/lit/pull/4260 (for use in tests)

This moves the `isLitTaggedTemplateExpression` template utility from the compiler to the analyzer. The analyzer will need these same utilities to detect lit-html template expressions for template analysis.

### DIFF
--- a/packages/labs/analyzer/src/test/server/lit-html/template_test.ts
+++ b/packages/labs/analyzer/src/test/server/lit-html/template_test.ts
@@ -26,7 +26,6 @@ for (const lang of languages) {
 
   test('isLitHtmlTemplateTag', ({getModule, analyzer, typescript}) => {
     const elementAModule = getModule('element-a')!;
-    console.log('elementAModule.jsPath', elementAModule.sourcePath);
     const decl = elementAModule.declarations[0];
 
     // get to the lit-html template tag
@@ -51,8 +50,6 @@ for (const lang of languages) {
       ),
       true
     );
-
-    // console.log('template', returnStatement.expression);
   });
 
   test.run();
--- a/packages/labs/analyzer/src/test/server/lit-html/template_test.ts
+++ b/packages/labs/analyzer/src/test/server/lit-html/template_test.ts
@@ -1,0 +1,59 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {suite} from 'uvu';
+// eslint-disable-next-line import/extensions
+import * as assert from 'uvu/assert';
+import type ts from 'typescript';
+
+import {
+  AnalyzerTestContext,
+  languages,
+  setupAnalyzerForTest,
+} from '../utils.js';
+import {ClassDeclaration} from '../../../lib/model.js';
+import {isLitTaggedTemplateExpression} from '../../../lib/lit-html/template.js';
+
+for (const lang of languages) {
+  const test = suite<AnalyzerTestContext>(`lit-html tests (${lang})`);
+
+  test.before((ctx) => {
+    setupAnalyzerForTest(ctx, lang, 'basic-elements');
+  });
+
+  test('isLitHtmlTemplateTag', ({getModule, analyzer, typescript}) => {
+    const elementAModule = getModule('element-a')!;
+    console.log('elementAModule.jsPath', elementAModule.sourcePath);
+    const decl = elementAModule.declarations[0];
+
+    // get to the lit-html template tag
+    const renderMethod = (decl as ClassDeclaration).getMethod('render')!;
+    const statement = renderMethod.node.body!.statements[0];
+
+    assert.is(typescript.isReturnStatement(statement), true);
+    const returnStatement = statement as ts.ReturnStatement;
+    assert.ok(returnStatement.expression);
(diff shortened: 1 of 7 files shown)


packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -337,3 +339,19 @@ export const getSuperClass = (
   );
   return undefined;
 };
+
+export const isConstructorFieldInitializer = (
```
rictic, line 343: Slightly unfortunate that this doesn't require that the expression is a direct child of a constructor

Could we instead write this like:

```typescript
export const getConstructorFieldInitializers = *(
  member: ts.ConstructorDeclaration>, 
  typescript: typeof ts
): IterableIterator<ConstructorFieldInitializer> => { ... };
```

or the forEach style equivalent

packages/labs/analyzer/src/lib/javascript/classes.ts
```diff
@@ -119,48 +122,45 @@ export const getClassMembers = (
...
-          node.expression.operatorToken.kind ===
-            typescript.SyntaxKind.EqualsToken &&
-          typescript.isPropertyAccessExpression(node.expression.left) &&
-          node.expression.left.expression.kind ===
-            typescript.SyntaxKind.ThisKeyword
+          typescript.isExpressionStatement(statement) &&
+          isConstructorFieldInitializer(statement.expression, typescript)
         ) {
-          const name = node.expression.left.name.getText();
+          const name = statement.expression.left.name.getText();
```
rictic, line 130: here and elsewhere: getText() is almost always a mistake to call outside of debug statements, because it exists on every node but doesn't have a reasonable value for a lot of them (and when called on artificial nodes that weren't in the original source code it may throw an error)

packages/labs/analyzer/src/lib/model.ts
```diff
@@ -316,29 +327,32 @@ export interface VariableDeclarationInit extends DeclarationInit {
 }
 
 export class VariableDeclaration extends Declaration {
-  readonly node:
+  readonly type: Type | undefined;
+  declare readonly node:
```
rictic, line 331: instead, what if we gave Declaration a type parameter with a default value of ts.Node and the subclasses like VariableDeclaration could set it when extending?

packages/labs/analyzer/src/lib/javascript/classes.ts
justinfagnani: Sorry, I forgot to set the correct base branch for this PR. This change is actually in https://github.com/lit/lit/pull/4260 and even there it's only a refactor of existing logic into a function call for better readability. The function is already called only against direct child statements of the constructor body in practice, within a loop against those body children.

Another refactor that could accomplish your goal would be to pass the constructor body node in to verify that it's the parent, however I'm not 100% sure that that check is correct. I want to verify against TypeScript's field inference because things like `(this.foo = 1)` might count as a field even though they're in a nested expression.

packages/labs/analyzer/src/lib/javascript/classes.ts
justinfagnani: This is already used in a bunch of places. I can file an issue to change them.

rictic: Yeah, we should honestly add a lint rule to catch and prevent calls to getText(). We did that on the internal version of grimlock and it prevented a bunch of issues.

Andrew actually ran into an issue _today_ where .getText() did the wrong thing (when called on a SourceFile it doesn't include all the content of the file (!), it misses leading comments. SourceFile has a .text property with the full text of the file)

packages/labs/analyzer/src/test/server/lit-html/template_test.ts
```diff
@@ -0,0 +1,59 @@
...
+for (const lang of languages) {
+  const test = suite<AnalyzerTestContext>(`lit-html tests (${lang})`);
+
+  test.before((ctx) => {
+    setupAnalyzerForTest(ctx, lang, 'basic-elements');
+  });
+
+  test('isLitHtmlTemplateTag', ({getModule, analyzer, typescript}) => {
+    const elementAModule = getModule('element-a')!;
+    console.log('elementAModule.jsPath', elementAModule.sourcePath);
```
rictic, line 29: console.log

packages/labs/analyzer/src/test/server/lit-html/template_test.ts
```diff
@@ -0,0 +1,59 @@
...
+    assert.is(
+      isLitTaggedTemplateExpression(
+        expression,
+        analyzer.typescript,
+        analyzer.program.getTypeChecker()
+      ),
+      true
+    );
+
+    // console.log('template', returnStatement.expression);
```
rictic, line 55: delete here too

rictic (approved):
LGTM, but for reals, every .getText() is a bug waiting to happen

packages/labs/analyzer/src/lib/javascript/classes.ts
AndrewJakubowicz: image

43081j (approved):

PR #4263, opened 2023-10-06
Add missing deps and other minor fixups to get CI green

rictic:


packages/labs/gen-manifest/package.json
```diff
@@ -19,7 +19,8 @@
       "dependencies": [
         "../../tests:build",
         "../analyzer:build",
-        "../gen-utils:build"
+        "../gen-utils:build",
+        "../../lit-element:build"
```
justinfagnani, line 23: I don't see that anything in this package imports lit-element... why was this needed?

packages/labs/gen-wrapper-react/src/test-gen/generate_test.ts
```diff
@@ -64,7 +64,7 @@ test('basic wrapper generation', async () => {
   // Pack the generated package here, as `test-output` package.json will
   // reference the generated tarball here by filename; `test-output:installSelf`
   // depends on these tests run by `test:gen`.
-  await packPackage(outputPackage);
+  console.log(await packPackage(outputPackage));
```
justinfagnani, line 67: remove?

package.json
```diff
@@ -282,7 +282,7 @@
     "rollup-plugin-terser": "^7.0.2",
     "typescript": "~5.2.0",
     "uvu": "^0.5.6",
-    "wireit": "^0.14.0"
+    "wireit": "^0.9.5"
```
AndrewJakubowicz, line 285: Why downgrade wireit?

packages/labs/gen-wrapper-react/src/test-gen/generate_test.ts
rictic: +1, done

AndrewJakubowicz (approved):

augustjk (approved):

packages/labs/gen-manifest/package.json
rictic: We're building a manifest with types for an example that imports lit, so we need the types for lit

packages/labs/gen-wrapper-react/src/test-gen/generate_test.ts
rictic: Ah, yes, done

package.json
rictic: Was wondering if that would fix, it did not

PR #4267, opened 2023-10-06
[labs/analyzer] Add template parser to analyzer

justinfagnani:
Builds on #4261 

Adds a `parseLitTemplate()` function to `lib/lit-html/template.ts`. This returns an extension of a parse5 DocumentFragment that contains parts and strings, and where some nodes like comments and attributes are annotated with Lit parts. The parts have references to their expressions to allow easy traversal into nested expressions and templates.

I didn't yet port over the similar code in SSR or the compiler. These might have some different requirements (like parallel traversal or re-writing comment markers). Other future extensions could be detecting common invalid binding locations for linting.


packages/labs/analyzer/src/lib/lit-html/template.ts
```diff
@@ -11,8 +11,24 @@
  */
 
 import type ts from 'typescript';
+import {_$LH} from 'lit-html/private-ssr-support.js';
+import {parseFragment} from 'parse5';
+// import type {Attribute} from 'parse5/dist/common/token.js';
```
AndrewJakubowicz, line 16: Intended to leave in?

AndrewJakubowicz (commented):
Very quick initial look.
Is it worth finding ways to reduce the number of type assertions?

packages/labs/analyzer/src/lib/lit-html/template.ts
```diff
@@ -163,3 +179,201 @@ const isResolvedIdentifierLitHtmlTemplate = (
...
+export const parseLitTemplate = (
+  node: ts.TaggedTemplateExpression,
+  ts: TypeScript,
+  _checker: ts.TypeChecker
+): LitTemplate => {
+  const strings = getTemplateStrings(node, ts);
+  const values = ts.isNoSubstitutionTemplateLiteral(node.template)
+    ? []
+    : node.template.templateSpans.map((s) => s.expression);
+  const parts: Array<PartInfo> = [];
```
kevinpschaaf, line 259: It might be good to clarify the use cases for the API's here. i.e. whether we intend the DOM AST that comes out to be mutable or not, since the `parts` list will only be the "as-parsed" values.

This could either be made a getter that reifies the parts list each access (slower), or could add an "invalidateParts" API?

And are you planning on adding something to go from a `LitTemplate` back to a `ts.TaggedTemplateExpression`? (that'll presumably need ^^)

kevinpschaaf: It's obv fine for some use cases, where you just want to statically analyze only, but for others the `parts` sort of seems not that useful.

packages/labs/analyzer/src/lib/lit-html/template.ts
```diff
@@ -11,8 +11,24 @@
...
+  isCommentNode,
+  isElementNode,
+  traverse,
+} from '@parse5/tools';
+
+const {getTemplateHtml, marker, markerMatch, boundAttributeSuffix} = _$LH;
 
 type TypeScript = typeof ts;
+// TODO (justinfagnani): use the real type from parse5?
+type Attribute = Element['attrs'][number];
```
43081j, line 31: you can probably import `Token` from parse5 and use `Token.Attribute`

packages/labs/analyzer/src/lib/lit-html/template.ts
43081j: fwiw in the lint side of things, we don't need to mutate nodes but we do need a way to translate between nodes or positions. i.e. we need a way to go for a parse5 node to an estree node (whether by reference or just by knowing the position)

packages/labs/analyzer/src/lib/lit-html/template.ts
```diff
@@ -10,14 +10,42 @@
...
+  type DocumentFragment,
+  type Element,
+  type Node,
+  type TextNode,
+} from '@parse5/tools';
+
+const {getTemplateHtml, marker, markerMatch, boundAttributeSuffix} = _$LH;
+
 type TypeScript = typeof ts;
+export type Attribute = Element['attrs'][number];
```
43081j, line 43: I think you can import `Token` from parse5 and use `Token.Attribute`

.changeset/fuzzy-lemons-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/analyzer': minor
+---
+
+Add a template parser
```
rictic, line 5: Duplicate changeset

packages/labs/analyzer/src/lib/lit-html/template.ts
```diff
@@ -163,3 +179,201 @@ const isResolvedIdentifierLitHtmlTemplate = (
...
+}
+
+export interface LitTemplateAttribute extends Attribute {
+  litPart: PartInfo;
+}
+
+export const parseLitTemplate = (
+  node: ts.TaggedTemplateExpression,
+  ts: TypeScript,
+  _checker: ts.TypeChecker
```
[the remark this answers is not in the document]
justinfagnani: I have it here for consistency with other functions that need the checker, because it's kind of unusual that this function _doesn't_ need the checker. In case we need to use the checker in here in the future, it won't be a breaking change to public consumers.

packages/labs/analyzer/src/lib/lit/template.ts
```diff
@@ -0,0 +1,561 @@
...
+  const nodeMarkerLength = nodeMarker.length;
+
+  // TODO (justinfagnani): to support server-only templates that include
+  // non-fragment-parser supported tags (<html>, <body>, etc) we need to
+  // inspect the string and conditionally use parse() here.
+  const ast = parseFragment(html.toString(), {
+    sourceCodeLocationInfo: true,
+  });
+
+  traverse(ast, {
```
e111077, line 349: Why not do some pre-processing of the template strings before traversing. e.g.

- Create a map of original-positions to processed-positions and vice versa
- iterate through the strings char by char
- store index of EVERY char in both maps
- When we get to the end of the string and to an expression replace the expression with a comment
  - original position is += the full text length of the expression
  - processed position is the length of your marker text
 
Now, if you have a map of all positions mapped from one template to another, when you traverse the parse5 tree and visit a node, it doesn't matter _much_ what the node is, as long as it has a sourceCodeLocation. The only times you'd have to care is if it's an element. In this case you'd have to also update `el.sourceCodeLocation.{attrs,startTag,endTag}`

The negatives here are that you are making a map that is memory size of O(n) where n is the max length of the template literal but it should be a map of `Map<number,number>`, so how bad can it be???

packages/labs/analyzer/src/lib/lit/template.ts
```diff
@@ -0,0 +1,584 @@
...
+  isElementNode,
+  isTextNode,
+} from '@parse5/tools';
+
+const {getTemplateHtml, marker, markerMatch, boundAttributeSuffix} = _$LH;
+
+type TypeScript = typeof ts;
+
+// Why, oh why are parse5 types so weird? We re-export them to make them easier
+// to use.
```
43081j, line 40: you should be able to pull these from `@parse5/tools` itself now

i think it re-exports the defaults

packages/labs/analyzer/src/lib/lit/decorators.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Utilities for analyzing ReactiveElement decorators.
+ */
+
+import type ts from 'typescript';
+
+export type TypeScript = typeof ts;
+
+/**
+ * Returns true if the decorator site is a simple called decorator factory of
+ * the form `@decoratorName()`.
```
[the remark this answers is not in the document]
justinfagnani: This was a file move that Git or GitHub didn't pick up. I'd rather keep this PR limited to the template parser.

packages/labs/analyzer/src/lib/lit/decorators.ts
```diff
@@ -0,0 +1,82 @@
...
+): decorator is PropertyDecorator =>
+  isNamedDecoratorFactory(ts, decorator, 'property');
+
+/**
+ * A narrower type for ts.Decorator that represents the shape of an analyzable
+ * `@customElement('x')` callsite.
+ */
+interface PropertyDecorator extends ts.Decorator {
+  readonly expression: ts.CallExpression;
+}
```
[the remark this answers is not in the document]
justinfagnani: Same as the other comment about this being a file move.

packages/labs/analyzer/src/lib/lit/template.ts
```diff
@@ -0,0 +1,561 @@
...
+  const strings = getTemplateStrings(node, ts);
+  const values = ts.isNoSubstitutionTemplateLiteral(node.template)
+    ? []
+    : node.template.templateSpans.map((s) => s.expression);
+  const templateSpans = ts.isNoSubstitutionTemplateLiteral(node.template)
+    ? []
+    : node.template.templateSpans;
+
+  const parts: Array<PartInfo> = [];
+  const [html, boundAttributeNames] = getTemplateHtml(strings, 1);
```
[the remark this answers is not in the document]
justinfagnani: Thanks! This was the plan from a long time ago :) I hope it hold up actually.

packages/labs/analyzer/src/lib/lit/template.ts
```diff
@@ -0,0 +1,561 @@
...
+
+  // Depth-first node index
+  let nodeIndex = 0;
+
+  // Adjustments to source locations based on the difference between the
+  // binding expression lengths in TypeScript source vs the marker replacements
+  // in the prepared and parsed HTML.
+
+  // TODO (justinfagnani): implement line and column adjustments
+  let lineAdjust = 0;
```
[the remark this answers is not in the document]
justinfagnani: parse5 includes lines, so I didn't want to leave them unadjusted and incorrect.

packages/labs/analyzer/src/lib/lit/template.ts
justinfagnani: I don't fully follow this approach. Can you explain it to me on a call?

packages/labs/analyzer/src/lib/lit/template.ts
justinfagnani: I don't see where to get all of these types from.

packages/labs/analyzer/package.json
```diff
@@ -67,7 +67,7 @@
     },
     "test:server:uvu": {
       "#comment": "The quotes around the file regex must be double quotes on windows!",
-      "command": "uvu test/server \"_test\\.js$\" -i \"lit-html/\"",
+      "command": "uvu test/server \"_test\\.js$\" -i \"lit/template_test\"",
```
rictic, line 70: What's with these changes?

justinfagnani: Just getting the uvu tests run by uvu and the node tests to be run run by node. I want to convert everything to node soon.

PR #4276, opened 2023-10-09
Document value and error fields on Task.

rictic:


justinfagnani (approved):

augustjk (approved):
Confirming no changeset so this'll just go in undocumented with the stable release.

PR #4282, opened 2023-10-10
Make @query not cache null results before first render (#4237)

MaxArt2501:
Do-over of #4259 after being closed when branch 3.0 has been closed. This was the description:

This change makes the @query decorator not cache the result if it's null and before the first render.
The result might be not null in case of a Declarative Shadow DOM.


AndrewJakubowicz:
Sorry for the slow response here. Thank you for re-creating this PR. I'm manually importing and testing this PR against our internal test suite to see if there is any test failures that need to be handled.

Will report back tomorrow with results.

packages/reactive-element/src/test/decorators-modern/query_test.ts
```diff
@@ -96,11 +96,11 @@ import {assert} from '@esm-bundle/chai';
     assert.notEqual(el.span, el.renderRoot.querySelector('span'));
   });
 
-  test('returns cached value when accessed before first update', async () => {
+  test('does not cache null values when accessed before first update', async () => {
     const notYetUpdatedEl = new C();
     assert.equal(notYetUpdatedEl.divCached, null);
     container.appendChild(notYetUpdatedEl);
```
AndrewJakubowicz, line 102: In addition to `null` not being cached, it should emit the warning. Could you also please add a test that the DEV_MODE warning is triggered here. 

You can reuse the pattern used in the property decorator tests: https://github.com/lit/lit/blob/main/packages/reactive-element/src/test/decorators/property_test.ts#L502-L514

packages/reactive-element/src/decorators/query.ts
```diff
@@ -64,6 +82,19 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
     descriptor?: PropertyDescriptor
   ) => {
     const doQuery = (el: Interface<ReactiveElement>): V => {
+      if (DEV_MODE && !el.hasUpdated) {
+        const name =
+          typeof nameOrContext === 'object'
+            ? nameOrContext.name
+            : nameOrContext;
+        issueWarning(
+          'early-query-field-access',
```
AndrewJakubowicz, line 91: This will result in the warning linking to `lit.dev/msg/early-query-field-access`, and we don't yet have any information there. Instead could this be an empty string, and `issueWarning` should only add the link if a code is provided.

Copy the code pattern in lit-html: https://github.com/lit/lit/blob/main/packages/lit-html/src/lit-html.ts#L221-L223

Or:

```
    warning += code
      ? ` See https://lit.dev/msg/${code} for more information.`
      : '';
```

AndrewJakubowicz (commented):
This is looking great, thank you for working on this. I've tested internally and it looks like this is a safe change!

I've left some small changes. Would you also be able to add a Changeset entry for this PR?
Thank you!

MaxArt2501:
@AndrewJakubowicz Will do! Thank you for your review 👍

MaxArt2501:
@AndrewJakubowicz I added a "patch" changeset for @lit/reactive-element. LMK if you need me to change it.

.changeset/quick-moons-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+`@query` properties with the cache flag set do not cache null results before the first update anymore
```
AndrewJakubowicz, line 5: Suggesting a slightly longer changeset:

```suggestion
Fix a bug where accessing a `@query` decorated field with the `cache` flag set before the first update would result in `null` being cached permanently. `null` will no longer be cached before the first update and in `DEV_MODE` now raises a warning.
```

AndrewJakubowicz (approved):
Thank you!

AndrewJakubowicz:
Ah, I just noticed there is a test failure being reported with `Task`. The test is `Tasks can depend on host rendered DOM`, and it's failing because the Task is depending on a `@query` decorated field. This means it accesses that field before the first update which triggers the render.

I think the warning should be scoped to only `cache: true` cases, and that'll solve the test failure.

packages/reactive-element/src/decorators/query.ts
```diff
@@ -64,6 +84,19 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
     descriptor?: PropertyDescriptor
   ) => {
     const doQuery = (el: Interface<ReactiveElement>): V => {
+      if (DEV_MODE && !el.hasUpdated) {
```
AndrewJakubowicz, line 87: To resolve the test failure, and to be more specific for the caching `null` case, I propose adding an `&& cache` here.

augustjk: Interesting. The warning itself is generic that the getter was called before first update, and not specific to the `cache` option.

I wonder if the failing task test should be updated instead to exclude this warning instead of opting out this warning. (also interesting to see that the task test does highlight one way the query decorated property might get accessed before first update).

augustjk: I could also be convinced to just only warn when `cache` is true too. It's more impactful to raise attention if `null` is getting cached where as the warning for none-caching scenarios might just be excess noise.

AndrewJakubowicz: Given that a test has shown a valid reason to access a `@query` decorated field before the first update (in a Task dependency array). I wouldn't want to silence the warning in only that test case, because we may end up having users that get the warning via a valid usage of `Task` depending on a `@query` field.

I think the warning makes sense if there is something bad going on, but the fix to `cache` behavior resolves the worst possible bugs. I propose removing the warnings from this PR.

augustjk: I agree, it seems fine to not warn when there's a legit case to encounter it. @rictic originally suggested in the issue and PR to add a dev mode warning, what do you think about removing it?

.changeset/quick-moons-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
```
augustjk, line 3: We bump these packages together.

```suggestion
---
'@lit/reactive-element': patch
'lit-element': patch
'lit': patch
---
```

packages/reactive-element/src/decorators/query.ts
MaxArt2501: There are some combinations I've considered for the warning. We could issue it if `cache` is true, but also when the query result is actually `null`. Would that be acceptable?

augustjk: Caching null sounds like an unintended behavior so warning in that case sounds reasonable to me.

AndrewJakubowicz: Yes please. Making the warning specific to: `cache: true` and result is `null`, is a perfect time for a warning. It's much more specific to the issue you are fixing.
Thanks!

packages/reactive-element/src/decorators/query.ts
```diff
@@ -64,10 +84,24 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
...
+        const name =
+          typeof nameOrContext === 'object'
+            ? nameOrContext.name
+            : nameOrContext;
+        issueWarning(
+          '',
+          `@query'd field ${JSON.stringify(String(name))} for selector ` +
+            `'${selector}' has been accessed before the first update. This ` +
+            `yields a certain null result if the renderRoot tree has not ` +
+            `been provided beforehand (e.g. via Declarative Shadow DOM).`
```
AndrewJakubowicz, line 98: Thanks for bearing with these changed, I think the last change required is to update the warning to relate to the situation that was handled. E.g., You are preventing the caching of a `null` value.

MaxArt2501: I was just wondering if the message should closely reflect the conditions for issuing the warning 👍 
I've kept a brief explanation of the cause since there isn't a detailed explanation in the documentation yet. I think it could be removed later to keep the message concise.

augustjk (approved):

AndrewJakubowicz (approved):
Nice work!

packages/reactive-element/src/decorators/query.ts
```diff
@@ -64,10 +84,25 @@ export function query(selector: string, cache?: boolean): QueryDecorator {
...
+      if (DEV_MODE && result === null && cache && !el.hasUpdated) {
+        const name =
+          typeof nameOrContext === 'object'
+            ? nameOrContext.name
+            : nameOrContext;
+        issueWarning(
+          '',
+          `@query'd field ${JSON.stringify(String(name))} with the 'cache' ` +
+            `flag set for selector '${selector}' has been accessed before ` +
+            `the first update and yieled null. This is expected if the ` +
```
AndrewJakubowicz, line 97:
```suggestion
            `the first update and returned null. This is expected if the ` +
```

packages/reactive-element/src/test/decorators-modern/query_test.ts
```diff
@@ -96,11 +96,27 @@ import {assert} from '@esm-bundle/chai';
...
 
-  test('returns cached value when accessed before first update', async () => {
+  test('does not cache null values when accessed before first update', async () => {
     const notYetUpdatedEl = new C();
     assert.equal(notYetUpdatedEl.divCached, null);
+
+    if (globalThis.litIssuedWarnings != null) {
+      assert(
+        [...globalThis.litIssuedWarnings].find((w) =>
+          /@query'd field "divCached" with the 'cache' flag set for selector '#blah' has been accessed before the first update and yieled null\. This is expected if the renderRoot tree has not been provided beforehand \(e\.g\. via Declarative Shadow DOM\)\. Therefore the value hasn't been cached\./.test(
```
AndrewJakubowicz, line 106:
```suggestion
          /@query'd field "divCached" with the 'cache' flag set for selector '#blah' has been accessed before the first update and returned null\. This is expected if the renderRoot tree has not been provided beforehand \(e\.g\. via Declarative Shadow DOM\)\. Therefore the value hasn't been cached\./.test(
```

packages/reactive-element/src/test/decorators/query_test.ts
```diff
@@ -90,12 +90,28 @@ import {assert} from '@esm-bundle/chai';
...
 
-  test('returns cached value when accessed before first update', async () => {
+  test('does not cache null values when accessed before first update', async () => {
     const notYetUpdatedEl = new C();
     assert.equal(notYetUpdatedEl.divCached, null);
+
+    if (globalThis.litIssuedWarnings != null) {
+      assert(
+        [...globalThis.litIssuedWarnings].find((w) =>
+          /@query'd field "divCached" with the 'cache' flag set for selector '#blah' has been accessed before the first update and yieled null\. This is expected if the renderRoot tree has not been provided beforehand \(e\.g\. via Declarative Shadow DOM\)\. Therefore the value hasn't been cached\./.test(
```
AndrewJakubowicz, line 100:
```suggestion
          /@query'd field "divCached" with the 'cache' flag set for selector '#blah' has been accessed before the first update and returned null\. This is expected if the renderRoot tree has not been provided beforehand \(e\.g\. via Declarative Shadow DOM\)\. Therefore the value hasn't been cached\./.test(
```

AndrewJakubowicz (approved):
Nit: Spelling issue in `yieled`. I've added the three suggestions to replace it. I changed it to `returned` because `yielded` has generator connotations due to the `yield` keyword.

MaxArt2501:
Thank you, folks, for all the support!

PR #4284, opened 2023-10-11
Allow @query() decorated fields to be null

justinfagnani:
Fixes a user-reported issue on Discord where this kind of field produced an error:

```ts
  @query('dt-dialog') private accessor dialogDce: DtDialog | undefined;
```

because `DtDialog | undefined` is not assignable to `Element`.

`@query()` can return `null`, not `undefined`, so this PR adds `null` to the field type and the error can be fixed by writing the field like:

```ts
  @query('dt-dialog') private accessor dialogDce!: DtDialog | null;
```

Note that the non-null assertion `!` is necessary.

### DIFF
--- a/.changeset/lemon-lamps-shave.md
+++ b/.changeset/lemon-lamps-shave.md
@@ -4,4 +4,4 @@
 'lit': patch
 ---
 
-Allow null to be in the type of @query() decorated fields
+Allow `null` to be in the type of `@query()` decorated fields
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz (approved):
Thanks!

.changeset/lemon-lamps-shave.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/reactive-element': patch
+'lit-element': patch
+'lit': patch
+---
+
+Allow null to be in the type of @query() decorated fields
```
augustjk, line 7: Nit: the `@` prefixed names can tag random users when the changelog are on github unless written as code.
```suggestion
Allow `null` to be in the type of `@query()` decorated fields
```

augustjk (approved):

PR #4285, opened 2023-10-11
Align renderRoot type change with PatchableElement

AndrewJakubowicz:
### Context

I was looking into the `renderRoot`, and `createRenderRoot` type change in 3.0. We possibly missed some locations:

 - `PatchableLitElement`: In Lit element hydration support
 - `PatchableReactiveElement`: In polyfill support

### Test plan

Unit tests. I don't have a good grasp of who consumes these types and what sort of testing would be best.


augustjk:
Hmm it doesn't look like those types are actually getting used inside the patch implementations as accessing `.prototype` turns it all into `any`.

AndrewJakubowicz:
> Hmm it doesn't look like those types are actually getting used inside the patch implementations as accessing `.prototype` turns it all into `any`.

Interesting. So maybe we could remove them? (I haven't checked yet)

PR #4286, opened 2023-10-12
[context] Fix erroneous JSDoc description on `@provide()` decorator

augustjk:
User reported erroneous JSDoc on the `@provide()` decorator
https://discord.com/channels/1012791295170859069/1012791295942590597/1161821452983603240


packages/context/src/lib/decorators/provide.ts
```diff
@@ -16,14 +16,10 @@ import {ContextProvider} from '../controllers/context-provider.js';
...
+ * making it respond to any `context-request` events from its children consumer.
  *
  * @param context A Context identifier value created via `createContext`
- * @param multiple An optional boolean which when true allows the value to be updated
- *   multiple times.
- *
- * See: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
  *
  * @example
  *
```
AndrewJakubowicz, line 25: Nit: in the example directly below, change `consume` to `provide` (in the imports, L27).

AndrewJakubowicz (approved):
Thank you!

packages/context/src/lib/decorators/provide.ts
augustjk: Thank you for catching that!

augustjk: I also want to change usage example below this.
```js
 * class MyElement {
 *   @provide({context: loggerContext})
 *   logger;
 *
 *   doThing() {
 *     this.logger.log('thing was done');
 *   }
 * }
```
`doThing()` is more what a consumer would do. Provider should set an initial value on the decorated property.. but going with the theme here would that be something like this?
```js
 * class MyElement {
 *   @provide({context: loggerContext})
 *   logger = new DefaultLogger();
 * }
 ```

Also wondering if the logger dependency injection theme here is better than just some data object.

AndrewJakubowicz: Yes. I much prefer this because it's also more similar to the example we give in the lit.dev documentation.

augustjk: Thanks! Updated to match the example on lit.dev docs.

PR #4291, opened 2023-10-13
[labs/virtualizer] Update version range for lit

graynorton:
Virtualizer should work with both Lit 2 and Lit 3; update `package.json` to reflect this.

### DIFF
--- a/.changeset/metal-cougars-brake.md
+++ b/.changeset/metal-cougars-brake.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/virtualizer': patch
+---
+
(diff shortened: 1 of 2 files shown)


augustjk (approved):
Thank you! I was looking at whether we need to specify a specific minor version of lit 2 but it appears virtualizer will work with 2.0.0 so this is good.

AndrewJakubowicz (approved):
Excellent! Thank you so much!

justinfagnani (approved):
thanks!

steverep:
Can this also be applied to other packages (e.g. `labs/context`, `labs/motion`, etc.)?  They all have the same issue.

PR #4293, opened 2023-10-14
feat(labs): support elements with no reactive props

43081j:
For each of the wrapper generators, this adds support for elements with no reactive props.

For example:

```ts
class Foo extends LitElement {
  render() {
    return html`no props!`;
  }
}
```

We basically conditionally output the code any reactive property related logic depends on, rather than always outputting it.

## Vue

When generating vue wrappers, we produce two relevant things:

- a `Props` interface (**empty when no properties exist**)
- a `vDefaults.created` function for setting the defaults of these props

If no props exist, we end up with an empty interface (`{}`), which means `keyof Props` is now `never`, leading to various type problems in `vDefaults.created` (which tries to key into props).

For example:

```ts
      interface Props {}
      const vueProps = defineProps<Props>();
      const defaults = reactive({} as Props);
      const vDefaults = {
        created(el: any) {
          for (const p in vueProps) {
            defaults[p as keyof Props] = el[p]; // ERROR: `never` can't be used to key into `defaults`
          }
        }
      };
```

I have changed this to conditionally output `vDefaults.created` if there are props. So without props we will now get:

```ts
      interface Props {}
      const vueProps = defineProps<Props>();
      const defaults = reactive({} as Props);
      const vDefaults = {};
```

## Angular

When generating angular wrappers, we produce a constructor and some properties like so:

```ts
private _el: FooElement;
private _ngZone: NgZone;
constructor(e: ElementRef, ngZone: NgZone) {
  this._el = e.nativeElement;
  this.ngZone = ngZone;
  // ...
}
```

When no props exist, this constructor is basically empty other than the assignments. Which leads to type errors in strict mode, since `_el` and `_ngZone` are unused.

I have changed this to do the following:

- Only output `_el` if we have properties _or_ events (existence of one of these results in code which uses `_el`)
- Only output `_ngZone` if we have properties (only property setters use `_ngZone`)
- Prefix unused parameters with `_` in the constructor (e.g. `constructor(_e: ElementRef)`)

## Open questions/feedback

As you can see, this has resulted in fiddlier code. I wonder if we should slightly refactor how we build up these templates to make such conditional logic a lot easier.

Do we want to think about doing that in this PR instead of the approach i've taken?

Also, in the vue code, i don't fully understand how `vDefaults` gets consumed. Can we skip it entirely? is it ok to have just cut out `created`? etc.

### DIFF
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
@@ -1,31 +1,14 @@
 <script setup lang="ts">
-import {h, useSlots, reactive} from 'vue';
+import {h, useSlots} from 'vue';
 import {assignSlotNodes, Slots} from '@lit-labs/vue-utils/wrapper-utils.js';
 import '@lit-internal/test-element-a/element-without-props.js';
 
 export interface Props {}
 
-const vueProps = defineProps<Props>();
-
-const defaults = reactive({} as Props);
-const vDefaults = {};
-
-let hasRendered = false;
-
 const slots = useSlots();
 
 const render = () => {
-  const eventProps = {};
-
-  const props = eventProps as typeof eventProps & Props;
-  for (const p in vueProps) {
-    const v = vueProps[p as keyof Props];
-    if (v !== undefined || hasRendered) {
-      (props[p as keyof Props] as unknown) = v ?? defaults[p as keyof Props];
-    }
-  }
-
-  hasRendered = true;
+  const props = {} as Props;
 
   return h('element-without-props', props, assignSlotNodes(slots as Slots));
 };
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
@@ -8,9 +8,10 @@ export interface Props {}
 const slots = useSlots();
 
 const render = () => {
-  const props = {} as Props;
+  const eventProps = {};
+  const props = eventProps as typeof eventProps & Props;
 
   return h('element-without-props', props, assignSlotNodes(slots as Slots));
 };
 </script>
-<template><render v-defaults /></template>
+<template><render /></template>
--- a/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
+++ b/packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
@@ -1,0 +1,33 @@
+<script setup lang="ts">
+import {h, useSlots, reactive} from 'vue';
+import {assignSlotNodes, Slots} from '@lit-labs/vue-utils/wrapper-utils.js';
+import '@lit-internal/test-element-a/element-without-props.js';
+
+export interface Props {}
+
+const vueProps = defineProps<Props>();
+
+const defaults = reactive({} as Props);
+const vDefaults = {};
+
+let hasRendered = false;
+
+const slots = useSlots();
+
+const render = () => {
+  const eventProps = {};
+
+  const props = eventProps as typeof eventProps & Props;
+  for (const p in vueProps) {
+    const v = vueProps[p as keyof Props];
+    if (v !== undefined || hasRendered) {
+      (props[p as keyof Props] as unknown) = v ?? defaults[p as keyof Props];
+    }
+  }
+
+  hasRendered = true;
+
+  return h('element-without-props', props, assignSlotNodes(slots as Slots));
+};
+</script>
+<template><render v-defaults /></template>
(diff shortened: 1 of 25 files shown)


kevinpschaaf (approved):
LGTM

e111077:
@kevinpschaaf is there anything blocking merge since your last review?

This PR seems to be blocking #4147 which is blocking the MWC doc site

augustjk (commented):
I haven't really worked much in these parts of the code but this looks mostly fine to me.

My thoughts on the Open Questions part of the the PR description:

I think better ergonomics for conditional logic can be thought out separate from this. It's a bit gnarly but not unintelligible as is.

The `vDefaults` object appears to be the syntax for [custom directives in Vue](https://vuejs.org/api/sfc-script-setup.html#using-custom-directives) that's consumed in the template with `v-defaults` attribute. So if it's empty, the whole thing and the `v-defaults` attribute in the template can likely go away.

packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
```diff
@@ -0,0 +1,33 @@
...
+const render = () => {
+  const eventProps = {};
+
+  const props = eventProps as typeof eventProps & Props;
+  for (const p in vueProps) {
+    const v = vueProps[p as keyof Props];
+    if (v !== undefined || hasRendered) {
+      (props[p as keyof Props] as unknown) = v ?? defaults[p as keyof Props];
+    }
+  }
```
augustjk, line 26: It actually seems like an empty `Props` interface means the whole `vueProps` and code looping through it might be unnecessary too?

43081j: i did think this too but wasn't sure if some vue magic needed this stuff to exist

i'll update 👍

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -89,6 +106,32 @@ const getElementTypeImports = (declaration: LitElementDeclaration) => {
...
+const renderPropsParam = (
+  reactiveProperties: Map<string, ModelProperty>,
+  events: Map<string, EventModel>
+) => {
+  const hasProps = reactiveProperties.size > 0;
+  if (hasProps) {
+    return javascript`
+      const eventProps = ${renderEvents(events)};
+
+      const props = eventProps as (typeof eventProps & Props);
```
augustjk, line 118: It might be possible have no reactive props but non-zero events, in which case we want to keep this part.

Conditions here getting a bit more gnarly here, but perhaps we should also do a `hasEvents = events.size > 0` and account for that too?

`hasPros && hasEvents` is as is.

!hasProps && hasEvents` should still include
```js
const eventProps = ${renderEvents(events)};
const props = eventProps as (typeof eventProps & Props);
```
and `!hasProps && !hasEvents` can be the empty object.

packages/labs/gen-wrapper-vue/goldens/test-element-a/src/ElementWithoutProps.vue
```diff
@@ -0,0 +1,16 @@
...
+
+const slots = useSlots();
+
+const render = () => {
+  const props = {} as Props;
+
+  return h('element-without-props', props, assignSlotNodes(slots as Slots));
+};
+</script>
+<template><render v-defaults /></template>
```
augustjk, line 16: can't comment on the code that generates this but we should also drop the `v-defaults` attribute here when we have no props.

augustjk (changes requested):

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
43081j: can you re-review? it may be simpler than that

i just unconditionally output `eventProps` and `props` now, since they will be empty objects anyway. and only output the loop through vue props when we have reactive props

augustjk: yes doing it binary like that to unconditionally output `eventProps` seems fine. right now i'm only seeing an empty `props` though, unless i'm looking in the wrong place?
https://github.com/lit/lit/blob/4d229f29ee61efb7ab51a31699edab3208782072/packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts#L130-L132

43081j: sorry forgot a `git push`! should be there now

augustjk (approved):
Awesome, thank you!

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
augustjk: I see it, thank you!

PR #4295, opened 2023-10-17
[labs/ssr-client] Fix comment on hydrate function

augustjk:
Fixing some error in the comment walking through an example scenario found while going through it with @AndrewJakubowicz 

Not doing a separate changeset as I don't think it really warrants a version bump and a new release.


AndrewJakubowicz (approved):
Thank you!

Issue #4265, opened 2023-10-06
[labs/ssr] svg templates are surrounded with empty <svg></svg> tags when rendered


pft:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

Any svg``-template is rendered wrapped in `<svg>...</svg>` when rendered with ssr.

This makes it impossible to create a valid svg, unless we do post-processing on the output.

### Reproduction

```
import { render } from "@lit-labs/ssr/lib/render-lit-html.js";
import { svg } from "lit";
import { Readable } from "stream";

const readable = Readable.from(
  render(svg`<svg xmlns="http://www.w3.org/2000/svg"
    height="229.6"
    width="300"
    viewBox="-16 0 250 229.6">${[
      svg`<path d="M116,33.9L116,38.06666666666667C116,42.23333333333333,116,50.56666666666666,116,58.9C116,67.23333333333333,116,75.56666666666668,116,79.73333333333333L116,83.9"></path>`,
      svg`<path d="M116,33.9L116,38.06666666666667C116,42.23333333333333,116,50.56666666666666,116,58.9C116,67.23333333333333,116,75.56666666666668,116,79.73333333333333L116,83.9"></path>`
    ]}</svg>`)
);

const chunks = [];

readable.on("readable", () => {
  let chunk;
  while (null !== (chunk = readable.read())) {
    chunks.push(chunk);
  }
});

readable.on("end", () => {
  const content = chunks.join("");
  console.log(content);
});

```

outputs

```
<!--lit-part h4LS1XLQZw0=--><svg><svg xmlns="http://www.w3.org/2000/svg"
    height="229.6"
    width="300"
    viewBox="-16 0 250 229.6"><!--lit-part--><!--lit-part XW/ddiU0ieA=--><svg><path d="M116,33.9L116,38.06666666666667C116,42.23333333333333,116,50.56666666666666,116,58.9C116,67.23333333333333,116,75.56666666666668,116,79.73333333333333L116,83.9"></path></svg><!--/lit-part--><!--lit-part XW/ddiU0ieA=--><svg><path d="M116,33.9L116,38.06666666666667C116,42.23333333333333,116,50.56666666666666,116,58.9C116,67.23333333333333,116,75.56666666666668,116,79.73333333333333L116,83.9"></path></svg><!--/lit-part--><!--/lit-part--></svg></svg><!--/lit-part-->
```

### Workaround

[...]

PR #4298, opened 2023-10-17
[labs/ssr] Fix svg templates getting surrounded by empty svg tags

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4265

### Context

`TemplateResultType.SVG` will wrap svg template fragments with an `<svg>` element on the client so the svg elements get parsed in the svg parser state.
On the server if we do this, it will emit excess wrapping svg elements.

The fix is either, 1. add plucking logic to the server, to pluck out the wrapping svg element. Or, a much simpler change which is treat the svg template exactly the same as the html template.

### Test plan

Tested with a unit test and an integration test.


packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -297,6 +297,26 @@ export const tests: {[name: string]: SSRTest} = {
     stableSelectors: ['div', 'span'],
   },
 
+  'ChildPart accepts TemplateResult with SVG type': {
```
augustjk, line 300:
```suggestion
  // Regression test for https://github.com/lit/lit/issues/4265
  // Ensures we do not get extra svg tags around `svg` tagged templates.
  'ChildPart accepts TemplateResult with SVG type': {
```

augustjk (approved):
Thank you!

PR #4299, opened 2023-10-17
Update dependency version range to accept Lit v2

augustjk:
Affected packages:

- '@lit-labs/scoped-registry-mixin': patch
- '@lit-labs/preact-signals': patch
- '@lit-labs/observers': patch
- '@lit/localize-tools': patch
- '@lit-labs/motion': patch
- '@lit-labs/router': patch
- '@lit/localize': patch

I haven't touched SSR and related packages yet as they'll require a more specific range of v2 to work.

I tested that these packages work with the lower version by manually installing `lit@2.0.0` or `@lit/reactive-element@1.0.0` within the package directory and running their tests.

### DIFF
--- a/packages/localize/package.json
+++ b/packages/localize/package.json
@@ -89,7 +89,7 @@
     "/init/"
   ],
   "dependencies": {
-    "@lit/reactive-element": "^2.0.0",
+    "@lit/reactive-element": "^1.0.0 || ^2.0.0",
     "lit": "^2.0.0 || ^3.0.0"
   }
 }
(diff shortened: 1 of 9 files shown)


AndrewJakubowicz (approved):
👏 👏

packages/localize/package.json
```diff
@@ -90,6 +90,6 @@
   ],
   "dependencies": {
     "@lit/reactive-element": "^2.0.0",
```
rictic, line 92: should this be broadened too?

rictic (approved):

packages/localize/package.json
augustjk: oops, yes it should!

PR #4300, opened 2023-10-17
Update devDependencies to not reference pre versions of internal packages

augustjk:
Found while going through package.json files.
Doing an empty changeset as dev dependency updates don't affect end users.


AndrewJakubowicz (approved):

AndrewJakubowicz:
If we wanted to be very conservative we could do a small patch changeset entry. In the case that a dev dependency has some impact on the published package.

AndrewJakubowicz (approved):

PR #4306, opened 2023-10-18
Fix dependencies to use stable instead of pre-release vesrions

augustjk:
Affected packages:

- '@lit-labs/gen-wrapper-vue': patch
- '@lit/reactive-element': patch
- '@lit-labs/testing': patch
- '@lit-labs/nextjs': patch
- 'lit-element': patch

These packages still had -pre versions as production dependencies seemingly missed by changeset during release. Did them separately from https://github.com/lit/lit/pull/4300 as these would require a version bump so downstream users don't install the prerelease versions.

`@lit-labs/testing` was also missed from https://github.com/lit/lit/pull/4299 in upping the version range.

### DIFF
--- a/.changeset/friendly-rabbits-arrive.md
+++ b/.changeset/friendly-rabbits-arrive.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/testing': patch
+---
+
+Update version range for `lit` dependency to include v2. This allows projects still on lit v2 to use this package without being forced to install lit v3.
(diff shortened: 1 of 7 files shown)


rictic (approved):

AndrewJakubowicz (approved):

augustjk:
Edited to add 'lit-element' too.

Now the only outstanding packages that need a pass are labs/ssr and related ones like labs/eleventy-plugin-lit, labs/ssr-client, labs/ssr-dom-shim.

PR #4309, opened 2023-10-19
Improve types for (un)compiled template results

rictic:
Adds two new types: `UncompiledTemplateResult` and `MaybeCompiledTemplateResult`. Currently `UncompiledTemplateResult` is the same as `TemplateResult`, and `MaybeCompiledTemplateResult` is the union of the compiled and uncompiled types.

We'd like for code to be able to handle the `html` and `svg` functions returning a MaybeCompiledTemplateResult. Changing that type will be a major breaking change though, and we don't want to do a major version so quickly, but this lays the ground work, and provides future-compatible types that people can use when they want to be explicit about which type they mean, because while TemplateResult will change, MaybeCompiled, Uncompiled, and Compiled will all have the same meanings in the next major version.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -473,6 +473,15 @@ export type UncompiledTemplateResult<T extends ResultType = ResultType> = {
   values: unknown[];
 };
 
+/**
+ * This is a template result that may be either uncompiled or compiled.
+ *
+ * In the future, TemplateResult will be this type. If you want to explicitly
+ * note that a template result is potentially compiled, you can reference this
+ * type and it will continue to behave the same through the next major version
+ * of Lit. This can be useful for code that wants to prepare for the next
+ * major version of Lit.
+ */
 export type MaybeCompiledTemplateResult<T extends ResultType = ResultType> =
   | UncompiledTemplateResult<T>
   | CompiledTemplateResult;
@@ -492,8 +501,8 @@ export type MaybeCompiledTemplateResult<T extends ResultType = ResultType> =
  * In Lit 4, this type will be an alias of
  * MaybeCompiledTemplateResult, so that code will get type errors if it assumes
  * that Lit templates are not compiled. When deliberately working with only
- * one, use either CompiledTemplateResult or UncompiledTemplateResult
- * explicitly.
+ * one, use either {@linkcode CompiledTemplateResult} or
+ * {@linkcode UncompiledTemplateResult} explicitly.
  */
 export type TemplateResult<T extends ResultType = ResultType> =
   UncompiledTemplateResult<T>;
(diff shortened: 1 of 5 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -466,13 +466,38 @@ const COMMENT_PART = 7;
...
+ * and type of template (html or svg).
+ *
+ * `TemplateResult` objects do not create any DOM on their own. To create or
+ * update DOM you need to render the `TemplateResult`. See
+ * [Rendering](https://lit.dev/docs/components/rendering) for more information.
+ *
+ * In Lit 4, this type will be an alias of
+ * MaybeCompiledTemplateResult, so that code will get type errors if it assumes
+ * that Lit templates are not compiled. When deliberately working with only
+ * one, use either CompiledTemplateResult or UncompiledTemplateResult
```
AndrewJakubowicz, line 495: Nit: Could make `CompiledTemplateResult` and `UncompiledTemplateResult` use `@linkcode` so they link to the respective docs.

packages/lit-html/src/lit-html.ts
```diff
@@ -466,13 +466,38 @@ const COMMENT_PART = 7;
...
  */
-export type TemplateResult<T extends ResultType = ResultType> = {
+export type UncompiledTemplateResult<T extends ResultType = ResultType> = {
   // This property needs to remain unminified.
   ['_$litType$']: T;
   strings: TemplateStringsArray;
   values: unknown[];
 };
 
+export type MaybeCompiledTemplateResult<T extends ResultType = ResultType> =
```
AndrewJakubowicz, line 476: Nit: This could use a JSDoc to clarify why this is used. Probably want to call out that this should ideally be the type of `html` or `svg` when you're trying to target both uncompiled and compiled environments.

packages/lit-starter-ts/src/test/my-element_test.ts
```diff
@@ -16,7 +17,9 @@ suite('my-element', () => {
   });
 
   test('renders with default values', async () => {
-    const el = await fixture(html`<my-element></my-element>`);
+    const el = await fixture(
+      html`<my-element></my-element>` as UncompiledTemplateResult
+    );
```
AndrewJakubowicz, line 22: I'm not clear why the changes in `my-element_test.ts` are needed? It seems outside the scope of `lit-html`?

packages/lit-html/src/lit-html.ts
rictic: Good point! Done

packages/lit-html/src/lit-html.ts
rictic: Oh nice!

packages/lit-starter-ts/src/test/my-element_test.ts
rictic: As part of this change I also fixed all the type errors that would arise if we changed TemplateResult to be MaybeCompiledTemplateResult

AndrewJakubowicz (approved):
Nice!

PR #4310, opened 2023-10-19
feat(lit-html/directives/when.ts): improve typing

### LINKED ISSUE #4085: [lit-html] when directive should have better type def
### Should this be an RFC?

### Description

It would be useful for the `when` directive to guard and properly type  its checked value.
For example:

```ts
const val: number|null|undefined = 2;
const  fn = (v: number)=>v;
when(
val,
 // type error because val can be undefined
 ()=>fn(val) 
)
```
In reality `val` can't be undefined because of the checks in `when`. 

A possible non breaking change solution could be to pass our guard types thru a param to the trueCase and falseCase fn.

```ts
const val: number|null|undefined = 2;
const  fn = (v: number)=>v;
when(
val,
 // when passes value to trueCase and `v`  is Exclude<typeof val, undefined|null|false>
 (v)=>fn(v) 
)
```

megheaiulian:
Improve typing for the `when` directive making it more dynamic.

```ts
let user = {name: string} | undefined;

when(
  user,
  (user) => user.name
)
```

In the trueCase and falseCase functions the condition is correctly type
and we no longer need a non-null assertion.

Fixes #4085

### DIFF
--- a/.changeset/kind-insects-shake.md
+++ b/.changeset/kind-insects-shake.md
@@ -1,0 +1,6 @@
+---
+"lit-html": patch
+"lit": patch
+---
+
+feat(lit-html/directives/when.ts): improve typing
--- a/.changeset/kind-insects-shake.md
+++ b/.changeset/kind-insects-shake.md
@@ -3,4 +3,4 @@
 'lit': patch
 ---
 
-Improve typing for the `when` directive.
+The `when()` directive now calls the case functions with the provided condition value as an argument. This allows the narrowing of types for the condition value based on its truthiness when used as a parameter for the case function.
--- a/.changeset/kind-insects-shake.md
+++ b/.changeset/kind-insects-shake.md
@@ -1,6 +1,6 @@
 ---
-"lit-html": patch
-"lit": patch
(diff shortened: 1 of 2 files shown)


.changeset/kind-insects-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit-html': patch
+'lit': patch
+---
+
+Improve typing for the `when` directive.
```
augustjk, line 6:
```suggestion
The `when()` directive now calls the case functions with the provided condition value as an argument. This allows the narrowing of types for the condition value based on its truthiness when used as a parameter for the case function.
```

augustjk (approved):
Thank you for the PR!
The change is pretty minimal but does affect runtime and usage will look different if you want to take advantage of this. Whether we want to document this everywhere I think is up for debate but it should be fine to merge this and leave it as an undocumented feature in the interim.
I did leave a suggestion to make at least the changelog entry more verbose as the one place to document this behavior change.

AndrewJakubowicz (approved):

Issue #4289, opened 2023-10-13
[labs/ssr-client] hydrate-lit-html can break when compiled with SWC


augustjk:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

This was noticed while testing `@lit-labs/nextjs`.
Next.js uses SWC for production JS builds (can optionally use terser instead).

Now that we emit ES2021 and keep logical assignments `??=`, this issue https://github.com/swc-project/swc/issues/8119 started happening.

These lines
https://github.com/lit/lit/blob/89a5b0882b3048e3e95a22eb739c649adc9de055/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts#L164-L167

got minified to
```js
            r ??= n = A(t, o, l, i),
            s ??= o
```
with the side-effectful `openChildPart()` call moved behind the conditional assignment.

### Reproduction

.

### Workaround

While this may or may not be fixed in SWC, an easy fix on our end is to not use logical nullish assignment for this line in particular to make sure `openChildPart()` always gets called.

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

1.1.4

### Browser/OS/Node environment

n/a

PR #4312, opened 2023-10-20
[labs/ssr-client] Avoid nullish logical assignment in hydrate-lit-html

augustjk:
Fixes https://github.com/lit/lit/issues/4289

While swc did fix this upstream it's not released yet and I don't know how long it takes for that to make it into Next.js's version of swc.

It seems like a tiny lift for us to add this to make it work now.

### DIFF
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -163,7 +163,7 @@ export const hydrate = (
       }
       // Create a new ChildPart and push it onto the stack
       currentChildPart = openChildPart(rootValue, marker, stack, options);
-      // Using nullish logical assignment below can cause some minifier to move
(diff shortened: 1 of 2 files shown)


packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -163,7 +163,12 @@ export const hydrate = (
       }
       // Create a new ChildPart and push it onto the stack
       currentChildPart = openChildPart(rootValue, marker, stack, options);
-      rootPart ??= currentChildPart;
+      // Using nullish logical assignment below can cause some minifier to move
```
rictic, line 166:
```suggestion
      // Using nullish logical assignment below can cause next.js's swc to move
```

More specific, so we know what to check when we're thinking about undoing this later

rictic (approved):

AndrewJakubowicz (approved):
Nice!

PR #4318, opened 2023-10-21
Draft: Trying out DOMParser for lit-html

augustjk:
DOMParser actually returns a document, and can weirdly move comments around if we don't specifically wrap the text in a `<body>` tag, so I'm re-parenting the nodes from the document body to a new document fragment.

The template instance now stores the document fragment instead of the template element in `.el`.

The parser always need to run with text/html, even for svg tags, as we want created elements in the document to actually be SVGElements so we still have to do the `<svg>` element wrapping.

For polyfill-support, since `ShadyDOM.prepareTemplateStyles` expects a template element, I'm then reparenting the fragment back into a template element, letting ShadyDOM prepare, then reparent again.

These re-parentings probably will make things slower.

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -856,7 +856,7 @@ const getTemplateHtml = (
 export type {Template};
 class Template {
   /** @internal */
-  static parser = new DOMParser();
+  static parser: DOMParser;
 
   /** @internal */
   el!: DocumentFragment;
@@ -1001,6 +1001,9 @@ class Template {
     _options?: RenderOptions | undefined,
     type?: ResultType
   ) {
+    if (this.parser === undefined) {
+      this.parser = new DOMParser();
+    }
     const doc = this.parser.parseFromString(
       '<body>' + (html as unknown as string) + '</body>',
       'text/html'
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -47,7 +47,7 @@ export namespace LitUnstable {
       kind: 'template prep';
       template: Template;
       strings: TemplateStringsArray;
-      clonableTemplate: HTMLTemplateElement;
+      clonableTemplate: DocumentFragment;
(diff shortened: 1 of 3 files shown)


packages/lit-html/src/lit-html.ts
```diff
@@ -999,10 +996,27 @@ class Template {
...
-    el.innerHTML = html as unknown as string;
-    return el;
+  static createElement(
+    html: TrustedHTML,
+    _options?: RenderOptions | undefined,
+    type?: ResultType
+  ) {
+    if (this.parser === undefined) {
+      this.parser = new DOMParser();
+    }
```
justinfagnani, line 1006:
```suggestion
    this.parser ??= new DOMParser();
```

packages/lit-html/src/lit-html.ts
```diff
@@ -999,10 +996,27 @@ class Template {
...
+    const doc = this.parser.parseFromString(
+      '<body>' + (html as unknown as string) + '</body>',
```
justinfagnani, line 1008: This concerns me. Maybe all element types work in `<body>`, but they definitely do in `<template>`

augustjk: Yeah I don't like it either but couldn't think of a better work around. I noticed test failures without this and remembered I had to do the same thing for labs/testing here: https://github.com/lit/lit/blob/cabe72a863a5de14a4bbca384374db748dd9b4c5/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts#L85-L94

Any top level comment gets moved outside the `<body>` tag by the DOM Parser.
image

sorvell: Might be relevant: https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody

augustjk:
Closing as this is no longer relevant.
We might instead consider switching `el.innerHTML =` with `e.setHTMLUnsafe()`, which will also parse and attach declarative shadow DOM.

Issue #4307, opened 2023-10-19
[reactive-element] "Multiple versions of Lit loaded." warning in dev mode


johnhunter:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

Upon upgrading to lit@3 and running a dev build I get the following console warning

```
Multiple versions of Lit loaded. Loading multiple versions is not recommended.
```

The warning is triggered in `reactive-element.ts`

```
// IMPORTANT: do not change the property name or the assignment expression.
// This line will be used in regexes to search for ReactiveElement usage.
(global.reactiveElementVersions ??= []).push('2.0.0');
if (DEV_MODE && global.reactiveElementVersions.length > 1) {
  issueWarning!(
    'multiple-versions',
    `Multiple versions of Lit loaded. Loading multiple versions ` +
      `is not recommended.`
  );
}
```
And inspecting the `reactiveElementVersions` value I find `['2.0.0', '2.0.0']` suggesting that reactive-element might be getting loaded twice. This might be explained by the dependencies tree which has been mentioned in issue #4305.

```
└─┬ lit@3.0.0
  ├── @lit/reactive-element@2.0.0
  └─┬ lit-element@4.0.0
    └── @lit/reactive-element@2.0.0
```

### Reproduction

**Steps to reproduce**

1. Clone https://github.com/johnhunter/wc-brewery-app
2. run `npm install`
3. run `npm start`
4. Open http://localhost:8000/ in browser with devtools console enabled
5. In the browser console run `globalThis.reactiveElementVersions`

**Expected**

- No "Multiple versions of Lit loaded." warnings
- reactiveElementVersions array is `['2.0.0']`

**Actual**

- "Multiple versions of Lit loaded." warning is shown
- reactiveElementVersions array is `['2.0.0', '2.0.0']`

<img width="807" alt="image" src="https://github.com/lit/lit/assets/219160/7ae1b8df-b8f5-4dd3-8a24-4b4e115a404a">

### Workaround

None. This does not appear to cause issues other than the dev mode warnings

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

Failing in lit@3.0.0, working in lit@2.8.0

### Browser/OS/Node environment

Browser: Chrome / Brave 118.0.5993.88 
OS: MacOS 14.0
Node: 18.17.1
NPM: 9.6.7

justinfagnani:
Is this:
```
└─┬ lit@3.0.0
[...]

PR #4319, opened 2023-10-23
[labs/virtualizer] correctly specify lit dependency as a peer

Roaders:
currently lit is specified as a "prod" dependency rather than a peer dependency. This will lead more developers to ending up with the "Muliple versions of lit loaded" error message

### DIFF
--- a/packages/labs/virtualizer/package.json
+++ b/packages/labs/virtualizer/package.json
@@ -159,7 +159,9 @@
     "tachometer": "^0.7.0"
   },
   "dependencies": {
-    "lit": "^2 || ^3",
     "tslib": "^2.0.3"
+  },
+  "peerDependencies": {
+    "lit": "^2 || ^3"
   }
 }


augustjk:
Thank you for the PR but semantically `lit` is a true dependency for `@lit-labs/virtualizer` as it is directly used to create the `lit-virtualizer` element as well as the `virtualize` directive.

A peer dependency would be more indicative that this package doesn't use the that package directly but is meant to be used along side it. That is not the case here.

The duplicate version may still happen just due to how npm chooses to dedupe the package. I have a suggestion to try and coerce npm to dedupe the version you want by installing the version to dedupe in your root package here: https://github.com/lit/lit/issues/4307#issuecomment-1773245478

Roaders:
Sorry but that is not what a peer depednency means. It is possible to specify an optional peer if you like but that requires the `peerDependenciesMeta` to mark them as optional. IN this case you are saying "this dependency is not required but if it's installed you need to have this version".
Peers without this are not optional.

If you use a normal dependency rather than a peer dependency then you can end up with this dependency tree:

-lit: 0.x
- @lit-labs/virtualizer: 1.x
-  - lit: 2.x

with no warnings that something bad is happening. if the lit dependency is specified as a peer by `@lit-labs/virtualizer` then npm will signal that there is an error and that it cannot be installed with a lit version of 1.

augustjk:
Hmm.. I see what you mean there. So the difference then is whether we want npm to warn/error when a `lit` package with a version outside the range specified in `peerDependencies` is present?

I am still torn about this though. `lit-virtualizer` is a self contained element and the version of `lit` it uses is an implementation detail. It is valid to use this custom element with any version of Lit or even non-Lit projects. If you're using the `virtualize` directive, then the peer argument holds more ground there as it's meant to be used inside lit-html templates of certain versions. @justinfagnani what do you think?

justinfagnani:
I agree with your assessment... I'm torn. If it were two packages it would be a bit more clear.

Roaders:
The reason I raised this was because we were getting the multiple versions of lit installed error. This was before the range was increased to include lit@2.
If virtualiser can safely use its own version then a prod dependency rather than a peer is fine - but decs will still het the error about multiple versions.
IMHO a peer is better to remove the warning and to keep the dependency tree simple.

augustjk:
Yeah I'm becoming more convinced by this idea and feel we should consider this, not just for virtualizer but for our other packages too. Thank you for raising this!

rictic:
So, going by the npm documentation, `lit` is a dependency of `@lit-labs/virtualizer`, not a peer dependency.

> In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.

– https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies

Since `@lit/virtualizer` directly imports `lit` but `lit` would never import `@lit/virtualizer`, this seems like it's not the intended use case of `peerDependencies`.

We also don't want to _error_ if there are multiple versions of lit, or if `@lit-labs/virtualizer` needs to install its own version. Our ideal outcome is that `npm` just tried harder not to introduce more versions, but that's not a feature of `npm` that I'm aware of.

Flipping the question around, if we adopted this change, why wouldn't we make every dependency of every package a peer dependency? And if not every dep, then how would we determine which should be peers and which shouldn't?

Worth noting that the reason that Lit warns about multiple versions in the page generally applies to every library. It's that it's wasteful, you're shipping more bytes over the wire than you (probably) need to.

augustjk:
We've discussed this at length in our engineering meeting this morning.

We're going to close this PR for the following reasons:
- This particular package vends the `<lit-virtualizer>` element meant to be used stand alone making `lit` semantically a dependency, not a peer dependency.
- Having multiple versions of lit warning is not considered a breakage. The warning is added only in dev mode purely to surface an opportunity for optimization.

However, this did surface a good discussion and we will consider marking peer dependencies where it semantically makes sense for our other packages. We would also like to do some more research on how other tools like yarn and pnpm handle peer deps and whether we might cause unwarranted install conflicts.

We also have action items from this to improve the multiple version warning and documentation behind that like https://github.com/lit/lit.dev/issues/455.

Thanks again for the PR and sorry that we're not able to accept this. We really do appreciate the conversation here.

Roaders:
Thanks for the detailed reply. I would like to counter with 2 points though:

- The fact that the warning is at dev time only does not swing the argument at all for me. Pretty much all the time users are not going to be looking at the dev console and if they are they are not going to care or understand the messages there. It is devs at dev time that will see the message and burn time looking into it. Just FYI 2 developers on our small team of less than 10 both independentaly saw this message and spent time investigating it.
- What do you consider the downside of marking it as a peer?

IMHO to support the developers who will be using this it gives a better developer experience to mark this as a peer and not a prod dependency.

augustjk:
For point 1, that is totally on us to make the message more clear. The warning should be a non-scary message and the url it points to should clearly state if you should care about it or not, with ways to resolve it should you care. We're trying to err on the side of giving devs more info, but we should do so without wasting their time.

Point 2, the potential downside is causing npm/pnpm/yarn install errors due to unreconcilable peer dependency when it should be perfectly valid to just have a nested dependency of specified version.

For virtualizer in particular, we worry for users wanting to use just the `<lit-virtualizer>` element out of the package. Since `lit` is merely an implementation detail for this web component, it should be able to be used with `lit-html@1`, `lit@2`, `lit@3` or really any version of it, or no version of it.

My personal gut feeling is that the risk of this causing errors for end users is very very low but as a team we don't have the confidence to say everything will be fine.

Another reason we're hesitant to immediate jump on this is that marking everything as a dependency has been working fine for us so far for a long time, and we haven't heard from our users wanting this until now. Also, some of this feels like we're trying to solve bad npm behavior that it's failing to dedupe packages that could be deduped, regardless of whether it's a dependency or a peer dependency.

We have https://github.com/lit/rfcs/issues/31 so we can keep this on our radar. More feedback there or info on peer dependency behavior for npm/pnpm/yarn would be helpful data for us to prioritize this.

PR #4322, opened 2023-10-23
Get the Analyzer fully running in browsers

justinfagnani:
This PR fixes a few cases where the analyzer was relying on Node-specific behavior, so that the analyzer can now run cleanly in browsers.

To test this the PR creates separate test:server and test:browser scripts where the browser test uses a Rollup-built JS module for typescript and a copy of the server tests' InMemoryAnalyzer itself with some fixes for running outside node.

The InMemoryAnalyzer code is a bit messy, but it's test code. I think we should probably offer a supported InMemoryAnalyzer at some point, derived from out test versions.

The test changes are in the first two commits, the fixes in the third.


packages/labs/analyzer/rollup.config.typescript.js
```diff
@@ -0,0 +1,25 @@
...
+      file: 'test/browser/typescript.js',
+      format: 'esm',
+    },
+    plugins: [
+      commonjs({
+        ignore: (_id) => true,
+      }),
+    ],
+  },
+];
```
rictic, line 25: neat!

packages/labs/analyzer/rollup.config.typescript.js
```diff
@@ -0,0 +1,25 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
rictic, line 1: Why is this extraneous? It's called on line 20

rictic (commented):
Do we want to remove NodeJS typings from the analyzer package? Or only make them available in specific files?

That wouldn't help with typescript.sys but hopefully we won't run into that again

packages/labs/analyzer/rollup.config.typescript.js
justinfagnani: eslint was yelling that `@rollup/plugin-commonjs` should be in `dependencies` instead of `devDependencies`. I didn't take the time to see if a config change would help.

packages/labs/analyzer/rollup.config.typescript.js
justinfagnani: All thanks to @aomarks !

packages/labs/analyzer/src/test/browser/analyzer_test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+suite('Analyzer', () => {
+  test('works', () => {
+    const elementPath = '/src/test-element.ts' as AbsolutePath;
+    const analyzer = new InMemoryAnalyzer({
+      typescript: ts,
+      lang: 'ts',
+      files: {
+        [elementPath]: `
+      import {LitElement, html, css} from 'lit';
```
kevinpschaaf, line 14: I expect that the declaration assertion below is a `ClassDeclaration` and not a `LitElementClassDeclaration` because stuff in the typechecker probably goes awry when it can't resolve this `lit` import? But we can't really test that you _could_ preload the cache with `lit` without coming up with a good scheme for how to do that, which I grok is out of scope of this PR for now?

packages/labs/analyzer/package.json
```diff
@@ -35,9 +37,25 @@
       ],
       "clean": "if-file-deleted"
     },
+    "build:typescript": {
```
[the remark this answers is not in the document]
kevinpschaaf: or `build:tslib-for-browser` or even `build:tslib-esm`...?

packages/labs/analyzer/tsconfig.json
```diff
@@ -25,7 +25,7 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitOverride": true,
     "skipLibCheck": true,
-    "types": ["node"]
+    "types": ["node", "mocha"]
```
justinfagnani, line 28: I think ultimately the safest thing here would be to have separate tsconfigs:
1. The main library code, no node typings
2. The node test code and package-analyzer function, with node typings
3. _Maybe_ a third with browser typings if they were needed in browser tests.

Not having Node typings would have caught the place where we imported `path`.

packages/labs/analyzer/src/test/browser/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+      cache[normalize(fileName)].version.toString(),
+    getScriptSnapshot: (fileName: string) => {
+      if (isLib(fileName)) {
+        console.log('requested lib', fileName);
+        return undefined;
+        // return fs.existsSync(fileName)
+        //   ? ts.ScriptSnapshot.fromString(
+        //       fs.readFileSync(path.normalize(fileName), 'utf-8')
+        //     )
+        //   : undefined;
```
kevinpschaaf, line 122: I think this is for stuff like `lib.dom.d.ts`, so I guess this is another spot where (something) will need to pre-load the in-memory cache with this stuff too, same as the `deps`.

packages/labs/analyzer/src/test/browser/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+  [index: string]: string;
+}
+
+// Note that because some paths come into the language host filesystem
+// abstractions in posix format and others may come in in OS-native
+// format, we normalize all paths going in/out of the in-memory cache to
+// posix format. There is apparently no path lib method to do this.
+// https://stackoverflow.com/questions/53799385/how-can-i-convert-a-windows-path-to-posix-path-using-node-path
+// const normalize = (p: string) => p.split(path.sep).join(path.posix.sep);
+const normalize = (p: string) => p;
```
kevinpschaaf, line 26: Remove this? I guess we don't have to worry about Window-style paths getting in in the browser?

kevinpschaaf (approved):
LGTM

packages/labs/analyzer/rollup.config.typescript.js
augustjk: there's a way to escape that here. (the labs/react rule should be removed as it no longer applies)
https://github.com/lit/lit/blob/cabe72a863a5de14a4bbca384374db748dd9b4c5/.eslintrc.json#L41-L47

augustjk: or here
https://github.com/lit/lit/blob/cabe72a863a5de14a4bbca384374db748dd9b4c5/.eslintrc.json#L82-L104

packages/labs/analyzer/package.json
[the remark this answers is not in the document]
justinfagnani: Went with `build:ts-bundle`. `tslib-` sounds like something for the tslib package.

packages/labs/analyzer/rollup.config.typescript.js
justinfagnani: I applied both of @augustjk's suggested changes

packages/labs/analyzer/package.json
```diff
@@ -46,6 +64,21 @@
...
+      "dependencies": [
+        "build",
+        "build:ts-bundle",
+        "../../tests:build"
+      ],
+      "env": {
+        "BROWSERS": {
+          "external": true
+        }
+      },
```
augustjk, line 79:
```suggestion
    "test:browser": {
      "command": "node ../../tests/run-web-tests.js \"test/browser/**/*_test.js\" --config ../../tests/web-test-runner.config.js",
      "dependencies": [
        "build",
        "build:ts-bundle",
        "../../tests:build"
      ],
      "env": {
        "BROWSERS": {
          "external": true
        },
        "MODE": "dev"
      },
```

packages/labs/analyzer/package.json
```diff
@@ -35,9 +37,25 @@
...
     "test": {
+      "dependencies": [
+        "test:server",
+        "test:browser"
+      ]
+    },
+    "test:server": {
       "#comment": "The quotes around the file regex must be double quotes on windows!",
-      "command": "cross-env NODE_OPTIONS=--enable-source-maps uvu test \"_test\\.js$\"",
+      "command": "cross-env NODE_OPTIONS=--enable-source-maps uvu test/server \"_test\\.js$\"",
```
augustjk, line 58:
```suggestion
      "command": "uvu test/server \"_test\\.js$\"",
      "env": {
        "NODE_OPTIONS": "--enable-souce-maps"
      },
```

.changeset/eleven-ducks-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/analyzer': major
```
augustjk, line 2: It's currently on 0.10.0 so minor can be breaking too.
```suggestion
'@lit-labs/analyzer': minor
```

augustjk (approved):

Issue #4288, opened 2023-10-13
[infra] Benchmarks workflow size check fails on forks


augustjk:
See https://github.com/lit/lit/actions/runs/6512809155/job/17691182042?pr=4233

PR workflows running from forks cannot create comments. This should be skipped, or be a separate workflow that runs on a different hook like the benchmark-report.

Logically I feel like size check should be separate from benchmarks too.

PR #4323, opened 2023-10-24
[infra] Split out size check workflow from benchmarks

augustjk:
Fixes https://github.com/lit/lit/issues/4288

I followed a similar pattern as the benchmarks report workflow added here https://github.com/lit/lit/pull/4041

Workflows running on forks can't use actions that write comments on the PR. This moves the size check to its own workflow, and a separate workflow that triggers on the completion of that size check is responsible for writing the comment.

The original size check wrote a markdown file on failure for the comment message so that is added as an artifact, similar to how benchmark results are done.

We won't see this work on this PR as workflows that are triggered by other workflows only run off of a protected main branch for security as these workflows have elevated access. It'll show up for PRs after this is merged to main.

### DIFF
--- a/.github/workflows/sizecheck-report.yaml
+++ b/.github/workflows/sizecheck-report.yaml
@@ -1,0 +1,40 @@
+name: Report Sizecheck Results
+
+on:
+  workflow_run:
+    workflows: [Sizecheck]
+    branches: ['**']
+    types: [completed]
+
+jobs:
+  on-success:
+    name: Report sizecheck success
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Comment on size report empty
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: 'The size of lit-html.js and lit-core.min.js are as expected.'
+          comment_tag: check-size # ensures we overwrite old failure comment if present
+
+  on-failure:
+    name: Report sizecheck failure
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      # Download the artifact from the triggering workflow that contains the
+      # size check results to report
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: ${{ github.event.workflow.id }}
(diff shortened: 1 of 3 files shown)


.github/workflows/sizecheck-report.yaml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      # Download the artifact from the triggering workflow that contains the
+      # size check results to report
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: ${{ github.event.workflow.id }}
+          run_id: ${{ github.event.workflow_run.id }}
+          name: sizecheck
+          path: scripts
```
AndrewJakubowicz, line 34: Should this path be the markdown file? `scripts/check-size-out.md`?

AndrewJakubowicz (approved):
Thank you :D

.github/workflows/sizecheck-report.yaml
augustjk: This one specifies the directory location where the artifact files are unzipped into. I think omitting it would be okay here too which will unzip the files at the root and next step could just find the file from the root, but I just decided to keep the directory name that it had from the workflow for no particular reason.

rictic: If the path here is scripts and the path for the uploaded file is scripts/check-size-out.md will that write the file to scripts/scripts/check-size-out.md?

augustjk: No. The upload artifact action here https://github.com/lit/lit/blob/04d16b181c292f0951097fce691ea182bb14b29c/.github/workflows/sizecheck.yaml#L27-L32
specifies the path to the file, and the individual file will be uploaded without the directory structure. https://github.com/actions/upload-artifact#upload-an-individual-file

augustjk: We do the same thing for benchmarks here with paths to individual files
https://github.com/lit/lit/blob/cabe72a863a5de14a4bbca384374db748dd9b4c5/.github/workflows/benchmarks.yml#L86-L95
that are downloaded to `results` directory and read from there
https://github.com/lit/lit/blob/cabe72a863a5de14a4bbca384374db748dd9b4c5/.github/workflows/benchmarks-report.yaml#L33-L48

PR #4326, opened 2023-10-25
[localize] Loosen type of `@localized()` decorator

augustjk:
This alleviates type errors when there's a different version of `reactive-element` that `@lit/localize` is using versus what the user's component is using.

Because `localized` is a class decorator, the standard `Interface` with `Omit` or `Pick` was not usable as there were static properties on it that conflict. I tried `Pick<typeof ReactiveElement, 'addInitializer'>` but the initializer type references `ReactiveElement` directly which breaks because of `renderRoot`.

I also removed the dependency on `@lit/reactive-element` and just import everything from `lit`.

I do wonder though.. `@lit/localize` only uses the `lit` package for types/interfaces. It exports utilities strictly meant to be used with `lit` projects and nothing is standalone. This is a prime candidate for marking `lit` as a peer dependency as described in https://github.com/lit/rfcs/issues/31.

I tried it locally by updating the package.json to specify lit as a peer dependency, without the loosened type fix `npm pack` and installed the tarball in a lit 2 test project that was having type errors before. It no longer errors as it did not create a nested duplicate `@lit/reactive-element`. But, to be fair, this was also the case with just expanding the dependency version to include lit 2 even as a non-peer regular dependency as npm was able to dedupe it then too.

### DIFF
--- a/.changeset/healthy-dingos-pay.md
+++ b/.changeset/healthy-dingos-pay.md
@@ -1,0 +1,5 @@
+---
+'@lit/localize': patch
+---
+
+Loosen type for the `@localized()` decorator to accept cross version `ReactiveElement`s. Also remove dependency on `@lit/reactive-element` as it is already covered by the `lit` dependency.
--- a/.changeset/healthy-dingos-pay.md
+++ b/.changeset/healthy-dingos-pay.md
@@ -2,4 +2,4 @@
 '@lit/localize': patch
 ---
 
-Loosen type for the `@localized()` decorator to accept cross version `ReactiveElement`s. Also remove dependency on `@lit/reactive-element` as it is already covered by the `lit` dependency.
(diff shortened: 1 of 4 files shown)


.changeset/healthy-dingos-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/localize': patch
+---
+
+Loosen type for the `@localized()` decorator to accept cross version `ReactiveElement`s. Also remove dependency on `@lit/reactive-element` as it is already covered by the `lit` dependency.
```
AndrewJakubowicz, line 5: Nit: Worth mentioning that it'll now work with any `ReactiveControllerHost`?

AndrewJakubowicz (approved):
Awesome!

.changeset/healthy-dingos-pay.md
augustjk: Thank you, excellent suggestion! Made it more descriptive as well as what situation it fixes.

PR #4329, opened 2023-10-26
[context] Only depend on `@lit/reactive-element`

augustjk:
We had both `lit` and `@lit/reactive-element` listed as dependency but really all we need for the implementation is `@lit/reactive-element` for the types. This would be a good candidate to consider making it a peer dependency as well.

`lit` is moved to dev dependency as it is used for testing.


justinfagnani (approved):

AndrewJakubowicz (approved):
Awesome!

PR #4336, opened 2023-10-27
[reactive-element] document why experimental decorators need to read auto-accessors late

AndrewJakubowicz:
Documentation to address https://github.com/lit/rfcs/pull/20#discussion_r1372449696

This ensures we have a link somewhere close to the logic that references the relevant open issues.

This is a low risk, documentation only change.


augustjk (approved):
Awesome, thank you!

PR #4337, opened 2023-10-27
[infra] Re-enable wireit github cache and make tests-local parallel 2

augustjk:
Full run of tests-local is taking ~28 minutes.

Wireit github caching was skipped to see if there was erroneous caching of artifacts but we're still seeing test build failures with not finding type defs and such so these have not been helping against that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       WIREIT_LOGGER: quiet-ci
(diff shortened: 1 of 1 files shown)


justinfagnani (approved):

AndrewJakubowicz (approved):

augustjk:
This probably shouldn't be merged. Seems like the build step can be successful without having all the output (like test files) and that faulty result can get cached. The underlying issue needs to be resolved before we re-enable github caching.

PR #4287, opened 2023-10-12
[context] Broaden the host type of context controllers to be compatible with Lit 2

justinfagnani:
This adds `lit@^2.7.5` and `@lit/reactive-element@^1.6.2` as dependencies of context, and builds the tests with those packages mapped in in a separate tsconfig.

That highlighted the type error in the controllers, which I fixed by changing the host type to `ReactiveControllerHost & HTMLElement`.

PR #4339, opened 2023-10-27
[infra] Don't build lit 2 by default in context package

justinfagnani:
I missed this in this commit: https://github.com/lit/lit/pull/4287/commits/366e6b8525993d70a75154f5fa190ea2d8a5398d


augustjk (approved):

rictic:
Change description says "don't install" but this means "don't build"

PR #4340, opened 2023-10-27
[context, react] Update dependency on graduated package to `^1.0.0`

augustjk:
Our original intention was to have the labs packages automatically track the first major version of the graduated package so it'll automatically receive updates. This got erroneously pinned to the exact version for initial release as we were moving/exiting pre-release mode.

I also took this opportunity to add `^` dependencies in examples/ projects so they don't get a version bump every release.


AndrewJakubowicz (approved):
Awesome dependency gardening and improving!

Issue #4332, opened 2023-10-26
Add dev mode warning when static values are detected without using static `html` tag.


AndrewJakubowicz:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

It's easy to accidentally use the wrong `html` tagged template literal when reaching for static directives.

E.g.

```ts
import { html, render } from 'lit'; // This should be the html imported from 'lit/static-html.js'
import { unsafeStatic } from 'lit/static-html.js';

// Does not throw warning/error, and renders `attribute="[object Object]"`
render(html`<h1 attribute="${unsafeStatic('test')}">test</h1>`, document.body);
```

The above example ends up rendering  `attribute="[object Object]"`, when it should throw a dev mode warning.

### Alternatives and Workarounds

Correctly use `html` TTL imported from static when using static values in your template.

E.g.

```ts
import { render } from 'lit'; // This should be the html imported from 'lit/static-html.js'
import { html, unsafeStatic } from 'lit/static-html.js';

// Renders correctly.
render(html`<h1 attribute="${unsafeStatic('test')}">test</h1>`, document.body);
```

PR #4345, opened 2023-10-28
[lit-html] Add dev mode warning when static values are detected in non static templates

AndrewJakubowicz:
Fixes https://github.com/lit/lit/issues/4332

### Why

Given:

```
import { html, render } from 'lit'; // This should be the html imported from 'lit/static-html.js'
import { unsafeStatic } from 'lit/static-html.js';

// Does not throw warning/error, and renders `attribute="[object Object]"`
render(html`<h1 attribute="${unsafeStatic('test')}">test</h1>`, document.body);
```

The example ends up rendering attribute="[object Object]", when it should throw a dev mode warning.

### Fix

Check that values do not have `_$litStatic$` and raise a dev mode warning if the key is present.

New warning is:

```
Static values 'literal' or 'unsafeStatic' cannot be used as values to non-static templates.
Please use the static 'html' template. See https://lit.dev/docs/templates/expressions/#static-expressions
```

### DIFF
--- a/.changeset/six-planets-notice.md
+++ b/.changeset/six-planets-notice.md
@@ -1,0 +1,6 @@
+---
+'lit': patch
+'lit-html': patch
+---
+
+Add a dev mode warning if a static value such as `literal` or `unsafeStatic` is detected within the non-static `html` template. These should only be used with the static `html` template imported from `lit-html/static.js` or `lit/static-html.js`.
--- a/.changeset/six-planets-notice.md
+++ b/.changeset/six-planets-notice.md
@@ -3,4 +3,4 @@
 'lit-html': patch
 ---
 
-Add a dev mode warning if a static value such as `literal` or `unsafeStatic` is detected within the non-static `html` template. These should only be used with the static `html` template imported from `lit-html/static.js` or `lit/static-html.js`.
+Add a dev mode warning if a static value such as `literal` or `unsafeStatic` is detected within the non-static `html` tag function. These should only be used with the static `html` tag function imported from `lit-html/static.js` or `lit/static-html.js`.
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -519,8 +519,8 @@ const tag =
         values.some((val) => (val as {_$litStatic$: unknown})?.['_$litStatic$'])
       ) {
         console.warn(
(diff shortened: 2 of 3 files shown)


justinfagnani:
So, more that I think about this some more, I'm not sure if this is the right check. Static values are just strings that are tagged so we can differentiate them in static templates. They actually _could_ be usable in regular templates if used in allowed binding positions. So checks for disallowed binding positions should suffice. In those we could suggest static templates.

The question is if we need to do anything to support static values, it of doing so would be confusing.

@rictic ?

.changeset/six-planets-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+'lit': patch
+'lit-html': patch
+---
+
+Add a dev mode warning if a static value such as `literal` or `unsafeStatic` is detected within the non-static `html` template. These should only be used with the static `html` template imported from `lit-html/static.js` or `lit/static-html.js`.
```
augustjk, line 6:
```suggestion
Add a dev mode warning if a static value such as `literal` or `unsafeStatic` is detected within the non-static `html` tag function. These should only be used with the static `html` tag function imported from `lit-html/static.js` or `lit/static-html.js`.
```

packages/lit-html/src/lit-html.ts
```diff
@@ -511,6 +511,19 @@ const tag =
...
+      // Import static-html.js results in a circular dependency which g3 doesn't
+      // handle. Instead we know that static values must have the field
+      // `_$litStatic$`.
+      if (
+        values.some((val) => (val as {_$litStatic$: unknown})?.['_$litStatic$'])
+      ) {
+        console.warn(
+          `Static values 'literal' or 'unsafeStatic' cannot be used as values to non-static templates.\n` +
+            `Please use the static 'html' template. See https://lit.dev/docs/templates/expressions/#static-expressions`
+        );
```
augustjk, line 524: I wonder if the existing `issueWarning()` infrastructure can be used instead of a direct `console.warn` so we don't do multiple warnings if there are multiple static values found.

AndrewJakubowicz: 100% agree. good catch!

augustjk (approved):

justinfagnani (approved):

PR #4348, opened 2023-10-30
[labs/compiler] Fix elementParts being compiled as attributeParts

AndrewJakubowicz:
### Context

This PR fixes two issues.

1. The compiled lit_html tests were not running on CI. I only discovered the error via local testing.
2. Element parts were not being handled accurately.

#### Element part issue

Element parts were being ordered by `realName`, which would prioritize ordering attribute parts first, followed by element parts.
This is also not aligned with how lit-html detects the parts.

Fix is to look at attribute names to detect element vs attribute parts!

### Test plan

Test with golden test and lit_html tests locally.
Then also added the compiled lit_html tests to test config so now it should run in CI.

### DIFF
--- a/packages/labs/compiler/src/lib/template-transform.ts
+++ b/packages/labs/compiler/src/lib/template-transform.ts
@@ -320,7 +320,7 @@ class CompiledTemplatePass {
                 attributesToRemove.add(attr);
                 if (isAttributePart) {
                   const realName = attrNames[attrNameIndex++];
-                  if (!realName) {
+                  if (realName === undefined) {
                     throw new Error(
                       `Internal Error: realName is not defined. Please file an issue at https://github.com/lit/lit/issues/new/choose`
                     );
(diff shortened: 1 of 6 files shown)


justinfagnani (approved):

packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -314,13 +314,17 @@ class CompiledTemplatePass {
...
-              ) {
+              const isAttributePart = attr.name.endsWith(boundAttributeSuffix);
+              const isElementPart = attr.name.startsWith(marker);
+              if (isAttributePart || isElementPart) {
                 attributesToRemove.add(attr);
-                const realName = attrNames[attrNameIndex++];
-                if (realName !== undefined) {
+                if (isAttributePart) {
+                  const realName = attrNames[attrNameIndex++];
+                  if (!realName) {
```
augustjk, line 323: Nit: on behalf of Justin
```suggestion
                  if (realName === undefined) {
```

augustjk (approved):

packages/labs/compiler/src/lib/template-transform.ts
AndrewJakubowicz: Appreciated!

Issue #4290, opened 2023-10-13
[labs/nextjs] Option to add template-shadowroot ponyfill


augustjk:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Other/unknown (please mention in description)

### Description

Currently labs/nextjs doesn't do anything  with regards to template-shadowroot ponyfill so browsers that don't support declarative shadow DOM won't work unless users add and apply the ponyfill themselves.

It would be nice to do this for them.

### Alternatives and Workarounds

Adding something like below
```js
// _app.tsx

if (!isServer) {
  if (!HTMLTemplateElement.prototype.hasOwnProperty('shadowRootMode')) {
    console.log('DSD not supported');
    const {hydrateShadowRoots} = await import(
      '@webcomponents/template-shadowroot'
    );
    hydrateShadowRoots(document.body);
  }
}
```

This could be added into or along side the client version of `@lit-labs/ssr-react/enable-lit-ssr.js` import apply the ponyfill.

Top level await was not a supported feature in Next.js until recently. We may want to just do an import up top to avoid the incompatibility or the wait time for the extra request and code execution.

PR #4354, opened 2023-11-01
[labs/nextjs] Add option to apply Declarative Shadow DOM polyfill

augustjk:
Fixes https://github.com/lit/lit/issues/4290

The solution here proposes a new option to pass when creating the plugin which will include a script that applies the `@webcomponents/template-shadowroot` ponyfill to the client bundle.

I am not happy with the option name `includeDSDPolyfill` but couldn't think of anything better at the moment.

This was tested manually by using our `/examples/nextjs` project in the monorepo. With the option disabled, the example errors when viewed in Firefox due to the presence of the `<template>` elements. With the option enabled, it no longer fails and gets hydrated correctly.

TODO:
- [x] Update README documenting this


packages/labs/nextjs/src/index.ts
```diff
@@ -6,16 +6,29 @@
 
 import type {NextConfig} from 'next';
 
-// TODO(augustjk) Consider adding option for adding template shadowroot polyfill
-// automatically
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-interface LitSsrPluginOptions {}
+interface LitSsrPluginOptions {
+  includeDSDPolyfill?: boolean;
```
AndrewJakubowicz, line 10: Could this be documented, so when someone is investigating the options bag type, they get documentation for this config.

packages/labs/nextjs/src/lib/apply-dsd-polyfill.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {hydrateShadowRoots} from '@webcomponents/template-shadowroot';
+
+if (!HTMLTemplateElement.prototype.hasOwnProperty('shadowRootMode')) {
```
AndrewJakubowicz, line 11: Should this be conditional, or always applied?

AndrewJakubowicz (commented):
Awesome!

packages/labs/nextjs/src/lib/apply-dsd-polyfill.ts
augustjk: This is the feature detection so the ponyfill is only applied for browsers that don't support it. https://developer.chrome.com/articles/declarative-shadow-dom/#feature-detection-and-browser-support

packages/labs/nextjs/src/index.ts
augustjk: Yes, good call. I'd like to do that when we settle on a name.
Apparently I'm supposed to do strict camel casing https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined so `includeDsdPolyfill`.

 @michaelwarren1106 suggested on discord maybe spelling it out like `polyfillDeclarativeShadowDom` which uses "polyfill" as a verb rather than doing "include". Naming is very hard.

justinfagnani: `addDeclarativeShadowDomPolyfill` isn't that bad really, IMO. I personally rather config be very explicit for future readers who didn't set it up in the first place.

AndrewJakubowicz (approved):
Awesome!

Is it worth documenting anywhere what happens if someone has already manually loaded the ponyfill?

augustjk:
> Is it worth documenting anywhere what happens if someone has already manually loaded the ponyfill?

Fair point. I add a note to the changelog for any users doing their own. I think with the option being documented in the README, new users who want their own polyfilling will know to set it to false.

PR #4355, opened 2023-11-01
[labs/virtualizer] Fix new failure case for #3815

graynorton:
This PR addresses a new failure case for #3815 (which I've just reopened). PR includes an additional test, along with the fix.

### DIFF
--- a/.changeset/six-fishes-clean.md
+++ b/.changeset/six-fishes-clean.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/virtualizer': patch
+---
+
+Fix masonry layout bug (new failure case for #3815)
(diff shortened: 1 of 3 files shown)


augustjk (approved):
Honestly I'm not really sure about how the range maps work here. I'm guessing it's storing indices of items that should be viewable, by the key which is like some cross section of possible scroll positions normalized with some granularity?

Based on issue description, it sounds like the change to making it an iteration will catch the first/last item to show landing somewhere in the middle of the viewable range and the test seems good.

graynorton:
Yes, @augustjk, that's exactly right. Here's a quick set of slides that describes in slightly more detail:

https://docs.google.com/presentation/d/1-hBY_Z6BVPlqHXRRdEsC5US5iqyR_7SS2Zkqe56Ye3g/edit?usp=sharing

Issue #4363, opened 2023-11-06
[labs/ssr-react] Wrong hydration ordering  for nested custom elements


augustjk:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

When SSR-ing nested custom elements, the nested ones should get `defer-hydration` attribute so that it won't hydrate until its parents hydrate.

Due to a bug in the server rendering code used by `labs/ssr-react`, nested custom elements were not getting that attribute.

Fix is to push into the `customElementHostStack` here in addition to the `customElementInstanceStack`:
https://github.com/lit/lit/blob/b2030b44e75d32b2e8433a8977848a672c995014/packages/labs/ssr-react/src/lib/node/render-custom-element.ts#L71

The host stack length is used for checking whether `defer-hydration` should be emitted.

PR #4364, opened 2023-11-06
[labs/ssr-react] Fix hydration ordering for nested custom elements

augustjk:
Fixes https://github.com/lit/lit/issues/4363

I also kept the changes I made to the `nextjs` example to reproduce the error.. It might not be really necessary but also doesn't hurt to keep it around.

Testing for this looks specifically for `defer-hydration` attribute on SSRed string on a child custom element that is rendered by a parent custom element.

The test we had formerly called "nested element" actually tests that a custom element that is  React child in the provided JSX renders properly so I've renamed to "custom element child" instead.

### DIFF
--- a/.changeset/brave-moles-check.md
+++ b/.changeset/brave-moles-check.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr-react': patch
+---
+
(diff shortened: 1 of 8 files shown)


AndrewJakubowicz (approved):
Beautiful!

PR #4367, opened 2023-11-06
[infra] Fix sizecheck actions to save PR number for report

augustjk:
I made an incorrect assumption in https://github.com/lit/lit/pull/4323

The action used for create comments https://github.com/thollander/actions-comment-pull-request does not work from a `workflow_run` triggered action as it can't pull the PR to comment on from the context of the workflow.

This is apparently not straight forward to do purely within GitHub actions.
https://github.com/orgs/community/discussions/45415
https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow

Even GHA's own doc shows using an artifact to pass the PR number to another workflow: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow

I cobbled together a solution from digging through above example and answers to save the PR number from the triggering workflow into an artifact and using it in the reporting workflow. Previously we would only have uploaded an artifact on size check failure which generates a markdown message, but now we just do it for both success and failure.

Side note: My original assumption that it should work was based on benchmarks report workflow working fine in a `workflow_run` triggered action but turns out they have bespoke logic to find the PR to comment on in that case https://github.com/andrewiggins/tachometer-reporter-action/blob/fef5976dd7c57c673de2391f1c2cfbfab7a9c415/src/utils/github.js#L8-L53

An alternative solution, instead of using an artifact, would be to do a logic similar to one done in the `tachometer-reporter-action` to write a GHA script within this workflow to use the commit head and query for a matching PR, but since we were already using an artifact, figured this was the easier solution.

### DIFF
--- a/.github/workflows/sizecheck.yaml
+++ b/.github/workflows/sizecheck.yaml
@@ -21,12 +21,21 @@ jobs:
       - name: NPM install
         run: npm ci
 
+      # This is needed to be able to read it from sizecheck-report.yaml to
+      # comment on the correct PR
+      - name: Save PR number
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: echo $PR_NUMBER > ./scripts/pr_number.txt
+
       - name: Generate size report
         run: npm run benchmark:size
 
-      - name: Upload failure artifact
-        if: ${{ failure() }}
+      - name: Upload artifacts
+        if: ${{ success() || failure() }}
         uses: actions/upload-artifact@v3
         with:
           name: sizecheck
-          path: scripts/check-size-out.md
+          path: |
+            scripts/check-size-out.md
+            sdripts/pr_number.txt
--- a/.github/workflows/sizecheck.yaml
+++ b/.github/workflows/sizecheck.yaml
@@ -38,4 +38,4 @@ jobs:
           name: sizecheck
           path: |
             scripts/check-size-out.md
-            sdripts/pr_number.txt
+            scripts/pr_number.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/sizecheck.yaml
```diff
@@ -21,12 +21,21 @@ jobs:
...
-        if: ${{ failure() }}
+      - name: Upload artifacts
+        if: ${{ success() || failure() }}
         uses: actions/upload-artifact@v3
         with:
           name: sizecheck
-          path: scripts/check-size-out.md
+          path: |
+            scripts/check-size-out.md
+            sdripts/pr_number.txt
```
AndrewJakubowicz, line 41: Nit: `sdripts` -> `scripts`

AndrewJakubowicz (approved):
It's quite difficult to test GitHub Actions. Thank you for doing this!

.github/workflows/sizecheck.yaml
augustjk: Oops.. I totally caught myself having written `sripts/pr_number.txt` there but typoed the typo fix 🤦 .

augustjk:
> It's quite difficult to test GitHub Actions. Thank you for doing this!

Indeed... especially the `workflow_run` triggered ones since they only work once in `main` branch. It should be low risk though, I'll test this against one of the open PRs to see if this works.

augustjk:
This looks to have been successful https://github.com/lit/lit/actions/runs/6778067338/job/18422880623

PR #4371, opened 2023-11-07
Allow ContextProvider to be used with HTMLElement, fixes #4361

### LINKED ISSUE #4361: [context] Global Context
### Should this be an RFC?

Context (@lit/context)

### Description

It seems like it would be quite straightforward to extend the Context API so that a truly global context provider can be set up on document.

Something like this:
```js
class AppComponent1 extends LitElement {
    // note use of 'document' instead of 'this'
    _userProvider = new ContextProvider(document, userContext); 
    // or perhaps a new class:
    _userProvider = new GlobalContextProvider(document, userContext); 
}
```

And then components can consume a global context value in exactly the same way as a normal one:
```js
class AppComponent2 extends LitElement {
    static properties = {
        user: {}
    }
    _userConsumer = new ContextConsumer(this, userContext, value => {
        this._user = value;
    }, true);
}
```

My reasoning for this is that I have a largely static page, with different `LitElement` components in different areas. For example a navigation bar (which needs to know whether the user is logged in or not to show login/logout buttons), and a small section of interactive content further down the page, which depends on the login.

It doesn't really make sense to turn the whole largely static page into a `LitElement`, but it would be nice to have a standard way to share some context between all components on the page. Since context is identified by the identity of the context object, there doesn't seem to be any danger of breaking the isolation between separate components that do not know about the context, and this should be completely backwards compatible with the existing context API.

It's possible to do this outside of Lit, but would be much neater to re-use the internal `ValueNotifier` class, or just modify the existing `ContextProvider` class to accept the document, and it seems like something that would be generally useful.

If this is something which would be welcome, then I'm happy to have a stab at implementing it.

(I don't think this is a substantial change, but correct me if I'm wrong)

autopulated:
Based on https://github.com/lit/lit/issues/4361#issuecomment-1797106854, allow ContextProvider to be created on raw `HTMLElement`s.

I'm new to typescript, and also got a bit lost trying to understand the build and release process, so if someone could point me in the right direction for adding tests to this, I'd appreciate it!


packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -82,7 +82,12 @@ export class ContextProvider<
       this.context = contextOrOptions as T;
     }
     this.attachListeners();
-    this.host.addController(this);
+    // We may be constructed with a plain DOM element to provide context to a
+    // whole DOM tree of elements, in this case the host DOM element doesn't
+    // need know about this controller.
+    if ((host as ReactiveControllerHost).addController !== undefined) {
+      this.host.addController(this);
+    }
```
augustjk, line 90: Optional chaining is shorter here.
```suggestion
    this.host.addController?.(this);
```

The comment should be moved to the jsdoc above the `ContextProvider` and more user facing.

Something like..
> The controller may also be attached to any HTML element in which case it's up to the user to call `hostConnected()` when attached to the DOM. This is done automatically for any custom elements implementing `ReactiveControllerHost`.

augustjk (changes requested):
Thanks for the PR!

In additional to the optional chaining suggestion, the `ReactiveElementHost` type here
https://github.com/lit/lit/blob/449bc2815593e8150b0737ed0190ba2c9843ba66/packages/context/src/lib/controllers/context-provider.ts#L45

should be changed to
```ts
type ReactiveElementHost = Partial<ReactiveControllerHost> & HTMLElement;
```
to make `ReactiveControllerHost` properties optional.

As for testing, a separate suite similar to this one
https://github.com/lit/lit/blob/449bc2815593e8150b0737ed0190ba2c9843ba66/packages/context/src/test/context-request_test.ts#L58
probably would work. The consumer can be shared but the setup can attach the controller directly to a container div.

README should also be updated here https://github.com/lit/lit/blob/main/packages/context/README.md#providing-a-context
to document the additional use case.

The doc site https://lit.dev/docs/data/context/ should also be updated but that can be separate task.

cc @sorvell for any additional thoughts?

augustjk (approved):
Thank you! I fixed the initial value in the test and added a changeset file for the version changelog.

.changeset/smooth-countries-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/context': minor
```
AndrewJakubowicz, line 2: Just clarifying for myself that we're not updating `@lit-labs/context` anymore.

packages/context/README.md
```diff
@@ -133,6 +133,30 @@ export class MyApp extends LitElement {
...
+      log: (msg) => {
+        console.log(`[global] ${msg}`);
+      },
+    },
+);
+```
+
+If the provider is being added when there is already a consumer registered with
+a parent of the specified element or with a `ContextRoot`, then
+`.hostConnected()` must be called on the provider after creating it.
```
AndrewJakubowicz, line 158: Nit: Should there be a slight addition of why it must be called.
E.g., to ensure any consumers are updated to use their nearest provider correctly?

[non blocking comment]

AndrewJakubowicz (approved):
Awesome!

.changeset/smooth-countries-work.md
augustjk: Correct. Latest `@lit-labs/context` is just a re-export of `@lit/context@1`.

packages/context/README.md
augustjk: Yeah I was debating if we should add some more there. Good call out.

augustjk:
Thank you @autopulated for the contribution!

klasjersevi:
Very nice addition! 👍 

I did something similar in my project and I'm looking forward to fully transition to this instead, but there is one important thing missing imo; the ability to unregister the event listeners from the HTMLElement. This is especially important if you want to use context listeners on rather persistent elements like document.body or document.documentElement (which usually never gets removed or replaced). Otherwise, the listeners will most likely live forever.

```js
public detachListeners() {
  this.host.removeEventListener('context-request', this.onContextRequest);
  this.host.removeEventListener('context-provider', this.onProviderRequest);
}
```

Please also consider exporting the `ReactiveElementHost` type.

As for now, I have added the functionality above in my project by extending the class.

PR #4372, opened 2023-11-07
[chore] Update changeset for template result type change

augustjk:
We should add `lit` package as well when bumping `lit-html`.


AndrewJakubowicz (approved):
Thanks!

PR #4375, opened 2023-11-08
[labs/virtualizer] Fix #4346

### LINKED ISSUE #4346: [labs/virtualizer] offsetParent issue report – virtualizer not working

### Description

User reported that virtualizer stopped working after the introduction of: https://github.com/lit/lit/commit/f84963d80f890f6ddaa46720622dcc366663cb6a#diff-1e2b7f3328e9761f16f9252acbf58364a193256cd562b17c47522317f8520b22R556-R564

Report states that after this change, Virtualizer stopped working within a `position: fixed;` container (within a `mwc-dialog` specifically).
Theory is that `offsetParent` is not precise enough to check whether an element is displayed or not.

### Reproduction

Repro: https://lit.dev/playground/#gist=a7c2a4f12422be727d0a6464482f8e0a

It looks like `offsetParent` with fixed element within shadowDOM doesn't work. In the repro, virtualizer works if the fixed element is in the light DOM.

### Workaround

A possible workaround is to use [the offsetParent](https://github.com/josepharhar/offsetparent-polyfills/blob/main/offsetParent-polyfill.js) polyfill.

### Is this a regression?

### Affected versions

Introduced in https://github.com/lit/lit/commit/f84963d80f890f6ddaa46720622dcc366663cb6a

No repro yet.

graynorton:
Fix #4346 by not relying on `offsetParent` to detect the case where a virtualizer is hidden because itself or one of its ancestors is `display: none`.

It turns out that, due to some Shadow-DOM-specific behavior (see [comment](https://github.com/lit/lit/issues/4346#issuecomment-1785809477) for background reading), `offsetParent` returned `null` in cases where a (visible) virtualizer was slotted into a shadow tree beneath a `position: fixed` ancestor, making this an unreliable check.

Instead of using the `offsetParent` check, we will rely on existing Virtualizer functionality to avoid rendering children when a virtualizer is hidden. In determining which children to render, a virtualizer calculates the size of the on-screen region it needs to fill, taking into account its position in the document and the current scroll position. This calculation already returns `width` and `height` of `0` when the virtualizer is hidden, which should be a signal not to render any children. However, for reasons not documented and long forgotten, we have been forcing the `width` and `height` passed to the layout to be at least `1px` each. I am quite confident that this was a hacky workaround for some no-longer-existent issue; I know that it predates much significant work on Virtualizer's rendering pipeline and that removing it doesn't cause any tests to fail.

Added and updated some relevant tests.

### DIFF
--- a/packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
@@ -4,11 +4,13 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {LitElement, html} from 'lit';
+import {customElement} from 'lit/decorators.js';
 import {array, ignoreBenignErrors} from '../helpers.js';
 import {LitVirtualizer} from '../../lit-virtualizer.js';
 import {grid} from '../../layouts/grid.js';
 import {styleMap} from 'lit/directives/style-map.js';
-import {expect, html, fixture} from '@open-wc/testing';
+import {expect, html as testingHtml, fixture} from '@open-wc/testing';
 
 interface NamedItem {
   name: string;
@@ -42,7 +44,7 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor', 
     // We use the grid layout here because it conveniently will not
     // render any items if it calculates that it doesn't have sufficient
     // space, making our testing job easier.
-    const container = await fixture(html`
+    const container = await fixture(testingHtml`
       <div id="container" style=${styleMap(containerStyles)}>
         <div id="scroller" style=${styleMap(scrollerStyles)}>
           <lit-virtualizer
@@ -87,7 +89,7 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor', 
       overflow: 'auto',
     };
 
-    const container = await fixture(html`
+    const container = await fixture(testingHtml`
       <div id="container" style=${styleMap(containerStyles)}>
         <div id="scroller" style=${styleMap(scrollerStyles)}>
           <lit-virtualizer
@@ -115,5 +117,59 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor', 
     scroller.scrollTo(0, scroller.scrollHeight);
     await virtualizer.layoutComplete;
     expect(virtualizer.textContent).to.contain('Item 99');
+  });
+});
+
+@customElement('simple-dialog')
+export class SimpleDialog extends LitElement {
+  render() {
+    return html`<div style="position: fixed; width: 200px;">
+      <slot></slot>
+    </div> `;
+  }
+}
+
+describe('Virtualizer renders properly when it is slotted into a position: fixed ancestor', () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  // Regression test for https://github.com/lit/lit/issues/4346
+  it('should render children', async () => {
+    const container = await fixture(testingHtml`
+      <simple-dialog>
+        <lit-virtualizer
+          .items=${_100Items}
+          .renderItem=${({name}: NamedItem) => html`<p>${name}</p>`}
+        ></lit-virtualizer>
+      </simple-dialog>
+    `);
+
+    const virtualizer = container.querySelector('lit-virtualizer')!;
+    expect(virtualizer).to.be.instanceOf(LitVirtualizer);
+
+    // In practice, we'll time out if we fail here because the `layoutComplete`
+    // promise will never be fulfilled.
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 0');
+  });
+});
+
+describe('Virtualizer renders properly when it is position: fixed', () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  it('should render children', async () => {
+    const virtualizer = (await fixture(testingHtml`
+      <lit-virtualizer
+        style="position: fixed; width: 200px;"
+        .items=${_100Items}
+        .renderItem=${({name}: NamedItem) => html`<p>${name}</p>`}
+      ></lit-virtualizer>
+    `)) as LitVirtualizer;
+
+    expect(virtualizer).to.be.instanceOf(LitVirtualizer);
+
+    // In practice, we'll time out if we fail here because the `layoutComplete`
+    // promise will never be fulfilled.
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 0');
   });
 });
--- a/packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
+++ b/packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
@@ -1,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ignoreBenignErrors} from '../helpers.js';
+import {html, LitElement} from 'lit';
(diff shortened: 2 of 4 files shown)


packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -4,11 +4,13 @@
...
  */
 
+import {LitElement, html} from 'lit';
+import {customElement} from 'lit/decorators.js';
 import {array, ignoreBenignErrors} from '../helpers.js';
 import {LitVirtualizer} from '../../lit-virtualizer.js';
 import {grid} from '../../layouts/grid.js';
 import {styleMap} from 'lit/directives/style-map.js';
-import {expect, html, fixture} from '@open-wc/testing';
+import {expect, html as testingHtml, fixture} from '@open-wc/testing';
```
augustjk, line 13:
```suggestion
import {expect, fixture} from '@open-wc/testing';
```

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -42,7 +44,7 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor',
     // We use the grid layout here because it conveniently will not
     // render any items if it calculates that it doesn't have sufficient
     // space, making our testing job easier.
-    const container = await fixture(html`
+    const container = await fixture(testingHtml`
```
augustjk, line 47:
```suggestion
    const container = await fixture(html`
```

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -87,7 +89,7 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor',
       overflow: 'auto',
     };
 
-    const container = await fixture(html`
+    const container = await fixture(testingHtml`
```
augustjk, line 92:
```suggestion
    const container = await fixture(html`
```

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -117,3 +119,57 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor',
...
+    </div> `;
+  }
+}
+
+describe('Virtualizer renders properly when it is slotted into a position: fixed ancestor', () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  // Regression test for https://github.com/lit/lit/issues/4346
+  it('should render children', async () => {
+    const container = await fixture(testingHtml`
```
augustjk, line 137:
```suggestion
    const container = await fixture(html`
```

packages/labs/virtualizer/src/test/scenarios/fixed-position-ancestor.test.ts
```diff
@@ -117,3 +119,57 @@ describe('Virtualizer behaves properly when it has a position: fixed ancestor',
...
+    await virtualizer.layoutComplete;
+    expect(virtualizer.textContent).to.contain('Item 0');
+  });
+});
+
+describe('Virtualizer renders properly when it is position: fixed', () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  it('should render children', async () => {
+    const virtualizer = (await fixture(testingHtml`
```
augustjk, line 160:
```suggestion
    const virtualizer = (await fixture(html`
```

packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ignoreBenignErrors} from '../helpers.js';
+import {html, LitElement} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import '../../lit-virtualizer.js';
+import {expect, html as testingHtml, fixture} from '@open-wc/testing';
```
augustjk, line 11:
```suggestion
import {expect, fixture} from '@open-wc/testing';
```

packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    `;
+  }
+}
+
+describe("Don't render any children if the virtualizer is hidden", () => {
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  it('should not render any children when initially hidden', async () => {
+    const el = await fixture<VirtualizerHider>(
+      testingHtml`
```
augustjk, line 48:
```suggestion
      html`
```

packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      `
+    );
+    const virtualizer = el.shadowRoot!.querySelector('lit-virtualizer')!;
+    await new Promise((resolve) => setTimeout(resolve, 500));
+    expect(virtualizer.children.length).to.equal(0);
+  });
+
+  it('should not render any children when subsequently hidden', async () => {
+    const el = await fixture<VirtualizerHider>(
+      testingHtml`
```
augustjk, line 59:
```suggestion
      html`
```

AndrewJakubowicz (approved):
This looks great to me. I'm also running internal tests via cl/580594865 (as additional verification).
Thanks for adding those awesome tests!

packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  ignoreBenignErrors(beforeEach, afterEach);
+
+  it('should not render any children when initially hidden', async () => {
+    const el = await fixture<VirtualizerHider>(
+      testingHtml`
+        <virtualizer-hider hidden></virtualizer-hider>
+      `
+    );
+    const virtualizer = el.shadowRoot!.querySelector('lit-virtualizer')!;
+    await new Promise((resolve) => setTimeout(resolve, 500));
```
augustjk, line 53: Not a huge fan of the timeout here but I understand we can't use `layoutComplete` when nothing is rendered initially. We did talk about making a change to this in the past.. perhaps that can be revisited at some point.

augustjk (approved):
LGTM. Just a nit about aliasing the `html` from `@open-wc/testing` that is unnecessary.

Thank you for adding a regression test for `display: none` issue as well!

packages/labs/virtualizer/src/test/scenarios/hidden.test.ts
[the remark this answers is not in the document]
AndrewJakubowicz: Ah, saw Augustine's comment https://github.com/lit/lit/pull/4375#discussion_r1387102543.
Resolving.

AndrewJakubowicz: This was a very helpful review comment, and may be worth adding as an inline comment.

augustjk: Filed https://github.com/lit/lit/issues/4376

augustjk: In case it's easier to resolve these by accepting suggestions
```suggestion
    // We can't await on layoutComplete when first render is empty
    // See https://github.com/lit/lit/issues/4376
    await new Promise((resolve) => setTimeout(resolve, 500));
```

Issue #4357, opened 2023-11-02
[labs/ssr] opt-in de-dupe styles in declarative shadow DOM


AndrewJakubowicz:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

SSR (@lit-labs/ssr)

### Description

For reference, a platform solution: https://github.com/WICG/webcomponents/issues/939

Currently, each instance of a component requires all its styles to be inlined into the DSD. This means if you have the same component on the page multiple times, you repeat the styles per instance. This leads to duplicated styles being sent down the network.

Before the platform proposal exists, we should offer a solution.

### Proposed solution

This is not a full design, just some rough notes initially.
We can add an de-dupe styles option to Lit SSR.

SSR could inline a JavaScript utility, such as a custom element definition into the top of the page. E.g. `<style-module>`.
Then when emitting styles in DSD, wrap the first instance of a DSD style in `<style-module>`:

```
<style-module style-id="abc">
  <style>
    ...
   </style>
</style-module>
```

If these styles will be duplicated, any future emit of the styles contains just the `<style-module style-id="abc">` without any style contents.

This custom element fetches the styles from a data store, and adds them to the DSD.

### Considerations

This should be an option that can be turned on and off, because this solution requires JavaScript be run on the page – but the benefit is a reduced payload. For clients who run JavaScript, this would allow for a reduced network payload.

### Alternatives and Workarounds

Keep duplicating styles until browser platform adds support for this.

bivens-dev:
Just spitballing a potentially JS free solution, could the same thing be achieved using link tags the components styles? 

Is it possible to reference a style tag on the page using ids and anchor links as a place extract common styles to.

justinfagnani:
One way to do this is to add `<lit-adoptable-style>` and `<lit-adopt-styles>`tags. The former would register a stylesheet in a global cache, the later would read from the cache and add styles to its root node's `adoptedStyleSheets`.

There's an [issue open](https://github.com/webcomponents-cg/community-protocols/issues/72#issuecomment-2835940612) in WCCG for a community protocol around this, but I don't think it's necessary. We can just do this in Lit SSR and it will be compatible with any elements, since they won't have to know about either style-related custom element here, ie it'll work with any declarative shadow root, not just Lit-based elements.

[...]

PR #4378, opened 2023-11-09
[lit-labs/ssr] Shared and de-duplicated declarative styles utility

AndrewJakubowicz:
Issue: https://github.com/lit/lit/issues/4357

### Background

Currently, there is no way to deduplicate styles in the declarative shadow DOM. This PR provides an opt-in mechanism (if JavaScript is enabled), to deduplicates declarative shadow DOM styles from LitElement's static `styles` array.

### Design

During SSR, if this feature is enabled, each `css` tag present in the static styles array will be servers side rendered into a `<style>` element once. This element will be wrapped with the deduplicate element.

After the first instance of the styles have been encountered, any subsequent instances encountered during SSR rendering will only send down the deduplication element with an id attribute which references the first instance.

During parse time, the deduplication element can synchronously pull in the styles and attach them. This avoids any FOUC.

I've made it so Constructable style sheets are used if available, falling back to cloning the style element directly.

### Semantic benefits

In the case where the browser supports constructable style sheets, this utility provides semantics closer to what we would like SSR to exhibit. Shared styles will use the same underlying style sheet instance.

Thus this utility also exhibits better SSR styles semantics.

If the browser doesn't support adopting constructable style sheets, the fallback behavior is to insert a style element.

### Benchmarks

**tl;dr**: Performance impact is unclear. The most obvious win is in `gzip` and `uncompressed` network size.

This is a hard feature to measure, because it impacts network size, but also will result in changes in runtime characteristics. E.g., less style tokens are parsed, but a style element or constructable style sheet is added instead.

Below are measurements I took using `material-web.dev` as an example of a Lit SSR'd website.

| Test   |      No compression      |  gzip | brotli |
|----------|-------------|------|------|
| [button.html](https://material-web.dev/components/button/) |  993kB | 45kB | 23kB |
| button.html (with deduplicated styles) |  327kB (-67%)  | 31kB (-31%) | 23kB (-0%) |

What is the impact on browser performance timings. Note this was done locally using Chrome developer tools, and I jotted down the rough numbers after a couple refreshes. This is for ballpark figures.

| Test   |    Parse HTML time (ms)   | FCP | DCL |
|----------|-------------|---------|----|
| button catalog page |  7.6ms (self 7.0ms) | 200ms | 183ms |
| button.html (with deduplicated styles) |  10.02ms (self 5.73ms) | 225ms | 241ms |

The increased parse HTML time makes sense with deduplicated styles because we are applying the styles during parse time via the new element. This seems to slightly increase all other timings.

### Risks

Doesn't work in no JavaScript environments, this is a userland solution to the following proposal: https://github.com/WICG/webcomponents/issues/939


packages/labs/ssr/src/lib/declarative-style-dedupe.ts
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        const obs = new MutationObserver(() => {
+            obs.disconnect();
+            this.parsedCallback();
+        });
+        obs.observe(this.parentNode, {childList: true, subtree: true});
+    }
+
+    parsedCallback() {
```
justinfagnani, line 43: I don't think you need this. Just put the style tag (not the text content) into the styleCache. By the time any other style-module requests the styles they will have necessarily been fully parsed.

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
```diff
@@ -0,0 +1,93 @@
...
+        const styleEl = document.createElement('style');
+        styleEl.textContent = styleContents;
+        this.append(styleEl);
+    }
+}
+customElements.define('style-module', StyleModule);
+</script>`;
+  }
+
+  getStyleHash(styles: CSSResultOrNative[]): string {
```
justinfagnani, line 63: I don't think you need a hash, unless we're trying to make this work over multiple requests. I think you should be able to use stylesheet identity on the server and a simple counter for the id.

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -89,11 +89,32 @@ export class LitElementRenderer extends ElementRenderer {
     const styles = (this.element.constructor as typeof LitElement)
       .elementStyles;
     if (styles !== undefined && styles.length > 0) {
-      yield '<style>';
-      for (const style of styles) {
-        yield (style as CSSResult).cssText;
+      let styleHashId: string | null = null;
```
justinfagnani, line 92: I think we're going to need to iterate through all the styles and emit a separate `<style-module>` for each one. That way they dedupe to an equivalent set of CSSStyleSheet objects that the components actually use.

There's even a future build optimization we could do where we rewrite the components to load styles from style cache if we know they've been SSR'ed so that we don't include them in the JS bundle when they're already on the page.

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
```diff
@@ -0,0 +1,93 @@
...
+        // to be cached.
+        if (this.hasAttribute("need-cached-styles")) {
+            return this.parsedCallback();
+        }
+
+        const obs = new MutationObserver(() => {
+            obs.disconnect();
+            this.parsedCallback();
+        });
+        obs.observe(this.parentNode, {childList: true, subtree: true});
```
justinfagnani, line 40: to definitively determine a tag close you can inject a `<script>` tag after it that does something like `document.currentScript.previousSibling.loadStyles()`

AndrewJakubowicz: Awesome! This is super helpful. To cache the first `style` element, we need the caching to happen after the `style-module` closing tag has been parsed (thus the `style` element exists within it).

I didn't realize `connectedCallback` ran at **tag open** time and was quite perplexed when there was no `style` element anywhere.

packages/labs/ssr/src/lib/lit-element-renderer.ts
AndrewJakubowicz: Great comment! Can definitely do that.

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
AndrewJakubowicz: After reading https://github.com/lit/lit/pull/4378#discussion_r1387367228 I can see an approach where no parse timing needs to be managed.
Still, this was a great comment to learn from. Thank you!

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
AndrewJakubowicz: Yep – this simplifies the client a lot. Thank you! Will do this!

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
justinfagnani: yeah, you don't need to cache the style text upon the first style module element, just cache the element. But you do need to know when a style-module has no children so you can _fetch_ the style text. Knowing a tag will have no more children is.... hard.

But you actually don't need a separate script tag in this case - you could just use a different attribute. I believe all browsers will have parsed attributes by the time connectedCallback is called, so you can read a `load-styles="id"` attribute and then just go get the styles right then. I'm sure that'll be faster than having possibly 100s of script tags on the page.

AndrewJakubowicz:
The primary reason for making this prototype was to measure if there would be any benefit. I've now measured against a page in material-web.dev which uses Lit SSR.

Method:

I `npm pack`ed a new eleventy ssr plugin which contained the de-duping styles prototype and ran it on [material-web.dev](https://material-web.dev). I visually inspected that the styles were not duplicated and also ensured the page worked great.
Most importantly, I could inspect the file size of the generated catalog.

| Test   |      No compression      |  gzip | brotli |
|----------|-------------|------|------|
| button.html |  993kB | 45kB | 23kB |
| button.html (with style-module) |  327kB (-67%)  | 31kB (-31%) | 23kB (-0%) |


Conclusion:

Brotli is pretty fantastic and handling duplicated text content, and thus de-duping styles provides less benefit than I initially expected. The gzip benefit of 30% is compelling but I'd like feedback to see if this benefit is worth pursuing.

justinfagnani:
67% smaller is a _lot_ less to parse, regardless of compression. And it's a lot less for any downstream processing: the server, browser, any transforms or middleware. So we should at least measure first paint, but even if we didn't detect a huge improvement there I think this should still be an option because we can make it semantically correct (with some client-side changes).

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
```diff
@@ -0,0 +1,109 @@
...
+      this._cachedSS.replaceSync(this.children[0].textContent);
+      return this._cachedSS;
+    }
+
+    connectedCallback() {
+      const styleId = this.getAttribute('style-id');
+      const cachedStyleSheet = StyleModule.styleCache.get(styleId);
+      if (cachedStyleSheet == null) {
+        StyleModule.styleCache.set(styleId, this);
+        return;
```
sorvell, line 70: So, if the style is here, the element just uses it? That's incorrect since it won't be shared. Instead, the element should remove the style and adopt the sheet.

sorvell: Consider instead, putting the `<style>` before this element so that it's available at `connectedCallback` time. Then it can be processed here.

sorvell: Additionally, remove the de-duping element then the stylesheet is adopted so it doesn't pollute the element's DOM.

sorvell: This seems to work, see: https://lit.dev/playground/#gist=078a12b72391e1638f6cfb09391b7688

sorvell (changes requested):

sorvell:
IMO, the main reason to add this feature is to make the styling behave as authored, using adopted stylesheets.

The performance testing is mostly important to ensure it doesn't regress performance, but if it makes FCP faster, that's great.

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
AndrewJakubowicz: The names need to be fixed. But `cachedStyleSheet` here is referring to the `StyleModule` that contains the style definition – the first instance of StyleModule.

If there is not already a StyleModule in the cache, then we must be the first one containing a definition.

AndrewJakubowicz: I edited my comment above to strike it out. I am confused as to why the first instance with the concrete `<style>` element is wrong if it isn't adopted?

I'd love to understand your very first comment better "So, if the style is here, the element just uses it? That's incorrect since it won't be shared. Instead, the element should remove the style and adopt the sheet.".

Why is it more correct to remove the style and adopt the sheet?

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -423,6 +424,58 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+        `<!--/lit-part-->`
+    );
+  });
+
+  test('duplicated element with static styles using styles de-duplicate utility', async () => {
+    const {render, duplicatedElementWithStaticStyles} = await setup();
+    const result = await render(duplicatedElementWithStaticStyles, {
+      dedupeStyles: new DeclarativeStyleDedupeUtility(),
+    });
+    const ssrElement = `<test-static-styles><template shadowroot="open" shadowrootmode="open"><style>
```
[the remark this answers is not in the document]
AndrewJakubowicz: Addressed. I accidentally left in broken tests that I was using to see that my change made the improvement.

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -423,6 +424,58 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+  </style><!--lit-part--><!--/lit-part--></template></test-static-styles>`;
+    assert.is(
+      result,
+      `<!--lit-part KqQuRUrghX8=-->` +
+        ssrElement +
+        ssrElement +
+        `<!--/lit-part-->`
+    );
+  });
+
```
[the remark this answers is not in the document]
AndrewJakubowicz: Test has been added.

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -21,6 +22,14 @@
  */
 const emptyVmGlobal = {};
 
+/**
+ * Removes the first script tag in HTML. Useful when testing the rendered DOM
+ * if `DeclarativeStyleDedupeUtility` is being used. It injects a large script
+ * tag into the rendered result.
+ */
+const removeDedupeScriptTagHelper = (rawHtml: string) =>
+  rawHtml.replace(/<script>(.|\s)*?<\/script>/gm, '');
```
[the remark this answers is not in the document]
AndrewJakubowicz: Test only code.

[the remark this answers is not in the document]
AndrewJakubowicz: Test only code.

packages/labs/ssr/src/lib/declarative-style-dedupe.ts
AndrewJakubowicz: We resolved at Lit Eng meeting! Thank you!

AndrewJakubowicz:
@sorvell I've addressed your great feedback. This change still needs docs, however I'd love your thoughts now that the logic has been fixed and works semantically correctly.

Thank you!

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5512,6 +5513,191 @@ export const tests: {[name: string]: SSRTest} = {
...
+            return customElements.whenDefined('lit-ssr-style-dedupe');
+          },
+          html: {
+            root: `<dd-constructable-stylesheet></dd-constructable-stylesheet>`,
+            'dd-constructable-stylesheet': ``,
+          },
+        },
+      ],
+      expectMutationsOnFirstRender: true, // The test setup manually attaches the de-dupe script tag.
+      stableSelectors: ['dedupe-constructable-stylesheet'],
```
augustjk, line 5577: should this be `'dd-constructable-stylesheet'`?

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -423,6 +432,117 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+    background-color: blue;
+  }
+</style><lit-ssr-style-dedupe style-id="1" style="display:none;"></lit-ssr-style-dedupe><style>
+  :host {
+    margin: 8px;
+  }
+</style><lit-ssr-style-dedupe style-id="3" style="display:none;"></lit-ssr-style-dedupe><!--lit-part--><!--/lit-part--></template></test-static-styles-array><test-static-styles-array><template shadowroot="open" shadowrootmode="open"><lit-ssr-style-dedupe style-id="1" style="display:none;"></lit-ssr-style-dedupe><lit-ssr-style-dedupe style-id="3" style="display:none;"></lit-ssr-style-dedupe><!--lit-part--><!--/lit-part--></template></test-static-styles-array><!--/lit-part-->`
+    );
+  });
+
```
augustjk, line 545: Could we test that sharing the same instance of the dedupe utility allows deduping across multiple render calls?

marcmalerei:
@AndrewJakubowicz, can you provide any updates on this?

prashantpalikhe:
It would be great to be able to opt in to this feature. We use a security package that parses the HTML and adds nonces to any script/links etc. And right now, the HTML size is really a problem there. Would be amazing to have no repeated styles.

marcmalerei:
Hi @AndrewJakubowicz ,

I hope this message finds you well. We are currently working on a project that heavily relies on the Lit framework, and we have encountered a challenge that we believe could be addressed by this pull request.

Context and Need:
Our project involves rendering pages statically, and as it stands, the current approach to styles in Lit is proving to be a significant bottleneck. Specifically, the inlined styles within each statically rendered component are causing our pages to bloat, resulting in increased load times and negatively impacting the overall performance and user experience.

It would be great if we can use this feature in a near future. 

Kind Regards

Marcel

prashantpalikhe:
@marcmalerei do you comrpess the HTML? gzip/brotli should compress these duplicate styles really well and significantly reduce the performance overhead.

marcmalerei:
> @marcmalerei do you comrpess the HTML? gzip/brotli should compress these duplicate styles really well and significantly reduce the performance overhead.

@prashantpalikhe, yes we do but if you have hundreds of the same component it still leads to a bloat of html. is there any progress here?

marcmalerei:
@AndrewJakubowicz, any progress on this topic?

HaniTaher:
@AndrewJakubowicz , can we have an update on this? For heavily visited websites that relays on SSR, this would be nice feature.

PR #4379, opened 2023-11-09
[infra] Split changeset status workflow out of lint

augustjk:
I didn't like that lint and changeset were combined.
While we do somewhat enforce the changeset check and add empty changesets when needed, I would like to see if lint and format checks fail separately from the changeset check.

I don't think doing an extra checkout and npm install is that costly.

Splitting out made it so the changeset will only run on pull requests, not on pushes to main. I also updated the condition so that it should continue to skip when the PR author is lit-robot, not just the initiator of the workflow. The previous condition would not skip the check and fail when one of us pushed up fixes to the release PR branch.

I also discovered changesets authors do not recommend this failing CI step (https://github.com/changesets/changesets/blob/main/docs/checking-for-changesets.md) since there's a bot that will check and comment on the packages being bumped. I think it does serve value for us as we may have now been conditioned to rely on the failure to remind us, and rarely check the comment.

Also took the opportunity to remove the 3.0 branch from the release workflow target.

### DIFF
--- a/.github/workflows/changeset-status.yaml
+++ b/.github/workflows/changeset-status.yaml
@@ -1,0 +1,31 @@
+name: Changeset Status
+
+on: [pull_request]
+
+jobs:
+  # Enforce that all PRs that change packages need changesets. Changes
+  # without changesets result in this job failing.
+  changeset:
+    runs-on: ubuntu-latest
+    # Skip if author is lit-robot which means it's a release PR that doesn't
+    # need this check.
+    if: ${{ github.event.pull_request.user.login != 'lit-robot' }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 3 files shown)


.github/workflows/changeset-status.yaml
```diff
@@ -0,0 +1,31 @@
+name: Changeset Status
```
justinfagnani, line 1: Why have this in a separate workflow vs another job in the tests workflow?

augustjk: Mainly for ease of having the trigger just be on pull_request rather than also on push to main that the Tests workflow has. We could keep it together and add an extra condition for this job to check for the triggering action.

AndrewJakubowicz (approved):
I like the separation of concerns and specificity. Thank you!

Issue #4374, opened 2023-11-08
[lit/react] React wrapped component typescript error with property named 'size' in Preact


ajmchambers:
### Which package(s) are affected?

React (@lit/react)

### Description

Getting a typescript error when using SlButton with [size](https://github.com/shoelace-style/shoelace/blob/next/src/components/button/button.component.ts#L64-L65) property in preact (react is fine).

```tsx
<SlButton size="small">Button</SlButton>
```

This is the text I see when hovering over size in vscode:

```
Type '"small"' is not assignable to type '((number | SignalLike<number | undefined>) & ("small" | "medium" | "large")) | undefined'.ts(2322)

jsx.d.ts(2316, 3): The expected type comes from property 'size' which is declared here on type 'IntrinsicAttributes & Omit<ComponentProps<SlButton, { onSlBlur: EventName<SlBlurEvent>; onSlFocus: EventName<SlFocusEvent>; onSlInvalid: EventName<SlInvalidEvent>; }>, "ref"> & RefAttributes<...> & Readonly<...>'

(property) size?: ((number | JSXInternal.SignalLike<number | undefined>) & ("small" | "medium" | "large")) | undefined
```

### Reproduction

Here is a stackblitz example:
https://stackblitz.com/edit/vitejs-vite-fhterf?file=src%2Fapp.tsx&terminal=dev

Note:
- modified vite preact typescript preset
- added preact/compat types to tsconfig

### Workaround

Can cast it as any to get rid of error:

```tsx
<SlButton size={"small" as any}>Button Text</SlButton>
```

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

lit: 3.0.2, @lit/react: 1.0.1

### Browser/OS/Node environment

Browser: Edge 119.0.2151.44
OS: Windows 11 23H2 22631.2506
Node: v18.18.0
npm: 9.4.2

augustjk:
This looks specifically due to `preact`'s typing of `HTMLAttributes` which includes this line: https://github.com/preactjs/preact/blob/b4a1cc206b3519d80b931134933c917b0f08178c/src/jsx.d.ts#L2316

Probably wanting to account for https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size

I don't see it in `@types/react`'s equivalent here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e2351cb03d18c520b908836dd38365efd563c104/types/react/index.d.ts#L1944

[...]

PR #4381, opened 2023-11-09
[react] Prefer element property type over React built-in type

augustjk:
Fixes https://github.com/lit/lit/issues/4374

### DIFF
--- a/packages/react/src/create-component.ts
+++ b/packages/react/src/create-component.ts
@@ -61,9 +61,9 @@ type ElementProps<I> = Partial<Omit<I, keyof HTMLElement>>;
 // Acceptable props to the React component.
 type ComponentProps<I, E extends EventNames = {}> = Omit<
   React.HTMLAttributes<I>,
-  // Omit keyof E to prefer provided event handler mapping over React's
-  // built-in event handler props.
-  keyof E
+  // Prefer type of provided event handler props or those on element over
+  // built-in HTMLAttributse
+  keyof E | keyof ElementProps<I>
 > &
   EventListeners<E> &
   ElementProps<I>;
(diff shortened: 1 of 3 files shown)


packages/react/src/create-component.ts
```diff
@@ -61,9 +61,9 @@ type ElementProps<I> = Partial<Omit<I, keyof HTMLElement>>;
 // Acceptable props to the React component.
 type ComponentProps<I, E extends EventNames = {}> = Omit<
   React.HTMLAttributes<I>,
-  // Omit keyof E to prefer provided event handler mapping over React's
-  // built-in event handler props.
-  keyof E
+  // Prefer type of provided event handler props or those on element over
+  // built-in HTMLAttributse
```
AndrewJakubowicz, line 65: `HTMLAttributse` -> `HTMLAttributes`

packages/react/src/test/create-component_test.tsx
```diff
@@ -241,6 +241,22 @@ suite('createComponent', () => {
...
+  test.skip('Prefer element property type over built-in HTMLAttribute', async () => {
+    const TestComponent = createComponent({
+      react: React,
+      tagName: 'my-component',
+      elementClass: class MyComponent extends ReactiveElement {
+        color: number = 0;
+      },
+    });
+
+    // `color` is `string | undefined` in React.HTMLAttribute.
```
AndrewJakubowicz, line 254: `HTMLAttribute` -> `HTMLAttributes`

packages/react/src/test/create-component_test.tsx
```diff
@@ -241,6 +241,22 @@ suite('createComponent', () => {
...
+    // It should be happy with number though because that's how it's typed in
+    // the element above.
+    <TestComponent color={1} />;
```
AndrewJakubowicz, line 257: Could you add a test beneath this that is expected to fail if now a `string` is passed in (since number was expected)?

E.g.:

```
@ts-expect-error
<TestComponent color={"red"} />
```

AndrewJakubowicz (commented):
Awesome!

packages/react/src/create-component.ts
augustjk: Sorry my fingers are dyslexic 😥

AndrewJakubowicz (approved):
Awesome!

PR #4382, opened 2023-11-10
[labs/ssr] Add a serverhtml function, for templates that are only rendered on the server

rictic:


packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * This is a performance optimization, causing the server not to emit comment
+ * markers for updating the dynamic parts of the server-rendered DOM.
+ * This results in fewer DOM nodes on the client. In most cases the difference
+ * will be small.
+ *
+ * It also allows for data binding into raw text elements like <title> and
+ * <textarea>.
+ *
+ * Currently, a server-only template cannot contain a normal Lit template,
```
jimsimon, line 39: Is this statement at odds with the description in the README?

> serverhtml template won't send down marker comments, so they can't be updated on the client, however if you render a normal Lit template inside a serverhtml template, then it can be hydrated and updated. Likewise, if you render a custom element inside a serverhtml template, it will automatically hydrate and update like normal.

jimsimon (commented):
Are there any security implications with using this new API? If so, it might be a good idea to add that info to the README.

Also, do we need to update the website docs?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -829,4 +920,66 @@ function* renderAttributePart(
...
+) {
+  if (isCompiledTemplateResult(result)) {
+    return result._$litType$.h.join('${...}');
+  }
+  return result.strings.join('${...}');
+}
+
 const getLast = <T>(a: Array<T>) => a[a.length - 1];
+
+function willExecute(node: Element | Template): boolean {
```
AndrewJakubowicz, line 938: Should some assumptions be documented on this function. E.g., this only holds true on server templates?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -829,4 +920,66 @@ function* renderAttributePart(
...
+  if (!/script/i.test(node.tagName)) {
+    return false;
+  }
+  if (isTemplateNode(node)) {
```
AndrewJakubowicz, line 942: Won't this be always false, as `<template>` would be caught in the !script regex above?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -489,7 +512,46 @@ const getTemplateOpcodes = (result: TemplateResult) => {
...
+            }
+            const text = child.value;
+            const textStart = child.sourceCodeLocation!.startOffset;
+            flushTo(textStart);
+            const markerRegex = new RegExp(marker.replace(/\$/g, '\\$'), 'g');
+            for (const mark of text.matchAll(markerRegex)) {
+              flushTo(textStart + mark.index!);
+              if (dangerous) {
+                throw new Error(
+                  `Found binding inside an executable <script> tag in a server-only template. For security reasons, this is not supported, as it could allow an attacker to execute arbitrary JavaScript.`
```
AndrewJakubowicz, line 536: Is there a workaround? `unsafe` directive?
The error should provide a way to provide executable code if you know what you're doing and want to do it.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -620,7 +695,21 @@ function* renderTemplateResult(
         break;
       case 'child-part': {
         const value = result.values[partIndex++];
-        yield* renderValue(value, renderInfo);
+        let isValueHydratable = hydratable;
+        if (isTemplateResult(value)) {
+          isValueHydratable = isHydratable(value);
+          if (!isValueHydratable && hydratable) {
+            throw new Error(
+              `A server-only template can't be rendered inside an ordinary, hydratable template. The outer template was:
```
AndrewJakubowicz, line 703: Great errors! May be easier to state that a server-only template can only be rendered at the top level or within other server-only templates.

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
AndrewJakubowicz, line 3: Nit: 2023

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,67 @@
...
+  type TemplateResult,
+} from 'lit-html';
+
+/**
+ * A lit-html template that can only be rendered on the server, and cannot be
+ * hydrated.
+ *
+ * This is a performance optimization, causing the server not to emit comment
+ * markers for updating the dynamic parts of the server-rendered DOM, and
+ * supports rendering into raw text elements like <title> and <textarea>.
```
AndrewJakubowicz, line 19: Leading with `performance optimization` may make people try and prefer this over the normal template, which can become an issue if they try and nest the server template in a client template.

I think we should lead with the features.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
...
+
+const serverClientHydrationTest: SSRTestDescription = {
+  render(title: string, c1: string, c2: string) {
+    return serverhtml`
+        <!doctype html>
+        <html>
+          <head><title>${title}</title></head>
+          <body>
+            <div id="one">${this.renderFns?.renderOne(c1)}</div>
+            <div id="two">${this.renderFns?.renderTwo(c2)}</div>
```
AndrewJakubowicz, line 5604: Optional chaining required?

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
...
+            <div id="one"><h1>c1</h1></div>
+            <div id="two"><h2>c2</h2></div>
+          </body>
+        </html>
+        `,
+    },
+    {
+      // Update #one
+      renderFn: 'renderOne',
+      args: ['hydrated c1'],
```
AndrewJakubowicz, line 5652: "updated c1"?

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
...
+            <div id="one"><h1>hydrated c1</h1></div>
+            <div id="two"><h2>c2</h2></div>
+          </body>
+        </html>
+        `,
+    },
+    {
+      // Update #two
+      renderFn: 'renderTwo',
+      args: ['hydrated c2'],
```
AndrewJakubowicz, line 5685: Technically already hydrated. Maybe would prefer "updated c2" here.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
...
+        },
+      },
+    ],
+    stableSelectors: ['server-only-attr'],
+    serverOnly: true,
+  },
+};
+
+const serverClientHydrationTest: SSRTestDescription = {
+  render(title: string, c1: string, c2: string) {
```
AndrewJakubowicz, line 5597: This is an awesome test

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -545,6 +552,133 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+    assert.is(result, `<div>onetwothree</div>`);
+  });
+
+  test('server-only template inside server-only template', async () => {
+    const {render, serverOnlyInsideServerOnly} = await setup();
+    const result = await render(serverOnlyInsideServerOnly);
+    assert.is(result, `<div>Server only</div>`);
+  });
+
+  test(`server-only template can't render a normal template`, async () => {
```
AndrewJakubowicz, line 580: Server only can render a normal template?

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -545,6 +552,133 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+  });
+
+  test('server-only template into raw element', async () => {
+    const {render, serverOnlyRawElementTemplate} = await setup();
+    const result = await render(serverOnlyRawElementTemplate);
+    assert.is(
+      result,
+      `
+    <title>No comments inside</title>
+    <textarea>This also works (kinda).</textarea>
```
AndrewJakubowicz, line 607: Why `(kinda)`?

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -545,6 +552,133 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
...
+      () => render(renderServerOnlyScript),
+      /Found binding inside an executable <script> tag in a server-only template\./
+    );
+    assert.throws(
+      () => render(renderServerOnlyScriptDeep),
+      /Found binding inside an executable <script> tag in a server-only template\./
+    );
+  });
+
+  test('server-only template into non-executing script', async () => {
```
AndrewJakubowicz, line 671: Worth adding an escape hatch in this PR for providing known safe executable scripts?
Also, what about `<script>` inside a `<template>` within a server template?

AndrewJakubowicz (commented):
This is amazing!

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,67 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
 top-level template can be used to pass data to the top-level components, and
 that template can be loaded and hydrated on the client to apply the same data.
 
+## Server-only templates
+
+You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags.
+
+serverhtml templates can be composed, and combined, and they support almost all features that normal Lit templates do, with the exception of features that don't have a pure HTML representation, like event handlers or property bindings.
```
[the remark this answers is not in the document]
rictic: Good call!

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,67 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
 top-level template can be used to pass data to the top-level components, and
 that template can be loaded and hydrated on the client to apply the same data.
 
+## Server-only templates
+
+You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags.
```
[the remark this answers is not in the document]
rictic: Clarified

At least for now we don't support binding into script tags that will execute, because this PR had already grown way too large

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,67 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+
+serverhtml templates can be composed, and combined, and they support almost all features that normal Lit templates do, with the exception of features that don't have a pure HTML representation, like event handlers or property bindings.
+
+serverhtml template won't send down marker comments, so they can't be updated on the client, however if you render a normal Lit template inside a serverhtml template, then it can be hydrated and updated. Likewise, if you render a custom element inside a serverhtml template, it will automatically hydrate and update like normal.
```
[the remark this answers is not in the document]
rictic: Good point! Done. I wanted to include a mention of markers, so I added that as a sentence in the first paragraph, for why someone might want to use them.

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,67 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+import {RenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
+import './app-shell.js';
+import {contentTemplate} from './content-template.js';
+
+const pageInfo = {
+  /* ... */
+};
+
+const ssrResult = render(serverhtml`
+  <html>
```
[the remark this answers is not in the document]
rictic: Yes!

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,67 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+      <script type="text/json" id="page-info">
+        ${JSON.stringify(pageInfo)}
+      </script>
+    </body>
+  </html>
+`);
+
+// ...
+
+context.body = Readable.from(ssrResult);
```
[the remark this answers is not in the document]
rictic: Ah, good call. Done

packages/labs/ssr/src/lib/render-value.ts
rictic: I haven't proven one out. I'd like to do that as a follow up, this PR is too large as it is!

jimsimon: I agree that we'll want to provide a workaround or escape hatch here if the current unsafe directives don't work. As an example, some of our scripts that get included on every page are intentionally inlined and we inject them via a pattern like:
```
<script nonce=${...} type="module">
  ${inlineScriptContents}
</script>
```

packages/labs/ssr/src/lib/render-value.ts
rictic: Added that as another phrasing to this error message.

packages/labs/ssr/src/lib/render-value.ts
rictic: True, was using it for type narrowing, rewrote our script tag check instead.

packages/labs/ssr/src/lib/render-value.ts
rictic: Yes, and thinking about it, a more specific name is called for as well.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -489,7 +512,46 @@ const getTemplateOpcodes = (result: TemplateResult) => {
           ops.push({
             type: 'custom-element-shadow',
           });
+        } else if (
+          !hydratable &&
+          /^(title|textarea|script)$/.test(node.tagName)
+        ) {
+          const dangerous = willExecute(node);
+          console.log(`dangerous: ${dangerous} – node: `, node);
```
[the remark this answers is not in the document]
rictic: Expanded instead :)

packages/labs/ssr/src/lib/render-value.ts
jimsimon: I can see this causing some headaches for us down the road. Would this be a valid workaround if I have a server-only template that I want to use inside a normal Lit template?

```
const normalChildTemplate = html`<h1>Hello World</h1>`
const serverChildTemplate = serverHtml`<span>${normalChildTemplate}</span>`

const rootTemplate = html`<div>${normalChildTemplate}</div>`
const otherServerTempate = serverHtml`<p>${serverChildTemplate}</p>`
```

This seems like it would work as a mechanism for sharing templates? I ask because being able to share templates between client and server is one of the big draws of Lit SSR for us (at Reddit).

packages/labs/ssr/src/lib/server-template.ts
rictic: Good catch, done

packages/labs/ssr/src/lib/server-template.ts
rictic: Good point. I dropped the docs on the module-local `SERVER_ONLY` constant, and used the description from the README to document the `serverhtml` function.

packages/labs/ssr/src/lib/server-template.ts
rictic: Good catch, yes, that was an out of date comment from an earlier commit, before I added support for normal templates inside server-only templates. Rewrote the comment.

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Yes, the type system requires it.

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Thanks!

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Done!

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Good point, done!

packages/labs/ssr/src/test/lib/render-lit_test.ts
rictic: Ah, yeah! I changed the implementation but not the name

packages/labs/ssr/src/test/lib/render-lit_test.ts
rictic: IMO this PR is already too big, and that's a fairly fringe use case. Let's do that separately.

A script inside a template is safe, but harder to detect, and there aren't clear use cases that I can see.

packages/labs/ssr/src/test/lib/render-lit_test.ts
rictic: Ah, good point. It works with the full promise of server-only templates. You can render them once, but you can't update them!

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,67 @@
...
+ * elements behave the same in both server-only templates and normal templates,
+ * and by default they will hydrate.
+ *
+ * A `serverhtml` template can't be rendered inside a normal Lit template.
+ */
+export function serverhtml(
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): ServerRenderedTemplate {
+  const value = baseHtml(strings, ...values) as ServerRenderedTemplate;
```
AndrewJakubowicz, line 47: How do serverhtml templates handle directives? I assume they treat them in the same way as SSR'd directives.

Should any directives that cannot ever work on the server also be an error? E.g. any ElementPart directives have no `render` representation and so do not have an SSR representation.

packages/labs/ssr/src/lib/render-value.ts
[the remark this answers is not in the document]
AndrewJakubowicz: Good call :P

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
...
+    'body',
+    'div#one',
+    'div#two',
+    'h1',
+    'h2',
+  ],
+  serverOnly: true as const,
 };
+tests['client templates inside server templates can hydrate'] =
+  serverClientHydrationTest;
```
augustjk, line 5712: Was there a particular reason for not inlining this test object?

augustjk (commented):
Very nice! I just have minor nits, mostly around docs.

packages/labs/ssr/src/lib/server-template.ts
augustjk: bindings in element part silently do nothing. since server only templates will throw on property and event binding for attribute parts, it doesn't seem like a bad idea for it to throw on element part binding too? though in the future we have opportunities to add server only directives like spreading attributes there.

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,68 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+`serverhtml` templates can be composed, and combined, and they support almost all features that normal Lit templates do, with the exception of features that don't have a pure HTML representation, like event handlers or property bindings.
+
+`serverhtml` templates can only be rendered once, so they can't be updated on the client. However if you render a normal Lit template inside a serverhtml template, then it can be hydrated and updated. Likewise, if you render a custom element inside a serverhtml template, it will automatically hydrate and update like normal.
+
+```js
+import {html} from 'lit';
+import {render, serverhtml} from '@lit-labs/ssr';
+import {RenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
+import './app-shell.js';
+import {contentTemplate} from './content-template.js';
```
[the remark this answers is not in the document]
rictic: Added a comment to the code, how's that?

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,68 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+
+        // Hydrate content template
+        const pageInfo = JSON.parse(document.getElementById('page-info').textContent);
+        hydrate(contentTemplate(pageInfo.description), document.querySelector('#content'));
+        // #content element can now be efficiently updated
+      </script>
+      <!-- Pass data to client. -->
+      <script type="text/json" id="page-info">
+        ${JSON.stringify(pageInfo)}
+      </script>
```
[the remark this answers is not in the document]
rictic: Added a comment. This pattern, of putting the data after the module script, is what we want to recommend I think, so that the browser can be downloading dependencies while also streaming down JSON data.

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,68 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+          hydrateShadowRoots,
+        } from './node_modules/@webcomponents/template-shadowroot/template-shadowroot.js';
+        import {hydrate} from '@lit-labs/ssr-client';
+        import {contentTemplate} from './content-template.js';
+        if (!hasNativeDeclarativeShadowRoots()) {
+          hydrateShadowRoots(document.body);
+        }
+
+        // Load and hydrate app-shell lazily
+        import('./app-shell.js');
```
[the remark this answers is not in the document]
rictic: I find that somewhat convincing, but on the other hand, if we're gonna have one example of server-side scripts, it seems good for it to be a reasonably full one, that shows everything working together.

packages/labs/ssr/package.json
```diff
@@ -172,6 +174,22 @@
         "../../tests/web-test-runner.config.js"
       ],
       "output": []
+    },
+    "test:build": {
```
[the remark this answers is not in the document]
rictic: Ah, yeah, that's better

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5511,4 +5512,201 @@ export const tests: {[name: string]: SSRTest} = {
       stableSelectors: ['le-internals-hydrate'],
     };
   },
+  'Server-only template works': {
```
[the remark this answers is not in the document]
rictic: Ah, yes!

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Yeah, so that we can reference `this.renderFns` in `render`, because a test factory won't have a `renderFns` field, but a test description will

rictic:
> bindings in element part silently do nothing. since server only templates will throw on property and event binding for attribute parts, it doesn't seem like a bad idea for it to throw on element part binding too? though in the future we have opportunities to add server only directives like spreading attributes there. 

Good point! Threw an error in that case and added a test.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -480,6 +480,10 @@ const getTemplateOpcodes = (result: TemplateResult) => {
                 useCustomElementInstance: node.isDefinedCustomElement,
               });
             } else {
+              if (!hydratable) {
+                throw new Error(`Server-only templates don't support element parts, as their API does not currently give them any way to render anything on teh server. Found in template:
```
AndrewJakubowicz, line 484: Nit: `teh` -> `the`.

Although I do kinda like `teh` :P

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  html as baseHtml,
+  type CompiledTemplateResult,
+  type TemplateResult,
+} from 'lit-html';
+
+const SERVER_ONLY = 1;
+
+export interface ServerRenderedTemplate extends TemplateResult {
+  $_litServerRenderMode: typeof SERVER_ONLY;
```
AndrewJakubowicz, line 16: If I understand this PR correctly, it is possible to render a `ServerRenderedTemplate` with `lit-html`'s `render` function, since it's a TemplateResult with an extra field that `lit-html` would ignore.

Is this by design so server templates could be shared and rendered as client templates as well (with the caveats that come with client templates)?

If this is by design it may be worth calling out. Technically you could use a `serverhtml` template within a client template if it's only occurring on the client and not during SSR.

If not intended, it may be worth throwing a descriptive error.

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,73 @@
...
+  html as baseHtml,
+  type CompiledTemplateResult,
+  type TemplateResult,
+} from 'lit-html';
+import {isServer} from 'lit-html/is-server.js';
+
+const SERVER_ONLY = 1;
+
+export interface ServerRenderedTemplate extends TemplateResult {
+  $_litServerRenderMode: typeof SERVER_ONLY;
```
augustjk, line 17: Should this brand follow the shareable private field pattern described here? https://github.com/lit/lit/blob/main/CONTRIBUTING.md#private-property-prefixing

And be similar to `_$litType$` ?
```suggestion
  _$litServerRenderMode$: typeof SERVER_ONLY;
```

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -44,6 +45,11 @@ export function serverhtml(
   strings: TemplateStringsArray,
   ...values: unknown[]
 ): ServerRenderedTemplate {
+  if (!isServer) {
+    throw new Error(
+      `serverhtml templates can only be rendered on the server, they cannot be rendered in the browser. Use the html function for templates that need to be rendered from the browser. This check is based on the "node" export condition: https://nodejs.org/api/packages.html#conditional-exports`
```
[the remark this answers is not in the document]
rictic: Yeah, but in that case I think we'll just want to update this error message. Fundamentally, we want `isServer` to be `false` in the browser, and true in all those various server environments. The export condition is just our initial implementation.

packages/labs/ssr/src/lib/server-template.ts
rictic: Oh, yes! Good catch!

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,70 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+          hydrateShadowRoots,
+        } from './node_modules/@webcomponents/template-shadowroot/template-shadowroot.js';
+        import {hydrate} from '@lit-labs/ssr-client';
+        import {getContent} from './content-template.js';
+        if (!hasNativeDeclarativeShadowRoots()) {
+          hydrateShadowRoots(document.body);
+        }
+
+        // Load and hydrate app-shell lazily
+        import('./app-shell.js');
```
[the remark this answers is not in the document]
rictic: Yeah, that should be fine right, to hydrate the content template and the app-shell element asynchronously, like we're doing here?

Or maybe I don't follow your comment?

packages/labs/ssr/src/lib/render-value.ts
rictic: Initial tests are showing that unsafeHTML works out of the box, so once this lands you should be able to write:

```typescript
serverhtml`
  ${unsafeHTML(`
      <script nonce=${...} type="module">
        ${inlineScriptContents}
      </script>`)}
`;
```

packages/labs/ssr/README.md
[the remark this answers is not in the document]
augustjk: Yeah assuming all the `<app-shell>` is doing would be project its children to a `<slot>`, upgrading it and hydrating content shouldn't depend on each other. It might be possible that hydrating content before element is upgraded could cause FOUC, which awaiting on the element upgrade could prevent.

I don't think it's a big enough worry to change anything here.

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,70 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
...
+## Server-only templates
+
+You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags like `<script type="text/json">`. They are also slightly more efficient than normal Lit templates, because the generated HTML doesn't need to include markers for updating.
+
+`serverhtml` templates can be composed, and combined, and they support almost all features that normal Lit templates do, with the exception of features that don't have a pure HTML representation, like event handlers or property bindings.
+
+`serverhtml` templates can only be rendered once, so they can't be updated on the client. However if you render a normal Lit template inside a serverhtml template, then it can be hydrated and updated. Likewise, if you place a custom element inside a serverhtml template, it can be hydrated and update like normal.
+
+```js
+import {html} from 'lit';
```
augustjk, line 188: This import is unused in this file and can be removed.
```suggestion
```

packages/labs/ssr/README.md
```diff
@@ -175,6 +176,70 @@ attributes or via a side-channel mechanism. This is in no way fundamental; the
 top-level template can be used to pass data to the top-level components, and
 that template can be loaded and hydrated on the client to apply the same data.
 
+## Server-only templates
+
+You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags like `<script type="text/json">`. They are also slightly more efficient than normal Lit templates, because the generated HTML doesn't need to include markers for updating.
+
+`serverhtml` templates can be composed, and combined, and they support almost all features that normal Lit templates do, with the exception of features that don't have a pure HTML representation, like event handlers or property bindings.
+
+`serverhtml` templates can only be rendered once, so they can't be updated on the client. However if you render a normal Lit template inside a serverhtml template, then it can be hydrated and updated. Likewise, if you place a custom element inside a serverhtml template, it can be hydrated and update like normal.
```
AndrewJakubowicz, line 185: Nit–non blocking docs. Can always be addressed in followup.

"`serverhtml` templates should only be rendered once, so they can't be updated on the client.".

This could be misconstrued as: you can use them somehow on the client as long as they are only rendered once. May be worth clarifying again here that `serverhtml` templates throw if they are used on the client.

augustjk (approved):

packages/labs/ssr/README.md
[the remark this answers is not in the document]
augustjk: Yeah we can tweak this later as we go. With more browsers shipping DSD, I think we can eventually remove the `template-shadowroot` usage from these examples and just have a small dedicated section for it.

packages/labs/ssr/README.md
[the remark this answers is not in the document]
augustjk: Comment looks good. I propose we remove the import mentioned here: https://github.com/lit/lit/pull/4382#discussion_r1393385349

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -489,7 +516,49 @@ const getTemplateOpcodes = (result: TemplateResult) => {
           ops.push({
             type: 'custom-element-shadow',
           });
+        } else if (
+          !hydratable &&
+          /^(title|textarea|script)$/.test(node.tagName)
```
AndrewJakubowicz, line 521: Should `style` be added here. (It's also handled as a raw text element by `lit-html`)?

packages/labs/ssr/README.md
rictic: Ah, good point! Did another pass

packages/labs/ssr/README.md
rictic: Good catch! Done

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -761,7 +859,7 @@ function* renderTemplateResult(
     }
   }
 
-  if (partIndex !== result.values.length) {
+  if (partIndex !== result.values.length && hydratable) {
     throwErrorForPartIndexMismatch(partIndex, result);
```
AndrewJakubowicz, line 863: Why are part index mismatches ok with server templates?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -829,4 +927,72 @@ function* renderAttributePart(
...
+    return result._$litType$.h.join('${...}');
+  }
+  return result.strings.join('${...}');
+}
+
 const getLast = <T>(a: Array<T>) => a[a.length - 1];
+
+/**
+ * Returns true if the given node is a <script> node that the browser will
+ * automatically executeif it's rendered on server-side, outside of a
```
AndrewJakubowicz, line 947: Nit: executeif -> "execute if"

packages/labs/ssr/src/lib/server-template.ts
```diff
@@ -0,0 +1,73 @@
...
+ * like `<script type="text/json">`. They are also slightly more efficient than
+ * normal Lit templates, because the generated HTML doesn't need to include
+ * markers for updating.
+ *
+ * `serverhtml` templates can be composed, and combined, and they support
+ * almost all features that normal Lit templates do, with the exception of
+ * features that don't have a pure HTML representation, like event handlers or
+ * property bindings.
+ *
+ * `serverhtml` templates can only be rendered once, so they can't be updated
```
AndrewJakubowicz, line 36: "can't be updated on the client". Maybe here is where we add the text around "can't be used on the client at all".

packages/labs/ssr/src/lib/render-value.ts
rictic: Done!

packages/labs/ssr/src/lib/render-value.ts
rictic: Ah, I think that was from an old and worse prototype! Thank you for catching this!

AndrewJakubowicz (approved):
AMAZING!

packages/labs/ssr/src/lib/render-value.ts
rictic: Good point! Treating similarly to `<script>`, with unsafeHTML as the escape hatch.

packages/labs/ssr/src/lib/server-template.ts
rictic: Updated!

Issue #4266, opened 2023-10-06
ReactiveController: adding/removing controllers in lifecycle methods affects other controllers


sorvell:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

If a `ReactiveController` removes itself or adds another controller during a lifecycle method like `hostUpdate` or `hostUpdated`, this can prevent other controllers from properly running.

### Reproduction

The initial text of the button in [this repro](https://lit.dev/playground/#gist=78ecf665bbece46f1eaa8e55b3ebd138) should be the following and then clicking the button should not change the text.

Expected:
```
controller1.count: 1
controller2.count: 1
```

Actual:
```
controller1.count: 1
controller2.count: 0
```
(after clicking once)
```
controller1.count: 1
controller2.count: 1
```

### Workaround

The issue is that the controllers are stored in an Array and to process the controllers' lifecycle, that array is `forEach`'d. Mutating the array while it's iterating unexpectedly alters the iteration. 

The only workaround is not to alter the controllers list inside a controller lifecycle method.

This could possibly be addressed by (1) copying the controllers array before iterating, or (2) making the controllers a `Set`.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

All

### Browser/OS/Node environment

Browser: All

justinfagnani:
Ouch. Yeah, it's a concurrent modification bug. Set seems like the right way to go.

PR #4388, opened 2023-11-10
Controllers can be added/removed during lifecycle.

sorvell:
Fixes https://github.com/lit/lit/issues/4266. Controllers are maintained via a Set instead of an Arry, allowing them to be iterated stably while being modified.

### DIFF
--- a/.changeset/strong-ants-think.md
+++ b/.changeset/strong-ants-think.md
@@ -4,4 +4,4 @@
 'lit-element': patch
 ---
 
-Controllers can be added/removed during lifecycle (#4266).
+Fixes bug where adding or removing controllers during a reactive controller lifecycle would affect the execution of other controllers (#4266). Controllers can now be added/removed during lifecycle without affecting others.
--- a/.changeset/strong-ants-think.md
+++ b/.changeset/strong-ants-think.md
@@ -1,0 +1,7 @@
+---
+'@lit/reactive-element': patch
+'lit': patch
+'lit-element': patch
+---
+
+Controllers can be added/removed during lifecycle (#4266).
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -993,7 +993,7 @@ export abstract class ReactiveElement
   /**
    * Set of controllers.
    */
-  private __controllers?: ReactiveController[];
+  private __controllers?: Set<ReactiveController>;
 
   constructor() {
     super();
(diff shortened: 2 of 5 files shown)


scripts/check-size.js
```diff
@@ -9,7 +9,7 @@ import * as fs from 'fs';
 // it's likely that we'll ask you to investigate ways to reduce the size.
 //
 // In either case, update the size here and push a new commit to your PR.
-const expectedLitCoreSize = 15423;
+const expectedLitCoreSize = 15405;
```
AndrewJakubowicz, line 12: Size win! :D

.changeset/strong-ants-think.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/reactive-element': patch
+'lit': patch
+'lit-element': patch
+---
+
+Controllers can be added/removed during lifecycle (#4266).
```
AndrewJakubowicz, line 7: This may be a breaking change for anyone relying on the broken behavior. I wonder if the changelog should elaborate more on the bug this fixes.

E.g., adding or removing controllers during lifecycle callbacks seems to halt updating other controllers prior to this fix.

AndrewJakubowicz (approved):
This is an excellent bug fix. I checked the current behavior for myself with a small [minimal repro](https://lit.dev/playground/#gist=c57bc726fb5bb3704285aa0524f44be8) , and the current behavior definitely feels like a bug and is non intuitive.

And this improvement comes with tests and a size win!

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1045,9 +1045,7 @@ export abstract class ReactiveElement
    * @category controllers
    */
   removeController(controller: ReactiveController) {
-    // Note, if the indexOf is -1, the >>> will flip the sign which makes the
-    // splice do nothing.
-    this.__controllers?.splice(this.__controllers.indexOf(controller) >>> 0, 1);
+    this.__controllers?.delete(controller);
```
rictic, line 1048: Oh nice, looks like Sets are very well behaved when mutated during iteration: https://lit.dev/playground/#gist=f424f3aab96d7c839d43c2c5815c8a10

rictic (approved):

.changeset/strong-ants-think.md
augustjk: How about something like:
```suggestion
Fixes bug where adding or removing controllers during a reactive controller lifecycle would affect the execution of other controllers (#4266). Controllers can now be added/removed during lifecycle without affecting others.
```

augustjk (approved):
Thank you! I went ahead and merged main to resolve the size check conflict after the other PR got merged, and removed the extra changeset file.

.changeset/strong-ants-think.md
AndrewJakubowicz: I like it!

PR #4389, opened 2023-11-10
[labs/ssr] fix: RenderResultReadable skipping async work

AndrewJakubowicz:
### Context

Fix a race condition bug with RenderResultReadable.

### What is the bug?

The `RenderResultReadable` class lets us efficiently stream synchronous or async iterables. It's architecture is [designed to be as optimal as possible](https://github.com/lit/lit/pull/3467#issuecomment-1319448671), so synchronous streams go fast and have minimal overhead from the async handling.

Our `RenderResultReadable.prototype._read` implementation is async, allowing us to `await` values in the stream.
From [Node.js documentation](https://nodejs.org/api/stream.html#readable_readsize): "`_read()` will be called again after each call to [`this.push(dataChunk)`](https://nodejs.org/api/stream.html#readablepushchunk-encoding) once the stream is ready to accept more data."

This is also the behavior back to [Node.js v10](https://nodejs.org/docs/latest-v10.x/api/stream.html#stream_readable_read_size_1).

What happens for the test case below (without the fix) when we are reading from the stream:

```ts
new RenderResultReadable([
    'a',
    new Promise((res) => setTimeout(res, 50)).then((_) => ['b']),
    'c',
]);
```

 1. Node.js calls `_read()`. We synchronously `push('a')`.
 2. Our `_read` encounters the promise and awaits the promise.
 3. Node.js calls `_read()`, and we synchronously `push('c')`, then we loop and synchronously call `push(null)` terminating the stream.

Thus the promise is skipped.

### Fix

Once `_read()` is called, it will not be called again until `readable.push()` has been called. This makes it safe for us to noop a `_read()` call while a Promise is waiting. Then once that promise resolves, it will result in `push()` being called with the resolved data and we're back in a consistent state.

### Test plan

I added two tests that reproduce without the `_waiting` flag.

### DIFF
--- a/.changeset/tasty-sloths-pump.md
+++ b/.changeset/tasty-sloths-pump.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+fix a race condition in `RenderResultReadable` which could skip async work.
--- a/.changeset/tasty-sloths-pump.md
+++ b/.changeset/tasty-sloths-pump.md
@@ -2,4 +2,4 @@
 '@lit-labs/ssr': patch
 ---
 
-fix a race condition in `RenderResultReadable` which could skip async work.
+Fix a race condition in `RenderResultReadable` which could skip async work.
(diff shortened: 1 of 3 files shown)


augustjk (approved):
Nice find and fix!

Issue #1434, opened 2020-11-13
[lit-ssr] Decide on strategy for handling data-mismatch during hydration between server and client


kevinpschaaf:
## Description
Currently the `lit-ssr` code has a few TODOs and open questions around how to handle data mismatch between what was rendered on the server and what is hydrated on the client. In some cases we could fairly cheaply detect the mismatch and then re-render the correct value on the client, but at the limit doing this in all cases would require more diffing than we do today and potentially impact hydration performance.

Note that mismatches can occur today when rendering `LitElement`s in top-level `lit-html` templates that have data-bindings to the element, since the element needs to hold off its initial update/hydration step until `hydrate` is called on the outer template. This is an instance of needing hydration coordination for custom elements, and should be solved by the "pending hydration" fix in https://github.com/Polymer/lit-html/pull/1388.

## Acceptance criteria
Decide on a strategy and break down the work to be done into tasks.

tpluscode:
Having bumped into such a mismatch a few times already, I think I would appreciate a fallback which does indeed replace the server-side DOM when it occurs.

Obviously, that should be a last-resort and authors should make sure that it does not happen but currently the hydrations simply breaks and leaves a page potentially broken in case of a mismatch

R3D-CODE:
I've started to investigate Lit's SSR packages, and till now just read `render.ts` and `render-value.ts`
But when I was reading `render.ts` wrote this:
1. Component lifecycles optimization
2. Asynchronous data fetching
3. Minimizing the amount of markup sent over the network --> code splitting, lazy loading, server-side caching
4. Size and format optimization --> images and other media assets, etc
5. CSR dynamic components that dont need to be pre-rendered on the server
6. CDN or other caching mechanism for static assets
7. Accessibility and SEO best practices 
8. Streams api
Hope helps

kevinpschaaf:
Seems like we just need to do the work of adding an "allow different initial data" flag that:
1. For child parts:
  1.a. When value is a TemplateResult, detects when a template digest differs and skips to the end part marker for that template instance, clears everything in between, and then renders the new TemplateResult
  1.b. When value is an iterable, detects when more need to be rendered (and CSRs those) or fewer (and clears them)
  1.c. When value is a (coerced) string, sets textContent during hydration (either unconditionally or after diffing, depending on what's fastest)
2. For attribute parts, calls setAttribute() during hydration (either unconditionally or after a getAttribute() diff, depending on what's fastest)

[...]

PR #4390, opened 2023-11-11
[labs/ssr] Add `serverUpdateComplete` promise to ReactiveController interface for SSR async work

AndrewJakubowicz:
### Context

This is split out of https://github.com/lit/lit/pull/1690 and implements a small interface which allows a ReactiveController to do async work on the server.

The demo works and can be tested with `npm run demo:global` from the `/labs/ssr` directory.

### Test plan

TODO

### DIFF
--- a/packages/labs/ssr-client/src/controllers/server-controller.ts
+++ b/packages/labs/ssr-client/src/controllers/server-controller.ts
@@ -1,0 +1,9 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export interface ServerController {
+  serverUpdateComplete: Promise<unknown>;
+}
--- a/packages/labs/ssr/src/lib/lit-element-renderer.ts
+++ b/packages/labs/ssr/src/lib/lit-element-renderer.ts
@@ -14,10 +14,11 @@ import {
 import {renderValue} from './render-value.js';
 import type {RenderInfo} from './render-value.js';
 import type {RenderResult} from './render-result.js';
+import type {ServerController} from '@lit-labs/ssr-client/controllers/server-controller.js';
 
 export type Constructor<T> = {new (): T};
 
-const {attributeToProperty, changedProperties} = _$LE;
+const {attributeToProperty, changedProperties, getControllers} = _$LE;
 
 /**
  * ElementRenderer implementation for LitElements
@@ -85,6 +86,20 @@ export class LitElementRenderer extends ElementRenderer {
   }
 
   override *renderShadow(renderInfo: RenderInfo): RenderResult {
+    const serverControllers = getControllers(this.element)
+      ?.map((c: Partial<ServerController>) => c.serverUpdateComplete)
+      .filter((p: Promise<unknown> | undefined) => !!p);
+    if (serverControllers?.length > 0) {
+      const continuation = Promise.allSettled(serverControllers).then((_) =>
+        this._renderShadowContents(renderInfo)
+      );
+      yield continuation;
+    } else {
+      yield* this._renderShadowContents(renderInfo);
+    }
+  }
+
+  private *_renderShadowContents(renderInfo: RenderInfo): RenderResult {
     // Render styles.
     const styles = (this.element.constructor as typeof LitElement)
       .elementStyles;
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -273,6 +273,8 @@ export const _$LE = {
   },
   // eslint-disable-next-line
   _$changedProperties: (el: LitElement) => (el as any)._$changedProperties,
+  // eslint-disable-next-line
+  _$controllers: (el: LitElement) => (el as any)._$controllers,
 };
 
 // IMPORTANT: do not change the property name or the assignment expression.
--- a/package-lock.json
+++ b/package-lock.json
@@ -3220,6 +3220,141 @@
         "url": "https://github.com/sponsors/isaacs"
       }
     },
+    "node_modules/@next/swc-darwin-arm64": {
+      "version": "13.3.1",
+      "resolved": "https://registry.npmjs.org/@next/swc-darwin-arm64/-/swc-darwin-arm64-13.3.1.tgz",
+      "integrity": "sha512-UXPtriEc/pBP8luSLSCZBcbzPeVv+SSjs9cH/KygTbhmACye8/OOXRZO13Z2Wq1G0gLmEAIHQAOuF+vafPd2lw==",
+      "cpu": [
+        "arm64"
+      ],
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
+    },
+    "node_modules/@next/swc-darwin-x64": {
+      "version": "13.3.1",
+      "resolved": "https://registry.npmjs.org/@next/swc-darwin-x64/-/swc-darwin-x64-13.3.1.tgz",
+      "integrity": "sha512-lT36yYxosCfLtplFzJWgo0hrPu6/do8+msgM7oQkPeohDNdhjtjFUgOOwdSnPublLR6Mo2Ym4P/wl5OANuD2bw==",
+      "cpu": [
+        "x64"
+      ],
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
+    },
+    "node_modules/@next/swc-linux-arm64-gnu": {
+      "version": "13.3.1",
+      "resolved": "https://registry.npmjs.org/@next/swc-linux-arm64-gnu/-/swc-linux-arm64-gnu-13.3.1.tgz",
+      "integrity": "sha512-wRb76nLWJhonH8s3kxC/1tFguEkeOPayIwe9mkaz1G/yeS3OrjeyKMJsb4+Kdg0zbTo53bNCOl59NNtDM7yyyw==",
+      "cpu": [
+        "arm64"
+      ],
+      "optional": true,
+      "os": [
+        "linux"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
+    },
+    "node_modules/@next/swc-linux-arm64-musl": {
+      "version": "13.3.1",
+      "resolved": "https://registry.npmjs.org/@next/swc-linux-arm64-musl/-/swc-linux-arm64-musl-13.3.1.tgz",
+      "integrity": "sha512-qz3BzjJRZ16Iq/jrp+pjiYOc0jTjHlfmxQmZk9x/+5uhRP6/eWQSTAPVJ33BMo6oK5O5N4644OgTAbzXzorecg==",
+      "cpu": [
+        "arm64"
+      ],
+      "optional": true,
+      "os": [
+        "linux"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
+    },
+    "node_modules/@next/swc-linux-x64-gnu": {
+      "version": "13.3.1",
+      "resolved": "https://registry.npmjs.org/@next/swc-linux-x64-gnu/-/swc-linux-x64-gnu-13.3.1.tgz",
+      "integrity": "sha512-6mgkLmwlyWlomQmpl21I3hxgqE5INoW4owTlcLpNsd1V4wP+J46BlI/5zV5KWWbzjfncIqzXoeGs5Eg+1GHODA==",
+      "cpu": [
+        "x64"
+      ],
+      "optional": true,
+      "os": [
+        "linux"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
+    },
+    "node_modules/@next/swc-linux-x64-musl": {
+      "version": "13.3.1",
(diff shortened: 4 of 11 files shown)


packages/labs/ssr-client/src/controllers/server-controller.ts
```diff
@@ -0,0 +1,20 @@
...
+ * ServerController extends the Reactive Controller interface to allow
+ * asynchronous work to occur during SSR on the server.
+ */
+export interface ServerController extends ReactiveController {
+  /**
+   * `serverUpdateComplete` is accessed on the server during SSR. When
+   * implemented, SSR rendering will pause until all the controllers on this
+   * ReactiveController's host have settled.
+   */
+  serverUpdateComplete: Promise<unknown>;
```
rictic, line 19: I think we want to say `undefined | Promise<unknown>`

e.g. Task is a controller that might or might not want to produce a promise here, based on configuration

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -85,6 +86,22 @@ export class LitElementRenderer extends ElementRenderer {
   }
 
   override *renderShadow(renderInfo: RenderInfo): RenderResult {
+    const serverControllers = Array.from<Partial<ServerController>>(
```
rictic, line 89: serverControllers isn't right, more like serverControllerUpdatePromises I think

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -85,6 +86,22 @@ export class LitElementRenderer extends ElementRenderer {
...
+      getControllers(this.element) ?? []
+    )
+      .map((c: Partial<ServerController>) => c.serverUpdateComplete)
+      .filter((p: Promise<unknown> | undefined) => !!p);
+    if (serverControllers?.length > 0) {
+      const continuation = Promise.allSettled(serverControllers).then((_) =>
```
rictic, line 95: Hm, this means that if a serverController's updateComplete promise rejects, we will silently ignore that error, which isn't good. I think we should at least console.error log it.

rictic, line 95: Hm, this means that if a serverController's updateComplete promise rejects, we will silently ignore that error, which isn't good. I think we should at least console.error log it.

More broadly, how do we handle errors while generating an SSR response? How should we?

packages/lit-element/src/lit-element.ts
```diff
@@ -273,6 +273,8 @@ export const _$LE = {
   },
   // eslint-disable-next-line
   _$changedProperties: (el: LitElement) => (el as any)._$changedProperties,
+  // eslint-disable-next-line
+  _$controllers: (el: LitElement) => (el as any)._$controllers,
```
rictic, line 277: hm, for these three I think we want to give them explicit return type annotations, because otherwise the return type is `any`

packages/labs/ssr/src/lib/lit-element-renderer.ts
AndrewJakubowicz: Alternatively, I can change to `all` so it rejects.

packages/labs/ssr-client/src/controllers/server-controller.ts
augustjk: Should it just be optional?

packages/labs/ssr/src/lib/lit-element-renderer.ts
AndrewJakubowicz: Done

packages/labs/ssr/src/lib/lit-element-renderer.ts
AndrewJakubowicz: Changed to Promise.all. Will also add a test for this case.

AndrewJakubowicz: Turns out, it's not easy to add a test for this case, as it's a catastrophic failure. But that is better than a silent failure, and matches more closely how we fail in other areas.

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -5512,6 +5513,167 @@ export const tests: {[name: string]: SSRTest} = {
...
+  /******************************************************
+   * ReactiveController tests
+   ******************************************************/
+
+  'ReactiveController: Constructor runs': () => {
+    return {
+      registerElements() {
+        class BasicController implements ReactiveController {
+          // To satisfy the ReactiveController interface.
+          declare hostConnected: undefined;
```
augustjk, line 5526: interesting. i guess even if they're all optional, it expects at least one of the lifecycle methods to have been implemented to satisfy the interface, none of which gets called server side.
i guess this or a no-op method would be fine too.

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -700,6 +707,65 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
       render(renderServerOnlyElementPart);
     }, /Server-only templates don't support element parts/);
   });
+
+  test('all ServerControllers must resolve for an element to complete rendering', async () => {
```
augustjk, line 711: Very nice tests!

augustjk (approved):
Just a few nits and comments but looks good to me!

sorvell:
Before committing to this direction, I think it's probably a good idea to have a plan for how to address async directives and server only templates.

Also, pushing this logic into controllers feels slightly odd since controllers may or may not have anything to do with being async or affecting server rendering.

It seems more natural to let an element itself make this decision. Random idea: await the result of update on the server.

```js
async update(changed) {
  if (isServer) {
    await this.task.taskComplete;
    await this.otherAsyncThingComplete;
  }
  super.update(changed);
}
```

AndrewJakubowicz:
The intention for this feature is to provide a low level utility which can later be used by controllers (such as `Task`) to do async work during SSR (without the user needing to modify their element).

The multiple highlighted async use-cases that this specific PR unlocks is mostly because Controllers will initially be the one place where async work can happen. We're starting there because we want to start proving async work out starting with Task. We suspect that there will be some subtlety in doing async work well, and so we'll want to package that up into reusable pieces, and controllers seem like a good way of doing that.

Is there concern that adding this capability in ReactiveControllers may block server directives or awaiting on the element?

sorvell:
The concern is about exposing the API at the right level where knowledge exists about what to do on the server. It seems like this makes more sense at the element level than the controller.

> The intention for this feature is to provide a low level utility which can later be used by controllers (such as Task) to do async work during SSR (without the user needing to modify their element).

Some tasks should be awaited for server rendering, others should not. Although this might be determined by a specialized task itself, it's more likely this depends on the setup in the consuming element... unless I'm missing something about the intended automation.

By analogy, you wouldn't add a feature to a self-driving car that allowed a wheel to automatically stop the car, instead you'd let it report speed and tire pressure so the system could decide what makes sense to do...

rictic:
@sorvell Yeah, not all Tasks will want to run on the server, that'll be a config. And agreed that elements should be able to do async work on the server too, but currently what does async work look like in an element? Because the server should be doing fundamentally the same thing as the client, just optionally it'll await it before first render. And right now our best pattern for async work on an element is the Task controller, so it makes sense to target that as the first API, IMO.

rictic (approved):
Excellent code, excellent tests!

I'd like to hear out Steve's perspective before merging, but this is otherwise good to me

sorvell:
I could easily see this becoming deprecated or just unnecessary once we have a  more comprehensive pattern for elements. IMO, this need means we need to design a way for elements to have an async trap before an update.

Towards that end...

Elements can schedule updates using `scheduleUpdate` and potentially do async work there, but it's not a convenient override point because it's not part of the controller lifecycle. 

This is why I suggested `update` as the spot. If we go that way, we'd want to change `performUpdate` to:

1. await the result of update if there is one.
2. return that result so the update itself can wait

Or something like that. 

Given that some design work seems warranted, I'm just a bit uncertain about pushing this through as is.

justinfagnani:
I don't think we want to make `performUpdate()` async. That brings all kinds of questions of what to do with concurrent updates (namely, continue and enqueue new update vs abort), and we're actively trying to move away from async `performUpdate()` by adding `scheduleUpdate()` and deprecating returning a Promise from `performUpdate()`.

> The concern is about exposing the API at the right level where knowledge exists about what to do on the server. It seems like this makes more sense at the element level than the controller.

I think both need to be involved, because we want the system to work correctly when an element uses a controller that must do async work for initial render. Task may be configurable, but there are other abstractions that might say fetch initial data asynchronously and always want to force a wait.

I share the concern about this API being ready to commit to though. Even on a bike-shedding axis, I'm wary of adding too many "server" specific names. Is there some more general concept we're really getting at there, like "initial"*?

So I think we should hold off on merging this a bit. I'd like to review (sorry, been very busy lately!) and I think we should include in the same branch *real* usage of the API, like updating the real Task implementation to use it, to prove that it suits at least the current needs we have. (and maybe we could make this a private API that only Task has access to initial to gather feedback?)

I'd also like to think through if there's a version of this API where the element is more in charge (but we don't make `performUpdate()` async). This might look like having a new method on ReactiveControllerHost to wait for initial render until a Promise resolves. It's not a hugely different, API really, it just puts the additional responsibility on the host and reduces how much SSR needs to know about controllers.

AndrewJakubowicz:
Great discussion here! To mark that this PR is not ready to merge I'm downgrading this PR back to a draft.

Thank you Justin and Steve for your reviews!

One option that Peter suggested was to add a `labs/server-task` package which adds the SSR functionality to Task (without impacting the stable Task package). I'll branch off this branch to make the server-task easier to review.

sorvell:
> I don't think we want to make performUpdate() async. That brings all kinds of questions of what to do with concurrent updates (namely, continue and enqueue new update vs abort)

IMO, we've already taken this on directly by having `scheduleUpdate` support asynchrony. There's already code to handle concurrent updates (that `scheduleUpdate`'s response participates in): the next update is blocked on the current update completing. 

AFAIR, we deprecated `performUpdate`'s asynchrony specifically because we added `scheduleUpdate`.

sorvell:
To respect `autoRun` and the `args` function, I think `ServerLoadedTask` would need to do something like:

```ts
class ServerLoadedTask extends Task implements ServerController {
  get serverUpdateComplete() {
    this.hostUpdate();
    if (this.status === TaskStatus.PENDING) {
      return this.taskComplete;
    }
  }
}
```

sorvell:
This is maybe more about `Task` than the `ServerController` interface, but one thing that bugs me is that the easy way to use `Task` typically results in a 2x update even when the task really shouldn't run. Consider this [slight modification of the playground example](https://lit.dev/playground/#gist=3482e280c9c6d064aae24f3b50f51aed). 

Here the package displayer is a separate element that takes the `packageName` as a property. It should only run the task if there's a `packageName` given. The code avoids the expensive fetch but does run the task and return the initial state, which provokes an additional update. While cheap, this isn't ideal and is probably worth avoiding, and if we do it on the server, that'll make the element render async, which again I think worth avoiding.

it's possible to finagle `autoRun` (commented in the example) to avoid this, but this requires involving the element, which would mean that it couldn't naively just use `ServerLoadedTask`.

justinfagnani:
@sorvell 

> > I don't think we want to make performUpdate() async. That brings all kinds of questions of what to do with concurrent updates (namely, continue and enqueue new update vs abort)
> 
> IMO, we've already taken this on directly by having `scheduleUpdate` support asynchrony. There's already code to handle concurrent updates (that `scheduleUpdate`'s response participates in): the next update is blocked on the current update completing.
> 
> AFAIR, we deprecated `performUpdate`'s asynchrony specifically because we added `scheduleUpdate`.

Yeah, async `scheduleUpdate()` is fine, but sync `performUpdate()` is very useful, and we've had custom discussions even this week around using that for some niche cases.

> This is maybe more about Task than the ServerController interface, but one thing that bugs me is that the easy way to use Task typically results in a 2x update even when the task really shouldn't run.

This is a side effect of the whole task function-calling call tree being async functions. We could rewrite things a bit so that `_performTask()`, `run()`, and the user's task function are _optionally_ async. That would allow returning `initialState` or any data, or throwing synchronously, which would play _slightly_ nicer with SSR.

It's a small optimization though, saving one microtask per task that has the next state synchronously available.

I think Task that works with the element being in control of waiting would look something like this, btw:

```ts
  async run(args?: T) {
    //...
    this.status = TaskStatus.PENDING;
    let result!: R | typeof initialState;
    let error: unknown;

    // Request an update to report pending state.
    if (this.autoRun === 'afterUpdate') {
      // Avoids a change-in-update warning
      queueMicrotask(() => this._host.requestUpdate());
      
      // afterUpdate tasks don't make sense in SSR, right?
      if (isServer) {
        console.log('skipping afterUpdate task in SSR');
      }
    } else {
      // get the host to report current task status like PENDING
      this._host.requestUpdate();
      // Tell the host to update after the task is complete?
      this._host.requestDeferredUpdate(this.taskComplete);
      // note, accessing this.taskComplete unconditionally like this defeats the lazy Promise creation we added
    }
  }
```

danielhjacobs:
Due to the use of the phrase "resolve https://github.com/lit/lit/issues/1434" in the PR description this PR is linked as fixing that issue, even though when reading the description fully it looks like that is not true.

PR #4398, opened 2023-11-15
Rename the serverhtml function to just html, in line with static html

rictic:
This makes three types of html function: normal, static, and server-only.

We want this named `html` so that syntax highlighting works by default, and to be consistent with the static `html` function.


packages/labs/ssr/README.md
```diff
@@ -178,16 +178,16 @@ that template can be loaded and hydrated on the client to apply the same data.
 
 ## Server-only templates
 
-You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags like `<script type="text/json">`. They are also slightly more efficient than normal Lit templates, because the generated HTML doesn't need to include markers for updating.
+`@lit-labs/ssr` also exports an `html` function for defining server-only templates. You can also write templates that will only render on the server. These templates can be used for rendering full documents, including the doctype, and rendering into elements that Lit normally cannot, like `<title>`, `<textarea>`, `<template>`, and non-executing `<script>` tags like `<script type="text/json">`. They are also slightly more efficient than normal Lit templates, because the generated HTML doesn't need to include markers for updating.
```
AndrewJakubowicz, line 181: Nit: "function" -> "tag function"?

packages/labs/ssr/src/test/integration/tests/basic.ts
```diff
@@ -42,7 +42,7 @@ import {
 
 import {anyHtml, SSRTest, SSRTestDescription} from './ssr-test.js';
 import {AsyncDirective} from 'lit/async-directive.js';
-import {serverhtml} from '../../../lib/server-template.js';
+import {html} from '../../../lib/server-template.js';
```
AndrewJakubowicz, line 45: On line 9 there is an import `html` from 'lit'. I think this should be `{html as serverhtml}`

packages/labs/ssr/src/test/test-files/render-test-module.ts
```diff
@@ -9,7 +9,7 @@ import {repeat} from 'lit/directives/repeat.js';
 import {classMap} from 'lit/directives/class-map.js';
 import {LitElement, css, PropertyValues} from 'lit';
 import {property, customElement} from 'lit/decorators.js';
-import {serverhtml} from '../../lib/server-template.js';
+import {html} from '../../lib/server-template.js';
```
augustjk, line 12: This one should also be `html as serverhtml` as it clashes with the regular `html`.

packages/labs/ssr/README.md
rictic: Rephrased!

packages/labs/ssr/src/test/integration/tests/basic.ts
rictic: Ah, yep. Fixed

packages/labs/ssr/src/test/test-files/render-test-module.ts
rictic: Ah, yep, fixed

augustjk (approved):
🎉

AndrewJakubowicz (approved):

PR #4409, opened 2023-11-20
async-replace correctly re-renders when value is unchanged

### LINKED ISSUE #4408: [lit-html] Re-rendering an `asyncReplace` clears its value.

### Description

When a template using an `asyncReplace` is re-rendered without changing the iterable or its current value in the `asyncReplace`, the value is incorrectly cleared.

### Reproduction

https://lit.dev/playground/#gist=257917d929b16bd6f3109c81f121b992

Click the `requestUpdate` button

Expected:  the asyncReplace value should not change
Actual: the asyncReplace value is cleared (it is replaced if the iterable emits another value)

### Workaround

None.

### Is this a regression?

### Affected versions

3.x

All

sorvell:
Fixes #4408.

### DIFF
--- a/packages/lit-html/src/test/directives/async-replace_test.ts
+++ b/packages/lit-html/src/test/directives/async-replace_test.ts
@@ -184,9 +184,7 @@ suite('asyncReplace', () => {
     render(t(iterable), container);
     assert.equal(stripExpressionMarkers(container.innerHTML), '<div></div>');
 
-    const wait = iterable.push('hello');
-    render(t(iterable), container);
-    await wait;
+    await iterable.push('hello');
     assert.equal(
       stripExpressionMarkers(container.innerHTML),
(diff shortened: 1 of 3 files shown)


augustjk (approved):
Nice!

packages/lit-html/src/test/directives/async-replace_test.ts
```diff
@@ -179,6 +179,30 @@ suite('asyncReplace', () => {
     assert.equal(stripExpressionMarkers(container.innerHTML), '<div>bar</div>');
   });
 
+  test('renders the same itetable value when re-rendered with no new value emitted', async () => {
```
AndrewJakubowicz, line 182:
```suggestion
  test('renders the same iterable value when re-rendered with no new value emitted', async () => {
```

PR #4413, opened 2023-11-22
Remove unused internal parameters to requestUpdate

justinfagnani:
I noticed this while re-reading some of the code. This was for distinguishing initial values while we were still trying to not reflect initial values for standard decorators.

### DIFF
--- a/.changeset/odd-buckets-attend.md
+++ b/.changeset/odd-buckets-attend.md
@@ -1,5 +1,7 @@
 ---
 '@lit/reactive-element': patch
+'lit-element': patch
+'lit': patch
 ---
 
 Remove unused internal parameters to `requestUpdate()`
--- a/.changeset/odd-buckets-attend.md
+++ b/.changeset/odd-buckets-attend.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Remove unused internal parameters to `requestUpdate()`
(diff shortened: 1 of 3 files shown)


.changeset/odd-buckets-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
```
AndrewJakubowicz, line 2: reactive-element patch should also patch `lit`:

```suggestion
'@lit/reactive-element': patch
'lit': patch
```

AndrewJakubowicz (approved):
LGTM!

Thank you. Just need to update the size and changeset!

.changeset/odd-buckets-attend.md
augustjk: I think we should also do `lit-element` as it depends on `@lit/reactive-element`

augustjk (approved):

.changeset/odd-buckets-attend.md
AndrewJakubowicz: Updated suggestion. Thank you!

AndrewJakubowicz:
I addressed code review feedback on changeset and Lit expected size.
There was a minor GitHub outage that blocked Actions from running so I created an empty commit to kickstart the actions.
Once they pass I'll merge.

.changeset/odd-buckets-attend.md
AndrewJakubowicz: done

Issue #4417, opened 2023-11-29
[ssr] invalid locations error when using server only templates


thescientist13:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

Per [this thread](https://github.com/lit/lit/issues/2441) on use cases for Lit SSR, the takeaway was that a special server only template function has now been made available that would relax some of the restrictions on the client side version of `html`, thus allowing something like [this](https://github.com/lit/lit/issues/2441#issuecomment-1816903951) to be server rendered now
```js
import {render, html} from '@lit-labs/ssr';

const template = ({ title, lang }) => html`
  <!DOCTYPE html>
  <html lang="${lang}">
    <head>
      <!-- Works within a server only template -->
      <title>${title}</title>
    </head>
    <body>
      <p>${title}</p>
    </body>
  </html>
`;
```

However, based on my experiments
```js
import { render, html } from '@lit-labs/ssr';
import { collectResult } from '@lit-labs/ssr/lib/render-result.js'

const template = ({ title, lang }) => html`
  <!DOCTYPE html>
  <html lang="${lang}">
    <head>
      <!-- Works within a server only template -->
      <title>${title}</title>
    </head>
    <body>
      <p>${title}</p>
    </body>
  </html>
`;

const ssrResult = render(template({ title: 'My Website', lang: 'en'}));
const ssrContent = await collectResult(ssrResult);

console.log({ ssrContent });
```

doing this will result in an [invalid locations error](https://lit.dev/docs/templates/expressions/#invalid-locations)
```sh
file:///Users/owenbuckley/Workspace/github/lit-v3-ssr-demo/node_modules/@lit-labs/ssr/lib/render-value.js:605
        throw new Error(errorMsg);
              ^
Error:
    Unexpected final partIndex: 2 !== 3 while processing the following template:


  <!DOCTYPE html>
  <html lang="${...}">
    <head>
      <!-- Works within a server only template -->
      <title>${...}</title>
    </head>
    <body>
      <p>${...}</p>
    </body>
  </html>


    This could be because you're attempting to render an expression in an invalid location. See
    https://lit.dev/docs/templates/expressions/#invalid-locations for more information about invalid expression
    locations.
```

### Reproduction

I have a repro repo setup here you can check out
https://github.com/thescientist13/lit-v3-ssr-demo/tree/master#document-%EF%B8%8F

Just clone, run `npm ci` and then run `npm run demo:document`.  The relevant code is in _document.js_

### Workaround

[...]

PR #4418, opened 2023-11-29
[labs/ssr] allow binding attributes to html tag when using server templates

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4417

### Why

Server templates reuse a lot of the code paths of non server templates. In the server we convert templates to an HTML ast with parse5's `parseFragment` function.

However, this results in the following:

```html
<script type="module">
import {parseFragment, serialize} from 'https://cdn.jsdelivr.net/npm/parse5@7.1.2/+esm';

console.log(serialize(parseFragment("<html lang='ko'><p>Content</p></html>")));
// → <p>Content</p>
</script>
```

The attribute part thus cannot be added, and a part mismatch error is thrown.

### Fix

For server templates, use parse5's `parse` function. This handles document tags such as `html`. In fact it puts everything in a document.

```html
<script type="module">
import {parse, serialize} from 'https://cdn.jsdelivr.net/npm/parse5@7.1.2/+esm';

console.log(serialize(parse("<html lang='ko'><p>Content</p></html>")));
// → <html lang="ko"><head></head><body><p>Content</p></body></html>
</script>
```

### Test plan

Added a regression test, and a composition test that shows an extreme level of server template composition. This ensures that using parse5's `parse` doesn't cause issues when composing server templates.

### Questions

What about those extra nodes that parse5 creates? Such as `<body>` in the example above?
  - If we were generating comment nodes (which we are not with server templates), the node index would be wrong. This isn't a problem with server templates as they result in a string of html without comment node metadata.

Do those extra document nodes get sent down the wire?
  - They do not–and the unit tests confirm this.

### DIFF
--- a/.changeset/lazy-beans-tap.md
+++ b/.changeset/lazy-beans-tap.md
@@ -1,0 +1,6 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix server template throwing when encountering an attribute binding on the
+`html` tag. This is now handled correctly.
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -51,7 +51,7 @@ import {
 
 import {escapeHtml} from './util/escape-html.js';
 
-import {parseFragment} from 'parse5';
+import {parseFragment, parse} from 'parse5';
 import {
   isElementNode,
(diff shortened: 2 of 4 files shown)


augustjk (approved):
LGTM

PR #4429, opened 2023-12-05
fix: support for className with whitespace and so on

leemove:
fix  https://github.com/lit/lit/issues/3095 

use classname. a lib be used in vue and react.

### DIFF
--- a/packages/lit-html/package.json
+++ b/packages/lit-html/package.json
@@ -490,7 +490,8 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


augustjk:
Thanks for the PR but we shouldn't have to add a new dependency for this, and this wouldn't work on updates either.

We need to discuss whether the solution presented in https://github.com/lit/lit/pull/3096 works and doesn't add too much weight.

Issue #4435, opened 2023-12-06
Property values are not available in the connectedCallback when using createComponent from @lit/react


vonsa:
### Which package(s) are affected?

React (@lit/react)

### Description

The properties are only assigned on useLayoutEffect which is too late to make them available inside of the connectedCallback.

I've solved it, adding this issue to link to my PR.

### Reproduction

If you need this even with the PR, let me know

### Workaround

Fixed it inside of @lit/react

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

3.1.0

### Browser/OS/Node environment

Browser: Chrome
OS: Mac (M1 Pro)

augustjk:
Related: https://github.com/lit/lit/issues/4023#issuecomment-1792627956

We don't want to set all props coming from JSX as attributes by passing them to `React.createElement`, as not all properties are meant to be attributes nor can be stringified, i.e. you get things like `[object Object]` as attribute value. Unfortunately that makes it so that properties are only set after the element is mounted as that's when we can get a ref to imperatively set them.

The linked comment above suggests maybe setting non-object props as attributes for the initial render so it's available to the underlying element for first render and connected. It's not a perfect solution though as it's sprouting potentially unneeded attributes.

I don't know if React ever gives an opportunity for ref access before it's mounted for the initial render. Certainly not for function components. I need to test if https://react.dev/reference/react/Component#getsnapshotbeforeupdate lifecycle for class components might do that.

vonsa:
I've updated https://github.com/lit/lit/pull/4436, for me currently not a priority to contribute this.

Some additional thoughts that came to mind:
- Use native HTML/JavaScript instead of React (I don't know intricate details of the frameworks, so not sure if this is viable)
- Wrap the React component in a lit component that uses firstUpdated to render the actual component and pass the props inside of the web component itself (seems too complex and not close enough to the native browser implementation)

Only adding the string attributes can be a partial solution, but I'm hesitant to implement it because it can also add more confusion on why only strings are working and it does not solve the root problem.

augustjk:
Confirmed that React class component life cycle `getSnapshotBeforeUpdate` (which replaces `componentWillUpdate`) will not work since it does not get called before first render. Without React giving us a ref before it's mounted, the only way we can control that is by imperatively creating the underlying custom element ourselves but that adds a whole other set of complications.

[...]

PR #4436, opened 2023-12-06
fix: Set lit component properties before rendering to the DOM when using @lit/react createComponent

vonsa:
[#4435](https://github.com/lit/lit/issues/4435)

**Solution:**
Adding the elementProps when using React.createElement makes sure that the properties are set before the connectedCallback runs.

**Notes:**
By adding this change, tests in @lit/labs/ssr-react were breaking because now the user inputted properties will show up on the element when you call ReactDOMServer.renderToString.

If you have any questions or suggestions please let me know.

### DIFF
--- a/.changeset/tall-tools-pay.md
+++ b/.changeset/tall-tools-pay.md
@@ -1,0 +1,6 @@
+---
+'@lit-labs/ssr-react': patch
+'@lit/react': patch
+---
+
+Set lit component properties before rendering to the DOM when using @lit/react createComponent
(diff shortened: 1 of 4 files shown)


tbroyer:
This is changing the [intended](https://lit.dev/docs/frameworks/react/#why-are-wrappers-needed) and [documented](https://lit.dev/docs/frameworks/react/#how-it-works) behavior by treating everything (except events) as attributes, which:
* will break if:
  * values aren't strings or easily convertible to/from string (converter might not be able to parse the value)
  * the attribute and property don't have the same name (will break your use-case here, as it would set an attribute with the name of the property, so the property wouldn't have been set at the time of `connectedCallback` either)
* will double the work for properties, as they'd be set as attributes **and** then handled again as properties in the `useLayoutEffect` callback

vonsa:
Alright, I haven't found any simple solution, maybe any of you know but I will be dropping this as I have different priorities. Should I close this PR?

PR #4439, opened 2023-12-07
Set up token permissions for GitHub Workflows

### LINKED ISSUE #4438: Set up token permissions for GitHub Workflows
### Should this be an RFC?

### Description

Hi again (previous issues #3668). 

Still regarding the topic on "securing the CI", I'd like to suggest to limit the permissions given to the workflows used.

My suggestion here is basically set up the permissions to the github workflows as read only on the [top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) and any write permission be given at the [run level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions).

This is necessary due to a behavior of github workflow to grant to GITHUB_TOKEN **write** permissions to all types of permissions, regardless of they being used or not. In case of the workflow getting compromised, an attacker can exploit this permissions.

Thus, it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

I'll soon submit a PR to show the changes I listed above. Please, feel free to reach me out in case of any doubts or concerns.

It is possible to see the permissions granted to the workflow looking into the GITHUB_TOKEN permissions line on the Set up step: https://github.com/lit/lit/actions/runs/7107353937/job/19348624321#step:1:17

joycebrum:
Closes #4438

Workflow runs with the given permissions:

- benchmarks.yaml https://github.com/joycebrum/lit/actions/runs/7120241683
- changeset-status.yaml https://github.com/joycebrum/lit/actions/runs/7120241687
- release-image.yaml https://github.com/joycebrum/lit/actions/runs/7120191834
- benchmarks-report.yaml https://github.com/joycebrum/lit/actions/runs/7120322262
- sizecheck-report.yaml https://github.com/joycebrum/lit/actions/runs/7120268829
- sizecheck.yaml https://github.com/joycebrum/lit/actions/runs/7120241690
- tests.yaml https://github.com/joycebrum/lit/actions/runs/7120235759

The remaining workflows I did not run since it is not supposed to work on forks. Let me know if I missed anything.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,9 +5,14 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   release:
     # Don't run on forks. We can and should only release from the main repo.
+    permissions:
+      contents: write  # for Git to git push
     if: github.repository == 'lit/lit'
 
     name: Release
--- a/.github/workflows/benchmarks-report.yaml
+++ b/.github/workflows/benchmarks-report.yaml
@@ -8,11 +8,17 @@ on:
       - completed
       - requested
 
+permissions:
+  contents: read
+
 jobs:
   # Optional job to update existing comments with a "benchmarks are running" text
   report_running:
     name: Report benchmarks are in-progress
     runs-on: ubuntu-latest
+    permissions:
+      issues: write # needed to write comments on PRs 
+      pull-requests: write # needed to write comments on PRs 
     # Only add the "benchmarks are running" text when the workflow_run starts
     if: ${{ github.event.action == 'requested' }}
     steps:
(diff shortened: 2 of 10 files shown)


.github/workflows/release.yaml
```diff
@@ -5,9 +5,14 @@ on:
...
       - main
 
+permissions:
+  contents: read
+
 jobs:
   release:
     # Don't run on forks. We can and should only release from the main repo.
+    permissions:
+      contents: write  # for Git to git push
```
augustjk, line 15: This workflow also creates a PR like this one https://github.com/lit/lit/pull/4399 via `changesets/action`, would it require `pull_requests: write` ?

augustjk (commented):
Thank you very much! One question below with the `release.yaml` workflow which only runs on approval from `main` branch.

.github/workflows/release.yaml
joycebrum: Yeah, it would 
I'll update it.

joycebrum: Fix it. Thanks!

augustjk (approved):
Thanks!

PR #4444, opened 2023-12-10
[labs/compiler]: Support for different lit html sources

### LINKED ISSUE #4269: [compiler] Make lit-html source configurable
### Should this be an RFC?

### Description

According to the docs:
image

> "Re-exports of html from other modules are not supported."

This will be an issue for us at ING, because our design system re-exports all lit stuff, and teams are expected to consume lit from there instead.

In `eslint-plugin-lit-a11y`, we make the `litHtmlSource` configurable:
image

Would it be possible to do something similar? e.g.:

```js
compileLitTemplates({
  litHtmlSources: ['our-design-system']
})
```

Unknown

redfox-mx:
Add some interfaces to allow compile re exported html sources.

Fixes #4269

### DIFF
--- a/packages/labs/compiler/src/lib/template-transform.ts
+++ b/packages/labs/compiler/src/lib/template-transform.ts
@@ -65,6 +65,18 @@ interface TemplateInfo {
   variableName: ts.Identifier;
 }
 
+export interface CompileLitTemplateOptions {
+  /**
+   * Allows you to use `html` imported from modules other than lit.
+   * It is especially useful when re-exporting `html` tagged templates.
+   */
+  litHtmlSources?: string[];
+}
+
+export interface CompiledTemplatePassOptions {
+  allowedImports?: string[];
+}
+
 /**
  * Matches the raw text elements.
  *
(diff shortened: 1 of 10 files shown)


packages/labs/compiler/src/lib/template-transform.ts
```diff
@@ -65,6 +65,18 @@ interface TemplateInfo {
   variableName: ts.Identifier;
 }
 
+export interface CompileLitTemplateOptions {
+  /**
+   * Allows you to use `html` imported from modules other than lit.
+   * It is especially useful when re-exporting `html` tagged templates.
+   */
+  litHtmlSources?: string[];
```
AndrewJakubowicz, line 73: This is a great improvement! One open question is whether we should allow more than a static string here, e.g. maybe a regex?

How are relative module imports like the following handled?

```js
import {myHtml as html} from './reexport.js';
```

redfox-mx: > One open question is whether we should allow more than a static string here, e.g. maybe a regex?

I think, It probably be useful in the future. But now I can't think about a specific use case. Maybe for some bundlers, usually import statements are transformed and could be "useful" to tackle this transformation. I'm not really sure :C 

> How are relative module imports like the following handled?

For typescript './reexport.js' is the given importer specifier, so if you need to be compiled, You need to specify './reexport.js' (extension is important) as a litHtmlSource.

```js
import typescript from '@rollup/plugin-typescript';
import {compileLitTemplates} from '@lit-labs/compiler';

export default {
  // ...
  plugins: [
    typescript({
      transformers: {
        before: [
          compileLitTemplates({ litHtmlSources: './reexport.js' })
        ],
      },
    }),
    // other rollup plugins
  ],
};

``` 

Maybe some scenarios with relative import for the same re-exported html template can be suitable for regex... but usually, these cases are covered by import aliases.

What do yo think about use `createFilter` from [@rollup/plugin-utils](https://github.com/rollup/rollup-pluginutils) ?

redfox-mx: update: maybe we can support micromatch expressions and regex as createFilter, and not all rollup specific cases

redfox-mx:
Hi @AndrewJakubowicz!

I close this PR since import specifier now is delegated to lit-analizer, so, for now, it does not make sense for me. Thanks for your time. Bye ✋🏼

PR #4451, opened 2023-12-12
Fix security alerts:

rictic:
These three seem legit, though unlikely to be exploited.

https://github.com/lit/lit/security/code-scanning/3

https://github.com/lit/lit/security/code-scanning/2

https://github.com/lit/lit/security/code-scanning/13


augustjk (approved):
Thank you!

PR #4456, opened 2023-12-14
[rollup-plugin-minify-html-literals] update to rollup ^4.0.0

vdegenne:


packages/labs/rollup-plugin-minify-html-literals/package.json
```diff
@@ -86,7 +86,7 @@
     "typescript": "^2.9.2 || ^3.0.0 || ^4.0.0 || ^5.0.0"
   },
   "peerDependencies": {
-    "rollup": "^0.65.2 || ^1.0.0 || ^2.0.0 || ^3.0.0"
+    "rollup": "^0.65.2 || ^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0"
```
augustjk, line 89: Thank you for the PR! I wonder if we might remove support for some older versions of rollup before publishing. I have no idea if the plugin as it is even works with the older versions. I have a vague memory that some changes were required to support v3.. no clue if that still makes it backwards compatible. It would be good if we have integration tests for major versions of rollup that this should work with.That could be separate PR too.

vdegenne: Correct, some changes have been made for v3 support (https://github.com/asyncLiz/rollup-plugin-minify-html-literals/commit/470561c035b495e3d175ff5a6450e59a94400d94).
All I can tell is it works for v3 and v4 but you're right backward compatibility could be broken.
How do you even start testing these things?

Issue #3572, opened 2023-01-17
[labs/virtualizer] Maximum size of virtualizer is artificially limited


usergenic:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

At some point in the past, we ran into strange (and bad) behaviors in one or more browsers when sizing a virtualizer beyond a certain number of pixels.

As a clumsy workaround, we are currently clamping the maximum height to an arbitrary value that was determined through ad hoc experimentation.

We need to investigate more deeply to see if we can solve this problem in a fundamental way, or (failing that) at least come up with a better workaround.

Note this issue was copied from the prior uni-virtualizer repository https://github.com/PolymerLabs/uni-virtualizer/issues/140

### Reproduction

Load this playground example: https://lit.dev/playground/#gist=346adebe77f9950d2fbc90e6758f4d79
Scroll all the way to the bottom of the list in the playground's Result pane (you'll probably want to do this by dragging the thumb)

Expected Result:

You can scroll all the way to the 1,000,000th item

Actual Result:

You can't

### Workaround

Have not determined whether a workaround is necessary or possible.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

0.7.2

### Browser/OS/Node environment

TBD

lehwark:
This is still an issue - tested in Firefox and Chrome. I am trying to virtualize a list with ~ 1 million items and depending on the height of the items, I can only display ~ 200.000. Is there any workaround?

scherler:
It seems it depends as well on the browser you are using
![image](https://github.com/lit/lit/assets/596701/1b1732b0-f839-4d21-81fc-09ce1d044743)

PR #4458, opened 2023-12-19
[raise_max_size] Maximum size of virtualizer is artificially limited

scherler:
- [ ] measure the max height and use that number instead of the static limit
- [ ]  see what the item max per browser are
- [ ] find a way to still render everything that is not in the limits (e.g. shrink artificially height so all fits in or something else)

https://github.com/lit/lit/issues/3572


scherler:
image
^ chrome now does the 1m 
image
so does opera
image
however, FF does less than before

https://lit.dev/playground/#gist=0cabcba351eefd12c113302a3d1f6d5c

scherler:
image
There is a hard cut in FF

PR #4461, opened 2023-12-21
Adds optional error callback to `customElement` decorator; provides `define` helper

### LINKED ISSUE #4455: Provide assistance for dealing with conflicting element definitions
### Should this be an RFC?

### Background

Custom elements have the super power of interoperability; they work anywhere the web does. There's just one big problem: the global registry. 

Scoped custom elements registries are the (hopefully medium term) platform answer and there's a [polyfill](https://www.npmjs.com/package/@webcomponents/scoped-custom-element-registry) and [helper](https://open-wc.org/docs/development/scoped-elements/) [libraries](https://github.com/lit/lit/tree/main/packages/labs/scoped-registry-mixin).

There are, however, a few use cases that are still problematic:
1. conflicting elements in a scope, especially and perhaps exclusively **_the global scope_**.
2. use in legacy systems (aka existing web **_frameworks_**) that do not use new scoped APIs, `shadowRoot.createElement / importNode`
3. environments that **_cannot rely on a polyfill_** due to maintenance concerns or just policy restrictions.

If not addressed, these issues can become barriers to adoption as has been evident in [recent](https://discord.com/channels/1012791295170859069/1016088590373101689/1183015266024689714) [feedback](https://discord.com/channels/1012791295170859069/1182032821682982983).

### Current

Lit [recommends](https://lit.dev/docs/tools/publishing/#self-define-elements) defining elements using the `@customElement` decorator which calls `customElements.define` and generates an exception if a name is in use. This halts script execution, and when a large amount of script is bundled, this can easily result in the entire page becoming non-functional.

### Proposal

Provide a `define` module which would be used by `@customElement`. This is inspired by and similar to the utility in [Shoelace](https://github.com/shoelace-style/shoelace/blob/next/src/internal/shoelace-element.ts#L101-L131).

[Prototype implementation](https://lit.dev/playground/#gist=6e0c9cfc921ffa53de893f07492c6a9d). Summary:

* `define(name, ctor, onError)`: tries to define an element, trapping exceptions and calling an error callback `onError(error, name, ctor)`. 
* Possibly the `onError` would also have a default implementation which called `console.error(error)`. This provides (scary) error feedback but does not halt script execution.
* `assertPackageVersion(name, ctor)`: an onError optional helper which does a `package`/`version` comparison on the existing and new constructors and issues a warning or error. For example `MdButton` might have package = `@material/web`,  `version = '1.1.0'` which could silently be defined if an existing package = `@material/web`,  `version = '1.1.1'` was defined but might error if a `version = '2.0.0'` was defined.

1. scoped custom element registries.
2. require users to resolve dependencies and make their own builds

sorvell:
Fixes #4455.

* Adds optional error callback to `customElements` decorator.
* Provides a `define` helper with an optional error callback for javascript use.
* Provides a `warn` error callback which issues a `console.warn`
* Provides a `warnPackageVersion` error callback which issues a `console.warn` when an incompatible package and/or version is defined.
* A provided error callback is called when `customElements.deine` fails with the `error`, `tagName`, and `constructor`; providing any error callback catches the define exception.


.changeset/small-pots-rest.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit/reactive-element': patch
+'lit': patch
+'lit-element': patch
+---
```
augustjk, line 5: This is a new feature so maybe it warrants a minor bump rather than patch?
```suggestion
---
'@lit/reactive-element': minor
'lit': minor
'lit-element': minor
---
```

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+/**
+ *
+ * @param tagName The tag name of the custom element to define.
+ * @param ctor The constructor of the custom element to define
+ * @param onError An optional function which can handle a define exception
+ * error. If provided, the exception is caught.
+ * @param onError
+ * @returns void
+ */
+export const define = (
```
augustjk, line 37: This could be used by users that aren't using the decorator but they'd have to import it from `lit/decorators/custom-element.js` or `lit/decorators.js`.

Would it maybe make sense to move `define`, `warn`, and `warnPackabeVErsion` to a separate submodule like `@lit/reactive-element/define.js`, which will be mirrored in `lit-element` and `lit`?

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
 
 import type {Constructor} from './base.js';
 
+declare global {
+  interface CustomElementRegistry {
+    getName: (ctor: CustomElementConstructor) => string | null;
```
justinfagnani, line 18: This won't be needed after we update TypeScript.

packages/reactive-element/src/decorators/custom-element.ts
justinfagnani: I think we should do exactly that.

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+  }
+}
+
+export type DefineErrorCallback = (
+  error: Error,
+  tagName: string,
+  ctor: CustomElementConstructor
```
justinfagnani, line 25: I think tag name and constructor should be on the error object. We can make an Error subclass for this.

We also need both the existing and new constructors on the error so that handlers can introspect them for compatibility.

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+ * @param ctor The constructor of the custom element to define
+ * @param onError An optional function which can handle a define exception
+ * error. If provided, the exception is caught.
+ * @param onError
+ * @returns void
+ */
+export const define = (
+  tagName: string,
+  ctor: CustomElementConstructor,
+  onError?: DefineErrorCallback
```
justinfagnani, line 40: I would put the `onError` callback in an options object. And we might want to pass that options object to the callback. This would let users place metadata in the object, like version number:

```ts
import {customElement, warnOnIncompatibleVersions} from 'lit/decorators.js';

@customElement('x-foo', {onError: warnOnIncompatibleVersions, version: '1.0.0'}
export class XFoo extends LitElement {}
```

Authors would be responsible for keeping the version in sync, hopefully with a build tool.

We could vend a common sentinel value to use as the version so that common tools could be made to replace it:

```ts
import {customElement, warnOnIncompatibleVersions, packageJsonVersion} from 'lit/decorators.js';

@customElement('x-foo', {onError: warnOnIncompatibleVersions, version: packageJsonVersion}
export class XFoo extends LitElement {}
```

I think when the version is packageJsonVersion, the comparison _always_ fails, so that in dev you get the warnings and in prod you get them unless you set up your build system correctly.

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+) => {
+  if (customElements.getName?.(ctor) === tagName) {
```
justinfagnani, line 42: This should be the same and has wider support:

```suggestion
  if (customElements.get(tagName) === ctor) {
```

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+    return;
+  }
+  try {
+    customElements.define(tagName, ctor);
+  } catch (error) {
+    if (onError === undefined) {
+      throw error;
+    } else {
+      onError?.(error as Error, tagName, ctor);
```
justinfagnani, line 51:
```suggestion
      onError(error as Error, tagName, ctor);
```

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+  ctor: CustomElementConstructor,
+  onError?: DefineErrorCallback
+) => {
+  if (customElements.getName?.(ctor) === tagName) {
+    return;
+  }
+  try {
+    customElements.define(tagName, ctor);
+  } catch (error) {
+    if (onError === undefined) {
```
justinfagnani, line 48: I think we should consider firing some kind of ErrorEvent on window (the owner of the registry - we could do this on the scope in the future), and checking for defaultPrevented. This should be a community protocol, but we could ship first and report back how well it works before doing that.

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+    }
+  }
+};
+
+export const warn = (error: Error) => {
+  console.warn(error);
+};
+
+// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+const semVerRe =
```
justinfagnani, line 61: I'd love to see a test of this with just a bunch of version strings and assertions on the parsed parts.

packages/reactive-element/src/decorators/custom-element.ts
```diff
@@ -13,6 +13,125 @@
...
+  _error: Error,
+  tagName: string,
+  ctor: CustomElementConstructor
+) => {
+  const {package: definedPackage, version: definedVersion} =
+    (customElements.get(tagName) ?? {}) as PackageVersion;
+  const {package: newPackage, version: newVersion} = ctor as PackageVersion;
+  const packageOk = definedPackage === newPackage;
+  let versionOk = definedVersion === newVersion;
+  // versions match if using semver and "patch" is only diff.
```
justinfagnani, line 107: This is tricky. If the existing version is the same major and an equal or greater minor, they should be compatible.

But semver ranges are much more expressive, and without them we limit the expressiveness here. Pre-release versions do not follow the same rules either.

I'd be more comfortable if the semver stuff was left out of the first version, and maybe put in a labs package. We can vend a function that simply warns if the versions aren't exactly the same for now, or no default error handler at all.

justinfagnani (changes requested):
Awesome @sorvell!

PR #4466, opened 2023-12-27
fix[@lit-labs/gen-wrapper-vue]: move rollup variabels to valid position

tibnor:
In @lit-labs/gen-wrapper-vue the generated vite.config.ts was not valid. This PR corrects the invalid values. See rollup documentation here: https://rollupjs.org/configuration-options/#preserveentrysignatures

I'm unsure what the intended preserveEntrySignatures value should be. But "true" is not a valid value.

Current implementation produces this warning during build of the generated wrapper project.
vite v3.2.7 building for production...
Unknown input options: preserveModulesRoot. Allowed options: acorn, acornInjectPlugins, cache, context, experimentalCacheExpiry, external, inlineDynamicImports, input, makeAbsoluteExternalsRelative, manualChunks, maxParallelFileOps, maxParallelFileReads, moduleContext, onwarn, perf, plugins, preserveEntrySignatures, preserveModules, preserveSymlinks, shimMissingExports, strictDeprecations, treeshake, watch


augustjk (changes requested):
Good catch, thank you for the fix! I do observe the vite build warning about unknown config option even during our test. It was just silently being ignored.

The test goldens should be updated here: https://github.com/lit/lit/blob/25fbfba9c0f1b97d720a981831c59c08472ba6ee/packages/labs/gen-wrapper-vue/goldens/test-element-a/vite.config.ts#L20-L28

augustjk (approved):
Thank you!

PR #4473, opened 2024-01-03
Add a warning when binding this.requestUpdate to an event listener

rictic:
This is an uncommon use case to do in real code, but a reasonable thing to try when experimenting and debugging, so we want to help someone out with a warning here.

### DIFF
--- a/.changeset/ninety-bananas-sneeze.md
+++ b/.changeset/ninety-bananas-sneeze.md
@@ -1,5 +1,7 @@
 ---
 '@lit/reactive-element': patch
+'lit-element': patch
+'lit': patch
 ---
 
 Add a warning in dev mode when binding this.requestUpdate directly as an event listener.
--- a/.changeset/ninety-bananas-sneeze.md
+++ b/.changeset/ninety-bananas-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
+
+Add a warning in dev mode when binding this.requestUpdate directly as an event listener.
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1246,6 +1246,12 @@ export abstract class ReactiveElement
   ): void {
     // If we have a property key, perform property update steps.
     if (name !== undefined) {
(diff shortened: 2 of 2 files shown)


.changeset/ninety-bananas-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
```
augustjk, line 3: we should bump core packages that depend on reactive-element
```suggestion
---
'@lit/reactive-element': patch
'lit-element': patch
'lit': patch
---
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1246,6 +1246,12 @@ export abstract class ReactiveElement
   ): void {
     // If we have a property key, perform property update steps.
     if (name !== undefined) {
+      if (DEV_MODE && (name as unknown) instanceof Event) {
+        issueWarning(
+          `request-update-event`,
```
augustjk, line 1251: We should leave this short code as an empty string unless we intend to actually have a page on lit.dev for it. I do think it's nicer to have a short code if this needs further clarification over the accompanying message. If we intend to keep it, we should make note of it here: https://github.com/lit/lit.dev/issues/1262

augustjk (commented):
Seems like a hyper specific case to be looking for, but I suppose there's no harm in adding just a dev mode warning.

packages/reactive-element/src/reactive-element.ts
rictic, line 1251:
```suggestion
          ``,
```

rictic: Ah, good point. Since this is a super specific thing that's fairly self-explanatory when you do hit it, I don't think we need a page on lit.dev for it. Removed the short code.

rictic:
> Seems like a hyper specific case to be looking for, but I suppose there's no harm in adding just a dev mode warning.

It is, but the check is easy, and the behavior was weird enough that it had me stumped for a bit.

augustjk (approved):

AndrewJakubowicz (approved):
This is very subtle. Nice!

Issue #4471, opened 2024-01-02
Evaluate replacing Sauce Labs with Github Actions


rictic:
There's some friction with Google Legal and the Sauce Labs, and honestly the sauce tunnel situation has been a steady stream of distracting issues for years. Is GitHub Actions good enough to replace it? This issue tracks investigating that.

justinfagnani:
I'd love to record our understanding of how Playwright engine versions relate to released browser versions and how we can test all of our supported versions, especially the older versions.

augustjk:
With playwright, the browser version is tied to playwright version and is not configurable. https://playwright.dev/docs/browsers
You can see what browser versions come with playwright in their release note https://playwright.dev/docs/release-notes#browser-versions
To test across different browser versions would mean installing different versions of playwright within GHA before running the test command. This also means we're not really testing on latest browsers unless we update playwright.

webdriver.io appears to have the ability to configure versions for chrome and firefox. https://webdriver.io/docs/capabilities/#automate-different-browser-channels
but Safari seems to be limited to the technology preview version that must be installed on the machine.

justinfagnani:
What about BrowserStack? Is that any better than Sauce?

augustjk:
Also costs $ and will require some approval process. Unsure if legal needs to be involved if we want to apply for the open source license for either BrowserStack or SauceLabs.

PR #4474, opened 2024-01-03
[infra] Remove Sauce workflow

augustjk:
Our sauce key currently does not work causing the workflow to always fail.
~~I opted to convert the trigger to be manual rather than deleting the whole workflow for now in case we want to re-enable it sometime soon, but we can always just delete and recreate this from the git history too.~~
This PR now deletes thew sauce workflow yaml file.


AndrewJakubowicz (approved):

rictic (commented):
We have source control, no need to comment things out. IMO let's just delete it.

Also, related context: https://github.com/lit/lit/issues/4471

AndrewJakubowicz:
Woot!

PR #4476, opened 2024-01-04
feat: allow to set props as attributes in react wrapper

### LINKED ISSUE #4023: [labs/react]Using @lit-labs/react to wrap components in an SSR environment(Remix) will result in attributes not being immediately passed to the component

React (@lit-labs/react)

### Description

image
image

Attributes will pass to the components which wrapped by `CreateComponent` after about 500ms, it will make a obvious flash.

### Reproduction

https://stackblitz.com/edit/remix-run-remix-nhexf3?file=app%2Fcomponent%2Fbutton.js

### Workaround

Use components directly?

### Is this a regression?

### Affected versions

@lit-labs/react: 1.2.1; lit: 2.7.6;

node: v19.9.0
react: 17.x or 18.x

jeripeierSBB:
Fixes #4023

It's an attempt to bring #4023 further, no final solution. 

In the current solution there is a probably unaccepted way to access the properties to attribute map. How could this be done better?
Also, tests are missing, but I'd provide them if architecture of solution would be clear.


justinfagnani:
Thanks for the PR!

I have some questions about the problem and potential fix in #4023

Before committing to this I want to make sure that this actually solves the FOUC problem with Remix, so we'd need a test that fails on the current observable behavior (not just a test that attributes are present). And I'd want to see if there's another solution, like maybe not using a layout effect.

Issue #4478, opened 2024-01-05
Component attributes don't render properly when using refs in combination with SSR


mwildeboer:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element), SSR (@lit-labs/ssr)

### Description

While upgrading Lit 2.7.2 to the latest version (3.1.0) we encountered a bug. We render our components on the server using the `@lit-labs/ssr` package and for testing the ssrFixture from `@lit-labs/testing/fixtures.js`  

We encountered a bug with dynamic attributes becoming `undefined` while making use of the `ref` directive. 

```js
  render() {
    return html`<div
      ${ref(this._testRef)}
      class="some-div"
      id=${this._id}
      title=${this._title}
    >
      Hello World
    </div>`;
  }
```

For an complete example of our code, please check out our reproduction that gave the following HTML output:
<img width="918" alt="Screenshot 2024-01-05 at 13 00 45" src="https://github.com/lit/lit/assets/923084/4776a839-3c14-43c7-9261-341ca3e2a6ab">

### Reproduction

You can find a reproduction here: https://github.com/mwildeboer/lit-ssr-bug. We encountered the bug in both our development environment and the Playwright tests.

### Workaround

We fixed our bug by moving the ref declaration below the attributes like so:

```js
  render() {
    return html`<div
      class="some-div"
      id=${this._id}
      title=${this._title}
      ${ref(this._testRef)}
    >
      Hello World
    </div>`;
  }
```

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

2.7.2 -> 3.1.0

### Browser/OS/Node environment

Playwright 1.40.1 and latest Chrome
MacOS 14.2.1
Node version: v18.18.2

augustjk:
Was this actually working before? I can't imagine any change in Lit 3 causing this suddenly breaking. I would think if it's a regression, it was some change to the ssr code. In any case, I can indeed reproduce the issue separately as well and will look into a fix. There must be something going on with counting the binding when an element part is involved.

PR #4479, opened 2024-01-09
[labs/ssr] Fix attribute part bindings following element part binding

augustjk:
Fixes https://github.com/lit/lit/issues/4478

The attribute index for accessing the generated case sensitive name was erroneously being incremented on element part bindings, causing a shift of attribute names.


AndrewJakubowicz (approved):
Very excellent. Nice tests!

PR #4481, opened 2024-01-09
[labs/ssr-client] Provide better error message on hydration value mismatch

### LINKED ISSUE #4477: Error in createAttributeParts of hydrate-lit-html.js when attempting to server render a library component in Next.js

### Description

My team is creating a design system of Lit components, which we export both as web components and React-wrapped web components.  We'd like to enable full SSR capabilities for the React wrapped components of our library at the very least, though eventually it'd be great to do both.

I've gotten to the point where I can see the DSD and things are mostly working as intended, but I'm hitting an `Unhandled runtime error` with the `hydrate-lit-html.js` script from @lit-labs/ssr-client when I include our primary-navigation component in a Next.js app.
image

I'm having a hard time understanding what's causing this issue and I'd appreciate any insights you all might have.  I've included a reproduction to show what's going on.

Please let me know if you have any issues reproducing the issue or if I can help explain in more detail.

### Reproduction

You can find a stripped-down reproduction of our library here: https://github.com/keisha-rw/lit-components-lib
And a Next.js application that shows the issue can be found here: https://github.com/keisha-rw/next-learn
(Most of the time, our React consumers are using Next.js applications, so my repro is for Next.js.)

The README of the library explains how to view the issue (you should just need to pull down the repos and run the given commands). 

You can find the primary navigation component that hits this issue here: https://github.com/keisha-rw/lit-components-lib/blob/main/packages/web-components/src/lib/components/primary-navigation/primary-navigation.ts

### Workaround

### Is this a regression?

### Affected versions

1.1.5 (@lit-labs/ssr-client)

Browser: any
Node version: 18.17.0
Next.js version: 14.0.4 (though it happens in v13 as well)

augustjk:
Fixes #4477

### DIFF
--- a/.changeset/few-garlics-dress.md
+++ b/.changeset/few-garlics-dress.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr-client': patch
+---
+
+Provide better error message in `createAttributeParts()` when an error occurs due to hydration value mismatch.
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -437,7 +437,13 @@ const createAttributeParts = (
       state.templatePartIndex++;
     }
   } else {
-    throw new Error('internal error');
+    // TODO(augustjk): This message may need to be updated based on
+    // hydration strategy. See https://github.com/lit/lit/issues/1434
+    throw new Error(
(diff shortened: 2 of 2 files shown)


AndrewJakubowicz (approved):
I like it! Is there a test exercising this error path? (optional)

augustjk:
> Is there a test exercising this error path? (optional)
No tests currently exist for error cases or any hydration mismatch. I think it's fine as is until we actually start handling hydration mismatches differently.

Issue #4446, opened 2023-12-11
@lit/react package is not setting properties passed to the component during Vitest tests


tylerbmyers:
### Which package(s) are affected?

React (@lit/react)

### Description

When running tests in Vitest and using any version of @lit/react or @lit-labs/react versions 2 and up, any lit element wrapped with `createComponent` is not setting the properties of the web component that are passed as props.

This works fine in a real browser using Vite, and seems to be only an issue in the tests. The other thing that is interesting is that if I copy and paste the `create-component.ts` file into my project and import it directly, it works just fine, so my guess is that something with the minified file is not compatible with Vitest for some reason.

### Reproduction

I have created [this](https://github.com/tylerbmyers/lit-react-vitest-issue) repo with a simple Lit web component to reproduce the issue. In `App.test.tsx` you will see the component that is created using the `@lit/react` package is not setting the `name` property, but the component created using the direct import of `create-component.ts` is setting the `name` property.

### Workaround

My current workaround is to use `@lit-labs/react` version 1 as this works fine with Vitest.

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

Failing any version of @lit/react or @lit-labs/react versions 2 and up

### Browser/OS/Node environment

Node version: 20.10.0

augustjk:
From the fact that local copy of `create-component.ts` working, the likely culprit is that the Vitest build is using the "node" export condition when resolving the import while building the test file which removes the `useLayoutEffect()` that's responsible for updating the property here: https://github.com/lit/lit/blob/a365002dedb560cc0f1c68a74ea6817aebc4f971/packages/react/src/create-component.ts#L270-L297

There might be some configuration changes that can be done to make sure it uses the default browser build, but I haven't closely looked into how or if that plays nice with jsdom/happydom. It looks like Vitest also has an experimental mode to run tests in the browser with webdriverio and maybe that'll bring in browser builds too.

augustjk:
The whole point of using export condition is that it the end-user should not have to worry about where they're importing from when you're writing code that's meant to work in both SSR and in browsers.

This is unfortunately another downside/limitation of trying to test browser code in Node.

According to Vitest docs, having `environment: 'jsdom'` should automatically use the "web" transform mode. But it seems it's still using nodejs resolution for the initial entry into the external dependency which starts of picking the wrong entrypoint (the node export condition in our case).

Some configs I tried below:

```js
// https://vitejs.dev/config/
export default defineConfig({
[...]

PR #4485, opened 2024-01-10
[infra] Add browser export conditions to package.json

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4446

### Context

This is a compromise so that testing in Vitest works. As part of this work I explored simplifying the package.json exports using subpath patterns. Unfortunately the [ecosystem doesn't support those very well, so it's been punted](https://github.com/lit/lit/pull/4483).

- First commit contains the script.
- Second commit contains the result of running the script.

Marked this a patch change, as it's a minor "bug fix" workaround for testing in Jest and Vitest.

This script has two modes:

 - `test`: won't write any files, but instead checks that no files need to be fixed. Is run as part of our CI.
 - `fix`: writes to package.json files.

### Test plan

The prior browser/node entrypoints are tested via our CI.

The script itself is lightly tested via node assertions. The script won't run if an assertion fails.

### Risk

This changes the entrypoints in some of our package.json configurations, so it is a slightly risky change. We've manually verified that adding the browser export above node works correctly. Similarly if our test suite passes, then this should not be a regression as we use the exports in our tests.

### DIFF
--- a/packages/internal-scripts/package.json
+++ b/packages/internal-scripts/package.json
@@ -18,8 +18,7 @@
   },
   "scripts": {
     "build": "wireit",
-    "build:ts": "wireit",
-    "test": "wireit"
+    "build:ts": "wireit"
   },
   "wireit": {
     "build": {
@@ -36,12 +35,6 @@
       ],
       "output": [
         "lib"
-      ]
-    },
-    "test": {
-      "command": "(cd ../../ && node packages/internal-scripts/bin/add-browser-exports.js test)",
-      "dependencies": [
-        "build"
       ]
     }
   },
--- a/packages/internal-scripts/package.json
+++ b/packages/internal-scripts/package.json
@@ -12,6 +12,7 @@
   "license": "BSD-3-Clause",
   "type": "module",
   "bin": {
+    "add-browser-exports": "bin/add-browser-exports.js",
     "treemirror": "bin/treemirror.js",
     "release-image": "bin/release-image.js"
   },
--- a/packages/internal-scripts/package.json
+++ b/packages/internal-scripts/package.json
@@ -18,7 +18,8 @@
   },
   "scripts": {
     "build": "wireit",
-    "build:ts": "wireit"
+    "build:ts": "wireit",
+    "test": "wireit"
   },
   "wireit": {
     "build": {
@@ -35,6 +36,12 @@
       ],
       "output": [
         "lib"
+      ]
+    },
+    "test": {
+      "command": "(cd ../../ && node packages/internal-scripts/bin/add-browser-exports.js test)",
+      "dependencies": [
+        "build"
       ]
     }
   },
--- a/packages/internal-scripts/src/add-browser-exports.ts
+++ b/packages/internal-scripts/src/add-browser-exports.ts
@@ -16,7 +16,7 @@
 import assert from 'assert';
 import {execSync} from 'child_process';
 import {readFile, writeFile} from 'fs/promises';
-import {resolve} from 'path';
+import {resolve, relative} from 'path';
 
 testScriptIntegrity();
 
@@ -56,6 +56,21 @@ const absolutePackageJsonFilePaths = packageJsonFiles.map((relativePath) =>
 );
 
 export const run = async () => {
+  const [, , command] = process.argv;
+  if (!validateCommand(command)) {
+    console.error(
+      `
+USAGE
+    add-browser-exports <command>
+
+COMMANDS
+    test    Ensure this script doesn't need to fix any package.json files.
+    fix     Write export fixes to package.json files.
+`.trim()
+    );
+    process.exit(1);
+  }
+  let hadTestFailure = false;
   const work: Promise<void>[] = [];
   for (const packageJsonPath of absolutePackageJsonFilePaths) {
     work.push(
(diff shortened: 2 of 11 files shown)


packages/internal-scripts/src/add-browser-exports.ts
```diff
@@ -0,0 +1,213 @@
...
+        if (exports === undefined) {
+          return;
+        }
+        // `"exports": "./index.js"` can be skipped.
+        if (typeof exports === 'string') {
+          return;
+        }
+        const fixedExports = optionallyAddBrowserExportCondition(exports);
+        parsedPackageJson['exports'] = fixedExports;
+        await writeFile(
```
rictic, line 77: Consider doing nothing if the exports are the same before and after (maybe optionallyAddBrowserExportCondition could communnicate whether it made any changes as part of its return), and if something does need to be changed, writing it to the console like:

```
Updated packages/lit-html/package.json
```

packages/internal-scripts/src/add-browser-exports.ts
```diff
@@ -0,0 +1,213 @@
+/**
```
rictic, line 1: I'd like if this could be run as a test script as part of `npm run test`. The test wouldn't ever write anything to the filesystem, it'd just print something and exit with a nonzero exit code if this script would normally have written something (and mention that you can run the script to fix the failure).

It's a lightweight way to be sure we maintain this invariant in our package.json files

rictic (changes requested):

packages/internal-scripts/src/add-browser-exports.ts
AndrewJakubowicz: Excellent comment! I can definitely add that!

packages/internal-scripts/src/add-browser-exports.ts
AndrewJakubowicz: Love it! Can do!

AndrewJakubowicz: Done!

packages/internal-scripts/src/add-browser-exports.ts
AndrewJakubowicz: Done, I've set this up to run as part of the top level `test` script.

Verified by manually adding an error that it will signal a failure. Attached screenshot:

Screenshot 2024-01-10 at 2 31 06 PM

packages/internal-scripts/src/add-browser-exports.ts
```diff
@@ -0,0 +1,272 @@
...
+        dirty = true;
+        const browserExports: Partial<Exports['']['browser']> = {};
+        for (const nodeKey of Object.keys(exportValue.node) as Array<
+          keyof typeof exportValue.node
+        >) {
+          const exportPath = exportValue[nodeKey];
+          assert.notStrictEqual(
+            exportPath,
+            undefined,
+            `Any key/value exports defined in a node conditional export needs a browser export. Got undefind for "exports.node.${nodeKey}".`
```
rictic, line 163:
```suggestion
            `Any key/value exports defined in a node conditional export needs a browser export. Got undefined for "exports.node.${nodeKey}".`
```

rictic (approved):

rictic:
This is great work!

packages/internal-scripts/package.json
```diff
@@ -35,6 +37,12 @@
       "output": [
         "lib"
       ]
+    },
+    "test": {
+      "command": "(cd ../../ && node packages/internal-scripts/bin/add-browser-exports.js test)",
+      "dependencies": [
+        "build"
+      ]
```
augustjk, line 45: nit: seems a bit misleading for a script called "test" here to be running a particular script.

can this be moved to a script on the monorepo package.json maybe
```json
  "scripts": {
    "test:package-exports": "wireit",
  },
  "wireit": {
    "test:package-exports": {
      "command": "add-browser-exports test", // adding to .bin allows this command
      "dependencies": {
        "./packages/internal-scripts:build",
      },
    },
  },
```
and add this to the wireit dependencies of `"test:not-starter-kits"`?

packages/internal-scripts/src/add-browser-exports.ts
```diff
@@ -0,0 +1,272 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+// This script exists to perform a one time operation of adding the "browser"
+// exports field to relevant package.json files. Motivated by:
+// https://github.com/lit/lit/issues/4446
+//
+// The script is idempotent.
+//
+// # Instructions
+// After building the script run:
```
augustjk, line 13:
```suggestion
// After building the script run from monorepo root:
```

packages/internal-scripts/src/add-browser-exports.ts
```diff
@@ -0,0 +1,272 @@
...
+    node?: {
+      development?: string;
+      default?: string;
+    };
+    development?: string;
+    default?: string;
+  };
+}
+
+const result = execSync(`find . -name package.json`);
```
augustjk, line 43: would this maybe save at least grabbing `package.json` from root node_modules directory?
```suggestion
const result = execSync(`find ./packages -name package.json`);
```

packages/internal-scripts/package.json
AndrewJakubowicz: I like this!

packages/internal-scripts/src/add-browser-exports.ts
augustjk: although keeping it `.` means you can run this from any package sub directory if you care to just test that one package. 🤔 feel free to ignore this one.

AndrewJakubowicz: yes! Great suggestion!

augustjk (approved):

packages/internal-scripts/package.json
AndrewJakubowicz: Done! Thank you for helping me get it working (and in package-lock).

augustjk (approved):

Issue #4486, opened 2024-01-10
Preact-Signals SignalWatcher mixin does not support classes that inherit from abstract classes.


stevevarga:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

Package: @lit-labs/preact-signals

Error
```
Argument of type 'typeof SettingsPageElement' is not assignable to parameter of type 'ReactiveElementConstructor'.
  Cannot assign an abstract constructor type to a non-abstract constructor type.ts(2345)
```

The SignalWatcher mixin uses the ```new``` constructor, which means that all classes it extends need to have constructors.

If it extends a class that has an abstract class somewhere in it's inheritance, it will fail because that class does not have a constructor.

As of typescript 4.2 adding ```abstract``` in front of ```new``` fixes the issue. 

Please see https://github.com/microsoft/TypeScript/issues/29653#issuecomment-825595155 for context.

### Reproduction

https://lit.dev/playground/#sample=examples%2Fhello-world&gist=79b9cf55d033756dd7c607ca3e5e6d66

### Workaround

The fix is to use ```abstract new``` when writing mixins.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

@lit-labs/preact-signals 1.0.1

### Browser/OS/Node environment

Chrome OS X, Typescript 5.3.2

PR #4487, opened 2024-01-10
[lit-labs/preact-signals] add abstract class support to SignalWatcher

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4486

For context see the issue. It is now possible to pass an abstract class to the mixin and have it type check.

### Test plan

Added type and runtime unit tests to ensure that both concrete and abstract classes are watched correctly.

### DIFF
--- a/packages/labs/preact-signals/src/lib/signal-watcher.ts
+++ b/packages/labs/preact-signals/src/lib/signal-watcher.ts
@@ -7,8 +7,10 @@
 import type {ReactiveElement} from 'lit';
 import {effect} from '@preact/signals-core';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-type ReactiveElementConstructor = new (...args: any[]) => ReactiveElement;
+type ReactiveElementConstructor = abstract new (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ...args: any[]
+) => ReactiveElement;
 
 /**
  * Adds the ability for a LitElement or other ReactiveElement class to
@@ -18,7 +20,7 @@ type ReactiveElementConstructor = new (...args: any[]) => ReactiveElement;
 export function SignalWatcher<T extends ReactiveElementConstructor>(
   Base: T
 ): T {
-  return class SignalWatcher extends Base {
+  abstract class SignalWatcher extends Base {
     private __dispose?: () => void;
 
     override performUpdate() {
@@ -67,5 +69,6 @@ export function SignalWatcher<T extends ReactiveElementConstructor>(
       super.disconnectedCallback();
       this.__dispose?.();
     }
-  };
+  }
+  return SignalWatcher;
 }
--- a/packages/labs/preact-signals/src/test/signal-watcher_test.ts
+++ b/packages/labs/preact-signals/src/test/signal-watcher_test.ts
@@ -97,4 +97,72 @@ suite('SignalWatcher', () => {
     assert.equal(el.shadowRoot?.querySelector('p')?.textContent, 'count: 3');
     assert.equal(readCount, 2);
   });
+
+  test('type-only test where mixin on an abstract class preserves abstract type', () => {
+    if (true as boolean) {
+      // This is a type-only test. Do not run it.
+      return;
+    }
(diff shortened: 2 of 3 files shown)


packages/labs/preact-signals/src/lib/signal-watcher.ts
```diff
@@ -18,7 +20,7 @@ type ReactiveElementConstructor = new (...args: any[]) => ReactiveElement;
 export function SignalWatcher<T extends ReactiveElementConstructor>(
   Base: T
 ): T {
-  return class SignalWatcher extends Base {
+  abstract class SignalWatcher extends Base {
```
rictic, line 23: Hm, should SignalWatcher itself be abstract? I guess probably so.

However technically you could do:

```js
class MyElement extends LitElement {...}

customElements.define('my-element', SignalWatcher(MyElement));
```

rictic (approved):

packages/labs/preact-signals/src/lib/signal-watcher.ts
augustjk: The test below "class returned from signal-watcher should be directly instantiable if non-abstract" seems to suggest this is possible as is.

The tests show an interesting behavior. While I would've expected returning an abstract class here would always result in an abstract class that's non-instantiable, it seems it's still instantiable if the class provided to the mixin is non-abstract?

So then the reason for the addition of `abstract` in this line is to preserve the abstract behavior if the class provided to the mixin is also abstract. I don't know if that's something we strictly need to preserve, but seems unlikely to be a problem.

packages/labs/preact-signals/src/test/signal-watcher_test.ts
```diff
@@ -97,4 +97,72 @@ suite('SignalWatcher', () => {
...
+    const TestElFromAbstractSignalWatcher = SignalWatcher(BaseEl);
+    // @ts-expect-error cannot instantiate an abstract class.
+    new TestElFromAbstractSignalWatcher();
+
+    // This is fine, passed in class is not abstract.
+    const TestElFromConcreteClass = SignalWatcher(LitElement);
+    new TestElFromConcreteClass();
+  });
+
+  test('can mixin an abstract class and watch a signal', async () => {
```
augustjk, line 122: abstract classes compile to regular classes in JS so I think type checking is good enough if other tests already cover runtime behavior with classes.

packages/labs/preact-signals/src/lib/signal-watcher.ts
AndrewJakubowicz: `customElements.define('my-element', SignalWatcher(MyElement));` is possible and tested by this PR [here](https://github.com/lit/lit/pull/4487/files#diff-9e4d56fda3753440f204ca2f7f5fbaad99a1809237a396a74a28e7b80034a7a7R148-R167). 

> The tests show an interesting behavior. While I would've expected returning an abstract class here would always result in an abstract class that's non-instantiable, it seems it's still instantiable if the class provided to the mixin is non-abstract?

This is accurate. If the passed in class is non-abstract, then the resulting class from the mixin is non-abstract. This is due to the type signature of the mixin, which takes in `Base: T`, and returns `T`. So the return type matches the exact class passed in.

> So then the reason for the addition of abstract in this line is to preserve the abstract behavior if the class provided to the mixin is also abstract.

This is required in order to pass in an abstract class with abstract members. Otherwise, within the mixin we'd get a type error suggesting that abstract methods/fields need to be implemented. By marking the mixin class `abstract`, the abstract methods do not need to be implemented by the mixin.

augustjk: Thank you for the excellent explanation!

packages/labs/preact-signals/src/test/signal-watcher_test.ts
AndrewJakubowicz: Great point. I'll remove this test!

AndrewJakubowicz: Done

augustjk (approved):
🎉

Issue #4468, opened 2023-12-28
[labs/ssr-dom-shim] Implement toggleAttribute


Lattyware:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

SSR (@lit-labs/ssr), Other/unknown (please mention in description)

### Description

Currently the DOM shim doesn't implement [`toggleAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute), which specifically is breaking [Material Web components](https://github.com/material-components/material-web/tree/main) (specifically at least the checkbox and radio buttons) when trying to render them with SSR, as [they use it](https://github.com/material-components/material-web/blob/main/labs/behaviors/form-associated.ts#L241), getting `TypeError: this.toggleAttribute is not a function` and blowing up on the render.

This function seems to be a reasonable one to include in the shim given the other attribute methods currently implemented, and it can be easily implemented in terms of the others, so it feels like a small change (unless I'm missing something).

### Alternatives and Workarounds

My current work around for this is injecting a polyfill into the shim, i.e:

```ts
function toggleAttribute(
  this: Element,
  name: string,
  force?: boolean | undefined,
): boolean {
  switch (force) {
    case true: {
      this.setAttribute(name, "");
      return true;
    }
    case false: {
      this.removeAttribute(name);
      return false;
    }
    default:
      return toggleAttribute.bind(this)(name, !this.hasAttribute(name));
  }
}

class PatchedLitElementRenderer extends LitElementRenderer {
  constructor(tagName: string) {
    super(tagName);
    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
    if (this.element.toggleAttribute === undefined) {
      this.element.toggleAttribute = toggleAttribute;
    }
  }
}

// ...

render(template, {
    elementRenderers: [PatchedLitElementRenderer],
})
```

augustjk:
Thanks for the report! I agree it seems reasonable to include that for the `Element` shim.

PR #4493, opened 2024-01-12
[lit-labs/ssr-dom-shim] shim Element.prototype.toggleAttribute method

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4468

This adds [`Element.prototype.toggleAttribute(name, force?)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute) to the SSR ElementShim.

### Test plan

Unit tested in the node environment using `uvu`. Note, the package originally didn't have unit tests, so I've also set up a new `test` rule.

### Assumptions about casing

Documenting this as I was going to implement lower-casing of attribute names until I spoke with Augustine.

Ignoring casing is by design in the shim because these shimmed APIs can be used on SVG elements where casing is preserved. I believe the justification is we preserve casing during SSR, and then the browser will lowercase the attributes if in the HTML namespace. E.g. if we SSR `<div cLaSs="example">`, it'll be parsed into `<div class="example">`. But `<svg viewBox=...>` will be preserved as `<svg viewBox=...>`.

### DIFF
--- a/packages/labs/ssr-dom-shim/src/index.ts
+++ b/packages/labs/ssr-dom-shim/src/index.ts
@@ -61,6 +61,17 @@ const ElementShim = class Element {
   removeAttribute(name: string) {
     attributesForElement(this).delete(name);
   }
+  toggleAttribute(name: string, force?: boolean): boolean {
+    if (force === undefined) {
+      force = !this.hasAttribute(name);
+    }
+    if (force) {
+      this.setAttribute(name, '');
+    } else {
+      this.removeAttribute(name);
+    }
+    return force;
+  }
   hasAttribute(name: string) {
     return attributesForElement(this).has(name);
   }
--- a/.changeset/fresh-fishes-do.md
+++ b/.changeset/fresh-fishes-do.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr-dom-shim': minor
+---
+
+Add `toggleAttribute` to the Element shim.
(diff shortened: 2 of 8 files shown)


rictic (approved):

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -61,6 +61,17 @@ const ElementShim = class Element {
...
+  toggleAttribute(name: string, force?: boolean): boolean {
+    if (force === undefined) {
+      force = !this.hasAttribute(name);
+    }
+    if (force) {
+      this.setAttribute(name, '');
+    } else {
+      this.removeAttribute(name);
+    }
+    return force;
```
justinfagnani, line 73: This isn't quite correct, since if an attribute is already set and `force` is `true` the attribute should be left unmodified: https://dom.spec.whatwg.org/#dom-element-toggleattribute This code sets it to `''`.

I think this should work, but it needs a test for that case:

```suggestion
    if (this.hasAttribute(name) {
      // Step 5
      if (force === undefined || !force) {      
        this.removeAttribute(name);
        return false;
      }
    } else {
      // Step 4
      if (force === undefined || force) {
        // Step 4.1
        this.setAttribute(name, '');  
        return true;
      } else {
        // Step 4.2
        return false;
      }
    }
    // Step 6
    return true;
```

I _think_ this handles boolean conversion correctly. ie, `undefined` is the real missing value and `null` treated as falsey.

augustjk (approved):
Excellent!

justinfagnani (changes requested):

packages/labs/ssr-dom-shim/src/index.ts
AndrewJakubowicz: Exceptional catch! Thank you! Fixed in https://github.com/lit/lit/pull/4494

PR #4494, opened 2024-01-12
[lit-labs/ssr-dom-shim] fix toggleAttribute clearing prior values if force is true

AndrewJakubowicz:
This addresses the bug pointed out by @justinfagnani in https://github.com/lit/lit/pull/4493#discussion_r1450976231

A new unit test ensures the bug is covered.
Changeset is empty since this will be covered by the changeset adding `toggleAttribute`.


augustjk (approved):
Nice! Thanks @justinfagnani for catching this.

Issue #4490, opened 2024-01-12
[localize] `extract` command breaks translation targets with reordered expression placeholders


Benexx:
### Which package(s) are affected?

Localize (@lit/localize)

### Description

When using `@lit/localize` with for example ```msg(str`You have ${amount} licenses for ${product}.`)``` it will generate the following XLF.

``` XML
<trans-unit id="xxxxxxx">
  <source>You have <x id="0" equiv-text="${amount}"/> licenses for <x id="1" equiv-text="${product}"/>.</source>
</trans-unit>
```

Now we want for a translation a different order of the variables because it sounds better (example):
``` XML
<trans-unit id="xxxxxxx">
  <source>You have <x id="0" equiv-text="${amount}"/> licenses for <x id="1" equiv-text="${product}"/>.</source>
  <target>Of the selected product <x id="1" equiv-text="${product}"/> you now own <x id="0" equiv-text="${amount}"/> licenses.</target>
</trans-unit>
```

If you now use `lit-localize extract && lit-localize build` it will reset the variable `id` (not the `equiv-text`):
``` XML
<trans-unit id="xxxxxxx">
  <source>You have <x id="0" equiv-text="${amount}"/> licenses for <x id="1" equiv-text="${product}"/>.</source>
  <target>Of the selected product <x id="0" equiv-text="${product}"/> you now own <x id="1" equiv-text="${amount}"/> licenses.</target>
</trans-unit>
```

And using this translation with `amount=20` and `product=professional` will result in:
Source: `You have 20 licenses for professional.`
Target: `Of the selected product 20 you now own professional licenses.` *wrong order* 

### Reproduction

The playground does not support localize, create a string with 2 variables and reorder the variable in the translation will result in the issue

### Workaround

You could rewrite the translation to use the same variables order but this is not ideal in all scenarios and also annoying for translators.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

 @lit/localize@0.12.1

### Browser/OS/Node environment

OS: Latest windows
NPM: Latest
Node: Latest

augustjk:
Thanks for the report. I see the problem is that the swapping the location in `<target>` tag does work, but subsequently doing a `lit-localize extract` will break it by changing the position `id` in the `<target>`.

Here's a repro: https://stackblitz.com/edit/stackblitz-starters-fgmj2h?file=xliff%2Fen_GB.xlf
Everything works fine originally but running `lit-localize extract` will break it.

PR #4503, opened 2024-01-18
[localize-tools] Preserve existing placeholder id/index on extract

augustjk:
Fixes https://github.com/lit/lit/issues/4490

Instead of creating a new placeholder index on write which will always be in the order seen, we should use the existing index for untranslatables which could be out of order if they're read from an already translated target.


AndrewJakubowicz (approved):
Thank you!

PR #4504, opened 2024-01-18
More linter package setup

justinfagnani:
Installs typescript-eslint and adds a couple of fake rules to get testing setup for rules that do and don't require type information.


justinfagnani:
This doesn't yet pass the new tests because there's a project with the test fixture - it's complaining that the fixture's tsconfig doesn't contain the fixture's .ts file, even though it appears to me that it does. I haven't figured this out yet - any guesses are much appreciated!

.changeset/dirty-beans-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-lit': patch
+---
+
+More linter package setup
```
augustjk, line 5: We should do empty changesets until a package is actually released. There's no benefit in this line being adding to the changelog or bumping package version that changeset will do. We have git history for what was changed until release.
```suggestion
---
---
```

packages/labs/eslint-plugin/tsconfig.json
```diff
@@ -4,10 +4,10 @@
     "composite": true,
     "tsBuildInfoFile": "tsconfig.tsbuildinfo",
     "target": "es2020",
```
augustjk, line 6: Since this caught my eye, the rest of our packages all do es2021.
```suggestion
    "target": "es2021",
```

packages/labs/eslint-plugin/src/test/rules/rule-tester.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {RuleTester} from '@typescript-eslint/rule-tester';
+import {describe as suite, test} from 'node:test';
```
augustjk, line 8: If using the node test runner, we should bump the node engine in package.json to `16.17.0` which is when it was added.
https://github.com/lit/lit/blob/546d75d9e26f4aefa3bf91d114a70a3c3fb05da2/package.json#L6

augustjk (commented):
Commenting only on the package meta stuff.

packages/labs/eslint-plugin/src/test/rules/rule-tester.ts
justinfagnani: Noted, though there's a difference between the Node version required for development and version required to run the library code. I'm not sure what version of Node eslint, typescript-eslint, or our code here require actually.

.changeset/dirty-beans-serve.md
justinfagnani: good point :)

packages/labs/eslint-plugin/src/test/rules/rule-tester.ts
augustjk: Ah, excellent point about dev requirement being different. Dependency packages _should_ have their own engines specified, maybe? In any case, no action needed with regards to the original comment.

43081j (approved):

PR #4506, opened 2024-01-19
Initial labs/tsserver-plugin package setup

justinfagnani:
This sets up a tsserver plugin that proxies the prior language service, and is debuggable, but adds no features of its own.


43081j (approved):

packages/labs/tsserver-plugin/tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "compilerOptions": {
+    "incremental": true,
+    "composite": true,
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
+    "target": "es2022",
+    "lib": ["es2022"],
+    "module": "commonjs",
+    "rootDir": "./src",
+    "moduleResolution": "node",
```
43081j, line 10: can this and `module` be `nodenext` since the `package.json` doesn't have a `type: "module"`?

packages/labs/tsserver-plugin/example/tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compilerOptions": {
+    // This matches the package name given in this package.json
+    "plugins": [
+      {
+        "name": "@lit-labs/tsserver-plugin"
```
rictic, line 6: This works? Cool

packages/labs/tsserver-plugin/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    // TODO (justinfagnani): the property copying below is creaky and assumes
+    // everything is a method. Can we use prototypal inheritance or a real
+    // Proxy instead?
+    const proxy: ts.LanguageService = Object.create(null);
+
+    for (const k of Object.keys(info.languageService) as Array<
+      keyof ts.LanguageService
+    >) {
+      const x = info.languageService[k]!;
```
rictic, line 24: Do we want to fix this into a const? Could `info.languageService` be mutated by another system?

rictic (approved):

packages/labs/tsserver-plugin/example/tsconfig.json
justinfagnani: yeah - I got this from the tsserver-plugin example repo link'ed from the TS wiki. It's nice. The only problem is that the circular symlink required blows up eslint right now so you have to remove the example's node_modules/ folder to lint.

packages/labs/tsserver-plugin/src/index.ts
justinfagnani: I suppose it could? I probably shouldn't though.

I don't love this whole proxying setup - I just copied it from the starter kit. I want to have some test coverage before I muck with it though.

PR #4511, opened 2024-01-22
[internal-scripts] Add execute permission to add-browser-exports.js

augustjk:
Noticed a git diff after doing `npm ci` on main.


augustjk:
duplicate of https://github.com/lit/lit/pull/4510

PR #4515, opened 2024-01-23
[labs/ssr] fix patched directives memory leak

AndrewJakubowicz:
# Context

A memory leak was reported in Lit SSR.

Found by testing with https://github.com/lit/lit/pull/4514, and going test by test.
 - The `repeat` directive test contained this leak. After this PR the directive test no longer leaks.

So far no other test shows memory leak characteristics.

### Why?

Because we have a cache of `WeakMap<non patched directive, patched directive>`. This is an issue because we can pull a patched directive off the `value`, and then it won't be found in the WeakMap. Because it's in the value of the WeakMap and not the `key`. Thus the patched directive gets patched again and stored in the WeakMap. This repeats for the number of patch calls resulting in a nested doll of patched directive constructors.

### Fix

~~I implemented a super simple naive fix. Just flag the patched directive ctor with a sentinel property. Now when pulling the directive constructor off the `value`, we can immediately check if it's patched by looking at the sentinel.~~

Fix was updated per Justin's feedback. It now patches the `_$resolve` field. This simplifies a lot of the book keeping code. Which has a much nicer memory footprint.

However, this fix does need to be backwards compatible hence the new API, `patchDirectiveResolve`. This allows for the following:

- User has lit-html and ssr, and bumps lit-html only. The older ssr package can still use `overrideDirectiveResolve`.
- User has lit-html and ssr, and bumps ssr. This will need to depend on the version of lit-html that introduces `patchDirectiveResolve`, v3.1.2 (unreleased).

### Test plan

No memory test for this change because memory tests are expensive and slow. Instead this is covered by existing behavior tests. I can add tests if required.
The memory characteristics of this change were verified by pulling this change into https://github.com/lit/lit/pull/4514

Plan is to commit a memory test as a separate PR.

### DIFF
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -74,6 +74,10 @@ declare module 'parse5/dist/tree-adapters/default.js' {
   }
 }
 
+interface MaybePatchedDirective extends DirectiveClass {
+  patchedDirective: true | undefined;
+}
+
 const patchedDirectiveCache = new WeakMap<DirectiveClass, DirectiveClass>();
 
 /**
@@ -83,7 +87,10 @@ const patchedDirectiveCache = new WeakMap<DirectiveClass, DirectiveClass>();
 const patchIfDirective = (value: unknown) => {
   // This property needs to remain unminified.
   const directiveCtor = getDirectiveClass(value);
-  if (directiveCtor !== undefined) {
+  if (
+    directiveCtor !== undefined &&
+    (directiveCtor as MaybePatchedDirective).patchedDirective === undefined
+  ) {
     let patchedCtor = patchedDirectiveCache.get(directiveCtor);
     if (patchedCtor === undefined) {
       patchedCtor = overrideDirectiveResolve(
@@ -94,6 +101,7 @@ const patchIfDirective = (value: unknown) => {
           return patchIfDirective(directive.render(...values));
         }
       );
+      (patchedCtor as MaybePatchedDirective).patchedDirective = true;
       patchedDirectiveCache.set(directiveCtor, patchedCtor);
     }
     // This property needs to remain unminified.
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -7,11 +7,7 @@
  */
 
 import type {TemplateResult, ChildPart, CompiledTemplateResult} from 'lit';
-import type {
-  Directive,
-  DirectiveClass,
-  DirectiveResult,
-} from 'lit/directive.js';
+import type {Directive} from 'lit/directive.js';
 
 import {nothing, noChange} from 'lit';
 import {PartType} from 'lit/directive.js';
@@ -29,8 +25,7 @@ const {
   marker,
   markerMatch,
   boundAttributeSuffix,
-  overrideDirectiveResolve,
-  setDirectiveClass,
+  patchDirectiveResolve,
   getAttributePartCommittedValue,
   resolveDirective,
   AttributePart,
@@ -67,6 +62,7 @@ import {reflectedAttributeName} from './reflected-attributes.js';
 
 import type {RenderResult} from './render-result.js';
 import {isHydratable} from './server-template.js';
+import type {Part} from 'lit-html';
 
 declare module 'parse5/dist/tree-adapters/default.js' {
   interface Element {
@@ -74,38 +70,20 @@ declare module 'parse5/dist/tree-adapters/default.js' {
   }
 }
 
-interface MaybePatchedDirective extends DirectiveClass {
-  patchedDirective: true | undefined;
+function ssrResolve(this: Directive, _part: Part, values: unknown[]) {
+  // Since the return value may also be a directive result in the case of nested
+  // directives, we may need to patch that as well.
+  return patchIfDirective(this.render(...values));
 }
 
-const patchedDirectiveCache = new WeakMap<DirectiveClass, DirectiveClass>();
-
 /**
- * Looks for values of type `DirectiveResult` and replaces its Directive class
- * with a subclass that calls `render` rather than `update`
+ * Looks for values of type `DirectiveResult` and patches its Directive class
+ * such that it calls `render` rather than `update`.
  */
 const patchIfDirective = (value: unknown) => {
-  // This property needs to remain unminified.
   const directiveCtor = getDirectiveClass(value);
-  if (
-    directiveCtor !== undefined &&
-    (directiveCtor as MaybePatchedDirective).patchedDirective === undefined
-  ) {
-    let patchedCtor = patchedDirectiveCache.get(directiveCtor);
-    if (patchedCtor === undefined) {
-      patchedCtor = overrideDirectiveResolve(
-        directiveCtor,
-        (directive: Directive, values: unknown[]) => {
-          // Since the return value may also be a directive result in the case of
-          // nested directives, we may need to patch that as well
-          return patchIfDirective(directive.render(...values));
-        }
-      );
-      (patchedCtor as MaybePatchedDirective).patchedDirective = true;
-      patchedDirectiveCache.set(directiveCtor, patchedCtor);
-    }
-    // This property needs to remain unminified.
-    setDirectiveClass(value as DirectiveResult, patchedCtor);
+  if (directiveCtor !== undefined) {
+    patchDirectiveResolve(directiveCtor, ssrResolve);
   }
   return value;
 };
--- a/packages/lit-html/src/private-ssr-support.ts
+++ b/packages/lit-html/src/private-ssr-support.ts
@@ -59,7 +59,11 @@ export const _$LH = {
     },
   patchDirectiveResolve: (
     directiveClass: new (part: PartInfo) => Directive,
-    resolveOverrideFn: Directive['_$resolve']
+    resolveOverrideFn: (
+      this: Directive,
+      _part: Part,
+      values: unknown[]
+    ) => unknown
   ) => {
     if (directiveClass.prototype._$resolve !== resolveOverrideFn) {
       directiveClass.prototype._$resolve = resolveOverrideFn;
--- a/packages/lit-html/src/private-ssr-support.ts
+++ b/packages/lit-html/src/private-ssr-support.ts
@@ -27,6 +27,9 @@ import type {
   TemplateInstance,
 } from './lit-html.js';
 
+// Contains either the minified or unminified `_$resolve` Directive method name.
+let resolveName = null;
+
 /**
  * END USERS SHOULD NOT RELY ON THIS OBJECT.
  *
@@ -65,12 +68,28 @@ export const _$LH = {
       values: unknown[]
     ) => unknown
   ) => {
-    // TODO: This could be optimized such that `_$resolve` is only patched on
-    // the Directive base class once. This should reduce class shape changes and
-    // be slightly more optimal. The hazard is that we need to handle both the
-    // unminified and minified property keys.
     if (directiveClass.prototype._$resolve !== resolveOverrideFn) {
-      directiveClass.prototype._$resolve = resolveOverrideFn;
+      resolveName ??= directiveClass.prototype._$resolve.name;
+      let proto = directiveClass.prototype;
+      for (; proto !== Object.prototype; proto = Object.getPrototypeOf(proto)) {
+        if (proto.hasOwnProperty(resolveName)) {
+          (proto as unknown as {[key: string]: typeof resolveOverrideFn})[
+            resolveName
+          ] = resolveOverrideFn;
+          return;
+        }
+      }
+      // Nothing was patched which indicates an error. The most likely error is
+      // that somehow both minified and unminified lit code passed through this
+      // codepath. This is possible as lit-labs/ssr contains its own lit-html
+      // module as a dependency for server rendering client Lit code. If a
+      // client contains multiple duplicate Lit modules with minified and
+      // unminified exports, we currently cannot handle both.
+      throw new Error(
+        `Internal error: It is possible  that both dev mode and production mode` +
+          ` Lit was mixed together during SSR. Please file a bug at: ` +
+          `https://github.com/lit/lit/issues/new/choose`
+      );
     }
   },
   setDirectiveClass(value: DirectiveResult, directiveClass: DirectiveClass) {
--- a/packages/lit-html/src/private-ssr-support.ts
+++ b/packages/lit-html/src/private-ssr-support.ts
@@ -88,9 +88,9 @@ export const _$LH = {
       // client contains multiple duplicate Lit modules with minified and
       // unminified exports, we currently cannot handle both.
       throw new Error(
-        `Internal error: It is possible  that both dev mode and production mode` +
-          ` Lit was mixed together during SSR. Please file a bug at: ` +
-          `https://github.com/lit/lit/issues/new/choose`
+        `Internal error: It is possible that both dev mode and production mode` +
+          ` Lit was mixed together during SSR. Please comment on the issue: ` +
+          `https://github.com/lit/lit/issues/4527`
       );
     }
   },
--- a/packages/lit-html/src/private-ssr-support.ts
+++ b/packages/lit-html/src/private-ssr-support.ts
@@ -59,11 +59,7 @@ export const _$LH = {
     },
   patchDirectiveResolve: (
     directiveClass: new (part: PartInfo) => Directive,
-    resolveOverrideFn: (
-      this: Directive,
-      _part: Part,
-      values: unknown[]
-    ) => unknown
+    resolveOverrideFn: Directive['_$resolve']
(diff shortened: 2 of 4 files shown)


rictic (approved):

augustjk (approved):
Good find and fix!

I looked into whether it might be better to add the sentinel property in `overrideDirectiveResolve` but that lives in `lit-html` exported via private-ssr-support so it seems better to keep everything in this ssr package.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -83,7 +87,10 @@ const patchedDirectiveCache = new WeakMap<DirectiveClass, DirectiveClass>();
 const patchIfDirective = (value: unknown) => {
   // This property needs to remain unminified.
   const directiveCtor = getDirectiveClass(value);
-  if (directiveCtor !== undefined) {
+  if (
+    directiveCtor !== undefined &&
+    (directiveCtor as MaybePatchedDirective).patchedDirective === undefined
```
justinfagnani, line 92: Why can't this just check the `patchedDirectiveCache`?

```suggestion
!patchedDirectiveCache.has(directiveCtor)
```

Otherwise consider a WeakSet?

justinfagnani (changes requested):

packages/labs/ssr/src/lib/render-value.ts
augustjk: It's the first check that happens within this if block, but the problem was that `directiveCtor` could be an already patched one which isn't added as the cache map key.

after patching the directive we do
```ts
patchedDirectiveCache.set(directiveCtor, patchedCtor);
```
so only the unpatched `directiveCtor` serves as the key.

I entertained the idea of also adding
```ts
patchedDirectiveCache.set(patchedCtor, patchedCtor);
```
which would make lookup from the cache catch already patched ones, but wasn't sure if that's really better than a sentinel property on the patched ctor.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -83,7 +87,10 @@ const patchedDirectiveCache = new WeakMap<DirectiveClass, DirectiveClass>();
 const patchIfDirective = (value: unknown) => {
   // This property needs to remain unminified.
```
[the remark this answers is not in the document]
augustjk: there's another one down in (new) line 107. it looks like old residue before the helper `getDirectiveClass()` and `setDirectiveClass()` were introduced so they could be cleaned up!

packages/labs/ssr/src/lib/render-value.ts
justinfagnani: hmm... I think we need some more comments here and make a name change? I would have thought that patching a directive would have actually patched the directive ctor we received, so that there isn't a _new_ class. We're instead _replacing_ the directive.

I am curious how a replaced directive re-enters here. When is a DirectiveResult passed to multiple renders?

Regardless, I get the problem now, and I think there may be a much simpler solution to this whole thing, which is to monkey-patch `_$resolve` on `Directive`. That method is used by all directives. We just have to check that we patched `Directive` in case we get multiple copies of lit-html in play.

AndrewJakubowicz: Great discussion. I'll explore a simpler fix.

AndrewJakubowicz: Previously this was implemented as a patch to the Directive prototype: https://github.com/lit/lit/pull/1599/commits/73930f40ac1d1033eed205d723498f6792a4ddba#diff-b8cfd0fdb8410f04f051c052e6ce548e5403f407d6954d48df02b7c5bdb0e506R75-R85

And `_$resolve` was patched directly. See change here: https://github.com/lit/lit/pull/1599/commits/96dfb42234a96782ceab5b3cfa091310020b4927#diff-f57f5f57247e8497a15da483cc93126465912bec0dd886454289befb86a72cbdL36-L41

History of this change is captured in the PR: https://github.com/lit/lit/pull/1599

augustjk: > I am curious how a replaced directive re-enters here. When is a DirectiveResult passed to multiple renders?

Hmm... the DirectiveResult would be in the values array of a template, and we mutate that value, so I think if that same template is getting SSRed over multiple requests, this could happen?

> monkey-patch `_$resolve` on `Directive`

Catching all copies might be hard.. we can only dig into the prototype of the directive class we got passed in. and we need this baked into `lit-html/private-ssr-support.js` due to the stable property mangling.

justinfagnani: I think there's another way to work around the issue of multiple lit-html libraries:

```ts
function ssrResolve(part: Part, props: Array<unknown>): unknown {
  return this.render(props);
}

const patchDirectiveResolve = (value: unknown) => {
  const directiveCtor = getDirectiveClass(value);
  if (directiveCtor._$resolve !== ssrResolve) {
    // even better, walk up the proto chain to Directive
    directiveCtor._$resolve = ssrResolve;
  }
};
```

This will work against any compatible version of the lit-html library.

AndrewJakubowicz:
Great feedback which has been incorporated. Description has also been updated. Thank you!

packages/labs/ssr/src/lib/render-value.ts
AndrewJakubowicz: Done!

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,14 @@ export const _$LH = {
         return resolveOverrideFn(this, values);
       }
     },
+  patchDirectiveResolve: (
+    directiveClass: new (part: PartInfo) => Directive,
+    resolveOverrideFn: Directive['_$resolve']
+  ) => {
+    if (directiveClass.prototype._$resolve !== resolveOverrideFn) {
+      directiveClass.prototype._$resolve = resolveOverrideFn;
```
augustjk, line 65: Maybe needed?
```suggestion
    // This property needs to remain unminified.
    if (directiveClass.prototype['_$resolve'] !== resolveOverrideFn) {
      directiveClass.prototype['_$resolve'] = resolveOverrideFn;
```

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,14 @@ export const _$LH = {
         return resolveOverrideFn(this, values);
       }
     },
+  patchDirectiveResolve: (
```
augustjk, line 60: Unsure if it's worth going a bit more verbose to communicate the condition
```suggestion
  patchDirectiveResolveIfNeeded: (
```

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,14 @@ export const _$LH = {
...
+    directiveClass: new (part: PartInfo) => Directive,
+    resolveOverrideFn: Directive['_$resolve']
```
augustjk, line 62: `_$resolve` is marked `/** @internal */` so won't be in the declaration, making it break when used on exported things. Unfortunately I think we need to revert the type simplification back to the explicit one.

AndrewJakubowicz: Good catch.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -67,37 +62,28 @@ import {reflectedAttributeName} from './reflected-attributes.js';
...
-          // Since the return value may also be a directive result in the case of
-          // nested directives, we may need to patch that as well
-          return patchIfDirective(directive.render(...values));
-        }
-      );
-      patchedDirectiveCache.set(directiveCtor, patchedCtor);
-    }
-    // This property needs to remain unminified.
-    setDirectiveClass(value as DirectiveResult, patchedCtor);
+    patchDirectiveResolve(directiveCtor, ssrResolve);
```
justinfagnani, line 86: Let's try to get @kevinpschaaf's eyes on this in case he remembers why we didn't go this way before.

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,18 @@ export const _$LH = {
         return resolveOverrideFn(this, values);
       }
     },
+  patchDirectiveResolve: (
+    directiveClass: new (part: PartInfo) => Directive,
```
justinfagnani, line 61: I think you can use `typeof Directive` here, which might be more descriptive

packages/labs/ssr/src/lib/render-value.ts
justinfagnani, line 86: walking up the proto chain until you get to the proto with `_$resolve` on it (ie, `Directive`) would mean we only path once per copy of lit-html, which will be nicer on the VM due to fewer class shape changes.

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: I think _$resolve does need to be minified. It should be minified to `S`.

packages/labs/ssr/src/lib/render-value.ts
AndrewJakubowicz: That's fair. Is it worth moving forward with the first solution that adds a simple brand just to fix the memory leak. It's not the most elegant, but it's a very self contained fix.

Then we can explore a larger scale fix separately? This can unblock users impacted by the current memory leak.

packages/lit-html/src/private-ssr-support.ts
augustjk: oops you're totally right!

packages/labs/ssr/src/lib/render-value.ts
AndrewJakubowicz: Gave this an attempt and it didn't go super well.
Instead documented this as a future enhancement.

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,22 @@ export const _$LH = {
...
+    resolveOverrideFn: (
+      this: Directive,
+      _part: Part,
+      values: unknown[]
+    ) => unknown
+  ) => {
+    // TODO: This could be optimized such that `_$resolve` is only patched on
+    // the Directive base class once. This should reduce class shape changes and
+    // be slightly more optimal. The hazard is that we need to handle both the
+    // unminified and minified property keys.
```
justinfagnani, line 71: The minification comment seems either to also apply to this approach, and `overrideDirectiveResolve`.  If this `patchDirectiveResolve` function is called against a minified lit-html Directive class, it won't work, right? We should try to trigger that in a test.

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -57,6 +57,22 @@ export const _$LH = {
         return resolveOverrideFn(this, values);
       }
     },
+  patchDirectiveResolve: (
```
justinfagnani, line 60: We want to remove `overrideDirectiveResolve` at some point... but older versions of ssr may depend on a newer version of lit-html... when can we remove it?

packages/lit-html/src/private-ssr-support.ts
augustjk: Andrew and I were discussing this in person yesterday. I believe the `overrideDirectiveResolve` works fine with minification specifically because it's the override method name is minified by the same process. Climbing the prototype and looking for that method could be difficult because `hasOwnProperty()` or `Object.hasOwn()` takes a string.

packages/lit-html/src/private-ssr-support.ts
augustjk: I think we get to remove it when we want to consider the current version of `@lit-labs/ssr` unsupported. Perhaps we can add a deprecation warning that logs when `overrideDirectiveResolve` is called that encourages users to update `@lit-labs/ssr`? They also should do that since there's a memory leak.

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: The minification comment issue that I was expecting is specifically for string literal handling, e.g.: `hasOwnProperty("_$resolve")`. Where I'm not sure `_$resolve` will become `S`. But should check.

justinfagnani: There's a separate minification issue where the version of lit-html that ssr imports could be minified and the version that contains the directive not minified, or vice-versa. This could happen with say, older versions of Shoelace that bundle the production build, or any situation where the client-side app is built separately from the server app.

AndrewJakubowicz: Ah yeah. I think I have actually encountered that issue and it currently exists.

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: Great comment! The positive of adding API here is that this doesn't increase `lit-html`'s size as it's only pulled from in SSR. The negative is that there is a larger API surface here.

We would want to be very slow at removing `overrideDirectiveResolve `, and potentially leave it in until most people are using Lit 3. If we want to continue down this approach I could add a comment to document this.

Alternatively, we could revert back to the very first implementation where we patch fix the memory leak from only the ssr side, without any modifications to `private-ssr-support`.

justinfagnani: I think one takeaway is that overrideDirectiveResolve isn't low-level enough, and we should check that patchDirectiveResolve doesn't have the same problem. Is there's a lower-level primitive we could add and move the patching itself into ssr?

AndrewJakubowicz: I'm quite motivated to fix the underlying issue since the leak is pretty serious. Wdyt about this very simple fix that I've split out: https://github.com/lit/lit/pull/4528

Then we can work on a much more robust fix that also handles minification correctly.

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -65,12 +68,28 @@ export const _$LH = {
       values: unknown[]
     ) => unknown
   ) => {
-    // TODO: This could be optimized such that `_$resolve` is only patched on
-    // the Directive base class once. This should reduce class shape changes and
-    // be slightly more optimal. The hazard is that we need to handle both the
-    // unminified and minified property keys.
     if (directiveClass.prototype._$resolve !== resolveOverrideFn) {
-      directiveClass.prototype._$resolve = resolveOverrideFn;
+      resolveName ??= directiveClass.prototype._$resolve.name;
```
justinfagnani, line 72: A longer variable name might be more clear:
```suggestion
      resolveMethodName ??= directiveClass.prototype._$resolve.name;
```

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -65,12 +68,28 @@ export const _$LH = {
...
+      let proto = directiveClass.prototype;
+      for (; proto !== Object.prototype; proto = Object.getPrototypeOf(proto)) {
```
justinfagnani, line 74: I personally prefer while loops over for loops w/o an initializer.

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -65,12 +68,28 @@ export const _$LH = {
...
+        if (proto.hasOwnProperty(resolveName)) {
+          (proto as unknown as {[key: string]: typeof resolveOverrideFn})[
```
justinfagnani, line 76: Can you eliminate the case by typing `resolveName` as `keyof Directive`?

packages/lit-html/src/private-ssr-support.ts
```diff
@@ -65,12 +68,28 @@ export const _$LH = {
...
+      }
+      // Nothing was patched which indicates an error. The most likely error is
+      // that somehow both minified and unminified lit code passed through this
+      // codepath. This is possible as lit-labs/ssr contains its own lit-html
+      // module as a dependency for server rendering client Lit code. If a
+      // client contains multiple duplicate Lit modules with minified and
+      // unminified exports, we currently cannot handle both.
+      throw new Error(
+        `Internal error: It is possible  that both dev mode and production mode` +
+          ` Lit was mixed together during SSR. Please file a bug at: ` +
```
justinfagnani, line 90: I would file the issue now and link to it so that we don't get many people filing the same issue.

justinfagnani (approved):
nice work!

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: Whoops, fixed with an initializer.

AndrewJakubowicz: Done!

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: Done!

packages/lit-html/src/private-ssr-support.ts
AndrewJakubowicz: See changed wording. Thank you!

PR #4516, opened 2024-01-24
[eslint-plugin] Add attribute names rule

justinfagnani:
This PR should be easier to review commit-by-commit. The last commit adds the actual attribute-names rule, the rest of them set things up. I bumped typescript and a few testing dependencies due to package duplication that broke the build while installing some packages.

### Type-awareness

This implementation of attribute-names requires a type-checker. This is to detect Lit subclasses and decorators. I think this is a good tradeoff for now, especially as we can support subclasses, ReactiveElement, re-exports, etc (work on that TBD). We almost certainly do want to be able to use different heuristics to detect these things in places where they don't want to run the type-checker, so we can make some detection logic pluggable. 

typescript-eslint recommends an option or two different rules in this case:

> We recommend against changing rule logic based solely on whether `services.program` exists. In our experience, users are generally surprised when rules behave differently with or without type information. Additionally, if they misconfigure their ESLint config, they may not realize why the rule started behaving differently. Consider either gating type checking behind an explicit option for the rule or creating two versions of the rule instead.

https://typescript-eslint.io/developers/custom-rules/#typed-rules

We'll have to figure that out.

### Copyright assignment

@43081j I originally copied the rule from https://github.com/43081j/eslint-plugin-lit/blob/master/src/rules/attribute-names.ts but ended up rewriting most of it to work against the Lit analyzer. I did copy the tests, then added the correct Lit imports to satisfy the analyzer's detection of LitElement and decorators.

The commits to this repo need to be copyright Google (unless they're a compatible OSS license, credited, and linked). I think the rule implementation qualifies. The test is written by you and you've signed the CLA, so I believe you can just reassign copyright to Google at will with your approval here. Or I can write new tests if necessary.


packages/labs/eslint-plugin/src/lib/util.ts
```diff
@@ -0,0 +1,65 @@
...
+  ESLintUtils,
+  ParserServicesWithTypeInformation,
+} from '@typescript-eslint/utils';
+// TODO (justinfagnani): get from ESLintUtils if possible?
+// See q: https://discord.com/channels/1026804805894672454/1084238921677946992/1199174067459198986
+import ts from 'typescript';
+import * as path from 'path';
+
+export const createRule = ESLintUtils.RuleCreator(
+  (name: string) => `https://lit.dev/eslint-plugin/${name}`
```
augustjk, line 18: is this worth a TODO or an issue for actually adding docs to lit.dev ?

packages/labs/eslint-plugin/src/lib/util.ts
```diff
@@ -0,0 +1,65 @@
...
+export const getAnalyzer = (services: ParserServicesWithTypeInformation) => {
+  const program = services.program;
+  let analyzer = analyzerByProgram.get(program);
+  if (analyzer === undefined) {
+    analyzer = new Analyzer({
+      typescript: ts,
+      getProgram: () => program,
+      fs: ts.sys,
+      path,
+      // basePath,
```
augustjk, line 32: should this be removed?

packages/labs/eslint-plugin/src/lib/util.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+  return undefined;
+};
+
+export const isTrueLiteral = (e: ts.Expression): e is ts.TrueLiteral =>
+  e.kind === ts.SyntaxKind.TrueKeyword;
+
+export const isFalseLiteral = (e: ts.Expression): e is ts.FalseLiteral =>
+  e.kind === ts.SyntaxKind.TrueKeyword;
```
augustjk, line 59:
```suggestion
  e.kind === ts.SyntaxKind.FalseKeyword;
```

packages/labs/eslint-plugin/src/rules/attribute-names.ts
```diff
@@ -0,0 +1,112 @@
...
+        const tsNode = services.esTreeNodeToTSNodeMap.get(node);
+        const declaration = getDeclarationForNode(analyzer, filename, tsNode);
+        if (
+          declaration === undefined ||
+          !declaration.isLitElementDeclaration()
+        ) {
+          return;
+        }
+        const properties = declaration.reactiveProperties;
+        // const propertyMap = getPropertyMap(node);
```
augustjk, line 58: should this be removed?

packages/labs/testing/package.json
```diff
@@ -108,7 +108,7 @@
   "dependencies": {
     "@lit-labs/ssr": "^3.1.8",
     "@lit-labs/ssr-client": "^1.1.4",
-    "@web/test-runner-commands": "^0.6.1",
+    "@web/test-runner-commands": "^0.9.0",
```
augustjk, line 111: nothing that this change warrants a bump to the labs/testing package.

packages/labs/eslint-plugin/package.json
```diff
@@ -33,7 +36,7 @@
       "clean": "if-file-deleted"
     },
     "test": {
-      "command": "node --enable-source-maps --test-reporter=dot --test test/**/*_test.js",
+      "command": "node --enable-source-maps --test-reporter=spec --test test/**/attribute-names_test.js",
```
augustjk, line 39: is narrowing the --test glob intentional?

augustjk (commented):
Commenting minor things that stood out to me. I'll defer to other reviewers for the actual rule implementation.
Additionally noting that packages like labs/cli and labs/compiler that have direct dependency on typescript will also need version bumps with this.

augustjk:
It is curious that this has size regression for core packages. Could changing the TS version change the emitted JS?

packages/labs/eslint-plugin/src/lib/util.ts
justinfagnani: ah, yes

justinfagnani:
> It is curious that this has size regression for core packages. Could changing the TS version change the emitted JS?

@augustjk I was wondering the same thing 🤷‍♂️

packages/labs/eslint-plugin/package.json
justinfagnani: nope!

43081j:
> typescript-eslint recommends an option or two different rules in this case

i would have two rules i think, a typed and untyped one (which share some logic). then just publish a typed/untyped recommended config

> The test is written by you and you've signed the CLA, so I believe you can just reassign copyright to Google at will with your approval here

sounds good to me.

i'm not too concerned with the copyright, copying things over here, etc. as long as im still able to exist as a maintainer alongside you all (which it seems i am)

package-lock.json
```diff
@@ -21614,9 +23667,10 @@
       }
     },
     "node_modules/rollup-plugin-terser/node_modules/terser": {
-      "version": "5.19.2",
+      "version": "5.27.0",
```
augustjk, line 23670: Looks like this terser update might be responsible for the bundle size change. I tried updating typescript alone and it didn't change the output but this branch does have a diff. Diff looks innocuous. Somehow previous version lucked out and didn't require an alias for `H`, but after update it does.
image

43081j (approved):
looks good to me 🥳

packages/labs/eslint-plugin/src/test/rules/attribute-names_test.ts
```diff
@@ -0,0 +1,104 @@
...
+          camelCase = 'foo';
+        }`,
+      errors: [
+        {
+          line: 4,
+          column: 11,
+          messageId: 'casedPropertyWithoutAttribute',
+        },
+      ],
+    },
```
rictic, line 102: Huh, I hadn't seen this rule before. Do we like this rule? Do we want to recommend it?

The casedAttribute message seems good, but the casedPropertyWithoutAttribute doesn't seem great to me.

packages/tests/src/web-test-runner.config.ts
```diff
@@ -138,7 +138,17 @@ See https://wiki.saucelabs.com/display/DOCS/Platform+Configurator for all option
...
+            args: [
+              '--js-flags=--expose-gc',
+              '--enable-precise-memory-info',
+              '--disable-background-timer-throttling',
+              '--disable-background-media-suspend',
+              '--disable-backgrounding-occluded-windows',
+              '--disable-input-event-activation-protection',
+              '--disable-renderer-backgrounding',
+              '--enable-automation',
+              '--disable-ipc-flooding-protection',
```
rictic, line 150: where did these new flags come from?

packages/labs/eslint-plugin/src/index.ts
```diff
@@ -4,19 +4,14 @@
...
  */
 
-import type {Rule, ESLint} from 'eslint';
-import {noKevin} from './rules/no-kevin.js';
+import type {ESLint} from 'eslint';
 
-export const rules = {
-  'no-kevin': noKevin as unknown as Rule.RuleModule,
-};
+export const rules = {};
```
rictic, line 9: Should this be empty?

justinfagnani: nope!

packages/tests/src/web-test-runner.config.ts
justinfagnani: I found these in a few places that were documenting trying to work around Chrome's background tab throttling. I'm trying to serperate and/or revert the changes causing the problems to unblock this PR.

packages/labs/eslint-plugin/src/test/rules/attribute-names_test.ts
justinfagnani: https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/attribute-names.md

I'm planning on porting all the existing rules to ease migration to the new version. I'm sure there is some overlap with Rune's rules, and some we might want to deprecate going forward. I agree that a lot of perfectly fine code may have a camelCased property name and be making the choice to use the lower-case / case insensitive attribute name and shouldn't be flagged. In those cases we may want additional configuration for this rule, and we may want a migration-oriented configuration and not include this in the recommended set.

I should open an RFC on how the plan for all of this.

PR #4518, opened 2024-01-24
chore/document minimum node version requirement for contributing guidelines

thescientist13:
### Overview
Was noticing that when running test cases I was getting this error
```sh
✗ npm run test

> test
> npm run test:not-starter-kits && npm run test:lit-starter-js && npm run test:lit-starter-ts


> test:not-starter-kits
> wireit


❌ [packages/labs/eslint-plugin:test] exited with exit code 9. Output:

node: bad option: --test-reporter=dot
❌ 1 script failed.
```

Per discussions on the Discord, it was determined that a minimum Node version was required as the project uses Node's built-in test runner, which only [came out in Node `20.x`](https://nodejs.org/api/test.html)

### Changes
1. Added a callout to the _CONTRIBUTING.md_ for expected minimum NodeJS version (`18.15.0`)
1. Update the `engines` field in _package.json_ to enforce this

----

~~Let me know your thoughts on the _.nvmrc_ file and / or version, happy to set that to whatever you prefer (or I can remove it)~~

Also, I don't think this change requires a changeset but can certainly generate one of those if desired.


augustjk (commented):
Thank you for the PR! Apologies for having to go through that error. Usage of Node test runner was a recent addition and we should've included that info for contributors.

The problematic command line argument `--test-reporter` was introduced in Node ~~v18.5.0~~ v18.15.0 (https://nodejs.org/api/test.html#test-reporters) so that should be our minimum Node version. I confirmed the project can be built with that.

I am hesitant about adding an `.nvmrc` though since that forces a very specific version rather than a semver match: https://github.com/nvm-sh/nvm/issues/2562

thescientist13:
> The problematic command line argument --test-reporter was introduced in Node v18.5.0 (https://nodejs.org/api/test.html#test-reporters) so that should be our minimum Node version. I confirmed the project can be built with that.

Interesting, I (thought) I was originally testing on `18.15.0` when I ran into that issue but tested again on that version and seems fine now. 😅 

> I am hesitant about adding an .nvmrc though since that forces a very specific version rather than a semver match: https://github.com/nvm-sh/nvm/issues/2562

Fair enough.

Maybe instead we bump the `engine` in the root [_package.json_](https://github.com/lit/lit/blob/main/package.json#L6) to `18.5.0` instead?

augustjk:
> Maybe instead we bump the engines in the root [package.json](https://github.com/lit/lit/blob/main/package.json#L6) to 18.5.0 instead?

I did have that thought (although with a wrong version) https://github.com/lit/lit/pull/4504#discussion_r1456739756 but found it reasonable that the version of Node to use the code is different than the version of Node to develop it.

I suppose it does make a difference if we're putting the engine on the monorepo package.json rather than the individual packages.

augustjk (changes requested):
@thescientist13 We discussed this in today's eng meeting and agree that setting the node version in engines in the monorepo root package.json is good.

If you could change the min node version there to >=18.15.0 instead of the .nvmrc, that would be great!

thescientist13:
All set!

augustjk (approved):
Awesome, thank you!!

CONTRIBUTING.md
```diff
@@ -102,7 +102,14 @@ If you've already signed a CLA but are still getting bothered by the awfully ins
 
 ## Set up
 
+You'll need at least [**Node** >= `18.15.0`](https://nodejs.org/en) installed on your machine.
+
+Once the minimum Node version is installed, you can continue with the rest of the repo setup.
+
 ```bash
+node -v
+v18.15.0
```
AndrewJakubowicz, line 111: In context with the rest of this script, this line could be interpreted as a command to type in.
Maybe make it a shell comment? e.g.:  `# v18.15.0`.

AndrewJakubowicz (approved):
Awesome thank you!

CONTRIBUTING.md
augustjk: Considering the context, I think it's fine. :)

Issue #4513, opened 2024-01-23
[labs/ssr] 3.2.0 to 3.2.1 breaks `<td colspan=${'foo')>` in server html template


JulianCataldo:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

Updating from 3.2.0 to 3.2.1 breaks `td` `colspan` binding.

### Reproduction

Repro:

```ts
import { html, render } from '@lit-labs/ssr';
import { collectResultSync } from '@lit-labs/ssr/lib/render-result.js';

const test = html` <td colspan=${2}>test</td> `;

console.log(collectResultSync(render(test)));
```

Error:

```
error: 
    Unexpected final partIndex: 0 !== 1 while processing the following template:

     <td colspan=${...}>test</td> 
```

### Workaround

none

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

3.2.0 to 3.2.1 

### Browser/OS/Node environment

node 1.0.25
bun v20.11.0

augustjk:
Thank you for reporting. I can confirm this is indeed a regression introduced by https://github.com/lit/lit/pull/4421

It seems `parse` from `parse5` swallows bare `<td>` tags so we lose all attribute bindings.
https://astexplorer.net/#/gist/161a7023b229715f0ba3f23f09a14fe4/fc8a336ba1088b61ae4b427027b7cf72de29f426

We tried to test for these https://github.com/lit/lit/blob/273ad4e23b8ec97f1a5015dbf398104f535f9c34/packages/labs/ssr/src/test/test-files/render-test-module.ts#L338 but failed to do so with attribute bindings.

I think we need to reconsider using `parse` over `parseFragment`.

thescientist13:
I wonder if the issue, at least from my understanding of the parse docs, is that if your string contains:
- an `<html>` tag (a "complete") document, you would need to use [`parse`](https://parse5.js.org/functions/parse5.parse.html) 
- for anything else, you would want to use [`parseFragment`](https://parse5.js.org/functions/parse5.parseFragment.html)

So perhaps [the detection](https://github.com/lit/lit/pull/4421/files#diff-993b92b15e276bc22253e71a6283fda6fe62ccf8a3d3846276e73808765b2eb1R315) needs to be something like this instead?
```js
const ast = (html.indexOf('html') >= 0 ? parse : parseFragment)(String(html), {
  sourceCodeLocationInfo: true,
});
```

[...]

PR #4519, opened 2024-01-25
[labs/ssr] fix server templates dropping top level `td` elements

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4513

### Context

Using `parse` from parse5 everywhere for server templates is not ideal, and unfortunately has consequences for some cases such as the issue template containing a top level `td` element. Parsing this as a page level HTML causes the element to be removed (which matches browser behavior). However, we instead want to treat this case as a fragment.

### Fix

This is a possible fix. I'm not super stoked about it but it does improve the situation.

Basically, there are 3 tags that I could find that `parseFragment` will not handle as a top-level element: `html`, `head`, and `body`.
Thus, detect those cases and only use `parse` if one of these top-level tags is detected. Otherwise use `parseFragment`.

### Test plan

Added a regression test for the template flagged in the issue.

### Risk

There's technically some other edge cases that can fall out of this. For example, the following template will not work correctly due to the HTML comment throwing off the regex:

```js
html`<!-- My nifty comment --><head>...</head>`
```

This might be acceptable as the situation has been improved by this PR.

### DIFF
--- a/.changeset/tough-crews-cover.md
+++ b/.changeset/tough-crews-cover.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix a bug where a top level 'td' tag would be removed from a server template.
--- a/packages/labs/ssr/src/lib/render-value.ts
+++ b/packages/labs/ssr/src/lib/render-value.ts
@@ -224,6 +224,14 @@ type Op =
   | PossibleNodeMarkerOp;
 
 /**
+ * Any of these top level tags will be removed by parse5's `parseFragment` and
+ * will cause part errors if there are any part bindings. For only these tags,
+ * we use a page-level template `parse`.
+ */
+const REGEXP_TEMPLATE_STARTS_WITH_PAGE_TAG =
+  /^\s*(<(!doctype|html|head|body))/i;
+
+/**
  * For a given TemplateResult, generates and/or returns a cached list of opcodes
  * for the associated Template.  Opcodes are designed to allow emitting
  * contiguous static text from the template as much as possible, with specific
@@ -285,11 +293,9 @@ type Op =
  * - `custom-element-close`
  *   - Pop the CE `instance`+`renderer` off the `customElementInstanceStack`
  */
-const getTemplateOpcodes = (
-  result: TemplateResult,
-  isServerTemplate = false
-) => {
+const getTemplateOpcodes = (result: TemplateResult) => {
   const template = templateCache.get(result.strings);
+  console.log('got undefined template', result.strings);
   if (template !== undefined) {
     return template;
   }
@@ -302,6 +308,9 @@ const getTemplateOpcodes = (
   );
 
   const hydratable = isHydratable(result);
+  const htmlString = String(html);
+  const isPageLevelTemplate =
+    REGEXP_TEMPLATE_STARTS_WITH_PAGE_TAG.test(htmlString);
 
   /**
    * The html string is parsed into a parse5 AST with source code information
@@ -311,7 +320,7 @@ const getTemplateOpcodes = (
    * Server Templates need to use `parse` as they may contain document tags such
    * as `<html>`.
    */
-  const ast = (isServerTemplate ? parse : parseFragment)(String(html), {
+  const ast = (isPageLevelTemplate ? parse : parseFragment)(htmlString, {
     sourceCodeLocationInfo: true,
   });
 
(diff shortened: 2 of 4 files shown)


justinfagnani:
I suspect there's a bit more we can do here with more advanced parts of the parse5 API. Looking at the implementation of `parseFragment()` it seems like we can get at the root node and see attributes on it:

```ts
import {Parser} from 'parse5';

const parser = Parser.getFragmentParser();
parser.tokenizer.write('<html foo><div></div></html>', true);

// This has the `foo` attribute
console.log(parser.treeAdapter.getFirstChild(parser.document));

// This only has the <div>
console.log(parser.getFragment());
```

I'm not sure yet how to tie this into a fix, but it seems like more of the information is there internally.

AndrewJakubowicz:
I agree Justin! I'll take a look a little more deeply into what parse5 gives us tomorrow.

AndrewJakubowicz:
@justinfagnani I think an inexpensive regex approach is probably going to be the best to capture the couple page level elements. Using `parseFragment` specifically on templates containing a top level `<html>` doesn't work well (even if manually rigged up) because parse5 removes the `sourceCodeLocation`. I haven't found a way to populate the source code location, which is important because we use it to `flushTo` and `skipTo` around Lit markers.

augustjk (approved):
Unless we unship the `parse()` change to regress on the page level element support, I like this is a strict improvement of what we currently have.

Trying to not use `parse()` all together could be explored on top of this change too.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -223,6 +223,14 @@ type Op =
   | CustomElementClosedOp
   | PossibleNodeMarkerOp;
 
+/**
+ * Any of these top level tags will be removed by parse5's `parseFragment` and
+ * will cause part errors if there are any part bindings. For only these tags,
+ * we use a page-level template `parse`.
+ */
+const REGEXP_TEMPLATE_STARTS_WITH_PAGE_TAG =
+  /^\s*(<(!doctype|html|head|body))/i;
```
justinfagnani, line 232: I think this will handle leading comments too:
```suggestion
  /^(\s*|<!--.*-->)*(<(!doctype|html|head|body))/is;
```

justinfagnani:
@AndrewJakubowicz thanks for looking into it!

I think the limitations of using non-fragment should be documented along with the tag. I modified the regex to handle comments (I think) so maybe there wouldn't be any to mention.

packages/labs/ssr/src/lib/render-value.ts
AndrewJakubowicz: Great suggestion. I've updated regex and added a bunch of tests. Things look pretty great with the comment regex addition.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -302,16 +307,22 @@ const getTemplateOpcodes = (
   );
 
   const hydratable = isHydratable(result);
+  const isServerTemplate = !hydratable;
```
augustjk, line 310: Nit: the rest of the code uses `hydratable` directly as basically meaning server template. e.g. https://github.com/lit/lit/blob/9c02b3876dc927c6a82b4420411256ecbb47c08c/packages/labs/ssr/src/lib/render-value.ts#L460-L470

I don't know why the naming was chosen as such then, but I don't see a need to introduce another boolean, even though I much prefer the name `isServerTemplate`.

augustjk (approved):
Thank you!

.changeset/tough-crews-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/ssr': patch
+---
+
+Fix a bug where a top level 'td' tag would be removed from a server template.
```
augustjk, line 5:
```suggestion
Fix a bug where server templates with attribute bindings on certain element tags like `<td>` would throw an error during server render.
```

packages/labs/ssr/src/lib/render-value.ts
AndrewJakubowicz: Done!

.changeset/tough-crews-cover.md
AndrewJakubowicz: Should the fact that it's a top-level tag be mentioned?

```suggestion
Fix a bug where server templates with attribute bindings on certain top-level element tags like `<td>` would throw an error during server render.
```

augustjk: Oh yes! I somehow dropped the top-level part as i was rewording. The placement makes it a bit ambiguous though. How do you like
```suggestion
Fix a bug where server templates with attribute bindings on certain element tags like `<td>` used top-level would throw an error during server render.
```

AndrewJakubowicz: Love it!

justinfagnani (approved):

Issue #4194, opened 2023-09-13
[lit-html] Detect duplicate attributes in DEV_MODE


filimon-danopoulos-stratsys:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

When a property is accidentally duplicated it is perceived that Lit silently "fails". It would be nice to get a warning about this in DEV_MODE as it can lead to obscure bugs that are hard to track down otherwise. 

See the full discussion from Discord here: https://discord.com/channels/1012791295170859069/1149635522646515774

Here is a small case where I would like to get a warning: https://lit.dev/playground/#gist=932be48698b63a7b9d38a78b75a8a3ef

When the template looks like this: 

```html
<textarea 
      ?disabled=${this.disabled} 
      ?disabled=${this.disabled} 
      placeholder=${this.placeholder}
></textarea>
```

Not sure if an explanation is necessary but the `placeholder` will not render properly as the duplicate properties are removed by the `<template>` resulting in a discrepancy between parts and values. Thomas Boyer explained it well in the attached Discord thread. 

If the template is large, with event handlers, `classMap` and `styleMap`, a single duplicate line can be really hard to track down and is unintuitive as the cause of the resulting bug. A message like `> Possible duplicate attribute detected.` would be very helpful. Ideally with some way to identify where said duplication might exist. 

### Alternatives and Workarounds

This could also be solved by expanded editor support where heavier analysis can be done during development.

PR #4523, opened 2024-01-26
[lit-html] Add DEV_MODE error if duplicate attribute bindings are encountered

AndrewJakubowicz:
Fixes: https://github.com/lit/lit/issues/4194

### Context

A lot of the context is in the issue, but in short, if an attribute binding is duplicated, such as in the following example:

```html
<textarea 
      ?disabled=${this.disabled} 
      ?disabled=${this.disabled} 
      placeholder=${this.placeholder}
></textarea>
```

The result is that Lit parts get out of sync with the values (because attributes are de-duped by the browser).

### Fix

I've implemented this as a DEV_MODE error, but I'm open to making it a DEV_MODE warning.

The following template: `<input ?disabled=${true} ?disabled=${false}>`, will throw:

```
Detected duplicate attribute bindings of attribute: '?disabled', in the template: `<input ?disabled=${...} ?disabled=${...}>`
```

### DIFF
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1041,7 +1041,9 @@ class Template {
       // If there was a duplicate attribute on a tag, then when the tag is
       // parsed into an element the attribute gets de-duplicated. We can detect
       // this mismatch if we haven't precisely consumed every attribute name
-      // when preparing the template. This works because `attrNames` is built from the template string and `attrNameIndex` comes from processing the resulting DOM.
+      // when preparing the template. This works because `attrNames` is built
+      // from the template string and `attrNameIndex` comes from processing the
+      // resulting DOM.
       if (attrNames.length !== attrNameIndex) {
         throw new Error(
           `Detected duplicate attribute bindings. This occurs if your template ` +
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -1045,7 +1045,7 @@ class Template {
       if (attrNames.length !== attrNameIndex) {
         throw new Error(
           `Detected duplicate attribute bindings. This occurs if your template ` +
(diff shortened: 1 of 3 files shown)


justinfagnani (commented):
lgtm!!

justinfagnani:
I think an error is correct here. The template will throw at rendering time anyway.

AndrewJakubowicz:
Note: Converted to draft because this collapses duplicate attributes across individual tags which is invalid behavior.

Added a test case: `<input ?disabled=${...}><input ?disabled=${...}>`. This should not error (and it does with the current naive implementation).

AndrewJakubowicz:
I fixed the implementation to be correct. Throw if there are attribute names that were not considered for template preparation (detects attribute mismatch).

Adding the attribute that caused the attribute mismatch is going to be tricky as duplicate attribute names can exist across multiple tags. I think this implementation is strictly an improvement, and we provide the template in the error that caused the issue. This will reduce debugging time for user's who encounter this error.

augustjk (approved):
Ooohhh very clever to check that we've consumed everything from `attrNames`!

packages/lit-html/src/lit-html.ts
```diff
@@ -1036,6 +1036,26 @@ class Template {
       }
       nodeIndex++;
     }
+
+    if (DEV_MODE) {
+      // If there was a duplicate attribute on a tag, then when the tag is
+      // parsed into an element the attribute gets de-duplicated. We can detect
+      // this mismatch if we haven't precisely consumed every attribute name
+      // when preparing the template.
```
sorvell, line 1044:
```suggestion
      // when preparing the template. This works because `attrNames` is built from the template string and `attrNameIndex` comes from processing the resulting DOM.
```

sorvell (approved):

justinfagnani (approved):

Issue #4502, opened 2024-01-18
[localize-tools] Validation of existing translation should not fail on changed expression


augustjk:
### Which package(s) are affected?

Localize (@lit/localize)

### Description

Reported by @kensternberg-authentik in https://github.com/lit/lit/discussions/4489

https://lit.dev/docs/localization/overview/#message-ids states the rules for message id generation and how they're used to deduplicate translation targets. This can be interpreted that changing the code inside expressions in your source code should not fail on future `extract` and `build` calls but there seems to exist some validation that compares the value of `equiv-text` attribute.

### Reproduction

See linked discussion above for sample and error message.

### Workaround

Manually update the xliff to match the `equiv-text`

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

@lit/localize-tools@0.7.1

### Browser/OS/Node environment

n/a

augustjk:
According to the comment here https://github.com/lit/lit/blob/546d75d9e26f4aefa3bf91d114a70a3c3fb05da2/packages/localize-tools/src/messages.ts#L108-L119 placeholders don't always contain just the expression code and can contain html too so we ignoring it might not be the best idea.

But perhaps this assumption doesn't hold when we consider multiple source `msg()` calls can map to a single trans-unit.

Need to investigate and consider this more carefully.

augustjk:
Got around to trying out some things and (re)discovering how lit-localize actually works. Thought dump here so I don't forget it all later.

When running `extract`, it'll always overwrite `<source>` of the translation unit with analysis from source code.

The validation and potential error happens when `build` is run, as it validates the placeholder in the translated `<target>` against the source code.

For runtime mode, it looks like localize currently uses the placeholder text from the `<target>` tag as it generates code. This is problematic if `<target>` tags have been modified and not verified against source.

For transform mode, localize only uses the index of the placeholder from `<target>` for positioning, but the actual placeholder content comes from source code.

Potential forward path here, we can make runtime mode behave like transform in that it'll always source placeholder text from source code. This removes the concern from the original comment above of arbitrary code injection from the translation targets. The only useful validation we'd really need then is just the number of placeholders so we know where to place them.

But without content validation, we won't detect any drift from source code to translation target for cases were the id was manually made. Would that actually matter? (This kind of drift wouldn't exactly happen with auto generated ID as it would be treated as a new translation unit then)

[...]

PR #4530, opened 2024-01-30
[localize] Make placeholder validation pass with different expressions

augustjk:
Fixes https://github.com/lit/lit/issues/4502

## How it used to be before #2405
In the past, the expressions in the translated message placeholder were directly used in generating the output code in transform mode.

##### foo.ts (source)
```js
msg(str`Hello ${user}!`);
```

##### es-419.xlf (translation)
```xml
<trans-unit id="s00ad08ebae1e0f74">
  <source>Hello <x id="0" equiv-text="${user}"/>!</source>
  <target>Hola <x id="0" equiv-text="${user}"/>!</target>
</trans-unit>
```
used the actual `equiv-text="${user}"` value from `<target>` to generate

##### foo.js (output)
```js
`Hello ${user}!`;
```

This meant it was possible for faulty or malicious JS expression to be injected to the output if translation placeholders were modified.

##### es-419.xlf (translation)
```xml
<trans-unit id="s00ad08ebae1e0f74">
  <source>Hello <x id="0" equiv-text="${user}"/>!</source>
  <target>Hola <x id="0" equiv-text="${alert('evil') || user}"/>!</target>
</trans-unit>
```
could produce

##### foo.js (output)
```js
`Hola ${alert('evil') || user}!`;
```

Runtime mode never had this issue since expressions are replaced by numbers in the output anyway and swap them with values given during runtime.

##### es-419.ts (runtime mode output)

```js
export const templates = {
  s00ad08ebae1e0f74: str`Hola ${0}!`,
}
```

## After #2405 

https://github.com/lit/lit/pull/2405 made it so that multiple `msg()` calls with different expressions can map to the same translation unit.

##### foo.ts (source)
```js
msg(str`Hello ${user}!`);
msg(str`Hello ${username}!`);
```
resulted in a single translation unit.
##### es-419.xlf (translation)
```xml
<trans-unit id="s00ad08ebae1e0f74">
  <source>Hello <x id="0" equiv-text="${user}"/>!</source>
  <target>Hola <x id="0" equiv-text="${user}"/>!</target>
</trans-unit>
```
The value of `equiv-text` is based on which expression is encountered first.

As such, it also made it so that we never use the expression from translation directly instead always grab the expression from the source code so that we'll produce the appropriate output (both `${user}` and `${username}` in above example). **This means strict validation of expressions in placeholders are now unnecessary.**

However, there remained some validation logic that would check that expressions in placeholders exactly as they are in the source code. If the source code were modified or the order of the template encountered changes while placeholders in the translations are not updated, errors like the follow would show up during build, even though the translations are still perfectly valid to use.

```
Placeholder error in es-419 localization of s00ad08ebae1e0f74: unexpected "${username}"
Placeholder error in es-419 localization of s00ad08ebae1e0f74: missing "${user}"
```

## What this PR does

The validation logic is now updated to replace all expressions to `"expr"` before comparison so a different expression wouldn't cause it to fail. We still want to make sure the rest of the string matches since placeholders can also include HTML that won't be translated, and placeholders without expressions are still used in output generation for runtime mode.

### DIFF
--- a/packages/localize-tools/src/messages.ts
+++ b/packages/localize-tools/src/messages.ts
@@ -139,26 +139,34 @@ export function validateLocalizedPlaceholders(
       // Note that two identical placeholders could appear in the same template,
       // and it matters how many of them there are, hence we use an array, not a
       // set (might be good to implement some form of multiset here).
-      const remainingProgramPlaceholders = [];
+      const remainingProgramPlaceholders: Array<{
+        raw: string;
+        normalized: string;
+      }> = [];
       for (const content of programMsg.contents) {
         if (typeof content !== 'string') {
-          remainingProgramPlaceholders.push(
-            replaceExpressionInTemplateString(content.untranslatable)
-          );
+          remainingProgramPlaceholders.push({
+            raw: content.untranslatable,
+            normalized: normalizeExpressionInTemplateString(
+              content.untranslatable
+            ),
+          });
         }
       }
 
       for (const content of localizedMsg.contents) {
         if (typeof content !== 'string') {
-          const placeholder = replaceExpressionInTemplateString(
+          const normalizedPlaceholder = normalizeExpressionInTemplateString(
             content.untranslatable
           );
-          const index = remainingProgramPlaceholders.indexOf(placeholder);
+          const index = remainingProgramPlaceholders.findIndex(
+            ({normalized}) => normalized === normalizedPlaceholder
+          );
           if (index === -1) {
             errors.push(
               `Placeholder error in ${locale} ` +
                 `localization of ${localizedMsg.name}: ` +
-                `unexpected "${placeholder}"`
+                `unexpected "${content.untranslatable}"`
             );
           } else {
             remainingProgramPlaceholders.splice(index, 1);
@@ -170,7 +178,7 @@ export function validateLocalizedPlaceholders(
         errors.push(
           `Placeholder error in ${locale} ` +
             `localization of ${localizedMsg.name}: ` +
-            `missing "${placeholder}"`
+            `missing "${placeholder.raw}"`
         );
       }
     }
@@ -179,24 +187,19 @@ export function validateLocalizedPlaceholders(
 }
 
 /**
- * Given a template string, replace all expression with a provided string (or
- * "expr" if none provided).
+ * Given a template string, replace all expression with "expr". Used to compare
+ * static parts of the template string.
  *
  * e.g. `hello ${foo} world ${bar}` -> `hello ${expr} world ${expr}`
  */
-function replaceExpressionInTemplateString(
-  templateString: string,
-  expression = 'expr'
-): string {
+function normalizeExpressionInTemplateString(templateString: string): string {
   const template = parseStringAsTemplateLiteral(templateString);
   if (ts.isNoSubstitutionTemplateLiteral(template)) {
     return template.text;
   }
-  const fragments: string[] = [];
-  fragments.push(template.head.text);
-  for (let i = 0; i < template.templateSpans.length; i++) {
-    fragments.push('${' + expression + '}');
-    fragments.push(template.templateSpans[i].literal.text);
+  let normalizedString = template.head.text;
+  for (const span of template.templateSpans) {
+    normalizedString += '${expr}' + span.literal.text;
   }
-  return fragments.join('');
+  return normalizedString;
 }
--- a/packages/localize-tools/src/messages.ts
+++ b/packages/localize-tools/src/messages.ts
@@ -4,8 +4,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import * as ts from 'typescript';
+import ts from 'typescript';
 import type {Locale} from './types/locale.js';
+import {parseStringAsTemplateLiteral} from './typescript.js';
 
 /**
  * A message for translation.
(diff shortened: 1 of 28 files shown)


packages/localize-tools/src/messages.ts
```diff
@@ -169,3 +177,26 @@ export function validateLocalizedPlaceholders(
...
+
+/**
+ * Given a template string, replace all expression with a provided string (or
+ * "expr" if none provided).
+ *
+ * e.g. `hello ${foo} world ${bar}` -> `hello ${expr} world ${expr}`
+ */
+function replaceExpressionInTemplateString(
+  templateString: string,
+  expression = 'expr'
```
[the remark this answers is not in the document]
augustjk: Yes, good point. It s unnecessary at the moment. Also renamed to `normalizeExpressionInTemplateString` as well as using string concat rather than array.join().

packages/localize-tools/testdata/build-transform-xliff/input/xliff/es-419.xlf
```diff
@@ -46,6 +46,18 @@
   <source>&lt;Hello<ph id="0">&lt;b></ph>&lt;World &amp; Friends><ph id="1">&lt;/b></ph>!></source>
   <target>&lt;Hola<ph id="0">&lt;b></ph>&lt;Mundo &amp; Amigos><ph id="1">&lt;/b></ph>!></target>
 </trans-unit>
+<trans-unit id="s63f0bfacf2c00f6b">
+  <source>Hello</source>
+  <target>Hola</target>
+</trans-unit>
+<trans-unit id="s372f95c2b25986c8">
+  <source>Different <ph id="0">${user}</ph></source>
```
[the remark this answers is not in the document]
augustjk: `<source>` expression is generated by the `extract` command but never read again by lit-localize so it does not need to really match anything. It will be overwritten if `extract` command is run again and finds a new value from the source code.

packages/localize-tools/src/tests/e2e/placeholder-errors.test.ts
```diff
@@ -12,10 +12,8 @@ e2eGoldensTest(
   1,
   `One or more localized templates contain a set of placeholders (HTML or template literal expressions) that do not exactly match the source code, aborting. Details:
 
-Placeholder error in es-419 localization of extra-expression: unexpected "\${alert("evil")}"
-Placeholder error in es-419 localization of missing-expression: missing "\${name}"
-Placeholder error in es-419 localization of changed-expression: unexpected "\${alert("evil") || name}"
-Placeholder error in es-419 localization of changed-expression: missing "\${name}"
+Placeholder error in es-419 localization of extra-expression: unexpected "\${expr}"
+Placeholder error in es-419 localization of missing-expression: missing "\${expr}"
```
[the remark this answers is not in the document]
augustjk: Yes good call. I was able to find a way to keep the raw expression around to preserve the same error message behavior from before.

packages/localize-tools/src/tests/e2e/placeholder-errors.test.ts
```diff
@@ -14,8 +14,6 @@ e2eGoldensTest(
 
 Placeholder error in es-419 localization of extra-expression: unexpected "\${alert("evil")}"
 Placeholder error in es-419 localization of missing-expression: missing "\${name}"
-Placeholder error in es-419 localization of changed-expression: unexpected "\${alert("evil") || name}"
-Placeholder error in es-419 localization of changed-expression: missing "\${name}"
```
augustjk, line 18: changed expressions are no longer considered placeholder errors.

packages/localize-tools/testdata/build-transform-xliff/goldens/tsout/es-419/foo.js
```diff
@@ -57,6 +57,9 @@ html`&lt;Hola<b>&lt;Mundo &amp; Amigos&gt;</b>!&gt;`;
 // Expressions as attribute values should stay as expressions
 html`Hello <b foo=${'World'}>World</b>`;
 html`Hello <b foo=${`Mundo`}>World</b>`;
-html`<b foo=${'Hello'}>Hello</b><b bar=${`Mundo`}>World</b>`;
+html`<b foo=${`Hola`}>Hello</b><b bar=${`Mundo`}>World</b>`;
```
augustjk, line 60: This diff is technically unrelated to the changes in the PR but just due to a missing translation target that I noticed from the test file.

packages/localize-tools/testdata/build-transform-xliff/input/xliff/es-419.xlf
```diff
@@ -46,6 +46,18 @@
   <source>&lt;Hello<ph id="0">&lt;b></ph>&lt;World &amp; Friends><ph id="1">&lt;/b></ph>!></source>
   <target>&lt;Hola<ph id="0">&lt;b></ph>&lt;Mundo &amp; Amigos><ph id="1">&lt;/b></ph>!></target>
 </trans-unit>
+<trans-unit id="s63f0bfacf2c00f6b">
+  <source>Hello</source>
+  <target>Hola</target>
+</trans-unit>
```
augustjk, line 52: This is the missing translation target I found, mentioned above.

packages/localize-tools/src/messages.ts
```diff
@@ -135,22 +139,34 @@ export function validateLocalizedPlaceholders(
...
       for (const content of localizedMsg.contents) {
         if (typeof content !== 'string') {
-          const placeholder = content.untranslatable;
-          const index = remainingProgramPlaceholders.indexOf(placeholder);
+          const normalizedPlaceholder = normalizeExpressionInTemplateString(
+            content.untranslatable
+          );
+          const index = remainingProgramPlaceholders.findIndex(
+            ({normalized}) => normalized === normalizedPlaceholder
+          );
```
AndrewJakubowicz, line 164: nice!

AndrewJakubowicz (approved):
👏 👏 👏

PR #4520, opened 2024-01-25
Bump typescript

justinfagnani:
Breaking out part of #4516 to isolate test flakiness

PR #4531, opened 2024-01-31
[chore] Update typescript to 5.3.3

augustjk:
Seeing if we can update typescript with a smaller blast radius than https://github.com/lit/lit/pull/4520


augustjk:
Closing in favor of https://github.com/lit/lit/pull/4520

PR #4533, opened 2024-01-31
[infra/benchmarks] Detect chromedriver version

augustjk:
Use environment variable to detect which version of chromedriver to install during npm install
https://github.com/giggio/node-chromedriver?tab=readme-ov-file#detect-chromedriver-version

I'm hoping the npm cache behavior won't mess with the chromedriver version. It's not something that should change too often anyway.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -25,6 +25,7 @@ jobs:
         env:
           # Force installed chromedriver version to match the available chrome
           # version regardless of package-lock
+          # https://github.com/giggio/node-chromedriver?tab=readme-ov-file#detect-chromedriver-version
           DETECT_CHROMEDRIVER_VERSION: true
         run: npm ci
 


.github/workflows/benchmarks.yml
```diff
@@ -22,6 +22,10 @@ jobs:
       - uses: google/wireit@setup-github-actions-caching/v1
 
       - name: NPM install
+        env:
+          # Force installed chromedriver version to match the available chrome
+          # version regardless of package-lock
```
AndrewJakubowicz, line 27: Nit: link to https://github.com/giggio/node-chromedriver?tab=readme-ov-file#detect-chromedriver-version

This rocks! Awesome!

AndrewJakubowicz (approved):
🚀

.github/workflows/benchmarks.yml
augustjk: good suggestion, done!

Issue #4497, opened 2024-01-16
[react] reusal of react component instances across render cycles by react leads to properties not being updated correctly


mikonse:
### Which package(s) are affected?

React (@lit/react)

### Description

Let's assume a simple lit component with a disabled flag and a label `<my-input label="label" disabled></my-input>` and a minimal react wrapper created using `@lit/react` called `MyInput`.

Consider the following react code snippet from a simple react component:

```tsx
if (somecondition) {
  return <MyInput label="foo" disabled={true} />
}

return <MyInput label="bar" />
```

If our the component containing the above code is rendered across two render cycles, `somecondition` being true in the first cycle, false in the second, something interesting happens.
After the second render cycle the DOM will contain 

```html
<my-input label="bar" disabled=""></my-input>
```

This is obviously unexpected as from looking at the react code we'd expect the input to not be disabled at all.

The reason for this weird behavior lies (to some extent) in the way how react reuses component instances.
In the second render cycle react determines that it already rendered an instance of `MyInput` as the first element to be returned from our component in the first render cycle. It simply decides to reuse the instance of `MyInput` from the first cycle and just pass the new props ` { label: "bar" }`.
Since react reuses the component instance, the underlying lit component is also the same with `disabled = true` set in the first render cycle.

`@lit/react`, however, only updates component properties which are explicitly set in the props object passed to a component by react. See https://github.com/lit/lit/blob/546d75d9e26f4aefa3bf91d114a70a3c3fb05da2/packages/react/src/create-component.ts#L277-L285

Since the props object does not contain `disabled` as it has not been set explicitly on the second render of `MyInput`, the update is not propagated to the lit component by `@lit/react` leading to the value from the previous render cycle being set.

### Reproduction

[...]

Issue #4227, opened 2023-09-26
[labs/react] Lit Element in React Component Doesn't Apply Default Prop Value on Re-render without Property Set


rivajunior:
### Which package(s) are affected?

React (@lit-labs/react)

### Description

I am currently experiencing a problem with a component's conditional rendering. During the first render, the component features a property with a particular value. However, during the subsequent render, the property is not set, and the lit element employs the previously set value instead of the default value. I would like to apologize if my explanation is vague or unclear.

### Reproduction

You can see in the following sandbox that the Lit Element is expected to behave the same way as the React Component with default props. However, even though the component changes, the rendered values never change.

https://codesandbox.io/p/sandbox/gifted-dream-ptk4zn

### Workaround

I have created a `defaultProps` option

```ts
interface Options<I extends HTMLElement, E extends EventNames = {}> {
	react: typeof React;
	tagName: string;
	elementClass: Constructor<I>;
	events?: E;
	displayName?: string;
	defaultProps?: ComponentProps<I, E>;
}
```

And used this option to set the initial value of the `prevPropsRef` inside the `ReactComponent` returned by the `createComponent` call.

```ts
	const ReactComponent = React.forwardRef<I, Props>((props, ref) => {
		const prevPropsRef = React.useRef<Props | null>(defaultProps ?? null);
```

I made those changes in a local copy of the createComponent function.

Then, I needed to create an object to use to set the default lit element property values and to pass to the 

```ts
export const defaultButtonProps = {
	disabled: false,
	loading: false,
	type: 'button',
	variant: 'primary',
} as const;

createComponent({
	tagName: 'm-button',
	elementClass: MButton,
	displayName: 'Button',
	defaultProps: defaultButtonProps,
});
```

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

2.0.3

### Browser/OS/Node environment

Browser: Chrome 117.0.5938.92 
OS: MacOS 14.0
Node: v18.18.0
NPM: 10.1.0

justinfagnani:
When a property is no long present in the pros object, there's really no way to "remove" it from the element. The best we could do here is set it to `undefined`, though we would have to remember previous values in order to do that.

justinfagnani:
I like the defaults idea though... we have been considering something similar for restoring a value when you remove an attribute. If we add that then maybe React wrappers could utilize it automatically.

augustjk:
[...]

PR #4534, opened 2024-02-01
[react] "Unset" missing properties

augustjk:
Fixes https://github.com/lit/lit/issues/4497
Also related https://github.com/lit/lit/issues/4227

### What changed

We now collect previous element props in a map and actually check off ones that we update. We then go through any remaining and set them all as `undefined`.

#### 1. Is patch bump right for this?

It could be considered a bugfix, or it could even be considered breaking. Or a new "feature" of handling unset properties?

#### 2. Is `undefined` the correct behavior?

`undefined` feels most correct in terms of semantics of "unsetting", but I wanted to see what React's future custom element support would do and it seems `react-dom@experimental` will actually set the property as `null` if it's missing.
Playground: https://lit.dev/playground/#gist=001005b1cfb39fd03ccb37e841d256e5
(The current value of the property is logged in the console)

### Resolution

We'll include this in a patch release as the existing behavior is considered unexpected, therefore this is a bugfix.

Filed an issue with React regarding they're behavior of setting `null`. Setting to `undefined` seems like the semantically correct thing to do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,6 @@
         "./packages/labs/gen-wrapper-vue:test",
         "./packages/labs/nextjs:test",
         "./packages/labs/preact-signals:test",
-        "./packages/labs/react:test",
         "./packages/labs/rollup-plugin-minify-html-literals:test",
         "./packages/labs/ssr:test",
         "./packages/labs/ssr-dom-shim:test",
--- a/packages/react/src/test/create-component_test.tsx
+++ b/packages/react/src/test/create-component_test.tsx
@@ -334,10 +334,7 @@ suite('createComponent', () => {
       assert.equal(el.getAttribute('id'), 'foo');
       assert.equal(el.id, 'foo');
 
-      // We currently require passing `undefined` to "unset" rather than
-      // omitting the prop
-      // See https://github.com/lit/lit/issues/4227
-      render(<Tag id={undefined} />);
+      render(<Tag />);
       assert.equal(el.getAttribute('id'), null);
       assert.equal(el.id, '');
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -171,7 +171,6 @@
         "./packages/labs/gen-wrapper-vue:test",
         "./packages/labs/nextjs:test",
         "./packages/labs/preact-signals:test",
-        "./packages/labs/react:test",
```
augustjk, line 174: This feels like a band-aid. The test files in packages/labs/react are now stale. I don't think we really have reason to keep all the things in the labs directory for graduated packages like the tests and rollup builds.

AndrewJakubowicz: Strongly agree.

packages/react/src/test/create-component_test.tsx
```diff
@@ -490,17 +487,15 @@ suite('createComponent', () => {
     el.fire('foo');
```
AndrewJakubowicz, line 487: The line above this can be update to either no have the explicit `undefined`, or the comment above it can mention that both explicit `undefined` or omitting property will clear listener.

AndrewJakubowicz (approved):
Awesome!

Would be cool to link to the React issue (where they are currently setting to `null`).
Discussed options IRL.

augustjk:
Regarding React's current behavior of setting `null`, filed https://github.com/facebook/react/issues/28203

package.json
augustjk: There are. I think they can be done as a separate chore task. Nothing super pressing.

PR #4537, opened 2024-02-01
Convert analyzer template test to node:test

justinfagnani:
node:test is pretty nice, and Uvu has some rough edges. This is just a first test: new tests would be written to node:test and existing tests would be ported over, starting with the analyzer.


packages/labs/analyzer/src/test/server/lit-html/template_test.ts
```diff
@@ -4,53 +4,48 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {suite} from 'uvu';
+import {test, describe as suite} from 'node:test';
 // eslint-disable-next-line import/extensions
 import * as assert from 'uvu/assert';
```
augustjk, line 9: do we also want to replace this with `chai@5` which supports esm?

justinfagnani: oh, node has an assert library too, I just forgot to use it!

augustjk (approved):

packages/labs/analyzer/src/test/server/utils.ts
```diff
@@ -244,6 +244,30 @@ export const setupAnalyzerForTest = (
   }
 };
 
+export const setupAnalyzerForNodeTest = (lang: Language, pkg: string) => {
```
rictic, line 247: Can write setupAnalyzerForTest in terms of setupAnalyzerForNodeTest I think:

```ts
  try {
    Object.assign(ctx, setupAnalyzerForNodeTest(lang, pkg))
  } catch (error) {
    // Uvu has a bug where it silently ignores failures in before and after,
    // see https://github.com/lukeed/uvu/issues/191.
    console.error('uvu before error', error);
    process.exit(1);
  }
}

rictic (approved):
Nice! One suggestion

packages/labs/analyzer/src/test/server/utils.ts
justinfagnani: nice, done!

justinfagnani:
Arg, this is why the new tests aren't running on CI: https://github.com/nodejs/node/issues/50658#issuecomment-1806581766

Should we update Node to 21?

PR #4543, opened 2024-02-06
[lit-virtualizer] Fixes #4505 - Adds documentation for lit-virtualizer/grid

### LINKED ISSUE #4505: [labs/virtualizer] Document `grid` layout
### Should this be an RFC?

### Description

Virtualizer's `grid` layout is not currently documented; it should be.

This Playground contains most of the relevant info: https://lit.dev/playground/#gist=a5960ffd0cb67c61083b4df0f9372eeb

The relevant source is mostly in these files:

https://github.com/lit/lit/blob/main/packages/labs/virtualizer/src/layouts/shared/SizeGapPaddingBaseLayout.ts

https://github.com/lit/lit/blob/main/packages/labs/virtualizer/src/layouts/shared/GridBaseLayout.ts

N/A

JamesIves:
Adds documentation for `lit-virtualizer` `grid` layout.

Documentation was produced using examples from the Playground / code. Please let me know if you feel I need to go into more detail in any specific area. 

Closes #4505

### DIFF
--- a/packages/labs/virtualizer/README.md
+++ b/packages/labs/virtualizer/README.md
@@ -154,7 +154,129 @@ html`
 
 ### Using the `grid` layout
 
-TODO
+The grid layout lays out uniformly sized items in a regular grid. You specify an `itemSize`, which is either strictly observed or (if you use the flex option) matched as closely as possible while taking up all space along the layout's inline axis. [You can view an example of the grid layout here](https://lit.dev/playground/#gist=a5960ffd0cb67c61083b4df0f9372eeb).
+
+_Note: When using the grid layout, Virtualizer explicitly sets the dimensions of its child elements. This differs from the behavior of the default flow layout._
+
+```js
+  render() {
+    return html`
+      <lit-virtualizer
+        .layout=${grid({
+          itemSize: {width: '100px', height: '100px'}
+        })}
+        .items=${this.photos}
+        .renderItem=${photos => html`<img src=${photo.url}>`}
+      ></lit-virtualizer>
+    `;
+  }
+```
+
+#### Options
+
+The following options for available for the `grid` layout.
+
+#### gap
+
+The `gap` property determines spacing between items and can be either a single pixel value such as `8px` (the default), or by applying both a vertical and horizontal pair such as `8px 0`.
+
+Additionally `auto` can be used to split available space evenly between items, however this may only make sense for inline axis.
+
+```js
+  render() {
+    return html`
+      <lit-virtualizer
+        .layout=${grid({
+          itemSize: {width: '100px', height: '100px'},
+          gap: '0 16px'
+        })}
+        .items=${this.photos}
+        .renderItem=${photos => html`<img src=${photo.url}>`}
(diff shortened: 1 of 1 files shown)


packages/labs/virtualizer/README.md
```diff
@@ -154,7 +154,129 @@ html`
...
+
+#### Options
+
+The following options for available for the `grid` layout.
+
+#### gap
+
+The `gap` property determines spacing between items and can be either a single pixel value such as `8px` (the default), or by applying both a vertical and horizontal pair such as `8px 0`.
+
+Additionally `auto` can be used to split available space evenly between items, however this may only make sense for inline axis.
```
justinfagnani, line 183: Regarding "however this may only make sense for inline axis" does `auto` even do anything for the block axis? If not, I'd say "however this only works for the inline axis". If it _does_ do something for the block axis it'd be good to say what that is.

packages/labs/virtualizer/README.md
```diff
@@ -154,7 +154,129 @@ html`
...
+        .items=${this.photos}
+        .renderItem=${photos => html`<img src=${photo.url}>`}
+      ></lit-virtualizer>
+    `;
+  }
+```
+
+#### padding 
+
+The `padding` property determines the spaces around the edges and can either be a single value pair such as `8px 16px`, or four values to specify top, right, bottom, and left independently, for example `8px 12px 16px 24px`.
```
justinfagnani, line 202: "around the edges" around the edges of _items_? If so I'd say that, and if possible clarify the difference between padding and gap. Maybe "around each item" would be enough to imply the difference and most people should be familiar with the concepts from CSS?

justinfagnani (commented):
Thanks @JamesIves!!

A couple of small comments from me, but otherwise looks good. @graynorton can you take a quick pass?

packages/labs/virtualizer/README.md
JamesIves: I like that change 👍🏻

packages/labs/virtualizer/README.md
JamesIves: Tried to check on this, but I am running into some issues. I based this particular example on the playground, but it seems like the `auto` property may be bugged, or I'm just misunderstanding how it works.

```js
  render() {
    return html`
      <lit-virtualizer
        .layout=${grid({
          itemSize: {width: '100px', height: '100px'},
          gap: '0 auto',
          justify: 'space-between'
        })}
        .items=${this.photos}
        .renderItem=${photos => html`<img src=${photo.url}>`}
      ></lit-virtualizer>
    `;
  }
```

The following produces this error: ` Error: grid layout: column-gap cannot be set to 'auto' when direction is set to vertical`, but from what I can tell `direction` doesn't actually work for a grid layout, as changing it at all causes the virtualizer to disappear.

Perhaps I should omit documenting the auto property for now?

JamesIves:
Bumpity bump @graynorton

whjvenyl:
what‘s required to merge this one? @justinfagnani

PR #4545, opened 2024-02-09
Add missing dependency to eslint-plugin-lit

justinfagnani:
I caught this running lint locally. Not sure why it wasn't flagged on CI.

Edit(ajakubowicz):
- Applied an additional commit which fixes two wireit issues. 1. eslint-plugin should build the lit analyzer as a wireit dependency. 2. The lit analyzer wireit config has a subtle issue where running `build` after `build:ts-bundle` will delete the `typescript.js` file, because it's captured by the `build` outputs. Fix is to ignore it.


AndrewJakubowicz (approved):
It's interesting that it wasn't caught on CI. Is it possible that it implicitly works because lit-labs/analyzer is built by other tests.

A similar dependency that may be missing is a `wireit` dep on the analyzer build command as a dependency from the eslint-plugin `build` command.

Pre-approving!

rictic (approved):

PR #4547, opened 2024-02-12
feat(labs/gen-manifest): Added attributes to generated CEM manifest

maicol07:
This PR adds the attributes property to the CEM manifest generation

### DIFF
--- a/packages/labs/gen-manifest/src/index.ts
+++ b/packages/labs/gen-manifest/src/index.ts
@@ -24,6 +24,7 @@ import {
 } from '@lit-labs/analyzer';
 import {FileTree} from '@lit-labs/gen-utils/lib/file-utils.js';
 import type * as cem from 'custom-elements-manifest/schema';
+import {ReactiveProperty} from '@lit-labs/analyzer/lib/model.js';
 
 /**
  * For optional entries in the manifest, if the value has no meaningful value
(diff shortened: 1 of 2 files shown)


maicol07:
@kevinpschaaf will this be merged?

maicol07:
@kevinpschaaf any news?

maicol07:
@justinfagnani can you review this? Since @kevinpschaaf seems unavailable...

packages/labs/gen-manifest/src/index.ts
```diff
@@ -316,6 +319,23 @@ const convertEvent = (event: Event): cem.Event => {
...
+  return {
+    name:
+      typeof reactiveProperty.attribute === 'string'
+        ? reactiveProperty.attribute
+        : reactiveProperty.name,
+    type: transformIfNotEmpty(reactiveProperty.type, convertType) ?? {
+      text: 'unknown',
+    },
+    ...convertCommonInfo(reactiveProperty),
+    default: reactiveProperty.default,
```
justinfagnani, line 334: I don't think this is _entirely_ correct, but it's ok for now. The attribute default is really the property default run through the property converter. This is probably the same most of the time, but will be different for arrays and objects and non-default converters.

Also, for boolean attributes this isn't quite right, as often it'll say the default is false.

Ideally attributes don't even have defaults, because reflecting initial values is bad. But we're in a spot where they do for now.

So how about if we add the default here if the only if:
- The property reflects
- The type is String, Number, or Boolean
- There is a property default
- There is no custom converter being used

justinfagnani (changes requested):
Thanks!

One comment about defaults.

packages/labs/gen-manifest/src/index.ts
maicol07: @justinfagnani thanks! Done in 7a88ea5

maicol07:
> Thanks!
> 
> One comment about defaults.

@justinfagnani I've fixed the issue

maicol07:
@justinfagnani can this be reviewed?

PR #4558, opened 2024-02-22
Allow nonce to be provided in DOM node

Psvensso:
Allowing inline script execution is probably an even worse idea than inline styles. 
So to use a global script example like is probably not the nicest approach.
```html
<script>
 window.litNonce = 'abc123';
</script>
```

Suggest to support the nonce via a DOM element instead.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce#accessing_nonces_and_nonce_hiding

This PR add the abbility to create a style or script tag with your nonce code and keeps the global litNonce as fallback.

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -178,8 +178,12 @@ export const adoptStyles = (
   } else {
     for (const s of styles) {
       const style = document.createElement('style');
+      
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const nonce = (global as any)['litNonce'];
+      let nonce = (global as any)['litNonce'] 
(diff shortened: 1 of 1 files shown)


Psvensso:
This PR does NOT contain 
Tests
Updated doc.

It was created as a discussion point, 
if you think this is a valid approach i can update the PR to include tests and docs for a complete PR.

justinfagnani:
> Allowing inline script execution is probably an even worse idea than inline styles.

The idea is that the page can control the execution of that script with a nonce itself. Most pages with CSP will be disallowing inline script.

Reading a nonce off a style tag in the shadow root won't work because that's where we're trying to _write_ the nonce to. We need a place outside of the shadow root to read the nonce from. Reading from another script or style tag isn't correct either, because that nonce attribute is for the specific tag it lives on.

PR #4559, opened 2024-02-22
Update prettier

rictic:


rictic:
The first commit is an update of the package.json file and a mechanical `npm run format` run. The follow on commits are fixes on top of that.

justinfagnani (approved):

PR #4564, opened 2024-02-27
[labs/virtualizer] Guard `window` access to not error on import from non-browser

### LINKED ISSUE #4557: [labs/virtualizer] Virtualizer.ts throws reference error in Node.js environment

### Description

I encountered a slight problem with the following statement at the very beginning of Virtualizer.ts:

`let _ResizeObserver: typeof ResizeObserver | undefined = window?.ResizeObserver;`

During server-side rendering in a Node.js environment (e.g. a Next.js app) the window object, being a browser-specific global object, does not exist. When the code tries to access window directly without checking if it's defined, it results in a ReferenceError.

The `window?.ResizeObserver` usage suggests that it should safely check window before trying to access ResizeObserver. However, this syntax only works if window is null or undefined. If window is not declared at all (as is the case on the server), trying to evaluate window throws a ReferenceError.

I see two possible straightforward solutions to the problem. We could either keep the same elegant implementation but substitute `window?.ResizeObserver` with `globalThis?.ResizeObserver` or we could do something a bit more old-school like:

```
let _ResizeObserver: typeof ResizeObserver | undefined;

if (typeof window !== 'undefined') {
  _ResizeObserver = window.ResizeObserver;
}
```

#### Setup
Clone the sample project repo:
`git clone https://github.com/IMinchev64/lit-vitrualizer-nextjs.git`

Now run `npm install` and then run the development server with `npm run dev`.

#### Steps to Reproduce
  1. Open http://localhost:3000/ with your browser to see the sample page.
  2. Reload the page.
  3. Observe the errors in the server console/terminal

### Workaround

Avoiding use of window resolves the issue.

### Is this a regression?

### Affected versions

lit@3.1.1, @lit-labs/virtualizer@2.0.12

Browser: Microsoft Edge for Business Version 121.0.2277.112 (Official build) (64-bit),
OS: Windows 10
Node version: v21.1.0
npm version: 9.7.2

IMinchev64:
Fixes #4557


justinfagnani:
@IMinchev64 I'm taking a look. Just kicked off the tests.

justinfagnani:
@IMinchev64 this looks good. We just need to format the file and add a changeset file (run `npm run changeset`).

IMinchev64:
> @IMinchev64 this looks good. We just need to format the file and add a changeset file (run `npm run changeset`).

@justinfagnani  Is it fine now?

enesozturk:
Hey guts, any update to this? @IMinchev64 @kevinpschaaf. Thinking to use it in our product: https://github.com/lit/lit/issues/4557#issuecomment-2015077808

Thanks 🙏

justinfagnani:
@enesozturk I'm re-running CI now. Looks like the file may still need to be formatted though. cc @IMinchev64

enesozturk:
Thanks @justinfagnani, I'll be following the updates

IMinchev64:
@justinfagnani I finished running the lint and format scripts as per the [guidelines ](https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests) so now the formatting should be fine as well

justinfagnani:
@IMinchev64 the formatting looks like it's formatting way too many files. Do you have the latest dependencies? Make sure your branch is synced to make and run `npm ci`.

enesozturk:
Hey team, any update to this? @IMinchev64 @justinfagnani

IMinchev64:
> Hey team, any update to this? @IMinchev64 @justinfagnani

I'm currently looking into what went wrong with the formatting.

IMinchev64:
@justinfagnani I hope the formatting is correct now. It's just for the Virtualizer.ts now so fingers crossed all is good.

IMinchev64:
@justinfagnani Can you rerun the ci please?

enesozturk:
@justinfagnani @IMinchev64 seems CI is good, do you have a plan to release this soon? I'd be the first one try it 🙏

hrmcdonald:
Would love to get this update! This is a blocker for us using this in components right now as well.

packages/labs/virtualizer/src/Virtualizer.ts
```diff
@@ -34,7 +34,8 @@ import {ScrollerController} from './ScrollerController.js';
 // polyfill in the package; this bit of module state facilitates
 // a simple mechanism (see ./polyfillLoaders/ResizeObserver.js.)
 // for loading the polyfill.
-let _ResizeObserver: typeof ResizeObserver | undefined = window?.ResizeObserver;
+let _ResizeObserver: typeof ResizeObserver | undefined =
+  globalThis?.ResizeObserver;
```
augustjk, line 38: I don't think the `?` really does anything here just like it doesn't actually do anything for `window?.ResizeObserver` which would break if `window` is not defined in global. Either it's available in global or it's not.

```suggestion
  globalThis.ResizeObserver;
```

.changeset/rare-actors-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/virtualizer': major
+---
+
+Change window to globalThis in Virtualizer.ts to support SSR scenarios.
```
augustjk, line 5: I suppose this is technically a breaking change warranting a major version. Feels a bit weird to do a major bump just for this but if we're strictly following semver... I would also remove the wording that this "supports" SSR scenarios.

```suggestion
BREAKING: Replace top-level access to `window` with `globalThis` so that loading the module won't error in non-browser environments.

A polyfill for `globalThis` may be needed if you are targeting an environment that does not support it.
```

augustjk (commented):
I don't want to specifically block this from merging as is, but I don't think this specifically makes virtualizer *support* SSR. It may prevent it from erroring during SSR but I suspect it won't actually render anything in the server. I'm unsure if that might lead to a hydration mismatch client-side.

enesozturk:
Hey @graynorton, do you have any update on this? Me and few other people willing to try it. Virtualized list is something necessary while building performant apps. Appreciate if you can prio this 🙏🏽

hrmcdonald:
Hey @IMinchev64, I was asking about this one in the discord. It seems like bringing in `globalThis` would be a breaking change and require a major version update. 

@augustjk suggested instead of:
```typescript
let _ResizeObserver: typeof ResizeObserver | undefined =
  globalThis.ResizeObserver;
```

Going with something like:
```typescript
let _ResizeObserver: typeof ResizeObserver | undefined =
  typeof window !== 'undefined' ? window.ResizeObserver : undefined;
```

I tested this in a scenario with Next.js that throws errors because of the `window` reference with this change patched in and it worked out great to bypass the issue. I think if we can go this route it will be easier to get this merged and released.

IMinchev64:
@hrmcdonald I applied the suggestion form the comment. Thanks for the recommendation!

PR #4570, opened 2024-03-07
Fix the lit-html marker length to be consistently 9 characters.

rictic:
This was always how we intended it to behave, and since it didn't have a reliable length before this change this should be safe.


rictic:
Done

AndrewJakubowicz (approved):
Nice :D

rictic (commented):
LGTM

PR #4573, opened 2024-03-07
[benchmarks] Update chromedriver with fixed CDN url

augustjk:
Our benchmarks workflow now tries to auto download matching version of chromedriver for the chrome available in the github action runner.

But recently the CDN urls for the chromedriver has changed https://github.com/giggio/node-chromedriver/issues/448 causing pipeline failures.

We need the latest version of chromedriver that contains the fix https://github.com/giggio/node-chromedriver/commit/6c14fa780aeea39dc066ca9e45d8a85dd58c9af7 which will find and use the correct CDN url.


AndrewJakubowicz (approved):
Nice clean fix! :D

Issue #4410, opened 2023-11-21
[labs/nextjs] Support Next.js 14


augustjk:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

The plugin still should work with the Page router even in Next.js 14.

We could include it in our peer dependency sem ver to allow installs. We should still clearly document limitation with App router.

foadtkf:
what about the app router?
![image](https://github.com/lit/lit/assets/67990762/555b390d-9d0a-4ed4-b79e-ed46b169f41d)

augustjk:
@foadtkf support for app router is tracked in a separate issue https://github.com/lit/lit/issues/3657

augustjk:
Supporting v14 seems a bit more complicated as they're now bundling `react/jsx-dev-runtime` and `react/jsx-runtime` rather than it being external for us to use the normal module replacement plugin. See https://github.com/vercel/next.js/discussions/46470#discussioncomment-8375627 and https://github.com/vercel/next.js/pull/55362

We may need to make users specify `jsxImportSource` instead and also forego with wrapping the webpack externals config. This means we no longer have the ability to touch any jsx-runtime imports within node_modules code, e.g. npm installed react components that use bare Lit components within them. This was considered a pretty niche use case to support but nonetheless disappointing to have to give up.

Monkey-patching of `React.createElement` still seems to work.

augustjk:
Found a way to monkey-patch everything without resorting to module replacement in above linked PR.

Issue #3657, opened 2023-02-08
[labs/nextjs] Support for the `app` directory


augustjk:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Other/unknown (please mention in description)

### Description

Next.js v13 introduced the `app` directory, an alternative to the existing `pages` directory. https://nextjs.org/blog/next-13#new-app-directory-beta

This feature is considered beta and not recommended for production at this time.

The current implementation of the Lit SSR plugin for Next.js won't work for users using this as we're looking specifically for the `pages` directory for the entry points.

Supporting this may be as simple as adding to the test regex but will require testing.

### Alternatives and Workarounds

.

ericis:
@augustjk et al, this is no longer marked as `experimental` in the NextJS configuration when using "create-next-app".
https://github.com/vercel/next.js/blob/a7601213ab1bd2fd75e120907cd7c70278d7ebc7/packages/create-next-app/templates/app/js/next.config.js#L2

```js
/** @type {import('next').NextConfig} */
const nextConfig = {}

module.exports = nextConfig
```

The "old" way used to look like:

```js
/** @type {import('next').NextConfig} */
const nextConfig = {
  experimental: {
    appDir: true,
  },
}

module.exports = nextConfig
```

augustjk:
`app` directory support definitely looks worthwhile to pursue. It does appear that something with the App router breaks Declarative Shadow DOM parsing even for browsers that should support it, so it won't be as simple as updating the entrypoint regex.

I'll investigate and add updates here.

augustjk:
@datradito sorry I haven't gotten around to investigating fully yet as we've been focused on other things for the Lit 3.0 release. I can't say for sure when I'll get to it but it's high on my list of things to do after 3.0.

ChellappanRajan:
Hi @augustjk, when we consume the @labs/react wrapped web component in a Next.js application and use the app router, we encounter this error. Do you think this could be due to the app's layout?

![image](https://github.com/lit/lit/assets/22997191/63cba807-6ce9-4520-8059-14095d0a6bcd)

steveworkman:
There's a lot of [really good investigation in to this from @ceIia & others on Discord](https://discord.com/channels/1012791295170859069/1138460266376474726). In short:
- Reproduction: https://github.com/ceIia/lit-next-ssr
- `"use client"` makes the error go away, but switches to CSR
[...]

PR #4575, opened 2024-03-09
[labs/nextjs, labs/ssr-react, lit/react] Add support for Next.js v14 and App Router

augustjk:
Fixes https://github.com/lit/lit/issues/4410 and https://github.com/lit/lit/issues/3657

Apologies for the big PR. It'll be easier to review commit by commit.

## Explanation of changes:

re: f10d7eeff4f64bd418700d389786884e2913dc61

Next.js v14 had some changes to how it bundles its copy of React so how we were hijacking imports of `react/jsx-runtime` and `react/jsx-dev-runtime` with the Webpack module replacement were not working described here https://github.com/vercel/next.js/discussions/46470#discussioncomment-8375627

Since React is still all CommonJS, I've resorted to monkey-patching everything including the runtime JSX functions. Also, a workaround in how we access the module to patch was added to handle seemingly inconsistent es module interop behavior in webpack, e.g. `(React.default || React).createElement = ...`. This addresses errors like seen here https://github.com/lit/lit/issues/3657#issuecomment-1799934662

This has a nice benefit that all the hijacking/patching is possible with just a single import of `@lit-labs/ssr-react/enable-lit-ssr.js` which allows the `@lit-labs/nextjs` plugin to be a lot simpler (see next commit). Users can also choose not to use the plugin and just add this one line import themselves. It should also make integration with other React meta-framework, namely Remix, easier (need to test this).

~~I also added a small check to make sure we don't wrap `createElement` and the jsx functions if they're already wrapped. This is mainly useful for when both `@lit-labs/ssr-react/enable-lit-ssr.js` is installed, _and_ the `"jsxImportSource": "@lit-labs/ssr-react"` is set, as it happens in our integration test suite.~~ This doesn't seem to actually be a problem.

These changes only affect `@lit-labs/ssr-react` package.

---

re: 9702f9155d0a8d235553d79843b42ccf6cdcd80b

With the changes above, the plugin can become a lot simpler. I've removed all the hacky webpack externals modification to replace `react/runtime-jsx` at the module level since we're just monkey-patching it globally now.

I've updated the peer dependencies to now support `next@13` and `next@14`, and dropped `next@12`. I debated whether we should keep v12 support, but the changes here mainly apply to next v13 and v14 for the app router support and I haven't tested these changes in v12 (and we likely won't be going forward, see commit about adding example projects below). Users of v12 can stick with the current published version.

---

re: d5dcabb2972512cd13715072d08df0c62089c23d

While manually testing the Next v14 example, I noticed server rendering of `@lit/react` wrapped components had incorrect initial props for production builds. We had coordination between `@lit/react` and `@lit-labs/ssr-react` based on the patched `createElement` function name, but that name check was failing perhaps as it got lost during production build minification. (I noticed this a while back when I was testing out `@lit-labs/ssr-react` in Remix too.) `@lit-labs/ssr-react` will now add a global flag which `@lit/react` node build will check for.

---

re: 02a0c05930ca75144396000a500f594aeb0e6d3b

I renamed the existing `examples/nextjs` package to `nextjs-v13` and added `nextjs-v14` and `nextjs-v14-app` examples.
`nextjs-v14` is basically a direct copy of the existing v13 example just with the next version bumped. It still uses the pages router.
`nextjs-v14-app` was modified further from there to use the app router directory structure, using `layout.tsx` instead of `_app.tsx`, as well as moving the custom element and wrapped component behind the `'use client';` boundary. The Lit SSR patching does not work within React Server Components. This note is added to the example README as well as in the changelog entry up above.

## Testing

Testing was done manually by running each example projects and inspecting the page source to confirm declarative shadow DOM was server rendered, there were no React hydration errors, and components became interactive.
 
## Future work

It would be much better to have automated integration/e2e testing for these test nextjs projects.

While I updated the README for the `@lit-labs/nextjs` package, it would be good to have a fuller explanation of the caveats of app router support somewhere like on lit.dev's frameworks section.

## Notes for releasing

The `@lit-labs/ssr-react` dependency minimum version should be updated for the next release of `@lit-labs/nextjs`. Both are getting "breaking" version bumps (minor for 0. package) so changesets might already do this but adding a note here as a reminder to check it.


packages/labs/nextjs/src/index.ts
```diff
@@ -38,73 +39,21 @@ export = (pluginOptions: LitSsrPluginOptions = {}): NextConfig =>
...
           options: {
             imports,
           },
         });
 
-        if (isServer && nextRuntime === 'nodejs') {
-          // Next.js uses this externals setting to skip webpack bundling
-          // external code for server environments but we specifically want to
-          // catch react/jsx-runtime imports for external packages as well.
-          // This might change in a future update to Next.js.
```
AndrewJakubowicz, line 58: Love a delete! Awesome!

.changeset/chilled-plants-reflect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lit-labs/ssr-react': patch
+'@lit/react': patch
+---
+
+Use a global flag to detect whether `@lit-labs/ssr-react/enable-lit-ssr.js` was used for coordinated props when server rendering components made with `@lit/react`.
```
AndrewJakubowicz, line 6: Nit: `coordinated` -> `coordinating`

packages/react/src/create-component.ts
```diff
@@ -309,7 +309,8 @@ export const createComponent = <
       // element properties in a special bag to be set by the server-side
       // element renderer.
       if (
-        React.createElement.name === 'litPatchedCreateElement' &&
+        (React.createElement.name === 'litPatchedCreateElement' ||
```
AndrewJakubowicz, line 312: Why is the patched name not enough?

examples/nextjs-v14-app/README.md
```diff
@@ -0,0 +1,9 @@
+# Lit in Next.js example
+
+This is a barebones demonstration of a Web Component authored with Lit working
+in a Next.js project with the App Router. You can also see an example of `@lit/react` usage.
+
+It uses the plugin from `@lit-labs/nextjs` to enable deep server rendering of
+Lit components.
+
+Note: Both bare custom elements and `@lit/react` wrapped components are placed in files with the `'use client';` directive. Deeply server rendering Lit components with declarative shadow DOM is not supported in React Server Components as React Flight server rendered data will contain the `<template>` element causing React hydration errors, and custom element definitions will not get included with client bundle.
```
AndrewJakubowicz, line 9: `React Flight server`? I'm not able to find what this is referring to.

examples/nextjs-v14-app/README.md
```diff
@@ -0,0 +1,9 @@
+# Lit in Next.js example
```
AndrewJakubowicz, line 1: When testing this locally with `JavaScript Disabled`, the "Hello, Friend!" component renders, but the component beneath doesn't seem to render.
My expectation is that two components would SSR with DSD.


Edit: Could this be that I need to link to the new packages?

AndrewJakubowicz (commented):
Absolutely awesome work!

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
```diff
@@ -19,49 +19,50 @@ import type {
...
-  return function litPatchedCreateElement<P>(
-    type: ElementType<P>,
-    props: PropsWithChildren<P> | null,
-    ...children: ReactNode[]
-  ): ReactElement {
-    if (isCustomElement(type)) {
-      const {shadowContents, elementAttributes, templateAttributes} =
-        renderCustomElement(type, props);
+) =>
+  originalCreateElement.name !== 'litPatchedCreateElement'
```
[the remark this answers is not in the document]
augustjk: Thank you! I agree, short path first is better readability. Opted to remove the ternary implicit return and do an explicit early return for the short path.

The `if (isCustomElement(...))` is less amenable to this because there's another condition check after rendering the custom element, for which the final return is a fallback for so can't early return with that.

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
```diff
@@ -19,49 +19,50 @@ import type {
...
+    ? function litPatchedCreateElement<P>(
```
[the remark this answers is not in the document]
augustjk: That is a possibility. The name check here is unfortunately not robust, which is why I added a global flag in `@lit-labs/ssr-react/enable-lit-ssr.js`.

I'm not sure how robust this really needs to be. I added it primarily because our integration test runs the monkey-patching and the jsxImportSource modification in the same context, which would be discouraged in practice.

Actually, now that I think about it, it might be better these checks just need to happen at the site of the monkey patching or replacement. Not in the wrapping function itself. I'm going to try that out!

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
```diff
@@ -0,0 +1,125 @@
...
+              },
+            });
+
+            // Call `jsxs` instead of `jsx` so that React doesn't incorrectly
+            // interpret that the children array was dynamically made and warn about
+            // missing keys
+            return originalJsxs(type, {
+              ...props,
+              ...elementAttributes,
+              children: [templateShadowRoot, props.children],
```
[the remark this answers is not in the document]
augustjk: The idea is we might get JSX like
```html
<my-element>
  <p>light dom child</p>
</my-element>
```
The `<p>` (as a React Element representation) would already be in `props.children`. We're going to create a `<template>` element with `React.createElement()` that'll contain the DSD, which is `templateShadowRoot`. This needs to be added as a child to `my-element`, so we create/replace the `children` prop and prepend it.

Resulting server render by ReactDOM results in
```html
<my-element>
  <template shadowrootmode="open">...</template>
  <p>light dom child</p>
</my-element>
```

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
```diff
@@ -0,0 +1,125 @@
...
+              ...templateAttributes,
+              dangerouslySetInnerHTML: {
+                __html: [...shadowContents].join(''),
+              },
+            });
+
+            // Call `jsxs` instead of `jsx` so that React doesn't incorrectly
+            // interpret that the children array was dynamically made and warn about
+            // missing keys
+            return originalJsxs(type, {
```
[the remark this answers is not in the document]
augustjk: It's a good call. I think the answer is no, and as mentioned about I'll move the monkey-patch check for the other one outside of the `wrap` function too.

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
```diff
@@ -0,0 +1,125 @@
...
+      ) {
+        if (isCustomElement(type)) {
+          const {shadowContents, elementAttributes, templateAttributes} =
+            renderCustomElement(type, props);
+
+          if (shadowContents) {
+            const templateShadowRoot = createElement('template', {
+              ...templateAttributes,
+              dangerouslySetInnerHTML: {
+                __html: [...shadowContents].join(''),
```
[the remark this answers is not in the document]
augustjk: Good point, that is correct we only support sync SSR. It would be good to change this to `collectResultSync` which'll throw if any promises are seen.

.changeset/quiet-shirts-trade.md
```diff
@@ -0,0 +1,9 @@
+---
+'@lit-labs/nextjs': minor
+---
+
+Add support for Next.js 14 and App Router. No longer supports Next.js 12.
+
+Note: Components in the App Router are by default React Server Components. Deep SSR of Lit components does **not** work within server components as they result in React hydration mismatch due to the presence of the `<template>` element in React Flight serialized server component data, and the custom element definitions will not be included with the client bundle either.
```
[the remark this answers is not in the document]
augustjk: "Flight" appears to be React/Next team's code word for the system/format behind server component serialization, serving/parsing. I found reference to this name in comments and functions as I was diving through their codebases, e.g. https://github.com/vercel/next.js/blob/canary/packages/next/src/server/app-render/use-flight-response.tsx

This blog post explains it. https://timtech.blog/posts/react-server-components-rsc-no-framework/#react-server-%26-react-flight-apis

But there's no official doc that mentions it by name. Only react/next team members mentioning it in comments and PRs like https://github.com/vercel/next-react-server-components/issues/13#issuecomment-988438962

Since many would wonder what that means the same as you, I'll remove "Flight" and just call them RSC payloads or something like that.

packages/react/src/create-component.ts
augustjk: I haven't dug super into it but my suspicion is that production build can mangle the function name. I think it'd be fine to remove the name check all-together here too in favor of just checking the global flag. I left it more so as a holdover for those who won't immediately be updating `@lit-labs/ssr-react` but maybe that's over thinking it.

examples/nextjs-v14-app/README.md
augustjk: Answered in https://github.com/lit/lit/pull/4575#discussion_r1520451524
Will remove "Flight" from here too.

examples/nextjs-v14-app/README.md
augustjk: It appears to show fine for me. Depending on how you ran the example.. if you're doing this outside the monorepo, you'd need `@lit-labs/ssr-react`, and `@lit/react` also linked to the local copy, as well as `@lit-labs/nextjs`.

But run it within this branch of the monorepo and everything should already link to the monorepo copy.

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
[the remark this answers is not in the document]
augustjk: I removed the name check all-together. It doesn't actually seem to be meaningful.

packages/labs/ssr-react/src/lib/node/wrap-create-element.ts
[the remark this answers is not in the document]
augustjk: Ended up removing the name check. But still moved to explicit return over arrow function implicit return. Seems more readable that way.

.changeset/polite-rice-stare.md
```diff
@@ -0,0 +1,9 @@
+---
+'@lit-labs/ssr-react': minor
+---
+
+The Node build of `@lit-labs/ssr-react/enable-lit-ssr.js` will now also monkey-patch `react/runtime-jsx` to include logic for deeply SSR Lit components without modifying `jsxImportSource` in tsconfig.
```
justinfagnani:
```suggestion
The Node build of `@lit-labs/ssr-react/enable-lit-ssr.js` now also monkey-patches `react/runtime-jsx` to include logic for deeply server-rendering Lit components without modifying `jsxImportSource` in tsconfig.
```

packages/labs/ssr-react/README.md
```diff
@@ -12,15 +12,9 @@ This package provides tools to integrate [`@lit-labs/ssr`](../ssr/README.md) wit
...
-
-The classic JSX transform replaces JSX expressions with `React.createElement()` function calls. In the default mode, it requires that `React` is imported and available in the scope of the JSX file.
-
-This package provides a couple different ways to handle the classic runtime:
-
-#### Monkey patching `React.createElement()` (recommended)
+### Monkey patching React element creation (recommended)
 
-This package provides a module that, when imported in a server environment, has the side-effect of monkey patching `React.createElement()` to be enhanced to add the declarative shadow DOM output to registered custom elements. This can be imported at the entry point of the application before `React` is imported.
+This package provides the `@lit-labs/ssr-react/enable-lit-ssr.js` module which, when imported in a server environment, has the side-effect of monkey patching `React.createElement()` and runtime JSX functions to be enhanced to add the declarative shadow DOM output to registered custom elements. This can be imported at the entry point of the application before `React` is imported or any JSX containing Lit components is written.
```
justinfagnani, line 17:
```suggestion
This package provides the `@lit-labs/ssr-react/enable-lit-ssr.js` module which, when imported in a server environment, has the side-effect of monkey patching `React.createElement()` and runtime JSX functions to be enhanced to add the declarative shadow DOM output to registered custom elements. This can be imported at the entry point of the application before `React` is imported or any JSX containing Lit components is evaluated.
```

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
```diff
@@ -0,0 +1,125 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import type {jsx, jsxs} from 'react/jsx-runtime';
+import type {jsxDEV} from 'react/jsx-dev-runtime';
+import {createElement, type ReactNode, type ElementType} from 'react';
+import {renderCustomElement} from './render-custom-element.js';
+import {isCustomElement} from '../utils.js';
+
+export const wrapJsx = (originalJsx: typeof jsx, originalJsxs: typeof jsxs) =>
```
justinfagnani: comment?

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
```diff
@@ -0,0 +1,125 @@
...
+  originalJsx.name !== 'litPatchedJsx'
+    ? function litPatchedJsx<P extends {children?: ReactNode}>(
```
justinfagnani: Like @AndrewJakubowicz 's comment above, you might need to make sure that this name isn't mangled. You could also detect this function instance by setting and checking a well-known property on it, or assigning it to a property on `globalThis`.

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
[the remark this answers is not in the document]
justinfagnani: I think you should definitely do that. It'll help people out if we enable async components.

packages/labs/ssr-react/src/node/enable-lit-ssr.ts
```diff
@@ -11,7 +11,25 @@
...
+import ReactJSXDevRuntime from 'react/jsx-dev-runtime';
 import {wrapCreateElement} from '../lib/node/wrap-create-element.js';
+import {wrapJsx, wrapJsxDev, wrapJsxs} from '../lib/node/wrap-jsx.js';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-(React.createElement as any) = wrapCreateElement(React.createElement);
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// The `.default ||` is used to get around inconsistent behavior with how
+// tools like webpack seem to do es module interop.
+Object.assign((React as any).default || React, {
```
justinfagnani, line 23:
```suggestion
Object.assign((React as any).default ?? React, {
```

packages/labs/ssr-react/src/node/enable-lit-ssr.ts
```diff
@@ -11,7 +11,25 @@
...
+  createElement: wrapCreateElement(React.createElement),
+});
+if (process.env.NODE_ENV === 'production') {
+  Object.assign((ReactJSXRuntime as any).default || ReactJSXRuntime, {
```
justinfagnani, line 27:
```suggestion
Object.assign((ReactJSXRuntime as any).default ?? ReactJSXRuntime, {
```

packages/labs/ssr-react/src/node/enable-lit-ssr.ts
```diff
@@ -11,7 +11,25 @@
...
+    jsx: wrapJsx(ReactJSXRuntime.jsx, ReactJSXRuntime.jsxs),
+    jsxs: wrapJsxs(ReactJSXRuntime.jsxs),
+  });
+} else {
+  Object.assign((ReactJSXDevRuntime as any).default || ReactJSXDevRuntime, {
```
justinfagnani, line 32:
```suggestion
Object.assign((ReactJSXDevRuntime as any).default ?? ReactJSXDevRuntime, {
```

justinfagnani (approved):
lgtm!

packages/labs/ssr-react/src/lib/node/wrap-jsx.ts
augustjk: I removed the checking by function name so it is no longer a problem if the name gets minified.

PR #4586, opened 2024-03-19
[labs/redux] Reactive controller for subscribing to a Redux store

augustjk:
Putting this up as draft PR. There's a few more things needed here. Not sure if/when I might be able to get to it.

In playground example form: https://lit.dev/playground/#gist=79560a5bb4cde3a007c18742b2a9ec8a

### TODO
- Add tests.
- JSDoc on user facing exports.
- README.
- Need to consider if surfacing `dispatch` from the `Selector` controller instance makes sense. What if a component only needs access to `dispatch` without any state? Should there be a separate controller for `dispatch`, or dual-purpose `Selector` controller and make the `selector` function optional?
- Decorators for selector and dispatch for better ergonomics over directly using controllers? These would be more akin to `useSelector` and `useDispatch` hooks from `react-redux`.


examples/redux/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules/
+/dist/
```
justinfagnani, line 2: who's writing to `/dist/`, Vite? Leave a comment?

packages/labs/redux/LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2017 Google LLC. All rights reserved.
```
justinfagnani, line 3: 2024?

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+  @provide({context: storeContext})
+  private _store = store;
+
+  ...
+}
+```
+
+### Connect a child component to the provided store
+
+Note: The component must be a child of the provider above. Attempting to connect without a store provided via context will throw an error.
```
justinfagnani, line 67: I wonder if you want to mention ContextRoot here...

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Using the `Connector` controller
+
```
justinfagnani, line 70: Add a short description of the the Connector controller is first?

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+#### Using the `@connect()` and `@dispatch()` decorators
+
+```ts
+import {select, dispatch} from '@lit-labs/redux';
+import {increment} from './counter-slice.js';
+
+export class CounterElement extends LitElement {
+  @select((state) => state.counter.value)
+  _count;
```
justinfagnani, line 129: Make this `protected`?

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+  @dispatch()
```
justinfagnani, line 131: I find this a little odd. Is there a way to vend a utility function instead, so the code below can do:

```ts
  _incrementCount() {
    dispatch(this, increment());
  }
```

Or maybe the decorator could go on the dispatching method?

```ts
  @dispatch()
  _incrementCount() {
    return increment();
  }

```

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+  _dispatch;
```
justinfagnani, line 132: Make `protected`

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+  _incrementCount() {
```
justinfagnani, line 134: Make `private`

packages/labs/redux/README.md
```diff
@@ -0,0 +1,231 @@
...
+    `;
+  }
+}
+```
+
+The `@select()` decorator can take a selector function as an argument or an options object to be passed to the underlying `Connector` controller.
+
+The `@dispatch()` decorator takes no arguments.
+
+Note: Like all the Lit 3 decorators, these decorators will work as either TypeScript experimental decorators or [standard decorators](https://lit.dev/docs/components/decorators/#standard-decorators). When running as standard decorators, `@select()` decorated fields must also have the `accessor` keyword. `@dispatch()` decorated fields do not have this requirement.
```
justinfagnani, line 151: Standard decorators can decorate private fields too.

packages/labs/redux/src/lib/connector.ts
```diff
@@ -0,0 +1,138 @@
...
+  equalityCheck?: EqualityCheck;
+};
+
+/**
+ * Reactive controller which subscribes to a Redux store and ties changes to a
+ * custom element's reactive life cycle.
+ *
+ * This controller must be used with a provider that supplies the Redux store
+ * using the `storeContext` from this package using the [Context
+ * Protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md).
```
justinfagnani, line 54: Add link to `@lit/context` too

packages/labs/redux/src/lib/connector.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Static method to create a typed version of the `Connector` constructor.
+   * This allows type checking for the `selector` option, the selected value, as
+   * well as the `dispatch` method provided by the connector.
+   *
+   * @example
+   *
+   * ```ts
+   * // `AppStore` type gotten from created Redux store
```
justinfagnani, line 72: Remove "gotten"

packages/labs/redux/src/lib/connector.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  hostConnected() {
+    new ContextConsumer(this._host, {
+      context: storeContext,
+      subscribe: true,
+      callback: (store) => {
+        this._store = store as S;
+      },
+    });
+    if (this._store === undefined) {
```
justinfagnani, line 117: Since this context subscribes, is this error correct here? What if the page is using a ContextRoot? Maybe this should be some kind of debug warning.

packages/labs/redux/src/lib/connector.ts
```diff
@@ -0,0 +1,138 @@
...
+    host: ReactiveControllerHost & HTMLElement,
+    options?: ConnectorOptions<S, V>
+  ) {
+    this._host = (host.addController(this), host);
+    this._selector = options?.selector;
+    this._equalityCheck = options?.equalityCheck ?? tripleEquals;
+  }
+
+  hostConnected() {
+    new ContextConsumer(this._host, {
```
justinfagnani, line 110: I think you need to make the consumer in the constructor and hold a reference to it. That's how I usually show examples of controller composition.

The callback in this case will keep the consumer instance alive even after this method exits, which means on a disconnect/reconnect you will create another long-lived consumer, causing a memory leak. This could be noticable for connected components inside `repeat()` or `cache()`.

packages/labs/redux/README.md
augustjk: Yes, I did think this was awkward. Kind of figured we can iron out a bit more from user feedback while the package is in labs.

I considered the idea of `dispatch` being provided to a decorated method somehow like your second suggestion, which is similar in idea to [`react-redux`'s `mapDispatchToProps`](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-a-function).

packages/labs/redux/src/lib/connector.ts
augustjk: Would storing the consumer instance and checking to not re-create work? I think I had some timing issues when instantiating in the constructor because the context value (redux store) would not be immediately available.

packages/labs/redux/src/lib/connector.ts
augustjk: I had that thought too..

I was hoping if it might be possible to kind of force this requirement on users to force the context provider be always set up first because I otherwise we have to consider that without having the store on start up, the `selected` value could be undefined and users need to always guard for that.

So if a user selected for an object, they would basically have to always null check any property access to it for their first render. Maybe that's not the worst thing, but I really like the idea that, assuming the user provided the correct store type and a valid selector, they can always expect a certain value to exist and use as such within their component code.

packages/labs/redux/src/lib/selector.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type EqualityCheck = (a: unknown, b: unknown) => boolean;
+export type SelectorFunction<S, V> = (state: S) => V;
+
+export type SelectorOptions<S extends Store, V> = {
+  selector: SelectorFunction<ReturnType<S['getState']>, V>;
+  equalityCheck?: EqualityCheck;
+};
+
+export class Selector<S extends Store, V> implements ReactiveController {
```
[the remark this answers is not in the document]
augustjk: Yup! This comment is on an old version of the changes. It has been renamed to `Connector`.

packages/labs/redux/src/lib/connector.ts
justinfagnani: When composing controllers, you should call the inner `hostConnected()` from the outer `hostConnected()`, etc. That should make the timing the same as if you directly added a ContextConsumer to the host.

packages/labs/redux/README.md
pgoforth: Providing arguments to the decorated method would create additional mental overhead for the implementer, so I don't think that's the right approach. I would prefer option #2 because it keeps it simple, you don't have to worry about calling dispatch yourself, and it can cleanly work with Redux middleware like [redux-thunk](https://github.com/reduxjs/redux-thunk)

PR #4591, opened 2024-03-22
[labs/observers] Prevent failing in SSR

### LINKED ISSUE #4590: [labs/observers] Prevent failing in SSR

### Description

All observer controllers crash when used in SSR as their constructors don't count on the fact that the window object is not available in the server environment.

### Reproduction

Any component that uses some observer controller will fail in SSR:

```ts
@customElement('some-component')
class SomeComponent extends LitElement {
    readonly mutationController = new MutationController(this, {
        config: {
	    subtree: true,
	    characterData: true,
	    childList: true,
        })
}
```

### Workaround

You could just add an inline `isServer` check and not initialized the controller. Not very ergonomic though!

```diff
@customElement('some-component')
class SomeComponent extends LitElement {
-    readonly mutationController = new MutationController(this, {
+    readonly mutationController = isServer ? undefined : new MutationController(this, {
        config: {
	    subtree: true,
	    characterData: true,
	    childList: true,
        })
}
```

### Is this a regression?

### Affected versions

All versions

All Node versions

a11delavar:
This PR makes observer controllers not to initialize the observers when the constructor is called in SSR environment, preventing a crash in SSR environment. Fixes #4590.

The `isServer` check being before `window.xObserver` check is intentional as `window` won't exist in Node. Also there is no need to warn the developer in SSR as opposed to when we are in the browser but the observer is not supported.

There is clean-up potential after #3386 has been merged, namely adding an abstract `createObserver()` method to the base class for all common browser observers and moving the `isServer` check to the base class and conditionally calling it.


justinfagnani (approved):
Looks good to me. @sorvell wdyt?

packages/labs/observers/package.json
```diff
@@ -168,6 +168,7 @@
     "@types/trusted-types": "^2.0.2"
   },
   "dependencies": {
+    "lit-html": "^2.4.0",
```
augustjk, line 171: Semver should at least include Lit 3
```suggestion
    "lit-html": "^2.4.0 || ^3.0.0",
```
But I wonder if it's worth it to create a Node build for this package instead of adding this dependency?

sorvell (approved):

packages/labs/observers/package.json
augustjk: Also needs to update package-lock.json to reflect the added dependency for `npm ci` to work.

justinfagnani: We need to do one of these, so I'm going to un-approve.

justinfagnani (changes requested):
There's an important unresolved comment now

packages/labs/observers/package.json
a11delavar: Updated package and package-lock files to include lit-html 3.x

augustjk (approved):
Thanks!

justinfagnani (approved):
Thanks!

PR #4598, opened 2024-03-24
[lit/context] Use `target.set` for setting context values

jun-sheaf:
`target.set` is the correct way to set values in decorators. It supports nested decorators and private fields.

### DIFF
--- a/.changeset/thirty-shirts-peel.md
+++ b/.changeset/thirty-shirts-peel.md
@@ -2,4 +2,4 @@
 '@lit/context': patch
 ---
 
-Use `target.set` instead of member assignment
+Use `target.set` instead of member assignment in `@consume()` decorator.
--- a/.changeset/thirty-shirts-peel.md
+++ b/.changeset/thirty-shirts-peel.md
@@ -1,0 +1,5 @@
+---
+'@lit/context': patch
+---
+
+Use `target.set` instead of member assignment
--- a/packages/context/src/lib/decorators/consume.ts
+++ b/packages/context/src/lib/decorators/consume.ts
@@ -55,7 +55,7 @@ export function consume<ValueType>({
   ) => {
     if (typeof nameOrContext === 'object') {
       // Standard decorators branch
-      nameOrContext.addInitializer(function (): void {
+      nameOrContext.addInitializer(function () {
         new ContextConsumer(this, {
           context,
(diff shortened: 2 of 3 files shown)


justinfagnani:
Thanks @jun-sheaf!

We need a changeset file as part of the PR. See https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

jun-sheaf:
> Thanks @jun-sheaf!
> 
> We need a changeset file as part of the PR. See https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

Done.

.changeset/thirty-shirts-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/context': patch
+---
+
+Use `target.set` instead of member assignment
```
augustjk, line 5:
```suggestion
Use `target.set` instead of member assignment in `@consume()` decorator.
```

packages/context/src/lib/decorators/consume.ts
```diff
@@ -47,18 +47,19 @@ export function consume<ValueType>({
...
   subscribe?: boolean;
 }): ConsumeDecorator<ValueType> {
-  return (<C extends ReactiveElement, V extends ValueType>(
-    protoOrTarget: ClassAccessorDecoratorTarget<C, V>,
-    nameOrContext: PropertyKey | ClassAccessorDecoratorContext<C, V>
+  return ((
+    protoOrTarget: ClassAccessorDecoratorTarget<ReactiveElement, ValueType>,
+    nameOrContext:
+      | PropertyKey
+      | ClassAccessorDecoratorContext<ReactiveElement, ValueType>
```
augustjk, line 54: Nice. I guess the generic wasn't necessary since we're casting the type of the returned function anyway. Perhaps we want to do the same change to the `@provide()` decorator just for consistency's sake?

jun-sheaf: Could you push the change? I don't have time too, apologies!

augustjk: Sure thing.

augustjk (approved):
Thank you!

jun-sheaf:
Someone with write access has to merge this.

PR #4605, opened 2024-03-30
Add Task args as argument to Task callbacks

### LINKED ISSUE #4561: [task] Passing the parameters inside the task render callback
### Should this be an RFC?

Task (@lit/task)

### Description

Which package(s) are affected?
Task (@lit/task)

Description
It would be super useful to get the parameters inside the task render callback, for example to adjust the render output using these params.
For example: 
```js
import {Task, TaskStatus} from '@lit/task';
// ...

  @state()
  private _userId: number = -1;

  private _apiTask = new Task(
    this,
    ([userId]) =>
      fetch(`//example.com/api/userInfo?${userId}`).then((response) =>
        response.json()
      ),
    () => [this._userId]
  );

  render() {
    return html`
      <div>User Info</div>
      ${this._apiTask.render({
        pending: ([userId]) => html`Loading user info by id:${userId}...`,
        error: ([userId]) => html`Loading user by id:${userId} failed :(`,
        complete: (user) => html`${user.name}`,
      })}
      <!-- ... -->
    `;
  }
} 
```

Is this a regression?
No.

Affected versions
all

Browser/OS/Node environment
all

It's possible to achieve the same result by setting taskParams manualy in the LitElement instance. But I would like to avoid unecessary code
For example:

```js
import { Task, TaskStatus } from "@lit/task";
// ...

    private _apiParams: [string] = [""];

    private _apiTask = new Task(this, {
        async task([userId]) {
            this._apiParams = [userId];

            const response = await fetch(`//example.com/api/userInfo?${userId}`);
            const json = await response.json();
            return json;
        },
        autoRun: false,
    });

    /* I can now call this._apiTask.run() with different params */

    render() {
        return html`
            <div>User Info</div>
            ${this._apiTask.render({
                pending: ([userId]) => html`Loading user info by id:${userId}...`,
                error: ([userId]) => html`Loading user by id:${userId} failed :(`,
                complete: (user) => html`${user.name}`,
            })}
            <!-- ... -->
        `;
    }
}
```

Snapstromegon:
Fixes #4561

This adds the args from the args function as an argument to the callbacks like mentioned here: https://github.com/lit/lit/issues/4561#issuecomment-1964563630

This is my first contribution to the projects, so I tried following CONTRIBUTING.md, but developing on Windows is hard in this project, since executing tests is not possible (other scripts only work in WSL and not natively in windows - at least for me).

### DIFF
--- a/packages/task/src/task.ts
+++ b/packages/task/src/task.ts
@@ -37,11 +37,11 @@ export const initialState = Symbol();
 
 export type TaskStatus = (typeof TaskStatus)[keyof typeof TaskStatus];
 
-export type StatusRenderer<R> = {
-  initial?: () => unknown;
-  pending?: () => unknown;
-  complete?: (value: R) => unknown;
-  error?: (error: unknown) => unknown;
+export type StatusRenderer<R, D extends ReadonlyArray<unknown>> = {
+  initial?: (args?: D) => unknown;
+  pending?: (args?: D) => unknown;
+  complete?: (value: R, args?: D) => unknown;
+  error?: (error: unknown, args?: D) => unknown;
 };
 
 export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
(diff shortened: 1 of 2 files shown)


packages/task/src/task.ts
```diff
@@ -37,11 +37,11 @@ export const initialState = Symbol();
 
 export type TaskStatus = (typeof TaskStatus)[keyof typeof TaskStatus];
 
-export type StatusRenderer<R> = {
-  initial?: () => unknown;
-  pending?: () => unknown;
-  complete?: (value: R) => unknown;
-  error?: (error: unknown) => unknown;
+export type StatusRenderer<R, D extends ReadonlyArray<unknown>> = {
```
justinfagnani, line 40: For backwards compatibility: in case someone is using this type directly in an annotation, I think we need a default value for `D` so it can be omitted:

```suggestion
export type StatusRenderer<R, D extends ReadonlyArray<unknown> = ReadonlyArray<unknown>> = {
```

justinfagnani, line 40: We call the args type `T` elsewhere, so I'd either stick with that or give all the type parameters better names like `Result` and `Args`.

packages/task/src/task.ts
```diff
@@ -423,18 +423,26 @@ export class Task<
     return this._error;
   }
 
-  render<T extends StatusRenderer<R>>(renderer: T) {
+  render<S extends StatusRenderer<R, T>>(renderer: S) {
     switch (this.status) {
       case TaskStatus.INITIAL:
-        return renderer.initial?.() as MaybeReturnType<T['initial']>;
+        return renderer.initial?.(this._previousArgs) as MaybeReturnType<
```
justinfagnani, line 429: `_previousArgs` doesn't sound quite right anymore if we're using them for the current render. Maybe it should be named `_latestArgs`?

justinfagnani (changes requested):
Just one small required fix to keep this a backwards compatible change.

justinfagnani:
Thanks @Snapstromegon!

PR #4607, opened 2024-04-01
Fix link to constructible stylesheets explainer

peschee:
The original link seems to point to a 404 page. I've changed the link to point to the explainer markdown in the orginal repo.


packages/reactive-element/src/css-tag.ts
```diff
@@ -161,7 +161,7 @@ export const css = (
 /**
  * Applies the given styles to a `shadowRoot`. When Shadow DOM is
  * available but `adoptedStyleSheets` is not, styles are appended to the
- * `shadowRoot` to [mimic spec behavior](https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets).
+ * `shadowRoot` to [mimic spec behavior](https://github.com/WICG/construct-stylesheets/blob/main/explainer.md#example-usage).
```
justinfagnani, line 164: I think an MDN link is even better, now that it's documented there.

```suggestion
 * `shadowRoot` to [mimic the native feature](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets).
```

justinfagnani (changes requested):
Thanks!

This just needs an empty changeset file. See https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

peschee:
> Thanks!
> 
> This just needs an empty changeset file. See https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

Done

PR #4615, opened 2024-04-10
Add @lit-labs/signals package

justinfagnani:
WIP: Still some rough edges with trying to sync watch() updates and regular updates.


packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+      this.__shouldRender = true;
+      if (!this.__watcherUpdate) {
+        this.__forceUpdateSignal?.set(this.__forceUpdateSignal.get() + 1);
+      }
+      super.requestUpdate(name, oldValue, options);
+    }
+
+    override connectedCallback(): void {
+      if (!this.__isWatching) {
+        this.__isWatching = true;
```
sorvell, line 84: Why is the flag needed? when would this happen when `__isWatching` is true? And I believe you can use `Signal.subtle.hasSources(this.__watcher)` if the state is needed (although an explicit flag in that case might be simpler).

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+    private __watcherUpdate = false;
+    private __watcher = new Signal.subtle.Watcher(() => {
+      this.__watcherUpdate = true;
+      this.requestUpdate();
+      this.__watcherUpdate = false;
+    });
+    private __forceUpdateSignal = new Signal.State(0);
+    private __updateSignal = new Signal.Computed(() => {
+      this.__forceUpdateSignal.get();
+      super.performUpdate();
```
sorvell, line 40: I think you have to call `this.__watcher.watch()` here to make sure to re-activate the watcher. Make sure to avoid passing the signal due to https://github.com/proposal-signals/proposal-signals/issues/157

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    override requestUpdate(
+      name?: PropertyKey | undefined,
+      oldValue?: unknown,
+      options?: PropertyDeclaration<unknown, unknown> | undefined
+    ): void {
+      this.__shouldRender = true;
+      if (!this.__watcherUpdate) {
+        this.__forceUpdateSignal?.set(this.__forceUpdateSignal.get() + 1);
```
sorvell, line 77: Is the idea that this ensures the update occurs even if `performUpdate` didn't get any signals? (e.g. you didn't actually use any signals)? If so, let's comment that.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    protected override update(
+      changedProperties: PropertyValueMap<this> | Map<PropertyKey, unknown>
+    ): void {
+      if (this.__shouldRender) {
+        this.__shouldRender = false;
+        super.update(changedProperties);
+      } else {
+        this.__pendingWatches.forEach((d) => d.commmit());
```
sorvell, line 65: Calling commit here couples this pretty tightly with the directive. This is why I just passed a callback. No strong opinion though.

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,73 @@
...
+export class WatchDirective<T> extends AsyncDirective {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private __host?: SignalWatcher;
+  __signal?: Signal.State<T> | Signal.Computed<T>;
+  private __watcher = new Signal.subtle.Watcher(() => {
+    this.__host?.updateWatch(this as WatchDirective<unknown>);
+  });
+
+  commmit() {
+    this.setValue(Signal.subtle.untrack(() => this.__signal?.get()));
```
sorvell, line 21: Again here, I think you have to call `this.__watcher.watch()` to reset the watcher.

sorvell (changes requested):
For watch, let's consider @e111077 's feedback and maybe have a way to auto-wrap a function (maybe if it's not an event part?).

And I really like using the `html` and not manually watching but if we do that we probably want `unwatched` which is just basically an echoing directive.

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // @ts-expect-error: signal is unused, but the name appears in the signature
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  render(signal: Signal.State<T> | Signal.Computed<T>): T {
+    return undefined as T;
```
e111077, line 27: SSR sad, is there really no good solution here with the current proposal, even with an unwatch?

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,73 @@
...
+  override update(
+    part: Part,
+    [signal]: [signal: Signal.State<T> | Signal.Computed<T>]
+  ) {
+    this.__host ??= part.options?.host as SignalWatcher;
+    if (signal !== this.__signal) {
+      if (this.__signal !== undefined) {
+        this.__watcher.unwatch(this.__signal);
+      }
+      if (signal !== undefined) {
```
e111077, line 39: When is this false? If it's ever false won't line 49 throw?

e111077 (changes requested):

JosefJezek:
@justinfagnani any progress?

packages/labs/signals/README.md
```diff
@@ -0,0 +1,127 @@
+# @lit-labs/preact-signals
+
+Preact Signals integration for Lit.
```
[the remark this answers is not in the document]
justinfagnani: I rewrote the README

packages/labs/signals/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+  render() {
+    return html`
+      <p>The count is ${count.value}</p>
+      <button @click=${this._onClick}>Increment<button></button></button>
+    `;
+  }
+
+  private _onClick() {
+    count.value = count.value + 1;
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/labs/signals/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+  render() {
+    return html`
+      <p>The count is ${watch(count)}</p>
+      <button @click=${this._onClick}>Increment<button></button></button>
+    `;
+  }
+
+  private _onClick() {
+    count.value = count.value + 1;
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/labs/signals/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+  render() {
+    return html`
+      <p>The count is ${count}</p>
+      <button @click=${this._onClick}>Increment<button></button></button>
+    `;
+  }
+
+  private _onClick() {
+    count.value = count.value + 1;
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/labs/signals/src/lib/html-tag.ts
```diff
@@ -0,0 +1,44 @@
...
+ * automatically wrapping Signal instances in the `watch()` directive.
+ */
+export const withWatch =
+  (coreTag: typeof coreHtml | typeof coreSvg) =>
+  (strings: TemplateStringsArray, ...values: unknown[]): TemplateResult => {
+    // TODO (justinfagnani): use an alternative to instanceof when
+    // one is available. See https://github.com/preactjs/signals/issues/402
+    return coreTag(
+      strings,
+      ...values.map((v) => (v instanceof Signal.State || v instanceof Signal.Computed? watch(v) : v))
```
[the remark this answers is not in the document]
justinfagnani: @e111077 I mentioned signal-aware control flow directives as future work in the README.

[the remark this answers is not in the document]
justinfagnani: @sorvell I still don't see those published

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+  abstract class SignalWatcher extends Base {
+    // Watcher.watch() doesn't dedupe so we need to track this ourselves.
+    private __isWatching = false;
+    private __watcherUpdate = false;
+    private __watcher = new Signal.subtle.Watcher(() => {
+      this.__watcherUpdate = true;
+      this.requestUpdate();
+      this.__watcherUpdate = false;
+    });
+    private __forceUpdateSignal = new Signal.State(0);
```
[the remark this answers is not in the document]
justinfagnani: I rewrote a lot of this and added a bunch of comments.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+      this.__forceUpdateSignal.get();
+      super.performUpdate();
+    });
+    private __shouldRender = true;
+    private __pendingWatches = new Set<WatchDirective<unknown>>();
+
+    protected override performUpdate() {
+      // ReactiveElement.performUpdate() also does this check, so we want to
+      // bail early so we don't erroneously appear to not depend on any signals.
+      if (this.isUpdatePending === false) {
```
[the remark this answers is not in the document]
justinfagnani: removed this anyway

packages/labs/signals/src/lib/signal-watcher.ts
justinfagnani: This logic has been simplified. I had to destroy and recreate watchers because of a polyfill bug.

packages/labs/signals/src/lib/signal-watcher.ts
justinfagnani: Yes. Commented and moved this update to where the read is to make this more clear.

packages/labs/signals/src/lib/signal-watcher.ts
justinfagnani: We can change the protocol over time to make this more general.

packages/labs/signals/src/lib/watch.ts
justinfagnani: done

packages/labs/signals/src/lib/watch.ts
justinfagnani: Fixed

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.__isWatching = true;
+        this.__watcher.watch(this.__updateSignal);
+        this.requestUpdate();
+      }
+      super.connectedCallback();
+    }
+
+    override disconnectedCallback(): void {
+      this.__isWatching = false;
+      this.__watcher.unwatch(this.__updateSignal);
```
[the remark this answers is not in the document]
justinfagnani: I'm auto-unwatching when disconnected now, but there are some subtleties still to work out due to memory leaks and a bug that was reported that @sorvell made a fix for.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Adds the ability for a LitElement or other ReactiveElement class to
+ * watch for access to signals during the update lifecycle and trigger a new
+ * update when signals values change.
+ */
+export function SignalWatcher<T extends ReactiveElementConstructor>(
+  Base: T
+): T {
+  abstract class SignalWatcher extends Base {
+    // Watcher.watch() doesn't dedupe so we need to track this ourselves.
```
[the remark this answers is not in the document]
justinfagnani: n/a as this is a bit rewritten

packages/labs/signals/rollup.config.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/signals/rollup.config.js
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {litProdConfig} from '../../../rollup-common.js';
+import {createRequire} from 'module';
+
+export const defaultConfig = (options = {}) =>
+  litProdConfig({
+    packageName: createRequire(import.meta.url)('./package.json').name,
+    entryPoints: ['index', 'lib/signal-watcher', 'lib/watch'],
```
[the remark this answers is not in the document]
justinfagnani: Fixed. But I'm not so sure these should even be separate entrypoints anymore.

packages/labs/signals/src/lib/watch.ts
justinfagnani: rewritten

packages/labs/signals/README.md
```diff
@@ -0,0 +1,127 @@
...
+export class SignalExample extends SignalWatcher(LitElement) {
+  static styles = css`
+    :host {
+      display: block;
+    }
+  `;
+
+  render() {
+    return html`
+      <p>The count is ${count.value}</p>
```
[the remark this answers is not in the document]
justinfagnani: done

justinfagnani:
@JosefJezek just pushed a major update. Hopefully we land this soon.

justinfagnani:
@sorvell and @e111077 thanks for the first pass a while back! PTAL

I rewrote most of SignalWatcher and I think simplified things a lot, and I added a bunch of comments, so hopefully it's easier to follow now.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+      // We may want to use a WeakMap inside the watcher to avoid a strong
+      // reference to the element.
+      queueMicrotask(() => {
+        if (this.isConnected === false) {
+          this.__unwatch();
+        }
+      });
+    }
+
+    updateWatchDirective(d: WatchDirective<unknown>): void {
```
justinfagnani, line 166: I need to add a comment here, and since this for coordination with other code in this library, maybe make it protected and name it with an `_` prefix.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+    protected override update(
+      changedProperties: PropertyValueMap<this> | Map<PropertyKey, unknown>
+    ): void {
+      if (this.__doFullRender) {
+        // Force future updates to not perform full renders by default.
+        this.__doFullRender = false;
+        super.update(changedProperties);
+      } else {
+        // For a partial render, just commit the pending watches.
+        this.__pendingWatches.forEach((d) => d.commmit());
```
justinfagnani, line 124: Note, this performs the commits in directive update order. We probably want to do this is tree order.

justinfagnani:
When this is merged we'll need to create a Lit Labs page entry on lit.dev and a feedback discussion thread similar to https://github.com/lit/lit/discussions/4115

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+This is very exciting for web components, since it means that different web
+components that don't use the same libraries can interoperably watch and produce
+signals.
+
+It also means that many existing state management systems and observability
+libraries that might currently each require their own adapter to integrated with
+the Lit lifecycle, might converge on using standard signals so that we only need
+one Lit adapter, and eventually no adapter at all as support for signals is
+directly added to Lit
```
justinfagnani, line 25: I should probably mention that this is really integration with the signals _polyfill_, and that this package depends on it. Might want to mention that there really needs to be only one copy of the polyfill in an npm installation, otherwise signal tracking breaks.

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+### watch() directive
+
+The `watch()` directive accepts a single Signal and renders its value,
+subscribing to updates and updating the DOM when the signal changes.
+
+The `watch()` directive allows for very targeted updates of the DOM, which can
+be good for performance (but as always, measure!).
+
+Updates from `watch()` directives are batches and commited during the next
```
thepassle, line 123:
```suggestion
Updates from `watch()` directives are batched and commited during the next
```

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+    return html`
+      <p>The count is ${count}</p>
+      <button @click=${this._onClick}>Increment<button></button></button>
+    `;
+  }
+
+  private _onClick() {
+    count.set(count.get() + 1);
+  }
+}
```
thepassle, line 193: the example doesnt include the `withWatch` function?

packages/labs/signals/README.md
rictic: Yeah, I think we need to more clearly delineate "this is what we hope standard signals will bring" vs "this is what you get if you use @lit-labs/signals today"

packages/labs/signals/src/lib/html-tag.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Wraps a lit-html template tag function (`html` or `svg`) to add support for
+ * automatically wrapping Signal instances in the `watch()` directive.
+ */
+export const withWatch =
+  (coreTag: typeof coreHtml | typeof coreSvg) =>
+  (strings: TemplateStringsArray, ...values: unknown[]): TemplateResult => {
+    // TODO (justinfagnani): use an alternative to instanceof when
+    // one is available. See https://github.com/preactjs/signals/issues/402
```
rictic, line 23: Oh, preact signals added a method to do this. Is there anything in standard signals land?

packages/lit-html/src/async-directive.ts
```diff
@@ -362,6 +362,7 @@ export abstract class AsyncDirective extends Directive {
    * @param value The value to set
    */
   setValue(value: unknown) {
+    // this.__part.options?.host
```
rictic, line 365: unintended?

packages/labs/signals/src/lib/html-tag.ts
sorvell: https://github.com/proposal-signals/signal-polyfill/blob/main/src/wrapper.ts#L45

justinfagnani: That's not exported anywhere. There's an umbrella issue for that type of thing here: https://github.com/tc39/proposal-signals/issues/32

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+      super.disconnectedCallback();
+      // Clean up the watcher to avoid memory leaks from signals holding
+      // references to the element. This means that while disconnected, regular
+      // reactive property updates will trigger a re-render, but signal updates
+      // will not. To ensure that current signal usage is still correctly
+      // tracked, we re-enable watching in performUpdate() even while
+      // disconnected. From that point on, a disconnected element will be
+      // retained by the signals it accesses during the update lifecycle.
+      // We may want to use a WeakMap inside the watcher to avoid a strong
+      // reference to the element.
```
justinfagnani, line 158: @sorvell whether or not to disable watching while disconnected and re-enable it when updates trigger while disconnected is a big thing to decide here.

I don't love that if we disable watching that we diverge from how reactive updates work while disconnected. Maybe a WeakRef strategy is how we keep updates enabled while avoid the memory leak.

packages/lit-html/src/async-directive.ts
justinfagnani: oops!

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type WatchDirectiveFunction = <T>(
+  signal: Signal.State<T> | Signal.Computed<T>
+) => DirectiveResult<typeof WatchDirective<T>>;
+
+/**
+ * Renders a signal and subscribes to it, updating the part when the signal
+ * changes.
+ *
+ * watch() can only be used in a reactive element that applies the
```
sorvell, line 89: It would be nice if this could work without reactive element. Considering how to facilitate that:
1. we could have `update` not use `untrack` if there's no host and then have the watcher call `commit` in a queued microtask (to avoid the read exception).
2. OR create a synthetic host that implements the `updateWatchDirective` protocol in a microtask, perhaps keyed via a WeakMap using the `part.options` (if that's always available?).

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,92 @@
...
+      });
+      const watcher = (this.__watcher = new Signal.subtle.Watcher(() => {
+        this.__host?.updateWatchDirective(this as WatchDirective<unknown>);
+        watcher.watch();
+      }));
+      watcher.watch(this.__computed);
+    }
+  }
+
+  private __unwatch() {
```
sorvell, line 36: Should this remove the directive from the host's pending update list?

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: That does seem a lot better because it would ensure that a signal referenced by a child doesn't do unnecessary work that a parent would make superfluous (like DOM removal).

I'm not sure how to do it efficiently, and let's definitely leave a comment if we don't resolve it. Possibly we could leverage the fact that any full render will be in order and push onto an array on the host in that case. Then this ordered array would always be iterated here and only pending watches would be executed.

packages/labs/signals/src/test/signal-watcher_test.ts
```diff
@@ -0,0 +1,230 @@
...
+    count.set(1);
+    await el.updateComplete;
+    assert.equal(
+      el.shadowRoot?.querySelector('p')?.textContent,
+      'count: 2',
+      'B'
+    );
+    assert.equal(readCount, 2);
+
+    // Disconnect the element
```
sorvell, line 131: Add a test like this: https://github.com/lit/lit/pull/4707/files#diff-9e4d56fda3753440f204ca2f7f5fbaad99a1809237a396a74a28e7b80034a7a7R106

sorvell (changes requested):

e111077 (changes requested):

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+    private __pendingWatches = new Set<WatchDirective<unknown>>();
+
+    protected override performUpdate() {
+      // Always enable watching before an update, even if disconnected, so that
+      // we can track signals that are accessed during the update.
+      this.__watch();
+      // Force an uncached read of __performUpdateSignal
+      this.__forcingUpdate = true;
+      this.__forceUpdateSignal?.set(this.__forceUpdateSignal.get() + 1);
+      this.__forcingUpdate = false;
```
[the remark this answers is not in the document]
e111077: That, or supply an equality fn to the signal that is computed from __forcingUpdate

packages/labs/signals/src/lib/watch.ts
e111077: +1 to supporting lit-html + signals outside of ReactiveElement for lit-html standalone users. Could possibly lead to powerful server-powered template renderers

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+type SignalWatcherInterface = SignalWatcher;
+
+/**
+ * Adds the ability for a LitElement or other ReactiveElement class to
+ * watch for access to signals during the update lifecycle and trigger a new
+ * update when signals values change.
+ */
+export function SignalWatcher<T extends ReactiveElementConstructor>(
```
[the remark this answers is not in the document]
justinfagnani: At we can actually address this and not apply the mixin twice by checking the prototype chain

packages/labs/signals/src/lib/watch.ts
sorvell: It can cause an exception if we don't remove the directive. When a part is removed, its `parentNode` can become `null`. This could happen via some parent part, and then when this directive calls `setValue` it has no `parentNode` and this is invalid. 

```js
  html`${this.selected ? html`${watch(this.message}}` : ''}`

File: packages/labs/signals/src/lib/signal-watcher.ts (hunk 21 above)
Comment (sorvell) [reply packages/labs/signals/src/lib/signal-watcher.ts:124]:
```suggestion
        this.__pendingWatches.forEach((d) => d.commit());
```

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: This is definitely pretty tricky. Aside from the memory issue there's also just the problem of indeterminate unbounded work. This typically only happens with shared state, and we're otherwise relatively careful to unhook any sharing in disconnected (e.g. [context](https://github.com/lit/lit/blob/main/packages/context/src/lib/controllers/context-consumer.ts#L87)).

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+      changedProperties: PropertyValueMap<this> | Map<PropertyKey, unknown>
+    ): void {
+      if (this.__doFullRender) {
+        // Force future updates to not perform full renders by default.
+        this.__doFullRender = false;
+        super.update(changedProperties);
+      } else {
+        // For a partial render, just commit the pending watches.
+        this.__pendingWatches.forEach((d) => d.commmit());
+        this.__pendingWatches.clear();
```
justinfagnani, line 125: Oops, we need to clear this outside of the branch. Need to add a test.

justinfagnani: done

packages/labs/signals/src/lib/signal-watcher.ts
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+signals read in `shouldUpdate()`, `willUpdate()`, `update()`, `render()`,
+`updated()`, `firstUpdated()`, and reactive controller's `hostUpdate()` and
+`hostUpdated()`.
+
+This effectively makes the the return result of `render()` a computed signal.
+
+```ts
+import {LitElement, html} from 'lit';
+import {customElement, property} from 'lit';
+import {SignalWatcher, signal} from '@lit-labs/preact-signals';
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+element watch all signal accesses during the element's reactive update
+lifecycle, then triggers an element update when signals change. This includes
+signals read in `shouldUpdate()`, `willUpdate()`, `update()`, `render()`,
+`updated()`, `firstUpdated()`, and reactive controller's `hostUpdate()` and
+`hostUpdated()`.
+
+This effectively makes the the return result of `render()` a computed signal.
+
+```ts
+import {LitElement, html} from 'lit';
```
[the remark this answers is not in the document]
justinfagnani: This example is using the default Lit html tag

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+export class SignalExample extends SignalWatcher(LitElement) {
+  static styles = css`
+    :host {
+      display: block;
+    }
+  `;
+
+  render() {
+    return html`
+      <p>The count is ${count.get()}</p>
```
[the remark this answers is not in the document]
justinfagnani: This is required still here, since it's using the default `html` tag

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+  private _onClick() {
+    count.set(count.get() + 1);
+  }
+}
+```
+
+Elements should not _write_ to signals in these lifecycle methods or they might
+cause an infinite loop.
+
+### watch() directive
```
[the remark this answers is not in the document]
justinfagnani: I don't want to recommend it because it's not going to work with the type-checking tools until they're updated. They're going to complain that State<string> is not assignable to string and such.

packages/labs/signals/README.md
justinfagnani: Yeah, there's no `withWatch()` example here because the `html` tag exported by this package has already applied it. `withWatch()` is kind of advanced.

packages/labs/signals/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export * from 'signal-polyfill';
+export * from './lib/signal-watcher.js';
+export * from './lib/watch.js';
+export * from './lib/html-tag.js';
+
+export const State = Signal.State;
+export const Computed = Signal.Computed;
+
+export const signal = <T>(value: T, options?: Signal.Options<T>) =>
```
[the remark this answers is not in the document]
justinfagnani: This is my thinking:

The proposal's API is pretty verbose compared to framework library signal APIs, which are published in modules and don't use namespace objects. The intention is that frameworks will wrap the API with more convenient ones. We don't have a framework, but we can offer a couple of really small convenience methods so that devs can use a single import and a shorter syntax and automatically use the native feature in the future.

Compare:

```ts
import {Signal} from 'signal-polyfill';
import {SignalWatcher} from '@lit-labs/signals';

const count = new Signal.State(0);
```

to:

```ts
import {SignalWatcher, state} from '@lit-labs/signals';

const count = state(0);
```

This might not matter much to someone also using signals outside of Lit and are already importing the polyfill, but it'll be nice for developers only using signals with Lit, as if it were a feature of the "framework".

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+  Base: T
+): T {
+  abstract class SignalWatcher extends Base implements SignalWatcherInterface {
+    private __watcher?: Signal.subtle.Watcher;
+
+    private __watch() {
+      if (this.__watcher === undefined) {
+        // We create a fresh computed instead of just re-using the existing one
+        // because of https://github.com/proposal-signals/signal-polyfill/issues/27
+        this.__performUpdateSignal = new Signal.Computed(() => {
```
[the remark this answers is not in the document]
justinfagnani: I need to update the bug. In testing here I needed a fresh computed too. For memory management, seems good to release the tracking state of the computed as well though.

packages/labs/signals/README.md
```diff
@@ -0,0 +1,208 @@
...
+export class SignalExample extends SignalWatcher(LitElement) {
+  static styles = css`
+    :host {
+      display: block;
+    }
+  `;
+
+  render() {
+    return html`
+      <p>The count is ${watch(count)}</p>
```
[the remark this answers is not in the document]
justinfagnani: The previous SignalWatcher example should show that you don't need watch()

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+ * update when signals values change.
+ */
+export function SignalWatcher<T extends ReactiveElementConstructor>(
+  Base: T
+): T {
+  abstract class SignalWatcher extends Base implements SignalWatcherInterface {
+    private __watcher?: Signal.subtle.Watcher;
+
+    private __watch() {
+      if (this.__watcher === undefined) {
```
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Set of watch directives that have been updated since the last update.
+     * These will be committed in update() to ensure that the latest value is
+     * rendered and that all updates are batched.
+     */
+    private __pendingWatches = new Set<WatchDirective<unknown>>();
+
+    protected override performUpdate() {
+      // Always enable watching before an update, even if disconnected, so that
```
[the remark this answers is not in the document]
justinfagnani: I removed this thinking that we now always force a read, but thinking about it I guess we could miss a signal update. I'll add it back, but I we need to devise a test scenario here.

packages/labs/signals/src/lib/signal-watcher.ts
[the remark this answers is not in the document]
justinfagnani: An equality function, even of `() => false` won't cause the computed to re-evaluate the compute function. We'd need something like: https://github.com/tc39/proposal-signals/issues/151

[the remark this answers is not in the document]
justinfagnani: @sorvell I used to use a counter for this, but I found the named boolean flag to be easier to understand myself.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+      name?: PropertyKey | undefined,
+      oldValue?: unknown,
+      options?: PropertyDeclaration<unknown, unknown> | undefined
+    ): void {
+      this.__doFullRender = true;
+      super.requestUpdate(name, oldValue, options);
+    }
+
+    override connectedCallback(): void {
+      this.__watch();
```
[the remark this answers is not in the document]
justinfagnani: ah, no, because we call requestUpdate(); Removed.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -0,0 +1,176 @@
...
+      // Clean up the watcher to avoid memory leaks from signals holding
+      // references to the element. This means that while disconnected, regular
+      // reactive property updates will trigger a re-render, but signal updates
+      // will not. To ensure that current signal usage is still correctly
+      // tracked, we re-enable watching in performUpdate() even while
+      // disconnected. From that point on, a disconnected element will be
+      // retained by the signals it accesses during the update lifecycle.
+      // We may want to use a WeakMap inside the watcher to avoid a strong
+      // reference to the element.
+      queueMicrotask(() => {
```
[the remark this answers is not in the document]
justinfagnani: The microtask is there to debounce moves that don't just directly append the element to another location. I believe this can happen in repeat()-like scenarios where code sees that an item is not present at the original location, removes and caches it, then re-adds it later at a new location. It'll also be true if we ever made that shared repeat() cache we've talked about for being able to move items between lists.

packages/labs/signals/src/index.ts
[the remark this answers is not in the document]
sorvell: Why not this? `export * from 'signal-polyfill';`

Just `new Signal.State(0)` v. `state(0)` alone doesn't seem worth obfuscating the connection between this package the the proposal.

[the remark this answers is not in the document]
justinfagnani: > Why not this? `export * from 'signal-polyfill'`;

Because that exports a namespace object, which is not that idiomatic for modules.

You still have this pattern:

```ts
import {SignalWatcher, Signal} from '@lit-labs/signals';

const count = new Signal.State(0);
```

That might not seem so bad with one signal, but with the `watch()` pattern you'll possibly need to create a lot of computed signals to get targeted updates for derived data:

```ts
import {SignalWatcher, state, computed} from '@lit-labs/signals';

export class MyElement extends SignalWatcher(LitElement) {
  #count = state(0);

  #countPlusOne = computed(() => this.#count.get() + 1);

  render() {
    return html`<p>One-indexed count: ${watch(this.#couplePlusOne)}`;
  }
}
```

Here I think `state()` and `computed()` start pulling their weight, given how cheap they are to vend. Again, the standard API is purposefully not made to be ergonomic because they assume frameworks will wrap it, and because the API hangs off a global instead of being exported from a module.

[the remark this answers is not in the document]
sorvell: > Again, the standard API is purposefully not made to be ergonomic because they assume frameworks will wrap it

This just doesn't make sense to me as anyone reading any documentation has to carefully connect the dots.

However, I think a better reason to wrap is that the proposal is in flux and we don't users to have to change their code if it changes.

packages/labs/signals/src/lib/signal-watcher.ts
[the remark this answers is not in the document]
sorvell: OK, let's then make sure to comment that the purpose of `__forcingUpdate` is to ensure a non-signals provoked update gets performed.

packages/labs/signals/src/lib/watch.ts
sorvell: Via chat, we resolved not to block on adding this but instead to add a comment stating this will be added in the future based on feedback.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -67,24 +77,32 @@ export function SignalWatcher<T extends ReactiveElementConstructor>(
...
-        // or a module global to prevent this closure from keeping the enclosing
-        // scopes alive.
-        const el = elementForWatcher.get(this) as
-          | SignalWatcherInternal
-          | undefined;
-        if (el?.__forcingUpdate === false) {
+        // (the watcher) or a module global to prevent this closure from keeping
+        // the enclosing scopes alive, which would keep the element alive. So
+        // The only two references are `this` and `elementForWatcher`.
+        const el = elementForWatcher.get(this);
```
e111077, line 83: Would we ever be in the situation where multiple versions of this library might cause a cache miss here? Do we care about that case?

justinfagnani: Great question, but I think the answer is no. We're never dealing in other elements here, only the `this` element, so the cache and element code are always in the same module.

justinfagnani:
@sorvell @e111077 I implemented the manual memory management code in the last two commits, if you can review that. 

I'll review the last round of comments ASAP, if there are any left unresolved.

e111077 (commented):
The memory approach is clever. Wonder if we might run into issues down the line with watchers running unnecessarily because of FinalizationRegistry timing. Though only things I could think of that might run into that are users writing brittle code that is probably in need of a rewrite anyway

justinfagnani:
All the current test failures are from virtualizer. Re-running tests hasn't made them pass, so I'm not sure if they're flaky.

justinfagnani:
I tried to update test dependencies in https://github.com/lit/lit/pull/4758 to see if that would help the virtualizer tests. That didn't work locally at least, so I'll keep re-running the test job for now.

packages/labs/signals/README.md
justinfagnani: Added `withWatch()` example.

packages/labs/signals/src/lib/signal-watcher.ts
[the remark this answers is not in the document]
justinfagnani: Added more comments.

packages/labs/signals/src/lib/signal-watcher.ts
justinfagnani: Added comment, renamed with a `_`, and added an `@internal` annotation. I think this method is particularly subject to change. We probably might to add a RenderOption to lit-html for signal update scheduling so that `watch()` could work outside of LitElement.

packages/labs/signals/src/lib/watch.ts
```diff
@@ -0,0 +1,92 @@
...
+  private __signal?: Signal.State<T> | Signal.Computed<T>;
+
+  private __watcher?: Signal.subtle.Watcher;
+
+  // We have to wrap the signal in a computed to work around a bug in the
+  // signal-polyfill: https://github.com/proposal-signals/signal-polyfill/issues/27
+  private __computed?: Signal.Computed<T | undefined>;
+
+  private __watch() {
+    if (this.__watcher === undefined) {
```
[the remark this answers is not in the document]
justinfagnani: done

packages/labs/signals/src/lib/watch.ts
justinfagnani: done

packages/labs/signals/src/lib/watch.ts
justinfagnani: I added a comment to `_updateWatchDirective()` suggesting how we might move forward on this.

justinfagnani (commented):
PTAL!

packages/labs/signals/src/test/signal-watcher_test.ts
justinfagnani: Just a note to myself here that I didn't add this test yet.

packages/labs/signals/src/lib/signal-watcher.ts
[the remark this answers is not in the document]
justinfagnani: Added a comment to https://github.com/proposal-signals/signal-polyfill/issues/27

justinfagnani:
@graynorton I had to disable the virtualizer tests here because they wouldn't pass even on multiple test re-runs. This PR doesn't change any core or virtualizer code, nor does it update any relevant dependencies int he package lock. I'm not sure how to move forward there. Any advice would be helpful!

packages/labs/signals/README.md
```diff
@@ -146,19 +180,36 @@ export class SignalExample extends SignalWatcher(LitElement) {
...
+then those updates are commited directly _without_ a full template render. If
+any other changes triggered the reactive update, then the whole template is
+re-rendered, along with the latest signal values.
+
+This approach preserves both DOM coherence and targetted updates, and coalesces
+updates when both signals and reactive properties change.
+
+`watch()` must be used in conjunction with the `SignalWatcher` mixin.
+
+You can mix and match targetted updates with `watch()` directive and
```
e111077, line 208:
```suggestion
You can mix and match targeted updates with `watch()` directive and
```

packages/labs/signals/README.md
```diff
@@ -183,19 +234,36 @@ export class SignalExample extends SignalWatcher(LitElement) {
...
 }
 ```
 
+#### `withWatch()`
+
 `withWatch()` is a function that wraps an `html` tag function with the
-auto-watching functionality. This allows you to compose this wrapper with other
-html-tag wrappers like Lit's `withStatic()` static template wrapper.
+auto-watching functionality. The `html` tag exported by `@lit-labs/signals` is
+simply the core lit-html template tag wrapped with `withWatch()` as a
```
e111077, line 251: Arthur once told me to try to not use words like “simply” or “just”. Maybe something like:

> The `html` tag exported by `@lit-labs/signals` is a convenient export of the core lit-html template tag wrapped with `withWatch()`.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -225,22 +234,45 @@ export function SignalWatcher<T extends ReactiveElementConstructor>(
       // the update lifecycle.
       //
       // We use queueMicrotask() to ensure that this cleanup does not happens
```
e111077, line 236:
```suggestion
      // We use queueMicrotask() to ensure that this cleanup does not happen
```

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -225,22 +234,45 @@ export function SignalWatcher<T extends ReactiveElementConstructor>(
...
-      // because of synchronous moves in the DOM.
+      // because moves in the DOM within the same task, such as removing an
```
e111077, line 237:
```suggestion
      // because of moves in the DOM within the same task, such as removing an
```

e111077 (approved):
All my comments at this point are just doc nits. Code looks good to me

sorvell (approved):

PR #4616, opened 2024-04-11
[labs/cli] Fix npm install failing in Winows

augustjk:
The latest Node security fix broke the windows tests (https://github.com/lit/lit/actions/runs/8639152432/job/23685369172) as the workaround for spawning npm/npx by adding .cmd for Windows is considered insecure: https://github.com/nodejs/node/commit/9095c914ed 

I removed the workaround and opted to go through the shell for the two places we use these. It's probably a bit less performant now for having to go through the shell but it's just for testing.


rictic (approved):

augustjk:
Added a changeset for `labs/cli` since this actually does change runtime behavior for npm installing.

PR #4637, opened 2024-05-02
[lit-html] Add MathML support with the math template tag

### LINKED ISSUE #4636: [lit-html] MathML suppport
### Should this be an RFC?

### Description

MathML has very good browser support now: https://developer.mozilla.org/en-US/docs/Web/MathML#browser_compatibility

We should support MathML fragments with a new `math` template tag like we do with SVG and the `svg` tag.

None. Without parsing fragments inside a `<math>` tag, elements will not be parsed into the correct namespace.

justinfagnani:
Fixes #4636

This adds a `mathml` template tag to enable writing MathML fragment templates, just like how we have the `svg` template tag for SVG fragment templates.

The change adds another TemplateResult type for MATHML, and just creates a `<math>` wrapper element for fragment templates like we do for SVG.

I had to update the `@open-wc/testing` dependency of the starter kits because 3.x had hard-coded the template result types. 4.x works with the new type. So there is the possibility that that this will break type-checking of other libraries that hard-code the types. Hopefully they can upgrade quickly if they need to. Our policy is to not consider type-only breaks as semver major, so this change is semver minor.

cc @lukewarlow @bkardell 😄


packages/lit-html/src/lit-html.ts
```diff
@@ -607,10 +610,37 @@ export const html = tag(HTML_RESULT);
...
+ *
+ * const eq = html`
+ *   <math>
+ *     ${num}
+ *   </math>`;
+ * ```
+ *
+ * The `math` *tag function* should only be used for MathML fragments, or
+ * elements that would be contained **inside** an `<math>` HTML element. A
+ * common error is placing an `<math>` *element* in a template tagged with the
```
lukewarlow, line 633: Super nit: an `<math>` -> a `<math>`

packages/lit-html/src/lit-html.ts
```diff
@@ -929,9 +962,9 @@ class Template {
     walker.currentNode = this.el.content;
 
     // Re-parent SVG nodes into template root
```
augustjk, line 964:
```suggestion
    // Re-parent SVG or MathML nodes into template root
```

.changeset/beige-lamps-camp.md
```diff
@@ -0,0 +1,7 @@
+---
+'lit-html': minor
+'lit': minor
+'lit-element': minor
```
augustjk, line 4: Nothing here touches the `lit-element` package so I don't think it needs a bump there.
```suggestion
'lit-html': minor
'lit': minor
```

augustjk (commented):
Non-blocking, but maybe some cursory SSR tests would be nice?
e.g. https://github.com/lit/lit/blob/56cb2ca4a86f2c7f6af757c612f05e50d1966615/packages/labs/ssr/src/test/integration/tests/basic.ts#L303
https://github.com/lit/lit/blob/56cb2ca4a86f2c7f6af757c612f05e50d1966615/packages/labs/ssr/src/test/lib/render-lit_test.ts#L392

Also, should there be an `unsafeMath()` directive, like `unsafeSVG()`?

.changeset/beige-lamps-camp.md
justinfagnani: I thought we were bumping `lit` and `lit-element` along with `lit-html` so that if you wanted to ensure you got a new feature but were only depending on `lit` or `lit-element` you could bump your own dependency version.

augustjk: Hmm.. I thought it was more so about making sure `lit` gets bumped, but yeah, I guess `lit-element` does have a dependency on `lit-html` so it would be a good signal for users that just depend on the `lit-element` package. Nevermind!

justinfagnani:
@augustjk Thanks! I added `unsafeMath()` and SSR tests

justinfagnani:
There's a small question of the name of the template tag. I chose `math` because I think it's a lot easier to read and write then `mathml`.

`html`, `svg`, and `css` are more clear about what their content are than just `math`, and that might help with built-in language support in tools. For instance, `mathml` might conceivable trigger XML or even MathML specific highlighting. `math` might not - though we can highlight it in our tools.

augustjk (approved):
Thanks!

Regarding the tag name, I have a preference for `mathml` as the full name of the markup language, but not strong enough to block. Maybe an online poll on discord or twitter?

If we do change, I suppose the unsafe directive name should change to `unsafeMathML` too.

justinfagnani:
> Regarding the tag name, I have a preference for `mathml`

Let's ask Discord before merging.

It's not unchangeable at least. Just a couple of byte cost to export two names.

lukewarlow:
Math being the root element name for the MathML namespace makes me lean towards math as the tagged template name too fwiw.

justinfagnani:
> Math being the root element name for the MathML namespace makes me lean towards math as the tagged template name too fwiw.

That's a good point. `<svg>` -> `svg`, `<math>` -> `math`

justinfagnani:
Discord seems to be leaning heavily towards `mathml`

scripts/check-size.js
```diff
@@ -9,8 +9,8 @@ import * as fs from 'fs';
 // it's likely that we'll ask you to investigate ways to reduce the size.
 //
 // In either case, update the size here and push a new commit to your PR.
-const expectedLitCoreSize = 15443;
-const expectedLitHtmlSize = 7258;
+const expectedLitCoreSize = 15503;
```
justinfagnani, line 12:
```suggestion
const expectedLitCoreSize = 15505;
```

scripts/check-size.js
```diff
@@ -9,8 +9,8 @@ import * as fs from 'fs';
...
+const expectedLitHtmlSize = 7313;
```
justinfagnani, line 13:
```suggestion
const expectedLitHtmlSize = 7315;
```

augustjk (approved):
Change from `math` to `mathml` LGTM

augustjk:
@justinfagnani Do we also need a static version of `mathml` tag function?

justinfagnani:
> @justinfagnani Do we also need a static version of `mathml` tag function?

done!

PR #4646, opened 2024-05-12
Ensure ref value is always undefined when disconnected

### LINKED ISSUE #4622: ref should not provide element when disconnected

### Description

The `ref` directive supplies an element when it runs update and it is in a disconnected state. This does not make sense since `ref` supplies `undefined` when disconnected and the element again when reconnected. Since this behavior implies that `ref` is intended to supply `undefined` when disconnected, it's incorrect for this to change while it is disconnected.

This can cause practical problems if code depends on installing/removing behavior based on connection state. See [discord discussion](https://discord.com/channels/1012791295170859069/1228620722859937792).

Note also that checking `refElement.isConnected` is *not* a workaround since directives are run in a connected state regardless of whether or not the host context is connected.

### Reproduction

In the repro example here https://lit.dev/playground/#gist=fe6d7deeba80b76873eb057819ed2bc4
1. open the console
2. click the remove button
3. click the increment button
4. click the add button

Note that:
A. After (3) the ref callback is supplied when an element. This is unexpected.
B. After (4) the ref callback gets undefined and then the element. This is unexpected.

### Workaround

Gate behavior on the host's connection state.

### Is this a regression?

### Affected versions

3.x

All

sorvell:
Fixes #4622

### DIFF
--- a/.changeset/violet-phones-look.md
+++ b/.changeset/violet-phones-look.md
@@ -4,4 +4,4 @@
 'lit': patch
 ---
 
-ref value always undefined when disconnected
+The value provided by the `ref()` directive will always be `undefined` when the element is disconnected.
--- a/.changeset/violet-phones-look.md
+++ b/.changeset/violet-phones-look.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+ref value always undefined when disconnected
--- a/.changeset/violet-phones-look.md
+++ b/.changeset/violet-phones-look.md
@@ -1,5 +1,7 @@
 ---
+'lit-element': patch
 'lit-html': patch
+'lit': patch
 ---
 
 ref value always undefined when disconnected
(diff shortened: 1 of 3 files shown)


justinfagnani (approved):

.changeset/violet-phones-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
```
augustjk, line 3:
```suggestion
---
'lit-element': patch
'lit-html': patch
'lit': patch
---
```

.changeset/violet-phones-look.md
```diff
@@ -0,0 +1,5 @@
...
+
+ref value always undefined when disconnected
```
augustjk, line 5:
```suggestion
The value provided by the `ref()` directive will always be `undefined` when the element is disconnected.
```

augustjk (commented):
Minor suggestions on the changeset

augustjk (approved):

PR #4648, opened 2024-05-16
Enable Babel decorator tests and add esbuild tests

justinfagnani:
The Babel tests were disabled because Babel hadn't implemented metadata and the had some [bugs that we triggered during tests](https://github.com/babel/babel/issues/16292). This PR updates Babel and enables the tests.

The esbuild tests are new, since esbuild [just shipped decorator are metadata support](https://github.com/evanw/esbuild/issues/104).

We don't strictly need the tests through multiple compilers if the compilers are spec compliant, but the compile author have wanted to see real-world usage work since decorators are a complicated feature, and this is a good way to catch regressions. When decorators are shipping natively, we could reduce our test load.


augustjk (approved):
Nice!

PR #4679, opened 2024-06-22
[localize] Move setup code to own file to resolve import loop

### LINKED ISSUE #4655: Import loop in @lit/localize

Localize (@lit/localize)

### Description

Hi there 👋,

Thanks for all the wonderful work! I noticed a little something:

`lit-localize.js` and `init/*.js` have an import loop.

There is already a comment to drop the lines:
https://github.com/lit/lit/blob/abf30b3e895ea5d833f6d9559612e2b1ba47580d/packages/localize/src/lit-localize.ts#L15C1-L22C37

However, the comment is only regarding the bundle size, not the import loop.

### Reproduction

I noticed this while using a post-processors, that traverses imports. It's much too complicated to set up, but you can simply check the imports manual and verify the issue:
https://github.com/search?q=repo%3Alit%2Flit+import+%7B_installMsgImplementation%7D+from+%27..%2Flit-localize.js%27%3B&type=code

### Workaround

I used esbuild to bundled it. Most bundlers will ignore the loop.

### Is this a regression?

### Affected versions

latest

OS independent issue.

### Potential solutions

1. Remove the code, that's marked for deprecation (clean, but risky).
2. Move the `_installMsgImplementation` into a separate file (safe).

codingjoe:
Localize has an import loop because submodules are importing from their root. To resolve the loop, `msg`, `installed` and `_installMsgImplementation` are moved to a new submodule.

Imports to `msg` remain intact, through an explicit export.

### DIFF
--- a/.changeset/chilly-melons-laugh.md
+++ b/.changeset/chilly-melons-laugh.md
@@ -1,5 +1,5 @@
 ---
-"@lit/localize": patch
+'@lit/localize': patch
 ---
 
 Move localize setup to resolve import loop
--- a/.changeset/chilly-melons-laugh.md
+++ b/.changeset/chilly-melons-laugh.md
@@ -1,0 +1,5 @@
+---
+"@lit/localize": patch
+---
+
+Move localize setup to resolve import loop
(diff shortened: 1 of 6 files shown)


augustjk (approved):
This is great, thank you!

PR #4681, opened 2024-06-23
[all] Update Rollup packages

justinfagnani:
Updating Rollup and related packages, initially to try to get rid of this error message we have on npm install:

<details>
  <summary>npm install error</summary>

```
➜  lit git:(main) ✗ npm ci
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @rollup/pluginutils@3.1.0
npm warn Found: rollup@4.18.0
npm warn node_modules/rollup
npm warn   dev rollup@"^4.18.0" from the root project
npm warn   11 more (@rollup/plugin-commonjs, @rollup/plugin-inject, ...)
npm warn
npm warn Could not resolve dependency:
npm warn peer rollup@"^1.20.0||^2.0.0" from @rollup/pluginutils@3.1.0
npm warn node_modules/rollup-plugin-sourcemaps/node_modules/@rollup/pluginutils
npm warn   @rollup/pluginutils@"^3.0.9" from rollup-plugin-sourcemaps@0.6.3
npm warn   node_modules/rollup-plugin-sourcemaps
npm warn
npm warn Conflicting peer dependency: rollup@2.79.1
npm warn node_modules/rollup
npm warn   peer rollup@"^1.20.0||^2.0.0" from @rollup/pluginutils@3.1.0
npm warn   node_modules/rollup-plugin-sourcemaps/node_modules/@rollup/pluginutils
npm warn     @rollup/pluginutils@"^3.0.9" from rollup-plugin-sourcemaps@0.6.3
npm warn     node_modules/rollup-plugin-sourcemaps
npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
```

</details>

It turns out that rollup-plugin-sourcemaps is seemingly unmaintained right now and has [this issue open for Rollup 3 and 4 compatibility](https://github.com/maxdavidson/rollup-plugin-sourcemaps/issues/127). A user reported that an npm override for `@rollup/pluginutils` fixed the error though, so I applied that as well.


justinfagnani:
Interesting - this adds 231 to lit-core.min.js. I think we'll still want to update dependencies, but understand what's going on here. Maybe `@rollup/plugin-terser` has some different defaults than `rollup-plugin-terser` did?

augustjk:
Manual inspection of the bundle shows the difference in bytes are due to some variable names being minified to 2 chars instead of 1. It seems to be variation from the change in the version of the `terser` packaged used by the rollup plugin. I updated the underlying `terser` package to latest as well which reduced the size gain a bit to 217 bytes instead of 231.

The failing tests were due to the `@rollup/plugin-place` not actually replacing the `DEV_MODE` flag for the bundled `polyfill-support.js`. Setting `preventAssignment: false` seems to make it work again. Not entirely sure why, maybe related to it being declared with `var`, but I didn't want to dig too further.

I also updated the rollup and related packages for the localize examples and starter kits.

.changeset/tender-penguins-admire.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
justinfagnani, line 2: @augustjk I suppose we should include the starter kits here, since updating dev dependencies is more relevant than for our own packages.

I so wonder if we should mention the rollup and terser updates for our packages as well though, since conceivably they could effect people if something goes wrong. wdyt?

packages/lit-html/package.json
```diff
@@ -460,9 +460,7 @@
         "tsconfig.compiled-tests.json"
       ],
       "output": [
-        "compiled/**/*.{js,js.map,d.ts,d.ts.map}",
-        "tsconfig.tsbuildinfo",
```
justinfagnani, line 464: How did you notice this?

justinfagnani (commented):
Your changes LGTM @augustjk. Thanks!

packages/lit-html/package.json
augustjk: I saw that on repeat builds wireit was detecting output files modified. I checked the outputs and found the compile build doesn't generate the tsbuildinfo files.

augustjk (approved):

PR #4682, opened 2024-06-23
[all] Update typescript to 5.5.0

### LINKED ISSUE #4686: [labs/gen-vue-wrapper] Stop renaming output files
### Should this be an RFC?

### Description

We currently generate a script that renames the files output by the vue-tsc compiler to remove the `.vue` part of the extension. This doesn't appear to be idiomatic. Packages that published compiled Vue components to npm use the default output filenames.

For example, `@mcler/v-calendar`, published components like Calendar.vue.js: https://unpkg.com/browse/@mcler/v-calendar@1.0.4/dist/es/components/Calendar/

One motivation for renaming the script is that the output files changed with a Vue update, and the rename script doesn't work well enough for the tests. Removing the script simplifies things and is more resilient to future changes.

You can remove the rename script.

justinfagnani:
Based on #4681

This updates TypeScript to 5.5.

The ElementInternals interface was updated in typescript's lib.dom.d.ts, which required a couple of changes to our SSR DOM shim and revealed a typo in our interface properties.

There's currently a build error due to dependencies of lit-html not finding the definition for TrustedHTML that's in lit-html's public interface. I'm not sure yet if there's a new check in TypeScript, or if a change in the .d.ts emit has caused a type import or reference to `@types/trusted-types` to be omitted. Adding `skipLibCheck` to dependents of lit-html fixes the problem, but we don't want to require that our dependents do that.

Closes #4686


justinfagnani:
Looks like `lit-html.d.ts` is now missing this line:

```ts
/// <reference types="trusted-types" />
```

That was automatically put there by the typescript compiler, so I'm not sure if we have to manually add it, or if there's a compiler setting that needs to be enabled now.

justinfagnani:
Interesting that Rollup update, plus this TypeScript update, add 140 bytes to lit-core.min.js in total, meaning that this one reduces the bundle size by 91 bytes.

rictic:
[Looks like](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e7a2bddf627c5e3c36e3769a66c412dbf67fae4d/types/trusted-types/lib/index.d.ts#L3C47-L3C64) we can now import the trusted types… types with something like:

```ts
import type {...} from '@types/trusted-types/lib';
```

(or maybe it's from `'trusted-types/lib'`?)

rictic:
> this one reduces the bundle size by 91 bytes.

That's pretty surprising!

justinfagnani:
> > this one reduces the bundle size by 91 bytes.
> 
> That's pretty surprising!

I know! Maybe there's a better emit for nullish coalescing or optional chaining?

justinfagnani:
> > > this one reduces the bundle size by 91 bytes.
> > 
> > 
> > That's pretty surprising!
> 
> I know! Maybe there's a better emit for nullish coalescing or optional chaining?

Hmm. Somehow in a rebase against `update-rollup` the 91 bytes came back. Looking into it.

justinfagnani:
Oh, looking back I think this PR had accidentally reverted some of the Rollup updates.

justinfagnani:
In the process of updating TypeScript, and therefore some other dependencies in our wrapper generators, I hit a snag that makes me think we need to stop renaming vue-tsc output files: https://github.com/lit/lit/issues/4686

It'd be ideal for that to be a separate PR, but I can just add it here to get the update done.

rictic (approved):

Githubant123:
hello

Issue #3517, opened 2022-12-13
[labs/router] Hash routing doesn't work


FluorescentHallucinogen:
### Which package(s) are affected?

`@lit-labs/router`

### Description

The `pathname` for `URLPattern` works in `@lit-labs/router`, but the `hash` for some reason doesn't. See the reproduction below.

This is not a bug in the native browser implementation of `URLPattern` or in the `urlpattern-polyfill`:

```js
new URLPattern({ hash: "one" }).test("http://localhost:8080")
// returns `false`

new URLPattern({ hash: "one" }).test("http://localhost:8080/#one")
// returns `true`

new URLPattern({ hash: "one" }).test("http://localhost:8080#one")
// returns `true`

new URLPattern({ hash: "one" }).test("http://localhost:8080/something/#one")
// returns `true`
```

In the `@lit-labs/router` source code I've found this: https://github.com/lit/lit/blob/main/packages/labs/router/src/routes.ts#L32.

Does this mean that nothing except `pathname` and `search` including `hash` is implemented?

### Reproduction

[...]

PR #4685, opened 2024-06-26
[labs/router] Extend lit-router to support hash-based routing

maischus:
lit-router currently only uses the `pathname` to perform the routing. This change extends the routing to `search` and `hash` part of an URL.

This should fix lit/lit#3517 and allow `urlpattern-polyfill` to work correctly (kenchris/urlpattern-polyfill#129) for hash-based routing.

@justinfagnani What are your thoughts on this approach? `tailGroup` has not been adapted yet. I would like to wait for your thoughts before continuing.


xseman:
Hey, I'd like to help with this. If the author isn't active, I'd be happy to continue implement the feature

PR #4687, opened 2024-06-27
[labs/virtualizer] Fix a null pointer exception happening in corner cases when virtualizer is destroyed and recreated quikcly.

alanbuzek:
This solves an error we've been facing internally:

When virtualizer component is destroyed and then recreated quickly, there appears to be a race condition where the children are sometimes null (such as when we change 'tabs' within our lit application, where in each tab there is a lit-virtualizer component). In those cases, we get the following error: `TypeError: Cannot read properties of null (reading 'observe')` with the stack trace pointing the modified line.

Given that from a type-safety perspective, `this._childrenRO` can be null in those cases, it makes sense to turn to add the optional binding.


justinfagnani:
@alanbuzek can you describe the error you were getting?

Ideally, we'd have an issue open first, and a test in the PR. I know that virtualizer is a bit hard to test, but I'd at least like to have an understanding of why this element can be undefined, and what the consequences of that are. Is this something that fixes itself after the next update?

We at least need a changeset in the PR.

cc @graynorton

alanbuzek:
Hi @justinfagnani, thanks for your reply. I added more detailed explanation on the error it's solving, please let me know if it makes sense. Also reached out via email with more (internal) context and added a context. Please let me know what you think, thanks!

graynorton:
@alanbuzek Thank you for the PR!

In reviewing your change, it appears to me that the error you're seeing is happening when a Virtualizer's mutation observer fires after the Virtualizer has been disconnected, resulting in a call to `_finishDOMUpdate()` after `_childrenRO` has been reset to `null`.

Your change will prevent the specific exception you're seeing, but I think we'd actually prefer to have `_finishDOMUpdate()` do nothing at all when the Virtualizer is not connected.

Since we have another spot in the code (in `_updateLayout()`) where we are guarding based on an indirect check of connection state, it seems better for clarity to track connection state explicitly and use that state to guard both `_finishDOMUpdate()` and `_updateLayout()`.

I've pushed a commit to your branch that does this, and have confirmed that all of our existing Virtualizer tests are passing with this change. Can you take a look and verify that this change also eliminates the errors you're seeing in your app?

If so, then I'll go ahead and approve the PR and we can get it merged.

cc @rictic

alanbuzek:
@graynorton Thanks so much for looking into this in detail and coming up with a better solution. We tested your proposed change and it indeed solves the issues we've been facing. You can approve the PR and merge, as suggested. Thanks!

graynorton (approved):
Approved—thanks again, @alanbuzek!

Issue #4377, opened 2023-11-08
[labs/virtualizer] Should `renderItem()` allow returning non `TemplateResult` values?


augustjk:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Virtualizer (@lit-labs/virtualizer)

### Description

I noticed that `RenderItemFunction` type only allows `TemplateResult` as a return value.
https://github.com/lit/lit/blob/462350126d51d56f9211d50db43762d81a8b8ed1/packages/labs/virtualizer/src/virtualize.ts#L40-L43

This function is basically passed into the `repeat()` directive which allows `unknown` as a return value, which is technically correct as `lit-html` can handle rendering values besides template results.

### Alternatives and Workarounds

Wrap the return with `html` tag function, even if not needed, to satisfy types.

PR #4691, opened 2024-06-29
[labs/virtualizer] improve type and supporting documentation

Garbee:
First, this fixes the `RenderItemFunction` type so it can return any stringable value. Just as lit expects internally for rendering.

Second, the documentation in the readme is updated to reflect the current type definition. As well as adding extra guidance on using the generic to satisfy more strict TypeScript configurations.

Fixes: #4377


Garbee:
I set the version bump to a `patch` for this. Since it doesn't *restrict* the return result further. We're relaxing it to accept more values. Nothing will break and nothing new is being added.

PR #4720, opened 2024-08-01
[chore] Update expected sizes

augustjk:
I think the size updates between https://github.com/lit/lit/pull/4637 and https://github.com/lit/lit/pull/4681 may have happened separately leaving the latest merge to not be up to date.

We could force PRs to always be up to date with main before merging to prevent this from happening in the future, but that seems too restrictive when size changes to core packages should be rare. I think we had that enabled when we were preparing for 3.0 release and it caused a lot of unneeded churn keeping PR branches up to date.


rictic (approved):
Always happy to approve a size reduction

PR #4725, opened 2024-08-06
Ensures initial change old value is always `undefined` when using `@property`

### LINKED ISSUE #4721: Initial changed value incorrect when using legacy decorators and a wrapped property

### Description

The legacy `@property` adds [wrapped](https://github.com/lit/lit/blob/main/packages/reactive-element/src/decorators/property.ts#L75) to the property options when an accessor already exists on the element. This happens (1) when the user has implemented a getter/setter or (2) when `accessor` is used to define the property.

When `ReactiveElement` processes initial changes, any properties with this flag set are automatically added to the `changedProperties` map. However, the *current* value [is added to the map](https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L1429). This diverges from the behavior of all other cases when `undefined` is provided as the initial value.

Simply changing this to pass `undefined` to `this._$changeProperty` should normalize the behavior.

### Reproduction

[Playground repro](https://lit.dev/playground/#gist=5db77883d71d6eab2783e691820d6e20), see console output.

### Workaround

Nope

### Is this a regression?

### Affected versions

3.x

All

sorvell:
Fixes #4721.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1421,12 +1421,11 @@ export abstract class ReactiveElement
         .elementProperties;
       if (elementProperties.size > 0) {
         for (const [p, options] of elementProperties) {
-          if (
-            options.wrapped === true &&
-            !this._$changedProperties.has(p) &&
-            this[p as keyof this] !== undefined
-          ) {
-            this._$changeProperty(p, this[p as keyof this], options);
+          if (options.wrapped === true && this[p as keyof this] !== undefined) {
+            // Note, deleting an existing value here ensures
+            // the initial old value is seeded as `undefined`
+            this._$changedProperties.delete(p);
+            this._$changeProperty(p, undefined, options);
           }
         }
       }
--- a/packages/reactive-element/src/test/decorators/property_test.ts
+++ b/packages/reactive-element/src/test/decorators/property_test.ts
@@ -488,6 +488,53 @@ suite('@property', () => {
     assert.deepEqual(el._observedZot2, {value: 'zot', oldValue: ''});
   });
 
+  test('reports initial changes to properties', async () => {
+    let changedProperties: PropertyValues | undefined;
+
+    class E extends ReactiveElement {
+      @property() unValued?: string;
+      @property() valued = 'valued';
+
+      _valuedAccessor = 'valuedAccessor';
+
+      @property()
+      accessor valuedAutoAccessor = 'valuedAutoAccessor';
+
+      @property()
+      set valuedAccessor(v) {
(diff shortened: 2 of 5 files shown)


sorvell:
Note, it does not appear that we are running `std-decorator` tests for TS, but we are for babel. I cannot find this output folder in my build (once I clear the cache) but see that this output folder ie [expected](https://github.com/lit/lit/blob/main/packages/reactive-element/package.json#L291)

sorvell:
There has been discussion of adding a feature to configure if a default value is seen as an initial change. With this PR normalizing the behavior of initial changes, such a feature is easier to add.

This is possible _**except**_ in cases when (1) using legacy decorators + emulated class fields or (2) JS based static properties and setting an initial value in the constructor.

In other cases, e.g. using `accessor ...` or a manual `get/set`, we could add an option like `skipDefaultChange` to avoid an initial default value from appearing in the `changedProperties` Map delivered in the update cycle.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1421,12 +1421,11 @@ export abstract class ReactiveElement
...
         for (const [p, options] of elementProperties) {
-          if (
-            options.wrapped === true &&
-            !this._$changedProperties.has(p) &&
-            this[p as keyof this] !== undefined
-          ) {
-            this._$changeProperty(p, this[p as keyof this], options);
+          if (options.wrapped === true && this[p as keyof this] !== undefined) {
+            // Note, deleting an existing value here ensures
+            // the initial old value is seeded as `undefined`
```
justinfagnani, line 1426: When would `_$changedProperties` have something else in it?

packages/reactive-element/src/test/decorators/property_test.ts
```diff
@@ -488,6 +488,53 @@ suite('@property', () => {
...
+    let changedProperties: PropertyValues | undefined;
+
+    class E extends ReactiveElement {
+      @property() unValued?: string;
+      @property() valued = 'valued';
+
+      _valuedAccessor = 'valuedAccessor';
+
+      @property()
+      accessor valuedAutoAccessor = 'valuedAutoAccessor';
```
justinfagnani, line 501: this is covered by the "modern" test, yeah?

justinfagnani (approved):

packages/reactive-element/src/reactive-element.ts
justinfagnani: Hmm, I would think we would return the getter's default value for the old value in that case, but I see how we can't do that if we want the initial value of an accessor to have undefined as its previous value and can't identify an initializer.

PR #4730, opened 2024-08-13
[labs/testing] Add option for init script for ssr worker to allow registering Node.js hooks

kyubisation:
This PR adds an option to the `litSsrPlugin` to allow defining worker init modules, which are sequentially imported at the start of the SSR worker. This allows registering Node.js hooks (https://nodejs.org/api/module.html#customization-hooks) or general setup.

### DIFF
--- a/packages/labs/testing/README.md
+++ b/packages/labs/testing/README.md
@@ -43,9 +43,7 @@ Signature
 `litSsrPlugin(options: LitSsrPluginOptions = {}): TestRunnerPlugin<Payload>`
 
 - `options: object` - Options object containing the following properties
-  - `workerInitModules: string` - Relative paths to modules to be loaded before
-    the fixture modules.
-    This allows registering e.g. Node.js ESM hooks or general setup.
+  - `workerInitModules: string` - Modules to be loaded before the fixture modules. This allows registering e.g. Node.js ESM hooks or general setup. Accepts absolute paths, paths relative to the working directory, or external package modules.
 
 ### Fixtures
 
--- a/packages/labs/testing/README.md
+++ b/packages/labs/testing/README.md
@@ -38,6 +38,15 @@ export default {
 };
 ```
 
+Signature
+
+`litSsrPlugin(options: LitSsrPluginOptions = {}): TestRunnerPlugin<Payload>`
+
+- `options: object` - Options object containing the following properties
+  - `workerInitModules: string` - Relative paths to modules to be loaded before
+    the fixture modules.
+    This allows registering e.g. Node.js ESM hooks or general setup.
+
 ### Fixtures
 
 The package exports functions that will generate fixtures by passing the
--- a/packages/labs/testing/src/test/init-script-hooks.ts
+++ b/packages/labs/testing/src/test/init-script-hooks.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {existsSync, readFileSync} from 'node:fs';
+import type {LoadHook, ResolveHook} from 'node:module';
+import {fileURLToPath} from 'node:url';
+
+import * as ts from 'typescript';
+
+/**
+ * Note: This is not a performant hook implementation,
+ * but serves its purpose for testing.
+ */
+
+const root = new URL('../', import.meta.url).href;
+const tsconfigPath = new URL('./tsconfig.json', root).href;
+const tsconfig = ts.getParsedCommandLineOfConfigFile(
+  fileURLToPath(tsconfigPath),
+  {},
+  ts.sys as unknown as ts.ParseConfigFileHost
+);
+
+const exists = (url: string) => existsSync(fileURLToPath(url));
+const assertTypeScriptFilePath = (
+  url: string,
+  tsUrl = url.replace(/.js$/, '.ts')
+) => (tsUrl && url && !exists(url) && exists(tsUrl) ? tsUrl : null);
+
+export const resolve: ResolveHook = (specifier, context, nextResolve) => {
+  const url =
+    (specifier.startsWith('.') || specifier.startsWith(root)) &&
+    !specifier.includes('/node_modules/') &&
+    context.parentURL?.startsWith(root)
+      ? assertTypeScriptFilePath(new URL(specifier, context.parentURL).href)
+      : null;
+  return url
+    ? {format: 'module', shortCircuit: true, url}
+    : nextResolve(specifier, context);
+};
+
+export const load: LoadHook = (url, context, nextLoad) => {
+  if (
+    url.startsWith(root) &&
+    !url.includes('/node_modules/') &&
+    url.endsWith('.ts')
+  ) {
+    const source = readFileSync(fileURLToPath(url), 'utf8');
+    const result = ts.transpileModule(source, {
+      compilerOptions: tsconfig?.options,
+    });
+    return {format: 'module', shortCircuit: true, source: result.outputText};
+  }
+  return nextLoad(url, context);
+};
--- a/packages/labs/testing/src/test/init-script-hooks.ts
+++ b/packages/labs/testing/src/test/init-script-hooks.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022 Google LLC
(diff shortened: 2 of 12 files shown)


kyubisation:
I'm not quite confident with the naming of `initScript` and open to suggestions for changes.

kyubisation:
@augustjk I updated the PR as discussed during the community call. Can I ask you for a review? 
If you would like to discuss it via call, I am reachable via Discord during CEST (currently UTC+2) time zone business hours (or later).

bennypowers:
this would be very helpful with the eleventy plugin - we could avoid tsc in eleventy.before and save a lot of time restarting the dev server after one-line changes

kyubisation:
@augustjk Sorry to ping you here, but if at all possible, I would like to get this into the next release. Could you have a look at this?

augustjk:
> @augustjk Sorry to ping you here, but if at all possible, I would like to get this into the next release. Could you have a look at this?

No need to be sorry! I apologize for not being on top of it. I'll take a look at it this evening.

augustjk (changes requested):
Thank you so much for this work! Sorry for the very long delay in review.

So from my recollection of the meeting where we discussed this, I think a good approach for this would be not to add the `typeScript` option to the plugin, just keep the `workerModules` option like how you had originally, and we provide the TS hook as a separate export that users could add to the `workerModules` option, maybe something like `workerModules: ['@lit-labs/testing/register-swc-hook']`. I see some residue of a `typescript-worker.js` in the PR, maybe you were thinking the same?

Also thinking I like having "init" in the option name to be clear these are loaded _before_ the component definition modules.. would `workerInitModules` be too long?

A problem with this is it won't be possible to provide a separate tsconfig. So perhaps we provide also provide the hook file directly and document how to write a module that registers the hook providing the tsconfig path as data.

My biggest hesitation with baking the hook directly in as a plugin option is that the choice of `swc` is quite opinionated and I don't want that to solidify as the canonical TS compiler we use here. Just vending the hook makes it clearer that it's one option, but not _the_ option for compiling typescript.

And this is just out of curiosity, why `swc` versus `esbuild`? I think I've seen more `esbuild` use than `swc` outside of nextjs. It's good to see `swc` has also implemented standard decorators with metadata, but oh my that large emit size.. 😵 

Edit to add: I overlooked the part where you linked to amaro which also uses swc, so maybe that answers the why.

augustjk:
> Hey there, FWIW I implemented an alternative approach to this in [RedHat-UX/red-hat-design-system#1992](https://github.com/RedHat-UX/red-hat-design-system/pull/1992). Take a look if you're interested.

Thank you for sharing! [TSX](https://tsx.is/dev-api/) actually looks really robust.

It's making me rethink about just adding allow TS option for the plugin that just uses their [`tsImport`](https://tsx.is/dev-api/ts-import).

kyubisation:
> Thank you so much for this work! Sorry for the very long delay in review.

No problem, thanks for the review 😃 

> So from my recollection of the meeting where we discussed this, I think a good approach for this would be not to add the `typeScript` option to the plugin, just keep the `workerModules` option like how you had originally, and we provide the TS hook as a separate export that users could add to the `workerModules` option, maybe something like `workerModules: ['@lit-labs/testing/register-swc-hook']`. I see some residue of a `typescript-worker.js` in the PR, maybe you were thinking the same?

Yes, this was the initial plan. However, how would we provide configuration (e.g. tsconfig.json path) for the separate module, without there being a disconnect between `litSsrPlugin` options and additionally having to add that separate module. I went for convenience instead. But I'm open for suggestions.

> Also thinking I like having "init" in the option name to be clear these are loaded _before_ the component definition modules.. would `workerInitModules` be too long?

I'm fine with that 👍 

> A problem with this is it won't be possible to provide a separate tsconfig. So perhaps we provide also provide the hook file directly and document how to write a module that registers the hook providing the tsconfig path as data.

See answer above.

> My biggest hesitation with baking the hook directly in as a plugin option is that the choice of `swc` is quite opinionated and I don't want that to solidify as the canonical TS compiler we use here. Just vending the hook makes it clearer that it's one option, but not _the_ option for compiling typescript.

That's fair, but the swc compiler in WebAssembly form is the one used by the Node.js experimental TypeScript support (without decorators). Due to this it would be closely aligned with native Node.js behavior.
We could also add additional implementations, although I'm not sure whether this could not already be covered by e.g. ts-node and tsx.
 
> And this is just out of curiosity, why `swc` versus `esbuild`? I think I've seen more `esbuild` use than `swc` outside of nextjs. It's good to see `swc` has also implemented standard decorators with metadata, but oh my that large emit size.. 😵
> 
> Edit to add: I overlooked the part where you linked to amaro which also uses swc, so maybe that answers the why.

Yeah, that's the primary reason. (It's also nice to have technically a single dependency, versus the dozen platform/arch specific dependencies esbuild adds, although it's a bit hypocritical, as I use esbuild in a lot of my projects.)

kyubisation:
@bennypowers Thanks for the input. 😃 👍 
I did consider tsx/esbuild, but went with swc due to the Node.js alignment.

kyubisation:
@augustjk To unblock this PR, I could split the TypeScript part into a separate PR. I only implemented the TypeScript part, as it was requested during a community call.

augustjk:
> @augustjk To unblock this PR, I could split the TypeScript part into a separate PR. I only implemented the TypeScript part, as it was requested during a community call.

Sure let's do that. Sorry for the multiple back and forth.

kyubisation:
@augustjk Done. Please have another look.

kyubisation:
@augustjk The build seems to crash due to an issue in packages/labs/signals, which I doubt is caused by this change. Do you have an idea as to why?

kyubisation:
@augustjk The breaking build was only a problem locally, from a previous state, after the rebase. Should work fine as it is.

packages/labs/testing/README.md
```diff
@@ -38,6 +38,15 @@ export default {
...
 ```
 
+Signature
+
+`litSsrPlugin(options: LitSsrPluginOptions = {}): TestRunnerPlugin<Payload>`
+
+- `options: object` - Options object containing the following properties
+  - `workerInitModules: string` - Relative paths to modules to be loaded before
+    the fixture modules.
+    This allows registering e.g. Node.js ESM hooks or general setup.
```
augustjk, line 48:
```suggestion
  - `workerInitModules: string` - Modules to be loaded before the fixture modules. This allows registering e.g. Node.js ESM hooks or general setup. Accepts absolute paths, paths relative to the working directory, or external package modules.
```

packages/labs/testing/src/test/init-script-hooks.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
augustjk, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

packages/labs/testing/src/test/init-script.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
augustjk, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

packages/labs/testing/src/test/init-script_test.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
augustjk, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

packages/labs/testing/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     "tsBuildInfoFile": ".tsbuildinfo",
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
-    "types": ["mocha"]
+    "types": ["mocha"],
+    "skipLibCheck": true
```
augustjk, line 23: was this addition necessary?

augustjk (commented):
Thank you! Just a few minor things.

packages/labs/testing/tsconfig.json
kyubisation: Yes, as the chokidar types fail with the Node.js types update.

augustjk (approved):
Thank you!

PR #4731, opened 2024-08-13
[labs/testing] Add fallback for call site location detection for webkit

### LINKED ISSUE #4067: [labs/testing] Call site location by stack trace might not work in webkit

kyubisation:
The fixtures currently look for the wtr-session-id query string. This PR adds a fallback, when wtr-session-id is not found, to look for the first URL that does not contain node_modules or the path to the fixture file.

Fixes #4067

### DIFF
--- a/.changeset/rude-cherries-cover.md
+++ b/.changeset/rude-cherries-cover.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/testing': patch
+---
+
+Add fallback for call site location detection for webkit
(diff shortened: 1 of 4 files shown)


augustjk (approved):
Thank you!

PR #4732, opened 2024-08-13
Monorepo playground

graynorton:
**Edit: Ready for approval**

Per discussions in the review here and in the last Lit Eng meeting, I've removed the commits for the illustrative playground examples and am taking this PR out of Draft status—it's now ready for approval.

---

As discussed in a recent Lit Eng meeting, here's a draft PR exploring what it might look like to add a playground to the Lit monorepo.

See playground/README.md for rationale and details.

Making this a draft primarily because I included a couple of additional commits, adding specific playground examples for the purpose of illustration; for cleanliness, we'd probably want to remove those before landing this PR.


playground/p/issues/4672/_import.info.json
```diff
@@ -0,0 +1,11 @@
+{
+  "source": "https://lit.dev/playground/#gist=aece301df423847e72b1a85e1e9caf75",
```
sorvell, line 2: Is this intended to be committed? If so, it seems weird to name it by issue.

playground/scripts/template/package.json
```diff
@@ -0,0 +1 @@
+{}
```
sorvell, line 1: Is it possible to include defaults here that make the example installable standalone?

playground/README.md
```diff
@@ -0,0 +1,187 @@
+# Playground
```
sorvell, line 1: Prefer not to call this `playground` as it's confusingly like the lit.dev playground but has different functionality and is for different use cases.

I think it would make the most sense to just integrate this with the existing `examples` folder, if possible.

playground/README.md
```diff
@@ -0,0 +1,187 @@
...
+  tripped up by edge cases
+
+  ### Implementation details
+
+- The `@lit-internal/playground` package is a workspace within
+  the monorepo so that its dependencies (TypeScript and Vite)
+  are installed via `npm install` at the monorepo root
+- Individual examples are not workspaces; each example's
+  external dependencies (if any) are installed locally
+- Examples use the [Vite](https://vitejs.dev) dev server
```
sorvell, line 176: The default tsconfig/vite setup doesn't support using the `accessor` keyword. Given our need to support this for standard decorators and its superior integration with class fields, it'd be great to make this work.

I tried some different settings but wasn't able to make it work, either via configuring ts or esbuild.

.gitignore
```diff
@@ -7,4 +7,4 @@ node_modules/
 /temp
 .idea
 /scripts/check-size-out.md
-
+/playground/p/**/package-lock.json
```
sorvell, line 10: Can we add a folder here for playgrounds to be auto-ignored and call it out in the readme?

e.g. `/playground/scratch/**`

playground/README.md
graynorton: I see the use cases for this and the Lit.dev playground as highly aligned, actually. The Lit.dev playground is optimized for small, targeted examples that demonstrate a particular feature, technique, issue, etc. It tries to reduce the friction around creating and running such examples to as close to zero as possible, and that's exactly what this PR is trying to do in the context of the monorepo. I wanted the name of the folder in the monorepo to convey this similarity of purpose. For that reason, I briefly considered similar-but-not-quite-the-same names (e.g. "sandbox"), and am still open to something like that, but I found the benefit of association that came with using exactly the same name to outweigh whatever confusion might arise from the fact that the implementation is different.

In contrast, the things we currently have in the monorepo's `examples` folder are (at least currently) for demonstrating and testing integrations of Lit into various external stacks. They tend to be more complex, with nested subfolders, and each one has its own "infrastructure" (dev server, config files, etc.), whereas my vision for the monorepo playground is (again, like the Lit.dev playground) for the individual examples not to be concerned with any of that stuff.

If we are especially concerned about adding folders to the top level of the monorepo, I could imagine nesting `playground` (or whatever we end up deciding call it) inside `examples`, but its feels more like a peer than a child to me.

playground/scripts/template/package.json
graynorton: The Lit libraries are intentionally left out of the `package.json` so that examples will automatically default to using the versions from the current branch of the monorepo.

I wasn't thinking of standalone examples as a P0/P1 use case, but if we want to support that, I think the best way of doing it would be to add an `export:standalone` script that adds deps into the exported `package.json`.

playground/p/issues/4672/_import.info.json
graynorton: Yes. In working on Virtualizer issues, I've found that committing a minimal repro, followed by the fix, makes for a very convenient PR review process. A reviewer can check out the PR branch, temporarily detach HEAD to view the repro without the fix, then switch back to the tip of the branch to review the fix.

I've also found through experience that when a repro case is part of a workflow like this, it's easiest to use the issue number as the name of the repro, to minimize cognitive load.

.gitignore
graynorton: Sure, that's a good idea!

playground/README.md
graynorton: Thanks for reporting—I'll take a look and see if I can make it work!

graynorton: I just pushed an update that allows standard decorators to be used on a per-example basis, and added an example to illustrate. It requires inclusion of a `tsconfig.json` in the example folder, but that can be a one-liner that just imports a standard-decorator-friendly base config, which I've added to the `playground` folder.

If this is too much of a pain, we might consider adding an option to the `npm run new` script that would add the `tsconfig.json` upon creation.

.gitignore
graynorton: Added!

sorvell (approved):

PR #4734, opened 2024-08-16
perf(context): skip expensive computation of Event.composedPath()

sorin-davidoi:
While profiling some Lit components we noticed that `Event#composedPath()` is in the hot path and that the browser spends a significant amount of time computing it.

For a particular use case (initial render of hundreds of components), ~29% of the time is spent in this function alone. A slight change to the bail out logic reduces this to ~17% (basically don't call it unless absolutely necessary). Keeping manual track of the original source of the event reduces this overhead completely.

Furthermore, this will also slightly reduce memory usage and time spent in garbage collection (for the same use case, Firefox Profiler shows a reduction in memory allocated by this function from ~998KB to ~511KB with the bailout logic and the complete elimination of this allocations with the manual tracking of the source of the event).

While this change is unlikely to make a difference for most consumers, it should slightly reduce the overhead of event dispatching, which seems to be non-trivial when lots of components that are both providers and consumers are present in the DOM.

Note that the prospect of `Event#composedPath()` being a potential performance bottleneck was raised in https://github.com/lit/lit/pull/2858#discussion_r873855309, where it was correctly dismissed. But https://github.com/lit/lit/pull/4014 reworked the implementation, and the original dismissal reason ("[...] is limited to host that consume and provide the same context, so it's not called on all events") no longer applies.

### DIFF
--- a/packages/context/src/lib/context-root.ts
+++ b/packages/context/src/lib/context-root.ts
@@ -97,7 +97,8 @@ export class ContextRoot {
     // Note, it's important to use the initial target
     // since that's the requesting element and the event may be re-targeted
     // to an outer host element.
-    const element = event.contextTarget as HTMLElement;
+    const element = (event.contextTarget ??
+      event.composedPath()[0]) as HTMLElement;
     const callback = event.callback;
 
     let pendingContextRequests = this.pendingContextRequests.get(event.context);
--- a/packages/context/src/lib/context-root.ts
+++ b/packages/context/src/lib/context-root.ts
@@ -80,7 +80,7 @@ export class ContextRoot {
       } else {
         // Re-dispatch if we still have the element and callback
         element.dispatchEvent(
-          new ContextRequestEvent(event.context, callback, true)
+          new ContextRequestEvent(event.context, element, callback, true)
         );
       }
     }
--- a/packages/context/src/lib/context-root.ts
+++ b/packages/context/src/lib/context-root.ts
@@ -94,10 +94,10 @@ export class ContextRoot {
       return;
     }
 
-    // Note, it's important to use the initial target via composedPath()
+    // Note, it's important to use the initial target
     // since that's the requesting element and the event may be re-targeted
     // to an outer host element.
-    const element = event.composedPath()[0] as HTMLElement;
+    const element = event.contextTarget as HTMLElement;
     const callback = event.callback;
 
     let pendingContextRequests = this.pendingContextRequests.get(event.context);
--- a/packages/context/src/lib/controllers/context-provider.ts
+++ b/packages/context/src/lib/controllers/context-provider.ts
@@ -97,23 +97,13 @@ export class ContextProvider<
     if (ev.context !== this.context) {
       return;
     }
-    let consumerHost: Element | undefined;
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
-    if (ev.target === this.host) {
-      // The check on composedPath (on top of ev.target) is to cover cases
-      // where the consumer is in the shadowDom of the provider.
-      consumerHost = ev.composedPath()[0] as Element;
-      if (consumerHost === this.host) {
-        return;
-      }
+    if (ev.contextTarget === this.host) {
+      return;
     }
     ev.stopPropagation();
-    this.addCallback(
-      ev.callback,
-      consumerHost ?? (ev.target as Element),
-      ev.subscribe
-    );
+    this.addCallback(ev.callback, ev.contextTarget, ev.subscribe);
   };
 
   /**
@@ -162,7 +152,7 @@ export class ContextProvider<
       }
       seen.add(callback);
       consumerHost.dispatchEvent(
-        new ContextRequestEvent(this.context, callback, true)
+        new ContextRequestEvent(this.context, consumerHost, callback, true)
       );
     }
     ev.stopPropagation();
--- a/packages/context/src/lib/controllers/context-provider.ts
+++ b/packages/context/src/lib/controllers/context-provider.ts
@@ -97,19 +97,23 @@ export class ContextProvider<
     if (ev.context !== this.context) {
       return;
     }
+    let consumerHost: Element | undefined;
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
     if (ev.target === this.host) {
-      return;
-    }
-    // The check on composedPath (on top of ev.target) is to cover cases
-    // where the consumer is in the shadowDom of the provider.
-    const consumerHost = ev.composedPath()[0] as Element;
-    if (consumerHost === this.host) {
-      return;
+      // The check on composedPath (on top of ev.target) is to cover cases
+      // where the consumer is in the shadowDom of the provider.
+      consumerHost = ev.composedPath()[0] as Element;
+      if (consumerHost === this.host) {
+        return;
+      }
     }
     ev.stopPropagation();
-    this.addCallback(ev.callback, consumerHost, ev.subscribe);
+    this.addCallback(
+      ev.callback,
+      consumerHost ?? (ev.target as Element),
+      ev.subscribe
+    );
   };
 
   /**
--- a/packages/context/src/lib/controllers/context-provider.ts
+++ b/packages/context/src/lib/controllers/context-provider.ts
@@ -99,9 +99,11 @@ export class ContextProvider<
     }
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
-    // The check on composedPath (as opposed to ev.target) is to cover cases
-    // where the consumer is in the shadowDom of the provider (in which case,
-    // event.target === this.host because of event retargeting).
+    if (ev.target === this.host) {
+      return;
+    }
+    // The check on composedPath (on top of ev.target) is to cover cases
+    // where the consumer is in the shadowDom of the provider.
     const consumerHost = ev.composedPath()[0] as Element;
     if (consumerHost === this.host) {
       return;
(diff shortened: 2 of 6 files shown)


sorvell:
It would be great to add a benchmark test (in the benchmarks package) that validates this change.

Also, it might be worth considering supporting an optional `rootElement` on these events, with the idea that this would be purely an implementation detail to facilitate optimization.

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -94,13 +94,16 @@ export class ContextProvider<
...
+      return;
+    }
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
     // The check on composedPath (as opposed to ev.target) is to cover cases
     // where the consumer is in the shadowDom of the provider (in which case,
     // event.target === this.host because of event retargeting).
     const consumerHost = ev.composedPath()[0] as Element;
-    if (ev.context !== this.context || consumerHost === this.host) {
+    if (consumerHost === this.host) {
```
justinfagnani, line 106: When the consumer is in the light DOM of the provider, I think we can further reduce the calls to `ev.composedPath()` by checking target first:

```suggestion
    if (event.target === this.host && ev.composedPath()[0] === this.host) {
```

Same below.


We could also avoid `composedPath()` altogether if we added another property to the ContextRequestEvent, like `contextTarget`. We'd set this to the dispatching host before dispatch and check it here. If the property doesn't exist, then we fallback to composed path to be compatible with the context protocol.

If that's really beneficial, we can bring `.contextTarget` to the protocol, and also open a standards issue for a property like `composedTarget` for these cases when you really don't need to whole path.

justinfagnani:
@sorin-davidoi thanks for finding this!

Did you see the same perf hit on all browsers?

Westbrook:
While this can effect _all_ events, for sure, for initial render bound events (like `context-request`) would it be useful to submit an addition to [the Context Protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md) that included the `composedTarget` on the custom event so that the path does not need to be computed to complete the workflow proposed by the protocol?

Westbrook:
Not squatting is a good call! I'm just looking to make sure this (and any of the other learnings this implementation has had) get brought back to the protocol. 🙏🏼

packages/context/src/lib/controllers/context-provider.ts
sorin-davidoi: > When the consumer is in the light DOM of the provider, I think we can further reduce the calls to ev.composedPath() by checking target first

Good spot, applied the change to both places :+1:

sorin-davidoi:
> Did you see the same perf hit on all browsers?

Saw it in Chrome as well, didn't check WebKit-based browsers.

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -94,13 +94,18 @@ export class ContextProvider<
...
     // Only call the callback if the context matches.
+    if (ev.context !== this.context) {
+      return;
+    }
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
-    // The check on composedPath (as opposed to ev.target) is to cover cases
-    // where the consumer is in the shadowDom of the provider (in which case,
-    // event.target === this.host because of event retargeting).
+    if (ev.target === this.host) {
```
justinfagnani, line 102: I think this needs to be 
```suggestion
    if (ev.target !== this.host) {
```

It's only if the target and the host are the same that we need to disambiguate with the composed path check. If they're different, then we know the consumer and provider are not the same element.

packages/context/src/lib/controllers/context-provider.ts
justinfagnani: If you're up to it I do think it'd be better to put a `contextTarget` property on ContextEvent. I suspect that a majority of context use is with consumers in the shadow root of the provider, so this small addition check here won't help them that much, while `contextTarget` could eliminate the composed path check for all Lit-to-Lit context scenarios.

justinfagnani (changes requested):

sorin-davidoi:
> If you're up to it I do think it'd be better to put a `contextTarget` property on ContextEvent. I suspect that a majority of context use is with consumers in the shadow root of the provider, so this small addition check here won't help them that much, while `contextTarget` could eliminate the composed path check for all Lit-to-Lit context scenarios.

I gave it a shot, not entirely sure I've made the right changes though.

sorin-davidoi:
@justinfagnani Can you perhaps trigger the CI? The context tests are passing for me locally but some of the others are timing out (also on `main`, so probably related to my setup and not the changes).

packages/context/src/lib/context-root.ts
```diff
@@ -94,10 +94,10 @@ export class ContextRoot {
       return;
     }
 
-    // Note, it's important to use the initial target via composedPath()
+    // Note, it's important to use the initial target
     // since that's the requesting element and the event may be re-targeted
     // to an outer host element.
-    const element = event.composedPath()[0] as HTMLElement;
+    const element = event.contextTarget as HTMLElement;
```
justinfagnani, line 100: To be interoperable with the context protocol we need to fallback to `composedPath()[0]` when `contextTarget` isn't set.

```suggestion
    const element = (event.contextTarget ?? event. composedPath()[0]) as HTMLElement;
```

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -94,17 +97,16 @@ export class ContextProvider<
...
+      return;
+    }
     // Also, in case an element is a consumer AND a provider
     // of the same context, we want to avoid the element to self-register.
-    // The check on composedPath (as opposed to ev.target) is to cover cases
-    // where the consumer is in the shadowDom of the provider (in which case,
-    // event.target === this.host because of event retargeting).
-    const consumerHost = ev.composedPath()[0] as Element;
-    if (ev.context !== this.context || consumerHost === this.host) {
+    if (ev.contextTarget === this.host) {
```
justinfagnani, line 105:
```suggestion
    const consumerHost = ev.contextTarget ?? ev.composedPath()[0] as Element;
    if (consumerHost === this.host) {
```

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -94,17 +97,16 @@ export class ContextProvider<
...
       return;
     }
     ev.stopPropagation();
-    this.addCallback(ev.callback, consumerHost, ev.subscribe);
+    this.addCallback(ev.callback, ev.contextTarget, ev.subscribe);
```
justinfagnani, line 109:
```suggestion
    this.addCallback(ev.callback, consumerHost, ev.subscribe);
```

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -117,13 +119,12 @@ export class ContextProvider<
...
+      return;
+    }
     // Also, in case an element is a consumer AND a provider
     // of the same context it shouldn't provide to itself.
-    // We use composedPath (as opposed to ev.target) to cover cases
-    // where the consumer is in the shadowDom of the provider (in which case,
-    // event.target === this.host because of event retargeting).
-    const childProviderHost = ev.composedPath()[0] as Element;
-    if (ev.context !== this.context || childProviderHost === this.host) {
+    if (ev.contextTarget === this.host) {
```
justinfagnani, line 127:
```suggestion
    const childProviderHost = ev.contextTarget ?? ev.composedPath()[0] as Element;
    if (childProviderHost === this.host) {
```

justinfagnani (changes requested):

justinfagnani:
> @justinfagnani Can you perhaps trigger the CI? The context tests are passing for me locally but some of the others are timing out (also on `main`, so probably related to my setup and not the changes).

I don't have the usual button available to start the CI? It could be the length of time since the last changes? If you push new changes for my comments, I think I'll be able to again.

sorin-davidoi:
What incantation is needed to get a Tachometer Benchmark Results like [this one](https://github.com/lit/lit/pull/4769#issuecomment-2357278016)?

augustjk:
> What incantation is needed to get a Tachometer Benchmark Results like [this one](https://github.com/lit/lit/pull/4769#issuecomment-2357278016)?

I'm not sure why PR doesn't have that comment but the benchmarks only test the core lit-html, reactive-element, and lit-element packages so they won't give us any insight for the changes here.

augustjk (approved):
Changes look good to me. @justinfagnani what do you think?
As we land this, we should also make a PR to update the community protocol.

sorin-davidoi:
@augustjk @justinfagnani We're still seeing `Event.composedPath` taking up a significant amount of time in the performance traces we take - is there any chance of getting this merge or should I close this?

justinfagnani (approved):
Thanks!

robrez:
I have an app with some mixed tech:  some components aren't lit-based.  Early in the dom tree I have a lit-based context provider.  Some of the non-lit components deeper in the tree have been using types / classes exported by lit/context to perform manual event dispatching:   `new ContextRequestEvent(context, callback, subscribe?)`

Since `1.1.4` that ctor is broken. This easy to fix in my example by having those "newer-uppers" use the 4-arg ctor.. there aren't many of them.  Reading above, it sounds like this change is likely to make it into the context protocol literature

Thought I'd mention it:  maybe other wacky consumers like mine

augustjk:
oops, that's a good point. this is indeed a breaking change in terms of the constructor signature across versions, though it does handle the lack of `contextTarget` field in the event itself. looks like we do indeed export the `ContextRequestEvent` from the package rather than leaving it as an internal implementation detail 😨 

i've made a note to get these changes into the protocol here https://github.com/webcomponents-cg/community-protocols/issues/50#issuecomment-2673555681

PR #4736, opened 2024-08-19
chore(labs/rollup-plugin-minify-html-literals): update dependencies

43081j:
Updates various dependencies to the latest version, including rollup to 4.x.

### DIFF
--- a/packages/labs/rollup-plugin-minify-html-literals/package.json
+++ b/packages/labs/rollup-plugin-minify-html-literals/package.json
@@ -53,11 +53,13 @@
       ]
     },
     "test": {
-      "command": "nyc mocha -r ts-node/register -r source-map-support/register test/*.js",
+      "command": "nyc mocha -r source-map-support/register test/*.js",
       "dependencies": [
         "build"
       ],
-      "files": [],
+      "files": [
+        "package.json"
+      ],
       "output": [
         ".nyc_output/"
       ]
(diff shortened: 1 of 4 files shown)


Westbrook (approved):
LGTM.

packages/labs/rollup-plugin-minify-html-literals/package.json
```diff
@@ -53,11 +53,13 @@
...
     },
     "test": {
-      "command": "nyc mocha -r ts-node/register -r source-map-support/register test/*.js",
+      "command": "nyc mocha -r source-map-support/register test/*.js",
       "dependencies": [
         "build"
       ],
-      "files": [],
+      "files": [
+        "package.json"
```
justinfagnani, line 61: package.json doesn't need to be listed in files: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

justinfagnani (approved):
thanks!

packages/labs/rollup-plugin-minify-html-literals/package.json
43081j: this was wireit's `files` so changing the `nyc` config would trigger a cache clear

i dropped `nyc` in the end but we do still have a c8 config in the same place. upto you, it was just a convivence thing while i was running locally

bencbradshaw (approved):
Looks good. Thanks for the PR!

augustjk (approved):
Nice, thank you!

packages/labs/rollup-plugin-minify-html-literals/package.json
justinfagnani: Ah, ok. I misread which files property this was.

Since you dropped nyc, do you still need the `"nyc"` section below?

augustjk: that part is done in https://github.com/lit/lit/pull/4737

Issue #4745, opened 2024-08-21
[labs/analyzer] Support native private fields


justinfagnani:
### Which package(s) are affected?

Analyzer (@lit-labs/analyzer)

### Description

Native private fields are ignored and log a warning:

```
src/lib/tree-view/tree-node.ts:63:3 - warning TS548001: @lit-labs/analyzer only supports analyzing class properties named with plain identifiers. This property was ignored.

63   #slotController = new SlotController(this);
```

### Reproduction

Use a private field in an element

### Workaround

None needed - private fields can be safely ignored.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

All

### Browser/OS/Node environment

n/a

PR #4746, opened 2024-08-22
[labs/analyzer] Handle class private fields in LitElement analysis

justinfagnani:
Fixes https://github.com/lit/lit/issues/4745

This change allows private fields to be handled, but because they're not likely to be reactive properties, they're effectively ignored in this code path.

### DIFF
--- a/packages/labs/analyzer/src/lib/lit-element/properties.ts
+++ b/packages/labs/analyzer/src/lib/lit-element/properties.ts
@@ -37,13 +37,14 @@ export const getProperties = (
   let staticProperties;
 
   for (const prop of propertyDeclarations) {
-    if (!ts.isIdentifier(prop.name)) {
+    if (!ts.isIdentifier(prop.name) && !ts.isPrivateIdentifier(prop.name)) {
       analyzer.addDiagnostic(
         createDiagnostic({
           typescript: ts,
           node: prop,
           message:
-            '@lit-labs/analyzer only supports analyzing class properties named with plain identifiers. This ' +
+            '@lit-labs/analyzer only supports analyzing class properties ' +
+            'named with plain identifiers, or private class fields. This ' +
             'property was ignored.',
           category: ts.DiagnosticCategory.Warning,
           code: DiagnosticCode.UNSUPPORTED,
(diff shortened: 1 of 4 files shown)


packages/labs/analyzer/src/lib/lit-element/properties.ts
```diff
@@ -37,13 +37,14 @@ export const getProperties = (
   let staticProperties;
 
   for (const prop of propertyDeclarations) {
-    if (!ts.isIdentifier(prop.name)) {
+    if (!ts.isIdentifier(prop.name) && !ts.isPrivateIdentifier(prop.name)) {
```
justinfagnani, line 40: @kevinpschaaf questions on this function, or you remember: it's named `getProperties()`, but only returns the reactive properties. 1) Should it be renamed? 2) Since reactive properties are kind of an implementation detail, and a getter/setter pair or different decorator could make a property reactive... are we handling those cases, or should we even distinguish reactive properties from others? Is this to do something else like find attributes?

packages/labs/analyzer/test-files/ts/properties/src/element-a.ts
```diff
@@ -99,6 +99,11 @@ export class ElementA extends LitElement {
   @property()
   union: LocalClass | HTMLElement | ImportedClass;
 
+  @state()
+  private _stateField: number;
+
+  #privateField = 0;
```
aomarks, line 105: Is this enough to make the test fail without the patch (seems like there is no test code in this PR that explicitly checks it)

aomarks (approved):

packages/labs/analyzer/test-files/ts/properties/src/element-a.ts
justinfagnani: Yes, just the presence of the field causes a diagnostic before this change.

packages/labs/analyzer/src/lib/lit-element/properties.ts
kevinpschaaf: Yeah, iirc this was `properties` in the `@property` sense?  The term `field` is used in the code to refer to generic fields. But yeah could rename to `getReactiveProperties()`.

I think these would be used to determine the `attributes` for e.g. CEM (I see there's still a TODO in there), but also in the Angular wrapper to wrap them with `@Input()`.

PR #4749, opened 2024-08-27
Use consistent prettier config for lit-html package

augustjk:
When prettier was updated in https://github.com/lit/lit/pull/4559 the [default config for trailing comma](https://prettier.io/docs/en/options.html#trailing-commas) changed from "es5" to "all", adding a lot of trailing comma to function params, specifically for the `lit-html` package as it had its own `.prettierrc.json` file without that option set.

This is inconsistent with the monorepo root prettierrc which explicitly had `"trailingComma": "es5"`.

This PR removes `/packages/lit-html/src/.prettierrc.json` and adds the `"embeddedLanguageFormatting": "off"` to the root `.prettierrc.json` file which prevents static portions of the tagged template literal from getting formatted.

Alternatively, I considered just adding the explicit trailing comma es5 setting to the lit-html prettierrc but figured it would be best to just keep one config file for the whole monorepo.

Besides the removal of trailing commas from function params/args in `lit-html`, the embeddedLanguageFormatting affected only test files for the other repos due to the formatting of tagged template literals.


AndrewJakubowicz (approved):
Nice!

rictic (approved):
Consistency is good. Started a discussion on comma policy in the discord, but getting consistency with other packages is an improvement.

PR #4755, opened 2024-09-02
[labs/ssr] Implement SSR custom elements event handling

### LINKED ISSUE #3301: [labs/context] SSR compatibility
### Should this be an RFC?

Context (@lit-labs/context)

### Description

Before graduating `@lit-labs/context` and `@lit-labs/ssr` we should implement context support in SSR. This may be done via general event support in SSR as in #2309, or if that's not feasible, in some other way.

None

### LINKED ISSUE #3365: [labs/ssr] Events
We should have some form of support in SSR for events.

https://github.com/lit/lit/pull/2309 proposes an implementation. Questions (copied from that PR description):

* Can we have an event path of just the custom elements without materializing the built-ins that would be on the path with a full DOM structure? This seems ok if the only event patterns we support are for cross-custom-element communication, ie context.
* Where do elements install event listeners? I added a call to `connectedCallback()` in here, but that only doesn't throw because none of the test elements do anything with the DOM in `connectedCallback()`. That wouldn't be viable in general.
* Do we need to implement the capture phase?
* How could we implement the proper composed path? We would need some concept of slotted elements being in the custom element instance stack.

kyubisation:
This PR implements limited event handling for custom elements, by shimming `EventTarget` (with methods `addEventListener`, `removeEventListener` and `dispatchElement`) which is now extended by the `HTMLElement` SSR shim. It only considers and allows event handlers and event dispatching on custom element instances (e.g. event handlers in a html`...` template will not break, but also not work with SSR). Due to the streaming nature of SSR, it will also not allow mutating the parent from a child for rendering. This primarily enables a use case like `@lit/context`.

It works by building a minimal DOM tree on server side, which only considers custom element instances and slot elements, to respect the event path including Shadow DOM. The event properties are patched during the `dispatchEvent` call, to more accurately reflect the state of the event.
This PR also adds a global flag to run `connectedCallback` (and `hostConnected` on registered controllers) with SSR, which can be enabled with `globalThis.litSsrCallConnectedCallback = true`.
Additionally, a global property `globalThis.litServerRoot` has been implemented, which can be used as an alternative for `document.documentElement` or `document.body` to register global event handlers (e.g. for `@lit/context ContextProvider`).

Closes #3301
Closes #3365


packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,538 @@
...
+    } else if (eventListeners.has(callback)) {
+      return;
+    }
+
+    const normalizedOptions =
+      typeof options === 'object' && options ? options : {};
+    normalizedOptions.signal?.addEventListener('abort', () =>
+      this.removeEventListener(type, callback, options)
+    );
+    // We are ignoring capture, as we expecte separate instances of this class to be used for that.
```
jeripeierSBB, line 55:
```suggestion
    // We are ignoring capture, as we expect separate instances of this class to be used for that.
```

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -66,14 +66,19 @@ export class LitElementRenderer extends ElementRenderer {
   }
 
   override connectedCallback() {
+    // Currently we only call connectedCallback as an opt-in,
+    // as this could break consumers.
+    if (globalThis.litSsrCallConnectedCallback) {
+      this.element.connectedCallback();
```
sorvell, line 72: I'm not clear on why this is ok here. Doesn't it call `createRenderRoot` and enable asynchronous updating? Both seem problematic...

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,537 @@
...
+    // in this scenario.
+    let stopPropagation = false;
+    let stopImmediatePropagation = false;
+    let eventPhase = 0;
+    let currentTarget: EventTarget | null = null;
+    const originalStopPropagation = event.stopPropagation;
+    const originalStopImmediatePropagation = event.stopImmediatePropagation;
+    Object.defineProperties(event, {
+      target: {
+        value: this,
```
sorvell, line 103: Does this need to do event re-targeting? See https://dom.spec.whatwg.org/#retarget

sorvell (changes requested):

packages/labs/ssr/src/lib/lit-element-renderer.ts
kyubisation: What would you suggest? Wrapping the call to `createRenderRoot` and `enableUpdating` in an if with `isServer`?
A separate method (e.g. `connectedCallbackServer`) was rejected during a previous community call.

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: I was not sure whether this is necessary for our limited use case.
But I could probably adapt that. My interpretation is that I could track a target similar to eventPhase and currentTarget. To be spec compliant I would probably also have to calculate eventPhase dependent on that.

packages/labs/ssr/src/lib/lit-element-renderer.ts
sorvell: Try `LitElement.prototype.connectedCallback = function() {}` and probably should add a test that confirms the element does *not* try to update (e.g. implement firstUpdated and wait for a setTimeout and confirm it's not called).

kyubisation: My understanding was that we want to call the controllers (`this.__controllers?.forEach((c) => c.hostConnected?.());`), in order for the context package to work. So should it be `LitElement.prototype.connectedCallback = function() { this.__controllers?.forEach((c) => c.hostConnected?.()) }`?
Also where should I patch this? The first time it is called in `LitElementRenderer`?

sorvell: Yeah, that's fair. Err, `__controllers` would likely have to be added to `lit-element/private-ssr-support.js`.

Perhaps instead, we could no-op `createRenderRoot` and `enableUpdating` but `enableUpdating` is set on the instance so it's probably simplest just to add this before calling `connectedCallback()`:

```js
// Optionally call connectedCallback via setting: `litSsrCallConnectedCallback`
// Enable this flag to process events dispatched handled via connectedCallback.
if (globalThis.litSsrCallConnectedCallback) {
  // Prevent element from trying to create a shadowRoot or enable updating
  // since these are both specially handled for server rendering.
  this.element.createRenderRoot = function() {};
  this.element.enableUpdating = function() {};
  this.element.connectedCallback();
}
// ...
```

And we should add tests to confirm this behavior.

packages/labs/ssr-dom-shim/src/lib/events.ts
sorvell: Yeah, it's hard to say. We also don't have a shadowRoot for the composedPath, so that's another infidelity, right?

I guess I'd be inclined to add retargetting if it is simple since the actual target setting is almost as straightforward as currentTarget (set target to the element if it has a shadowRoot, which I think we know?), and yes that would be in both phases.

I'm ok skipping this and documenting the limitation if it becomes too hard to manage.

kyubisation: I'll have a look at what can be done. 👍

packages/labs/ssr/src/lib/lit-element-renderer.ts
kyubisation: Quick question; With this the controller hook `hostDisconnected` is called, but the `hostUpdate` hook, which is nested between `willUpdate` and `update`, will not be called. Is that acceptable for now?

sorvell: I assume you mean `hostConnected`? 

I'm not sure about `hostUpdate`. I think we can start with not calling it and just making sure it's commented clearly that it's not called.

kyubisation: Yes, I meant `hostConnected`. 😅 
How should I best document this? In the `ReactiveController` docs?

sorvell: In the code where `willUpdate` is specially called and perhaps add a bullet [here](https://github.com/lit/lit/tree/main/packages/labs/ssr#notes-and-limitations) about events and controllers?

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: Was a bit more complicated than I thought, but it should retarget correctly.

packages/labs/ssr/src/lib/lit-element-renderer.ts
kyubisation: Done so. Can you have another look?

Additional question for `@lit/context`; In the browser you can use e.g. `document.body` for `ContextRoot`. It would be helpful if we had something similar for the server. However, currently my only idea is to define a constant on `globalThis` (e.g. `serverEventRoot`), which would be added to the start of `RenderInfo.eventTargetStack`.
Then you could use `new ContextRoot().attach(isServer ? serverEventRoot : document.body)`.
What do you think? Do you have another idea?

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,567 @@
...
+ * but should work well enough for our use cases.
+ */
+
+type EventTargetInterface = EventTarget;
+
+const isCaptureEventListener = (
+  options: undefined | AddEventListenerOptions | boolean
+) => (typeof options === 'boolean' ? options : options?.capture ?? false);
+
+// Shim the global EventTarget object
```
justinfagnani, line 18: EventTarget already exists, I think we can extend it. See https://github.com/lit/lit/pull/2309/files#diff-faf9adf12117ea971af406ace7fdbb87f6c19ae279972892b67b5c0c520d61a4R40

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,567 @@
...
+const EventTargetShim = class EventTarget implements EventTargetInterface {
```
justinfagnani, line 19: I think we definitely need some tests for this.

packages/labs/ssr/src/lib/lit-element-renderer.ts
kyubisation: So don't touch `createRenderRoot` 👍 

As for `enqueueUpdate`; How so? It is a private method? Or what do you mean?

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,567 @@
...
+    if (this[kIsBeingDispatched]) {
+      return;
+    }
+    this[kType] = `${type}`;
+    this.#bubbles = !!bubbles;
+    this.#cancelable = !!cancelable;
+  }
+
+  stopImmediatePropagation() {
+    if (!isEvent(this))
```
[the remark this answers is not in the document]
kyubisation: This is basically taken from the Node.js implementation. For future comparison I left it mostly as it is.
But I can trim it down. (If the empty VM were to be abandoned, we could also remove the Event implementation completely.)

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: Yes, but the behavior is different, both with the capture phase and the property states (that are patched in this implementation).
Also it would fail in the empty VM mode.

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: This is indirectly and directly tested in the ssr package. Should I add additional tests in the shim package?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -861,6 +959,38 @@ And the inner template was:
...
+        } else if (host.element) {
+          let slots = elementSlotMap.get(host.element);
+          if (!slots) {
+            slots = new Map();
+            elementSlotMap.set(host.element, slots);
+          }
+          if (!slots.has(op.name)) {
+            const element = new HTMLElement() as HTMLElementShim;
+            element.__eventTargetParent = getLast(renderInfo.eventTargetStack);
+            slots!.set(op.name, element);
```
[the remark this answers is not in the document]
kyubisation: That is intended for unnamed slots. I'll try to add comments for that case.

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,567 @@
...
+      throw new Error('Value of "this" must be of type Event');
+    // Spec mention "stopImmediatePropagation should set both "stop propagation"
+    // and "stop immediate propagation" flags"
+    // cf: from https://dom.spec.whatwg.org/#dom-event-stopimmediatepropagation
+    this.stopPropagation();
+    this[kStop] = true;
+  }
+
+  preventDefault() {
+    if (!isEvent(this))
```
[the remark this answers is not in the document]
kyubisation: I did run `npm run format` and there was no change. But I assume you mean the missing brackets, which I will remove with the checks.

kyubisation:
Over the weekend I realized that this implementation will not work, if a consumer uses their own polyfill for `HTMLElement`. If that is a concern (and I feel like it would be), I would need to adapt the implementation.
A possible approach would be to patch the `HTMLElement`/`LitElement` prototype for the event related methods and apply the logic currently in the shim there.
@justinfagnani @sorvell Thoughts?

packages/labs/ssr/src/lib/lit-element-renderer.ts
sorvell: * `createRenderRoot` calls `adoptStyles` to add styles to the `shadowRoot`. Calling this is either bad or unncessary.
* It would probably be easiest to no-op `enableUpdating`.

kyubisation:
@justinfagnani @sorvell Thanks again for your reviews and time.

As discussed during our last call, I added tests and comments. Please have another look.

Open topics in my mind:

1. How to provide React support? I did a simple experiment with React Context, which could work. But from my limited understanding of React, I don't think it's possible to resolve slot assignment reliably and the implementation could just default to either unnamed slot or the parent custom element.
2. Should there be an event target root object for SSR? This would facilitate using `ContextProvider` with a root object both on client (e.g. document.body) and server.
3. Nesting slots does not seem to work (in Chrome? https://stackblitz.com/edit/stackblitz-starters-nodfpm?file=script.js ) or did I misunderstand the request?
4. Is any documentation lacking/missing? Limitations or usage guide for `@lit/context` in SSR?

kyubisation:
About the slot nesting; I simply forgot about white space with unnamed slots...
Also the event path always travels the full available, slotted path, regardless if it is "rendered" or not: https://stackblitz.com/edit/stackblitz-starters-atybzr?file=index.html

kyubisation:
@justinfagnani @sorvell I have added the discussed tests (and found a bug with the nested slot case) and rebased the branch on main. I have renamed the global property to `litServerRoot` to keep it simple, but I am open to rename this to something else.
Please have another look.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -784,6 +902,25 @@ And the inner template was:
...
+          // EventTarget polyfill uses these values to calculate the
+          // composedPath of a dispatched event.
+          // Note that the event target parent is either the unnamed/named slot
+          // in the parent event target if it exists or the parent event target
+          // if no matching slot exists.
+          const eventTarget = getLast(
+            renderInfo.eventTargetStack
+          ) as HTMLElementWithEventMeta;
+          const slotName = getLast(renderInfo.slotStack);
+          (instance.element as HTMLElementWithEventMeta).__eventTargetParent =
```
sorvell, line 917: Would this mean that in the case the elements are not slotted, a sibling's parent node would be incorrectly set to its `previousElementSibling`?

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -0,0 +1,474 @@
...
+  stopPropagation(): void {
+    this.#propagationStopped = true;
+  }
+
+  get isTrusted(): boolean {
+    return false;
+  }
+};
+
+Object.defineProperties(EventShim.prototype, {
```
sorvell, line 413: Why doesn't this redefine all the class accessors above?

sorvell (changes requested):
Just a couple questions, very close!

packages/labs/ssr-dom-shim/src/test/event-target-shim_test.ts
```diff
@@ -0,0 +1,341 @@
...
+  child.dispatchEvent(new Event(eventName, {bubbles: true}));
+  assert.equal(eventPath, [
+    'parent:0/capture/CAPTURING_PHASE/child:1',
+    'child:1/capture/AT_TARGET/child:1',
+    'child:1/non-capture/AT_TARGET/child:1',
+    'parent:0/non-capture/BUBBLING_PHASE/child:1',
+  ]);
+});
+
+test('should handle dispatched bubbling event with parent', () => {
```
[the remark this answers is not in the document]
kyubisation: Yes, thanks for the catch.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -107,8 +108,24 @@ const patchAnyDirectives = (
...
 
 const templateCache = new Map<TemplateStringsArray, Array<Op>>();
-/**
 
+// This is a map for which slots exist for a given custom element.
+// With a named slot, it is represented as a string with the name
+// and the unnamed slot is represented as undefined.
+const elementSlotMap = new WeakMap<
+  HTMLElement,
+  Map<string | undefined, HTMLElement>
```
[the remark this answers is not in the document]
kyubisation: Yes 👍

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -391,6 +444,20 @@ const getTemplateOpcodes = (result: TemplateResult) => {
...
+          node.parentNode &&
+          isElementNode(node.parentNode) &&
+          node.parentNode.isDefinedCustomElement
+        ) {
+          // When the parent node is a custom element we check for the presence
+          // or absence of the slot attribute. This allows us to track the
+          // event tree with the association of the slot path.
+          ops.push({
+            type: 'slotted-element-open',
+            name: node.attrs.find((a) => a.name === 'slot')?.value,
```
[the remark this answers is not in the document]
kyubisation: Correct. This is represented as `undefined`, i.e. the absence of the slot attribute.
I'll add another comment to the SlottedElementOpenOp type.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -571,10 +645,26 @@ const getTemplateOpcodes = (result: TemplateResult) => {
...
         ops.push({
           type: 'custom-element-close',
         });
+      } else if (node.tagName === 'slot') {
+        ops.push({
+          type: 'slot-element-close',
+        });
+      }
+      if (
+        node.parentNode &&
```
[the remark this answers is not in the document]
kyubisation: Sorry, I'm not sure I fully understand what you mean? I assume you mean adding `isElementNode(node)` to the check?
But what do you mean with `(incorrectly?) used`?

packages/labs/ssr/src/lib/render-value.ts
kyubisation: Sorry, I'm not sure I fully understand what you mean? Could you elaborate?

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: This was adapted from https://github.com/nodejs/node/blob/main/lib/internal/event_target.js
I'm afraid I couldn't tell you the reason.

kyubisation:
@sorvell Thank you for the review. Feel free to contact me via Discord, if you would like to discuss the open topics it in person.

packages/labs/ssr-dom-shim/src/lib/events.ts
sorvell: My bad, I haven't used `defineProperties` to modify properties, but apparently [it does so](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties) for existing properties.

packages/labs/ssr/src/lib/render-value.ts
bennypowers: it will default to eventTarget, which is the result of `getLast(renderInfo.eventTargetStack)` - can that end up being the previous element sibling?

kyubisation: Ah, I think I understand now what is meant.
No, because the entry is removed with the closing of the custom element or slot element down below.

.changeset/wise-donuts-check.md
```diff
@@ -0,0 +1,7 @@
+---
+'@lit-labs/ssr-dom-shim': minor
+'@lit-labs/ssr': minor
+'@lit-labs/ssr-react': patch
+---
+
+Implement SSR event handling
```
justinfagnani, line 7: Can you mention the option that enables this here?

justinfagnani (approved):
Thanks so much for your work an patience here!!

.changeset/wise-donuts-check.md
kyubisation: You mean like this?
```suggestion
Implement SSR event handling and an optional flag `globalThis.litSsrCallConnectedCallback` to call `connectedCallback` during SSR, if set to true.
```

PR #4757, opened 2024-09-04
docs(labs/observers): :memo: add example for resize-controller to readme

lloydrichards:
## Goals/Scope

I recently was trying out `ResizeController` while learning `lit` and struggled hard to find any current examples on how to use it.  The only example I found was from the youtube channel posted last year where they had a link to a [lit playground example](https://lit.dev/playground/#gist=97aea25901d5dd66c83f19d48ba6de77), but unfortunately this doesn't seem to work.

In an effort to help others learning `lit` and web components, I'm proposing additional examples to the `lit-labs/observers` documentation to make it easier to learn.

## Description

Have added a simple minimal example on how the to extract the width and height of the component using the `.getBoundingClientRect()` instead of `.clientWidth` like in the previous example.

As I've only worked up until now with the `ResizeController`, I've only added this new example, but maybe this PR could encourage others to add additional examples for `IntersectionObserver` and `PerformanceObserver` too.


packages/labs/observers/README.md
```diff
@@ -43,6 +45,32 @@ class MyElement extends LitElement {
...
+
+```ts
+import { ResizeController } from "@lit-labs/observers/resize-controller.js";
+// ...
+
+class MyElement extends LitElement {
+  private _resizeController = new ResizeController(this, {});
+
+  render() {
+    const width = this.getBoundingClientRect().width;
```
sorin-davidoi, line 58: I think the idea with `ResizeObserver` is that we can avoid calling `getBoundingClientRect`, which negatively affects performance due to the layout calculations it triggers.

The dimensions are already provided to `ResizeController`, and they can be extracted with something like this:


```typescript
private _resizeController = new ResizeController<DOMRectReadOnly>(this, {
    // Save into _resizeController.value only the current rect dimentions
    callback: (entries: ResizeObserverEntry[]) => entries[entries.length - 1].contentRect,
  });

render() {
  const width = this._resizeController.value.width;
  const height = this._resizeController.value.height;

  return html`<div class="container">${width}px x${height}px</div>`;
}
```

lloydrichards: Interesting 🤔 will definitely give this a go in my other project.  These are the sorts of examples/documentation that I'm missing I think to better use these packages

sorin-davidoi: > These are the sorts of examples/documentation that I'm missing I think to better use these packages

Agreed, though maybe a better default value for `callback` can also be provided? I would assume that most consumers would be interested in the current dimensions being exposed in `.value` (and not `undefined` as it is the case right now).

lloydrichards: For all my use-cases, I think this could be a better default. But can also imaging it could be confusing if some situation required getting the `borderBoxSize` or `contentBoxSize`.  Happy to test out any new changes that are made 🙏

PR #4758, opened 2024-09-06
Update test dependencies

justinfagnani:
I'm trying to see if updating test dependencies might affect browser timing / performance and help out the virtualizer tests. So far no luck on my local machine.


graynorton (approved):
I had to make another couple of fixes to get this working for me, so I pushed a commit to this branch.

Even though this didn't end up solving the mysteriously failing virtualizer tests, it seems like a good idea to update our test deps, given how stale many of them had gotten, especially since we've gone through the process of doing it now.

graynorton (commented):
Actually, tests seem to be failing in CI...it may not be worth moving forward with this, if it's going to take a lot of troubleshooting to get things back in a stable state.

justinfagnani:
@graynorton I'll try to figure out what's going on. It's some npm issue.

graynorton:
I'm also now seeing four `lit-html` memory leak tests failing on Chromium locally with this branch; I am not familiar with these tests, so not sure whether the failures are significant, or just noise:

image

justinfagnani:
It's possible that Chrome has introduced a memory leak - that's happened before, or it could have changed GC, or increased the memory footprint of elements... With no code changes from us, this failure doesn't indicate that we introduced a leak.

@rictic any guidance here?

PR #4763, opened 2024-09-13
[lit-html] add custom equality support to live

cristinecula:
Added the ability to provide a custom equality function as the second argument to live, to be able to handle cases where a type conversion happens internally.

### DIFF
--- a/packages/lit-html/src/directives/live.ts
+++ b/packages/lit-html/src/directives/live.ts
@@ -14,6 +14,24 @@ import {
 } from '../directive.js';
 import {isSingleExpression, setCommittedValue} from '../directive-helpers.js';
 
+type ValidPartType = 1 | 3 | 4;
+type Element = HTMLElement & {[key: string]: unknown};
+type GetValueFn = (element: Element, name: string) => unknown;
+type EqualFn = (value: unknown, elementValue: unknown) => boolean;
+
+const getValue: Record<ValidPartType, GetValueFn> = {
+  [PartType.PROPERTY]: (element, name) => element[name],
+  [PartType.BOOLEAN_ATTRIBUTE]: (element, name) => element.hasAttribute(name),
+  [PartType.ATTRIBUTE]: (element, name) => element.getAttribute(name),
+};
+
+const defaults: Record<ValidPartType, EqualFn> = {
+  [PartType.PROPERTY]: (value, elementValue) => value === elementValue,
+  [PartType.BOOLEAN_ATTRIBUTE]: (value, elementValue) =>
+    !!value === elementValue,
+  [PartType.ATTRIBUTE]: (value, elementValue) => elementValue === String(value),
+};
+
 class LiveDirective extends Directive {
   constructor(partInfo: PartInfo) {
     super(partInfo);
@@ -33,31 +51,25 @@ class LiveDirective extends Directive {
     }
   }
 
-  render(value: unknown) {
+  render(value: unknown, _equal?: EqualFn) {
     return value;
   }
 
-  override update(part: AttributePart, [value]: DirectiveParameters<this>) {
+  override update(
+    part: AttributePart & {
+      readonly type: ValidPartType;
+    },
+    [value, equal]: DirectiveParameters<this>
+  ) {
     if (value === noChange || value === nothing) {
       return value;
     }
-    const element = part.element;
+    const element = part.element as Element;
     const name = part.name;
+    const _equal = equal ?? defaults[part.type];
 
-    if (part.type === PartType.PROPERTY) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      if (value === (element as any)[name]) {
-        return noChange;
-      }
-    } else if (part.type === PartType.BOOLEAN_ATTRIBUTE) {
-      if (!!value === element.hasAttribute(name)) {
-        return noChange;
-      }
-    } else if (part.type === PartType.ATTRIBUTE) {
-      if (element.getAttribute(name) === String(value)) {
-        return noChange;
-      }
-    }
+    if (_equal(value, getValue[part.type](element, name))) return noChange;
+
     // Resets the part's value, causing its dirty-check to fail so that it
     // always sets the value.
     setCommittedValue(part);
@@ -88,6 +100,14 @@ class LiveDirective extends Directive {
  * `live()` should not be used when the binding will cause a type conversion. If
  * you use `live()` with an attribute binding, make sure that only strings are
  * passed in, or the binding will update every render.
+ *
+ * To customize the equality check, you can pass in a function as the second
+ * argument.
+ *
+ * ```js
+ * const eqAsString = (value, elementValue) => String(value) === elementValue)
+ * html`<input .value=${live(x, eqAsString}>`
+ * ```
  */
 export const live = directive(LiveDirective);
 
--- a/.changeset/wicked-candles-bathe.md
+++ b/.changeset/wicked-candles-bathe.md
@@ -1,0 +1,5 @@
+---
+'lit-html': minor
+---
+
+Added ability to customize the live equality function
--- a/packages/lit-html/src/test/directives/live_test.ts
+++ b/packages/lit-html/src/test/directives/live_test.ts
@@ -16,8 +16,8 @@ class LiveTester extends HTMLElement {
     return this._x;
   }
 
-  set x(v: string | undefined) {
-    this._x = v;
+  set x(v: unknown) {
+    this._x = String(v);
     this._setCount++;
   }
 }
@@ -212,6 +212,37 @@ suite('live directive', () => {
       await new Promise((resolve) => setTimeout(resolve, 0));
       assert.equal(el.getAttribute('hidden'), null);
       assert.equal(mutationCount, 1);
+    });
+  });
+
+  suite('custom equality', () => {
+    test('sets a non-changed property when there is a type conversion', () => {
+      const go = (x: number) =>
+        render(html`<live-tester .x="${live(x)}"></live-tester>}`, container);
+
+      go(1);
+      const el = container.firstElementChild as LiveTester;
+      assert.equal(el.x, '1');
+      assert.equal(el._setCount, 1);
+      go(1);
+      assert.equal(el.x, '1');
+      assert.equal(el._setCount, 2);
+    });
+
+    test('customize equality to handle type conversion', () => {
+      const eqAsString = (a: unknown, b: unknown) => String(a) === b;
+      const go = (x: number) =>
+        render(
+          html`<live-tester .x="${live(x, eqAsString)}"></live-tester>}`,
+          container
+        );
+      go(1);
+      const el = container.firstElementChild as LiveTester;
+      assert.equal(el.x, '1');
+      assert.equal(el._setCount, 1);
+      go(1);
+      assert.equal(el.x, '1');
+      assert.equal(el._setCount, 1);
     });
   });
 });


justinfagnani:
@cristinecula thanks for the PR

Was there an issue yet for this design? At a glance, this looks like a nice change, but we do ask that we have an issue or RFC to discuss additions before PRs: https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

I'll take a look at the change soon!

cristinecula:
Hi, @justinfagnani!

It's an issue I ran into during my work. Sorry, I should have reported it first. The gist of it is this:

I have a number input and I store it's value as a number. I use `live` to bind the value back to the `value` property. Because there is a type mismatch, `live` will update the property all the time. This causes UX issues when users try to edit decimals.

You can test it in this [reproduction](https://lit.dev/playground/#sample=examples%2Fproperties-has-changed&project=W3sibmFtZSI6Im15LWVsZW1lbnQudHMiLCJjb250ZW50IjoiaW1wb3J0IHsgTGl0RWxlbWVudCwgaHRtbH0gZnJvbSBcImxpdFwiO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50LCBwcm9wZXJ0eX0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuaW1wb3J0IHtsaXZlfSBmcm9tICdsaXQvZGlyZWN0aXZlcy9saXZlLmpzJztcblxuQGN1c3RvbUVsZW1lbnQoJ215LWVsZW1lbnQnKVxuY2xhc3MgTXlFbGVtZW50IGV4dGVuZHMgTGl0RWxlbWVudCB7XG5cbiAgQHByb3BlcnR5KCkgcHJpY2U_OiBudW1iZXI7XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYFxuICAgICAgPHA-U2V0IGEgcHJpY2U6XG4gICAgICA8aW5wdXQgdHlwZT1cIm51bWJlclwiIC52YWx1ZT0ke2xpdmUodGhpcy5wcmljZSl9IEBpbnB1dD0ke3RoaXMuX3ByaWNlQ2hhbmdlZH0-PC9wPlxuICAgICAgPHA-UHJpY2UgY2hvc2VuOiAke3RoaXMucHJpY2V9PC9wPlxuICAgIGA7XG4gIH1cblxuICBfcHJpY2VDaGFuZ2VkKGU6IEV2ZW50KSB7XG4gICAgY29uc3QgcHJpY2UgPSAoZS50YXJnZXQgYXMgSFRNTElucHV0RWxlbWVudCkudmFsdWVBc051bWJlcjtcbiAgICBpZiAocHJpY2UpIHtcbiAgICAgIHRoaXMucHJpY2UgPSBwcmljZTtcbiAgICB9XG4gIH1cbn1cblxuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IWRvY3R5cGUgaHRtbD5cbjxodG1sPlxuICA8aGVhZD5cbiAgICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL215LWVsZW1lbnQuanNcIj48L3NjcmlwdD5cbiAgICA8c3R5bGU-XG4gICAgICBib2R5IHtcbiAgICAgICAgZm9udC1mYW1pbHk6IHNhbnMtc2VyaWYnXG4gICAgICB9XG4gICAgPC9zdHlsZT5cbiAgPC9oZWFkPlxuICA8Ym9keT5cbiAgICA8bXktZWxlbWVudD48L215LWVsZW1lbnQ-XG4gIDwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4zLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl40LjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMy4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfV0).

Reproduction steps:
1. type 1.5 in the number input 
2. attempt to change the value to 1.2, by deleting 5 with backspace and typing 2
3. observe that the input now says 21 instead of 1.2 (on chromium, on firefox it will be 12)

My addition to live would allow me to customize the equality, such that instead of strict equality, I also do a type conversion to string, thus preventing lit from unnecessarily updating the value property.

---

While writing the issue description and reproduction gists, I've come to realize that I can bind to `valueAsNumber`, which [simply works out of the box](https://lit.dev/playground/#sample=examples%2Fproperties-has-changed&project=W3sibmFtZSI6Im15LWVsZW1lbnQudHMiLCJjb250ZW50IjoiaW1wb3J0IHsgTGl0RWxlbWVudCwgaHRtbH0gZnJvbSBcImxpdFwiO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50LCBwcm9wZXJ0eX0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuaW1wb3J0IHtsaXZlfSBmcm9tICdsaXQvZGlyZWN0aXZlcy9saXZlLmpzJztcblxuQGN1c3RvbUVsZW1lbnQoJ215LWVsZW1lbnQnKVxuY2xhc3MgTXlFbGVtZW50IGV4dGVuZHMgTGl0RWxlbWVudCB7XG5cbiAgQHByb3BlcnR5KCkgcHJpY2U_OiBudW1iZXI7XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYFxuICAgICAgPHA-U2V0IGEgcHJpY2U6XG4gICAgICA8aW5wdXQgdHlwZT1cIm51bWJlclwiIC52YWx1ZUFzTnVtYmVyPSR7bGl2ZSh0aGlzLnByaWNlKX0gQGlucHV0PSR7dGhpcy5fcHJpY2VDaGFuZ2VkfT48L3A-XG4gICAgICA8cD5QcmljZSBjaG9zZW46ICR7dGhpcy5wcmljZX08L3A-XG4gICAgYDtcbiAgfVxuXG4gIF9wcmljZUNoYW5nZWQoZTogRXZlbnQpIHtcbiAgICBjb25zdCBwcmljZSA9IChlLnRhcmdldCBhcyBIVE1MSW5wdXRFbGVtZW50KS52YWx1ZUFzTnVtYmVyO1xuICAgIGlmIChwcmljZSkge1xuICAgICAgdGhpcy5wcmljZSA9IHByaWNlO1xuICAgIH1cbiAgfVxufVxuXG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhZG9jdHlwZSBodG1sPlxuPGh0bWw-XG4gIDxoZWFkPlxuICAgIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vbXktZWxlbWVudC5qc1wiPjwvc2NyaXB0PlxuICAgIDxzdHlsZT5cbiAgICAgIGJvZHkge1xuICAgICAgICBmb250LWZhbWlseTogc2Fucy1zZXJpZidcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuICA8L2hlYWQ-XG4gIDxib2R5PlxuICAgIDxteS1lbGVtZW50PjwvbXktZWxlbWVudD5cbiAgPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ). This fixes my original issue, rendering this PR unnecessary. I guess it could still be useful in some situations. What do you think?

cristinecula:
Here is an example where having a custom live equality function makes a lot of sense:

When you want to store a number as fixed point notation.

[Playground](https://lit.dev/playground/#sample=examples%2Fproperties-has-changed&project=W3sibmFtZSI6Im15LWVsZW1lbnQudHMiLCJjb250ZW50IjoiaW1wb3J0IHsgTGl0RWxlbWVudCwgaHRtbH0gZnJvbSBcImxpdFwiO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50LCBwcm9wZXJ0eX0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuaW1wb3J0IHtsaXZlfSBmcm9tICcuL2xpdmUuanMnO1xuXG5AY3VzdG9tRWxlbWVudCgnbXktZWxlbWVudCcpXG5jbGFzcyBNeUVsZW1lbnQgZXh0ZW5kcyBMaXRFbGVtZW50IHtcblxuICBAcHJvcGVydHkoKSBwcmljZT86IHN0cmluZztcblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgXG4gICAgICA8cD5TZXQgYSBwcmljZTpcbiAgICAgICAgPGlucHV0IHR5cGU9XCJudW1iZXJcIiAudmFsdWU9JHtsaXZlKHRoaXMucHJpY2UpfSBAaW5wdXQ9JHt0aGlzLl9wcmljZUNoYW5nZWR9PjwvcD5cbiAgICAgICA8cD5TZXQgYSBwcmljZSAoY3VzdG9tKTpcbiAgICAgICAgPGlucHV0IHR5cGU9XCJudW1iZXJcIiAudmFsdWU9JHtsaXZlKHRoaXMucHJpY2UsICh2YWx1ZTogc3RyaW5nLCBlbGVtZW50VmFsdWU6IHN0cmluZykgPT4gcGFyc2VGbG9hdCh2YWx1ZSkgPT09IHBhcnNlRmxvYXQoZWxlbWVudFZhbHVlKSl9IEBpbnB1dD0ke3RoaXMuX3ByaWNlQ2hhbmdlZH0-PC9wPlxuICAgICAgPHA-UHJpY2UgY2hvc2VuOiAke3RoaXMucHJpY2V9PC9wPlxuICAgIGA7XG4gIH1cblxuICBfcHJpY2VDaGFuZ2VkKGU6IEV2ZW50KSB7XG4gICAgY29uc3QgcHJpY2UgPSAoZS50YXJnZXQgYXMgSFRNTElucHV0RWxlbWVudCkudmFsdWVBc051bWJlcjtcbiAgICAgdGhpcy5wcmljZSA9IHByaWNlLnRvRml4ZWQoMyk7XG4gIH1cbn1cblxuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IWRvY3R5cGUgaHRtbD5cbjxodG1sPlxuICA8aGVhZD5cbiAgICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL215LWVsZW1lbnQuanNcIj48L3NjcmlwdD5cbiAgICA8c3R5bGU-XG4gICAgICBib2R5IHtcbiAgICAgICAgZm9udC1mYW1pbHk6IHNhbnMtc2VyaWYnXG4gICAgICB9XG4gICAgPC9zdHlsZT5cbiAgPC9oZWFkPlxuICA8Ym9keT5cbiAgICA8bXktZWxlbWVudD48L215LWVsZW1lbnQ-XG4gIDwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4zLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl40LjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMy4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfSx7Im5hbWUiOiJsaXZlLnRzIiwiY29udGVudCI6Ii8qKlxuICogQGxpY2Vuc2VcbiAqIENvcHlyaWdodCAyMDIwIEdvb2dsZSBMTENcbiAqIFNQRFgtTGljZW5zZS1JZGVudGlmaWVyOiBCU0QtMy1DbGF1c2VcbiAqL1xuXG5pbXBvcnQge0F0dHJpYnV0ZVBhcnQsIG5vQ2hhbmdlLCBub3RoaW5nfSBmcm9tICdsaXQtaHRtbCc7XG5pbXBvcnQge1xuICBkaXJlY3RpdmUsXG4gIERpcmVjdGl2ZSxcbiAgRGlyZWN0aXZlUGFyYW1ldGVycyxcbiAgUGFydEluZm8sXG4gIFBhcnRUeXBlLFxufSBmcm9tICdsaXQtaHRtbC9kaXJlY3RpdmUuanMnO1xuaW1wb3J0IHtpc1NpbmdsZUV4cHJlc3Npb24sIHNldENvbW1pdHRlZFZhbHVlfSBmcm9tICdsaXQtaHRtbC9kaXJlY3RpdmUtaGVscGVycy5qcyc7XG5cbnR5cGUgVmFsaWRQYXJ0VHlwZSA9IDEgfCAzIHwgNDtcbnR5cGUgRWxlbWVudCA9IEhUTUxFbGVtZW50ICYge1trZXk6IHN0cmluZ106IHVua25vd259O1xudHlwZSBHZXRWYWx1ZUZuID0gKGVsZW1lbnQ6IEVsZW1lbnQsIG5hbWU6IHN0cmluZykgPT4gdW5rbm93bjtcbnR5cGUgRXF1YWxGbiA9ICh2YWx1ZTogdW5rbm93biwgZWxlbWVudFZhbHVlOiB1bmtub3duKSA9PiBib29sZWFuO1xuXG5jb25zdCBnZXRWYWx1ZTogUmVjb3JkPFZhbGlkUGFydFR5cGUsIEdldFZhbHVlRm4-ID0ge1xuICBbUGFydFR5cGUuUFJPUEVSVFldOiAoZWxlbWVudCwgbmFtZSkgPT4gZWxlbWVudFtuYW1lXSxcbiAgW1BhcnRUeXBlLkJPT0xFQU5fQVRUUklCVVRFXTogKGVsZW1lbnQsIG5hbWUpID0-IGVsZW1lbnQuaGFzQXR0cmlidXRlKG5hbWUpLFxuICBbUGFydFR5cGUuQVRUUklCVVRFXTogKGVsZW1lbnQsIG5hbWUpID0-IGVsZW1lbnQuZ2V0QXR0cmlidXRlKG5hbWUpLFxufTtcblxuY29uc3QgZGVmYXVsdHM6IFJlY29yZDxWYWxpZFBhcnRUeXBlLCBFcXVhbEZuPiA9IHtcbiAgW1BhcnRUeXBlLlBST1BFUlRZXTogKHZhbHVlLCBlbGVtZW50VmFsdWUpID0-IHZhbHVlID09PSBlbGVtZW50VmFsdWUsXG4gIFtQYXJ0VHlwZS5CT09MRUFOX0FUVFJJQlVURV06ICh2YWx1ZSwgZWxlbWVudFZhbHVlKSA9PlxuICAgICEhdmFsdWUgPT09IGVsZW1lbnRWYWx1ZSxcbiAgW1BhcnRUeXBlLkFUVFJJQlVURV06ICh2YWx1ZSwgZWxlbWVudFZhbHVlKSA9PiBlbGVtZW50VmFsdWUgPT09IFN0cmluZyh2YWx1ZSksXG59O1xuXG5jbGFzcyBMaXZlRGlyZWN0aXZlIGV4dGVuZHMgRGlyZWN0aXZlIHtcbiAgY29uc3RydWN0b3IocGFydEluZm86IFBhcnRJbmZvKSB7XG4gICAgc3VwZXIocGFydEluZm8pO1xuICAgIGlmIChcbiAgICAgICEoXG4gICAgICAgIHBhcnRJbmZvLnR5cGUgPT09IFBhcnRUeXBlLlBST1BFUlRZIHx8XG4gICAgICAgIHBhcnRJbmZvLnR5cGUgPT09IFBhcnRUeXBlLkFUVFJJQlVURSB8fFxuICAgICAgICBwYXJ0SW5mby50eXBlID09PSBQYXJ0VHlwZS5CT09MRUFOX0FUVFJJQlVURVxuICAgICAgKVxuICAgICkge1xuICAgICAgdGhyb3cgbmV3IEVycm9yKFxuICAgICAgICAnVGhlIGBsaXZlYCBkaXJlY3RpdmUgaXMgbm90IGFsbG93ZWQgb24gY2hpbGQgb3IgZXZlbnQgYmluZGluZ3MnXG4gICAgICApO1xuICAgIH1cbiAgICBpZiAoIWlzU2luZ2xlRXhwcmVzc2lvbihwYXJ0SW5mbykpIHtcbiAgICAgIHRocm93IG5ldyBFcnJvcignYGxpdmVgIGJpbmRpbmdzIGNhbiBvbmx5IGNvbnRhaW4gYSBzaW5nbGUgZXhwcmVzc2lvbicpO1xuICAgIH1cbiAgfVxuXG4gIHJlbmRlcih2YWx1ZTogdW5rbm93biwgX2VxdWFsPzogRXF1YWxGbikge1xuICAgIHJldHVybiB2YWx1ZTtcbiAgfVxuXG4gIG92ZXJyaWRlIHVwZGF0ZShcbiAgICBwYXJ0OiBBdHRyaWJ1dGVQYXJ0ICYge1xuICAgICAgcmVhZG9ubHkgdHlwZTogVmFsaWRQYXJ0VHlwZTtcbiAgICB9LFxuICAgIFt2YWx1ZSwgZXF1YWxdOiBEaXJlY3RpdmVQYXJhbWV0ZXJzPHRoaXM-XG4gICkge1xuICAgIGlmICh2YWx1ZSA9PT0gbm9DaGFuZ2UgfHwgdmFsdWUgPT09IG5vdGhpbmcpIHtcbiAgICAgIHJldHVybiB2YWx1ZTtcbiAgICB9XG4gICAgY29uc3QgZWxlbWVudCA9IHBhcnQuZWxlbWVudCBhcyBFbGVtZW50O1xuICAgIGNvbnN0IG5hbWUgPSBwYXJ0Lm5hbWU7XG4gICAgY29uc3QgX2VxdWFsID0gZXF1YWwgPz8gZGVmYXVsdHNbcGFydC50eXBlXTtcbiAgICBcbiAgICBcbiAgICBjb25zb2xlLmxvZygndXBkYXRlJywgdmFsdWUpO1xuXG4gICAgaWYgKF9lcXVhbCh2YWx1ZSwgZ2V0VmFsdWVbcGFydC50eXBlXShlbGVtZW50LCBuYW1lKSkpIHJldHVybiBub0NoYW5nZTtcblxuICAgIC8vIFJlc2V0cyB0aGUgcGFydCdzIHZhbHVlLCBjYXVzaW5nIGl0cyBkaXJ0eS1jaGVjayB0byBmYWlsIHNvIHRoYXQgaXRcbiAgICAvLyBhbHdheXMgc2V0cyB0aGUgdmFsdWUuXG4gICAgY29uc29sZS5sb2coJ3NldENvbW1pdHRlZFZhbHVlJywgdmFsdWUpO1xuICAgIHNldENvbW1pdHRlZFZhbHVlKHBhcnQpO1xuICAgIHJldHVybiB2YWx1ZTtcbiAgfVxufVxuXG4vKipcbiAqIENoZWNrcyBiaW5kaW5nIHZhbHVlcyBhZ2FpbnN0IGxpdmUgRE9NIHZhbHVlcywgaW5zdGVhZCBvZiBwcmV2aW91c2x5IGJvdW5kXG4gKiB2YWx1ZXMsIHdoZW4gZGV0ZXJtaW5pbmcgd2hldGhlciB0byB1cGRhdGUgdGhlIHZhbHVlLlxuICpcbiAqIFRoaXMgaXMgdXNlZnVsIGZvciBjYXNlcyB3aGVyZSB0aGUgRE9NIHZhbHVlIG1heSBjaGFuZ2UgZnJvbSBvdXRzaWRlIG9mXG4gKiBsaXQtaHRtbCwgc3VjaCBhcyB3aXRoIGEgYmluZGluZyB0byBhbiBgPGlucHV0PmAgZWxlbWVudCdzIGB2YWx1ZWAgcHJvcGVydHksXG4gKiBhIGNvbnRlbnQgZWRpdGFibGUgZWxlbWVudHMgdGV4dCwgb3IgdG8gYSBjdXN0b20gZWxlbWVudCB0aGF0IGNoYW5nZXMgaXQnc1xuICogb3duIHByb3BlcnRpZXMgb3IgYXR0cmlidXRlcy5cbiAqXG4gKiBJbiB0aGVzZSBjYXNlcyBpZiB0aGUgRE9NIHZhbHVlIGNoYW5nZXMsIGJ1dCB0aGUgdmFsdWUgc2V0IHRocm91Z2ggbGl0LWh0bWxcbiAqIGJpbmRpbmdzIGhhc24ndCwgbGl0LWh0bWwgd29uJ3Qga25vdyB0byB1cGRhdGUgdGhlIERPTSB2YWx1ZSBhbmQgd2lsbCBsZWF2ZVxuICogaXQgYWxvbmUuIElmIHRoaXMgaXMgbm90IHdoYXQgeW91IHdhbnQtLWlmIHlvdSB3YW50IHRvIG92ZXJ3cml0ZSB0aGUgRE9NXG4gKiB2YWx1ZSB3aXRoIHRoZSBib3VuZCB2YWx1ZSBubyBtYXR0ZXIgd2hhdC0tdXNlIHRoZSBgbGl2ZSgpYCBkaXJlY3RpdmU6XG4gKlxuICogYGBganNcbiAqIGh0bWxgPGlucHV0IC52YWx1ZT0ke2xpdmUoeCl9PmBcbiAqIGBgYFxuICpcbiAqIGBsaXZlKClgIHBlcmZvcm1zIGEgc3RyaWN0IGVxdWFsaXR5IGNoZWNrIGFnYWluc3QgdGhlIGxpdmUgRE9NIHZhbHVlLCBhbmQgaWZcbiAqIHRoZSBuZXcgdmFsdWUgaXMgZXF1YWwgdG8gdGhlIGxpdmUgdmFsdWUsIGRvZXMgbm90aGluZy4gVGhpcyBtZWFucyB0aGF0XG4gKiBgbGl2ZSgpYCBzaG91bGQgbm90IGJlIHVzZWQgd2hlbiB0aGUgYmluZGluZyB3aWxsIGNhdXNlIGEgdHlwZSBjb252ZXJzaW9uLiBJZlxuICogeW91IHVzZSBgbGl2ZSgpYCB3aXRoIGFuIGF0dHJpYnV0ZSBiaW5kaW5nLCBtYWtlIHN1cmUgdGhhdCBvbmx5IHN0cmluZ3MgYXJlXG4gKiBwYXNzZWQgaW4sIG9yIHRoZSBiaW5kaW5nIHdpbGwgdXBkYXRlIGV2ZXJ5IHJlbmRlci5cbiAqXG4gKiBUbyBjdXN0b21pemUgdGhlIGVxdWFsaXR5IGNoZWNrLCB5b3UgY2FuIHBhc3MgaW4gYSBmdW5jdGlvbiBhcyB0aGUgc2Vjb25kXG4gKiBhcmd1bWVudC5cbiAqXG4gKiBgYGBqc1xuICogY29uc3QgZXFBc1N0cmluZyA9ICh2YWx1ZSwgZWxlbWVudFZhbHVlKSA9PiBTdHJpbmcodmFsdWUpID09PSBlbGVtZW50VmFsdWUpXG4gKiBodG1sYDxpbnB1dCAudmFsdWU9JHtsaXZlKHgsIGVxQXNTdHJpbmd9PmBcbiAqIGBgYFxuICovXG5leHBvcnQgY29uc3QgbGl2ZSA9IGRpcmVjdGl2ZShMaXZlRGlyZWN0aXZlKTtcblxuLyoqXG4gKiBUaGUgdHlwZSBvZiB0aGUgY2xhc3MgdGhhdCBwb3dlcnMgdGhpcyBkaXJlY3RpdmUuIE5lY2Vzc2FyeSBmb3IgbmFtaW5nIHRoZVxuICogZGlyZWN0aXZlJ3MgcmV0dXJuIHR5cGUuXG4gKi9cbmV4cG9ydCB0eXBlIHtMaXZlRGlyZWN0aXZlfTsifV0)

Try writing 123 in both inputs.

justinfagnani:
Thanks for the details @cristinecula. These are the things that would be great in an issue. You can even open an issue and PR at the some time so the issue can be the point of reference if there are changes in how we solve something. Like if we decide this is working fine, people might find that in a closed issue, but not a closed PR.

As for whether we still need this change... tough call. In general I like the feature, but I'm by default slightly hesitant to add more code w/o a clear use case example. Thanks for the second example. It behaves a little funny because of the rounding behavior of toFixed. If you type in 1.23456 it'll get rounded to 1.235 and from there any keystroke on 5 or greater will just increment the number by .001 :)

But I think the important takeaway focus management. live() gives us the ability to overwrite the binding value when Lit doesn't think it's necessary. This change gives us more ability to _not_ do that, which I don't think is usually a big deal - setting a property sometimes shouldn't matter much, but in input cases it can preserve focus when you don't set a new value, which is important.

I think it might be better to actually have more customizable focus management for inputs - like to restore the cursor to a specific index, but that seems more than what live() is responsible for.

So I think this might be useful, and it's not much code and not in core. I'll get to reviewing it now 👍

packages/lit-html/src/directives/live.ts
```diff
@@ -33,31 +51,25 @@ class LiveDirective extends Directive {
...
+  ) {
     if (value === noChange || value === nothing) {
       return value;
     }
-    const element = part.element;
+    const element = part.element as Element;
     const name = part.name;
+    const _equal = equal ?? defaults[part.type];
+
+    if (_equal(value, getValue[part.type](element, name))) return noChange;
```
justinfagnani, line 71: I think we can inline the getValue function to avoid some property lookups:

```suggestion
    const {type, name, element} = part;
    
    const currentValue =
      type === PartType.PROPERTY
        ? element[name as keyof typeof element]
        : type === PartType.ATTRIBUTE
          ? element.getAttribute(name)
          : element.hasAttribute(name);
    if (equal(value, currentValue)) {
      return noChange;
    }
```

packages/lit-html/src/directives/live.ts
```diff
@@ -33,31 +51,25 @@ class LiveDirective extends Directive {
...
+  render(value: unknown, _equal?: EqualFn) {
     return value;
   }
 
-  override update(part: AttributePart, [value]: DirectiveParameters<this>) {
+  override update(
+    part: AttributePart & {
+      readonly type: ValidPartType;
+    },
+    [value, equal]: DirectiveParameters<this>
```
justinfagnani, line 62: Defaults can be set in the destructuring parameter:

```suggestion
    [value, equal = defaults[part.type]]: DirectiveParameters<this>
```

packages/lit-html/src/directives/live.ts
```diff
@@ -88,6 +100,14 @@ class LiveDirective extends Directive {
  * `live()` should not be used when the binding will cause a type conversion. If
  * you use `live()` with an attribute binding, make sure that only strings are
  * passed in, or the binding will update every render.
+ *
+ * To customize the equality check, you can pass in a function as the second
```
justinfagnani, line 104: I think the previous paragraph needs to be updated a bit to refer to that as default behavior.

packages/lit-html/src/directives/live.ts
```diff
@@ -14,6 +14,24 @@ import {
 } from '../directive.js';
 import {isSingleExpression, setCommittedValue} from '../directive-helpers.js';
 
+type ValidPartType = 1 | 3 | 4;
+type Element = HTMLElement & {[key: string]: unknown};
```
justinfagnani, line 18: I don't think we should override the global `Element` type, and this type could remove a lot of type-checking in the whole file. We can cast at the access site.

justinfagnani (changes requested):

packages/lit-html/src/directives/live.ts
cristinecula: Updated.

packages/lit-html/src/directives/live.ts
cristinecula: Updated.

packages/lit-html/src/directives/live.ts
cristinecula: Updated. I'm not sure if the wording is clear enough, but I was not sure how to make it better.

cristinecula:
Thank you for the thorough analysis and feedback! I've updated the PR with the suggested changes.

That rounding input behavior is indeed funny. You have a keen eye to catch that!
It was actually a bug in the equality function. [This updated version](https://lit.dev/playground/#sample=examples%2Fproperties-has-changed&project=W3sibmFtZSI6Im15LWVsZW1lbnQudHMiLCJjb250ZW50IjoiaW1wb3J0IHsgTGl0RWxlbWVudCwgaHRtbH0gZnJvbSBcImxpdFwiO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50LCBwcm9wZXJ0eX0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuaW1wb3J0IHtsaXZlfSBmcm9tICcuL2xpdmUuanMnO1xuXG5AY3VzdG9tRWxlbWVudCgnbXktZWxlbWVudCcpXG5jbGFzcyBNeUVsZW1lbnQgZXh0ZW5kcyBMaXRFbGVtZW50IHtcblxuICBAcHJvcGVydHkoKSBwcmljZT86IHN0cmluZztcblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgXG4gICAgICA8cD5TZXQgYSBwcmljZTpcbiAgICAgICAgPGlucHV0IHR5cGU9XCJudW1iZXJcIiAudmFsdWU9JHtsaXZlKHRoaXMucHJpY2UpfSBAaW5wdXQ9JHt0aGlzLl9wcmljZUNoYW5nZWR9PjwvcD5cbiAgICAgICA8cD5TZXQgYSBwcmljZSAoY3VzdG9tKTpcbiAgICAgICAgPGlucHV0IHR5cGU9XCJudW1iZXJcIiAudmFsdWU9JHtsaXZlKHRoaXMucHJpY2UsICh2YWx1ZTogc3RyaW5nLCBlbGVtZW50VmFsdWU6IHN0cmluZykgPT4gdmFsdWUgPT09IHBhcnNlRmxvYXQoZWxlbWVudFZhbHVlKS50b0ZpeGVkKDMpKX0gQGlucHV0PSR7dGhpcy5fcHJpY2VDaGFuZ2VkfT48L3A-XG4gICAgICA8cD5QcmljZSBjaG9zZW46ICR7dGhpcy5wcmljZX08L3A-XG4gICAgYDtcbiAgfVxuXG4gIF9wcmljZUNoYW5nZWQoZTogRXZlbnQpIHtcbiAgICBjb25zdCBwcmljZSA9IChlLnRhcmdldCBhcyBIVE1MSW5wdXRFbGVtZW50KS52YWx1ZUFzTnVtYmVyO1xuICAgICB0aGlzLnByaWNlID0gcHJpY2UudG9GaXhlZCgzKTtcbiAgfVxufVxuXG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhZG9jdHlwZSBodG1sPlxuPGh0bWw-XG4gIDxoZWFkPlxuICAgIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vbXktZWxlbWVudC5qc1wiPjwvc2NyaXB0PlxuICAgIDxzdHlsZT5cbiAgICAgIGJvZHkge1xuICAgICAgICBmb250LWZhbWlseTogc2Fucy1zZXJpZidcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuICA8L2hlYWQ-XG4gIDxib2R5PlxuICAgIDxteS1lbGVtZW50PjwvbXktZWxlbWVudD5cbiAgPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9LHsibmFtZSI6ImxpdmUudHMiLCJjb250ZW50IjoiLyoqXG4gKiBAbGljZW5zZVxuICogQ29weXJpZ2h0IDIwMjAgR29vZ2xlIExMQ1xuICogU1BEWC1MaWNlbnNlLUlkZW50aWZpZXI6IEJTRC0zLUNsYXVzZVxuICovXG5cbmltcG9ydCB7QXR0cmlidXRlUGFydCwgbm9DaGFuZ2UsIG5vdGhpbmd9IGZyb20gJ2xpdC1odG1sJztcbmltcG9ydCB7XG4gIGRpcmVjdGl2ZSxcbiAgRGlyZWN0aXZlLFxuICBEaXJlY3RpdmVQYXJhbWV0ZXJzLFxuICBQYXJ0SW5mbyxcbiAgUGFydFR5cGUsXG59IGZyb20gJ2xpdC1odG1sL2RpcmVjdGl2ZS5qcyc7XG5pbXBvcnQge2lzU2luZ2xlRXhwcmVzc2lvbiwgc2V0Q29tbWl0dGVkVmFsdWV9IGZyb20gJ2xpdC1odG1sL2RpcmVjdGl2ZS1oZWxwZXJzLmpzJztcblxudHlwZSBWYWxpZFBhcnRUeXBlID0gMSB8IDMgfCA0O1xudHlwZSBFbGVtZW50ID0gSFRNTEVsZW1lbnQgJiB7W2tleTogc3RyaW5nXTogdW5rbm93bn07XG50eXBlIEdldFZhbHVlRm4gPSAoZWxlbWVudDogRWxlbWVudCwgbmFtZTogc3RyaW5nKSA9PiB1bmtub3duO1xudHlwZSBFcXVhbEZuID0gKHZhbHVlOiB1bmtub3duLCBlbGVtZW50VmFsdWU6IHVua25vd24pID0-IGJvb2xlYW47XG5cbmNvbnN0IGdldFZhbHVlOiBSZWNvcmQ8VmFsaWRQYXJ0VHlwZSwgR2V0VmFsdWVGbj4gPSB7XG4gIFtQYXJ0VHlwZS5QUk9QRVJUWV06IChlbGVtZW50LCBuYW1lKSA9PiBlbGVtZW50W25hbWVdLFxuICBbUGFydFR5cGUuQk9PTEVBTl9BVFRSSUJVVEVdOiAoZWxlbWVudCwgbmFtZSkgPT4gZWxlbWVudC5oYXNBdHRyaWJ1dGUobmFtZSksXG4gIFtQYXJ0VHlwZS5BVFRSSUJVVEVdOiAoZWxlbWVudCwgbmFtZSkgPT4gZWxlbWVudC5nZXRBdHRyaWJ1dGUobmFtZSksXG59O1xuXG5jb25zdCBkZWZhdWx0czogUmVjb3JkPFZhbGlkUGFydFR5cGUsIEVxdWFsRm4-ID0ge1xuICBbUGFydFR5cGUuUFJPUEVSVFldOiAodmFsdWUsIGVsZW1lbnRWYWx1ZSkgPT4gdmFsdWUgPT09IGVsZW1lbnRWYWx1ZSxcbiAgW1BhcnRUeXBlLkJPT0xFQU5fQVRUUklCVVRFXTogKHZhbHVlLCBlbGVtZW50VmFsdWUpID0-XG4gICAgISF2YWx1ZSA9PT0gZWxlbWVudFZhbHVlLFxuICBbUGFydFR5cGUuQVRUUklCVVRFXTogKHZhbHVlLCBlbGVtZW50VmFsdWUpID0-IGVsZW1lbnRWYWx1ZSA9PT0gU3RyaW5nKHZhbHVlKSxcbn07XG5cbmNsYXNzIExpdmVEaXJlY3RpdmUgZXh0ZW5kcyBEaXJlY3RpdmUge1xuICBjb25zdHJ1Y3RvcihwYXJ0SW5mbzogUGFydEluZm8pIHtcbiAgICBzdXBlcihwYXJ0SW5mbyk7XG4gICAgaWYgKFxuICAgICAgIShcbiAgICAgICAgcGFydEluZm8udHlwZSA9PT0gUGFydFR5cGUuUFJPUEVSVFkgfHxcbiAgICAgICAgcGFydEluZm8udHlwZSA9PT0gUGFydFR5cGUuQVRUUklCVVRFIHx8XG4gICAgICAgIHBhcnRJbmZvLnR5cGUgPT09IFBhcnRUeXBlLkJPT0xFQU5fQVRUUklCVVRFXG4gICAgICApXG4gICAgKSB7XG4gICAgICB0aHJvdyBuZXcgRXJyb3IoXG4gICAgICAgICdUaGUgYGxpdmVgIGRpcmVjdGl2ZSBpcyBub3QgYWxsb3dlZCBvbiBjaGlsZCBvciBldmVudCBiaW5kaW5ncydcbiAgICAgICk7XG4gICAgfVxuICAgIGlmICghaXNTaW5nbGVFeHByZXNzaW9uKHBhcnRJbmZvKSkge1xuICAgICAgdGhyb3cgbmV3IEVycm9yKCdgbGl2ZWAgYmluZGluZ3MgY2FuIG9ubHkgY29udGFpbiBhIHNpbmdsZSBleHByZXNzaW9uJyk7XG4gICAgfVxuICB9XG5cbiAgcmVuZGVyKHZhbHVlOiB1bmtub3duLCBfZXF1YWw_OiBFcXVhbEZuKSB7XG4gICAgcmV0dXJuIHZhbHVlO1xuICB9XG5cbiAgb3ZlcnJpZGUgdXBkYXRlKFxuICAgIHBhcnQ6IEF0dHJpYnV0ZVBhcnQgJiB7XG4gICAgICByZWFkb25seSB0eXBlOiBWYWxpZFBhcnRUeXBlO1xuICAgIH0sXG4gICAgW3ZhbHVlLCBlcXVhbF06IERpcmVjdGl2ZVBhcmFtZXRlcnM8dGhpcz5cbiAgKSB7XG4gICAgaWYgKHZhbHVlID09PSBub0NoYW5nZSB8fCB2YWx1ZSA9PT0gbm90aGluZykge1xuICAgICAgcmV0dXJuIHZhbHVlO1xuICAgIH1cbiAgICBjb25zdCBlbGVtZW50ID0gcGFydC5lbGVtZW50IGFzIEVsZW1lbnQ7XG4gICAgY29uc3QgbmFtZSA9IHBhcnQubmFtZTtcbiAgICBjb25zdCBfZXF1YWwgPSBlcXVhbCA_PyBkZWZhdWx0c1twYXJ0LnR5cGVdO1xuICAgIFxuICAgIFxuICAgIGNvbnNvbGUubG9nKCd1cGRhdGUnLCB2YWx1ZSk7XG5cbiAgICBpZiAoX2VxdWFsKHZhbHVlLCBnZXRWYWx1ZVtwYXJ0LnR5cGVdKGVsZW1lbnQsIG5hbWUpKSkgcmV0dXJuIG5vQ2hhbmdlO1xuXG4gICAgLy8gUmVzZXRzIHRoZSBwYXJ0J3MgdmFsdWUsIGNhdXNpbmcgaXRzIGRpcnR5LWNoZWNrIHRvIGZhaWwgc28gdGhhdCBpdFxuICAgIC8vIGFsd2F5cyBzZXRzIHRoZSB2YWx1ZS5cbiAgICBjb25zb2xlLmxvZygnc2V0Q29tbWl0dGVkVmFsdWUnLCB2YWx1ZSk7XG4gICAgc2V0Q29tbWl0dGVkVmFsdWUocGFydCk7XG4gICAgcmV0dXJuIHZhbHVlO1xuICB9XG59XG5cbi8qKlxuICogQ2hlY2tzIGJpbmRpbmcgdmFsdWVzIGFnYWluc3QgbGl2ZSBET00gdmFsdWVzLCBpbnN0ZWFkIG9mIHByZXZpb3VzbHkgYm91bmRcbiAqIHZhbHVlcywgd2hlbiBkZXRlcm1pbmluZyB3aGV0aGVyIHRvIHVwZGF0ZSB0aGUgdmFsdWUuXG4gKlxuICogVGhpcyBpcyB1c2VmdWwgZm9yIGNhc2VzIHdoZXJlIHRoZSBET00gdmFsdWUgbWF5IGNoYW5nZSBmcm9tIG91dHNpZGUgb2ZcbiAqIGxpdC1odG1sLCBzdWNoIGFzIHdpdGggYSBiaW5kaW5nIHRvIGFuIGA8aW5wdXQ-YCBlbGVtZW50J3MgYHZhbHVlYCBwcm9wZXJ0eSxcbiAqIGEgY29udGVudCBlZGl0YWJsZSBlbGVtZW50cyB0ZXh0LCBvciB0byBhIGN1c3RvbSBlbGVtZW50IHRoYXQgY2hhbmdlcyBpdCdzXG4gKiBvd24gcHJvcGVydGllcyBvciBhdHRyaWJ1dGVzLlxuICpcbiAqIEluIHRoZXNlIGNhc2VzIGlmIHRoZSBET00gdmFsdWUgY2hhbmdlcywgYnV0IHRoZSB2YWx1ZSBzZXQgdGhyb3VnaCBsaXQtaHRtbFxuICogYmluZGluZ3MgaGFzbid0LCBsaXQtaHRtbCB3b24ndCBrbm93IHRvIHVwZGF0ZSB0aGUgRE9NIHZhbHVlIGFuZCB3aWxsIGxlYXZlXG4gKiBpdCBhbG9uZS4gSWYgdGhpcyBpcyBub3Qgd2hhdCB5b3Ugd2FudC0taWYgeW91IHdhbnQgdG8gb3ZlcndyaXRlIHRoZSBET01cbiAqIHZhbHVlIHdpdGggdGhlIGJvdW5kIHZhbHVlIG5vIG1hdHRlciB3aGF0LS11c2UgdGhlIGBsaXZlKClgIGRpcmVjdGl2ZTpcbiAqXG4gKiBgYGBqc1xuICogaHRtbGA8aW5wdXQgLnZhbHVlPSR7bGl2ZSh4KX0-YFxuICogYGBgXG4gKlxuICogYGxpdmUoKWAgcGVyZm9ybXMgYSBzdHJpY3QgZXF1YWxpdHkgY2hlY2sgYWdhaW5zdCB0aGUgbGl2ZSBET00gdmFsdWUsIGFuZCBpZlxuICogdGhlIG5ldyB2YWx1ZSBpcyBlcXVhbCB0byB0aGUgbGl2ZSB2YWx1ZSwgZG9lcyBub3RoaW5nLiBUaGlzIG1lYW5zIHRoYXRcbiAqIGBsaXZlKClgIHNob3VsZCBub3QgYmUgdXNlZCB3aGVuIHRoZSBiaW5kaW5nIHdpbGwgY2F1c2UgYSB0eXBlIGNvbnZlcnNpb24uIElmXG4gKiB5b3UgdXNlIGBsaXZlKClgIHdpdGggYW4gYXR0cmlidXRlIGJpbmRpbmcsIG1ha2Ugc3VyZSB0aGF0IG9ubHkgc3RyaW5ncyBhcmVcbiAqIHBhc3NlZCBpbiwgb3IgdGhlIGJpbmRpbmcgd2lsbCB1cGRhdGUgZXZlcnkgcmVuZGVyLlxuICpcbiAqIFRvIGN1c3RvbWl6ZSB0aGUgZXF1YWxpdHkgY2hlY2ssIHlvdSBjYW4gcGFzcyBpbiBhIGZ1bmN0aW9uIGFzIHRoZSBzZWNvbmRcbiAqIGFyZ3VtZW50LlxuICpcbiAqIGBgYGpzXG4gKiBjb25zdCBlcUFzU3RyaW5nID0gKHZhbHVlLCBlbGVtZW50VmFsdWUpID0-IFN0cmluZyh2YWx1ZSkgPT09IGVsZW1lbnRWYWx1ZSlcbiAqIGh0bWxgPGlucHV0IC52YWx1ZT0ke2xpdmUoeCwgZXFBc1N0cmluZ30-YFxuICogYGBgXG4gKi9cbmV4cG9ydCBjb25zdCBsaXZlID0gZGlyZWN0aXZlKExpdmVEaXJlY3RpdmUpO1xuXG4vKipcbiAqIFRoZSB0eXBlIG9mIHRoZSBjbGFzcyB0aGF0IHBvd2VycyB0aGlzIGRpcmVjdGl2ZS4gTmVjZXNzYXJ5IGZvciBuYW1pbmcgdGhlXG4gKiBkaXJlY3RpdmUncyByZXR1cm4gdHlwZS5cbiAqL1xuZXhwb3J0IHR5cGUge0xpdmVEaXJlY3RpdmV9OyJ9XQ) does not have that issue.

I'm not sure if I should go and create an issue referencing this PR or not. Please let me know and I will do it ASAP!

cristinecula:
Hi @justinfagnani!

Do you need me to do any more changes related to this PR?

PR #4773, opened 2024-09-21
[labs/forms] WIP: Initial implementation

justinfagnani:
See RRFC: https://github.com/lit/rfcs/issues/43

This a rough sketch of a no-public-API FormAssociated mixin.


packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,227 @@
...
+) => {
+  class C extends base {
+    #internals: ElementInternals;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
+      super(...args);
+      const internals = this.attachInternals();
+      this.#internals = internals;
+      internalsCache.set(this, internals);
```
sorvell, line 97: This strikes me as a bit cumbersome v. just using a symbol based property and sharing that symbol (possibly even putting it on the constructor).

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,227 @@
...
+      internals.role = (this.constructor as FormAssociatedConstructor).role;
+    }
+
+    formDisabledCallback(disabled: boolean) {
+      if (disabled) {
```
sorvell, line 102: If internals is not private, perhaps this isn't needed since you could just do: `this[internals].ariaDisabled === 'true'`. Regardless, the specific set isn't needed since `isDisabled` could do the string check, right?

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,227 @@
...
+  ): ClassAccessorDecoratorResult<FormAssociated, V> => {
+    valueAccessors.set(context.metadata, context.access);
+
+    return {
+      get: target.get,
+      set(value: V) {
+        target.set.call(this, value);
+        getInternals(this)!.setFormValue(value);
+      },
+      init(value: V) {
```
sorvell, line 185: So this requires using standard decorators. Think that's ok?

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * A mixin that makes a ReactiveElement into a form-associated custom element.
+ *
+ * - ARIA role: Sets this element's ElementInternals role to the value of the
+ *   static `role` property.
+ * - Form value: The value of the element is stored in the field decorated with
+ *   the `@formValue()` decorator. This value field can have any name, but the
+ *   type must be assignable to `string | File | FormData | null`.
+ * - Form state: The state of the element is stored in the field decorated with
```
sorvell, line 64: I don't follow why this exists and is separate from the value. If it's to provide additional control, couldn't this just be done with value + user customization?

packages/labs/forms/src/form-associated.ts
justinfagnani: I was hoping to find a way to limit the call so that it could only be called once per subclass, like `attachInternals()`. Then if a subclass called it and stored the value, a user couldn't call it later and mess with the element.

packages/labs/forms/src/form-associated.ts
justinfagnani: The browser separates form state and value, in `setFormValue()` (which I need to call w/ state still) and `formStateRestoreCallback()`. This is to be able to save other state that isn't necessarily part of the submitted form value. You could serialize some transient UI state so when the form is restored you restore that UI.

I'm not sure how often it's used. It's weird to me that it has to have the same type as value. I can see how you can encode extra state in a string, but not how you can do that with a File object.

packages/labs/forms/src/form-associated.ts
justinfagnani: Konor pointed out that I can remove this and just use `element.matches(':disabled')` in `isDisabled()` too.

packages/labs/forms/src/form-associated.ts
justinfagnani: Just sketching things out for now and making sure they compile. I think we'll need to support experimental decorators though, and figure out how to handle the default value. Maybe we try out the pattern we're discussion for `@property()` here first.

packages/labs/forms/src/form-associated.ts
sorvell: ok, yeah. There's an example here: https://web.dev/articles/more-capable-form-controls#restoring-form-state

packages/labs/forms/src/form-associated.ts
sorvell: Sorry, not following. Limit the call to what, the property?

justinfagnani: Limit it to being called just once. `attachInternals()` throws if called a separate time, to help prevent outside access to it. An element can either call `attachInternals()` in the constructor (or set a static on the class, but I forget the name. Something like `static disableInternals = true`) to prevent anyone else from getting to internals.

I'd like to do the same for `getInternals()` so it doesn't bypass those protections.

justinfagnani: `disabledFeatures`. It's very poorly documented.

packages/labs/forms/src/form-associated.ts
justinfagnani: If we need user-configurable default values, then we'll have to have a separate decorator for defaults and won't need to rely on the initializer anymore.

packages/labs/forms/src/form-associated.ts
justinfagnani: I updated `getInternals()` to require a constructor that's on the proto chain be passed in, and allows only one call per proto. That can be worked around, but hopefully this prevents casual misuse.

packages/labs/forms/README.md
```diff
@@ -0,0 +1,168 @@
...
+### `@formStateGetter()`, `@formStateSetter()`, and `@formState()`
+
+Form state is an additional object that can be stored with a form that
+represents state that is not neccessarily part of the value and submitted with
+the form. If elements have additional state, they should pass it when setting
+form values and they should be able to derive a value from state.
+
+To enable this two decorators are provided to read and write form state:
+`@formStateGetter()` and `@formStateSetter()`. They are applied to a getter and
+setter to mark them as beind use to readn adn write state. When present,
```
43081j, line 80:
```suggestion
setter to mark them as being used to read and write state. When present,
```

I'm guessing this is what was meant

43081j (approved):

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,456 @@
...
+export type FormValue = string | File | FormData | null;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Constructor<T = object> = new (...args: any[]) => T;
+
+export interface FormAssociated extends ReactiveElement {
+  /**
+   * Implement this method to validate the element. Return a `ValidityResult`.
+   */
+  _getValidity?(): ValidityResult;
```
kyubisation, line 29: Code style question; Has this pattern (public properties/methods starting with underscore) been used previously? If not, would `protected` with a declared class be an option?

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,456 @@
...
+ */
+const _getInternals = (element: FormAssociated) => {
+  return internalsCache.get(element)!.internals;
+};
+
+/**
+ * Returns true if the element is disabled.
+ */
+export const isDisabled = (element: FormAssociated) =>
+  element.matches(':disabled');
```
kyubisation, line 146: This would probably break in SSR. I assume that is acceptable?

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,456 @@
...
+  formAssociated: true;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  new (...args: any[]): FormAssociated;
+}
+
+interface ValidityResult {
+  flags: Partial<Omit<ValidityState, 'valid'>>;
+  message?: string;
+  anchor?: HTMLElement;
```
kyubisation, line 58: I believe this breaks in webkit, if anchor is omitted (or is the element itself, from what I remember).
https://github.com/whatwg/html/issues/10155
https://bugs.webkit.org/show_bug.cgi?id=269832

kyubisation (commented):
Nice work! I generally like this approach. 👍

packages/labs/forms/src/form-associated.ts
kyubisation: This is apparently only a problem, if `setValidity` is called during constructor phase.

packages/labs/forms/src/form-associated.ts
theacodes: Another option is using symbols, which has a few benefits in terms of "hiding" the method from outsiders and making it explicit for implementers:

```typescript
export const getValidity = Symbol("getValidity");

export interface FormAssociated extends ReactiveElement {
    [getValidity]?(): ValidityResult;
}
```

packages/labs/forms/src/form-associated.ts
justinfagnani: This is a really good question. I can imagine something where we implement `.matches()` for a few specific cases like `:disabled` and have them delegate to attributes (and maybe do the form bits too?), otherwise there's no public API for this. We should open an issue for this.

packages/labs/forms/src/form-associated.ts
justinfagnani: Looks fixed now.

packages/labs/forms/src/form-associated.ts
justinfagnani: We have useDefault now, so I should see about making these work with experimental decorators, and getting the default value back out to use for form reset.

packages/labs/forms/src/form-associated.ts
```diff
@@ -0,0 +1,421 @@
...
+    constructor(...args: any[]) {
+      super(...args);
+      internalsCache.set(this, this.#internals);
+      this.#internals.role =
+        (this.constructor as FormAssociatedConstructor).role ?? null;
+    }
+
+    override attachInternals(): ElementInternals {
+      if (this.#attachInternalsCalled) {
+        throw new Error('attachInternals has already been called');
```
kyubisation: Should this maybe just call `super.attachInternals()` to provoke the native error?

justinfagnani: done

packages/labs/forms/src/form-associated.ts
justinfagnani: Added a TODO

Issue #4771, opened 2024-09-20
[react] Broken SSR compat without global HTMLElement shim


augustjk:
### Which package(s) are affected?

React (@lit/react)

### Description

The package includes a reference to a global `HTMLElement` here: https://github.com/lit/lit/blob/42260c2d3bb179756e109616338936de2a589801/packages/react/src/create-component.ts#L187

We already have a node build for `@lit/react` so we should make the node build not break by either removing the reference or bringing in a shim like we do for `@lit/reactive-element`.

Thanks to @hrmcdonald for finding the root cause.

### Reproduction

.

### Workaround

Include snippet below to entrypoint before any component is loaded.
```js
import {HTMLElement} from '@lit-labs/ssr-dom-shim';
globalThis.HTMLElement ??= HTMLElement;
```

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

.

### Browser/OS/Node environment

.

hrmcdonald:
I'd bet this is the same issue: https://github.com/lit/lit/issues/3407

PR #4774, opened 2024-09-23
[react] Prevent calling HTMLElement.prototype in SSR

kyubisation:
Currently in dev mode of the `@lit/react` package, reserved React properties are checked against the prototype of the given custom elements class. This used to work during SSR, as none of reserved keywords were implemented in the HTMLElement shim.
Now with #4553 `localName` is implemented in the shim, which is one of the reserved keywords, which causes `HTMLElement.prototype` to be evaluated in the code path and crashes during SSR.

This PR removes this check during SSR. Injecting the shim for the `node` exports was considered, but dismissed, as there is a potential difference between properties of the shim and the native Browser HTMLElement instance.

### DIFF
--- a/.changeset/fluffy-peaches-appear.md
+++ b/.changeset/fluffy-peaches-appear.md
@@ -1,0 +1,5 @@
+---
+'@lit/react': patch
+---
+
+Prevent calling HTMLElement.prototype in SSR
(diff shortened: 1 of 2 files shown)


augustjk:
Good find! I thought it might be the other `HTMLElement.prototype` here https://github.com/lit/lit/blob/42260c2d3bb179756e109616338936de2a589801/packages/react/src/create-component.ts#L185-L188 but that's only used inside `setProperty` which won't run during SSR.

Fixes #4771

augustjk (approved):

PR #4782, opened 2024-10-02
[all] Revert Rollup Terser plugin to rollup-plugin-terser

justinfagnani:
After updating the Rollup dependencies in https://github.com/lit/lit/pull/4681 our Terser minified name prefixing strategy started failing, meaning that we were outputting "plain" minified names (eg, `this.o`) instead of prefixed names (eg, `this._$Co`).

This is bad because for subclassable classes like ReactiveElement or LitElement, these minified names could clash with use subclass names. We haven't received any reports of this that I know of, but it's a possibility.

This PR simply revers our rollup plugin to `rollup-plugin-terser` instead of the newer and maintained `@rollup/plugin-terser`.

This is a temporary fix as `rollup-plugin-terser` isn't being maintained and eventually will pin us to old versions of Terser, or stop working in newer Rollups. If we can, we should make a reproduction and file a bug against `@rollup/plugin-terser`. The bug seems to be related to how the `nameCache` object is handled. It's a mutable object, so its identity needs to be stable across all Rollup runs.

In addition to this, I think we can also move away from this whole gambit of using the nameCache to prefix properties. Terser has a [`nth_identifier` option](https://terser.org/docs/options/#mangle-options) that allows complete customization of minified names. We can just prefix directly with that and remove all the name seeder code in our Rollup config.


rictic (approved):

justinfagnani:
oh, before merging, I think we need to update the changeset file to force releases of every package released since https://github.com/lit/lit/pull/4681

justinfagnani:
Scraping #4681, this is the list of packages released:

- @lit-labs/analyzer
- @lit-labs/compiler
- @lit-labs/gen-wrapper-vue
- lit
- lit-element
- lit-html
- @lit/localize-tools
- @lit-labs/cli
- @lit-labs/cli-localize
- @lit-labs/gen-manifest
- @lit-labs/gen-utils
- @lit-labs/gen-wrapper-react
- @lit-labs/observers
- @lit-labs/ssr-dom-shim
- @lit-labs/virtualizer
- @lit/localize
- eslint-plugin-lit
- @lit-labs/gen-wrapper-angular
- @lit/lit-starter-js
- @lit/lit-starter-ts

Of those, the client-side packages that use our Rollup config should be:
- lit
- lit-element
- lit-html
- @lit-labs/observers

@lit-labs/virtualizer uses Rollup, but has it's won separate config. @lit/localize doesn't have a Rollup config.

augustjk (approved):
With the added benefit of smaller size!

PR #4790, opened 2024-10-07
docs(context): Set correct language identifier

mgred:
`js` -> `ts`

This makes the highlighting of all code blocks equal.

### DIFF
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -138,7 +138,7 @@ useful to provide a context provider without introducing a custom element:
 
 #### **`my-app.js`**:
 
-```js
+```ts
 import {ContextProvider} from '@lit/context';
 import {loggerContext, Logger} from './logger.js';
 


packages/context/README.md
```diff
@@ -138,7 +138,7 @@ useful to provide a context provider without introducing a custom element:
 
 #### **`my-app.js`**:
 
-```js
+```ts
```
AndrewJakubowicz, line 141: Thank you for the change, but I'm not sure I understand it. The syntax highlighting should work as this code block is valid JavaScript. It makes sense for valid JavaScript examples to be highlighted as JavaScript instead of TypeScript.

mgred: Hey, I think I shot too fast. I thought there was a mistake with the highlighting, since everything was annotated with `ts`, but in this case it makes sense, because its about `my-app.js` :facepalm: 

Thanks for your answer and taking the time to review. I will close this.

PR #4799, opened 2024-10-17
Adds `noBooleanRender` to `RenderOptions`

### LINKED ISSUE #4788: [lit-html] Do not stringify boolean values in child parts
### Should this be an RFC?

### Description

The following syntax is not supported in lit-html as it will render `false` to the screen when condition doesn't match:

```tsx
html`<i>${condition && something}</i>`
```

lit-html's behavior of stringifying boolean values is not desirable. There is 0 benefit from rendering booleans as `"false"`, which is also what lit docs point out:
> A boolean value true will render 'true', and false will render 'false', **but rendering a boolean like this is uncommon**.

This quirk of lit-html makes migration from Stencil/React/Preact/Vue/Svelte or any other modern framework harder for no good reason - it is a source of bugs in production. Instead, modern rendering libraries ignore falsy values, which reduces the need for bottom values like `nothing`.

As an extension of https://github.com/lit/lit/issues/1559, I ask you to please reconsider the current behavior of rendering `false` to the screen.

I wrote a TypeScript transformer that tries to rewrite `${condition && something}` into `${condition ? something : nothing}` at build time. However, it doesn't catch all the cases, and adds complexity to our build pipeline.

It's no fun to see this in a production app 😓:

Screenshot 2024-10-06 at 12 10 47

sorvell:
Fixes #4788.

Currently html`${bool && 'hi'}` prints "false" when `bool` is `false`. Setting `noBooleanRender` to `true` in `RenderOptions` changes this behavior so that booleans are treated as a non-rendering value, and html`${bool && 'hi'}` prints nothing.

### DIFF
--- a/.changeset/sweet-fans-decide.md
+++ b/.changeset/sweet-fans-decide.md
@@ -1,0 +1,5 @@
+---
+'lit-html': minor
+---
+
+Adds `noBooleanRender` to `RenderOptions` to prevent booleans from rendering as strings.
--- a/package-lock.json
+++ b/package-lock.json
@@ -28628,7 +28628,7 @@
     },
     "packages/labs/signals": {
       "name": "@lit-labs/signals",
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "BSD-3-Clause",
       "dependencies": {
         "lit": "^2.0.0 || ^3.0.0",
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -711,6 +711,12 @@ export interface RenderOptions {
    * render to change the connected state of the part.
    */
   isConnected?: boolean;
+  /**
+   * By default boolean values render as strings ("true" or "false") in child
+   * parts. Set to `true` to treat these values as non-rendering, equivalent
+   * to setting `nothing`.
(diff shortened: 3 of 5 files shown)


leogreu:
Thank you! I tried it by adding it to a base class and it works as expected.

```js
import { html, css, LitElement } from "./lit/packages/lit/lit-core.min.js";

class BaseElement extends LitElement {
    renderOptions = {
        host: this,
        noBooleanRender: true
    };
}

export { html, css, BaseElement };
```

Do you have any thoughts on class field versus constructor-based initialization?

leogreu:
@sorvell @justinfagnani Just wanted to ask if there is anything new here? It would still be great to see this feature! 🙂

justinfagnani:
@leogreu I personally don't love that this adds 53 bytes to lit-html.

leogreu:
Really appreciate Lit's focus on a small bundle size! @justinfagnani 

But 53 bytes for a big increase in developer experience — plus many, many potentially removed `: undefined`'s (likely adding up to more than 53 bytes) compensate for this, in my opinion.

Please also see the latest comment on this issue: https://github.com/lit/lit/issues/4788#issuecomment-3370073791

PR #4803, opened 2024-10-20
[lab/analyzer] Dual-publish as CommonJS and standard modules

justinfagnani:
Some contexts where we want to use the analyzer, like tsserver plugins, are Common JS and sync everywhere, making it very difficult to consume standard modules. The easiest solution to dual-publish the analyzer.

`require(esm)` just landed in Node 23, and will eventually make it's way into Electron and VS Code, but that'll take quite a while.

### DIFF
--- a/.changeset/tidy-dragons-return.md
+++ b/.changeset/tidy-dragons-return.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/analyzer': minor
+---
+
+Dual publish as CommonJS and standard modules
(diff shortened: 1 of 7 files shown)


rictic (approved):
Changes look good to me. Have you tested it with `npm pack` or `npm link` and a consumer that uses require?

justinfagnani:
Sorry, for the changes, but I don't think I need this myself anymore. The tsserver-plugin depends on other standard modules, including lit-html, which I can't convert.

I figured out async loading of code in a plugin, so I worked around this...

justinfagnani:
I'm going to close this. If someone else needs CJS, we can resurrect it, but hopefully TS plugins get official async-loading support and `require(esm)` spreads fast enough to make this not necessary.

PR #4809, opened 2024-10-24
Fix undefined starting animation state in disconnected animation corner case

### LINKED ISSUE #4808: [motion] disconnected animation error on low end devices

Motion (@lit-labs/motion)

### Description

When application changes cause elements to be removed in quick succession that are animated with "out" animations, sometimes on low end devices the animation can create an error that prevents the elements from being removed.

A possible scenario: given a rendering like the following where `this.list` length is reduced quickly while animations may be playing.

```js
 this.list.map((item) => html`<li ${animate({out: fadeOut, stabilizeOut: true})}>...</li>`
```

The disconnected animation process performs the following steps:
1. on disconnection, wait for a `requestAnimationFrame` 
2. re-attach the disconnected element
3. move it to its previous position
4. run the "out" animation
5. remove the element.

On step (3), the previous position data for the element is used. However, in rare cases when the `requestAnimationFrame` is significantly delayed, the previous position data can be undefined, and this results in an error. That error prevents the "out" animation from running and the element from being removed.

### Reproduction

Corner case, hard to repro

### Workaround

N/A

### Is this a regression?

### Affected versions

@lit-labs/motion at least to 1.0.7

All

sorvell:
Fixes #4808.

### DIFF
--- a/packages/labs/motion/src/animate.ts
+++ b/packages/labs/motion/src/animate.ts
@@ -404,6 +404,7 @@ export class Animate extends AsyncDirective {
       return;
     }
     this.prepare();
+    const disconnectedFromValues = this._fromValues;
     await animationFrame();
     if (this._parentNode?.isConnected) {
       // put element back in DOM
@@ -420,6 +421,7 @@ export class Animate extends AsyncDirective {
         // TODO(sorvell): these nudges could conflict with existing styling
         // or animation but setting left/top should be rare, especially via
         // animation.
+        this._fromValues ??= disconnectedFromValues;
         const left =
           (this._fromValues!['left'] as number) - (shifted['left'] as number);
         const top =


packages/labs/motion/src/animate.ts
```diff
@@ -420,6 +421,7 @@ export class Animate extends AsyncDirective {
         // TODO(sorvell): these nudges could conflict with existing styling
         // or animation but setting left/top should be rare, especially via
         // animation.
+        this._fromValues ??= disconnectedFromValues;
```
justinfagnani, line 424: I think assigning back to `this` is a little hard to reason about. If directive instance state might change in bad ways while awaiting an animation frame, then I think that the critical state should be saved off before the `await animationFrame()` and reference to `this` be avoided after it, to avoid the "shearing".

Even if you don't do this for every field yet, I think it'd be good to structure it this way for `_fromValues`:

```ts
    this.prepare();
    const fromValues = this._fromValues;
    await animationFrame();
    // ...
    const left = fromValues['left'] ...
```

justinfagnani:
ping @sorvell

PR #4820, opened 2024-11-06
[labs/ssr] Update @types/node dependency range to prevent conflicts in consumer projects

kyubisation:
Currently the `@lit-labs/ssr` package uses `^16.0.0` as the dependency version for `@types/node`. This causes npm or yarn install in consuming projects to complain about version incompatibility, if the project uses a newer major version.
This PR changes the dependency version range to `>=18.0.0` (the oldest still supported LTS version).


packages/labs/ssr/package.json
```diff
@@ -228,7 +228,7 @@
     "@lit-labs/ssr-dom-shim": "^1.2.0",
     "@lit/reactive-element": "^2.0.4",
     "@parse5/tools": "^0.3.0",
-    "@types/node": "^16.0.0",
+    "@types/node": ">=18.0.0",
```
augustjk, line 231: perhaps this should be a dev dependency instead? i think moving it devDependencies would make it not install for downstream consumers.

kyubisation: Yes, but I wonder if that would break consumers? But I guess if consumers use TypeScript for this, they would have @types/node already installed?

kyubisation: Hm, `render-result-readable.ts` imports/exports `Readable` as part of the public API. I will test if that breaks, when using that in a consuming project and running type checking.

kyubisation: Ok, did a quick check. When we move it to a devDependency, it works when `skipLibCheck` in tsconfig.json is set to true. Otherwise it breaks the build.
Is that ok?
Alternatively we could also define it as an optional peer dependency.

packages/labs/ssr/package.json
```diff
@@ -228,14 +228,16 @@
...
     "enhanced-resolve": "^5.10.0",
     "lit": "^3.1.2",
     "lit-element": "^4.0.4",
     "lit-html": "^3.1.2",
     "node-fetch": "^3.2.8",
     "parse5": "^7.1.1"
   },
+  "peerDependencies": {
+    "@types/node": ">=18.0.0 <25.0.0"
+  },
```
louis-bompart, line 240: For pure JS consumers, we should not impose `@types/node` as a dependency in their project I think.

```suggestion
  },
  "peerDependenciesMeta": {
    "@types/node": {
      "optional": true
    }
  },
```

packages/labs/ssr/package.json
```diff
@@ -228,14 +228,16 @@
...
-    "@types/node": "^16.0.0",
     "enhanced-resolve": "^5.10.0",
     "lit": "^3.1.2",
     "lit-element": "^4.0.4",
     "lit-html": "^3.1.2",
     "node-fetch": "^3.2.8",
     "parse5": "^7.1.1"
   },
+  "peerDependencies": {
+    "@types/node": ">=18.0.0 <25.0.0"
```
louis-bompart, line 239: I think it might be a bit early to include Node24, as development on it as not started yet.
```suggestion
    "@types/node": ">=18.0.0 <24.0.0"
```

louis-bompart (commented):
Random review/opinion on a couple points ;)

packages/labs/ssr/package.json
kyubisation: I suggested this, but it was rejected. But I can add it.

packages/labs/ssr/package.json
kyubisation: While I agree, this will force us to update it every 6 months, for an API (`import {Readable} from 'stream'`) that is unlikely to change.
From my understanding there is not yet an automation to update dependencies, so someone will continuously have to remember to update the range.
This range will give us time until the end of next year, while Node.js 24 will be released in April (although not yet as LTS).

kyubisation:
@louis-bompart I updated the PR as suggested. Thanks for the review.

kyubisation:
Sorry, I did not realize updating the Node.js version breaks the custom vm types. This only happens, if you delete the previously built files.
I'll have a look, whether I can figure out the differences.

kyubisation:
@augustjk I have re-applied to changes the actual state of the main branch and resolved the `vm` types. Can you take another look?

packages/labs/ssr/custom_typings/node.d.ts
augustjk, line 1: @justinfagnani do you recall why we needed these custom types? maybe `@types/node` was lacking at the time, and it's safe to rely on that now.

kyubisation: With @types/node <= 16, these types were not defined. With @types/node 18 they were incompatible. With @types/node >= 20 they seem to be obsolete.

kyubisation:
The failing tests are timeouts. As the tests are successful locally, I assume a rerun should fix it.

kyubisation:
@augustjk Can I ask you to merge this?

kyubisation:
@augustjk Thank you very much! 😃

PR #4821, opened 2024-11-07
[labs/testing] Fix fixture base url resolution when running browsers in containers

kyubisation:
Currently when running browsers in a container (e.g. playwright), the resolution to localhost will fail.
This PR adds `host.containers.internal`, additionally to `localhost`, to the base resolution regex.
`host.containers.internal` is available per default in [Podman](https://podman.io/) and can easily be added to Docker (`docker run --add-host=host.containers.internal=host-gateway ...`).

### DIFF
--- a/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
+++ b/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
@@ -53,6 +53,7 @@ export async function ssrFixture<T extends HTMLElement>(
     // asyncFunctionResume@[native code]
     // @http://localhost:8000/test/my-element_test.js?wtr-session-id=aKWON-wBOBGyzb2CwIvmK:65:37
     const {stack} = new Error();
+    // host.containers.internal represents the host address when running browser/tests inside a container.
     const match =
       stack?.match(
         /http:\/\/(localhost|host\.containers\.internal).+(?=\?wtr-session-id)/
--- a/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
+++ b/packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
@@ -55,7 +55,7 @@ export async function ssrFixture<T extends HTMLElement>(
     const {stack} = new Error();
     const match =
       stack?.match(
-        /http:\/\/(localhost|host.containers.internal).+(?=\?wtr-session-id)/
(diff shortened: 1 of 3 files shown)


packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
```diff
@@ -54,7 +54,9 @@ export async function ssrFixture<T extends HTMLElement>(
     // @http://localhost:8000/test/my-element_test.js?wtr-session-id=aKWON-wBOBGyzb2CwIvmK:65:37
     const {stack} = new Error();
     const match =
-      stack?.match(/http:\/\/localhost.+(?=\?wtr-session-id)/) ??
+      stack?.match(
+        /http:\/\/(localhost|host.containers.internal).+(?=\?wtr-session-id)/
```
augustjk, line 58: should the `.` be escaped? i don't have a quick way of testing this. it probably works without it since it's any char, but seems more proper.
```suggestion
        /http:\/\/(localhost|host\.containers\.internal).+(?=\?wtr-session-id)/
```

packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
```diff
@@ -64,7 +66,11 @@ export async function ssrFixture<T extends HTMLElement>(
       // ssrFixture@http://localhost:8000/lib/fixtures/ssr-fixture.js:19:34
       // ssrNonHydratedFixture@http://localhost:8000/lib/fixtures/ssr-fixture.js:98:45
       // @http://localhost:8000/test/my-element_test.js:20:37
-      [...(stack?.matchAll(/http:\/\/localhost:?[^:)]+/gm) ?? [])]
+      [
+        ...(stack?.matchAll(
+          /http:\/\/(localhost|host.containers.internal):?[^:)]+/gm
```
augustjk, line 71: same question as above
```suggestion
          /http:\/\/(localhost|host\.containers\.internal):?[^:)]+/gm
```

packages/labs/testing/src/lib/fixtures/ssr-fixture.ts
kyubisation: Nice catch. Thanks!

kyubisation:
@augustjk I don't think this change is responsible for the build breaking. Do you have an idea?

kyubisation:
@augustjk Rebased. Thanks for the upstream fix. 👍

PR #4836, opened 2024-11-16
feat(task): improve type checking for tuples

### LINKED ISSUE #4829: [Lit/Task]: Improve type-checking by using `const`
### Should this be an RFC?

Task (@lit/task)

### Description

TypeScript 5.0 [added the `const` modifier for type variables](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters).

It can be adapted in`@lit/task` like so:

```diff
export class Task<
-  T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
+  const T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
-  R = unknown,
+  const R = unknown,
 > {
```

The advantage of this is that argument types and return types would be inferred better by TypeScript. Example:

```tsx
// Before this was inferred as Task<(string | undefined)[], string[]>
// With addition of "const" it would be Task<readonly [string | undefined], readonly [string]> - a much more accurate type
  _dataTask = new Task(
    this,
    async ([name]) => {
      if (name === undefined) {
        return "";
      }

      const response = await fetch(`example.com/${name}`);
      return [await response.text()];
    },
    () => [this.name],
  );
```

If `const` is not included in the `@lit/tasks` typings, it can be manually included in the component source code:

```diff
-    () => [this.name],
+    () => [this.name] as const,
```

however, not every TypeScript user would be aware of this, so some might get frustrated by bad type checking experience when using Lit Task

maxpatiiuk:
Fixes #4829

### DIFF
--- a/.changeset/wicked-rabbits-trade.md
+++ b/.changeset/wicked-rabbits-trade.md
@@ -1,0 +1,5 @@
+---
+'@lit/task': patch
+---
+
+Improve type inference of tuples returned by the args function being used as task function parameter.
--- a/packages/labs/task/src/test/task_test.ts
+++ b/packages/labs/task/src/test/task_test.ts
@@ -759,7 +759,7 @@ suite('Task', () => {
   });
 
   test('Elements only render once for pending tasks', async () => {
-    let resolveTask: (v: unknown) => void;
+    let resolveTask: (v: any) => void;
     let renderCount = 0;
     class TestElement extends ReactiveElement {
       task = new Task(this, {
--- a/packages/task/src/task.ts
+++ b/packages/task/src/task.ts
@@ -227,12 +227,12 @@ export class Task<
   private _rejectTaskComplete?: (e: unknown) => void;
   private _taskComplete?: Promise<R>;
 
+  constructor(host: ReactiveControllerHost, task: TaskConfig<T, R>);
   constructor(
     host: ReactiveControllerHost,
     task: TaskFunction<T, R>,
     args?: ArgsFunction<T>
   );
-  constructor(host: ReactiveControllerHost, task: TaskConfig<T, R>);
   constructor(
     host: ReactiveControllerHost,
     task: TaskFunction<T, R> | TaskConfig<T, R>,
--- a/packages/task/src/task.ts
+++ b/packages/task/src/task.ts
@@ -166,8 +166,8 @@ export interface TaskConfig<T extends ReadonlyArray<unknown>, R> {
  * ```
  */
 export class Task<
-  T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
-  R = unknown,
+  const T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
+  const R = unknown,
 > {
   private _previousArgs?: T;
   private _task: TaskFunction<T, R>;
--- a/packages/task/src/test/task_test.ts
+++ b/packages/task/src/test/task_test.ts
@@ -759,7 +759,7 @@ suite('Task', () => {
   });
 
   test('Elements only render once for pending tasks', async () => {
-    let resolveTask: (v: unknown) => void;
+    let resolveTask: (v: any) => void;
     let renderCount = 0;
     class TestElement extends ReactiveElement {
       task = new Task(this, {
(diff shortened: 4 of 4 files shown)


justinfagnani:
Thanks Max!

I'd like to see a type-only test or two to make sure that this is doing what we expect. I tried it locally and it doesn't seem to infer the args array as const automatically for me:

```ts
  test('type-only args type test', () => {
    const accept = <T>(x: T) => x;

    class TestElement extends ReactiveElement {
      foo: string = 'foo';
      bar: number = 123;

      task = new Task(this, {
        task: ([foo, bar]) => {
          accept<string>(foo); // Argument of type 'string | number' is not assignable to parameter of type 'string'.
          accept<number>(bar); // Argument of type 'string | number' is not assignable to parameter of type 'number'.
        },
        args: () => [this.foo, this.bar],
      });
    }
    TestElement;
  });
```

maxpatiiuk:
The 1st overload is typed correctly after the changes in this PR:

```ts
      task1 = new Task(
        this,
        ([string, number]) => {
          expectType<string>(string);
          expectType<number>(number);
        },
        () => [this.string, this.number]
      );
```

The 2nd overload is not inferred correctly. Appears to be a TypeScript limitation.

I also tried the following alternatives, which were not inferred correctly either:

```ts
// Defining constructors with generic type parameters as per https://github.com/microsoft/TypeScript/issues/10860
export type ITask = {
  new <
    const T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
    const R = unknown,
  >(
    host: ReactiveControllerHost,
    task: TaskFunction<T, R>,
    args?: ArgsFunction<T>
  ): Task<T, R>;
  new <
    const T extends ReadonlyArray<unknown> = ReadonlyArray<unknown>,
    const R = unknown,
  >(
    host: ReactiveControllerHost,
    task: TaskConfig<T, R>
  ): Task<T, R>;
// ... other members omitted
};
///...
    const T = Task as unknown as ITask;

      task = new T(this, {
        task: ([string, number]) => {
          expectType<string>(string); // Argument of type 'string | number' is not assignable to parameter of type 'string'.
          expectType<number>(number); // Argument of type 'string | number' is not assignable to parameter of type 'number'.
        },
        args: () => [this.string, this.number],
      });
```

```ts
// Defining a wrapper function
export function useTask<const T extends ReadonlyArray<unknown>, const R>(
  host: ReactiveControllerHost,
  task: TaskFunction<T, R>,
  args?: ArgsFunction<T>
): Task<T, R>;
export function useTask<const T extends ReadonlyArray<unknown>, const R>(
  host: ReactiveControllerHost,
  config: TaskConfig<T, R>
): Task<T, R>;
export function useTask<const T extends ReadonlyArray<unknown>, const R>(
  host: ReactiveControllerHost,
  task: TaskFunction<T, R> | TaskConfig<T, R>,
  args?: ArgsFunction<T>
): Task<T, R> {
  return new Task(host, task as TaskFunction<T, R>, args);
}

// ...

      task3 = useTask(this, {
        task: ([string, number]) => {
          expectType<string>(string); // Argument of type 'string | number' is not assignable to parameter of type 'string'.
          expectType<number>(number); // Argument of type 'string | number' is not assignable to parameter of type 'number'.
        },
        args: () => [this.string, this.number],
      });
```

However, as soon as I disable the 1st function overload, the 2nd one begins getting inferred correctly:
```ts
  constructor(host: ReactiveControllerHost, task: TaskConfig<T, R>) {

// ...

      task = new Task(this, {
        task: ([string, number]) => {
          expectType<string>(string); // all good!
          expectType<number>(number);
        },
        args: () => [this.string, this.number],
      });
```

(the useTask wrapper is also inferred correctly once I remove the 1st function overload)

We can open an issue in the TypeScript repository for improved `const` inference in function overloads, though I imagine it is complicated to make that work in a performant way.

Until then, I will add a type test for the 1st function overload - I imagine that's the more commonly used signature?

augustjk:
Maybe my machine is crazy, but I tried reversing the constructor overload order and it seems to correctly infer the types for both forms?? Our docs on lit.dev do prefer the object form for the second arg so it would be good for it to work there.

justinfagnani (approved):
Nice!

maxpatiiuk:
Hi
Thanks for merging this PR
Could you publish a new `@lit/task` release that includes this fix?

PR #4838, opened 2024-11-20
[lit-html] Use Node.moveBefore if it's available

justinfagnani:
`Node.moveBefore()` is an atomic version of `insertBefore` that is available in Chrome Canary behind a flag.

- Explainer: https://github.com/noamr/dom/blob/spm-explainer/moveBefore-explainer.md
- DOM issue: https://github.com/whatwg/dom/issues/1255
- Intent to Ship: https://groups.google.com/a/chromium.org/g/blink-dev/c/YE_xLH6MkRs/m/_7CD0NYMAAAJ

Using `moveBefore()` preserves the state of the element being moved, including focus, selection, iframe state, media state, etc.

We can adopt this API when available to make `repeat()` preserve the state of elements. The one place that needs to change is in `insertPart()` in `directive-helpers.ts`.

`moveBefore()` still calls `disconnectedCallback()` and `connectedCallback()` on the element being moved, so elements should still take care to not over-aggressively free up resources in `disconnectedCallback()`. A common pattern is to delay cleanup for a microtask and check `isConnected` before cleanup - that should still be done.

Another possible change it in `lit-html.ts` when we insert a node into a child part. Adoption `moveBefore()` there would mean that directly binding a node to one of multiple child bindings could be done so that if the binding that the node is in changes, it'll be atomic. That's probably an _very_ rare case today, though this could potentially be done by directives that move nodes between parents as in some drag and drop constructions we've seen (moving items between lists, etc).

https://github.com/lit/lit/blob/1eb179f69d663440fd2ebd3589b6f2808d87494f/packages/lit-html/src/lit-html.ts#L1495

One reason why we might want to try that anyway is perf - we need to test any perf difference between `insertBefore()` and `moveBefore()` but it seems _possible_ at least that `moveBefore()` could be faster if it's skipping work.

One thing to watch out for is that `moveBefore()` can throw in some cases. It doesn't seem like we can hit them in `repeat()`, but maybe in `cache()`. We might need an option to `insertPart()`.

I haven't yet figured out how to enable the feature in Playwright to test this, so do not merge.

### DIFF
--- a/packages/lit-html/src/directive-helpers.ts
+++ b/packages/lit-html/src/directive-helpers.ts
@@ -164,7 +164,12 @@ export const insertPart = (
       let start: Node | null = part._$startNode;
       while (start !== endNode) {
         const n: Node | null = wrap(start!).nextSibling;
-        wrap(container).insertBefore(start!, refNode);
+        const wrapper = wrap(container);
+        (wrapper.moveBefore ?? wrapper.insertBefore).call(
+          wrapper,
+          start!,
+          refNode
+        );
         start = n;
       }
     }
@@ -172,6 +177,12 @@ export const insertPart = (
 
   return part;
 };
+
+declare global {
+  interface Node {
+    moveBefore<T extends Node>(node: T, child: Node | null): T;
+  }
+}
 
 /**
  * Sets the value of a Part.


jimsimon:
Should this be considered a breaking change since it could impact the behavior of existing code that's using repeat? It seems unlikely, but someone could be relying on the existing behavior, right? It might be worth considering making this opt-in somehow.

packages/lit-html/src/directive-helpers.ts
```diff
@@ -173,6 +183,12 @@ export const insertPart = (
   return part;
 };
 
+declare global {
```
rictic, line 186: Augmenting globals is messy and surprising for users, consider instead casting to:

```ts
interface NodeWithMoveBefore extends Node {
  moveBefore?: <T extends Node>(node: T, child: Node | null) => T;
}
```

justinfagnani:
@jimsimon 

> Should this be considered a breaking change since it could impact the behavior of existing code that's using repeat? It seems unlikely, but someone could be relying on the existing behavior, right? It might be worth considering making this opt-in somehow.

Something to consider, though I don't love adding more and more options. I don't think people would be relying on the current behavior, since in our re-order algorithm you don't quite know which nodes are moved and which aren't. It's deterministic, but only certain nodes would have their state reset.

A larger danger, I suspect, is state being preserved in Chromium browsers with `moveBefore()` and not in others for a time, so if the developer doesn't test enough, they'll have slightly broken apps in Webkit and Firefox.

packages/lit-html/src/directive-helpers.ts
justinfagnani: Good point!

sorvell (approved):
Might we also want to use this for `ChildPart._insert`? In practice, I think it would just handle the case where users are manually inserting nodes into parts and moving them around.

I also added https://github.com/whatwg/dom/pull/1307#issuecomment-2491906616

packages/lit-html/src/directive-helpers.ts
```diff
@@ -162,9 +162,19 @@ export const insertPart = (
     }
     if (endNode !== refNode || parentChanged) {
       let start: Node | null = part._$startNode;
+      // moveBefore() cannot be called if either the new parent or the node to
+      // move is disconnected, so we fall back to insertBefore() in that case.
```
domfarolino, line 166: Note that this is not _entirely_ true; as of the latest commits to the spec PR, we throw if exactly one of the old or new parent are disconnected, but we DO support the disconnected->disconnected case, in response to some developer feedback. Just FYI!

nolanlawson:
> `moveBefore()` still calls `disconnectedCallback()` and `connectedCallback()` on the element being moved

According to [the explainer](https://github.com/noamr/dom/blob/spm-explainer/moveBefore-explainer.md), this is not the case?

> The author gets reflection of this in web components: a new optional `movedCallback`, which is invoked instead of `disconnectedCallback` and `connectedCallback`

(In the PR this seems to have been renamed to `connectedMoveCallback`: https://github.com/whatwg/dom/pull/1307)

sorvell:
> >moveBefore() still calls disconnectedCallback() and connectedCallback() on the element being moved

> According to [the explainer](https://github.com/noamr/dom/blob/spm-explainer/moveBefore-explainer.md), this is not the case?

Yes, my understanding is that disconnectedCallback/connectedCallback should *not* be called. However, in Chrome's current prototype they are still called so this is likely contributing confusion.

packages/lit-html/src/directive-helpers.ts
justinfagnani: @domfarolino thanks! Can you look at the new condition? I had to look at the spec for the Move operation to try to mirror the logic.

It'd be nice if there were a DOM API that did this check for us and called `insertBefore()` as a fallback. Since `moveBefore()` is doing the same checks, we're doing them twice, once in JS once in the DOM.

justinfagnani: I filed https://github.com/whatwg/dom/issues/1406 for moveOrInsertBefore()

packages/lit-html/src/directive-helpers.ts
```diff
@@ -162,17 +163,30 @@ export const insertPart = (
     }
     if (endNode !== refNode || parentChanged) {
       let start: Node | null = part._$startNode;
+      // moveBefore() cannot be called if the old parent and new parent do not
+      // have the same connected shadow-including-root, so we fall back to
```
domfarolino:
```suggestion
      // have the same shadow-including-root, so we fall back to
```

moveBefore() also cannot be called if they do not share the same _disconnected_ shadow-including-root! The shadow-including-root must be the same, regardless of connection status. The condition below is correct, it's just this documentation that I'd tweak.

packages/lit-html/src/directive-helpers.ts
domfarolino: Thanks! I'd personally like to see that API be introduced. But regarding the original question, the condition is right, I'd just tweak the documentation above it, as recommended in https://github.com/lit/lit/pull/4838#pullrequestreview-3278395601.

packages/lit-html/src/directive-helpers.ts
justinfagnani: Thanks @domfarolino!

btw, I've never benchmarked `getRootNode({composed: true})`, nor used it in a hot code path like this. I'm somewhat concerned about the performance impact of calling it twice here if the composed root has to be looked up with an ancestor walk. I'm wondering if I should do other checks first, like `_endNode.parentNode === container`.

And while I'm dreaming about more ergonomic move APIs, `node.hasSameRoot(otherNode, {composed: true})` might be nice if it can do a faster check than two composed root lookups.

domfarolino: The only reason you have to call `getRootNode({composed: true})` on each node is to cater to the disconnected case; that's the only time where the the composed roots can differ (provided the start and end containers are same document).

So if you want to speed things up, you could tighten your conditions to only exercise `moveBefore()` in the connected case (what most people care about), and you just have to run a `connected && connected` check, as well as check that the `ownerDocument`s match.

All of those are constant-time checks, and you'd only be limiting usage of `moveBefore()` to the case where the target node is connected, as a trade-off.

PR #4840, opened 2024-11-21
[labs/nextjs] Fix type of nextjs config wrapper

kyubisation:
Currently the nextjs config wrapper type mistakenly indicates it returns a NextConfig object. It however returns a callback, which receives and returns a NextConfig object.

### DIFF
--- a/.changeset/cold-bees-doubt.md
+++ b/.changeset/cold-bees-doubt.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/nextjs': patch
+---
+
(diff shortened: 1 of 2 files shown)


augustjk (approved):
thanks! i wonder why TS never complained about the implementation not matching annotation here.

kyubisation:
That is confusing me as well. First I thought that maybe the NextConfig type allows it, but that does not seem to be the case. Next I checked whether `export =` has some typing problems, but that isn't the case either.
TypeScript just doesn't seem to care for this...

Issue #4833, opened 2024-11-14
[labs/virtualizer] `renderItem` is called with `undefined` as the first argument


jpzwarte:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

By changing the `items` array to a subset of an earlier value and not scrolling the first item of the new array into view, i managed to get `renderItem` to be called with `undefined` as the first argument. I fixed this error by scrolling to the first item in the new array. I'll try to create a reproduction when i have time.

### Reproduction

TBD

### Workaround

Scroll the first item of the new `items` array into view.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

latest

### Browser/OS/Node environment

-

PR #4846, opened 2024-11-27
Virtualizer fixes

jpradelle:
Hi, I'm not sure you want to merge it as is. Purpose of this PR is to document issues I faced with virtualizer and fixes I made to have it working properly in my context: create a data table with virtualizer.

Please see code attached comments

### DIFF
--- a/packages/labs/virtualizer/src/Virtualizer.ts
+++ b/packages/labs/virtualizer/src/Virtualizer.ts
@@ -504,9 +504,7 @@ export class Virtualizer {
     for (let i = 0; i < children.length; i++) {
       const child = children[i];
       const idx = this._first + i;
-      if (this._itemsChanged || this._toBeMeasured.has(child)) {
-        mm[idx] = fn.call(this, child, this._items[idx]);
-      }
+      mm[idx] = fn.call(this, child, this._items[idx]);
     }
     this._childMeasurements = mm;
     this._schedule(this._updateLayout);
@@ -720,17 +718,14 @@ export class Virtualizer {
         if (child) {
           child.style.position = 'absolute';
           child.style.boxSizing = 'border-box';
-          child.style.transform = `translate(${left}px, ${top}px)`;
           if (width !== undefined) {
             child.style.width = width + 'px';
           }
           if (height !== undefined) {
             child.style.height = height + 'px';
           }
-          (child.style.left as string | null) =
-            xOffset === undefined ? null : xOffset + 'px';
-          (child.style.top as string | null) =
-            yOffset === undefined ? null : yOffset + 'px';
+          child.style.left = `${(xOffset === undefined ? 0 : xOffset) + left}px`;
+          child.style.top = `${(yOffset === undefined ? 0 : yOffset) + top}px`;
         }
       });
     }
--- a/packages/labs/virtualizer/src/layouts/flow.ts
+++ b/packages/labs/virtualizer/src/layouts/flow.ts
@@ -312,6 +312,8 @@ export class FlowLayout extends BaseLayout<BaseLayoutConfig> {
     } else {
       this._getItems();
     }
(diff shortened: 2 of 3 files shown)


packages/labs/virtualizer/src/Virtualizer.ts
```diff
@@ -720,17 +718,14 @@ export class Virtualizer {
...
           }
           if (height !== undefined) {
             child.style.height = height + 'px';
           }
-          (child.style.left as string | null) =
-            xOffset === undefined ? null : xOffset + 'px';
-          (child.style.top as string | null) =
-            yOffset === undefined ? null : yOffset + 'px';
+          child.style.left = `${(xOffset === undefined ? 0 : xOffset) + left}px`;
+          child.style.top = `${(yOffset === undefined ? 0 : yOffset) + top}px`;
```
jpradelle, line 728: In my context I don't want a new stacking context being created, this is an issue. Avoid using transform can fix that.
In addition I had to put `style="contain: size"` to the scrolling container, to avoid `layout` value https://github.com/lit/lit/blob/05691ba4848f1bcbec06df7dcd63ee024589ed47/packages/labs/virtualizer/src/Virtualizer.ts#L362

packages/labs/virtualizer/src/layouts/flow.ts
```diff
@@ -312,6 +312,8 @@ export class FlowLayout extends BaseLayout<BaseLayoutConfig> {
     } else {
       this._getItems();
     }
+    // Fix scroll size after rendering
+    this._updateScrollSize();
```
jpradelle, line 316: With items having different sizes, end of scrolling container was not aligned with bottom of last item. I needed to had this update scroll size, with code update in `_updateScrollSize`

packages/labs/virtualizer/src/layouts/flow.ts
```diff
@@ -360,8 +362,13 @@ export class FlowLayout extends BaseLayout<BaseLayoutConfig> {
...
     // pass to stabilize a previously started reflow, we will already
-    // have an anchor. If not, establish an anchor now.
-    if (this._anchorIdx === null || this._anchorPos === null) {
+    // have an anchor and anchor is in view. If not, establish an anchor now.
+    if (
+      this._anchorIdx === null ||
+      this._anchorPos === null ||
+      this._anchorPos < lower ||
+      this._anchorPos > upper
+    ) {
```
jpradelle, line 371: While scrolling, sometime the anchor was far away from the current scroll position. There is a fix after to move scrollbar to anchor position: https://github.com/lit/lit/blob/05691ba4848f1bcbec06df7dcd63ee024589ed47/packages/labs/virtualizer/src/layouts/flow.ts#L403
This can move it a lot and makes scroll moving away from mouse cursor. Adding a check to recompute a new anchor here makes scroll much more smooth.

packages/labs/virtualizer/src/virtualize.ts
```diff
@@ -74,7 +74,11 @@ class VirtualizeDirective<T = unknown> extends AsyncDirective {
     const itemsToRender: Array<T> = [];
 
     if (this._first >= 0 && this._last >= this._first) {
-      for (let i = this._first; i <= this._last; i++) {
+      for (
+        let i = this._first;
+        i <= this._last && i < this._items.length;
+        i++
+      ) {
         itemsToRender.push(this._items[i]);
```
jpradelle, line 82: I sometimes got `undefined` elements in `keyFunction` and `renderItem`. I think when item set change with less items, _last is recomputed after a render event, or something like that. This avoid `undefined` values. Maybe it would require a fix before being here.

jpradelle: I think this is this issue https://github.com/lit/lit/issues/4833

Issue #4767, opened 2024-09-17
Scrolling to a far away element using scrollIntoView then nothing gets rendered


MattiasJa:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

When scrolling to a far away element, for example using virtualizer.element(xyz).scrollIntoView then nothing gets rendered,

For example 30000 elements in the list and each element have height 37px. Scrolling to anything above 14984 not working.

(temp2 is the virtualizer ref in the browser console  )
temp2.element(14984).scrollIntoView({block: 'center', behavior: 'instant',})
OK, list scrolls and gets rendered correct.

temp2.element(14985).scrollIntoView({block: 'center', behavior: 'instant',})
NOK, list is blanked no elements rendered. Any value >= 14985
At the same time as this problem theres also a printput : err ResizeObserver loop completed with undelivered notifications.
in the console.log

Same problem with scrollToIndex.

Getting a ref to the scrollable element above <lit-virtualizer> and using:
scrollabeRef.scrollTop = 37 * 14985 is working

lit-virtualizer is rendered in a lit component as :

                <lit-virtualizer
                    .items=${this.options.option}
                    .renderItem=${renderItem}
                ></lit-virtualizer>

The Scrollable element is outside the host component.

### Reproduction

[...]

Issue #4789, opened 2024-10-07
[@lit-labs/virtualizer] Metrics cache is not reset when the `items` array is changed


jpzwarte:
### Which package(s) are affected?

Virtualizer (@lit-labs/virtualizer)

### Description

When the `items` array changes, the metrics cache in `flow.ts` is not cleared, leading to `_getAverageSize()` returning an incorrect value, which eventuallyt causes the virtualizer host element to have an incorrect `min-height`

### Reproduction

https://lit.dev/playground/#gist=1781355efc64eba1811131415cbdbfc0

### Workaround

You can clear the metrics cache from the outside, but you have to be *very* careful *when* you do this. If you do it too often, then things reset to the default `_itemSize` of 100px.

`this.#virtualizer?._layout?._metricsCache?.clear()`

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

Latest

### Browser/OS/Node environment

-

PR #4847, opened 2024-11-27
Add test for #4767

jpzwarte:


graynorton:
@jpzwarte Sorry, I accidentally merged this, meaning to merge your test for #4789. I've force-pushed the `gnorton/fix-4767` branch so that it no longer contains this test, but I think you'll need to open a new PR.

PR #4850, opened 2024-11-28
[labs/observers] Fix ResizeController's skipInitial behavior. Closes #4831

### LINKED ISSUE #4831: labs/observers: ResizeController's skipInitial option has no effect

### Description

Even when a `ResizeController`'s `skipInitial` flag is set to `true`, it triggers an initial update.

The test suite's test case for `skipInitial` falsely passes (it's false-green). We can get the the test to fail by waiting for two additional animation frames (or alternatively, 100 ms). During this time window, an update triggers, even though it shouldn't.

I ran into this while troubleshooting my app's performance. As I have a lot of elements with resize observers, the extraneous update slows down the initial page load.

### Reproduction

I created a pull request to make the test fail correct at [placeholder -- will edit].

### Workaround

```typescript
  _resizeObserver = new ResizeController(this, { skipInitial: true });
  _skipNextUpdate = true;

  shouldUpdate(_changedProperties: PropertyValues) {
    if (this._skipNextUpdate) {
      this._skipNextUpdate = false;
      return false;
    }
    return super.shouldUpdate(_changedProperties);
  }
}
```

### Is this a regression?

### Affected versions

@lit-labs/observers 2.0.4 and main (1eb179)

Browser: Chrome 132.0.6821.4 
OS: macOS 15.1

For failing tests:
Node: 22.3.0
npm: 10.8.1
OS: Linux

joliss:
The `ResizeController` would previously manually trigger an initial callback (which would get skipped if `skipInitial` is true), but this isn't necessary, as the native `ResizeObserver` already triggers an immediate callback when observing initially[1]. So this commit rewrites the logic to instead suppress the initial time the `ResizeObserver` fires for a given element if `skipInitial` is true.

Further improvements:

- When `controller.observe(el)` is called while disconnected, we don't immediately call `observer.observe(el)` , as this would cause a duplicate `.observe` call when reconnecting.
- The test suite would previously pass despite `skipInitial` not working properly. We add `await resizeComplete()` in a few places to wait two animation frames, so that we now test the behavior properly.

I considered whether `skipInitial` should just be removed altogether, but I think it's actually useful for Lit specifically: Without `skipInitial`, you'll often get an extraneous `updated()` lifecycle callback after the initial component render, as the `ResizeObserver` only triggers *after* the component has already been rendered. This behavior isn't very useful, and so `skipInitial` can help avoid the performance impact of this.

[1] Minor aside: According to the spec, this initial callback should only happen if the observed element is connected to the DOM. However, current browser implementations seem to trigger the callback [regardless](https://github.com/w3c/csswg-drafts/issues/11280). This distinction likely isn't important for `ResizeController`'s production code, as we only observe while the host component is connected. However, I took care to avoid relying on this behavior in the test suite, in case browsers change it to conform to the spec down the line.

-------------------

This is a fix for #4831.

I also tested this in the context of the app I'm working on, where it does resolve the issues I've had: With `skipInitial`, I'm now only getting a single `updated` call on initial render, and it reacts correctly to browser window resizes.

You might also want to merge this minor follow-up PR: #4852

Once this is merged, it would be lovely to get a new release of @lit-labs/observers so I can use it via npm.


packages/labs/observers/src/resize-controller.ts
```diff
@@ -65,13 +68,9 @@ export class ResizeController<T = unknown> implements ReactiveController {
...
   private _skipInitial = false;
-  /**
-   * Flag used to help manage calling the `callback` when observe is called
-   * in addition to when a mutation occurs. This is done to help setup initial
-   * state and is performed async by requesting a host update and calling
-   * `handleChanges` once by checking and then resetting this flag.
-   */
-  private _unobservedUpdate = false;
+  private _shouldSkipNextUpdate?: Set<Element> = undefined;
+  private _isConnected = false;
```
justinfagnani, line 72: I would remove this in favor of checking `this._host.isConnected`

packages/labs/observers/src/resize-controller.ts
```diff
@@ -38,9 +38,12 @@ export interface ResizeControllerConfig<T = unknown> {
...
   /**
-   * By default the `callback` is called without changes when a target is
-   * observed. This is done to help manage initial state, but this
-   * setup step can be skipped by setting this to true.
+   * The browser's `ResizeObserver` triggers immediately when a target is first
+   * observed. This behavior may be undesirable in a Lit component, as it may
+   * cause an extraneous `updated()` lifecycle callback after the initial
+   * render. If the performance impact of this is a concern, set `skipInitial`
+   * to true, which will cause the `ResizeController` to automatically ignore
+   * the first callback from the `ResizeObserver`.
```
justinfagnani, line 46: I think it'll read better in hover-over docs to have a short positive description of the effect of the flag, followed by a longer description of why.

Also "`updated()` lifecycle callback" is too specific, the whole lifecycle would be run again.

```suggestion
   * When `true`, ignores the first `ResizeController` callback for each
   * observed element.
   *
   * The browser's `ResizeObserver` triggers immediately when a target is first
   * observed. This behavior may be undesirable in a Lit component, as it may
   * cause an extraneous update lifecycle iteration after the initial
   * render. If the performance impact of this is a concern, set `skipInitial`
   * to true, which will cause the `ResizeController` to automatically ignore
   * the first callback from the `ResizeObserver`.
```

packages/labs/observers/src/resize-controller.ts
```diff
@@ -105,9 +103,23 @@ export class ResizeController<T = unknown> implements ReactiveController {
       return;
     }
     this._observer = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+      if (this._skipInitial) {
+        entries = entries.filter(
+          (entry) => !this._shouldSkipNextUpdate!.delete(entry.target)
```
justinfagnani, line 108: Ah my comment about a WeakSet wouldn't work if you need to get entries. You could maybe use WeakRefs. Seems a little complicated, but I would consider it unless we know for sure that we'll always get a ResizeObserver callback and be able to remove elements from the set.

Another option is to only skip the first callback from the ResizeObserver ever, regardless of the specific observed element. This would cover the `target: this` case and cases where `.observe()` is called in `firstUpdated()`.

Did you have a case where you needed to skip initial updates per separately observed element?

PR #4853, opened 2024-12-03
ci: add npm provenance

startracex:
I found this repository does not provide [provenance](https://docs.npmjs.com/generating-provenance-statements) for the packages published, so I added environment variables and permissions in ci. The provenance should be produced correctly.


.github/workflows/release.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
     permissions:
       contents: write  # for Git to git push
       pull-requests: write # to allow creating PR
+      id-token: write # for npm provenance
```
rictic, line 17: This new permission is necessary, according to the npm documentation: https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions

.github/workflows/release.yaml
```diff
@@ -66,6 +68,7 @@ jobs:
           HUSKY: 0
           GITHUB_TOKEN: ${{ secrets.LIT_ROBOT_AUTOMATION_PAT }}
           NPM_TOKEN: ${{ secrets.LIT_ROBOT_NPM_PUBLISH_ACCESS_TOKEN }}
+          NPM_CONFIG_PROVENANCE: true
```
rictic, line 71: This is the recommended way of generating provenance when not using `npm publish` directly: https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

rictic (approved):
Seems like a strict improvement, and I've verified that these changes follow the documentation.

rictic:
Thanks for the contribution!

Issue #4857, opened 2024-12-07
[@lit/react] Add React 19 support to reduce migration burden


maxpatiiuk:
### Should this be an RFC?

- [x] This is not a substantial change

### Which package is this a feature request for?

React (@lit/react)

### Description

`@lit/react` appears to work in React 19 in my limited testing. The only issue is that this line in package.json needs to be updated to add `|| 19`

https://github.com/lit/lit/blob/267d243aeec13a6f0e8184420919db4c519b8caf/packages/react/package.json#L199

I understand that React 19 supports custom elements, removing the need for wrappers. However, to smooth the ecosystem migration, it might be a good idea to support wrappers in React 19 too.

Use cases:
- Forcing people to refactor ooutur all wrappers usages and drop `@lit/react` before adopting React 19 is going to make the migration burden harder. Better let people gradually adopt React 19 and drop wrappers afterward.
- In a monorepo, some packages might update to React 19 before others
  - Example: updating storybook may require updating to React 19, while the code is not ready to
- Using wrappers may have better DX because IDEs can auto-import them on usages, and detect needless imports (assuming the package that uses `@lit/react` has separate entrypoint for each component wrapper, with a side effect import of the component inside) - not so for side-effect component imports.

### Alternatives and Workarounds

Use yarn/npm "resolutions" to force `@llit/react` to work with `@types/react@19`.

augustjk:
Indeed, we can and should widen the semver range for the types package.

We might even do better where we check `React.version` to see if it's >= 19 and pass all the props to React so that we don't have to do the imperative property setting in the `useLayoutEffect` for client rendering. This can prevent unwanted behavior where a component as its mounting momentarily has a default value for a property before it's set in the first layout effect.

We do have some special handling of element props for server rendering which we want to keep.

OvidijusParsiunas:
Hi. Just wanted to double check if _this issue_ is now redundant since the [following](https://github.com/lit/lit/pull/4865) PR has now been merged and deployed in [@lit/react](https://www.npmjs.com/package/@lit/react) version `1.0.7`. Just wanted to double check if there was anything else left to be done.

PR #4865, opened 2024-12-10
fix: Widen range for React types to include React 19

Artur-:
This allows using @lit/react with React 19

For #4857

### DIFF
--- a/.changeset/rare-cherries-bake.md
+++ b/.changeset/rare-cherries-bake.md
@@ -1,5 +1,0 @@
----
-"@lit/react": patch
----
-
-Widen range to support React 19
--- a/.changeset/rare-cherries-bake.md
+++ b/.changeset/rare-cherries-bake.md
@@ -1,0 +1,5 @@
+---
+"@lit/react": patch
+---
+
+Widen range to support React 19
--- a/package-lock.json
+++ b/package-lock.json
@@ -31843,7 +31843,7 @@
         "react-dom": "^18.2.0"
       },
       "peerDependencies": {
-        "@types/react": "17 || 18"
+        "@types/react": "17 || 18 || 19"
       }
     },
     "packages/reactive-element": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -31120,8 +31120,8 @@
         "uvu": "^0.5.6"
       },
       "peerDependencies": {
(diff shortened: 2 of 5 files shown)


kyubisation:
Should this also widen the dependency range in the ssr-react package?
https://github.com/lit/lit/blob/main/packages/labs/ssr-react/package.json#L54-L55

OvidijusParsiunas:
Hi, is there an estimated date when this will be reviewed? Thanks!

Artur-:
> Should this also widen the dependency range in the ssr-react package?

Looks like it should, I added it.

Artur-:
Is @justinfagnani still working on this project or is it somebody else?

.changeset/rare-cherries-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lit/react": patch
```
kyubisation, line 2: With the ssr react package also being updated, add it to the changeset

kyubisation:
I'll bring it up during the community call on Thursday.

.changeset/rare-cherries-bake.md
Artur-: Done, I think :)

kyubisation: Thanks 👍 
I think you can remove the original changeset, as a single changset file should be enough.

sorvell (approved):

package-lock.json
```diff
@@ -31843,7 +31843,7 @@
         "react-dom": "^18.2.0"
       },
       "peerDependencies": {
-        "@types/react": "17 || 18"
+        "@types/react": "17 || 18 || 19"
```
justinfagnani, line 31846: I would expect a few more changes here. Does the package lock need to be regenerated?

Artur-: The lock change for ssr-react was apparently missing, added what "npm install" updates

justinfagnani (approved):

augustjk (approved):
Thanks! I do want to explore detect React version and passing props directly to React for >= 19 as described in https://github.com/lit/lit/issues/4857#issuecomment-2524849452 but that can be done as a separate improvement.

PR #4869, opened 2024-12-17
Improve performance on trimOuterMarkers in Eleventy Lit plugin

zachleat:
Pretty conclusive that there is a hefty RegExp performance cost in the Eleventy Lit plugin, in the `trimOuterMarkers` function here: https://github.com/lit/lit/blob/267d243aeec13a6f0e8184420919db4c519b8caf/packages/labs/eleventy-plugin-lit/src/index.ts#L241-L245

Example Node CPU profile here (about 75% of this Eleventy project’s build is spent here):

image

The regular expressions seem to suggest that markers we’re looking for here are only found at the beginning and end of the content strings so I made some improvements using those assumptions.

Build time dropped significantly with the proposed changes (`20s` -> `~5s`)

cc @KonnorRogers @claviska

### DIFF
--- a/.changeset/metal-tigers-visit.md
+++ b/.changeset/metal-tigers-visit.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/eleventy-plugin-lit': patch
+---
+
+Improve performance on trimming outer markers from server rendered results.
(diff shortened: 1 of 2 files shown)


augustjk (approved):
Thank you! Excellent performance gains!

njfamirm:
It would be great if this package could be ready for production use with the help of these two packages.

PR #4870, opened 2024-12-17
[@lit/reactive-element] Provide ElementInternalHost mixin for attaching element internals

kyubisation:
Based on #4773 and my comment [here](https://github.com/lit/rfcs/pull/45#issuecomment-2504814462), I extracted the ElementInternal usage to its own mixin `ElementInternalsHost`.

```ts
import {ReactiveElement} from '@lit/reactive-element';
import {ElementInternalsHost, elementInternals} from '@lit/reactive-element/mixins/element-internals.js';

class MyElement extends ElementInternalsHost(ReactiveElement) {
  static override role = 'button';

  override connectedCallback(): void {
    super.connectedCallback();
    this[elementInternals].ariaPressed = 'false';
  }
}
```

### DIFF
--- a/.changeset/purple-foxes-turn.md
+++ b/.changeset/purple-foxes-turn.md
@@ -1,0 +1,6 @@
+---
+'@lit/reactive-element': minor
+'lit': minor
+---
+
+Provide ElementInternalHost mixin for attaching element internals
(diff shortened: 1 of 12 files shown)


kyubisation:
Discussed during community call. Preferred approach: https://lit.dev/playground/#gist=dd9d7c7bf865fb69b7fb0eea2a38a742

PR #4873, opened 2024-12-18
fix(reactive-element): preserve styles when moving elements between documents

### LINKED ISSUE #4868: Styles lost when moving elements between documents
### Should this be an RFC?

### Description

Hi. So I found out not long ago that you can't share CSSStyleSheet instances across document boundaries. Had a few issues when I tried to share a Lit component between the main page and iframes or document PIP windows. I was wondering if we could attempt any changes/additions that would detect and create compatible CSSStyleSheet instances for rendering contexts that required it if a LitElement happened to be instantiated there.

Here's a good workaround from @sorvell: https://lit.dev/playground/#gist=35429f7f8ff89216849c6d592e1dbc44
It's only ideal though if you're the author of the component. Things tend to get a bit hacky in my opinion if you're importing others' components.

paglobal:
Fixes #4868

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -172,9 +172,12 @@ export const adoptStyles = (
   styles: Array<CSSResultOrNative>
 ) => {
   if (supportsAdoptingStyleSheets) {
-    (renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) =>
-      s instanceof CSSStyleSheet ? s : s.styleSheet!
-    );
+    const sheets = styles.map(
+      (result) => (result as CSSResult).styleSheet ?? result
+    ) as CSSStyleSheet[];
+    const defaultView = renderRoot.ownerDocument.defaultView ?? window;
+    const viewSheets = sheetsForView(defaultView, sheets);
+    renderRoot.adoptedStyleSheets = viewSheets;
   } else {
     for (const s of styles) {
       const style = document.createElement('style');
@@ -187,6 +190,42 @@ export const adoptStyles = (
       renderRoot.appendChild(style);
     }
   }
+};
+
+const viewSheetCache = new WeakMap<
+  Window,
+  WeakMap<CSSStyleSheet, CSSStyleSheet>
+>();
+
+const sheetsForView = (view: Window, sheets: CSSStyleSheet[]) => {
+  let cache = viewSheetCache.get(view);
+  if (cache === undefined) {
+    viewSheetCache.set(view, (cache = new WeakMap()));
+  }
+
+  return sheets.map((sheet) => {
+    let viewSheet = cache.get(sheet);
+    if (viewSheet === undefined) {
+      if (
+        sheet instanceof
+        (view as unknown as {CSSStyleSheet: typeof CSSStyleSheet}).CSSStyleSheet
+      ) {
+        viewSheet = sheet;
+      } else {
+        const css = Array.from((sheet as CSSStyleSheet).cssRules).reduce(
+          (css, rule) => `${css} ${rule.cssText}`,
+          ''
+        );
+        viewSheet = new (
+          view as unknown as {CSSStyleSheet: typeof CSSStyleSheet}
+        ).CSSStyleSheet({baseURL: window.location.href});
+        viewSheet.replaceSync(css);
+      }
+      cache.set(sheet, viewSheet);
+    }
+
+    return viewSheet;
+  });
 };
 
 const cssResultFromStyleSheet = (sheet: CSSStyleSheet) => {
--- a/packages/reactive-element/src/test/css-tag_test.ts
+++ b/packages/reactive-element/src/test/css-tag_test.ts
@@ -9,6 +9,7 @@ import {
   CSSResult,
   unsafeCSS,
   supportsAdoptingStyleSheets,
+  adoptStyles,
 } from '@lit/reactive-element/css-tag.js';
 import {assert} from 'chai';
 
@@ -98,6 +99,64 @@ suite('Styling', () => {
       // document.body level.
       const bodyStyles = `${cssModule}`;
       assert.equal(bodyStyles.replace(/\s/g, ''), '.my-module{color:yellow;}');
+    });
+
+    test('`adoptStyles uses different cached `styleSheets` per document', async () => {
+      if (supportsAdoptingStyleSheets) {
+        const div = document.createElement('div');
+        document.body.appendChild(div);
+        const shadowRoot = div.attachShadow({mode: 'open'});
+        const styles = [
+          css`
+            div {
+              margin: 2px solid red;
+            }
+          `,
+          css`
+            div {
+              padding: 10px;
+            }
+          `,
+        ];
+        adoptStyles(shadowRoot, styles);
+        const adoptedStyleSheets_1 = shadowRoot.adoptedStyleSheets;
+        adoptStyles(shadowRoot, styles);
+        const adoptedStyleSheets_2 = shadowRoot.adoptedStyleSheets;
+        assert.notStrictEqual(adoptedStyleSheets_1, adoptedStyleSheets_2);
+        adoptedStyleSheets_1.forEach((_, index) => {
+          assert.strictEqual(
+            adoptedStyleSheets_1[index],
+            adoptedStyleSheets_2[index]
+          );
+        });
+        const iframe = document.createElement('iframe');
+        const html = '<body>Foo</body>';
+        document.body.appendChild(iframe);
+        iframe.contentWindow?.document.open();
+        iframe.contentWindow?.document.write(html);
+        iframe.contentWindow?.document.close();
+        iframe.onload = function () {
+          iframe.contentWindow?.document.body.appendChild(div);
+          adoptStyles(shadowRoot, styles);
+          const adoptedStyleSheets_3 = shadowRoot.adoptedStyleSheets;
+          adoptStyles(shadowRoot, styles);
+          const adoptedStyleSheets_4 = shadowRoot.adoptedStyleSheets;
+          assert.notStrictEqual(adoptedStyleSheets_3, adoptedStyleSheets_4);
+          adoptedStyleSheets_1.forEach((_, index) => {
+            assert.strictEqual(
+              adoptedStyleSheets_3[index],
+              adoptedStyleSheets_4[index]
+            );
+          });
+          assert.notStrictEqual(adoptedStyleSheets_2, adoptedStyleSheets_3);
+          adoptedStyleSheets_1.forEach((_, index) => {
+            assert.notStrictEqual(
+              adoptedStyleSheets_2[index],
+              adoptedStyleSheets_3[index]
+            );
+          });
+        };
+      }
     });
   });
 });
--- a/packages/reactive-element/src/test/css-tag_test.ts
+++ b/packages/reactive-element/src/test/css-tag_test.ts
@@ -119,10 +119,9 @@ suite('Styling', () => {
           `,
         ];
         adoptStyles(shadowRoot, styles);
-        const adoptedStyleSheets_1 = shadowRoot.adoptedStyleSheets;
+        const adoptedStyleSheets_1 = [...shadowRoot.adoptedStyleSheets];
         adoptStyles(shadowRoot, styles);
-        const adoptedStyleSheets_2 = shadowRoot.adoptedStyleSheets;
-        assert.notStrictEqual(adoptedStyleSheets_1, adoptedStyleSheets_2);
+        const adoptedStyleSheets_2 = [...shadowRoot.adoptedStyleSheets];
         adoptedStyleSheets_1.forEach((_, index) => {
           assert.strictEqual(
             adoptedStyleSheets_1[index],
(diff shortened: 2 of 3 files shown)


paglobal:
Not ready for review please.
Just realized I wasn't supposed to touch the `adoptStyles` function. I'm also supposed to implement an `adoptedCallback` in `ReactiveElement`.
Will make a commit and signal when it's ready for review.

Thanks.

packages/reactive-element/src/css-tag.ts
```diff
@@ -172,9 +172,12 @@ export const adoptStyles = (
...
 ) => {
   if (supportsAdoptingStyleSheets) {
-    (renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) =>
-      s instanceof CSSStyleSheet ? s : s.styleSheet!
-    );
+    const sheets = styles.map(
+      (result) => (result as CSSResult).styleSheet ?? result
+    ) as CSSStyleSheet[];
+    const defaultView = renderRoot.ownerDocument.defaultView ?? window;
+    const viewSheets = sheetsForView(defaultView, sheets);
```
sorvell, line 179: Let's not do this if `defaultView === window`?

packages/reactive-element/src/css-tag.ts
```diff
@@ -172,9 +172,12 @@ export const adoptStyles = (
   styles: Array<CSSResultOrNative>
 ) => {
   if (supportsAdoptingStyleSheets) {
```
sorvell, line 174: Factor this into `adoptStyleSheets` and return true if it actually adopts the sheets: L174-180

packages/reactive-element/src/css-tag.ts
```diff
@@ -189,6 +192,42 @@ export const adoptStyles = (
...
+  return sheets.map((sheet) => {
+    let viewSheet = cache.get(sheet);
+    if (viewSheet === undefined) {
+      if (
+        sheet instanceof
+        (view as unknown as {CSSStyleSheet: typeof CSSStyleSheet}).CSSStyleSheet
+      ) {
+        viewSheet = sheet;
+      } else {
+        const css = Array.from((sheet as CSSStyleSheet).cssRules).reduce(
```
sorvell, line 215: Use `cssResultFromStyleSheet` (defined in this module)

packages/reactive-element/src/test/css-tag_test.ts
```diff
@@ -99,5 +100,60 @@ suite('Styling', () => {
...
+        adoptedStyleSheets_1.forEach((_, index) => {
+          assert.strictEqual(
+            adoptedStyleSheets_1[index],
+            adoptedStyleSheets_2[index]
+          );
+        });
+        const iframe = document.createElement('iframe');
+        const html = '<body>Foo</body>';
+        document.body.appendChild(iframe);
+        iframe.contentWindow?.document.open();
```
sorvell, line 134: I don't think you need to do this. I think `iframe.contentDocument.body` should just exist.

sorvell (changes requested):
OK, I think I understand why you wanted to integrate this with `adoptStyles`. Since `adoptStyles`  is called via `createRenderRoot` which is by default in `connectedCallback`, if you create an element in the main document and immediately connect it to an iframe, it would need its styles to be fixed.

However, the fixup we need in `adoptedCallback` is not exactly the same:
1. In `createRenderRoot`, it's important for `adoptStyles` to fallback to creating `<style>` elements when adoptedStyleSheets are not supported.
2. in a`doptedCallback`, we can no-op when `adoptedStyleSheets` is not supported.

Given this, it may make sense to create a new method called `adoptStyleSheets` that only tries to adopt stylesheets and returns true if it does. Then call this in `ReactiveElement`'s `adoptedCallback`.

Also please add a test for using this in ReactiveElement.

packages/reactive-element/src/css-tag.ts
paglobal: Ok, I think I get where you're coming from. My only issue is that `window` may not be the origin of instantiation of the element.

paglobal: The coalescing was not present initially. I only added it because typescript kept yelling at me for the possibility of `defaultView` being null.

packages/reactive-element/src/css-tag.ts
paglobal: Will do.

packages/reactive-element/src/css-tag.ts
paglobal: Totally missed that. Thanks.

packages/reactive-element/src/test/css-tag_test.ts
paglobal: Oh I see. Affirmative.

paglobal:
I've implemented the requested changes although I'm unable to set up a test using `ReactiveElement` because of the `adoptedStyleSheets` issue I mentioned earlier. I've been at it for a bit. Could you please take a look at what I might be doing wrong? @sorvell

packages/reactive-element/src/css-tag.ts
```diff
@@ -189,6 +185,58 @@ export const adoptStyles = (
...
+ * Applies the given styles to a `shadowRoot` when `adoptedStyleSheets` is available and returns `true`.
+ * Otherwise, it does nothing and returns `false`.
+ */
+export const adoptStyleSheets = (
+  renderRoot: ShadowRoot,
+  styles: Array<CSSResultOrNative>
+) => {
+  if (supportsAdoptingStyleSheets) {
+    const sheets = styles.map(
+      (result) => (result as CSSResult).styleSheet ?? result
```
sorvell, line 198: To reduce size, let's try inlining the fix right here if and only if the sheet instanceof root.ownerDocument.defaultView.CSSStyleSheet is false. Let's also try not doing the caching.

sorvell (changes requested):
This adds `456 bytes` which I think it significantly too much. We need to try to make this as small as possible. You can run `checksize` on the repo to see this.

sorvell:
> For some reason, elements created with `ReactiveElement` don't seem to have `adoptedStyleSheets` on their `shadowRoot`. Is this right?

This should work. Check that the testing browser supports adoptedStylesheets.

justinfagnani:
To eliminate the size impact on ReactiveElement, and potentially make a solution available to any web component, I'd like to see us try helper functions instead:

```ts
import {prepareStyles, restoreStyles} from 'lit/move-to-document.js';

const el = document.createElement('my-element');
const pipWindow = await window.documentPictureInPicture.requestWindow();

prepareStyles(el);

pipWindow.document.body.append(el);

restoreStyles(el);
```

packages/reactive-element/src/css-tag.ts
paglobal: Sure...Will try that.

paglobal:
> This adds `456 bytes` which I think it significantly too much. We need to try to make this as small as possible. You can run `checksize` on the repo to see this.

Is that a cli tool we use?

paglobal:
> > For some reason, elements created with `ReactiveElement` don't seem to have `adoptedStyleSheets` on their `shadowRoot`. Is this right?
> 
> This should work. Check that the testing browser supports adoptedStylesheets.

That's where I'm confused. The testing browser does support it. In fact, it worked fine in the `adoptStyleSheets` test.

paglobal:
> To eliminate the size impact on ReactiveElement, and potentially make a solution available to any web component, I'd like to see us try helper functions instead:
> 
> ```ts
> import {prepareStyles, restoreStyles} from 'lit/move-to-document.js';
> 
> const el = document.createElement('my-element');
> const pipWindow = await window.documentPictureInPicture.requestWindow();
> 
> prepareStyles(el);
> 
> pipWindow.document.body.append(el);
> 
> restoreStyles(el);
> ```

Given the code delta, I guess we could just do helpers. Then we could leave a note in the docs suggesting to element authors to toss it into their `adoptedCallback`s to enable cross-document style preservation. Similar to what @sorvell has [here](https://discord.com/channels/1012791295170859069/1318220115040534639/1318366786315620402).

packages/reactive-element/src/css-tag.ts
paglobal: In light of @justinfagnani's suggestion, should I scrap this, or should I go on?

paglobal:
Does the code size check only apply to the core lit libraries?

sorvell:
Yeah, I think this isn't worth 0.5KB and we should try a helper. It would be nice to create in such a way that an author could opt into using the/a helper in `adoptedCallback` if desired.

> Does the code size check only apply to the core lit libraries?

Yeah.

justinfagnani:
Elements that wanted to opt into this behavior would have to also call a helper after their styles are set up. 

I think it'd be best if the helpers worked on any element with a shadow root, by accessing the adopted stylesheets.

paglobal:
> Yeah, I think this isn't worth 0.5KB and we should try a helper. It would be nice to create in such a way that an author could opt into using the/a helper in `adoptedCallback` if desired.
> 
> > Does the code size check only apply to the core lit libraries?
> 
> Yeah. 
> 
> 

Oh okay sure. No explicit change set required then? 

Also, where would this helper be implemented given that it's not supposed to be in core?

paglobal:
> Elements that wanted to opt into this behavior would have to also call a helper after their styles are set up. 
> 
> I think it'd be best if the helpers worked on any element with a shadow root, by accessing the adopted stylesheets.

Oh okay. Like in the example you provided? So that users could opt-in on their own if element authors don't? If so, I think that'd be quite helpful. The only problem I see there is closed `shadowRoot`.

sorvell:
@paglobal 
 > Also, where would this helper be implemented given that it's not supposed to be in core?

It can be in the reactive-element package (part of the core) as long as it's not auto-included and must be opted into, like the decorators.

> Like in the example you provided?

Here is an [example of something that might work](https://lit.dev/playground/#gist=3970bd087993d5f1140a4d9f88da9344). 

1. It provides a `CrossDocument` mixin which captures styles (on the element's shadow inclusive subtree) on connection and restores them on adoption.
2. It also provides `prepareStyles` and `restoreStyles` so they can be used generically.
3. It also makes cross document moves work on Firefox where otherwise elements break because they lose their custom element prototype.

paglobal:
> @paglobal
> 
> > Also, where would this helper be implemented given that it's not supposed to be in core?
> 
> It can be in the reactive-element package (part of the core) as long as it's not auto-included and must be opted into, like the decorators.
> 
> > Like in the example you provided?
> 
> Here is an [example of something that might work](https://lit.dev/playground/#gist=3970bd087993d5f1140a4d9f88da9344).
> 
> 1. It provides a `CrossDocument` mixin which captures styles (on the element's shadow inclusive subtree) on connection and restores them on adoption.
> 2. It also provides `prepareStyles` and `restoreStyles` so they can be used generically.
> 3. It also makes cross document moves work on Firefox where otherwise elements break because they lose their custom element prototype.

Nice! I didn't even know about the firefox thing. Should I go ahead and implement this in reactive element with some tests?

sorvell:
> Nice! I didn't even know about the firefox thing. Should I go ahead and implement this in reactive element with some tests?

In the package yes (not in r`eactive-element.ts` directly), as part of the `cross-document.ts` helper. Let's make a separate set of tests for that.

paglobal:
> @paglobal
> 
> > Also, where would this helper be implemented given that it's not supposed to be in core?
> 
> It can be in the reactive-element package (part of the core) as long as it's not auto-included and must be opted into, like the decorators.
> 
> > Like in the example you provided?
> 
> Here is an [example of something that might work](https://lit.dev/playground/#gist=3970bd087993d5f1140a4d9f88da9344).
> 
> 1. It provides a `CrossDocument` mixin which captures styles (on the element's shadow inclusive subtree) on connection and restores them on adoption.
> 2. It also provides `prepareStyles` and `restoreStyles` so they can be used generically.
> 3. It also makes cross document moves work on Firefox where otherwise elements break because they lose their custom element prototype.

I was going through the code in the playground once more in preparation for writing the implementation and then I came across something. We have a problem if the element doesn't start in the main document. See [here](https://lit.dev/playground/#gist=1a210463e34361d576607d4ba29c3a26).

paglobal:
Should we call `restoreStyles` right after `prepareStyles` in `connectedCallback`?

sorvell:
> We have a problem if the element doesn't start in the main document. See [here](https://lit.dev/playground/#gist=1a210463e34361d576607d4ba29c3a26).

I think the only way to fix this generally is with an intrusive patch, which may or may not be worth it. The attempt here was to makes elements we don't control (but possibly only Lit elements) functional cross-document. The problem is that any element could choose to create a shadowRoot and attach a (wrong document) stylesheet at any time, say in the constructor. We can't easily mitigate that from a parent element.

Here is a version that [applies a patch to make it work](https://lit.dev/playground/#gist=e9531fd350484c819de3491124c84b89). It uses a proxy for `adoptedStyleSheets` and patches `attachShadow` to always use this proxy. If we go this direction, we definitely need some robust testing.

paglobal:
> > We have a problem if the element doesn't start in the main document. See [here](https://lit.dev/playground/#gist=1a210463e34361d576607d4ba29c3a26).
> 
> I think the only way to fix this generally is with an intrusive patch, which may or may not be worth it. The attempt here was to makes elements we don't control (but possibly only Lit elements) functional cross-document. The problem is that any element could choose to create a shadowRoot and attach a (wrong document) stylesheet at any time, say in the constructor. We can't easily mitigate that from a parent element.
> 
> Here is a version that [applies a patch to make it work](https://lit.dev/playground/#gist=e9531fd350484c819de3491124c84b89). It uses a proxy for `adoptedStyleSheets` and patches `attachShadow` to always use this proxy. If we go this direction, we definitely need some robust testing.

Oh wow. It definitely works, but you're right. We will need some robust testing if we're going to go in this direction.

paglobal:
Hi guys. So I've been thinking about this issue and I wanted to suggest we tackle the issue in two parts:

- Make sure that Lit elements can be used in different documents at the same time without erroring (I anticipate this could be implemented in `adoptStyles` or somewhere else suitable within Lit core pretty trivially without much code delta)
- Enable Lit elements to be moved across documents while preserving their styles(this could be implemented in a utility outside of Lit core like where we left off but has proven to be less trivial and requiring of much more thought)

As such, I would like to close this PR in favor of handling the two parts in two separate PRs. Any thoughts please?

Issue #2631, opened 2022-03-11
[SSR] Can not use css files


daKmoR:
# Description

I can not use css files to style my component as css import assertions error on node.

# Steps to Reproduce

1. Write this code

👉  `style.css`

```css
:host {
  background: green;
}
```

👉  `index.js`

```ts
import { render } from "@lit-labs/ssr/lib/render-with-global-dom-shim.js";
import { html, LitElement, css } from 'lit';

import style from './style.css' assert { type: 'css' };

class MyElement extends LitElement {
  static styles = [style, css`
    :host {
      background: red;
    }
  `];

  render() {
    return html`
      <h2>MyElement</h2>
    `;
  }
}

customElements.define('my-element', MyElement);

const template = ({ title }) => html`
  <p>${title}</p>
  <my-element></my-element>
`;

const ssrResult = render(template({ title: "hello world" }));
let ssrString = '';

for (const s of ssrResult) {
  ssrString += s;
}

console.log(ssrString);
```

2. run `node index.js`

## Expected Results

SSR Result logged to console

## Actual Results

```
node index.js
node:internal/errors:465
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".css" for /Users/html/test-ssr/style.css
    at new NodeError (node:internal/errors:372:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:56:11)
    at defaultGetFormat (node:internal/modules/esm/get_format:83:38)
    at defaultLoad (node:internal/modules/esm/load:21:20)
    at ESMLoader.load (node:internal/modules/esm/loader:363:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:282:22)
    at new ModuleJob (node:internal/modules/esm/module_job:66:26)
    at ESMLoader.#createModuleJob (node:internal/modules/esm/loader:301:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:260:34)
    at async ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:81:21) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}

Node.js v17.6.0
```

## Browsers Affected

- Tested in node 16 & 17

justinfagnani:
[...]

PR #4879, opened 2024-12-20
[labs/ssr-dom-shim] Implement limited shim for CSSStyleSheet and CSS loader for Node.js

### LINKED ISSUE #4862: expose minimal `CSSStyleSheet` class in SSR DOM Shim
### Should this be an RFC?

SSR (@lit-labs/ssr)

### Description

Coming out of discussions in https://github.com/lit/lit/issues/2631#issuecomment-1065400805 and from the recent Lit engineering call, the topic of splitting the scope off for _just_ providing a DOM Shim for `CSSStyleSheet` (distinct from handling how to actually load a CSS file in a non-browser runtime) was proposed.   This would allow for the usage of CSS Module Scripts with custom elements when server-rendering.

You can see the current behavior in [this reproduction repo](https://github.com/thescientist13/lit-ssr-css-modules)
```css
/* greeting.css */
p { 
  color: blue;
}
```

```js
// greeting.js
import { html, css, LitElement } from 'lit';
import sheet from './greeting.css' with { type: 'css' }

export class SimpleGreeting extends LitElement {
  static styles = [ sheet ];

  // ...
}

customElements.define('simple-greeting', SimpleGreeting);
```

And the output from running the `build` script
```sh
SSR Stylesheet detected...
file:///Users/owenbuckley/Workspace/github/lit-ssr-css-modules/components/greeting/greeting.css:2
        const sheet = new CSSStyleSheet();sheet.replaceSync(`p { 
                      ^

ReferenceError: CSSStyleSheet is not defined
    at file:///Users/owenbuckley/Workspace/github/lit-ssr-css-modules/components/greeting/greeting.css:2:23
    at ModuleJob.run (node:internal/modules/esm/module_job:234:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:473:24)
    at async file:///Users/owenbuckley/Workspace/github/lit-ssr-css-modules/register.js:5:1
```

----

The proposal for a DOM Shim itself would effectively be to stub out no-ops for the main methods of `CSSStyleSheet`, so as to at least not fail in an SSR context.  (more work to improve their capabilities could be made over time if so desired)
```js
class CSSStyleSheet {
  insertRule() { }
  deleteRule() { }
  replace() { }
  replaceSync() { }
}
```

I thought I could just shim this in from first party code or at least try and patch it through _node_modules, but it still produced the same error, so I assume it has something to do with how Lit+SSR is registering / shielding globals?

```js
class CSSStyleSheet {
  insertRule() { }
  deleteRule() { }
  replace() { }
  replaceSync() { }
}

globalThis.CSSStyleSheet = globalThis.CSSStyleSheet ?? CSSStyleSheet;
```

Interestingly, I saw that [`CSSStyleSheet`](https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/dom-shim.ts#L53) was defined in the DOM Shim, yet we still get `CSSStyleSheet is undefined`? 🤔

kyubisation:
This PR adds a shim for CSSStyleSheet (and associated classes) and a CSS loader hook for Node.js.

**Limitation**
The CSS loader cannot be registered synchronously (i.e. if you import LitElement/the shim and a .css file from the same file, it will break). Due to this the loader needs to be registered via `--import @lit-labs/ssr-dom-shim/css-loader.js` via Node.js CLI or imported before dynamically importing components.

Closes #4862
Relates to #2631


packages/labs/ssr-dom-shim/src/lib/css-loader.ts
```diff
@@ -0,0 +1,42 @@
...
+ * When an attempt is made to import a CSS file/module, the CSS content
+ * is read and added to a CSSRule, which is provided in a CSSStyleSheet
+ * instance.
+ *
+ * https://nodejs.org/api/module.html#loadurl-context-nextload
+ */
+export const load: LoadHook = (url, context, nextLoad) => {
+  if (context.importAttributes.type === 'css') {
+    const filePath = fileURLToPath(url);
+    const css = readFileSync(filePath, 'utf8');
```
thescientist13, line 32: fwiw, this would also work as well, since Node *fs** APIs support an instance of a URL, though maybe the constructor overhead isn't worth it?  But also one less NodeJS specific API I suppose (`fileURLToPath`) 🤷 
```js
const css = await fs.readFile(new URL(url), 'utf-8');
```

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * When an attempt is made to import a CSS file/module, the CSS content
+ * is read and added to a CSSRule, which is provided in a CSSStyleSheet
+ * instance.
+ *
+ * https://nodejs.org/api/module.html#loadurl-context-nextload
+ */
+export const load: LoadHook = (url, context, nextLoad) => {
+  if (context.importAttributes.type === 'css') {
```
justinfagnani, line 30: Could we add a warning if there's a `.css` import without the correct import attribute? And can loaders compose so that we only warn if that import otherwise failed?

Also, what happens if Node gains support for CSSStyleSheet natively and this loader is registered? Is there a way to detect that case?

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
justinfagnani: I would prefer async fs APIs if they work. Will Node use the same thread for multiple imports if it yields?

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
```diff
@@ -0,0 +1,42 @@
...
+    const filePath = fileURLToPath(url);
+    const css = readFileSync(filePath, 'utf8');
+    const code = `
+      import {CSSStyleSheet} from '@lit-labs/ssr-dom-shim';
+      const sheet = new CSSStyleSheet();
+      sheet.replaceSync(\`${css.replaceAll('`', '\\`')}\`);
```
justinfagnani, line 36: I think we need to do more escaping here. `${...}` is problematic at least, and there are some octal escapes that need handling as well, so we might want to escape `\`. Needs tests.

packages/labs/ssr-dom-shim/src/test/types.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.css' {
```
justinfagnani, line 1: Let's link to https://github.com/microsoft/TypeScript/issues/46689 and https://github.com/microsoft/TypeScript/issues/46135 so we can hopefully remove this in the future.

justinfagnani:
> **Limitation** The CSS loader cannot be registered synchronously

This seems expected. We can't rely on import ordering because while eval happens in order, fetching and parsing happen in parallel.

> **To discuss**
> 
> * Should CSSStyleSheet be added to the global context?

Not by default, it should be installed with the other DOM globals.

> * How to proceed with the loader?
> * Should the logic in packages/reactive-element/src/css-tag.ts be adapted?

What do you mean with these last two points?

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
kyubisation: I'll change it as suggested.
However, it seems the future of hooks is synchronous: https://github.com/nodejs/node/issues/52219
But we can cross that bridge once that is widely available and/or the recommended approach.

kyubisation: "Will Node use the same thread for multiple imports if it yields?"
My understanding is that Node.js runs hooks on a singular, separate thread.

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
kyubisation: Yes, but only in the load hook. I'll implement a suggestion.

I am unaware of a possibility to detect what import attributes are natively supported. However, before registering our loader we could try a dynamic import of a css file and only register it, if that fails. That seems a bit hacky to me, but I am open to do that, if you would prefer that approach.

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
kyubisation: An alternative approach would be to do the `readFile` in the code template. This should circumvent the need to escape the value.
What do you think?

packages/labs/ssr-dom-shim/src/index.ts
```diff
@@ -4,12 +4,24 @@
...
+  CSSRule,
+  CSSRuleList,
+  CSSStyleSheet,
+  MediaList,
+  StyleSheet,
+} from './lib/css.js';
+
+// We want to provide CSSStyleSheet in the global scope, as
+// `new CSSStyleSheet()` can be called anywhere in consumer code.
+globalThis.CSSStyleSheet ??= CSSStyleSheetShim;
```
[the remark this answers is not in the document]
kyubisation: I'm not opposed to this, but is the philosophy behind this documented somewhere? My first instinct would be to make it "just work", as moving it into `getWindow()` requires consumers to first find and use this functionality, in contrast to e.g. customElements being globally available.

packages/labs/ssr-dom-shim/src/css-loader.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @example
+ *
+ * ```ts
+ *  import styles from 'my-styles.css' with {type: 'css'};
+ * ```
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with
+ * @see https://nodejs.org/api/module.html#customization-hooks
+ */
+register('./lib/css-loader.js', {parentURL: import.meta.url});
```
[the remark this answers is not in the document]
kyubisation: It is active immediately (with the limitation mentioned above). I can add it accordingly.

kyubisation:
> * How to proceed with the loader?

How to provide/integrate the loader. I think you answered this by suggesting to add it to `install-global-dom-shim.ts`.

> * Should the logic in packages/reactive-element/src/css-tag.ts be adapted?

Should usage of `CSSStyleSheet` be enabled in SSR?

packages/labs/ssr-dom-shim/src/css-loader.ts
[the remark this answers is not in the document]
kyubisation: `install-global-dom-shim.ts` is only used in deprecated code and not documented anywhere.
Is there any strategy for this in the future?

kyubisation:
I changed the terminology from `loader` to `hook` to align with the terminology used by Node.js

packages/labs/ssr-dom-shim/src/index.ts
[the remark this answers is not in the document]
justinfagnani: The globals are defined `getWindow()` so that they can be installed into different context than the one that imports the shim. Unless and until we remove support for VM modules we need to keep that ability. `install-global-dom-shim.js` exists for when you want to automatically install the globals into the importing context.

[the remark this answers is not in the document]
justinfagnani: btw, the `customElements` global is installed in `reactive-element.ts` (in addition to `installWindowOnGlobal()`) in the Node build: https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L60

We _could_ add that there, but the reason why `customElements` is there is because it's so common and we're just trying to make the basic element load in Node without errors. `new CSSStyleSheet()` is much less comment, and importing CSS less so since it's only supported in Chrome.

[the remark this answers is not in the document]
kyubisation: Yes, I noticed that. But yeah, that is reasonable.

kyubisation:
Hello @thescientist13 
Can you check whether this PR meets the criteria to close #4862?
I have the thought to create a follow-up PR to add an entry/import module `@lit-labs/ssr/dom-shim.js`, which would install the hook and the DOM shim globals. It could be used like `node --import @lit-labs/ssr/dom-shim.js my-script.js`. Would that help for your use case?

packages/labs/ssr-dom-shim/package.json
```diff
@@ -20,6 +20,10 @@
     ".": {
       "types": "./index.d.ts",
       "default": "./index.js"
+    },
+    "./css-hook.js": {
```
thescientist13, line 24: Just a heads up that there is an option to set an export condition
https://nodejs.org/api/module.html#enabling
> Where some-package has an ["exports"](https://nodejs.org/api/packages.html#exports) field defining the `/register` export to map to a file that calls `register()`, like the following _register-hooks.js_ example.

edit: though maybe this is what you indicated would be the follow up PR.

thescientist13:
Hey @kyubisation thanks so much for all your great work on this!

> Can you check whether this PR meets the criteria to close https://github.com/lit/lit/issues/4862?

It seems like it would, but I haven't had a case to use the `installWindowOnGlobal` yet in a Lit+SSR project so my experience / usage there is limited.

Inferring from the changes in this PR though, would I just simply need to add these lines to the top of my script in the [repro repo](https://github.com/thescientist13/lit-ssr-css-modules/blob/master/index.js)?
```js
// add these lines
import '@lit-labs/ssr-dom-shim/css-hook.js';
import {installWindowOnGlobal} from './dom-shim.js';
installWindowOnGlobal();

// rest of the code as normal
import fs from 'node:fs/promises';
import { html } from 'lit';
import { render } from '@lit-labs/ssr';
import { collectResult } from '@lit-labs/ssr/lib/render-result.js';
import './components/greeting/greeting.js';

//...
```

> I have the thought to create a follow-up PR to add an entry/import module @lit-labs/ssr/dom-shim.js, which would install the hook and the DOM shim globals. It could be used like node --import @lit-labs/ssr/dom-shim.js my-script.js. Would that help for your use case?

Yeah, I think that would be helpful for composition purposes, so consumers can cherry-pick whatever parts they want of this implementation, which I think is really nice for folks building meta-frameworks around Lit / Lit+SSR.

----

Left some other thoughts in the PR, feel free to take or leave them.

packages/labs/ssr-dom-shim/src/lib/css-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * https://nodejs.org/api/module.html#loadurl-context-nextload
+ */
+export const load: LoadHook = async (url, context, nextLoad) => {
+  if (context.importAttributes.type === 'css') {
+    // Convert the path to base64 to prevent any special characters
+    // from being falsely interpreted as code.
+    const base64url = btoa(url);
+    const code = `
+      import {readFile} from 'node:fs/promises';
```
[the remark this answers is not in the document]
kyubisation: I just wanted to avoid having to escape the string, as the example you linked to would break if `contents` contained a backtick.
I'll try to think of an alternative.

packages/labs/ssr-dom-shim/src/lib/css-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+      import {CSSStyleSheet} from '@lit-labs/ssr-dom-shim';
+      const url = new URL(atob('${base64url}'));
+      const css = await readFile(url, 'utf-8');
+      const sheet = new CSSStyleSheet();
+      sheet.replaceSync(css);
+      export default sheet;
+    `;
+    return {format: 'module', shortCircuit: true, source: code};
+  } else if (new URL(url).pathname.endsWith('.css')) {
```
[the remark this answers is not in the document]
kyubisation: @justinfagnani You asked for the warning. What is your opinion on this?

packages/labs/ssr-dom-shim/package.json
kyubisation: > edit: though maybe this is what you indicated would be the follow up PR?

Yes, basically. But to be able to use it from `@lit-labs/ssr` I already need a specific exports entry in `@lit-labs/ssr-dom-shim`. This also allows it to be used directly from this package, if so desired.

kyubisation:
> Inferring from the changes in this PR though, would I just simply need to add these lines to the top of my script in the [repro repo](https://github.com/thescientist13/lit-ssr-css-modules/blob/master/index.js)?
> 
> ```js
> // add these lines
> import '@lit-labs/ssr-dom-shim/css-hook.js';
> import {installWindowOnGlobal} from './dom-shim.js';
> installWindowOnGlobal();
> 
> // rest of the code as normal
> import fs from 'node:fs/promises';
> import { html } from 'lit';
> import { render } from '@lit-labs/ssr';
> import { collectResult } from '@lit-labs/ssr/lib/render-result.js';
> import './components/greeting/greeting.js';
> 
> //...
> ```

No, that would still break, as the imports to the hook and the component (which in turn imports the css file) are on the same "level".
You would either to dynamically import the/any component (e.g. `await import('./components/greeting/greeting.js');`) or register the hook via Node.js CLI/environment variable paramters (e.g. `node --import @lit-labs/ssr-dom-shim/css-hook.js my-script.js`, which I think you already adapted [here](https://github.com/thescientist13/lit-ssr-css-modules/blob/master/package.json#L16)).

packages/labs/ssr-dom-shim/src/lib/css-loader.ts
kyubisation: I have changed the code to just use JSON.stringify. From my limited understanding this should work in any use case.

packages/labs/ssr-dom-shim/src/css-hook.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @example
+ *
+ * ```ts
+ *  import styles from 'my-styles.css' with {type: 'css'};
+ * ```
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with
+ * @see https://nodejs.org/api/module.html#customization-hooks
+ */
+register('./lib/css-hook.js', {parentURL: import.meta.url});
```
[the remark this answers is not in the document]
thescientist13: yeah, that seems fine.  not a big deal either way

thescientist13:
> No, that would still break, as the imports to the hook and the component (which in turn imports the css file) are on the same "level".
You would either to dynamically import the/any component (e.g. await import('./components/greeting/greeting.js');) or register the hook via Node.js CLI/environment variable paramters (e.g. node --import @lit-labs/ssr-dom-shim/css-hook.js my-script.js, which I think you already adapted [here](https://github.com/thescientist13/lit-ssr-css-modules/blob/master/package.json#L16)).

Ah, OK.  I'm reviewing the README a bit with more your examples and think I get it now.  I assume this is because of the requirement to [exclude `CSSStyleSheet` from the SSR DOM Shim](https://github.com/lit/lit/pull/4879#discussion_r1894340259)?  I suppose if that was not the case, we wouldn't have to "externalize" this to userland like this?  Would certainly like to see if we can discuss this a bit during the next call.  Maybe we could change the [position on this](https://github.com/lit/lit/pull/4879#discussion_r1894710186) once [(CSS) Import Attributes reach full interop](https://github.com/web-platform-tests/interop/issues/703)? 🤞

packages/labs/ssr-dom-shim/src/lib/css-hook.ts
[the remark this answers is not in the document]
justinfagnani: I don't care that much. I don't think many other CSS loader hooks are out there that would handle the else case, so it'll just fail when Node sees the import.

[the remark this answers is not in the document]
kyubisation: I think we'll leave it in for now. If that pops up in a consuming framework, we can still remove it.

kyubisation:
@justinfagnani @thescientist13 I implemented a check whether the runtime environment support importing CSS files directly and whether hooks can be registered. Feel free to provide feedback, if you have concerns.

thescientist13 (approved):
Happy to test this out when it is ready / published 👍

Issue #4882, opened 2024-12-22
[labs/ssr] Move more DOM shims to ssr-dom-shim package


justinfagnani:
`@lit-labs/ssr/lib/dom-shim.ts` still includes some shims that should probably have been moved to `@lit-labs/ssr-dom-shim`, like ShadowRoot, Document, and CSSStyleSheet: https://github.com/lit/lit/blob/25962bf58f33f32abef6487689438bf095780b63/packages/labs/ssr/src/lib/dom-shim.ts#L39

https://github.com/lit/lit/pull/4879 is moving CSSStyleSheet already.

kyubisation:
- #4893
- #4894

PR #4893, opened 2025-01-15
[labs/ssr] Implement ShadowRoot and document/window in event path

### LINKED ISSUE #5178: Event target stack incorrect when `wa-page` is imported

SSR (@lit-labs/ssr)

### Description

I render a page where my elements are sharing some data using `@lit-labs/context`. When I put them in Web Awesome `wa-page`, I found that the context stepped propagating.

I traced it to issue with event targeting where the event target stack is incorrect, thus causing the event to never reach the right parent

### Reproduction

https://gist.github.com/tpluscode/3e5bdc0768489f8376379b055ddcfe6f

### Workaround

I found none

### Is this a regression?

### Affected versions

lit@3.3.1 @lit-labs/ssr@3.3.1 @lit-labs/context@0.5.1

- Brave 1.84.139 Chromium: 142.0.7444.163
- node 20.19.4
- npm 10.8.2
- macOS 15.3.2

kyubisation:
This PR implements ShadowRoot and document/window in the event path.
For this to work, `Node.getRootNode()` was implemented as part of the DOM shim, which allows to retrieve the host (if the element is in a Shadow DOM) or document. Due to this, Document and document were moved into the `@lit-labs/ssr-dom-shim` from the `@lit-labs/ssr` (relates to #4882).

This also fixes #5178, by not short circuiting on slots with duplicate names.


kyubisation:
@sorvell In the initial SSR events PR you asked for a tree representation for the event path; I tried to add this here: https://github.com/lit/lit/blob/f4d202c8e1391459a67a43a927f1e015b4bc21e2/packages/labs/ssr-dom-shim/src/lib/events.ts
Is this what you were asking for?

justinfagnani:
@kyubisation 

> Should Window/window also be part of the composedPath, as it is in the browser?

I think it should be, in case someone is doing path manipulation by index, and assuming that window is there. We do have a window object, we just don't put `window` on global.

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -4,289 +4,16 @@
...
-) => (typeof options === 'boolean' ? options : (options?.capture ?? false));
+type EventInterface = Event;
 
 // Event phases
 const NONE = 0;
 const CAPTURING_PHASE = 1;
 const AT_TARGET = 2;
 const BUBBLING_PHASE = 3;
 
-// Shim the global EventTarget object
```
justinfagnani, line 42: Are you using EventTarget from Node now?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -1256,3 +1233,18 @@ function isJavaScriptScriptTag(node: Element | Template): boolean {
   const willExecute = !safeTypeSeen;
   return willExecute;
 }
+
+function addEventTargetConfiguration(
```
justinfagnani: This function name sounds more generic than its behavior, which is slot specific. Can we name something like `addElementToEventPath()`?

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -1256,3 +1233,18 @@ function isJavaScriptScriptTag(node: Element | Template): boolean {
...
+  element: HTMLElement & Partial<EventTargetShimMeta>,
+  renderInfo: RenderInfo
+): void {
+  const eventTarget = renderInfo.eventTargetStack.at(-1)!;
+  const slotName = renderInfo.slotStack.at(-1);
+  element.__host = renderInfo.customElementHostStack.at(-1)?.element;
+  const eventTargetParent =
```
justinfagnani: A comment here would be good

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -1256,3 +1233,18 @@ function isJavaScriptScriptTag(node: Element | Template): boolean {
   const willExecute = !safeTypeSeen;
   return willExecute;
 }
+
+function addEventTargetConfiguration(
+  element: HTMLElement & Partial<EventTargetShimMeta>,
+  renderInfo: RenderInfo
+): void {
+  const eventTarget = renderInfo.eventTargetStack.at(-1)!;
+  const slotName = renderInfo.slotStack.at(-1);
```
justinfagnani, line 1236: Why aren't slots already on the eventTargetStack? That would simplify the following logic.

packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -4,289 +4,16 @@
...
- * This is not fully spec compliant (e.g. validation),
- * but should work well enough for our use cases.
- *
- * @see https://dom.spec.whatwg.org/#eventtarget
- * @see https://dom.spec.whatwg.org/#event
- * @see https://dom.spec.whatwg.org/#customevent
- */
+/* Adapted from Node.js https://github.com/nodejs/node/blob/main/lib/internal/event_target.js */
 
-export interface EventTargetShimMeta {
```
justinfagnani, line 18: This is still being imported elsewhere, where is it defined now?

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: No, we still have our own EventTarget implementation, but it is now in src/index.ts with the whole class chain.
We could theoretically patch the global EventTarget object, but I worry about side effects.

packages/labs/ssr-dom-shim/src/lib/events.ts
kyubisation: Moved to src/index.ts in the shim package.
The motivation for moving `EventTarget` and related symbols into src/index.ts, was because the event path and the `Node.getRootNode` implementation needed references to the document and window objects.
Due to that these symbols were originally defined in the src/lib/events.ts file (to avoid circular dependencies), which would not be where I would expect to find the declaration of this symbol.

packages/labs/ssr/src/lib/render-value.ts
kyubisation: I'm afraid I don't remember why I did it this way, but that should be possible. I'll try to adapt the logic. 👍

packages/labs/ssr/src/lib/render-value.ts
kyubisation: I added documentation to the function.
Do you think the current explanation works well enough?

packages/labs/ssr/src/lib/render-value.ts
justinfagnani: I think it might be worth a TODO to see if you can add slots to the eventTargetStack, or document why that can't be. It could be that because we don't process DOM in flattened tree order - we emit shadow roots, then light DOM, so we don't process a slot and then the elements assigned to it - we can't build the stack that way.

But maybe we can - if we know here the slot we should use, then should we not just update the event target stack here? The slot should be on the path for elements further down the stack too, yes? ie, this case:

```html
<el-a>
  <template shadowrootmode="open">
    <el-b><slot></slot></el-b>
  </template>
  <el-c><el-d></el-d></el-d>
</el-a>
```

In this case `<el-b>` needs to be on the stack for `<el-d>`. Will that work because `<el-c>`'s __eventTargetParent is set to the slot?

justinfagnani (approved):

packages/labs/ssr/src/lib/render-value.ts
kyubisation: This currently works.
From what I remember this is because slots are orthogonal to the DOM tree.
Looking at your example:
> el-a is rendered
-->  renders shadow DOM
----> renders el-b
----> renders slot (which has parent el-b)
----> slot is removed from stack
----> el-b is removed from stack
--> el-a shadow DOM has finished rendering
--> children of el-a are rendered and they need to dynamically check to which slot they are associated to get the associated event path

Due to this slots need to be handled separately from the eventTargetStack. I'll try to document this in the `__slot` property.

PR #4894, opened 2025-01-16
[labs/ssr-dom-shim] Add DOM shim implementation for `MutationObserver`, `ResizeObserver` and `IntersectionObserver`

kyubisation:
Relates to #4882

### DIFF
--- a/packages/labs/ssr-dom-shim/src/lib/observers.ts
+++ b/packages/labs/ssr-dom-shim/src/lib/observers.ts
@@ -1,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * This is a limited implemenation of the observer family of classes.
+ */
+
+type MutationObserverInterface = MutationObserver;
+
+/**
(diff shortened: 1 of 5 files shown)


packages/labs/ssr-dom-shim/src/lib/observers.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * This is a limited implemenation of the observer family of classes.
```
justinfagnani, line 8: It'd be good to say what the limitations are. Do the observers ever fire?

justinfagnani (approved):

packages/labs/ssr-dom-shim/src/lib/observers.ts
kyubisation: No, they are just noop's.
I will document that. 👍

kyubisation: Will additionally be addressed in lit.dev: https://lit.dev/docs/ssr/dom-emulation/

PR #4895, opened 2025-01-17
[reactive-element] Adds skipping initial attribute reflection for default values.

sorvell:
Adds support for https://github.com/lit/rfcs/issues/36 and helps facilitate https://github.com/shoelace-style/shoelace/discussions/2339.

Spawning an initial attribute can cause a NextJS hydration error, see [repro](https://codesandbox.io/p/devbox/nostalgic-monad-3ps3dz?workspaceId=ws_MXuotjJLEi62fRTssWSgdA).

To use, set `defaultValue` in property options, and avoid setting the default in the field/accessor.

```ts
@property({reflect: true, defaultValue: 'foo'})
variant: string;
```

Previously, this would spawn a `variant="foo"` attribute. With this PR and the above settings, it will not.

```html
<my-element></my-element>
```

In addition, the default value will be set when the attribute is removed.


justinfagnani:
Nice!

My main comment before looking at code is that I don't think we should have the option `skipReflectInitial`. Attributes should ideally _never_ reflect. If you give a default value, I see no benefit to reflecting it, even for styling.

If for some reason in the future enabling initial value reflection is important, we could add an `reflectInitial` option, but I wouldn't include it now.

I also think the option name should be `defaultValue`. `value` sounds too much like the value can't change. "default" implies better to me that that value will be used when the attribute is removed.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1275,22 +1298,44 @@ export abstract class ReactiveElement
   _$changeProperty(
     name: PropertyKey,
     oldValue: unknown,
-    options: PropertyDeclaration
+    options: PropertyDeclaration,
+    initialize?: boolean
```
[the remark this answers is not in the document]
sorvell: Not sure I follow, but it's tricky. We need this to avoid reflecting the initial value which we only want to do when it's called as part of initialization (decorator initializer or performUpdate when `wrapped` is true). That's why this state is passed in.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1257,7 +1269,18 @@ export abstract class ReactiveElement
       ).getPropertyOptions(name);
       const hasChanged = options.hasChanged ?? notEqual;
       const newValue = this[name as keyof this];
-      if (hasChanged(newValue, oldValue)) {
+      let changed = hasChanged(newValue, oldValue);
+      // Force an update if a reflecting property should have an attribute.
+      if (!changed && this.hasUpdated && options.reflect && newValue != null) {
```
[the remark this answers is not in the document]
sorvell: Tweaked checks and added comments to try to make this more clear.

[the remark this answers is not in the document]
sorvell: Resolving so we can revisit, if needed, on current code.

packages/reactive-element/src/decorators/property.ts
```diff
@@ -146,10 +146,8 @@ export const standardProperty = <C extends Interface<ReactiveElement>, V>(
         this.requestUpdate(name, oldValue, options);
       },
       init(this: ReactiveElement, v: V): V {
-        if (v !== undefined) {
-          this._$changeProperty(name, undefined, options);
-        }
-        return v;
+        this._$changeProperty(name, undefined, options, true);
+        return v ?? (options.defaultValue as V);
```
justinfagnani, line 150: What do you think about a dev mode warning that they have both a default value and an initializer?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -274,6 +274,14 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
    * @internal
    */
   wrapped?: boolean;
+
+  /**
+   * Default value for the property. When set, the property is
+   * initialized to this value and if `reflect` is set, the initial default
+   * value does *not* reflect. Avoid setting a value here and
+   * in the field/accessor.
```
justinfagnani, line 282:
```suggestion
   * The default value for the property.
   *
   * When set, the property is initialized to this value and if the `reflect`
   * option is `true`, the initial default value does *not* reflect. Subsequent
   * changes to the property will reflect, even if they are equal to the default
   * value.
   *
   * Avoid setting both a default value here and defining a field initializer or
   * setting the property in the constructor or connectedCallback.
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1257,7 +1265,25 @@ export abstract class ReactiveElement
...
       const hasChanged = options.hasChanged ?? notEqual;
       const newValue = this[name as keyof this];
-      if (hasChanged(newValue, oldValue)) {
+      let changed = hasChanged(newValue, oldValue);
+      // When there is no change, check a corner case that can occur when
+      // 1. there's a defaultValue which was not reflected
+      // 2. the property is subsequently set to the defaultValue.
+      // For example, `prop: {defaultValue: 'foo', reflect: true}`
+      // and el.prop = 'foo'. This should be considered a change if the
+      // attribute is not set.
```
justinfagnani, line 1274: Nice comment, thanks!

```suggestion
      // attribute is not set because we will now reflect the property to the attribute.
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1275,22 +1301,39 @@ export abstract class ReactiveElement
...
     // property to `__reflectingProperties`. This ensures setting
     // attribute + property reflects correctly.
-    if (options.reflect === true && this.__reflectingProperty !== name) {
+    if (reflect) {
       (this.__reflectingProperties ??= new Set<PropertyKey>()).add(name);
     }
   }
 
+  // Intiialize property's defaultValue. Do this without reflection.
+  __initializeDefaultValue(name: PropertyKey, options: PropertyDeclaration) {
```
justinfagnani, line 1328: Mark as private?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1421,12 +1464,21 @@ export abstract class ReactiveElement
         .elementProperties;
       if (elementProperties.size > 0) {
         for (const [p, options] of elementProperties) {
+          const {wrapped, defaultValue} = options;
+          if (wrapped !== true && defaultValue === undefined) {
+            continue;
+          }
+          const value = this[p as keyof this];
+          // Ensure any initially set value is seen as a change.
```
justinfagnani, line 1472: Do we want to do this for default values? It kind of makes sense to me to consider the initial default value _not_ changed.

Or maybe that's what this does? Might be clearer if the comments are inside the conditions? If so, maybe mention that in the jsdocs for the defaultValue option too.

justinfagnani (changes requested):

packages/reactive-element/src/decorators/property.ts
sorvell: I think it's not really possible to detect for all cases, and having a warning that sometimes shows just seems confusing and hard to maintain.

For example, decorator v. not, accessor v. field, instance shadowing, etc.

packages/reactive-element/src/reactive-element.ts
sorvell: Made the comments clearer. Properties with defaultValue set _are_ currently included in initial changes.

We could choose to change this but the number of different ways properties can be configured makes this really hard to manage. (In fact there are some issues with this currently: `accessor` defined properties are only handled specially with the decorator and not in JS.)

justinfagnani (approved):

packages/reactive-element/src/decorators/property.ts
justinfagnani: I hope in the future we can move to standard decorators only and this will be the place we warn. I guess not that many people would use it now. We can try to add a lint warning somewhere.

justinfagnani:
Awesome @sorvell!!

We need to update lit.dev for this feature. Can you do that?

sorvell:
Docs PR: https://github.com/lit/lit.dev/pull/1396

Lookwe69:
defaultValue shouldn't be a function ? It can be eager when the default value is an object. With a function, we can delay the creation only when needed.

sorvell:
[19963dd](https://github.com/lit/lit/pull/4895/commits/19963ddf26b5feb4359665171560c269bc9b219d) adds setting the default value when the attribute is removed.

It also changes the strategy used to ensure the default value is the value of the property at construction time. We need to modify the getter to do this since we can't set the value before the user's constructor runs (at least for `accessor` without standard decorators).

The default value is returned from the getter when it would otherwise return undefined **_and_** the setter for the property has been called (this is recorded via a set). Maintaining this set is a slight tax/complication. 

**_Alternatively_** the initializer could be used as the default value. This would avoid writing the getter and instead just use the normal initializer for it. This would require capturing the default value on the first **_set_** call on the property. The concern here is that the storage would be on the instance rather than property options and therefore consume excess memory. Note, while it could technically be on property options, doing so could have unexpected behavior since data would be implicitly shared between instances.

Note also, non-primitive values as defaults doesn't really make sense with the default in property options since the default value is shared between instances.

Issue #4877, opened 2024-12-20
[lit] Remove "Lit is in dev mode" warning


maxpatiiuk:
### Should this be an RFC?

- [x] This is not a substantial change

### Which package is this a feature request for?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

Lit emits a `Lit is in dev mode` warning when Lit is loaded using the development bundle, rather than production:

![Image](https://github.com/user-attachments/assets/2926362e-b9ec-4c53-ad2e-03382f5a4c9c)

I understand that this warning is useful for detecting cases when development mode of Lit is accidentally used in production, but I do not know how likely that is, or if it is worth the tradeoff.

From my understanding, Lit's dev mode is only enabled when the "development" export condition is set - thus, it is a quite deliberate opt-in step, so I am not sure how likely it is that development build is accidentally shipped to production.

Some negative consequences of this warning:
- Clutteres the console, adding noise and distracting from real warnings/errors
- Affects the console not only of the team working on a Lit package, but also of any consumer of that Lit-authored package - that is a lot of console noise. Some users might even be confused into thinking something is misconfigured/not working right because they see this warning
- Cluttering the console is especially bad when running isolated tests, as each test spams the console with this warning
- This warning can break tests. Many testing setups use [jest-fail-on-console](https://www.npmjs.com/package/jest-fail-on-console) or [vitest-fail-on-console](https://www.npmjs.com/package/vitest-fail-on-console) to fail the test when any error or warning is emitted - Lit's warning falsely fails tests and requires special configuration to avoid
- Some organizations have a 0 warning policy

### Alternatives and Workarounds

I understand that this warning can be disabled by adding the following code:
```js
globalThis.litIssuedWarnings ??= new Set();
globalThis.litIssuedWarnings.add(
  "Lit is in dev mode. Not recommended for production! See https://lit.dev/msg/dev-mode for more information.",
);
```

However, that is far from an ideal solution because the above code would need to be added by every consumer of my package.
Worse, since Lit's warning is emitted in global scope, the consumers have to make sure the above code is added before Lit is imported - which can be challenging in some apps (and the fact that imports are hoisted makes that even trickier)

justinfagnani:
This warning is important to us to help discourage misconfigured builds and deployments from shipping the debug code in Lit. Given that we include extra debug code, including validity checks that can effect performance, we really want to keep the warning.

[...]

PR #4901, opened 2025-01-21
feat(lit): make disabling dev mode warnings simpler

maxpatiiuk:
Makes it much easier to work around https://github.com/lit/lit/issues/4877.
By adding queueMicrotask, the warnings are no longer emitted in the global scope, which makes disabling them much easier (see referenced issue for details).

With this change, I am able to add code for disabling this warning to my runtime code - with that, the warning is be disabled for all consumers of my library without them having to do anything special in their test setup files.

The warning is still enabled by default in vanilla Lit for those who prefer it.

The original discussion was only about the "Lit is in dev mode" warning, but I made this change for other global scope warnings too for consistency.

## Testing

Verified that after this change I am able to silence Lit's dev mode warning regardless of the order of imports.

### DIFF
--- a/packages/lit-element/src/lit-element.ts
+++ b/packages/lit-element/src/lit-element.ts
@@ -279,9 +279,11 @@ export const _$LE = {
 // This line will be used in regexes to search for LitElement usage.
 (globalThis.litElementVersions ??= []).push('4.1.1');
 if (DEV_MODE && globalThis.litElementVersions.length > 1) {
-  issueWarning!(
-    'multiple-versions',
-    `Multiple versions of Lit loaded. Loading multiple versions ` +
-      `is not recommended.`
-  );
+  queueMicrotask(() => {
+    issueWarning!(
+      'multiple-versions',
+      `Multiple versions of Lit loaded. Loading multiple versions ` +
+        `is not recommended.`
+    );
+  });
 }
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -228,10 +228,12 @@ if (DEV_MODE) {
     }
   };
 
-  issueWarning(
-    'dev-mode',
-    `Lit is in dev mode. Not recommended for production!`
-  );
+  queueMicrotask(() => {
+    issueWarning(
+      'dev-mode',
+      `Lit is in dev mode. Not recommended for production!`
+    );
+  });
 }
 
 const wrap =
@@ -2197,11 +2199,13 @@ polyfillSupport?.(Template, ChildPart);
 // This line will be used in regexes to search for lit-html usage.
 (global.litHtmlVersions ??= []).push('3.2.1');
 if (DEV_MODE && global.litHtmlVersions.length > 1) {
-  issueWarning!(
-    'multiple-versions',
-    `Multiple versions of Lit loaded. ` +
-      `Loading multiple versions is not recommended.`
-  );
+  queueMicrotask(() => {
+    issueWarning!(
+      'multiple-versions',
+      `Multiple versions of Lit loaded. ` +
+        `Loading multiple versions is not recommended.`
+    );
+  });
 }
 
 /**
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -94,19 +94,21 @@ if (DEV_MODE) {
     }
   };
 
-  issueWarning(
-    'dev-mode',
-    `Lit is in dev mode. Not recommended for production!`
-  );
-
-  // Issue polyfill support warning.
-  if (global.ShadyDOM?.inUse && polyfillSupport === undefined) {
+  queueMicrotask(() => {
     issueWarning(
-      'polyfill-support-missing',
-      `Shadow DOM is being polyfilled via \`ShadyDOM\` but ` +
-        `the \`polyfill-support\` module has not been loaded.`
+      'dev-mode',
+      `Lit is in dev mode. Not recommended for production!`
     );
-  }
+
+    // Issue polyfill support warning.
+    if (global.ShadyDOM?.inUse && polyfillSupport === undefined) {
+      issueWarning(
+        'polyfill-support-missing',
+        `Shadow DOM is being polyfilled via \`ShadyDOM\` but ` +
+          `the \`polyfill-support\` module has not been loaded.`
+      );
+    }
+  });
 }
 
 /**
@@ -1671,9 +1673,11 @@ if (DEV_MODE) {
 // This line will be used in regexes to search for ReactiveElement usage.
 (global.reactiveElementVersions ??= []).push('2.0.4');
 if (DEV_MODE && global.reactiveElementVersions.length > 1) {
-  issueWarning!(
-    'multiple-versions',
-    `Multiple versions of Lit loaded. Loading multiple versions ` +
-      `is not recommended.`
-  );
+  queueMicrotask(() => {
+    issueWarning!(
+      'multiple-versions',
+      `Multiple versions of Lit loaded. Loading multiple versions ` +
+        `is not recommended.`
+    );
+  });
 }


maxpatiiuk:
cc @sorvell for review

sorvell (changes requested):
We need to add a test for this. To facilitate this, let's consider adding to the issueWarning tracker the code for the warning in addition to the warning itself.

Then I hope you could do something like:

```js
window.litIssuedWarnings ?? = new Set();
window.litIssuedWarnings.add(`dev-mode`);
```

maxpatiiuk:
Done - though it won't work fully till https://github.com/lit/lit.dev/issues/1350 is in place (which requires Lit team to update the dev website)
Until then, I will add fallback code in issuedWarnings to prevent empty warning code from disabling warnings for disperate empty code warning messages.

---

Also, I noticed that reactive-element already disables warnings based on a code (which does not match the warning code):
https://github.com/lit/lit/blob/6a232e92af9372892c7a916dd3d25947be674ee0/packages/reactive-element/src/reactive-element.ts#L1168-L1174

packages/lit-element/src/lit-element.ts
```diff
@@ -97,15 +97,17 @@ let issueWarning: (code: string, warning: string) => void;
 if (DEV_MODE) {
   // Ensure warnings are issued only 1x, even if multiple versions of Lit
   // are loaded.
-  const issuedWarnings: Set<string | undefined> =
-    (globalThis.litIssuedWarnings ??= new Set());
+  globalThis.litIssuedWarnings ??= new Set();
```
maxpatiiuk, line 100: some of your tests are re-assigning `globalThis.litIssuedWarnings`.
to better support that use case, rather than having local `issuedWarnings` variable, always check what is the current globalThis.litIssuedWarnings value

packages/lit-element/src/lit-element.ts
```diff
@@ -97,15 +97,17 @@ let issueWarning: (code: string, warning: string) => void;
...
   issueWarning = (code: string, warning: string) => {
     warning += ` See https://lit.dev/msg/${code} for more information.`;
-    if (!issuedWarnings.has(warning)) {
+    if (
+      !globalThis.litIssuedWarnings!.has(warning) &&
+      !globalThis.litIssuedWarnings!.has(code)
+    ) {
       console.warn(warning);
-      issuedWarnings.add(warning);
+      globalThis.litIssuedWarnings!.add(warning);
```
maxpatiiuk, line 110: While user now has the ability to disable warnings based on warning code, the issueWarning itself still only disables the warnings it emitted using warning content rather than code.
This is because, in case of reactive-element, the same warning message may be emitted once for each tag name (rather than once for entire app) - that seems like a useful behavior.

https://github.com/lit/lit/blob/6a232e92af9372892c7a916dd3d25947be674ee0/packages/reactive-element/src/reactive-element.ts#L1175-L1178

packages/lit-element/src/test/lit-element_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,45 @@
...
+import {LitElement} from 'lit-element';
+import {assert} from 'chai';
+
+// Note, since tests are not built with production support, detect DEV_MODE
+// by checking if warning API is available.
+const DEV_MODE = !!LitElement.enableWarning;
+
+if (DEV_MODE) {
+  globalThis.litIssuedWarnings ??= new Set();
+  globalThis.litIssuedWarnings!.add('dev-mode');
```
maxpatiiuk, line 16: dev-mode warning is now delayed by a microtask so we can disable the warning here (even after lit-element is already imported)

However, we can't do so inside of test/suiteSetup as they run later than a microtask.
Not an issue for my use case as disabling warnings in the global scope is now easy (so my library can do so for all consumers).

packages/reactive-element/src/test/reactive-element_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,78 @@
...
+      test('disabled warning of dev mode', () => {
+        assert.isFalse(
+          Array.from(litWarnings).some((v) => v?.includes('Lit is in dev mode'))
+        );
+        assert.isTrue(
+          Array.from(litWarnings).some((v) => v?.includes('dev-mode'))
+        );
+      });
+
+      (missingPlatformSupport ? test : test.skip)(
```
maxpatiiuk, line 61: is this still needed?

packages/lit-element/src/test/lit-element_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,45 @@
...
+      console.warn = (message: string) => warnings.push(message);
+    });
+
+    suiteTeardown(() => {
+      console.warn = consoleWarn;
+    });
+
+    const litWarnings = globalThis.litIssuedWarnings!;
+
+    test('dev mode waring was disabled', () => {
```
sorvell, line 33:
```suggestion
    test('dev mode warning was disabled', () => {
```

packages/lit-element/src/lit-element.ts
```diff
@@ -97,15 +97,17 @@ let issueWarning: (code: string, warning: string) => void;
 if (DEV_MODE) {
   // Ensure warnings are issued only 1x, even if multiple versions of Lit
   // are loaded.
-  const issuedWarnings: Set<string | undefined> =
-    (globalThis.litIssuedWarnings ??= new Set());
+  globalThis.litIssuedWarnings ??= new Set();
 
   // Issue a warning, if we haven't already.
```
sorvell, line 102:
```suggestion
  /** 
   * Issue a warning if we haven't already, based either on `code` or `warning`.
   * Warnings are disabled automatically only by `warning`; disabling via `code` 
   * can be done by users.
   */
```

packages/lit-html/src/lit-html.ts
```diff
@@ -215,23 +215,28 @@ let debugLogRenderId = 0;
 let issueWarning: (code: string, warning: string) => void;
 
 if (DEV_MODE) {
-  global.litIssuedWarnings ??= new Set();
+  globalThis.litIssuedWarnings ??= new Set();
 
   // Issue a warning, if we haven't already.
```
sorvell, line 220:
```suggestion
  /** 
   * Issue a warning if we haven't already, based either on `code` or `warning`.
   * Warnings are disabled automatically only by `warning`; disabling via `code` 
   * can be done by users.
   */
```

packages/lit-html/src/test/lit-html_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,47 @@
...
+    suiteTeardown(() => {
+      console.warn = consoleWarn;
+    });
+
+    const litWarnings = globalThis.litIssuedWarnings!;
+
+    test('dev mode waring was disabled', () => {
+      // Ensure lit-html package is imported
+      void html``;
+      assert.lengthOf(
```
sorvell, line 36: Add comment that this will check if the warning was issued.

packages/lit-html/src/test/lit-html_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,47 @@
...
+// Note, since tests are not built with production support, detect DEV_MODE
+// by checking if warning API is available.
+const DEV_MODE = !!globalThis.litIssuedWarnings;
+
+if (DEV_MODE) {
+  globalThis.litIssuedWarnings ??= new Set();
+  globalThis.litIssuedWarnings!.add('dev-mode');
+
+  suite('Can disable developer mode warning', () => {
+    const warnings: string[] = [];
```
sorvell, line 19: Is this used? Remove if not.

packages/lit-html/src/test/lit-html_dev_mode_test.ts
```diff
@@ -0,0 +1,52 @@
...
+        container.parentNode.removeChild(container);
+      }
+    });
+
+    const litWarnings = globalThis.litIssuedWarnings!;
+
+    test('warns for dev mode only 1x', () => {
+      // Ensure lit-html package is imported
+      void html``;
+      assert.equal(
```
sorvell, line 46: Add comment that this will check that warning was not issued.

packages/reactive-element/src/decorators/query.ts
```diff
@@ -20,17 +20,19 @@ let issueWarning: (code: string, warning: string) => void;
 if (DEV_MODE) {
   // Ensure warnings are issued only 1x, even if multiple versions of Lit
   // are loaded.
-  const issuedWarnings: Set<string | undefined> =
-    (globalThis.litIssuedWarnings ??= new Set());
+  globalThis.litIssuedWarnings ??= new Set();
 
   // Issue a warning, if we haven't already.
```
sorvell, line 25:
```suggestion
  /** 
   * Issue a warning if we haven't already, based either on `code` or `warning`.
   * Warnings are disabled automatically only by `warning`; disabling via `code` 
   * can be done by users.
   */
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -82,31 +82,35 @@ const polyfillSupport = DEV_MODE
 if (DEV_MODE) {
   // Ensure warnings are issued only 1x, even if multiple versions of Lit
   // are loaded.
-  const issuedWarnings: Set<string | undefined> = (global.litIssuedWarnings ??=
-    new Set());
+  globalThis.litIssuedWarnings ??= new Set();
 
   // Issue a warning, if we haven't already.
```
sorvell, line 87:
```suggestion
  /** 
   * Issue a warning if we haven't already, based either on `code` or `warning`.
   * Warnings are disabled automatically only by `warning`; disabling via `code` 
   * can be done by users.
   */
```

packages/reactive-element/src/test/reactive-element_dev_mode_disable_test.ts
```diff
@@ -0,0 +1,78 @@
...
+const DEV_MODE = !!ReactiveElement.enableWarning;
+
+if (DEV_MODE) {
+  globalThis.litIssuedWarnings ??= new Set();
+  globalThis.litIssuedWarnings!.add('dev-mode');
+  globalThis.litIssuedWarnings!.add('polyfill-support-missing');
+
+  suite('Can disable developer mode warning', () => {
+    let container: HTMLElement;
+    let warnings: string[] = [];
```
sorvell, line 21: Again, remove if unused.

sorvell (changes requested):

packages/lit-html/src/lit-html.ts
```diff
@@ -215,23 +215,28 @@ let debugLogRenderId = 0;
 let issueWarning: (code: string, warning: string) => void;
 
 if (DEV_MODE) {
-  global.litIssuedWarnings ??= new Set();
+  globalThis.litIssuedWarnings ??= new Set();
```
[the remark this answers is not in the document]
maxpatiiuk: I didn't realize `global` referred to a local variable rather than a Node.js global. Will revert

maxpatiiuk:
Thanks for the review! Applied all suggested fixes

sorvell (approved):

maxpatiiuk:
Merge conflict resolved

Issue #4675, opened 2024-06-16
[context] `@provide()` in standard decorator mode does not provide initial value


mrcsmcln:
### Which package(s) are affected?

Context (@lit/context)

### Description

I am creating an [HTML-based plugin](https://lit.dev/docs/data/context/#html-based-plugins) using Lit. I have a recursive custom element that works like this:

```html
<custom-element>
  <custom-element></custom-element>
</custom-element>
```

To pass data from one `<custom-element>` to the other down the tree, we need to define both a provider and a consumer on the same element. Using standard decorators, it looks like this:

```ts
import {html, css, LitElement} from 'lit';
import {createContext, provide, consume} from '@lit/context';
import {customElement, property} from 'lit/decorators.js';

const foo = createContext<string>(Symbol('foo'))
    
@customElement('custom-element')
export class CustomElement extends LitElement {
  @provide({ context: foo })
  accessor provider = 'provided consumer value'
      
  @consume({ context: foo })
  accessor consumer = 'default consumer value'

  render() {
    return html`
      ${this.consumer || 'undefined consumer value'}
      <slot></slot>
    `;
  }
}
```

I'd expect to see `provided consumer value` in the inner `<custom-element>`, but it displays `undefined consumer value` instead. With TypeScript decorators, there doesn't seem to be an issue but when using standard decorators, TypeScript throws an error:

```
Unable to resolve signature of property decorator when called as an expression.
The runtime will invoke the decorator with 3 arguments, but the decorator expects 1-2.
```

My best guess is that this is due to using `accessor`, but I don't know enough about accessors to make a firm determination.

### Reproduction

[...]

PR #4917, opened 2025-02-04
[context] Make @provide work with standard decorators

djrenren:
Fixes https://github.com/lit/lit/issues/4675

Previously this code added the context provider to the map in `addInitializer`, and the `init` callback expected it to be there. However the spec for accessor attributes states that the `addInitializer` callback is invoked *after* `init`, meaning the intial value was never assigned to the provider.

This is a pretty simple fix, but as a new contributor I couldn't quite figure out how to get the test running properly. Any tips there would be appreciated.


djrenren:
I have signed the CLA now, but I can't rerun the action to check it.

justinfagnani (commented):
This looks great, thanks!!

We'll need you to sign the CLA at: https://cla.developers.google.com/clas

And the PR needs a changeset. See https://github.com/lit/lit/blob/main/CONTRIBUTING.md#pull-requests

You can run `npm run changeset` and you will be prompted to enter a description for the change.

Other than that, LGTM!

packages/context/src/lib/decorators/provide.ts
```diff
@@ -65,7 +61,10 @@ export function provide<ValueType>({
           return protoOrTarget.set.call(this, value);
         },
         init(this: ReactiveElement, value: ValueType) {
-          controllerMap.get(this)?.setValue(value);
+          controllerMap.set(
```
sorvell, line 64: This doesn't address the legacy decorator version. We need to cover that as well, somehow.

djrenren: My understanding is that the legacy one already works fine (at least it's tested in `packages/context/src/test/context-provider_test.ts`

packages/reactive-element/src/test/decorators-modern/provide_test.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {Context, provide} from '@lit/context';
```
justinfagnani, line 7: This is the wrong place for this test. `@lit/reactive-element` doesn't depend on `@lit/context`. The test should be in `packages/context/src/text/std-decorators/context-provider_test.ts`.

The reason why this bug isn't covered there is because the test DOM always sets an attribute for the value being provided, so it always goes through the setter code path, not the initializer code path.

justinfagnani (changes requested):

justinfagnani:
@djrenren thanks for allowing commits from maintainers!

I pushed up new tests that I verified caught the bug before the fix. I also found and fixed a wireit config problem that could have prevented the standard decorator tests from running.

justinfagnani (approved):

PR #4934, opened 2025-02-19
Adds `useDefault` property option

sorvell:
**Note:** This is an alternative to https://github.com/lit/lit/pull/4895, only one should be merged.

Adds support for https://github.com/lit/rfcs/issues/36 and helps facilitate https://github.com/shoelace-style/shoelace/discussions/2339.

Spawning an initial attribute can cause a NextJS hydration error, see [repro](https://codesandbox.io/p/devbox/nostalgic-monad-3ps3dz?workspaceId=ws_MXuotjJLEi62fRTssWSgdA).

To use, set `useDefault` in property options, and make sure to set an initial value  in the field/accessor.

```ts
@property({reflect: true, useDefault: true})
accessor variant = 'foo';
```

Note, using `accessor` is encouraged but not required. This setting can also be used in `static properties` and on custom getters/setters.

Previously, this would spawn a `variant="foo"` attribute. With this PR and the above settings, it will not.

```html
<my-element></my-element>
```

In addition, the initial value will be set when the attribute is removed.


packages/reactive-element/src/reactive-element.ts
```diff
@@ -274,6 +274,15 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
    * @internal
    */
   wrapped?: boolean;
+
+  /**
+   * When set and if the `reflect` option is `true`, the initial default value
+   * does *not* reflect. Subsequent changes to the property will reflect, even
+   * if they are equal to the default value. The default value will not trigger
+   * an initial old value in the `changedProperties` map argument
+   * to update lifecycle methods.
```
justinfagnani, line 283:
```suggestion
   * When `true`, uses the initial value of the property as the default value,
   * which changes how attributes are handled:
   *  - The initial value does *not* reflect, even if the `reflect` option is `true`.
   *    Subsequent changes to the property will reflect, even if they are equal to the
   *     default value.
   *  - When the attribute is removed, the property is set to the default value
   *  - The initial value will not trigger an old value in the `changedProperties` map
   *    argument to update lifecycle methods.
   *
   * When set, properties must be initialized, either with a field initializer, or an
   * assignment in the constructor. Not initializing the property may lead to
   * improper handling of subsequent property assignments.
   *
   * While this behavior is opt-in, most properties that reflect to attributes should
   * use `useDefault: true` so that their initial values do not reflect.
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -676,6 +686,12 @@ export abstract class ReactiveElement
       (options as Mutable<PropertyDeclaration, 'attribute'>).attribute = false;
     }
     this.__prepare();
+    // Whether this property is wrapping accessors.
+    // Helps control the initial value change and reflection logic.
+    if (this.prototype.hasOwnProperty(name)) {
+      options = Object.create(options);
+      options.wrapped = true;
```
justinfagnani, line 693: Do we care about preserving the prototype relationship here? I would normally do:
```suggestion
      options = {...options, wrapped: true};
```

packages/reactive-element/src/reactive-element.ts
```diff
@@ -748,12 +764,10 @@ export abstract class ReactiveElement
...
     }
     return {
-      get(this: ReactiveElement) {
-        return get?.call(this);
-      },
+      get,
       set(this: ReactiveElement, value: unknown) {
         const oldValue = get?.call(this);
-        set!.call(this, value);
+        set?.call(this, value);
```
justinfagnani, line 770: Is this because we support decorating a getter without a setter now?

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1251,7 +1271,26 @@ export abstract class ReactiveElement
...
+      // When there is no change, check a corner case that can occur when
+      // 1. there's a initial value which was not reflected
+      // 2. the property is subsequently set to this value.
+      // For example, `prop: {useDefault: true, reflect: true}`
+      // and el.prop = 'foo'. This should be considered a change if the
+      // attribute is not set because we will now reflect the property to the attribute.
+      if (
+        !changed &&
+        options.reflect &&
+        newValue != null &&
```
justinfagnani, line 1284: Don't we need to run through the attribute converter to know if the property will be reflected? ie, if the converter returns `null`?

packages/reactive-element/src/reactive-element.ts
sorvell: There was a test that covered this case that we weren't previously exercising through here.

packages/reactive-element/src/reactive-element.ts
sorvell: Not sure we should bother since it's even more code? It's "correct" either way since the reflection will only happen if it should. In that rare case that it wouldn't reflect, we'd skip an update.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1251,7 +1271,26 @@ export abstract class ReactiveElement
...
+        options.useDefault &&
+        newValue === this.__defaultValues?.get(name)
+      ) {
+        const attr = (
+          this.constructor as typeof ReactiveElement
+        ).__attributeNameForProperty(name, options);
+        changed = !this.hasAttribute(attr!);
+      }
```
justinfagnani, line 1302:
```suggestion
      changed ||= 
        options.useDefault &&
        options.reflect &&
        newValue === this.__defaultValues?.get(name) &&
        !this.hasAttribute(ctor.__attributeNameForProperty(name, options)!);
```

You'll need `ctor`

justinfagnani (approved):
🎉 awesome!!

freshp86:
We are encountering an issue in the Chromium codeabase as part of migrating to using the `accessor` keyword (to remove our dependency to `useDefineForClassFields:false`) along with `reflect: true`, which sounds that this new `useDefault: true` option would solve?

Please see https://issues.chromium.org/issues/389737066#comment12 for more context or directly see this minimal Lit playground [repro](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy5qcyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcblxuZXhwb3J0IGNsYXNzIFNpbXBsZUdyZWV0aW5nIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHN0YXRpYyBzdHlsZXMgPSBjc3NgXG4gICAgOmhvc3QgZGl2IHtcbiAgICAgIGRpc3BsYXk6IG5vbmU7XG4gICAgfVxuICAgICAgXG4gICAgOmhvc3QoW3Nob3VsZFNob3ddKSBkaXYge1xuICAgICAgZGlzcGxheTogaW5pdGlhbDtcbiAgICB9XG4gIGA7XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYFxuICAgICAgICA8cD5cbiAgICAgICAgICBUaGUgZGl2IGJlbG93IHNob3VsZCBiZSB2aXNpYmxlIG9uIHN0YXJ0dXAsIGJ1dCBpdCBpc24ndC5cbiAgICAgICAgICBJdCBvbmx5IGJlY29tZXMgdmlzaWJsZSBhZnRlciB0b2dnbGluZyBzaG91bGRTaG93IHRvIGZhbHNlIGFuZCB0aGVuXG4gICAgICAgICAgYmFjayB0byB0cnVlICgyIGNsaWNrcyBiZWxvdykuIFNlZW1zIGxpa2UgYSBMaXQgYnVnP1xuICAgICAgICA8cD5cbiAgICAgICAgPGgyPnNob3VsZFNob3c6ICR7dGhpcy5zaG91bGRTaG93fTwvaDI-XG4gICAgICAgIDxidXR0b24gQGNsaWNrPVwiJHt0aGlzLm9uQ2xpY2tffVwiPkNsaWNrIG1lPC9idXR0b24-XG4gICAgICAgIDxkaXY-SSBhbSBvbmx5IHZpc2libGUgd2hlbiA6aG9zdChbc2hvdWxkU2hvd10pIG1hdGNoZXM8L2Rpdj5cbiAgICBgO1xuICB9XG4gIFxuICBzdGF0aWMgZ2V0IHByb3BlcnRpZXMoKSB7XG4gICAgcmV0dXJuIHtcbiAgICAgIHNob3VsZFNob3c6IHtcbiAgICAgICAgdHlwZTogQm9vbGVhbixcbiAgICAgICAgcmVmbGVjdDogdHJ1ZSxcbiAgICAgICAgdXNlRGVmYXVsdDogdHJ1ZSxcbiAgICAgIH0sXG4gICAgfTtcbiAgfVxuXG4gIC8vIFRoZSBjb2RlIGJlbG93IGlzIHdoYXQgVHlwZVNjcmlwdCBjb21waWxlciBlbWl0cyB3aGVuIHRoZSBmb2xsb3dpbmcgbGluZSBpcyB1c2VkLlxuICAvLyBhY2Nlc3NvciBzaG91bGRTaG93OiBib29sZWFuID0gdHJ1ZTtcblxuICAjc2hvdWxkU2hvd19hY2Nlc3Nvcl9zdG9yYWdlID0gdHJ1ZTtcbiAgZ2V0IHNob3VsZFNob3coKSB7IHJldHVybiB0aGlzLiNzaG91bGRTaG93X2FjY2Vzc29yX3N0b3JhZ2U7IH1cbiAgc2V0IHNob3VsZFNob3codmFsdWUpIHsgdGhpcy4jc2hvdWxkU2hvd19hY2Nlc3Nvcl9zdG9yYWdlID0gdmFsdWU7IH1cblxuICBvbkNsaWNrXygpIHtcbiAgICB0aGlzLnNob3VsZFNob3cgPSAhdGhpcy5zaG91bGRTaG93O1xuICB9XG59XG5cbmN1c3RvbUVsZW1lbnRzLmRlZmluZSgnc2ltcGxlLWdyZWV0aW5nJywgU2ltcGxlR3JlZXRpbmcpOyJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9zaW1wbGUtZ3JlZXRpbmcuanNcIj48L3NjcmlwdD5cbjwvaGVhZD5cbjxib2R5PlxuICA8c2ltcGxlLWdyZWV0aW5nPjwvc2ltcGxlLWdyZWV0aW5nPlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ)

> Note: This is an alternative to https://github.com/lit/lit/pull/4895, only one should be merged.

Completely ignoring the initial value would probably not work for us, as we are already getting failed tests when migrating to the `accessor` keyword due to this behavior.

justinfagnani:
@freshp86 can you open an issue?

It seems like the issues is that with experimental decorators and auto-accessors, the initial value isn't reflected? I'll have to check our tests for that case.

@sorvell on the linked playground, adding `wrapped: true` fixes the issue. Maybe this is a case of decorating the getter vs the setter.

sorvell:
@freshp86 This PR should also fix your issue, regardless of using `useDefault`. We were only properly handling `accessor` when using the `@property` decorator, not `static properties`. This was corrected [here](https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L701-L704) via this PR.

As @justinfagnani mentioned, while waiting for this release, you can add `wrapped: true` to the property options to correct the behavior. This workaround will be unnecessary in the next Lit release, but it is compatible with it ([repro with workaround](https://lit.dev/playground/#gist=3edcbf167b3203650a2bccbe9903a711))

freshp86:
> As @justinfagnani mentioned, while waiting for this release, you can add wrapped: true to the property options to correct the behavior. 

Thanks for the quick response! I verified that adding `wrapped: true` fixes both the Lit playground repro, as well as one of the tests that were failing in Chromium (will verify remaining tests shortly). I don't see the  `wrapped` option documented at https://lit.dev/docs/components/properties/#property-options though, so curious if this is considered part of the public API.

> This workaround will be unnecessary in the next Lit release, but it is compatible with it ([repro with workaround](https://lit.dev/playground/#gist=3edcbf167b3203650a2bccbe9903a711))

Do you have any estimate on when the next release will come out? In Chromium we still use v3.0.2 [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/node/package.json;l=33;drc=5211137c915706f5f4f55ad343f1aa1a798dab12), which is already behind latest 3.2.1, so we'll need to roll forward a few versions to get this fix.

In the meantime, do you think locally patching [these lines](https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L701-L704) that you referernced in our Chromium copy of Lit would be sufficient to fix the issue, without us having to update to later releases?

sorvell:
> In the meantime

I think it's probably simpler to use the undocumented `wrapped: true` rather than patching, and just update when we publish the next version. We can probably do a release this week.

freshp86:
Might have found an even easier fix for us, which does not require updating client code with `wrapped: true`, by automatically adding it during our `static override finalize() {...}` method for the `CrLitElement` class (which all other Lit element's extend from). See https://chromium-review.googlesource.com/c/chromium/src/+/6349561. I have verified this fixes previously failing tests, but still waiting to confirm that no other tests break because of it.

sorvell:
> Might have found an even easier fix for us, which does not require updating client code with wrapped: true

If you have a base class, probably the simplest way would be:

```js
static createProperty(name, options) {
  super.createProperty(name, {...options, wrapped: true});
}
```

freshp86:
> If you have a base class, probably the simplest way would be:

Don't we need to check for  `if (this.prototype.hasOwnProperty(name)) {...}` similar to your PR to only add the `wrapped: true` for such cases? In the superclass we still need to deal with both cases (using/not-using `accessor`). Also, don't we only need to add this for cases where `reflect: true` and not add it to all reactive porperties?

sorvell:
> Don't we need to check for if (this.prototype.hasOwnProperty(name)) {...}

It's only strictly needed there, but you don't need to discriminate it. Setting `wrapped` forces the property to be in the changed properties if it has a value on first update. Valued non-own properties are already in changed properties. 

> Also, don't we only need to add this for cases where reflect: true instead of adding it to all reactive porperties?

No, it should not be based on the value of `reflect`. As noted, setting `wrapped` ensures the property is in the changed properties. This makes reflection work and *also* ensures the property is included in changed properties (e.g. `updated(changed /* in this map */)` )

freshp86:
Thanks for the info! I will update my CL and tests at https://chromium-review.googlesource.com/c/chromium/src/+/6349561 and I might add you as a reviewer as well if you don't mind. Thanks!

sorvell:
> In short accessor properties while indeed present in changed properties when wrapped: true is used, their "previous" value reported is not undefined

Yeah, I should have mentioned that. This is also fixed in this change, and you can patch it in if you need, but it's also technically safe to do in your base class like this (since `_$*` properties are not renamed):

```ts
override protected performUpdate() {
  if (!this.isUpdatePending) {
    return;
  }
  // this loop is duplicated in the super call but the work won't occur twice due to the 
  // `_$changedProperties` check
  if (!this.hasUpdated) {
    const {elementProperties} = (this.constructor as typeof ReactiveElement);
      if (options.wrapped === true && !this._$changedProperties.has(p) &&
          this[p as keyof this] !== undefined) {
        this._$changeProperty(p, undefined, options);
      }
    }
  }
  super.performUpdate();
}
```

freshp86:
Thanks for the detailed response, this is indeed very helpful and does fix the issue!

Having said that, fixing the issue in our `CrLitElement` base is a bit complicated. Specifically, after verifying that the above works in the minimal repro, plugging it in `CrLitELemnet` didn't initially work, and took quite a bit to debug. The problem is that we operate on the minified version of `ReactiveElement` from `node_modules/@lit/reactive-element/reactive-element.js`, where the internal APIs `_$changedProperties` and `_$changeProperty` are minified to `_$AL` and `C` respectively (took a while to figure out why `_$changedProperties` was `undefined`). We also need to make TypeScript happy while using those internal APIs, so the end result looks as follows:
```ts
  override performUpdate() {
    if (!this.isUpdatePending) {
      return;
    }
    if (!this.hasUpdated) {
      interface WithInternalApis extends CrLitElement {
        // minified name for $changedProperties
        _$AL: PropertyValues;
        // minified name for _$changeProperty()
        C(p: PropertyKey, v: any, options: any): void;
      }

      const {elementProperties} = this.constructor as typeof ReactiveElement;
        if ((options as {wrapped?: boolean}).wrapped === true &&
            !(this as unknown as WithInternalApis)._$AL.has(p) &&
            this[p as keyof this] !== undefined) {
          (this as unknown as WithInternalApis).C(p, undefined, options);
        }
      }
    }
    super.performUpdate();
  }
```
I think this qualifies as a hack :> You can see the complete change at https://chromium-review.googlesource.com/c/chromium/src/+/6349561/7/third_party/lit/v3_0/cr_lit_element.ts.

So, the sooner we can get the official fix the sooner we can remove this hack. Rolling our Lit version to the latest Lit version is also blocked by https://github.com/lit/lit/issues/4943, so the current plan is 

1. Land the hacky fixes directly in `CrLitElement`, to unblock the migration to the `accessor` keyword (https://issues.chromium.org/issues/389737066). 
2. Roll our Lit version to 3.2.1 as soon as there is a fix for https://github.com/lit/lit/issues/4943 (we would like to avoid adding `skipLibCheck` across the entire repo).
3. Roll to 3.2.2 (or whichevere version will have these fixes) and remove the workarounds in `CrLitElement`.

Thanks again!

freshp86:
@sorvell Unfortunately discovered one more difference in changed properties behavior between wrapped and non-wrapped properties, when a value is passed from the parent, see new minimal repro [here](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy5qcyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIExpdEVsZW1lbnR9IGZyb20gJ2xpdCc7XG5cbmV4cG9ydCBjbGFzcyBTaW1wbGVHcmVldGluZyBleHRlbmRzIExpdEVsZW1lbnQge1xuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgXG4gICAgICAgPGRpdj5EaWZmZXJlbnQgY2hhbmdlZCBwcm9wZXJ0aWVzIGJlaGF2aW9yIGZvciB3cmFwcGVkIGFuZCBub24td3JhcHBlZCwgc2VlIERldlRvb2xzIGNvbnNvbGU8L2Rpdj5cbiAgICAgICA8aHI-XG4gICAgICAgPGRpdj4ke3RoaXMucHJvcDF9ICR7dGhpcy5wcm9wMn08L2Rpdj5cbiAgICBgO1xuICB9XG4gIFxuICBzdGF0aWMgZ2V0IHByb3BlcnRpZXMoKSB7XG4gICAgcmV0dXJuIHtcbiAgICAgIHByb3AxOiB7XG4gICAgICAgIHR5cGU6IFN0cmluZyxcbiAgICAgICAgd3JhcHBlZDogdHJ1ZSxcbiAgICAgIH0sXG4gICAgICBcbiAgICAgIHByb3AyOiB7XG4gICAgICAgIHR5cGU6IFN0cmluZyxcbiAgICAgIH0sXG4gICAgfTtcbiAgfVxuXG4gIC8vIFRoZSBjb2RlIGJlbG93IGlzIHdoYXQgVHlwZVNjcmlwdCBjb21waWxlciBlbWl0cyB3aGVuIHRoZSBmb2xsb3dpbmcgbGluZSBpcyB1c2VkLlxuICAvLyBhY2Nlc3NvciBwcm9wMTogYm9vbGVhbiA9IHRydWU7XG4gICNwcm9wMV9hY2Nlc3Nvcl9zdG9yYWdlID0gJ2luaXRpYWwnO1xuICBnZXQgcHJvcDEoKSB7IHJldHVybiB0aGlzLiNwcm9wMV9hY2Nlc3Nvcl9zdG9yYWdlOyB9XG4gIHNldCBwcm9wMSh2YWx1ZSkgeyB0aGlzLiNwcm9wMV9hY2Nlc3Nvcl9zdG9yYWdlID0gdmFsdWU7IH1cblxuICBjb25zdHJ1Y3RvcigpIHtcbiAgICBzdXBlcigpO1xuICAgIHRoaXMucHJvcDIgPSAnaW5pdGlhbCc7XG4gIH1cblxuICB3aWxsVXBkYXRlKGNoYW5nZWRQcm9wZXJ0aWVzKSB7XG4gICAgc3VwZXIud2lsbFVwZGF0ZShjaGFuZ2VkUHJvcGVydGllcyk7XG4gICAgY29uc29sZS5sb2coJy0tLXdpbGxVcGRhdGUnKTtcbiAgICBjb25zb2xlLmxvZyh1bmRlZmluZWQgPT09IGNoYW5nZWRQcm9wZXJ0aWVzLmdldCgncHJvcDEnKSk7IC8vIHByaW50cyBmYWxzZSwgYSBidWc_XG4gICAgY29uc29sZS5sb2codW5kZWZpbmVkID09PSBjaGFuZ2VkUHJvcGVydGllcy5nZXQoJ3Byb3AyJykpOyAvLyBwcmludHMgdHJ1ZSwgT0tcbiAgfVxuXG4gIHVwZGF0ZWQoY2hhbmdlZFByb3BlcnRpZXMpIHtcbiAgICBzdXBlci51cGRhdGVkKGNoYW5nZWRQcm9wZXJ0aWVzKTtcbiAgICBjb25zb2xlLmxvZygnLS0tdXBkYXRlZCcpO1xuICAgIGNvbnNvbGUubG9nKHVuZGVmaW5lZCA9PT0gY2hhbmdlZFByb3BlcnRpZXMuZ2V0KCdwcm9wMScpKTsgLy8gcHJpbnRzIGZhbHNlLCBhIGJ1Zz9cbiAgICBjb25zb2xlLmxvZyh1bmRlZmluZWQgPT09IGNoYW5nZWRQcm9wZXJ0aWVzLmdldCgncHJvcDInKSk7IC8vIHByaW50cyB0cnVlLCBPS1xuICB9XG5cbiAgcGVyZm9ybVVwZGF0ZSgpIHtcbiAgICBpZiAoIXRoaXMuaXNVcGRhdGVQZW5kaW5nKSB7XG4gICAgICByZXR1cm47XG4gICAgfVxuICAgIC8vIHRoaXMgbG9vcCBpcyBkdXBsaWNhdGVkIGluIHRoZSBzdXBlciBjYWxsIGJ1dCB0aGUgd29yayB3b24ndCBvY2N1ciB0d2ljZSBkdWUgdG8gdGhlIFxuICAgIC8vIGBfJGNoYW5nZWRQcm9wZXJ0aWVzYCBjaGVja1xuICAgIGlmICghdGhpcy5oYXNVcGRhdGVkKSB7XG4gICAgICBjb25zdCB7ZWxlbWVudFByb3BlcnRpZXN9ID0gdGhpcy5jb25zdHJ1Y3RvcjtcbiAgICAgIGZvciAoY29uc3QgW3AsIG9wdGlvbnNdIG9mIGVsZW1lbnRQcm9wZXJ0aWVzKSB7XG4gICAgICAgIGlmIChvcHRpb25zLndyYXBwZWQgPT09IHRydWUgJiYgIXRoaXMuXyRjaGFuZ2VkUHJvcGVydGllcy5oYXMocCkgJiZcbiAgICAgICAgICAgIHRoaXNbcF0gIT09IHVuZGVmaW5lZCkge1xuICAgICAgICAgIHRoaXMuXyRjaGFuZ2VQcm9wZXJ0eShwLCB1bmRlZmluZWQsIG9wdGlvbnMpO1xuICAgICAgICB9XG4gICAgICB9XG4gICAgfVxuICAgIHN1cGVyLnBlcmZvcm1VcGRhdGUoKTtcbiAgfVxufVxuXG5jdXN0b21FbGVtZW50cy5kZWZpbmUoJ3NpbXBsZS1ncmVldGluZycsIFNpbXBsZUdyZWV0aW5nKTsifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHNpbXBsZS1ncmVldGluZyBwcm9wMT1cImluaGVyaXRlZFwiIHByb3AyPVwiaW5oZXJpdGVkXCI-PC9zaW1wbGUtZ3JlZXRpbmc-XG48L2JvZHk-XG4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXRcIjogXCJeMy4wLjBcIixcbiAgICBcIkBsaXQvcmVhY3RpdmUtZWxlbWVudFwiOiBcIl4yLjAuMFwiLFxuICAgIFwibGl0LWVsZW1lbnRcIjogXCJeNC4wLjBcIixcbiAgICBcImxpdC1odG1sXCI6IFwiXjMuMC4wXCJcbiAgfVxufSIsImhpZGRlbiI6dHJ1ZX1d), where the 'previous' value of the 1st `willUpdate()` and `updated()` callbacks is not `undefined` (which was revealed by some of our Chromium tests failing when we deploy the 'accessor' keyword + our workaround more widely, which are failing [this assertion](https://source.chromium.org/chromium/chromium/src/+/main:ui/webui/resources/cr_elements/cr_icon/cr_iconset.ts;l=55;drc=6c63c64b3441a233a29ce3c521532c7d6dbc93f7)).

Do you know if the fix in this PR addresses this, or if this is just a problem with the temporary workaround we discussed in earlier comments. Is there a way to make these cases behave identically across wrapped and non-wrapped properties?

PR #4935, opened 2025-02-19
fix: widen React peer dependency range to include React 19

tibnor:
Earlier #4865 allows @lit/react to accept React 19 as peerdependency. This patch updates the generated package.json to include this version when using @lit-labs/gen-wrapper-react.

### DIFF
--- a/.changeset/orange-oranges-hammer.md
+++ b/.changeset/orange-oranges-hammer.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/gen-wrapper-react': patch
+---
+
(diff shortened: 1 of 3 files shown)


augustjk (commented):
thanks for the PR! the test goldens here should be updated as well https://github.com/lit/lit/blob/a66737fc9b861999b00ccad01edb925172b7f711/packages/labs/gen-wrapper-react/goldens/test-element-a/package.json#L14-L15

augustjk (approved):
Thanks!

Issue #4860, opened 2024-12-07
[@lit-labs/signals] `watch` prevents re-render when an unrelated signal changes


tjcrowder:
### Which package(s) are affected?

Other/unknown (please mention in description)

### Description

Using `@lit-labs/signals`, using `watch` on one signal sometimes prevents `render` from being called when an *unrelated* signal changes. It doesn't appear to happen with *all* other signals, but I have a case that replicates it. Here's the component:

```typescript
import { LitElement, css, html } from "lit";
import { customElement } from "lit/decorators.js";
import { signal, SignalWatcher, watch } from "@lit-labs/signals";

@customElement("lit-list")
export class LitListElement extends SignalWatcher(LitElement) {
    #items = signal<string[]>([]);

    #item = signal("");

    render() {
        const items = this.#items.get();
        console.log(`render: ${items.length}`);
        return html`<div>
            <div>
                <label
                    >Item:
                    <input
                        type="text"
                        .value=${watch(this.#item)}
                        @input=${this.#itemInput}
                /></label>
                <button @click=${this.#onAddItemClick}>Add</button>
            </div>
            <ul>
                ${items.map((item) => html`<li>${item}</li>`)}
            </ul>
        </div> `;
    }

    #itemInput(e: InputEvent) {
        const value = (e.target as HTMLInputElement).value;
        this.#item.set(value);
    }

    #onAddItemClick(e: MouseEvent) {
        e.preventDefault();
        const item = this.#item.get();
        const items = this.#items.get();
        const newItems = [...items, item];
        console.log(
            `Adding "${item}" to [${items.map((item) => `"${item}"`)}] (${
                items.length
            }) => [${newItems.map((item) => `"${item}"`)}] (${newItems.length})`
        );
        this.#items.set(newItems);
    }

    static styles = css`
        :host {
            display: block;
        }
    `;
}

declare global {
    interface HTMLElementTagNameMap {
        "lit-list": LitListElement;
    }
}
```

`this.#item` is a signal around a string for the current item being typed, and `this.#items` is a signal around an array for the list of items. With the code above, typing an item and clicking Add *does* add the item to `this.#items`, but doesn't cause a re-render. As I understand it, since we're reading the value of the `this.#items` signal within `render`, changing it should cause a re-render.

If you change the `.value` on the `input`, though, from `.value=${watch(this.#item)}` to `.value=${this.#item.get()}` (e.g., don't use `watch`), the problem goes away and the component re-renders when `this.#items` changes.

[...]

PR #4941, opened 2025-02-27
watch no longer triggers update; adds SignalWatcher.effect

sorvell:
Fixes https://github.com/lit/lit/issues/4860.

* `watch` no longer triggers a reactive update on a `SignalWatcher` element
* `watch` can be used without `SignalWatcher`
* adds `SignalWatcher.effect` which is called by default after rendering updates but can optionally include an options `{beforeUpdate: true}` to process the effect before any rendering updates.


packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,79 @@
...
-) => ReactiveElement;
 
 export interface SignalWatcher extends ReactiveElement {
-  _updateWatchDirective(d: WatchDirective<unknown>): void;
-  _clearWatchDirective(d: WatchDirective<unknown>): void;
+  _watcher?: Signal.subtle.Watcher;
+}
+
+interface EffectOptions {
+  beforeUpdate?: boolean;
```
justinfagnani, line 19: Can we have jsdocs on the options directly? It's nice when hovering over the option names in an IDE.

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,79 @@
...
+    effectWatcher.watch();
+  });
+});
+
+/**
+ * Executes the provided callback function when any of the signals it accesses
+ * change. If an options object is provided, the `element` property can be used
+ * to specify the element to associate the effect with. The `beforeUpdate`
+ * property can be used to specify that the effect should run before the element
+ * updates. The `manualDispose` property can be used to specify that the effect
```
justinfagnani: What's the use case for `manualDispose`?

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,79 @@
...
+ * should not be automatically disposed when the element is disconnected.
+ *
+ * @param callback
+ * @param options {element, beforeUpdate, manualDispose}
+ * @returns
```
justinfagnani: remove `@returns` or add the type?

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,79 @@
...
+  _watcher?: Signal.subtle.Watcher;
+}
+
+interface EffectOptions {
+  beforeUpdate?: boolean;
+  manualDispose?: boolean;
+}
+
+export interface ElementEffectOptions extends EffectOptions {
+  element?: SignalWatcher & SignalWatcherApi;
```
justinfagnani: JSDocs here?

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,79 @@
...
+    return () => {
+      effectWatcher.unwatch(computed);
+    };
+  } else {
+    return element._effect(callback, options);
+  }
+};
+
+interface SignalWatcherApi {
+  _effect(fn: () => void, options?: EffectOptions): () => void;
```
justinfagnani: Should this be marked `/** @internal */`?

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -48,24 +101,61 @@ const elementForWatcher = new WeakMap<
...
+      });
+      const pendingWatches = this._watcher
+        ?.getPending()
+        .filter(
+          (signal) =>
+            signal !== this.__performUpdateSignal && !this.__effects.has(signal)
+        );
+      beforeEffects.forEach((signal) => signal.get());
+      this.__performUpdateSignal?.get();
+      pendingWatches!.forEach((signal) => signal.get());
```
justinfagnani, line 118: Should the pending watches only be run if the full update didn't run? Otherwise those bindings will run twice?

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -87,25 +177,70 @@ export function SignalWatcher<T extends ReactiveElementConstructor>(
...
 
     private __unwatch() {
-      if (this.__watcher === undefined) {
+      if (this._watcher === undefined) {
         return;
       }
-      this.__watcher.unwatch(this.__performUpdateSignal!);
+      // We unwatch all signals that are not manually disposed, so that we don't
+      // keep the element alive by holding references to it.
+      this._watcher.unwatch(
```
justinfagnani, line 181: Do we do this manually because of manualDispose, or is there also no bulk way to dispose of the whole watcher?

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Happy to, but not sure where it goes. Can you point it out in the code?

justinfagnani: You just put them on the option property:

```ts
interface EffectOptions {
  /**
   * Whether to run the effect before the element's update. Defaults to `false`.
   */
  beforeUpdate?: boolean;

  /** ... */
  manualDispose?: boolean;
}
```

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Clarified docs

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Done

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Done

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: No, the get won't provoke it running twice if the value hasn't changed, right?

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Because of `manualDispose`.

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: @justinfagnani if we only run effect for element, should we just expose it directly here? Maybe as `signalEffect` to avoid taking the generic name?

justinfagnani: Yeah, I think `this.signalEffect()` is ok, thought maybe redundant since "effect" is strongly associated with signals in this world already. `this.updateEffect()` might be ok too? Or does "update" sound like a verb in that usage?

I do think an instance method is fine for this though. As opposed to `effect(this, () => ...)`.

sorvell: changed to `updateEffect`

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Done

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Added test for `manualDispose`

packages/labs/signals/src/lib/signal-watcher.ts
justinfagnani: ah, correct

packages/labs/signals/src/lib/signal-watcher.ts
```diff
@@ -4,23 +4,60 @@
...
+  /**
+   * By default, effects are automatically disposed when the element is
+   * disconnected. If `manualDispose` is set to `true`, the effect will not
+   * be automatically disposed, and you must call the returned function to
+   * dispose of the effect manually.
+   */
+  manualDispose?: boolean;
+}
+
+export interface ElementEffectOptions extends EffectOptions {
```
justinfagnani: Is this interface used anywhere?

justinfagnani (approved):
Awesome!

Two small comments left

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Removed, no longer needed.

packages/labs/signals/src/lib/signal-watcher.ts
sorvell: Removed, no longer needed.

Westbrook:
Getting into some finer testing of this change as we await it landing and wondered which of these is "correct".

In https://lit.dev/playground/#gist=5cfb3e893dbc5412ed1468483be5fa80 we get the render lifecycle on all three buttons, but only the Signal bindings on the first and last (signal powered) element.

In https://lit.dev/playground/#gist=3b14cb4a58de364ec7686f230a88435d we do not get the render lifecycle in either of the Signal bindings.

Maybe there's changes in this PR before I lifted the code, but wanted to confirm as we're working out how we want to consumer these capabilities in my team right now.

sorvell:
Forwarded from [discord](https://discord.com/channels/1012791295170859069/1068571268607127643/1435397282597175390)

* This should be removed: https://lit.dev/docs/data/signals/#:~:text=For%20coordination%20purposes,template%20are%20skipped
* We changed this since signals are not part of the element's reactive properties so if they triggered an update it would be an update without any visible changes to properties.
* We've added `updateEffect` which lets you use signal values when signals change. This method is coordinated with the rendering impact that any of those signals have on the element and can be either before or after them, based on options e.g. `updateEffect(() => console.log(mySIgnal.get()), {beforeUpdate: true})`

PR #4949, opened 2025-03-21
[reactive-element] fixes inconsistent initial values in changed properties

### LINKED ISSUE #4916: [reactive-element] Initial changed properties can be inconsistent

### Description

Reactive properties defined via the static properties block v. the property decorator can have inconsistent initial changed values.

This is true for example, for properties defined via `accessor` or via custom `get/set`.

### Reproduction

https://lit.dev/playground/#gist=9ea51856fe6b624f8dc1d37b62692dd4

See console output. All the of `*j` and `*t` properties should match.

### Workaround

Hack fix:

```js
static createProperty(name, options) {
  if (this.prototype.hasOwnProperty(name)) {
    options = {...options, wrapped: true};
  }
  super.createProperty(name, options);
}
  
_$changeProperty(name, oldValue, options) {
  oldValue = this.hasUpdated ? oldValue : undefined;
  super._$changeProperty(name, oldValue, options);
}
```

### Is this a regression?

### Affected versions

3.2.1

All

sorvell:
Fixes #4916

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1317,32 +1317,33 @@ export abstract class ReactiveElement
   _$changeProperty(
     name: PropertyKey,
     oldValue: unknown,
-    options: PropertyDeclaration,
+    {useDefault, reflect, wrapped}: PropertyDeclaration,
     initializeValue?: unknown
   ) {
     // Record default value when useDefault is used. This allows us to
     // restore this value when the attribute is removed.
-    if (options.useDefault && !(this.__defaultValues ??= new Map()).has(name)) {
+    if (useDefault && !(this.__defaultValues ??= new Map()).has(name)) {
       this.__defaultValues.set(
         name,
         initializeValue ?? oldValue ?? this[name as keyof this]
       );
       // if this is not wrapping an accessor, it must be an initial setting
       // and in this case we do not want to record the change or reflect.
-      if (options.wrapped !== true || initializeValue !== undefined) {
+      if (wrapped !== true || initializeValue !== undefined) {
         return;
       }
     }
     // TODO (justinfagnani): Create a benchmark of Map.has() + Map.set(
     // vs just Map.set()
     if (!this._$changedProperties.has(name)) {
+      oldValue = !this.hasUpdated && !useDefault ? undefined : oldValue;
       this._$changedProperties.set(name, oldValue);
     }
     // Add to reflecting properties set.
     // Note, it's important that every change has a chance to add the
     // property to `__reflectingProperties`. This ensures setting
     // attribute + property reflects correctly.
-    if (options.reflect === true && this.__reflectingProperty !== name) {
+    if (reflect === true && this.__reflectingProperty !== name) {
       (this.__reflectingProperties ??= new Set<PropertyKey>()).add(name);
     }
   }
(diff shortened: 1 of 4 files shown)


freshp86:
> @freshp86 can you help verify that this version would address the issues you're seeing? Thanks.

Will give it a shot and report back, but unfortunately trying non-released versions of Lit within local Chromium builds  is not very straight forward.

freshp86:
> > @freshp86 can you help verify that this version would address the issues you're seeing? Thanks.
> 
> Will give it a shot and report back, but unfortunately trying non-released versions of Lit within local Chromium builds is not very straight forward.

I have checked out Lit, patched this PR, and successfully ran `npm run build` locally. Having said that, in order to try this change in the context of Chromium I need to recreate the files that are distributed by `npm install lit` and I can't find them. If you can help me recreate those files locally, then I can try verifying whether these fixes are sufficient for our case. Thanks!

sorvell:
>  If you can help me recreate those files locally

For quick testing, maybe you can just copy the ReactiveElement and LitElement out and make your base class extend from those?

freshp86:
> For quick testing, maybe you can just copy the ReactiveElement and LitElement out and make your base class extend from those?

Thanks! I have partially verified that these would work for us. For the record here is how I went about it.

* Checked out Lit locally, patched this PR, and succesfully ran `npm run build`
* Copied the generated `packages/reactive-element/reactive-element.js`file to Chromium's `third_party/node/node_modules/@lit/reactive-element/reactive-element.js` folder
* Copied the generated `packages/lit-element/lit-element.js`file to Chromium's `third_party/node/node_modules/lit-element/lit-element.js` .
* Note that we still use `"lit": "3.0.2"` in our package.json, so the files above have been copied on top of that version.
* Reverted the workarounds we added at https://chromium-review.googlesource.com/c/chromium/src/+/6349561 and https://chromium-review.googlesource.com/c/chromium/src/+/6373768 (did not revert the updated tests for `CrLitElement` class)
* Re-built our tests and ran all `CrLitElement` tests which passed.

I am not able to actually spawn our trybots with the modified Lit in Chromium's `third_party/`, to run all of our tests with these changes, so there is still a chance something might have been missed if not covered by the `CrLitElement` unit tests, but the fact that `CrLitElement` tests are passing with the workarounds reverted seems very promising.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1317,32 +1317,33 @@ export abstract class ReactiveElement
...
       // and in this case we do not want to record the change or reflect.
-      if (options.wrapped !== true || initializeValue !== undefined) {
+      if (wrapped !== true || initializeValue !== undefined) {
         return;
       }
     }
     // TODO (justinfagnani): Create a benchmark of Map.has() + Map.set(
     // vs just Map.set()
     if (!this._$changedProperties.has(name)) {
+      oldValue = !this.hasUpdated && !useDefault ? undefined : oldValue;
```
justinfagnani, line 1339: I slightly prefer an if that doesn't reassign to the existing value, for readability:

```suggestion
      // On the initial change, the old value should be `undefined`, except
      // with `useDefault`
      if (!this.hasUpdated && !useDefault) {
        oldValue = undefined;
      }
```

justinfagnani (approved):

PR #4960, opened 2025-03-31
[labs/nextjs] Enable plugin rule settings granular control

### LINKED ISSUE #4952: [@lit-labs/nextjs] Enable setting the plugin rule settings for granular control

### Description

The `@lit-labs/nextjs` plugin webpack rule settings currently target the common entry points for typical Next.js apps: `pages/*.js`  and `app/*.js` (or the jsx|tsx|tx equivalents).

This works well for most applications, though I'm currently setting this plugin in an application with a slightly different Next.js setup. Instead of the entry points mentioned above, it has several, split into independent areas, like `menu/*.js`, `search/*.js`, `account/*.js`, and so on.

This is controlled by this line in the plugin: https://github.com/lit/lit/blob/main/packages/labs/nextjs/src/index.ts#L48

In order to bypass this limitation of the plugin, the most relevant parts of the plugin were copied into in the `next.config` file. 

This works to a certain extent, however, it is not ideal as it will not receive plugin updates for its code or dependencies. Another downside is to have to install the plugin dependencies such as `@lit-labs/ssr-react` in the host app.

If the plugin exposed the webpack rules settings, it'd give the necessary granular control, and having a copy of it in the `next.config` wouldn't be necessary:

```js
const withLitSSR = require('@lit-labs/nextjs')({
  webpackModuleRulesTest: /discovery\/index\.js/, // < This overrides the default set in line 48
});
```

The Next.js application in question is a large-scale one and it's rather unlikely its structure would be modified back to standard one.

Having this setting available for granular control would be rather beneficial for helping us in the integration of the Design System components developed with Lit.

Making a copy of the plugin code inside `next.config` and manually installing all of its dependencies.

fernandofranca:
Fixes #4952


.changeset/tall-gorillas-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/nextjs': minor
```
augustjk, line 2: this package is still on 0. major version so this non-breaking addition can be done with a patch.

```suggestion
'@lit-labs/nextjs': patch
```

packages/labs/nextjs/README.md
```diff
@@ -42,14 +42,17 @@ e.g.
...
-| `addDeclarativeShadowDomPolyfill` | `boolean` | If `true`, the client bundle will include a script that applies the [Declarative Shadow DOM polyfill](https://github.com/webcomponents/template-shadowroot). Defaults to `true`. |
+| Property                          | Type            | Description                                                                                                                                                                      |
+| --------------------------------- | --------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `addDeclarativeShadowDomPolyfill` | `boolean`       | If `true`, the client bundle will include a script that applies the [Declarative Shadow DOM polyfill](https://github.com/webcomponents/template-shadowroot). Defaults to `true`. |
+| `webpackModuleRulesTest`          | `RegExp`        | Allows to pass a specific RegExp to match the files to be processed by the plugin. Defaults to `/\/pages\/.*\.(?:j\|t)sx?$\|\/app\/.*\.(?:j\|t)sx?$/`.                           |
```
augustjk, line 54:
```suggestion
| `webpackModuleRulesTest`          | `RegExp`        | Configure the RegExp used to inject Lit SSR support. Ideally it should match the entrypoint to your routes. Defaults to `/\/pages\/.*\.(?:j\|t)sx?$\|\/app\/.*\.(?:j\|t)sx?$/`.                           |
```

augustjk (commented):
thanks! just a couple suggested changes.

augustjk (approved):

PR #4965, opened 2025-04-08
Add missing aria attributes.

rictic:


rictic:
Ah, this will need to be done with the next TypeScript compiler update

justinfagnani:
I think https://github.com/lit/lit/pull/4984/files supersedes this one

PR #4968, opened 2025-04-09
Minor changes to EventTarget polyfill for closure compiler.

rictic:


packages/labs/ssr-dom-shim/src/lib/events.ts
```diff
@@ -278,10 +274,9 @@ const EventTargetShim = class EventTarget
     }
     return finishDispatch();
   }
-};
+}
 
-const EventTargetShimWithRealType =
-  EventTargetShim as object as typeof EventTarget;
+const EventTargetShimWithRealType = EventTarget as object as typeof EventTarget;
```
kevinpschaaf, line 279: What is this doing, exactly?

kevinpschaaf, line 279: What is this doing, exactly? `EventTarget as object as typeof EventTarget`

kevinpschaaf (approved):

packages/labs/ssr-dom-shim/src/lib/events.ts
rictic: Oh, good catch. That should be `EventTarget as object as typeof globalThis.EventTarget`.

Fixed.

PR #4975, opened 2025-04-14
[lit-html] Remove some redundant code from removePart()

justinfagnani:
Also fixes a directive-helpers test that was incorrectly constructing its test case.

I noticed that `removePart()` was almost the same as `ChildPart.$_clear()` with the exception of removing the start node. I also optimized `ChildPart.$_clear()` a little and noticed that the `insertPart()` test was wrong.

### DIFF
--- a/packages/lit-html/src/test/directive-helpers_test.ts
+++ b/packages/lit-html/src/test/directive-helpers_test.ts
@@ -215,10 +215,16 @@ suite('directive-helpers', () => {
         // Create two parts and remove the first, then the second to make sure
         // that removing the first doesn't move the second's markers. This
         // fails if the parts accidentally share a marker.
-        const childPart2 = insertPart(part, undefined);
-        const childPart1 = insertPart(part, undefined, childPart2);
+        const childPart2 = insertPart(part);
+        const childPart1 = insertPart(part, childPart2);
+
+        // Check that the test is correctly inserting two different parts:
+        assert.notEqual(childPart1, childPart2);
+        assert.notEqual(childPart1._$endNode, childPart2._$startNode);
+
         removePart(childPart1);
         removePart(childPart2);
+
         return v;
       }
     }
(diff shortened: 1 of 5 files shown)


packages/lit-html/src/test/directive-helpers_test.ts
```diff
@@ -215,10 +215,16 @@ suite('directive-helpers', () => {
         // Create two parts and remove the first, then the second to make sure
         // that removing the first doesn't move the second's markers. This
         // fails if the parts accidentally share a marker.
-        const childPart2 = insertPart(part, undefined);
-        const childPart1 = insertPart(part, undefined, childPart2);
+        const childPart2 = insertPart(part);
+        const childPart1 = insertPart(part, childPart2);
```
justinfagnani, line 219: Previously the test mixed up the parameter for the refPart and the part to be inserted. Passing the third parameter meant that the second call wasn't making a new part.

rictic:
This passes all of Google's internal tests!

justinfagnani:
Thanks for running the google3 tests, @rictic!

jimfranke:
@justinfagnani It looks like this change causes some unforseen behaviour. When using the CustomElement below to reshuffle a list of items with varying length via the `repeat` directive by pressing the button, I see markers `<!---->` being added and never being removed within the `ul` element; resulting in hundreds of dangling markers. When reverting this PR everything seems to behave as expected.

Could you please verify? Thanks!

```
<!doctype html>
<html lang="en">

<head>
  <meta charset="utf-8" />
  <title>App</title>

  <script type="module">
    //*! Import 'html' and 'render' versions from this PR !*/
    
    // Import the latest 'repeat' directive
    import { repeat } from 'https://cdn.jsdelivr.net/npm/lit-html@3.3.0/directives/repeat.js';

    const fruits = [{
        name: 'apple',
        text: '🍎'
      },
      {
        name: 'banana',
        text: '🍌'
      },
      {
        name: 'grapes',
        text: '🍇'
      },
      {
        name: 'pineapple',
        text: '🍍'
      },
      {
        name: 'peach',
        text: '🍑'
      },
      {
        name: 'cherry',
        text: '🍒'
      },
      {
        name: 'greenapple',
        text: '🍏'
      },
      {
        name: 'melon',
        text: '🍉'
      },
      {
        name: 'strawberry',
        text: '🍓'
      },
    ];

    customElements.define(
      'x-app',
      class App extends HTMLElement {
        #fruits = [...fruits];

        connectedCallback() {
          this.render();
        }

        shuffleFruits = () => {
          const size = Math.floor(Math.random() * fruits.length) + 1;

          this.#fruits = fruits.sort(() => Math.random() - 0.5).slice(-size);
          console.log(this.#fruits.concat());
          this.render();
        };

        render() {
          render(
            html`
          <main>
            <button @click=${this.shuffleFruits}>Shuffle Fruits</button>

            <ul>
              ${repeat(
                this.#fruits,
                fruit => fruit.name,
                fruit =>
                  html`<li>
                    <label>
                      <input name=${fruit.name} type="checkbox" />
                      ${fruit.name} - ${fruit.text}
                    </label>
                  </li>`,
              )}
            </ul>
          </main>
        `,
            this,
          );
        }
      },
    );
  </script>
</head>

<body>
  <x-app></x-app>
</body>

</html>
```

justinfagnani:
@jimfranke since this PR is merged, can you open a new issue so we can be sure to track it?

justinfagnani:
I do think I see the problem here. We need some more tests for repeat and directive helpers.

jimfranke:
> @jimfranke since this PR is merged, can you open a new issue so we can be sure to track it?

Thanks, sure! https://github.com/lit/lit/issues/5010

PR #4976, opened 2025-04-14
fromAttribute can set the property to either null or undefined.

rictic:
Fixes a regression from https://github.com/lit/lit/pull/4934 which collapsed all converted nullish values to null.


justinfagnani (approved):

.changeset/many-grapes-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit/reactive-element': patch
+---
```
augustjk, line 3:
```suggestion
---
'lit': patch
'lit-element': patch
'@lit/reactive-element': patch
---
```

Issue #4273, opened 2023-10-09
Make @lit-labs/rollup-plugin-minify-html-literals available via npm


JamesIves:
### Should this be an RFC?

- [X] This is not a substantial change

### Which package is this a feature request for?

Other/unknown (please mention in description)

### Description

I would like to use [@lit-labs/rollup-plugin-minify-html-literals](https://github.com/lit/lit/tree/main/packages/labs/rollup-plugin-minify-html-literals) in my rollup configuration as [rollup-plugin-minify-html-literals](https://github.com/asyncLiz/rollup-plugin-minify-html-literals) no longer appears to be maintained, however it does not appear to be published on the npm registry.

```
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@lit-labs%2frollup-plugin-minify-html-literals - Not found
```

Is there any chance we could get a version of this published, or is there another way to install this? There's limited options for this, and multiple code options maintained by the same groups.

### Alternatives and Workarounds

The other plugin for this purpose seems to have multiple issues, as a result any project moving towards Node 20 has no alternative. I'm somewhat confused in terms of what is the preferred option to use here.

JamesIves:
Related issue: https://github.com/asyncLiz/rollup-plugin-minify-html-literals/issues/29

FluorescentHallucinogen:
+1

fernandopasik:
Any ETA when this npm package would be released?

redfox-mx:
Is anyone here? Any plans for this?

Frank3K:
> If not, are there any alternatives?

There's https://github.com/ouweiya/rollup-plugin-minify-html-literals-v3.

malobre:
There's also [rollup-plugin-minify-template-literals](https://github.com/explodingcamera/esm/tree/main/packages/rollup-plugin-minify-template-literals)

yhaiovyi-tripactions:
@malobre it's quite buggy

bnf:
> If not, are there any alternatives?

Hi, we've created a tiny lit template minimizer called [litnano](https://www.npmjs.com/package/litnano).
It is an asynchronous replacement for the synchronous (and unfortunately unreleased) minify-html-literals variants in the wild.

It uses [htmlnano](https://github.com/posthtml/htmlnano) (allows asynchronous css processing) and [cssnano](https://github.com/cssnano/cssnano) (tends to support modern css features earlier than clean-css) for minification and supports integration in rollup and webpack out of the box.

[...]

PR #4977, opened 2025-04-16
[labs/rollup-plugin-minify-html-literals] prepare for release

bencbradshaw:
Addresses https://github.com/lit/lit/issues/4273

* update README to clarify what this package is
* clean up CHANGELOG and link to previous for old entries
* remove `private:true` from `package.json`


JamesIves:
Awesome! Thank you for putting this together.

packages/labs/rollup-plugin-minify-html-literals/README.md
```diff
@@ -1,5 +1,18 @@
...
+2. [minify html literals](https://github.com/asyncLiz/minify-html-literals)
+3. [parse literals](https://github.com/asyncLiz/parse-literals)
+
+Changed since the originals:
+
+- all three projects have been merged into one
+- rollup 4 added to peerDependencies
+- testing is cleaned up and improved
+- source code has been refactored more closely follow Lit's code style
+
```
justinfagnani, line 15: I think that before talking about the project history, we should just say what the package does. New users might not care about the history bits.

```suggestion
A Rollup plugin to minify HTML and CSS markup inside JavaScript template literal strings.
```

Any history should go into an `# Acknowledgements` section at the end.

justinfagnani (changes requested):

packages/labs/rollup-plugin-minify-html-literals/README.md
bencbradshaw: Thanks Justin. I adjusted the README based on your feedback.

justinfagnani:
@bencbradshaw this needs a changeset, even an empty one (`npx changeset --empty`).

And the wireit failure might be because of a GitHub cache update. Is this branch sync'ed to the latest main?

augustjk (approved):
Thank you!
I reset the package version to `0.0.0` and added a changeset so the next release cycle will bump this up to 0.1.0 and add the changeset content to the changelog file.
https://github.com/lit/lit/pull/4979 should fix the CI issues.

PR #4979, opened 2025-04-25
[infra] update wireit

augustjk:
Fixes wireit errors in GHA. See https://github.com/google/wireit/issues/1297


justinfagnani:
@augustjk why disable the cache now on benchmarks?

augustjk:
Benchmarks also checkout to an older branch of lit where we had an older version of wireit so the cache causes errors.

justinfagnani:
> Benchmarks also checkout to an older branch of lit where we had an older version of wireit so the cache causes errors.

Wow. Ok, that's definitely necessary then!

justinfagnani (approved):
Look good. Can you add something to the README that mentions why the cache is disabled?

PR #4984, opened 2025-05-03
chore: update TypeScript dependencies to version 5.9 with related ARIAMixin changes

kyubisation:
This PR updates all TypeScript dependencies to version 5.9, adds .d.ts compilation checks for older TypeScript versions (see `packages/tests-typescript/package.json`) and adds newly added ARIAMixin properties (ariaColIndexText, ariaRelevant and ariaRowIndexText) to the ElementInternals SSR shim.

### DIFF
--- a/packages/labs/ssr-dom-shim/src/lib/element-internals.ts
+++ b/packages/labs/ssr-dom-shim/src/lib/element-internals.ts
@@ -4,8 +4,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+type StringKeys<T extends object> = {
+  [K in keyof T]: T[K] extends string | null ? K : never;
+}[keyof T];
+
 type ARIAAttributeMap = {
-  [K in keyof ARIAMixin]: string;
+  [K in StringKeys<ARIAMixin>]: string;
 };
 
 /**
@@ -66,6 +70,7 @@ type ElementInternalsInterface = ElementInternals;
 export const ElementInternalsShim = class ElementInternals
   implements ElementInternalsInterface
 {
+  ariaActiveDescendantElement = null;
   ariaAtomic = '';
   ariaAutoComplete = '';
   ariaBrailleLabel = '';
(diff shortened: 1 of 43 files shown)


packages/labs/ssr-dom-shim/src/lib/element-internals.ts
```diff
@@ -4,8 +4,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+type StringKeys<T extends object> = {
```
justinfagnani, line 7: Why was this necessary here? Did ARIAMixin add new non-string keys? Maybe leave a comment?

packages/localize-tools/src/messages.ts
```diff
@@ -99,7 +99,7 @@ export function sortProgramMessages(
   messages: ProgramMessage[]
 ): ProgramMessage[] {
   return messages.sort((a, b) => {
-    if (a.desc ?? '' !== b.desc ?? '') {
+    if ((a.desc ?? '') !== (b.desc ?? '')) {
```
justinfagnani, line 102: Was there a formatter update?

justinfagnani (approved):
Still LGTM! Just two small questions...

packages/labs/ssr-dom-shim/src/lib/element-internals.ts
kyubisation: Yes, e.g. https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/ariaControlsElements
I will add a comment.

packages/localize-tools/src/messages.ts
kyubisation: No, but TypeScript had an error that the middle part would always evaluate to a non-nullish value.
So maybe the question is whether the intention is `(a.desc ?? '') !== (b.desc ?? '')` or `a.desc ?? ('' !== b.desc) ?? ''`.
From my understanding it was the former.

PR #4988, opened 2025-05-14
fix(lit-html,lit,lit-element): adjust ssr directive patch condition

ADNolan:
### Description
Adjusted the comparison to use the Function.name property of the `_$resolve` function and the `resolveOverrideFn` in private ssr support to prevent attempted duplicative patching of a directive class.

It appears that after the first patching the directive, like `classMap`, that the conditional check remains `false`. The member/method on the class is still `_$resolve`, but its name is `ssrResolve`. Further processing of checking the Prototype of the directive with `proto.hasOwnProperty(resolveMethodName)` since `ssrResolve` doesn't exist. That path leads to an error being thrown and breaks SSR for that render.

Hopefully addresses Issue #4527

### DIFF
--- a/.changeset/five-deers-own.md
+++ b/.changeset/five-deers-own.md
@@ -1,7 +1,7 @@
 ---
-'lit-html': minor
-'lit': minor
-'lit-element': minor
+'lit-html': patch
+'lit': patch
+'lit-element': patch
 ---
 
 Adjusted the comparison to use the name property of the \_$resolve function and the resolveOverrideFn in private ssr support to prevent duplicated patching of the directive class.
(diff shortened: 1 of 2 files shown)


JamesIves:
@augustjk  👀

.changeset/five-deers-own.md
```diff
@@ -0,0 +1,7 @@
+---
+'lit-html': minor
+'lit': minor
+'lit-element': minor
```
augustjk, line 4: it's a fix for SSR support and not a new feature so patch seems appropriate

```suggestion
'lit-html': patch
'lit': patch
'lit-element': patch
```

augustjk (approved):
thank you for getting to the bottom of this!
i think this is a reasonable change. i suppose we never originally figured we could somehow get a new reference to the `ssrResolve` function if it's coming from the same `@lit-labs/ssr` code.

PR #4989, opened 2025-05-16
pin changeset action version

augustjk:
There's a bug in changesets/action that causes an error when using sub directories: https://github.com/changesets/action/issues/475#issuecomment-2871206995

Example of failed release workflow: https://github.com/lit/lit/actions/runs/14846949543/job/41682877633

Pinning to a previous working version until it's fixed.


augustjk:
action has been fixed and this is no longer needed

PR #4970, opened 2025-04-11
style(copilot-instructions): add a copilot instructions file

e111077:
Added copilot instructions. Please bikeshed, copilot instructions docs recommend a very short instructions

| Parent | Children |
| -- | -- |
| [#4991](https://github.com/lit/lit/pull/4991) | [#4971](https://github.com/lit/lit/pull/4971), [#4990](https://github.com/lit/lit/pull/4990) |

PR #4991, opened 2025-05-17
[ci] fix two broken CI issues

e111077:
CI is broken based on the kernel used by GH Actions. We are using unbuntu-latest. [See this CI error](https://github.com/lit/lit/actions/runs/15080547701/job/42397010351?pr=4970). Pretty sure we can trust the content we are running so we should be able to just turn the sandbox off since this just generates an image.

It was also broken because it seems that Vue introduced a new hidden slot with the slotname `__` which seems to hold metadata we were not expecting. [Related CI failure](https://github.com/lit/lit/actions/runs/15080964528/job/42397632767?pr=4991)

| Parent | Children |
| -- | -- |
| (main) | [#4970](https://github.com/lit/lit/pull/4970) |

### DIFF
--- a/.changeset/lovely-mangos-deny.md
+++ b/.changeset/lovely-mangos-deny.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/vue-utils': patch
+---
+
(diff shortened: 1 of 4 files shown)


justinfagnani (approved):
🎉

Issue #4986, opened 2025-05-09
[labs/ssr-react] guard against `enable-lit-ssr.js` wrapping createElement multiple times


augustjk:
### Should this be an RFC?

- [x] This is not a substantial change

### Which package is this a feature request for?

Other/unknown (please mention in description)

### Description

in server frameworks like next.js, it may be possible that [`enable-lit-ssr.js`](https://github.com/lit/lit/blob/main/packages/labs/ssr-react/src/node/enable-lit-ssr.ts) might run multiple times in the server, leading to `createElement` being wrapped multiple times.

part of the observation described https://github.com/lit/lit/issues/4527#issuecomment-2860407265 and from the stackblitz example shared there, refreshing the page causes error because multiple `<template shadowrootmode="open">` elements are added to a single custom element.

we're already setting a global flag from this so we could guard the patching of `createElement` behind that flag too.

### Alternatives and Workarounds

.

ADNolan:
I was tinkering with this a bit, not sure if the idea was/is to have a guard like this
```js
if (process.env.NODE_ENV === 'production' && !globalThis.litSsrReactEnabled) {
    Object.assign(ReactJSXRuntime.default ?? ReactJSXRuntime, {
        jsx: wrapJsx(ReactJSXRuntime.jsx, ReactJSXRuntime.jsxs),
        jsxs: wrapJsxs(ReactJSXRuntime.jsxs),
    });
}
else if(!globalThis.litSsrReactEnabled) {
  Object.assign(ReactJSXDevRuntime.default ?? ReactJSXDevRuntime, {
        jsxDEV: wrapJsxDev(ReactJSXDevRuntime.jsxDEV),
    });
}
```

This doesn't seem to solve the double shadowroot problem. It seems to originate from the `lit-element-hydrate-support` for the client.

First render:
```text
page.js:436 DEVELOPMENT: Lit-HTML Private SSR Support
17:08:42.753 page.js:392 Development Lit-HTML Directive
17:08:42.756 page.js:238 DEVELOPMENT: litElementHydrateSupport for client
17:08:42.766 page.js:370 Patching LitElement with litElementHydrateSupport
17:08:42.772 page.js:403 DEVELOPMENT: Client StyleMap Directive
17:08:42.774 page.js:51 litElementHydrateSupport: called connectedCallback <my-divider defer-hydration orientation=​"horizontal">​…​</my-divider>​
[...]

PR #4993, opened 2025-05-20
fix(@lit-labs/ssr-react,@lit-labs/nextjs): Enable ssr create element patch

ADNolan:
### Description
Added a guard to check if `React.createElement` was already patched with `litPatchedCreateElement` to prevent adding to the stack multiple calls to `litPatchedCreateElement`. 

Addresses issue #4986

### DIFF
--- a/packages/labs/ssr-react/src/node/enable-lit-ssr.ts
+++ b/packages/labs/ssr-react/src/node/enable-lit-ssr.ts
@@ -20,9 +20,12 @@ import {wrapJsx, wrapJsxDev, wrapJsxs} from '../lib/node/wrap-jsx.js';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 // The `.default ??` is used to get around inconsistent behavior with how
 // tools like webpack seem to do es module interop.
-Object.assign((React as any).default ?? React, {
-  createElement: wrapCreateElement(React.createElement),
-});
+if (React.createElement.name !== 'litPatchedCreateElement') {
+  Object.assign((React as any).default ?? React, {
+    createElement: wrapCreateElement(React.createElement),
+  });
+}
+
 if (process.env.NODE_ENV === 'production') {
   Object.assign((ReactJSXRuntime as any).default ?? ReactJSXRuntime, {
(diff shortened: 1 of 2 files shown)


ADNolan:
@augustjk just a ping since its React SSR related and we have had some comments back and forth on the issues.

packages/labs/ssr-react/src/node/enable-lit-ssr.ts
```diff
@@ -20,9 +20,12 @@ import {wrapJsx, wrapJsxDev, wrapJsxs} from '../lib/node/wrap-jsx.js';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 // The `.default ??` is used to get around inconsistent behavior with how
 // tools like webpack seem to do es module interop.
-Object.assign((React as any).default ?? React, {
-  createElement: wrapCreateElement(React.createElement),
-});
+if (React.createElement.name !== 'litPatchedCreateElement') {
```
justinfagnani, line 23: Couldn't `litPatchedCreateElement` be minified? I think it'd be safer to add a well-known property to the function in `wrap-create-element.ts` and check for that here. The property should be use computed property syntax to protect against minifiers and compilers.

augustjk: This is code meant to run in the server, though I suppose it's not impossible that server code could be minified.

digitalsadhu:
I have this issue and when I manually applied the fix proposed in this PR into my local node_modules, it solved the issue for me. 👍

packages/labs/ssr-react/src/node/enable-lit-ssr.ts
augustjk: Since we have confirmation this fixes the issue for others, I'm happy to merge this as is. We can address it then if minification becomes an issue.

augustjk (approved):
Thank you!

raoufswe:
Hi @augustjk @ADNolan 
Thanks so much for looking into this and getting it done 🙌
I'm keen to test the fix across a few different Next.js applications. Would it be possible to cut a release with the changes?

PR #4999, opened 2025-05-22
[reactive-element] Fix a bug in change detection with decorated standard private accessors

### LINKED ISSUE #4998: [reactive-element] Reactive properties don't handle private names properly

### Description

When a standard private field is decorated with `@state()` the name of the field is something like `'#foo'`, which `requestUpdate()` then uses to get the current value of the field, here: https://github.com/lit/lit/blob/6792b7ef422a090374e67581b5c8f07dd779edc9/packages/reactive-element/src/reactive-element.ts#L1289

`this['#foo']` does not return the value of `this.#foo`, it is an entirely different property, so the change detection logic fails, specifically when changing value from `undefined` (since the new value will appear to be `undefined` and so no change is detected.

A fix would be to allow `requestUpdate()` to take the new value as a parameter instead of looking it up by name.

### Reproduction

None yet. The playground doesn't support standard accessors yet.

### Workaround

Use non-standard TypeScript private fields.

### Is this a regression?

### Affected versions

3.x

All

justinfagnani:
Fixes #4998

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -1270,12 +1270,20 @@ export abstract class ReactiveElement
    * @param oldValue old value of requesting property
    * @param options property options to use instead of the previously
    *     configured options
+   * @param useNewValue if true, the newValue argument is used instead of
+   *     reading the property value. This is important to use if the reactive
+   *     property is a standard private accessor, as opposed to a plain
+   *     property, since private members can't be dynamically read by name.
+   * @param newValue the new value of the property. This is only used if
+   *     `useNewValue` is true.
    * @category updates
    */
   requestUpdate(
     name?: PropertyKey,
     oldValue?: unknown,
-    options?: PropertyDeclaration
+    options?: PropertyDeclaration,
+    useNewValue = false,
+    newValue?: unknown
   ): void {
     // If we have a property key, perform property update steps.
     if (name !== undefined) {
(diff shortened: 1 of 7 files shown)


packages/reactive-element/src/test/decorators-modern/decorators-standard/README.md
```diff
@@ -0,0 +1,3 @@
+This folder is excluded from the plain tsconfig that uses
+`experimentalDecorators: true` because it includes decorators on private class
+members which aren't allows by experimental decorators.
```
augustjk, line 3:
```suggestion
members which aren't allowed by experimental decorators.
```

packages/reactive-element/src/test/decorators-modern/decorators-standard/state_test.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
augustjk, line 3:
```suggestion
 * Copyright 2025 Google LLC
```

augustjk (approved):
need size check update as per comment above but otherwise looks good to me. suggestions for minor fixes.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1270,12 +1270,20 @@ export abstract class ReactiveElement
    * @param oldValue old value of requesting property
    * @param options property options to use instead of the previously
    *     configured options
+   * @param useNewValue if true, the newValue argument is used instead of
```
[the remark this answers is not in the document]
justinfagnani: I don't think so. Primitives can't be put in a WeakSet.

PR #5013, opened 2025-06-20
Generate herited properties in vue-wrapper

tibnor:


.changeset/hot-feet-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lit-labs/gen-wrapper-vue': patch
+---
+
+Add properties herited from mixins in vue wrapper
```
justinfagnani:
```suggestion
Add properties inherited from mixins in the Vue wrapper
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
   return '';
 };
 
+const getHeritageReactiveProperties = (
```
justinfagnani, line 149:
```suggestion
const getInheritedReactiveProperties = (
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -133,6 +167,8 @@ const wrapperTemplate = (
   wcPath: string
 ) => {
   const {tagname, events, reactiveProperties} = declaration;
+  const heritageProps = getHeritageReactiveProperties(declaration);
```
justinfagnani, line 191:
```suggestion
  const heritageProps = getInheritedReactiveProperties(declaration);
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
...
+): Map<string, ModelProperty> => {
+  const {heritage} = declaration;
+  if (
+    heritage == null ||
+    (heritage.superClass?.name === 'LitElement' && heritage.mixins.length === 0)
+  ) {
+    return new Map();
+  }
+  const props = heritage.mixins
+    .flatMap(
```
justinfagnani, line 160:
```suggestion
    .map(
```

`heritage.mixins` is a flat map of References, so we don't need `flatMap()` here. Even though `flatMap()` works, I prefer plain `map()` for clarity, since `flatMap()` to me implies that we're dealing with an array of arrays and I start wondering whether a depth of 1 works or not, etc.

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
...
+      (mixin) => (mixin.dereference() as MixinDeclaration).classDeclaration
+    )
+    .filter((mixin) => mixin.isLitElementDeclaration())
```
justinfagnani, line 163: I think this would read more clearly if the argument was named `mixinClass`

```suggestion
    .filter((mixinClass) => mixinClass.isLitElementDeclaration())
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
   return '';
 };
 
+const getHeritageReactiveProperties = (
+  declaration: LitElementDeclaration
+): Map<string, ModelProperty> => {
+  const {heritage} = declaration;
+  if (
+    heritage == null ||
```
justinfagnani, line 154: Out style is to always use `===` unless we know the value may be either `null` or `undefined`. In this case, it can't be `null`.

```suggestion
    heritage === undefined ||
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
...
+    .flatMap(
+      (mixin) => (mixin.dereference() as MixinDeclaration).classDeclaration
+    )
+    .filter((mixin) => mixin.isLitElementDeclaration())
+    .flatMap((mixin) => {
+      const props = Array.from(mixin.reactiveProperties.entries());
+      props.push(...getHeritageReactiveProperties(mixin).entries());
+      return props;
+    });
+  if (heritage.superClass != null) {
```
justinfagnani, line 169:
```suggestion
  if (heritage.superClass !== undefined) {
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
...
+  const props = heritage.mixins
+    .flatMap(
+      (mixin) => (mixin.dereference() as MixinDeclaration).classDeclaration
+    )
+    .filter((mixin) => mixin.isLitElementDeclaration())
+    .flatMap((mixin) => {
+      const props = Array.from(mixin.reactiveProperties.entries());
+      props.push(...getHeritageReactiveProperties(mixin).entries());
+      return props;
+    });
```
justinfagnani, line 168:
```suggestion
    .map((mixinClass) => {
      const props = Array.from(mixinClass.reactiveProperties.entries());
      props.push(...getHeritageReactiveProperties(mixinClass).entries());
      return props;
    });
```

packages/labs/gen-wrapper-vue/src/lib/wrapper-module-template-sfc.ts
```diff
@@ -124,6 +125,39 @@ const renderPropsParam = (reactiveProperties: Map<string, ModelProperty>) => {
...
+  if (heritage.superClass != null) {
+    const superClass =
+      heritage.superClass.dereference() as LitElementDeclaration;
+    if (superClass.isLitElementDeclaration()) {
+      props.push(...superClass.reactiveProperties.entries());
+    }
+    if (heritage.superClass.name !== 'LitElement') {
+      props.push(...getHeritageReactiveProperties(superClass).entries());
```
justinfagnani, line 176: The two recursive calls don't have test coverage, but they look correct. I think it's ok for now.

justinfagnani (changes requested):
Thanks you @tibnor! Sorry for the delay in review.

I made a bunch of small comments, but the overall shape looks great. I'll run the tests now.

PR #5035, opened 2025-08-13
Make our directives generic, for better type checking of lit templates

rictic:
If a directive's type is written such that it returns sufficiently specific DirectiveResults, the TypeScript type system can tell the effective type that a DirectiveResult will render to its part.

By itself, this doesn't do much, but I've also got a patch to rune's lit analyzer to better type check directives there.


packages/lit-html/src/directive.ts
```diff
@@ -16,8 +16,8 @@ export {
   PropertyPart,
 } from './lit-html.js';
 
-export interface DirectiveClass {
-  new (part: PartInfo): Directive;
+export interface DirectiveClass<RenderAs = unknown> {
```
justinfagnani: Why do you need this type parameter? You'd normally get the return type with `ReturnType<C['render']>`.

Is it so that you can feed in a type variable from an override type for `directive()`, like in Guard? If so, can you leave a comment?

justinfagnani (commented):
LGTM with one question.

I think you need to run format too.

packages/lit-html/src/directive.ts
rictic: We can actually use `infer` (and similar tricks in the TypeScript type system) to get around needing this at all. I've updated the PR to not need any changes to the common directive types, with a little care, a directive can just be written to put enough information on the DirectiveResults it makes.

packages/lit-html/src/directives/until.ts
```diff
@@ -107,6 +113,47 @@ export class UntilDirective extends AsyncDirective {
   }
 }
 
+interface Until {
+  <T>(val: T): DirectiveResult<typeof UntilDirective<T>>;
```
justinfagnani: Why can't this be:
```suggestion
  <T>(...args: Array<T>): DirectiveResult<typeof UntilDirective<T>>;
```
Won't TS infer `T` as a union of the arg types, just like you're doing manually?

rictic: Paired on fix.

justinfagnani (approved):

PR #5044, opened 2025-08-20
[lit-html] Make ClassInfo mutable

### LINKED ISSUE #5024: [lit-html] Provide write-able version of ClassInfo
### Should this be an RFC?

### Description

It would be useful to have a write-able version of ClassInfo&mdash;something like `ClassInfoBuilder` or similar&mdash;for constructing a ClassInfo object dynamically.

This would look like:

```typescript
export interface ClassInfoBuilder {
  [name: string]: string | boolean | number;
}

export type ClassInfo = Readonly<ClassInfoBuilder>;
```

And would help support patterns like:

```typescript
override render() {
  const classes: ClassInfoBuilder = {
    'class1': true,
    'class2': true,
  };
  // Conditionally add another class
  if (condition) classes['class3'] = true;

  // use it
  const containerClasses = classMap(classes);
  
  // etc.
}
```

Without a non-`readonly` version of ClassInfo, this would produce type errors like:

```
Index signature in type 'ClassInfo' only permits reading.
```

It's possible to use `any` for this, but that loses some type safety and clarity. Similarly, it's possible to define the -Builder version of ClassInfo, but it seems a little silly to define that interface per lit application.

ava-cassiopeia:
This allows for users to compose together a ClassList object dynamically. For example:

```typescript
override render() {
  const classes: ClassInfo = {
    'class1': true,
    'class2': true,
  };
  // Conditionally add another class
  if (condition) classes['class3'] = true;

  // use it
  const containerClasses = classMap(classes);
  
  // etc.
}
```

Closes #5024


.changeset/dirty-pillows-allow.md
```diff
@@ -0,0 +1,8 @@
+---
+"lit": minor
```
justinfagnani:
```suggestion
"lit": patch
```

justinfagnani (approved):

justinfagnani:
Thanks @ava-cassiopeia!

PR #5045, opened 2025-08-21
[labs/vscode-extension] Initial extension implementation

justinfagnani:
The "Hello World" of VS Code extensions...

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,7 +7,6 @@ node_modules/
 *.tgz
 *.tsbuildinfo
 .DS_Store
-.vscode/
 .wireit
 /temp
 .idea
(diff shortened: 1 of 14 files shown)


.eslintignore
```diff
@@ -7,7 +7,6 @@ node_modules/
 *.tgz
 *.tsbuildinfo
 .DS_Store
-.vscode/
```
rictic, line 10: Why remove this?

justinfagnani: Because I needed to check in launch.json and tasks.json for the extension. We had previously ignored it so we could have personal launches and tasks, but there must be some way to have both...

PR #5048, opened 2025-08-21
[labs/vscode-extension] Load the tsserver-plugin in the extension

justinfagnani:
The diagnostics don't always kick in automatically, but if you switch files, open a log, etc., they will.


kevinpschaaf (commented):
LGTM after https://github.com/lit/lit/pull/5049 lands

PR #5050, opened 2025-08-21
[labs/tsserver-plugin] getDefinitionAtPosition via HTMLElementTagNameMap

kevinpschaaf:


packages/labs/tsserver-plugin/src/lib/lit-language-service.ts
```diff
@@ -185,6 +203,44 @@ export const makeLitLanguageService = (
...
+      // Use the type checker to get the symbol for the ambient/global
+      // HTMLElementTagNameMap type.
+      const tagNameMapSymbol = checker.resolveName(
+        'HTMLElementTagNameMap',
+        undefined,
+        typescript.SymbolFlags.Interface,
+        false
+      );
+
+      if (tagNameMapSymbol?.declarations?.[0]) {
```
justinfagnani: This if check condition looks odd to me, can you just do:

```suggestion
      if (tagNameMapSymbol) {
```

PR #5052, opened 2025-08-21
Add func to get the effective type of a binding expression

rictic:
Introduce the `getLitExpressionType` function to determine the appropriate type for lit-html binding expressions, including sentinel symbols and directive results.

### DIFF
--- a/packages/labs/tsserver-plugin/src/lib/type-helpers/lit-expression-type.ts
+++ b/packages/labs/tsserver-plugin/src/lib/type-helpers/lit-expression-type.ts
@@ -1,0 +1,165 @@
+import type * as ts from 'typescript';
+
+/**
+ * Given a type of a binding expression, return the type that we should
+ * type check against.
+ *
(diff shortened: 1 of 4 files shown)


packages/labs/tsserver-plugin/src/lib/type-helpers/lit-expression-type.ts
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * This is usually the same as just returning the type, however, lit-html's
+ * rendering behavior does have a few special values.
+ *
+ * The `noChange` and `nothing` symbols are always allowed in any binding.
+ *
+ * DirectiveResults should be type checked as the return type of the `render`
+ * method on the directive class.
+ */
+export function getLitExpressionType(
```
justinfagnani, line 17: This should probably go in the analyzer. TODO

justinfagnani (approved):

PR #5053, opened 2025-08-21
[labs/tsserver-plugin] Fix noBindingLikeAttributeNames rule

kevinpschaaf:
We weren't actually checking that there wasn't a lit part.


packages/labs/tsserver-plugin/src/lib/rules/no-binding-like-attribute-names.ts
```diff
@@ -32,9 +33,10 @@ export const noBindingLikeAttributeNames = {
           element.attrs.forEach((attr) => {
             const {name} = attr;
             if (
-              name.startsWith('.') ||
-              name.startsWith('@') ||
-              name.startsWith('?')
+              (name.startsWith('.') ||
+                name.startsWith('@') ||
+                name.startsWith('?')) &&
+              (attr as LitTemplateAttribute).litPart === undefined
```
justinfagnani, line 39: fyi, the other check we could do is `!name.endsWith(boundAttributeSuffix)`

PR #5056, opened 2025-08-22
Test directives and unions a bit more, fix a few subtle bugs.

rictic:


packages/labs/tsserver-plugin/src/lib/diagnostic-codes.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Central definition of diagnostic codes.
+ */
+export const LitDiagnosticCode = {
+  BindingLikeAttributeName: 6301,
```
justinfagnani: Can we comment on why we chose these numbers?

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
```diff
@@ -0,0 +1,144 @@
...
+      const {tsNode} = litTemplate;
+
+      traverse(litTemplate, {
+        element(element: Element) {
+          for (const attr of element.attrs) {
+            if (!hasAttributePart(attr)) {
+              continue;
+            }
+            const part = attr.litPart;
+            if (part.type !== PartType.PROPERTY) {
```
justinfagnani, line 57: We could check element bindings here too - they just have to be a directive result.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
```diff
@@ -0,0 +1,144 @@
...
+            let rightHandSideType: ts.Type;
+            // If there are more than two strings, or if either of the strings
+            // are not empty string, then the type of the right hand side is
+            // 'string'.
+            if (part.strings.length > 2 || part.strings.some((s) => s !== '')) {
+              rightHandSideType = checker.getStringType();
+            } else {
+              if (part.expressions.length !== 1) {
+                throw new Error(
+                  'InternalError: Expected exactly one expression'
```
justinfagnani: This should be impossible, but want to leave any breadcrumbs for after-the-fact debugging? Like outputting the attribute test from the template?

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
```diff
@@ -0,0 +1,144 @@
...
+                typescript,
+                litLanguageService.getProgram()!
+              );
+            }
+            // get the type of the left hand side
+            const elementType = litLanguageService.getElementClassType(
+              element.tagName
+            );
+            if (elementType === undefined) {
+              // Unknown element. Other rules will warn in this case.
```
justinfagnani: Does the HTMLElementTagNameMap include built-in elements? Should we fall back to HTMLElement for unknown elements? ie, a `.id=${}` binding is fine actually, but if we don't know the element `.foo=${}` is bad.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
```diff
@@ -0,0 +1,144 @@
...
+
+            const propertySymbol = elementType.getProperty(part.name);
+            if (
+              propertySymbol == null ||
+              propertySymbol.valueDeclaration == null
+            ) {
+              const location = element.sourceCodeLocation?.attrs?.[attr.name];
+              const templateStart =
+                (typescript.isNoSubstitutionTemplateLiteral(tsNode.template)
+                  ? tsNode.template.getStart()
```
justinfagnani: I'm still not entirely sure the difference, but I've been trying to use `getFullStart()` after getting inconsistence results from other methods and properties.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
```diff
@@ -0,0 +1,144 @@
...
+              leftHandSide
+            );
+
+            if (isAssignable) {
+              continue;
+            }
+
+            const {name} = attr;
+            const location = element.sourceCodeLocation?.attrs?.[name];
+            const templateStart =
```
justinfagnani, line 134: This is repeated code... maybe we need helpers.

justinfagnani (approved):
🎉

justinfagnani:
I think we should merge this into main when it's prereqs are in, instead of into `tools-hackathon`

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
rictic: It does include built-ins, but yeah, good point, we should fall back to HTMLUnknownElement here.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
rictic: Yeah, and we need squiggle testing, but I'm punting that for now.

packages/labs/tsserver-plugin/src/lib/diagnostic-codes.ts
rictic: Done.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
rictic: Sure. Done.

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
rictic: Done

packages/labs/tsserver-plugin/src/lib/rules/no-unassignable-property-bindings.ts
rictic: Done, but as a separate rule, that's different enough that it doesn't feel like this rule.

justinfagnani (approved):

PR #5057, opened 2025-08-22
[labs/analyzer] Adjust attribute source locations in template parser

justinfagnani:
This is the same as https://github.com/lit/lit/pull/5054, just based on `main`.


justinfagnani:
Merging this since the other was already approved.

PR #5058, opened 2025-08-22
[localize] Fix "&", "<" and ">" getting replaced with html escape sequences.

### LINKED ISSUE #5012: [@lit/localize] A translation of "2 > 1" results in "2 &gt; 1" when using `msg("2 > 1")`

Localize (@lit/localize)

### Description

When adding translations containing the `>` char (didn't test any others), the char is HTML escaped when rendering in that translation (not when rendering the source language).

### Reproduction

When defining a translation like this:

``` js
return html`<span>${msg("2 > 1")}</span>`;
```

And add a translation:
``` xml
<trans-unit id="s59704991b8b78e98">
  <source>2 &gt; 1</source>
  <target>2 &gt; 1</target>
</trans-unit>
```

The rendered text is `2 &gt; 1` instead of `2 > 1`.

### Workaround

Did not test any workaround but I would guess that using 
``` JS
msg(html`2 &gt; 1`)
```
would work.

### Is this a regression?

### Affected versions

Tested with @lit/localize 0.12.2 and @lit/localize-tools 0.8.0

Browser: Firefox 139.0.1
OS: Windows 11 (24H2)

IIIMADDINIII:
Fixes #5012

"&", "<" and ">" do not need to be escaped in Template Literal Strings.
This causes invalid translations when these symbols are used in non HTML Context.
I think it is the code author/translator responsibility to properly escape special characters depending on the context.
lit/localize can not know in which context the strings are used.

### DIFF
--- a/packages/localize-tools/src/program-analysis.ts
+++ b/packages/localize-tools/src/program-analysis.ts
@@ -517,6 +517,8 @@ function replaceHtmlWithPlaceholders(
 
   const traverse = (node: ChildNode): void => {
     if (node.nodeName === '#text') {
+      // Copy source range to preserve escape sequences from the source text intact.
+      // parse5 does not preserve escape sequences, so node.value has html like escape sequences removed.
       const location = node.sourceCodeLocation!;
       const text = html.substring(location.startOffset, location.endOffset);
       components.push(text);
--- a/packages/localize-tools/src/program-analysis.ts
+++ b/packages/localize-tools/src/program-analysis.ts
@@ -6,7 +6,7 @@
 
 import ts from 'typescript';
 import * as parse5 from 'parse5';
-import {ChildNode, ParentNode, TextNode, CommentNode} from '@parse5/tools';
+import {ChildNode, ParentNode, CommentNode} from '@parse5/tools';
 import {ProgramMessage, Placeholder} from './messages.js';
 import {createDiagnostic} from './typescript.js';
 import {
--- a/packages/localize-tools/src/program-analysis.ts
+++ b/packages/localize-tools/src/program-analysis.ts
@@ -517,7 +517,8 @@ function replaceHtmlWithPlaceholders(
 
   const traverse = (node: ChildNode): void => {
     if (node.nodeName === '#text') {
-      const text = (node as TextNode).value;
+      const location = node.sourceCodeLocation!;
+      const text = html.substring(location.startOffset, location.endOffset);
       components.push(text);
     } else if (node.nodeName === '#comment') {
       components.push({
@@ -535,7 +536,7 @@ function replaceHtmlWithPlaceholders(
     }
   };
 
-  const frag = parse5.parseFragment(html);
+  const frag = parse5.parseFragment(html, {sourceCodeLocationInfo: true});
   for (const child of frag.childNodes) {
     traverse(child);
   }
--- a/packages/localize-tools/src/tests/transform.unit.test.ts
+++ b/packages/localize-tools/src/tests/transform.unit.test.ts
@@ -124,6 +124,26 @@ test('msg(string) translated', async () => {
   });
 });
 
+test('msg(string) with symbols', async () => {
+  // "\" in code needs to be double escaped (string inside a string).
+  await checkTransform(
+    'msg("Hello World \\\\ ${} ` < > &", {id: "foo"});',
+    '"Hello World \\\\ ${} ` < > &";'
+  );
+});
+
+test('msg(string) translated with symbols', async () => {
+  // "\" in code needs to be double escaped (string inside a string).
+  // "$", "`" only needs to be escaped in template literal string
+  await checkTransform(
+    'msg("Hello World \\\\ ${} ` < > &", {id: "foo"});',
+    '`Hola Mundo \\\\ \\${} \\` < > &`;',
+    {
+      messages: [{name: 'foo', contents: ['Hola Mundo \\ ${} ` < > &']}],
+    }
+  );
+});
+
 test('html(msg(string))', async () => {
   await checkTransform(
     'html`<b>${msg("Hello World", {id: "foo"})}</b>`;',
@@ -136,6 +156,27 @@ test('html(msg(string)) translated', async () => {
     'html`<b>${msg("Hello World", {id: "foo"})}</b>`;',
     'html`<b>Hola Mundo</b>`;',
     {messages: [{name: 'foo', contents: ['Hola Mundo']}]}
+  );
+});
+
+// Fix: "<", ">", "&" need to be escaped in html literal
+test('html(msg(string)) with symbols', async () => {
+  // "\" in code needs to be double escaped (string inside a string).
+  // "$", "`" only needs to be escaped in template literal string
+  await checkTransform(
+    'html`<b>${msg("Hello World \\\\ ${} ` < > &", {id: "foo"})}</b>`;',
+    'html`<b>Hello World \\\\ \\${} \\` < > &</b>`;' // should be: 'html`<b>Hello World \\\\ \\${} \\` &lt; &gt; &amp;</b>`;'
+  );
+});
+
+// Fix: "<", ">", "&" need to be escaped in html literal
+test('html(msg(string)) translated with symbols', async () => {
+  // "\" in code needs to be double escaped (string inside a string).
+  // "$", "`" only needs to be escaped in template literal string
+  await checkTransform(
+    'html`<b>${msg("Hello World \\\\ ${} ` < > &", {id: "foo"})}</b>`;',
+    'html`<b>Hola Mundo \\\\ \\${} \\` < > &</b>`;', // should be: 'html`<b>Hola Mundo \\\\ \\${} \\` &lt; &gt; &amp;</b>`;'
+    {messages: [{name: 'foo', contents: ['Hola Mundo \\ ${} ` < > &']}]}
   );
 });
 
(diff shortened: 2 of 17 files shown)


justinfagnani:
@aomarks any thoughts?

@IIIMADDINIII is there some kind of test you can add so this won't regress if it's a good fix?

IIIMADDINIII:
After digging a little deeper i figured out the following:

@justinfagnani There is already a test for this, but the expected value included the escape sequences:
https://github.com/lit/lit/pull/5058/files#diff-f69680f4d4214898d3d6247b7ad6a30a08aec9f7ff0f83fb699c59668cfe07efL347-L351

There also already was a test case for keeping escape sequences in html translations:
https://github.com/lit/lit/blob/main/packages/localize-tools/testdata/build-runtime-xliff/input/foo.ts#L59-L60
The Previous logic worked as follows:
 1. Extract would parse html Templates with html5 -> Removing escape sequences only from html Templates
 2. Write the Translations to XLIFF file -> Adding escape sequences as needed for it to be valid xml (all Types)
 3. Build would read XLIFF file form disk -> Removing the previously added Escape Sequences (all Types)
 4. escapeTextContentToEmbedInTemplateLiteral -> Always adding escape Sequences for some Characters (all Types)

This Pull request is changing this, by not removing the escape sequences during extract (copy source string instead of using the parser result). So the escape sequences do not need to be added back by escapeTextContentToEmbedInTemplateLiteral which is flawed.

Advantages of this solution:
- Not removing escape sequences which are in the source
- Not adding escape sequences which are not in the source
- I think it is expected behavior that translations do not mess with escape sequences
- All Types (`""`, `str`` `, `html`` `) have the same Behavior.

Disadvantages of this solution:
- HTML escape sequences in source are double escaped in XLIFF files (The & sign needs to be escaped in XLIFF)
- Changing XLIFF Source Translations for existing projects might be a breaking change

I would consider this a breaking change, because if some project has existing translations, these would need to be fixed to add the escape sequences which where previously removed. 
If a source string currently contains a `&lt;` and was already translated then the target is not double escaped.
So even if the sources are extracted again, the translations do not contain the double escapes and during build the escaping would be removed, creating invalid html.

I think this pull request implements the correct behavior, but if the braking change is not desired, It might be a good idea to instead only apply the escaping in escapeTextContentToEmbedInTemplateLiteral for html Templates.

packages/localize-tools/src/tests/transform.unit.test.ts
```diff
@@ -347,7 +347,7 @@ test('msg(string(msg(string))) translated', async () => {
 test('msg(string(<b>msg(string)</b>)) translated', async () => {
   await checkTransform(
     'msg(str`Hello <b>${msg("World", {id: "bar"})}</b>!`, {id: "foo"});',
-    '`Hola &lt;b&gt;Mundo&lt;/b&gt;!`;',
+    '`Hola <b>Mundo</b>!`;',
```
rictic, line 457: This test does look more correct to me. We should not be doing HTML escaping of expressions that only contain `msg()` calls, string literals, and `str` templates.

We do need to be careful about HTML that we emit into `html` templates, both because that is a security boundary for Lit (a maliciously crafted html template can execute arbitrary code. this is fine because html templates are themselves source code).

packages/localize-tools/testdata/build-runtime-xliff-ph/goldens/xliff/es-419.xlf
```diff
@@ -56,8 +56,8 @@
   <target>described 0</target>
 </trans-unit>
 <trans-unit id="h02c268d9b1fcb031">
-  <source>&lt;Hello<ph id="0">&lt;b></ph>&lt;World &amp; Friends><ph id="1">&lt;/b></ph>!></source>
-  <target>&lt;Hola<ph id="0">&lt;b></ph>&lt;Mundo &amp; Amigos><ph id="1">&lt;/b></ph>!></target>
+  <source>&amp;lt;Hello<ph id="0">&lt;b></ph>&amp;lt;World &amp;amp; Friends&amp;gt;<ph id="1">&lt;/b></ph>!&amp;gt;</source>
```
rictic, line 59: This looks less correct. Why is it double-escaped?

IIIMADDINIII: TL;DR
The original template for this translation is ``html`&lt;Hello<b>&lt;World &amp; Friends&gt;</b>!&gt;` `` which already includes escape sequences. To write this in an XML file, the `&` symbols need to be escaped. 

Long Version: 
I have the opinion, that localize should not change how I write my HTML in templates. Lets say I want to put the cent Symbol in a temlate (¢). But for some reason i need to escape it. I would write: ``html`&cent;` ``.

The old Version would convert all escape sequences back, so the `&cent;` would become a `¢` Symbol. This is then written to the translation file as a `¢` Symbol.
During build it would read the `¢` Symbol and output ``html`¢` `` to the source. So it is not the same as how I wrote the template.

To fix this, we need to preserve the original escape sequences. So instead off writing the `¢` symbol to the translation files we need to write `&cent;` to the translation file. Doing so will cause the XML serializer to escape the `&` Sign to make it valid XML. This results in `&amp;cent;` to be written to the translation file. If you open this with an XML viewer you will not See the double escape. 

The XML parser will convert the `&amp;cent;` back to `&cent;` while reading the file during build. So localize can See the original content and will emit `` html`&cent;` `` to the source. This is the behaivor i would expect.

justinfagnani: That explanation sounds reasonable to me

rictic: That makes sense to me. I'm surprised that no other tests needed changing here, just these .xlf files. Do we not use them in any other parts of the flow?

IIIMADDINIII: Because the behavior is mostly the same and most of the test do not include escape sequences, not a lot needed to be changed. If it helps, i could add some tests to check preserving escape sequences.

Regarding the Question: I do not know exactly. My understanding is, that the translation files are updated when extract is run. And when build is run it will do an extract first (without updating translations) to check for missing translations and then read the translations from the files.

IIIMADDINIII:
Is there anything i need to to to get this merged??
I am just asking, because i am not so familiar with open source development.

packages/localize-tools/src/program-analysis.ts
```diff
@@ -517,7 +517,8 @@ function replaceHtmlWithPlaceholders(
 
   const traverse = (node: ChildNode): void => {
     if (node.nodeName === '#text') {
-      const text = (node as TextNode).value;
+      const location = node.sourceCodeLocation!;
+      const text = html.substring(location.startOffset, location.endOffset);
```
justinfagnani, line 523: How is this different form `node.value`? Is it because it preserves the escape sequences? If so, a comment will help future editors.

packages/localize-tools/src/tests/transform.unit.test.ts
justinfagnani: Is this actually more correct? The source of `Hello <b>` is a `str` template, not an `html` template. So if this were rendered client-side, `Hello <b>` would be treated as untrusted and rendered as `Hola &lt;b&gt;`. This looks semantically incorrect and a potential security issue (even if a very edge case).

@IIIMADDINIII Is this changing because `str` templates are being passed to `escapeTextContentToEmbedInTemplateLiteral()`, or that they're being concatenated into containing `html` templates first without escaping?

@aomarks if you recall ay of this :)

justinfagnani (changes requested):
@IIIMADDINIII sorry for the long delay here, and thanks for this contribution!

The .xlf test file changes look good to me, but I'm concerned about the transform test. It looks to me like string content is being treated as HTML now, and that the previous escaping behavior was correct. Can you try to confirm that if possible?

packages/localize-tools/src/tests/transform.unit.test.ts
IIIMADDINIII: The reason i changed this test, is that there is no HTML involved here. There is no html template tag used so we can not assume we are working with HTML content. We attribute `<b>` to mean bold text in HTML but we do not know if it is used as HTML. It might be a custom placeholder or trying to hilight the letter b in an error log.
I believe that adding `msg()` to a string or html template should basically be a noop if no translation is provided (Not modifying the source content). This is was is tested here.

Every localized template goes through `escapeTextContentToEmbedInTemplateLiteral()`. This function escapes \, $ and ` because they have a special meaning in template strings. The &, <, > signs have no special meaning in template strings and should not need to be escaped. Because i fixed this, the result is now the way as i see as "correct".

The only security issue i can see, is if someone had a html template ``msg(html`&lt;script&gt;alert(1)&lt;/script&gt;`)`` and extracted and translated it with the current version.
If then they build the app after the this PR, the output would be ```html`<script>alert(1)</script>` ``` potentially running code not intended for execution. Because of this, this it should be considered a breaking change.
If this is not acceptable for an experimental package, I could implement a waring/error if this is detected in html templates during build.

packages/localize-tools/src/program-analysis.ts
IIIMADDINIII: Thank you. I have added a comment.

packages/localize-tools/src/tests/transform.unit.test.ts
justinfagnani: My apologies! I misread the expected result as being inside an `html` template, just from seeing the backticks I guess. Sorry for that. I agree this is the correct result :)

justinfagnani: I think the point @rictic is raising would be definitively answered by a test like "html(msg(string(<b>msg(string)</b>))) translated" - that is, this same test, but inside an `html` template. The goal there is to make sure that the unescaped `<b>` isn't inlined into the `html` template.

We don't see to have any tests that nest `html(msg(string))` yet, so I'm not sure if we even inline in those cases or not.

IIIMADDINIII: Now i understand what you mean. I just tested ```html`<b>${msg(`Hello World <test>`)}</b>`;``` which gets transformed in to ```html`<b>Hello World <test></b>``` which is definitely a problem.
But this is also the case before this PR as well. So this PR does not change this behavior.

In my opinion, the flattening/merging of the template literals is problematic. If we would remove this flattening, it would always produce the correct output. Else we need to implement proper sanitization which is normally done by the browser via a textNode. If you like, i can look in to this in a different issue/PR?

justinfagnani: Thanks for checking. If you have that test, would you mind adding it even if it's confirming the incorrect behavior for now? Then we'll add an issue. If you have time to look into this, that would be great!

> In my opinion, the flattening/merging of the template literals is problematic.

The flattening is done because bindings do have a cost in lit-html, even if the values are constant. Flattening removes the overhead of localization completely. Luckily, proper escaping of untrusted strings is not complicated.

IIIMADDINIII: I have added the tests. Hope this is now ready to be merged. Thank you verry much!

Monica-Nguyen:
@rictic Wondering if this can possibly be reviewed again? Our team is encountering the issue and the fix would ideally come from the source

Monica-Nguyen:
@rictic @kevinpschaaf @justinfagnani @IIIMADDINIII Hi! I wanted to do a gentle bump on this issue as our team is actively looking at shipping a patch for this, but it would have to be at each point of consumption and ideally it would come from the source. Was wondering if we can look again at whether there are any outstanding issues with the current approach and getting this one through?

Monica-Nguyen:
Will also tag in other Lit members for help on this matter: @augustjk @bicknellr @dfreedm @smalls @sorvell JIC the others are OOO or unavailable!

PR #5059, opened 2025-08-22
[labs/vscode-extension] Load the tsserver-plugin

justinfagnani:


packages/labs/tsserver-plugin/src/index.cts
```diff
@@ -9,6 +9,9 @@ const init: ts.server.PluginModuleFactory & {loadedPromise: Promise<void>} = ({
     create(info: ts.server.PluginCreateInfo) {
       const {logger} = info.project.projectService;
 
+      // TODO (justinfagnani): Figure out where this logs to or remove.
```
rictic: Did you figure it out?

packages/labs/vscode-extension/src/lib/lit-extension.ts
```diff
@@ -23,6 +23,25 @@ export class LitExtension {
 
     const {context} = this;
 
+    // Begin: Try to force our tsserver-plugin to show diagnostics automatically
```
rictic: This is now more about configuring our plugin, since you figured out the race condition that was making the other bit fail, yeah?

justinfagnani: Yeah, I don't think it does much for us at the moment, but I'm going to keep it because I think this is where we plug in config. I'll change the comment.

packages/labs/tsserver-plugin/src/index.cts
justinfagnani: removed

PR #5061, opened 2025-08-22
[labs/tsserver-plugin] Get definition at position via html element tag name map

justinfagnani:
https://github.com/lit/lit/pull/5050 but against `main`


justinfagnani:
Merging, since #5050 was already approved.

Issue #2282, opened 2021-11-04
[lit-html] Type error occurs when rendering on a SVGElement container


naoak:
# Description

Rendering on SVGElement containers are not allowed from lit@2.0.

# Steps to Reproduce

1. Write this code

```html
<!DOCTYPE html>
<head>
  <style>
    svg {
      width: 100px;
      height: 100px;
      background: #ddd;
    }
    circle {
      fill: green;
    }
  </style>
  <script type="module" src="./render-on-svg.js"></script>
</head>
<body>
  <svg></svg>
</body>
```

```ts
// render-on-svg.ts
import {html, css, svg, LitElement, TemplateResult, render} from 'lit';
import {customElement, property} from 'lit/decorators.js';

const svgEl = document.querySelector('svg');

render(svg`<circle cx="50" cy="50" r="36"></circle>`, svgEl);
```

2. See this output...

## Live Reproduction Link

[...]

PR #5064, opened 2025-08-24
[lit-html] Include SVGElement in render root union type

brysonbw:
## Changes
- Create exportable union type `RenderRootNode` to replace `HTMLElement | DocumentFragment` for reusability
- Add type `SVGElement` to render root element union type 
  - This resolves type error when passing an `SVGElement` as the `container` argument to `render(...)`, allowing SVG elements to be used as valid render roots
  
## Linked Issues Being Addressed
- #2282

### DIFF
--- a/.changeset/sour-heads-study.md
+++ b/.changeset/sour-heads-study.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Add type SVGElement to render root union type. Change resolves type error when passing an SVGElement as the container argument to render(...), allowing SVG elements to be used as valid render roots.
--- a/.changeset/sour-heads-study.md
+++ b/.changeset/sour-heads-study.md
@@ -2,4 +2,4 @@
 'lit-html': patch
 ---
 
-Add type SVGElement to render root union type. Change resolves type error when passing an SVGElement as the container argument to render(...), allowing SVG elements to be used as valid render roots.
+Add `SVGElement` to the `render()` container type, allowing SVG elements to be used as render roots without type errors.
--- a/packages/lit-html/src/lit-html.ts
+++ b/packages/lit-html/src/lit-html.ts
@@ -55,7 +55,7 @@ export namespace LitUnstable {
       kind: 'begin render';
       id: number;
       value: unknown;
-      container: HTMLElement | DocumentFragment;
+      container: RenderRootNode;
       options: RenderOptions | undefined;
       part: ChildPart | undefined;
     }
@@ -63,7 +63,7 @@ export namespace LitUnstable {
       kind: 'end render';
       id: number;
       value: unknown;
-      container: HTMLElement | DocumentFragment;
+      container: RenderRootNode;
       options: RenderOptions | undefined;
       part: ChildPart;
     }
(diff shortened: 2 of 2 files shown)


brysonbw:
In order to import code from other packages we have to add package to package’s `package.json` dependencies? I'm asking because I'm importing `RenderRootNode` type into `reactive-element.ts`. I’ve seen this approach used in other packages, but my context here is still new, so I might be missing a preferred alternative.

packages/lit-html/src/lit-html.ts
```diff
@@ -721,6 +721,10 @@ export interface RenderOptions {
    */
   isConnected?: boolean;
 }
+/**
+ * Element type that acts as a render root for DOM rendering.
+ */
+export type RenderRootElement = HTMLElement | SVGElement | DocumentFragment;
```
justinfagnani: I think this should be called `RenderRootNode` since it includes `DocumentFragment` (and therefore `ShadowRoot`).

brysonbw: Ok, smooth

brysonbw:
> In order to import code from other packages we have to add package to package’s `package.json` dependencies? I'm asking because I'm importing `RenderRootNode` type into `reactive-element.ts`. I’ve seen this approach used in other packages, but my context here is still new, so I might be missing a preferred alternative.

**Note To Self**: Try to build packages locally before pushing with `npm run test:dev`. Also, I think when adding and/or referencing a `lit` package within monorepo you have to (1) add it as a dependency in `package.json` and (2) include it's build script in the Wireit dependencies list

packages/reactive-element/package.json
```diff
@@ -453,7 +454,8 @@
     "/node/"
   ],
   "dependencies": {
-    "@lit-labs/ssr-dom-shim": "^1.4.0"
+    "@lit-labs/ssr-dom-shim": "^1.4.0",
+    "lit-html": "^3.3.0"
```
justinfagnani: I don't think we should add a new dependency here, especially just for a type.

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1116,7 +1117,7 @@ export abstract class ReactiveElement
    * @return Returns a node into which to render.
    * @category rendering
    */
-  protected createRenderRoot(): HTMLElement | DocumentFragment {
+  protected createRenderRoot(): RenderRootNode {
```
justinfagnani: I don't think we need to add SVGElement to this return type. Since a custom element can only extend HTMLElement, I think it'd be very unusual to render directly to inside an existing SVG.

brysonbw: I see, ok smooth

.changeset/sour-heads-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Add type SVGElement to render root union type. Change resolves type error when passing an SVGElement as the container argument to render(...), allowing SVG elements to be used as valid render roots.
```
justinfagnani:
```suggestion
Add `SVGElement` to the `render()` container type, allowing SVG elements to be used as render roots without type errors.
```

justinfagnani (approved):

justinfagnani:
Thanks @brysonbw!!

PR #5065, opened 2025-08-24
fix(lit-html): handle mock document without createTreeWalker

### LINKED ISSUE #4855: bug(lit-html): document mock check is too loose

### Description

When Stencil Jest unit tests are importing a lit-authored library, the following error occurs:

> TypeError: r.createTreeWalker is not a function

https://github.com/lit/lit/blob/267d243aeec13a6f0e8184420919db4c519b8caf/packages/lit-html/src/lit-html.ts#L716

lit-html provides a small mock for createTreeWalker, but the mock is only provided if document is not defined:

https://github.com/lit/lit/blob/267d243aeec13a6f0e8184420919db4c519b8caf/packages/lit-html/src/lit-html.ts#L349-L356

In case of Stencil, [mock document is provided](https://github.com/ionic-team/stencil/blob/main/src/mock-doc/document.ts#L13), but the mock does not have createTreeWalker.

I understand that the mock dom is not provided by Lit, but Lit could do a better job of detecting whether document implements createTreeWalker:

```suggestion
-   NODE_MODE && global.document === undefined
+   NODE_MODE && typeof global.document?.createTreeWalker !== 'function'
```

Since this code is stripped from production browser bundle, it should not increase core bundle size.

### Reproduction

1. [Open lit.dev](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiIvLyBTaW11bGF0ZSBTdGVuY2lsIG1vY2sgZG9ja1xuZG9jdW1lbnQuY3JlYXRlVHJlZVdhbGtlciA9IHVuZGVmaW5lZDtcbmltcG9ydCgnbGl0LWh0bWwnKS50aGVuKGNvbnNvbGUubG9nKS5jYXRjaChjb25zb2xlLmVycm9yKTsifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXRcIjogXCJeMy4wLjBcIixcbiAgICBcIkBsaXQvcmVhY3RpdmUtZWxlbWVudFwiOiBcIl4yLjAuMFwiLFxuICAgIFwibGl0LWVsZW1lbnRcIjogXCJeNC4wLjBcIixcbiAgICBcImxpdC1odG1sXCI6IFwiXjMuMC4wXCJcbiAgfVxufSIsImhpZGRlbiI6dHJ1ZX1d)
2. See error in the console

### Workaround

Create a Jest setup file with this code:

```js
// jestSetup.js
document.createTreeWalker ??= () => ({});
```

and use that file in stencil.config.ts:
```js
  testing: {
    setupFilesAfterEnv: ["<rootDir>/path/to/jestSetup.js"],
  },
```

### Is this a regression?

### Affected versions

lit-html 3.2.1

Node.js 20
Stencil 4.20.0

jitrodriguez:
Fixes #4855

Update the `document` selection logic in `lit-html.ts` to also check whether  
`document.createTreeWalker` is a function:  

```ts
const d =
  NODE_MODE && typeof global.document?.createTreeWalker !== 'function'
    ? ({ createTreeWalker() { return {}; } } as unknown as Document)
    : document;
```

If the function is missing, lit-html falls back to its internal mock.  

---

Did not add a unit test here, since the fallback path is decided at module evaluation time (const d = ...), and in the existing test files lit-html is already imported at the top.
This makes it impossible to toggle document.createTreeWalker before the module loads without introducing brittle dynamic-import tricks or extracting a new helper function, which I wanted to avoid to keep this fix minimal.

### DIFF
--- a/.changeset/chubby-moose-itch.md
+++ b/.changeset/chubby-moose-itch.md
@@ -5,4 +5,4 @@
 '@lit/reactive-element': patch
 ---
 
-Handle mock document without createTreeWalker (#4855)
+In Node builds, handle the case where document exists, but document.createTreeWalker does't (#4855)


.changeset/chubby-moose-itch.md
```diff
@@ -0,0 +1,8 @@
+---
+'lit-html': patch
+'lit': patch
+'lit-element': patch
+'@lit/reactive-element': patch
+---
+
+Handle mock document without createTreeWalker (#4855)
```
justinfagnani:
```suggestion
In Node builds, handle the case where document exists, but document.createTreeWalker does't (#4855)
```

PR #5067, opened 2025-08-25
fix(@lit-labs/nextjs): update webpackModuleRulesTest regex for cross-…

### LINKED ISSUE #5046: @lit-labs/nextjs renders only on client side in Windows (works fine on macOS)

SSR (@lit-labs/ssr)

### Description

In our component library, we have enabled deep rendering using @lit-labs/nextjs. It works fine on macOS. However, when running the same application on Windows, no error is shown, but all components fall back to client-side rendering. This is blocking us from proceeding further. Please help

### Reproduction

1. Clone this repo https://github.com/ChellappanRajan/lit-nextjs-ssr-issue
2. Run `npm install`
3. Run `npm run dev`
4. Open the application

   * On **macOS** → SSR works as expected
   * On **Windows** → Components fall back to client-side rendering

**Expected Behavior**
SSR should work the same on Windows as it does on macOS.

**Actual Behavior**
On Windows, components do not SSR and always render on the client.

### Workaround

None

### Is this a regression?

### Affected versions

Not sure

Node Version:v22.13.1

ChellappanRajan:
Fix #5046

### DIFF
--- a/.changeset/petite-grapes-sin.md
+++ b/.changeset/petite-grapes-sin.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/nextjs': patch
+---
+
+update webpackModuleRulesTest regex for cross-platform (Windows & Linux) path matching


packages/labs/nextjs/src/index.ts
```diff
@@ -34,7 +34,7 @@ export = (
 
         const {
           addDeclarativeShadowDomPolyfill = true,
-          webpackModuleRulesTest = /\/pages\/.*\.(?:j|t)sx?$|\/app\/.*\.(?:j|t)sx?$/,
+          webpackModuleRulesTest = /[\\/]pages[\\/].*\.(?:j|t)sx?$|[\\/]app[\\/].*\.(?:j|t)sx?$/,
```
justinfagnani: The only thing I worry about is `\` being a valid file name character on Unixes. Can we construct the regex dynamically based on which operating system we're running on?

packages/labs/nextjs/src/index.ts
```diff
@@ -34,7 +34,7 @@ export = (
...
           webpackModuleRulesExclude = [/next\/dist\//, /node_modules/],
```
justinfagnani: I think we need to update the separators here too.

justinfagnani (changes requested):
Thanks @ChellappanRajan!

Just a couple of comments, and this needs a changeset. Run: `npm run changeset`, select this package, choose "patch" and enter a description.

packages/labs/nextjs/src/index.ts
ChellappanRajan: Do you think it would be better to use the Node.js path package (https://nodejs.org/api/path.html#pathsep such as path.sep) to construct and check paths dynamically depending on the operating system?

justinfagnani: Yeah, that would be a good way to do it, along with the RegExp constructor.

ChellappanRajan: Thank you for confirming. I am currently away from my laptop. I will apply the changes and update the PR once I am back.

ChellappanRajan: Thanks for the feedback. I have addressed the comments — please take another look

PR #5070, opened 2025-08-27
[labs/tsserver-plugin] Optimize tests by reusing the language service.

rictic:
Each test file still runs in its own process so this is just for tests in the same file.


packages/labs/tsserver-plugin/src/test/lib/lit-expression-type_test.ts
```diff
@@ -3,25 +3,23 @@ import path from 'node:path';
...
-    target: ts.ScriptTarget.ES2022,
-    module: ts.ModuleKind.NodeNext,
-    moduleResolution: ts.ModuleResolutionKind.NodeNext,
-    strict: true,
-    esModuleInterop: true,
-    skipLibCheck: true,
-  };
-  return ts.createProgram([fakeTypesPath], options);
+function getProgram(): ts.Program {
+  using cleanup = getReusableTestProjectService();
```
justinfagnani, line 9: 😄

PR #5071, opened 2025-08-27
[labs/analyzer] Convert uvu tests to node:test

justinfagnani:
This is a large change by line count, but it was all fairly mechanical. I verified that the same number of tests run and pass after the conversion. Before was 17 node:test, 437 uvu. After 454 node:test.


packages/labs/analyzer/src/test/server/utils.ts
```diff
@@ -250,9 +228,17 @@ export const setupAnalyzerForNodeTest = (lang: Language, pkg: string) => {
   };
 };
 
-export interface AnalyzerModuleTestContext extends AnalyzerTestContext {
-  module: Module;
-}
+export const setupAnalyzerForNodeTestWithModule = (
```
rictic: Now that the migration is complete, consider renaming to just `setupAnalyzerForTestWithModule`

rictic (approved):

packages/labs/analyzer/src/test/server/utils.ts
justinfagnani: Oops! I meant to do that :)

PR #5074, opened 2025-08-29
[labs/ssr-client] Add digest cache

jimsimon:
This PR adds a cache for previously computed digest results. This should improve throughput when rendering the same `TemplateResult` multiple times.

### DIFF
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -466,7 +466,7 @@ const digestCache = new WeakMap<TemplateStringsArray, string>();
 // We don't care about cryptographic suitability.
 export const digestForTemplateResult = (templateResult: TemplateResult) => {
   let digest = digestCache.get(templateResult.strings);
-  if (digest) {
+  if (digest !== undefined) {
     return digest;
   }
 
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -449,6 +449,10 @@ const createAttributeParts = (
 
 // Number of 32 bit elements to use to create template digests
 const digestSize = 2;
+
+// Digest cache
+const digestCache = new WeakMap<TemplateStringsArray, string>();
+
 // We need to specify a digest to use across rendering environments. This is a
 // simple digest build from a DJB2-ish hash modified from:
 // https://github.com/darkskyapp/string-hash/blob/master/index.js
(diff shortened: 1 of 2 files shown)


packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -461,6 +465,11 @@ const digestSize = 2;
 //  - Easily specifiable and implementable in multiple languages.
 // We don't care about cryptographic suitability.
 export const digestForTemplateResult = (templateResult: TemplateResult) => {
+  let digest = digestCache.get(templateResult.strings);
+  if (digest) {
```
justinfagnani:
```suggestion
  if (digest !== undefined) {
```

We're pretty consistent about not relying on to-boolean conversions for clarity, safety, and performance. The direct check is faster.

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -449,6 +449,10 @@ const createAttributeParts = (
 
 // Number of 32 bit elements to use to create template digests
 const digestSize = 2;
+
+// Digest cache
+const digestCache = new WeakMap<TemplateStringsArray, string>();
```
justinfagnani, line 454: Let's make sure that we're not exacerbating the static template memory leak with this...

I think we're not because once the static template cache frees a template, the forged TemplateStringsArray will be able to be freed and this cache will release the entry.

jimsimon: I believe you're correct based on my understanding of how WeakMaps work in JS

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -478,5 +488,9 @@ export const digestForTemplateResult = (templateResult: TemplateResult) => {
   // for `btoa` when they set up their VM context, we instead inject an import
   // for `Buffer` from Node's built-in `buffer` module in our Rollup config (see
   // note at the top of this file), and use that.
-  return NODE_MODE ? Buffer.from(str, 'binary').toString('base64') : btoa(str);
+  digest = NODE_MODE
```
justinfagnani, line 491: btw, we should start using btoa in Node soon...

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
jimsimon: > Once an object used as a key has been collected, its corresponding values in any WeakMap become candidates for garbage collection as well — as long as they aren't strongly referred to elsewhere.
-MDN

So as long as the template is only associated with WeakMaps (or I'm assuming other "weak" data structures), we should be good.

PR #5080, opened 2025-08-31
Ignore a /.envrc file for managing working environment.

rictic:
I thought about checking in my nix based working dev environment setup but unless we used that as the basis for CI then it would get out of date, and getting a good github actions environemnt running locally so I can iterate on that turned into too much of a rathole, so for now let devs do what they like with a .envrc file. The /.direnv/ directory should always be git ignored, it just stores a versioned cache of dev environments.

Also ignore .actrc which is a user's local config for the act tool for running github actions locally. It usually has very setup-specific options, and should not be checked in.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,6 +14,9 @@ node_modules/
 /playground/p/**/package-lock.json
 /playground/p/scratch/*/
 !/playground/p/scratch/*.*
+/.direnv/
+/.envrc
+/.actrc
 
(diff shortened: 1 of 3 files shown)


justinfagnani:
Did you run ignore-sync?

rictic:
> Did you run ignore-sync?

I think it was run automatically on commit, it propagated changes to two other .ignore files.

PR #5097, opened 2025-09-16
[labs/ssr, labs/ssr-client] Add opt-in performance marking

jimsimon:
This PR adds performance markers for various pieces of the internal SSR implementation. We plan to use this at Reddit to better evaluate the real-world performance of parts of Lit SSR that we believe may benefit from further optimization.

### DIFF
--- a/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
+++ b/packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
@@ -275,7 +275,7 @@ const openChildPart = (
       throw new Error('compiled templates are not supported');
     }
     // Check for a template result digest
-    const markerWithDigest = `lit-part ${digestForTemplateResult(value)}`;
+    const markerWithDigest = `lit-part ${digestForTemplateResult(value).digest}`;
     if (marker.data === markerWithDigest) {
       const template = (
         ChildPart.prototype as ChildPart & {
@@ -467,7 +467,7 @@ const digestCache = new WeakMap<TemplateStringsArray, string>();
 export const digestForTemplateResult = (templateResult: TemplateResult) => {
   let digest = digestCache.get(templateResult.strings);
   if (digest !== undefined) {
-    return digest;
+    return {digest, cacheHit: true};
   }
 
   const hashes = new Uint32Array(digestSize).fill(5381);
@@ -492,5 +492,5 @@ export const digestForTemplateResult = (templateResult: TemplateResult) => {
     ? Buffer.from(str, 'binary').toString('base64')
     : btoa(str);
   digestCache.set(templateResult.strings, digest);
-  return digest;
+  return {digest, cacheHit: false};
 };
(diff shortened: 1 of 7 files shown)


packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -766,11 +784,33 @@ export function* renderValue(
...
-      yield `<!--lit-part ${digestForTemplateResult(
+      const {digest, cacheHit} = digestForTemplateResult(
         value as TemplateResult
-      )}-->`;
+      );
+      const shouldEmitPerformanceMetrics =
+        !cacheHit && options.emitPerformanceMetrics;
+      const startMark = mark(
+        shouldEmitPerformanceMetrics,
+        'ssr-digest-cold-creation-start'
```
justinfagnani, line 794: What's this measuring? The digest itself is calculated above, so this is only measuring the time it takes to assemble and yield a string.

Also, since this uses measure(), is there any benefit from also emitting marks? My understanding is that a measure is just a pair or related start/end marks anyway.

packages/labs/ssr-client/src/lib/hydrate-lit-html.ts
```diff
@@ -467,7 +467,7 @@ const digestCache = new WeakMap<TemplateStringsArray, string>();
 export const digestForTemplateResult = (templateResult: TemplateResult) => {
   let digest = digestCache.get(templateResult.strings);
   if (digest !== undefined) {
-    return digest;
+    return {digest, cacheHit: true};
```
justinfagnani, line 470: Rather than changing the return type, could we just measure digest creation directly here?

packages/labs/ssr/src/lib/render.ts
```diff
@@ -28,8 +37,12 @@ export type {RenderResult} from './render-result.js';
  */
 export function* render(
   value: unknown,
-  renderInfo?: Partial<RenderInfo>
+  renderInfo?: Partial<RenderInfo>,
+  renderOptions?: Partial<RenderOptions>
```
justinfagnani, line 41: Adding this parameter in some places makes this technically a breaking change (eg, element renderers need to forward the options to be correct). Can we add an option to RenderInfo? It's already a combined options and state object.

packages/labs/ssr/src/lib/render.ts
```diff
@@ -43,5 +56,14 @@ export function* render(
   if (isTemplateResult(value)) {
     hydratable = isHydratable(value);
   }
-  yield* renderValue(value, renderInfo as RenderInfo, hydratable);
+  yield* renderValue(value, renderInfo as RenderInfo, options, hydratable);
+
+  const endMark = mark(options.emitPerformanceMetrics, 'ssr-rsender-end');
+  measure(options.emitPerformanceMetrics, 'ssr-render-complete', {
```
justinfagnani, line 62: I don't think it's that much shorter to pass the option than to just do an if check right here. For simplicity, I'd lead towards an if and using the standard mark/measure APIs.

jimsimon:
Closing for now since the need for this has lessened. Will re-open if we decide to move forward with it later.

PR #5099, opened 2025-09-17
[labs/rollup-plugin-minify-html-literals] Replace unmaintained html-minifier with html-minifier-next

aaronmaxlevy:
This also fixes CVE-2022-37620 :)

### DIFF
--- a/.changeset/fifty-frogs-brake.md
+++ b/.changeset/fifty-frogs-brake.md
@@ -1,5 +1,5 @@
 ---
-"@lit-labs/rollup-plugin-minify-html-literals": patch
+'@lit-labs/rollup-plugin-minify-html-literals': patch
 ---
 
 Replacing unmaintained html-minifier with html-minifier-next
--- a/.changeset/fifty-frogs-brake.md
+++ b/.changeset/fifty-frogs-brake.md
@@ -1,0 +1,5 @@
+---
+"@lit-labs/rollup-plugin-minify-html-literals": patch
+---
+
+Replacing unmaintained html-minifier with html-minifier-next
--- a/package-lock.json
+++ b/package-lock.json
@@ -7714,16 +7714,16 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/@types/html-minifier": {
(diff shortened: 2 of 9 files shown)


justinfagnani:
We're going to have to take a look at this one, because I'm pretty wary of changing npm dependencies these days.

I'm personally not so worried about a ReDoS vulnerability being reported in html-minifier since the minifier is typically run on a developers own inputs. The minifier would have to be run on untrusted inputs for this to be a problem, which is conceivable (a hosted minifier, I suppose?) but probably very rare.

A few notes on `html-minifier-next`:
- It's very new - about 4 months
- It has few dependents on npm
- It has provenance set up
- It has benchmarks in the repo
- The dependencies are slightly different than html-minifier. I didn't look into them.

aaronmaxlevy:
@justinfagnani FWIW, the GitHub for the original project suggests moving to `html-minifier-next`, see https://github.com/kangax/html-minifier?tab=readme-ov-file#this-version-of-html-minifier-is-no-longer-maintained-for-an-up-to-date-version-which-includes-critical-security-fixes-please-use-html-minifier-next-by-j9t . That is what led to my specifically going down this path. If there is a different alternative you would prefer to switch to, I'd be happy to update my PR accordingly.

Re: the ReDoS vuln, I agree with you to an extent about the severity of the vulnerability, however the challenge for any company using Dependabot or other comparable vuln scanners is that CVE-2022-37620 will show as a high severity vuln, and that will cause a lot of headaches (e.g. as part of the process of being marked as a false positive, having to switch to a different library that doesn't bring in CVE-2022-37620, audit / compliance issues, etc.). 

Some companies even have a "proxy" that sits between internal systems and the NPM registry to block the download of any packages that include critical or high vulns.

I would also make the case that depending on an unmaintained package (e.g. the current `html-minifier` package) is itself inherently risky :)

justinfagnani (approved):

justinfagnani:
Thanks for the additional info @aaronmaxlevy!

aaronmaxlevy:
No problem! Thank you for your help with this :)

PR #5101, opened 2025-09-19
[labs/nextjs] skip patching when bundling for next edge runtime

mircoservices:
Added an early return to skip any webpack config modification when bundling for next's edge runtime. 

Addresses an issue we experienced when bundling in "node:22.11.0-alpine" where otherwise lit patches were applied to e.g. `.next/server/edge-instrumentation.js`.

This should not have any side-effects as the edge runtime is not used for any SSR processes.

### DIFF
--- a/packages/labs/nextjs/src/index.ts
+++ b/packages/labs/nextjs/src/index.ts
@@ -30,7 +30,17 @@ export = (
   (nextConfig: NextConfig = {}) => {
     return Object.assign({}, nextConfig, {
       webpack: (config, options) => {
-        const {isServer} = options;
+        const {isServer, nextRuntime} = options;
+
+        // Skip patching when bundling for next's edge runtime
+        if (nextRuntime === 'edge') {
+          // Apply user provided custom webpack config function if it exists.
+          if (typeof nextConfig.webpack === 'function') {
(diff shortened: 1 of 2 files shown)


packages/labs/nextjs/src/index.ts
```diff
@@ -30,7 +30,17 @@ export = (
   (nextConfig: NextConfig = {}) => {
     return Object.assign({}, nextConfig, {
       webpack: (config, options) => {
-        const {isServer} = options;
+        const {isServer, nextRuntime} = options;
+
+        // Skip patching when bundling for next's edge runtime
```
justinfagnani, line 35: Can you expand on this to say _why_ we should skip the patch in the edge runtime? Why should SSR not be run there?

mircoservices: We are using the edge runtime only for middleware and not in any of our pages. But saw that next's edge runtime instrumentation included the lit ssr patch and error'ed with: `ReferenceError: document is not defined`.

Might have been a bit naive to completely skip patching edge runtime, but it gave us the results we are looking for.

I can try working on a test demonstrating the behaviour that let to errors on our side.

PR #5103, opened 2025-09-22
[labs/ssr] Add performance benchmarking

jimsimon:
This PR adds a suite of performance benchmarks to Lit SSR help developers analyze the perf impact of their changes. Currently these need to be run manually on the developer's local machine before and after their changes.

The benchmarks try to simulate multiple different sizes and complexities of Lit templates representing something vaguely similar to Reddit's comments tree. This is easily Reddit's most complex and performance sensitive rendering scenario due to the heavy amounts of recursive nesting.

I wasn't able to use Tachometer like the existing benchmarks do, because we want these to run in Node and not in the browser. I originally wanted to use k6, but it doesn't support loading packages from node_modules. I ended up having AI implement a sort of custom test runner that simulates what k6 is trying to do. 

One thing that's still missing is the ability to have a scenario run multiple times concurrently. This is somewhat outside of the scope of what I currently need, so it's left as future work.

I'm happy to move this onto a proper benchmarking platform (e.g. Artillery, Gatling, etc) if there's a strong preference.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9011,22 +9011,6 @@
         "node": ">=10"
       }
     },
-    "node_modules/@wdio/config/node_modules/typescript": {
-      "version": "4.9.5",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.9.5.tgz",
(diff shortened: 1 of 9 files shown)


justinfagnani:
It would be nice if Tachometer could run Node benchmarks: https://github.com/google/tachometer/issues/274

PR #5106, opened 2025-09-25
[labs/ssr] Convert generators to a trampoline pattern

justinfagnani:
This gets rid of generators completely from the SSR code in favor of a thunk/trampoline approach where render functions can return thunks that are called to continue rendering. This keeps our ability to render synchronously or asynchronously, while improving perf.

Benchmarks from https://github.com/lit/lit/pull/5103 are merged into a separate branch https://github.com/lit/lit/tree/ssr-thunks-benchmarks and show an **18x** improvement compared to `main`.

This approach works by mapping a template's opcodes array into an array of strings or thunks and performing all render state access and mutation in those thunks. Consumers must iterate the array and when they get a thunk, they have to call it and process the return value. The return value can be undefined, a string, another render result, or a promise.

The thunk representation converts recursion and generators into a kind of continuation, and this appears to eliminate a *huge* amount of overhead.

Like the generator approach, synchronous consumers can throw when they encounter Promises, so we can implement `renderToString()` and continue to render synchronously inside of `React.createElement` patches, etc.

Asynchronous consumers can pause evaluation simply by not calling the next thunk yet. They must await the promises return by thunks, and can pause when they receive backpressure signals, etc. from downstream consumers.

The change is *mostly* backwards compatible. `render()` continues to return ` RenderResult`, which is still an iterable of strings or promises. A new function called `renderThunked()` is added that returns a ThunkedRenderResult`, which is an array of strings of thunks. `render()` calls `renderThunked()` and wraps the result in an iterable that uses a trampoline to read through the result for a small overhead.

`collectResult()`, `collectResultSync()`, and `RenderResultReadable` have been updated to accept both `RenderResult`s and `ThunkedRenderResult`s. Users who are already using those to consumer renders can just change their render call from `render()` to `renderThunked()` to get the best performance. All existing users will get a signification improvement either way though.

Further optimizations are likely possible. One idea is to reduce the number of closures produced by making a Renderer class that keeps the current render state on a stack and produces new values when calling next. Because of this, the `ThunkedRenderResult` type might not be very stable going forward. We may see a few major versions go by as we settle this down - this is labs after all!

One breaking change here is to the return type of the ElementRenderer methods. They now directly return a `ThunkedRenderResult`. This may be a good stable API to settle on, since if they return iterables of strings, we have to chain iterables (which we did with `yield*` previously) or change the return type to support iterables of iterables and make nested iterables responsible for kicking off execution of their render when `next()` is called - essentially a manual generator.

Custom ElementRenderers are more rare than direct SSR `render()` users, so it may be worth some churn here to land the drastic performance improvements ASAP, then adjust APIs in future major versions if needed.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "cSpell.words": ["renderable", "Thunked"]
+}
--- a/packages/labs/ssr/README.md
+++ b/packages/labs/ssr/README.md
@@ -268,7 +268,7 @@ has lower overhead compared to the `RenderResult` returned by `render()`.
 You should prefer `renderThunked()` over `render()`, which exists for backwards
 compatibility.
 
-In the future we may introduce APIs like `renderToStream()` and
+In the future we may introduce APIs like `renderToString()` and
 `renderToStream()` that eliminate the need to expose the underlying
 representation.
 
--- a/packages/labs/ssr/README.md
+++ b/packages/labs/ssr/README.md
@@ -36,22 +36,22 @@ including defining `customElements` on the global object.
 
 Web servers should prefer rendering to a stream, as they have a lower memory
 footprint and allow sending data in chunks as they are being processed. For this
-case use `RenderResultReadable`, which is a Node `Readable` stream
+case use `ThunkedRenderResultReadable`, which is a Node `Readable` stream
 implementation that provides values from `RenderResult`. This can be piped
 into a `Writable` stream, or passed to web server frameworks like [Koa](https://koajs.com/).
 
 ```js
 // Example: server.js:
 
-import {render} from '@lit-labs/ssr';
-import {RenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
+import {renderThunked} from '@lit-labs/ssr';
+import {ThunkedRenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
 import {myTemplate} from './my-template.js';
 
 //...
 
-const ssrResult = render(myTemplate(data));
+const ssrResult = renderThunked(myTemplate(data));
 // Assume `context` is a Koa.Context.
-context.body = new RenderResultReadable(ssrResult);
+context.body = new ThunkedRenderResultReadable(ssrResult);
 ```
 
 #### Rendering to a string
@@ -59,7 +59,7 @@ context.body = new RenderResultReadable(ssrResult);
 To render to a string, you can use the `collectResult` or `collectResultSync` helper functions.
 
 ```js
-import {render} from '@lit-labs/ssr';
+import {renderThunked} from '@lit-labs/ssr';
 import {
   collectResult,
   collectResultSync,
@@ -67,7 +67,7 @@ import {
 import {html} from 'lit';
 
 const myServerTemplate = (name) => html`<p>Hello ${name}</p>`;
-const ssrResult = render(myServerTemplate('SSR with Lit!'));
+const ssrResult = renderThunked(myServerTemplate('SSR with Lit!'));
 
 // Will throw if a Promise is encountered
 console.log(collectResultSync(ssrResult));
@@ -91,17 +91,17 @@ iterable that incrementally emits the serialized strings of the given template.
 ```js
 // Example: render-template.js
 
-import {render} from '@lit-labs/ssr';
-import {RenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
+import {renderThunked} from '@lit-labs/ssr';
 import {myTemplate} from './my-template.js';
 export const renderTemplate = (someData) => {
-  return render(myTemplate(someData));
+  return renderThunked(myTemplate(someData));
 };
 ```
 
 ```js
 // Example: server.js:
 
+import {ThunkedRenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
 import {renderModule} from '@lit-labs/ssr/lib/render-module.js';
 
 // Execute the above `renderTemplate` in a separate VM context with a minimal DOM shim
@@ -115,7 +115,7 @@ const ssrResult = await (renderModule(
 // ...
 
 // Assume `context` is a Koa.Context, or other API that accepts a Readable.
-context.body = new RenderResultReadable(ssrResult);
+context.body = new ThunkedRenderResultReadable(ssrResult);
 ```
 
 ## Client usage
@@ -202,8 +202,8 @@ Server-only templates can only be rendered on the server, they can't be rendered
 Here's an example that shows how to use a server-only template to render a full document, and then lazily hydrate both a custom element and a template:
 
 ```js
-import {render, html} from '@lit-labs/ssr';
-import {RenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
+import {renderThunked, html} from '@lit-labs/ssr';
+import {ThunkedRenderResultReadable} from '@lit-labs/ssr/lib/render-result-readable.js';
 import './app-shell.js';
 import {getContent} from './content-template.js';
 
@@ -211,7 +211,7 @@ const pageInfo = {
   /* ... */
 };
 
-const ssrResult = render(html`
+const ssrResult = renderThunked(html`
   <!DOCTYPE html>
   <html>
     <head><title>MyApp ${pageInfo.title}</head>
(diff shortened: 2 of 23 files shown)


justinfagnani:
Benchmark results using @jimsimon's PR: https://github.com/lit/lit/pull/5103

## `main` using generators:
Screenshot 2025-09-26 at 2 20 52 PM

## This change using `render()`
Screenshot 2025-09-26 at 2 23 32 PM

## This change using `renderThunked()`:
Screenshot 2025-09-26 at 2 40 42 PM

----

This is about an **18x** performance improvement for `renderThunked()` across the board 😱

`render()`, which is wrapping the thunked return type with a custom iterable for backwards compatibility, is about **14x** faster than `main`.

These improvements are so large that they deserve a lot of investigation to verify that they're real. Tests do pass and this is a non-breaking change now, so we should be able to publish and get real-world results.

----

The code to adapt the benchmarks to thunks is in https://github.com/lit/lit/tree/ssr-thunks-benchmarks

jimsimon:
I'll patch these changes into Reddit on Monday which should let us validate the performance improvement a bit. This is about 5 times faster than the "strings only" approach I currently have patched in, so we should see a shift in metrics.

justinfagnani:
I need to push a new changeset file that marks this as a breaking change. I forgot about the ElementRenderer change when I make the first changeset.

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -823,240 +832,293 @@ function* renderTemplateResult(
...
+          if (instance === undefined) {
+            throw new Error(
+              `Internal error: ${op.type} outside of custom element context`
+            );
+          }
+          // Perform any connect-time work via the renderer (e.g. reflecting any
+          // properties to attributes, for example)
+          if (instance.connectedCallback) {
+            instance.connectedCallback();
+          }
```
kyubisation: Nitpick, as you updated `getLast` and optional chaining has been supported about equally as long
```suggestion
          instance.connectedCallback?.();
```

packages/labs/ssr/src/lib/render.ts
```diff
@@ -43,5 +70,96 @@ export function* render(
...
+      return this.next();
+    }
+
+    // The value is a Promise. Convert to a Promise<RenderResult>:
+    this._waiting = true;
+    return {
+      done: false,
+      value: value.then((r) => {
+        if (typeof r === 'string') {
+          return r;
```
kyubisation: This might be a misunderstanding on my part, but should `this._waiting` not also be set to false in this early return?

packages/labs/ssr/src/test/lib/render-lit_test.ts
```diff
@@ -425,6 +425,7 @@ for (const global of [emptyVmGlobal, shimmedVmGlobal]) {
     );
   });
 
+  // eslint-disable-next-line no-only-tests/no-only-tests
```
kyubisation: Leftover?
```suggestion
```

packages/labs/ssr/src/test/lib/render-result-iterator_test.ts
```diff
@@ -0,0 +1,385 @@
...
+  const result1 = iterator.next();
+  assert.equal(result1.done, false);
+  assert.equal(result1.value, 'a');
+
+  const result2 = iterator.next();
+  assert.equal(result2.done, false);
+  assert.ok(result2.value instanceof Promise);
+
+  const resolvedValue = await result2.value;
+  assert.equal(resolvedValue, 'async-value');
```
kyubisation, line 157: This relates to my earlier question about `this._waiting`. For completeness, should you also assert `'b'` or add a test with `for await` with this scenario?

packages/labs/ssr/README.md
```diff
@@ -256,6 +256,22 @@ const ssrResult = render(html`
...
+
+The difference between the two is that `renderThunked()` returns a
+`ThunkedRenderResult`, which is a lower-level representation of a render that
+has lower overhead compared to the `RenderResult` returned by `render()`.
+
+You should prefer `renderThunked()` over `render()`, which exists for backwards
+compatibility.
+
+In the future we may introduce APIs like `renderToStream()` and
+`renderToStream()` that eliminate the need to expose the underlying
```
kyubisation, line 272: I am assuming `renderToStream()` should not be a repetition?

kyubisation (approved):
LGTM 👍 
Thank you very much for your work!
Looks very solid to me.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cSpell.words": ["renderable", "Thunked"]
```
rictic, line 2: neat!

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -823,240 +832,293 @@ function* renderTemplateResult(
...
+              part,
+              value,
+              partIndex
+            );
           }
-        }
-        partIndex += statics.length - 1;
+          let attributeResult: string | undefined = undefined;
+          // We don't emit anything on the server when value is `noChange` or
+          // `nothing`
```
rictic: This comment is out of date, we check for `nothing` inside the render*Part functions

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -823,240 +832,293 @@ function* renderTemplateResult(
...
+          // that may have been dynamically set by the renderer)
+          const result = instance.renderAttributes();
+          // If deferHydration flag is true or if this element is nested in
+          // another, add the `defer-hydration` attribute, so that it does not
+          // enable before the host element hydrates
+          if (
+            renderInfo.deferHydration ||
+            renderInfo.customElementHostStack.length > 0
+          ) {
+            result.push(' defer-hydration');
```
rictic: I don't think it's kosher to mutate the returned value of `instance.renderAttributes()`. No guarantee it will return a new array every time. Safer and almost as cheap to do:

```ts
result = [result, ' defer-hydration'];
```

packages/labs/ssr/src/lib/render-value.ts
```diff
@@ -823,240 +832,293 @@ function* renderTemplateResult(
...
+            shadowResult.push(
+              `<template shadowroot="${mode}" shadowrootmode="${mode}"${delegatesfocusAttr}>`
+            );
+            shadowResult.push(() => shadowContents);
+            shadowResult.push('</template>');
+            shadowResult.push(() => {
+              renderInfo.customElementHostStack.pop();
+            });
+          }
+          // renderInfo.customElementHostStack.pop();
```
rictic: Remove this comment

packages/labs/ssr/src/lib/render.ts
```diff
@@ -43,5 +70,96 @@ export function* render(
...
+    this._iterators = [result[Symbol.iterator]()];
+  }
+
+  next(): IteratorResult<string | Promise<RenderResult>, unknown> {
+    if (this._waiting) {
+      throw new Error(
+        'Cannot call next() while waiting for a Promise to resolve'
+      );
+    }
+    const iterator = this._iterators.at(-1);
```
rictic: Since most JS VMs don't have tail calls, can probably get a bit of speed out of putting most of this inside a `while (true)` so we can `continue;` instead of recursing.

packages/labs/ssr/src/lib/render.ts
```diff
@@ -43,5 +70,96 @@ export function* render(
...
+  return renderValue(value, renderInfo as RenderInfo, hydratable);
+}
+
+type InternalRenderResultIterator = Iterator<string | Thunk>;
+
+/**
+ * Wraps a ThunkedRenderResult to implement a RenderResult.
+ */
+export class RenderResultIterator
+  implements Iterator<string | Promise<RenderResult>>
```
rictic: IterableIterator?

packages/labs/ssr/src/lib/render.ts
justinfagnani: good catch!

packages/labs/ssr/src/test/lib/render-result-iterator_test.ts
justinfagnani: Asserting the `'b'` showed the bug. Thanks!

.vscode/settings.json
justinfagnani: Yeah, I need to do this for more words. It gets rid of a lot of blue marks in VS Code.

packages/labs/ssr/src/lib/render-value.ts
justinfagnani: It's just a question of where to perform the check then. Doing it here saves a function call...

justinfagnani: I just updated the comment, since each `render*Part` has slightly different `nothing` handling.

packages/labs/ssr/src/lib/render.ts
justinfagnani: True. It should be very minimal as calling `next()` will only recurse when entering or leaving a nested array or have a string of undefineds, and the stack unwinds when it returns. So you only get deep recursion if you have either a nested set of arrays that are all the first or last value of their parent. Or a string of undefined, which should be rare. Either way, I'll make the change :)

PR #5111, opened 2025-10-02
[labs/observers] Allow ResizeController to observe individual elements in Lit templates

### LINKED ISSUE #5110: [labs/observers] Add ability for ResizeController to observe elements in a template
### Should this be an RFC?

### Description

It would be very useful to be able to observe individual elements with ResizeController.

An example use case is an accordion-style widget where sections of it can be opened. If we need to observe the open sections (say to overlay an SVG), then we want to observe the section directly, not the host, and we want to be able to conditionally observe sections so that they're only observed when open.

I've forked ResizeController and used this API successfully:

```ts
ResizeController {
  observe(observe: boolean = true);
}
```

Example usage:
```ts
class MyAccordion extends LitElement {

  #resizeController = new ResizeController(this, {target: null});

  @property()
  accessor selectedItemId: string | undefined = undefined;

  @property({attribute: false})
  accessor items: Array<Item> | undefined = undefined;

  render() {
    return this.items?.map((i) => {
      const selected = i.id === this.selectedItemId;
      return html`
        <section
          class="${selected ? 'selected' : ''}"
          ${this.#resizeController.observe(selected)}
      ></section>
    `);
  }
}
```

You can't do this now with ResizeController, you have to use ResizeObserver directly.

justinfagnani:
Fixes #5110

This adds a `.target()` method to ResizeController that returns a directive that observes that element specifically.

### DIFF
--- a/packages/labs/observers/package.json
+++ b/packages/labs/observers/package.json
@@ -68,13 +68,14 @@
         "build:ts",
         "build:ts:types",
         "build:rollup",
+        "../../lit:build",
         "../../reactive-element:build"
       ]
     },
     "build:ts": {
       "command": "tsc --build --pretty",
       "dependencies": [
-        "../../lit-html:build:ts:types",
+        "../../lit:build:ts:types",
         "../../reactive-element:build:ts:types"
       ],
       "clean": "if-file-deleted",
@@ -169,6 +170,7 @@
     "@types/trusted-types": "^2.0.2"
   },
   "dependencies": {
+    "lit": "^3.2.0",
     "lit-html": "^3.2.0",
     "@lit/reactive-element": "^1.0.0 || ^2.0.0"
   },
--- a/.changeset/open-pets-tickle.md
+++ b/.changeset/open-pets-tickle.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/observers': minor
+---
+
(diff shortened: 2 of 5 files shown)


chen-ye:
you might want to consider updating the lit docs, which proposes a very similar architecture but with a different api, to mirror this: https://lit.dev/docs/composition/controllers/#controllers-that-own-directives

packages/labs/observers/package.json
```diff
@@ -169,6 +170,7 @@
     "@types/trusted-types": "^2.0.2"
   },
   "dependencies": {
+    "lit": "^3.2.0",
```
kyubisation: Why is lit needed here as a dependency? Looking at the changes, lit is only used in the test?

kyubisation (approved):
LGTM 👍

packages/labs/observers/package.json
justinfagnani: Good point. I'll move it to a dev dependency.

sorvell (approved):

justinfagnani:
@chen-ye 

> you might want to consider updating the lit docs, which proposes a very similar architecture but with a different api, to mirror this: https://lit.dev/docs/composition/controllers/#controllers-that-own-directives

This follows the same pattern mentioned in those docs. A method on the controller returns a directive call that passes the controller to the directive for coordination.

PR #5113, opened 2025-10-07
[lit-html] Allow the template cache to be customized when using the withStatic() tag generator

justinfagnani:
This adds an options object with a `getCache(strings)` option on it to `withStatic()`.

Authors can now customize the caching of static templates, including:
- Disabling caching completely (return `undefined`)
- Having separate caches per template callsite
- Use an LRU cache
- Use caches that collect hit rate metrics

This is done via `withStatic()` so that the caching strategy doesn't affect other static templates, and so that caching can be tailored to the specific template. A highly dynamic template could have caching disabled, for instance.

When combine with an `isServer` check, the cache can be customized per environment too.

### DIFF
--- a/packages/lit-html/src/static.ts
+++ b/packages/lit-html/src/static.ts
@@ -108,12 +108,44 @@ export const literal = (
 
 const stringsCache = new Map<string, TemplateStringsArray>();
 
+interface Cache {
+  get(s: string): TemplateStringsArray | undefined;
+  set(s: string, v: TemplateStringsArray): void;
+}
+interface WithStaticOptions {
+  /**
+   * Optional static template cache factory.
+   *
+   * Provide this option to customize caching, for example to provide a per
+   * template expression cache that doesn't grow without bound, use caches that
+   * collect metrics or to disable caching entirely by returning `undefined`.
+   *
+   * If this option is not provided, a global cache is used which grows without
+   * bound.
+   *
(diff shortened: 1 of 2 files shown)


packages/lit-html/src/static.ts
```diff
@@ -108,12 +108,44 @@ export const literal = (
...
+   * bound.
+   *
+   * If `undefined` is returned from `getCache`, no caching will be applied.
+   * This can be appropriate when the number of unique templates is known to be
+   * very large and the cache hit rate would be very low. This can happen in
+   * SSR scenarios where templates are generated dynamically based on user
+   * input.
+   */
+  getCache?: (strings: TemplateStringsArray) => Cache | undefined;
+}
```
jimsimon, line 133: If we end up merging https://github.com/lit/lit/pull/5118 as-is, then this description will probably need to be updated.

packages/lit-html/src/static.ts
```diff
@@ -108,12 +108,44 @@ export const literal = (
 
 const stringsCache = new Map<string, TemplateStringsArray>();
 
+interface Cache {
```
jimsimon, line 111: Any particular reason for creating a custom type and not just using the `Map` type?

jimsimon: Just realized that this is because WeakMap and Map don't have the same set of instance functions. This makes sense.

Issue #5114, opened 2025-10-09
Bug in `lit-element-hydrate-support.ts`, mutation of NamedNodeMap


zeroedin:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

When you SSR an element with element internals, the attributes are added to the DSD artifact:

```html
<my-element aria-label="foo" 
        hydrate-internals-aria-label="foo" 
        aria-roledescription="bar" 
        hydrate-internals-aria-roledescription="bar">
  ...
</my-element>
```

On hydration, these values should be removed from the DSD artifact, allowing the JavaScript definition to take over:

https://github.com/lit/lit/blob/5016b8f81091689282aa8a021c80e949f055e8ca/packages/labs/ssr-client/src/lit-element-hydrate-support.ts#L100-L110

However, the above code is reading from `this.attributes`, which will mutate in the loop, causing you to remove only half of the total attributes.

Instead, it should be setting a static array of the attributes first, then looping over the static array so that we don't mutate the Array we are reading from:

Something like:
```js
// Remove aria attributes added by internals shim during SSR
const attributes = [...this.attributes];
for (let i = 0; i < attributes.length; i++) {
  const attr = attributes[i];
  if (attr.name.startsWith(HYDRATE_INTERNALS_ATTR_PREFIX)) {
    const ariaAttr = attr.name.slice(
      HYDRATE_INTERNALS_ATTR_PREFIX.length
    );
    this.removeAttribute(ariaAttr);
    this.removeAttribute(attr.name);
  }
}
```

### Reproduction

Codepen shows only half the attributes being properly removed:

https://codepen.io/zeroedin/pen/PwZmMqZ

### Workaround

I have found a fix and will open a PR.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

Failing in @lit-labs/ssr-client@1.1.7

### Browser/OS/Node environment

Chrome: Version 141.0.7390.56 
MacOS: 26.0.1

PR #5115, opened 2025-10-09
[labs/ssr-client] fix loop mutation in hydrate support

zeroedin:
Fixes Array mutation problem in hydration-support

Addresses #5114


Lookwe69:
Are you sure `this.attributes` is an Iterable ? I thought it was a ArrayLike.
Is it performance related to not using for...of ?

packages/labs/ssr-client/src/lit-element-hydrate-support.ts
```diff
@@ -98,8 +98,9 @@ globalThis.litElementHydrateSupport = ({
     if (this._$needsHydration) {
       this._$needsHydration = false;
       // Remove aria attributes added by internals shim during SSR
-      for (let i = 0; i < this.attributes.length; i++) {
-        const attr = this.attributes[i];
+      const attributes = [...this.attributes];
+      for (let i = 0; i < attributes.length; i++) {
+        const attr = attributes[i];
```
justinfagnani: Since we only use the attribute name, this will be faster and also fix the bug if we use `getAttributeNames()` (which is not live):

```suggestion
      for (const attrName of this.getAttributeNames()) {
```

justinfagnani:
> Are you sure `this.attributes` is an Iterable ? I thought it was a ArrayLike. Is it performance related to not using for...of ?

NamedNodeMap is iterable, but it's also live (unlike an array iterator) so we can't use it directly here.

bennypowers (approved):

justinfagnani:
Thanks @zeroedin!! Sorry for the long review time!

PR #5118, opened 2025-10-10
[lit-html] Use a double-keyed LRU cache for server rendered static html

jimsimon:
This change switches static html to use a double keyed cache. The outer key is the pre-flattened template strings and the inner key is the final template string after all of the static parts have been merged into it (which was the previous behavior). It also implements the inner cache as a simple LRU with a relatively small max size. The idea here is that the ideal usage of static html has a small cardinality, but improper usage may have a high/infinite cardinality and we don't want to OOM the server (when using SSR) with an ever-growing cache.

### DIFF
--- a/.changeset/five-jobs-tell.md
+++ b/.changeset/five-jobs-tell.md
@@ -1,0 +1,5 @@
+---
+'lit-html': patch
+---
+
+Use a double-keyed LRU cache for server rendered static html


packages/lit-html/src/static.ts
```diff
@@ -106,7 +108,35 @@ export const literal = (
...
+  stringsCache = new Map<
+    TemplateStringsArray,
+    LRUCache<string, TemplateStringsArray>
+  >();
+} else {
+  stringsCache = new Map<
+    TemplateStringsArray,
+    Map<string, TemplateStringsArray>
+  >();
+}
```
kyubisation, line 139: Is there any difference here? Both paths initialize a Map and the type is given by `Cache`.
From my understanding this can simply be reduced to the following suggestion. Or am I missing something?
```suggestion
const stringsCache: Cache = new Map();
```

packages/lit-html/src/lru-cache.ts
```diff
@@ -0,0 +1,30 @@
...
+      throw new Error('maxSize must be greater than 0');
+    }
+    super();
+  }
+
+  override set(key: K, value: V): this {
+    super.delete(key);
+    super.set(key, value);
+    if (this.size > this.maxSize) {
+      const keyToDelete = this.keys().next().value;
```
kyubisation, line 13: During the community call there was a question about performance of accessing `this.keys().next().value`.
From a brief, naive test case this seems acceptable.

<details>
<summary>Performance test case</summary>

```js
const perfTest = (name, setup, test) => {
  console.log(`Starting performance test: ${name}`);
  const times = [];
  for (let i = 0; i < 100; i++) {
    const data = setup();
    const start = performance.now();
    test(data);
    times.push(performance.now() - start);
    if (i % 10 === 0) {
      console.log(`Completed ${i} runs`);
    }
  }
  const avgTime = times.reduce((a, b) => a + b, 0) / times.length;
  console.log(`Average time over 100 runs: ${avgTime.toFixed(6)} ms`);
}

perfTest(
  'Small map key access',
  () => new Map().set(1, 1).set(2, 2),
  (data) => data.keys().next().value
);
perfTest(
  'Large map key access',
  () => [...Array(100000).keys()].map(() => crypto.randomUUID()).reduce((m, k) => m.set(k, k), new Map()),
  (data) => data.keys().next().value
);
```
</summary>

packages/lit-html/src/static.ts
```diff
@@ -150,15 +180,25 @@ export const withStatic =
     }
 
     if (hasStatics) {
+      let innerCache = stringsCache.get(strings);
+      if (innerCache === undefined) {
+        if (isServer) {
+          innerCache = new LRUCache<string, TemplateStringsArray>(10);
+        } else {
+          innerCache = new Map<string, TemplateStringsArray>();
+        }
```
kyubisation, line 189: Code style nitpick; Feel free to ignore.
```suggestion
        innerCache = isServer
          ? new LRUCache<string, TemplateStringsArray>(10)
          : new Map<string, TemplateStringsArray>();
```

PR #5120, opened 2025-10-13
[reactive-element] Assume native constructible stylesheet support

justinfagnani:
DO NOT MERGE

Exploration of how much code we could remove if we assumed constructible stylesheet support, which is Baseline and available cross-browser since March 2023: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet

This change saves 650 bytes.

### DIFF
--- a/packages/reactive-element/src/css-tag.ts
+++ b/packages/reactive-element/src/css-tag.ts
@@ -4,20 +4,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const NODE_MODE = false;
-
-// Allows minifiers to rename references to globalThis
-const global = globalThis;
-
-/**
- * Whether the current browser supports `adoptedStyleSheets`.
- */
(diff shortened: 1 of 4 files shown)


packages/reactive-element/src/css-tag.ts
```diff
@@ -47,60 +31,36 @@ const cssTagCache = new WeakMap<TemplateStringsArray, CSSStyleSheet>();
...
         'CSSResult is not constructable. Use `unsafeCSS` or `css` instead.'
       );
     }
     this.cssText = cssText;
-    this._strings = strings;
   }
 
   // This is a getter so that it's lazy. In practice, this means stylesheets
   // are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet | undefined {
```
rictic, line 49: This can just be CSSStyleSheet, we can drop the `| undefined` here.

PR #5153, opened 2025-10-26
[labs/ssr] Experiment with `node:worker_threads` Worker

kyubisation:
This is a draft pull request to experiment with `node:worker_threads` Worker.

Findings:

- If a lot of workers are spawned, old workers might be killed: https://engineering.zalando.com/posts/2024/07/nodejs-tale-worker-threads.html
  Depending on use case it might make sense to use a single render Worker instance or multiple. Might require additional performance testing.
- It is not possible to transfer functions into workers, but objects work. Note: Some data might be lost, like the `raw` property of a `TemplateStringsArray`, as just the array is transferred.
- The current state of this PR would allow using a ready-made solution or just parts of it (e.g. use the provided Worker or bring your own, use the `registerRenderRequestHandler` or implement your own logic for handling messages). See `render-worker_test.ts`.

Questions:

- Should lit provide utility for using `node:worker_threads` Workers?


justinfagnani:
@kyubisation awesome stuff!

I think that we can make the shared cache work by making an interface for a sync or async cache:

```ts
interface LocalTemplateCache {
  kind: 'local';
  get(strings: TemplateStringsArray): Array<Op>;
  set(strings: TemplateStringsArray, ops: Array<Op>): void;
}

interface RemoteTemplateCache {
  kind: 'remote';
  get(strings: TemplateStringsArray): Promise<Array<Op>>;
  set(strings: TemplateStringsArray, ops: Array<Op>): Promise<void>;
}

type TemplateCache = LocalTemplateCache | RemoteTemplateCache;
```

Then we have to thread the sync-or-async nature through `getTemplateOpcodes()` and `renderTemplateResult()`. When using a RemoteTemplateCache `getTemplateOpcodes()` would have to return a function, pass that to `renderTemplateResult()` which returns a function into it it's ThunkedResult... something like that.

A tricky part is `CustomElementOpenOp`, which has a reference to an element constructor. I think a RemoteTemplateCache will want to just store the tag names on the remote end, then map to a constructor in the worker. So a RemoteTemplateCache needs two ends to work.

PR #5177, opened 2025-12-10
[labs/analyzer] Preserve declared union types

### LINKED ISSUE #5176: [labs/analyzer] Declared union types should be preserved

Analyzer (@lit-labs/analyzer)

### Description

This is crucial especially for generated Angular wrappers.

### Reproduction

Consider a property like

```ts
@property({ type: String, reflect: true })
public placement: 'top' | 'bottom' = 'top';
```

Currently, the analyzer would widen its type to `string`. Then, the Angular wrapper generator would emit

```ts
@Input()
set placement(v: string) {
  this._ngZone.runOutsideAngular(() => (this._el.placement = v));
}
```

which fails type checking:

```
error TS2322: Type 'string' is not assignable to type '"top" | "bottom"'.
```

### Workaround

Fixed by this three-line change, which we've successfully run in production for a year:
```diff
diff --git a/packages/labs/analyzer/src/lib/types.ts b/packages/labs/analyzer/src/lib/types.ts
index 803eb141..f6b7442f 100644
--- a/packages/labs/analyzer/src/lib/types.ts
+++ b/packages/labs/analyzer/src/lib/types.ts
@@ -84,8 +84,11 @@ export const getTypeForType = (
 ): Type => {
   const {typescript} = analyzer;
   const checker = analyzer.program.getTypeChecker();
-  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"'
-  type = checker.getBaseTypeOfLiteralType(type);
+  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"',
+  // but retain explicitly defined unions
+  if (!type.isUnion()) {
+    type = checker.getBaseTypeOfLiteralType(type);
+  }
   const text = checker.typeToString(type);
   const typeNode = checker.typeToTypeNode(
     type,
```

### Is this a regression?

### Affected versions

0.13.2

OS: Windows 11
Node version: v25.2.1
npm version: 11.6.2

ClaudioHoffmann:
Fixes #5176.

### DIFF
--- a/packages/labs/analyzer/src/lib/types.ts
+++ b/packages/labs/analyzer/src/lib/types.ts
@@ -84,8 +84,11 @@ export const getTypeForType = (
 ): Type => {
   const {typescript} = analyzer;
   const checker = analyzer.program.getTypeChecker();
-  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"'
-  type = checker.getBaseTypeOfLiteralType(type);
+  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"',
+  // but retain explicitly defined unions
+  if (!type.isUnion()) {
+    type = checker.getBaseTypeOfLiteralType(type);
+  }
   const text = checker.typeToString(type);
   const typeNode = checker.typeToTypeNode(
     type,
--- a/.changeset/honest-planes-send.md
+++ b/.changeset/honest-planes-send.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/analyzer': minor
+---
+
+Declared union types are now preserved instead of being widened to their base type. This fixes unexpected type errors in the property accessors created by the Angular wrapper generator.
--- a/packages/labs/analyzer/src/test/server/lit/properties_test.ts
+++ b/packages/labs/analyzer/src/test/server/lit/properties_test.ts
@@ -175,6 +175,22 @@ for (const lang of languages) {
       assert.equal(importedClass.module, 'external.js');
     });
 
+    test('property with string literal union', () => {
+      const property = element.reactiveProperties.get('stringLiteralUnion')!;
+      assert.equal(property.type?.text, '"hi" | "hello"');
+      assert.equal(property.type?.references.length, 0);
+      assert.equal(property.reflect, false);
+    });
+
+    test('reflected property with string literal union', () => {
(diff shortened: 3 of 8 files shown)


packages/labs/analyzer/src/lib/types.ts
```diff
@@ -84,8 +84,11 @@ export const getTypeForType = (
 ): Type => {
   const {typescript} = analyzer;
   const checker = analyzer.program.getTypeChecker();
-  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"'
-  type = checker.getBaseTypeOfLiteralType(type);
+  // Ensure we treat inferred `foo = 'hi'` as 'string' not '"hi"',
+  // but retain explicitly defined unions
+  if (!type.isUnion()) {
```
justinfagnani, line 89: Are there any other exceptions we should list, or should we explicitly only widen literal types? I'm a little surprised that `getBaseTypeOfLiteralType()` modifies union types, since they aren't literal types. Maybe it works recursively?

I'm also slightly confused that `type` is not already widened. I thought that mutable locations, like class fields, were already widened. I got that from the PR description: https://github.com/Microsoft/TypeScript/pull/10676

We are calling this via `getTypeForNode()` on the class field, not the initializer, so I would expect it to be widened already. Another maybe: maybe `getWidenedLiteralType()` is the right function to use here?

Either way, this is a great change and I want to merge it to get it out with the next release, but I wonder if we also want to go further.

justinfagnani (approved):
LGTM. Thanks @ClaudioHoffmann!

packages/labs/analyzer/src/lib/types.ts
ClaudioHoffmann: > I'm a little surprised that `getBaseTypeOfLiteralType()` modifies union types, since they aren't literal types. Maybe it works recursively?

It falls into the `type.flags & TypeFlags.Union` case and runs `getBaseTypeOfLiteralTypeUnion()`, which widens to `string` as you'd expect.

> Another maybe: maybe `getWidenedLiteralType()` is the right function to use here?

That function only widens types with the `ObjectFlags.RequiresWidening` bit. For most of our test cases, this wouldn't do anything and thus cause these tests to fail – `"hi"` stays `"hi"` and doesn't get widened to `string`, `0` stays `0` and doesn't get widened to `number`.

PR #5185, opened 2025-12-18
[labs/forms] Reset version to 0.0.0 before first release

justinfagnani:


.changeset/beige-hairs-teach.md
augustjk, line 1: Might be good to add a changeset for minor version “Initial release” instead an empty one?

PR #5186, opened 2025-12-18
chore(labs/cli-mcp-server): delete the MCP package

e111077:
deleting this package. It was never deployed but mostly because not quite sure MCP is the right approach. Anthropic has had some doubts about how MCP is exposed in the first place. see: https://www.anthropic.com/engineering/code-execution-with-mcp

LMK what you think. Personally I'd like to sit on MCP some more before making a publish, esp since we don't really offer much in our MCP server as of now


e111077:
waiting for CI

e111077:
i signed it

justinfagnani:
@e111077 that article seems to be arguing for a different way of calling MCP servers and tools, instead of replacing them or even changing the MCP protocol.

I'm fine with deleting this for now, but I think we could bring it back too.

justinfagnani:
The test failure is a flake in the motion package.

PR #5188, opened 2025-12-22
Add type inference for mapper callbacks in asyncReplace and asyncAppend directives

ArthurG94:
## Description

This PR enhances the type definition of the `asyncReplace` and `asyncAppend` directives to support proper TypeScript type inference when using the mapper function.

### Problem

Currently, when using `asyncReplace` or `asyncAppend` with a mapper function, TypeScript cannot infer the type of items yielded by the async iterable, resulting in the mapper's parameter being typed as `unknown`.

**Example:**

```typescript
class MyComponent extends LitElement {
    private myArr: {id: number; name: string}[] = [
        {id: 1, name: 'Alice'},
        {id: 2, name: 'Bob'},
        {id: 3, name: 'Charlie'},
    ];

    private async * generateItems() {
        for (const item of this.myArr) {
            yield item;
            await new Promise((resolve) => setTimeout(resolve, 100));
        }
    }

    public override render() {
        return html`
            <ul>
                ${asyncReplace(this.generateItems(), (item) => html`<li>${item.name}</li>`)}
            </ul>
        `;
    }
}
```

TypeScript throws an error: `'item' is of type 'unknown'` at `${item.name}`.

### Solution

This PR adds generic type parameters to both directives, allowing TypeScript to properly infer the type of values yielded by the async iterable and passed to the mapper function.

**Before:**
```typescript
(value: AsyncIterable<unknown>, _mapper?: Mapper<unknown> | undefined) => DirectiveResult<typeof AsyncReplaceDirective>
```

**After:**
```typescript
<V>(value: AsyncIterable<V>, _mapper?: Mapper<V> | undefined) => DirectiveResult<typeof AsyncReplaceDirective>
```

### DIFF
--- a/packages/lit-html/src/test/directives/types_are_inferrable_test.ts
+++ b/packages/lit-html/src/test/directives/types_are_inferrable_test.ts
@@ -141,6 +141,8 @@ if (false as boolean) {
 
     type VRendersAs = GetRenderAs<typeof v>;
     const vRendersAs = null! as VRendersAs;
+    // GetRenderAs returns the render method's return type.
+    // For asyncReplace, the render method always returns noChange, which is a Symbol.
     vRendersAs satisfies symbol;
     // @ts-expect-error
     vRendersAs satisfies string;
@@ -162,6 +164,8 @@ if (false as boolean) {
 
     type VRendersAs = GetRenderAs<typeof v>;
     const vRendersAs = null! as VRendersAs;
+    // GetRenderAs returns the render method's return type.
+    // For asyncAppend, the render method always returns noChange, which is a Symbol.
     vRendersAs satisfies symbol;
     // @ts-expect-error
     vRendersAs satisfies string;


justinfagnani:
Thanks for the PR, but we need better descriptions than "Enhanced..." _How_ was this enhanced?

Apologies if I'm wrong, but this sounds very much like an AI generated PR title, which means it may be an AI generated PR. While we don't have any policies regarding AI generated PRs yet, I think we'll likely land on something similar in spirit to this article: https://simonwillison.net/2025/Dec/18/code-proven-to-work/

So we want the burden for showing that the code works to be on the PR submitter. Write a good description that shows you know what it does, associate with an issue, try the code out yourself and report back that it does work, make sure there are tests etc.

I see this is a small type-only change, and it might be related to https://github.com/lit/lit/pull/5035 but it'd be good for reviewers for you to point that out and to add to the same type test that there.

ArthurG94:
Hi Justin,

Thanks for your feedback. You're right that the PR description was generated by AI, but I want to clarify that the actual code implementation was written by me. I apologize for the confusion.

This change is related to issue #5035. I’ve added tests following the same pattern as the existing ones.

I’ve also updated the PR title and description.

justinfagnani:
@ArthurG94 thanks for clarifying! Reviewing now...

packages/lit-html/src/test/directives/types_are_inferrable_test.ts
```diff
@@ -122,4 +124,46 @@ if (false as boolean) {
...
+      // @ts-expect-error - value should not be string
+      value satisfies string;
+      return `rendered-${value}`;
+    });
+
+    type VRendersAs = GetRenderAs<typeof v>;
+    const vRendersAs = null! as VRendersAs;
+    vRendersAs satisfies symbol;
+    // @ts-expect-error
+    vRendersAs satisfies string;
```
justinfagnani, line 148: I might be misunderstanding the tests at this point, but these seem backwards to me. The DirectiveResult `v` will render a `string`, yes? That would align with the other render-as tests.

justinfagnani:
The test failure is a flaky lit-labs/motion test.

packages/lit-html/src/test/directives/types_are_inferrable_test.ts
ArthurG94: The DirectiveResult returns the return type of the render method.
In the case of asyncReplace and asyncAppend, the render method always returns noChange which is a Symbol.

Honestly, I added this check because it's present in all the other tests, but it's really lines 134-140 that test the functionality

justinfagnani: Ah, thanks makes sense. Thanks. Could you add a comment in the test that says exactly that?

ArthurG94: Done

PR #5196, opened 2025-12-29
Publicize SignalWatcherApi so class extensions are happy in TS

Westbrook:
TS throws the following without this `export`...

> 'extends' clause of exported class 'ReveElement' has or is using private name 'SignalWatcherApi'.ts(4020)

image

### DIFF
--- a/packages/labs/signals/src/lib/signal-watcher.ts
+++ b/packages/labs/signals/src/lib/signal-watcher.ts
@@ -41,7 +41,7 @@ const effectWatcher = new Signal.subtle.Watcher(() => {
   });
 });
 
-interface SignalWatcherApi {
+export interface SignalWatcherApi {
   updateEffect(fn: () => void, options?: EffectOptions): () => void;
 }
 


justinfagnani:
Thanks @Westbrook!

It'd be really nice to have a test that would have caught this. I wonder why our current tests didn't?

Westbrook:
Do you have thoughts on a path to a test here? I _think_ it would have to be a test outside of the package that imported it across the project to not immediately be on the private side of this API at build time. Are there "centralized" tests in the repo that work that way anywhere? Happy to add a test, but somewhat unsure about this issue to begin with...

Westbrook:
We have to patch this locally to continue to leverage these tools, so if the team has any suggestions on triggering this sort of build error in repo I would love to hear them so I can get some tests added and this update landed. 🙇🏼‍♂️

pascalvos:
Screenshot 2026-01-14 at 11 06 45
i also found an issue now we dont support abstract classes anymore in new 0.2.0 i changed these 2 lines in the 0.2.0 version of the signal-watcher file in the node modules and now my code works again.

PR #5202, opened 2026-01-08
[labs/ssr-dom-shim] Add register css hook file to publishable files

kyubisation:
Currently the compilation result from `register-css-hook.ts` is not included in the package.json `files` list, which means these files will not be published.
This PR fixes this.


justinfagnani (approved):
Ooops! Thanks for the PR

PR #5205, opened 2026-01-09
Fix: minifying css inside @container queries

### LINKED ISSUE #5207: [@lit-labs/rollup-plugin-minify-html-literals] Does not minify @container queries correctly

@lit-labs/rollup-plugin-minify-html-literals

### Description

When we minify a template string that contains a css container query it does not minify correctly.

For example

Input
```js
    function someFunction() {
      return css\`
        :host {
          container-type: inline-size;
        }

        .some-class {
          padding: 16px;
        }
      
        @container (min-width: 768px) {
          .some-class {
            display: flex;
          }

          .another-class {
            font-size: 20px;
          }
        }
      \`;
    }
```

Actual Output
```js
    function someFunction() {
      return css`:host{container-type:inline-size}.some-class{padding:16px}.another-class{font-size:20px}`;
    }
```

Expected Output
```js
    function someFunction() {
      return css`:host{container-type:inline-size}.some-class{padding:16px}@container (min-width:768px){.some-class{display:flex}.another-class{font-size:20px}}`;
    }
```

### Reproduction

Inside `packages/labs/rollup-plugin-minify-html-literals/src/test/minify-html-literals.spec.ts` add  unit test

```ts
  const CONTAINER_QUERY_SOURCE = `
    function someFunction() {
      return css\`
        :host {
          container-type: inline-size;
        }

        .some-class {
          padding: 16px;
        }
      
        @container (min-width: 768px) {
          .some-class {
            display: flex;
          }

          .another-class {
            font-size: 20px;
          }
        }
      \`;
    }
  `;

  const CONTAINER_QUERY_SOURCE_MIN = `
    function someFunction() {
      return css\`:host{container-type:inline-size}.some-class{padding:16px}@container (min-width:768px){.some-class{display:flex}.another-class{font-size:20px}}\`;
    }
  `;

  test('should preserve CSS inside @container queries', async () => {
    const result = await minifyHTMLLiterals(CONTAINER_QUERY_SOURCE, {
      fileName: 'test.js',
    });
    assert.equal(typeof result, 'object');
    assert.equal(result!.code, CONTAINER_QUERY_SOURCE_MIN);
  });
```

### Workaround

### Is this a regression?

### Affected versions

0.1.1

node --version
v22.21.1

npm --version
10.9.4

cadamsdev:
Fixes #5207

### DIFF
--- a/.changeset/wise-dryers-brush.md
+++ b/.changeset/wise-dryers-brush.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/rollup-plugin-minify-html-literals': patch
+---
+
+Fix minifying css inside @container queries


packages/labs/rollup-plugin-minify-html-literals/src/test/minify-html-literals.spec.ts
```diff
@@ -133,7 +133,7 @@ suite('minify-html-literals', () => {
 
     function cssProperty(property) {
       const width = '20px';
-      return css\`.foo{font-size:1rem;width:\${width};color:\${property}}\`;
+      return css\`.foo{font-size:1rem;width:\${width}color:\${property}}\`;
```
cadamsdev, line 136: Regression in clean-css v5.3.3

I don't mind fixing this however...

[clean-css](https://github.com/clean-css/clean-css) is not maintained anymore. Can the Lit team fork clean-css and maintain it or how should we proceed?

bencbradshaw: I've also seen related issues with clean-css. Modern css features like`@container` and native nesting are broken after being minified. I have a half-completed branch with a replacement css minify strategy, using another library, that I could move forward with, that does fix all these issues and provides greater flexibility. I wasn't sure if there was interest for an update, but it seems like yes.

cadamsdev: @bencbradshaw Sounds good to me 👍

I think that would be a great update since [container queries](https://caniuse.com/css-container-queries) and [css nesting](https://caniuse.com/css-nesting) are fully supported on all major browsers. We should be able use them. 🙂

bencbradshaw: @cadamsdev see this PR I just put up, as an alternative
https://github.com/lit/lit/pull/5208

cadamsdev: @bencbradshaw Awesome thanks for doing that. I'll close this PR since I think switching to lightningcss is a better approach 😊

PR #5218, opened 2026-01-30
fix(lit-html): handle `undefined` arg on disconnect for ref directive 

### LINKED ISSUE #5217: Ref directive does not san-check its stored arg on disconnected/reconnected lifecycle

### Description

❓ is the ref directive supposed to accept an undefined for argument?
Looks like that yes based on the exported types (that are derived from [here](https://github.com/lit/lit/blob/f243134b226735320b61466cebdaf0c1e574bfa7/packages/lit-html/src/directives/ref.ts#L48)
However, both the disconnected & reconnected lifecyles are doing unsafe usage of the ref.

I'll be glad to bring a fix upstream for that 🙂
However, I see two possible fixes, and I'd like to see which one we want to go with:
Either we consider the issue to be the looseness of the argument, and restrict it from RefCallback|undefined to RefCallback
Or, we go the opposite way and we ensure the internal _ref property is used safely.****

### Reproduction

https://lit.dev/playground/#sample=examples%2Fhello-world&gist=0b002ac6df87f09ff2344481e719cce4
1. Open DevTools
2. Check the console
3. It throw a runtime error

### Workaround

Marshall value prior to using the directive, `<div ${arg ? ref(arg) : nothing}></div>`  

### Is this a regression?

### Affected versions

3.3.1

Browser: Chrome 144.0.7559.97
OS: macOS 26.2 (25C56)

louis-bompart:
When `ref` is called with `undefined` for arg and the element where the ref has been render in disconnect, the disconnect method tried to call `this._ref.value`, where `this._ref` being undefined, causing an error.

The `_updateRefValue` method in essence currently already sieve thru the various type of the `_ref` member. This PR handles the `undefined` case by returning early. 

Fixes #5217

### DIFF
--- a/packages/labs/vscode-extension/package.json
+++ b/packages/labs/vscode-extension/package.json
@@ -7,6 +7,12 @@
   "engines": {
     "vscode": "^1.103.0"
   },
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/lit/lit.git",
+    "directory": "packages/labs/vscode-extension"
+  },
   "categories": [
     "Programming Languages",
     "Snippets",


packages/lit-html/src/directives/ref.ts
```diff
@@ -51,7 +51,7 @@ class RefDirective extends AsyncDirective {
 
   override update(part: ElementPart, [ref]: Parameters<this['render']>) {
     const refChanged = ref !== this._ref;
-    if (refChanged && this._ref !== undefined) {
+    if (refChanged) {
```
louis-bompart, line 54: Maybe an over-optimization, but we would have been checking twice whether `this._ref` is undefined in this code path.
Given that `_updateRefValue` handles it after my change, I removed this part of the condition.

.changeset/sweet-garlics-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'lit-html': patch
+---
```
augustjk, line 4: to make sure the main lit package also gets a bump
```suggestion
---
'lit': patch
'lit-html': patch
---
```

louis-bompart: I did consider, but the changeset CLI does mention that the patch would've been propagated to the consumer of lit-html, such as lit.
But it ain't an issue either so 👍

Issue #5116, opened 2025-10-09
[context] poor performance with tree of providers and consumers using same context


sorvell:
### Which package(s) are affected?

Context (@lit/context)

### Description

If you have a situation like this:

```html
<a-provider id="container">
  <!-- n of these --> <a-subsribing-consumer></a-subsribing-consumer>
  <!-- m of these --> <a-provider></a-provider>
</a-provider>
```

The `container` will hear `m` `context-provider` events and for each, it will send `n` `context-request` events. This is  `m * n`, which can lead to poor performance, especially with heavy use of the same context, for example in a list.

The code [here](https://github.com/lit/lit/blob/main/packages/context/src/lib/controllers/context-provider.ts#L113-L158)  controls this behavior. It exists to account for the case when a provider in the middle of the tree between an existing subscription for a context should take over providing that context.

### Reproduction

https://lit.dev/playground/#gist=bee184ea9b654996e9fc4e5006921b27

### Workaround

Here is a hack fix for non-minified code:

```ts
(ContextProvider.prototype as any).attachListeners = function() {
  this.host.addEventListener('context-request', this.onContextRequest);
  let pending = false;
  this.host.addEventListener('context-provider', (e) => {
    if (!this.host.hasUpdated) {
      return
    }
    if (e.context !== this.context) {
      return;
    }
    const childProviderHost = e.contextTarget ?? e.composedPath()[0];
    if (childProviderHost === this.host) {
      return;
    }
    e.stopPropagation();
    if (pending) {
      return;
    }
    pending = true;
    this.host.updateComplete.then(() => {
      this.onProviderRequest(e);  
      pending = false;
    });
  });
} 
```

### Is this a regression?

No, always broken

### Affected versions

all

### Browser/OS/Node environment

All

PR #5226, opened 2026-02-13
Fixes #5221 and #5116 by re-dispatching only for relevant consumers

### LINKED ISSUE #5221: [context] Rendering a manually created context provider element in a consumer that is a light DOM child of a provider causes an infinite loop

Context (@lit/context)

### Description

Given a tree like the following:
```html
<provider>
    #shadow-root
    | <slot><consumer></slot>
    <consumer>
        #shadow-root
        | <provider>
        |     <!--Created with document.createElement("provider") in <consumer>'s render-->
        | </provider>
    </consumer>
</provider>
```
Consumer gets stuck in an infinite loop rerendering itself.

From my basic investigation, it appears to be caused by the `context-provider` event, but beyond that I don't know what causes it

### Reproduction

https://lit.dev/playground/#sample=examples%2Fhello-world&gist=14a287d908738aa5664a99a88c3dd561

### Workaround

Stopping propagation of the `context-provider` event via either `stopPropagation` or `stopImmediatePropagation`  on the consumer element causes the issue to disappear

### Is this a regression?

### Affected versions

lit@3.3.1, @lit/context@1.1.6

Browser: Librewolf 144.0-1
OS: Linux

sorvell:
When a provider receives a `context-provider` event, previously it would re-dispatch the `context-request` event to any subscribers. Now it only does so for providers added _between_ the existing provider and subscriber.

This can avoid a corner case infinite loop when rendering the subscriber can create a new provider. It is also signifcantly more efficient.

Two changes were made in the `context-provider` event handler `onProviderRequest`:
1. Ignore for consumer host's event path, this means the new provider is _within_ the consumer and cannot provide for it.
2. Ignore for new providers that are not ancestors of the consumer and the handling provider. This is determined by traversing up from the consumer.

### DIFF
--- a/packages/context/src/lib/controllers/context-provider.ts
+++ b/packages/context/src/lib/controllers/context-provider.ts
@@ -107,6 +107,10 @@ export class ContextProvider<
       return;
     }
     ev.stopPropagation();
+    // avoid reprocessing callback if it's already a subscriber
+    if (this.subscriptions.has(ev.callback)) {
+      return;
+    }
     this.addCallback(ev.callback, consumerHost, ev.subscribe);
   };
 
@@ -125,8 +129,7 @@ export class ContextProvider<
     }
     // Also, in case an element is a consumer AND a provider
     // of the same context it shouldn't provide to itself.
-    const eventPath = ev.composedPath();
-    const childProviderHost = ev.contextTarget ?? eventPath[0];
+    const childProviderHost = ev.contextTarget ?? ev.composedPath()[0];
     if (childProviderHost === this.host) {
       return;
     }
@@ -151,15 +154,12 @@ export class ContextProvider<
         continue;
       }
       seen.add(callback);
-      // if consumer is in the event path, then the new provider
-      // is inside it and shouldn't provide.
-      if (eventPath.includes(consumerHost)) {
-        continue;
-      }
-      // determine if consumer is "contained" within the new provider;
+      // Determine if consumer is "contained" within the new provider;
       // this is required for it to provide.
-      let needsContext = false,
-        current = consumerHost;
+      // Note, the walk here is significantly faster than sending the event and
+      // short-circuiting based on the path.
+      let needsContext = false;
+      let current = consumerHost;
       while (current) {
         needsContext = current === childProviderHost;
         if (needsContext || current === this.host) {


packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -125,7 +125,8 @@ export class ContextProvider<
     }
     // Also, in case an element is a consumer AND a provider
     // of the same context it shouldn't provide to itself.
-    const childProviderHost = ev.contextTarget ?? ev.composedPath()[0];
+    const eventPath = ev.composedPath();
```
justinfagnani: I think we need to be careful about unconditionally calling `ev.composedPath()`. There was a PR a while back that added ev.contextTarget just so we could avoid the cost of computing the composed path: https://github.com/lit/lit/pull/4734

If we just early return if `ev.contextTarget === this.host` I think we should be fine. But we also need comments that tell future editors about this. There's a spot in onContextRequest too. That should have been done in the previous PR, sorry!

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -150,9 +151,30 @@ export class ContextProvider<
...
-      );
+      // if consumer is in the event path, then the new provider
+      // is inside it and shouldn't provide.
+      if (eventPath.includes(consumerHost)) {
+        continue;
+      }
+      // determine if consumer is "contained" within the new provider;
+      // this is required for it to provide.
+      let needsContext = false,
+        current = consumerHost;
```
justinfagnani: Our style is to do a `let` per variable. I'm surprised this passed lint!


```suggestion
      let needsContext = false;
      let current = consumerHost;
```

packages/context/src/lib/controllers/context-provider.ts
```diff
@@ -150,9 +151,30 @@ export class ContextProvider<
...
+      while (current) {
```
justinfagnani, line 163: It's unfortunate that we can't just fire the event here, since this is re-implementing the event bubbling logic. Could we still fire it and handle it differently if this provider receives back that same event, which should be the case you're filtering out here?

justinfagnani (changes requested):
Awesome!

Just one important thing where we need to avoid computing `composedPath()` if possible.

justinfagnani:
You might want to mention this change over here: https://github.com/webcomponents-cg/community-protocols/issues/50#issuecomment-2673555681 where Westbrook is asking us to document any refinements we've made to the Context Protocol.

packages/context/src/lib/controllers/context-provider.ts
sorvell: I did some benchmarking and firing the event and short-circuting is signifcantly slower. I tried compareDocuemntPosition and contains with scope jumping and the current approach was best.

Also added a short-circuit in the request handler which helps slightly.

sorvell:
> You might want to mention this change over here: [webcomponents-cg/community-protocols#50 (comment)](https://github.com/webcomponents-cg/community-protocols/issues/50#issuecomment-2673555681)

see https://github.com/webcomponents-cg/community-protocols/issues/50#issuecomment-3906146106

PR #5240, opened 2026-02-20
[labs/ssr-dom-shim] Register CSSStyleSheet polyfill in global scope when using Node.js hook for importing CSS files

kyubisation:
Currently if the DOM shim is not installed (or not installed before the import context of any import to `lit` or `@lit/reactive-element`), the conversion of `CSSStyleSheet` is hard locked into resulting in undefined, which is skipped.
Due to this, during the registering of the CSS import hook, the `CSSStyleSheet` is added to the global scope.
Additionally, this changes the native CSS import feature detection to use a data url to avoid bundlers failing when inlining the CSS file or skipping it entirely.


thescientist13 (approved):
Thank you @kyubisation 🙌

packages/labs/ssr-dom-shim/src/register-css-hook.ts
```diff
@@ -6,9 +6,8 @@
 
 try {
   // Detect whether the environment supports importing CSS files.
-  const cssImportsSupported = await import('./detection.css', {
-    with: {type: 'css'},
-  })
+  const inlineCSS = 'data:text/css;base64,cHt0b3A6MDt9';
```
thescientist13, line 11: oh interesting approach here.  Is this due to the larger issue of the race condition, or just an isolated refactor around _detection.css_ in general?

kyubisation: Just an isolated refactor to not depend on an external CSS file. As it is also harder to polyfill, it's a clearer indication of native support.

kyubisation:
We discussed this during the community call:

- There was opposition to changing the client side code in [packages/reactive-element/src/css-tag.ts‎. The preference is in just adapting the Node.js code path.
- The situation that there can be an instance of `CSSStyleSheet` without it being in the global scope is seen as unfortunate. Some options were discussed:

1. The CSS hook for Node.js removes the import to the `CSSStyleSheet` DOM shim and warns/throws if `CSSStyleSheet` is not in the global scope (i.e. requires using the DOM shim or similar, at least for `CSSStyleSheet`)
2. The CSS hook registers `CSSStyleSheet` on hook registration in the global scope
3. Keep it as it is and just adapt css-tag.ts

@thescientist13 What is your use case/opinion? Personally I'm inclined to 1., but I'd like your input on this.

thescientist13:
@kyubisation 
Not sure if I am 100% following along but it sounds like it mostly a concern around `CSSStyleSheet` leaking into the global scope?

Regarding option 1:

> The CSS hook for Node.js removes the import to the CSSStyleSheet DOM shim and warns/throws if CSSStyleSheet is not in the global scope (i.e. requires using the DOM shim or similar, at least for CSSStyleSheet)

Are you saying that for [these example here](https://github.com/thescientist13/lit-ssr-css-modules/), I would just have manually add an `import` for the DOM shim script specifically, as opposed to Lit SSR just automatically loading it in a NodeJS environment?

Regarding option 3:

> Keep it as it is and just adapt css-tag.ts

What would adapting in this case look like?

----

Not sure if this is useful info at, but for my use case, I am (already) using Worker Threads to do SSR, mainly to isolate SSR code like this from the global scope, so not sure how that would factor into your decision making?

That said, I've been considering moving away from worker threads just because when using `--import` from the CLI to run Node, all these hooks get registered over and over for every worker thread, as opposed to shared, which i guess is by design.  (NodeJS thread for reference - https://github.com/nodejs/loaders/issues/201#issuecomment-2892712205)

That said, I do need something like Worker Threads for local development / live reloading to do module cache busting, since ESM has no module invalidation API like CJS did - https://github.com/nodejs/node/issues/49442

Maybe the VM approach Lit recommends would be my way to go?  Or I was thinking of using Cluster, but ideally I'm trying to find something (at least long term) that would be "web standard" / WinterTC compatible.  Maybe that's Shadow Realms?  🤔 

----

All that yapping to say, and apologies if I didn't actually answer / get your question, is that I'm pretty much not opposed to any particular option in general, though I wouldn't be opposed to #2 since I always love it when a library does the thing for me  😅   ((though I guess there is still the case of Lit wanting to future proof in the event Node actually supports `CSSStyleSheet` in the future?)

kyubisation:
@thescientist13 Thank you for your input. After some considerations I think I agree with you of making it "just work" with option 2. I'll implement a corresponding approach and update the PR. 👍

kyubisation:
@thescientist13 Also in #5153 I'm experimenting with Lit Worker support for Node.js. If you have the time to take a look at that and provide your input, it should help in shaping it for usability.

justinfagnani (approved):
🎉

PR #5246, opened 2026-02-27
update creation scope to use customElementRegistry, fixes #5245

### LINKED ISSUE #5245: [lit-html] Update support for scoped custom elements
### Should this be an RFC?

### Description

Lit's support for custom elements is outdated and should be updated to match the current spec. The updated version has shipped in Safari and is targeted for Chrome 146.

lit-html's renderOptions supports `creationScope`. The docs describe:

```
 * Node used for cloning the template (`importNode` will be called on this
 * node). This controls the `ownerDocument` of the rendered DOM, along with
 * any inherited context. Defaults to the global `document`.
```

This was based on `shadowRoot.importNode` which does not exist.

However, all elements has a `customElementRegistry` property which could be used here. Perhaps a non-breaking but additive way this could be supported is as follows:

```js
  options?.creationScope?.importNode(content, true) ?? 
  document.importNode(content, {customElementRegistry: options?.creaionScope?.customElementRegistry ?? window.customElements});
```

The `@lit-labs/scoped-registry-mixin` package should be updated based on whatever we do here.

N/A

sorvell:
Updates support for scoped custom elements registries. `RenderOptions.creationScope` is used to call `document.importNode` with the provided scope's `customElementRegistry`.

This is a non-breaking change. The original support for `creationScope.importNode` is preserved, but a deprecation warning is issued if `importNode` is implemented on `creationScope`.


packages/lit-html/src/lit-html.ts
```diff
@@ -1220,7 +1226,29 @@ class TemplateInstance implements Disconnectable {
...
+    // scoped elements spec that included `shadowRoot.importNode`.
+    // In dev mode, we issue a warning if `importNode` is used, since it's
+    // deprecated and will be removed in the next major version.
+    const {customElementRegistry} = options?.creationScope ?? {};
+    const fragment =
+      options?.creationScope?.importNode?.(content, true) ??
+      d.importNode(
+        content,
+        // the explicit undefined check ensures null can be supplied.
+        customElementRegistry === undefined ? true : {customElementRegistry}
```
justinfagnani: Can you just do:

```suggestion
        options?.creationScope ?? {}
```

At this point `options.creationScope` doesn't have `importNode` in it, or it's undefined. It or `{}` will be interpreted as truthy in older browsers and as an option without `selfOnly` in ones that support scoped registries.

Then you don't need the `{customElementRegistry} = options?.creationScope ?? {}` line above either.

Or maybe you could try:
```ts
const {creationScope = {}} = options ?? {};
```

And see if that's smaller.

packages/lit-html/src/lit-html.ts
```diff
@@ -1220,7 +1226,25 @@ class TemplateInstance implements Disconnectable {
...
+    // Moving forward, the `customElementRegistry` should be passed via
+    // `creationScope`, but for now we also support customization
+    // via `importNode`. This is for backwards compatibility with the older
+    // scoped elements spec that included `shadowRoot.importNode`.
+    // In dev mode, we issue a warning if `importNode` is used, since it's
+    // deprecated and will be removed in the next major version.
+    const {creationScope = {}} = options ?? {};
+    const fragment =
+      creationScope?.importNode?.(content, true) ??
+      d.importNode(content, creationScope ?? {});
```
justinfagnani: Because of the `creationScope = {}` above, creationScope is always defined now:

```suggestion
      d.importNode(content, creationScope);
```

Actually, you can either remove the `?? {}` here, _and_ the optional chaining on `creationScope?.importNode` above (change to `creationScope.importNode`), or you can keep those two and remove the `= {}` from `creationScope = {}` above. Then creation scope could be undefined and you only give it a fallback if we don't follow this new branch.

It also occurs to be that we could write this whole thing as:

```ts
    const {creationScope} = options ?? {};
    const fragment = (creationScope?.importNode ?? d.importNode)(content, creationScope ?? {});
```

The only maybe weird thing is that if `creationScope` has both `importNode` and `customElementRegistry` then we call our old-style `importNode` API, with a new style `customElementRegistry` arg. But it might be smaller after minification.

As far as nano-optimization goes, I _slightly_ suspect that letting `creationScope` be undefined and relying on optional chaining is a tiny bit better than looking up an undefined property on a fresh object.

sorvell: Removed the `= {}`, kept the old code path the same, and used `d.importNode(node, creationScope ?? true)`. 

* Note, with true or {} this was 3 bytes less and `true` is more obviously the older API.
* I think it's important to avoid changing the old code path at all to try to ensure BC. I, for one, often made bespoke use of creationScope.

packages/lit-html/src/lit-html.ts
```diff
@@ -1220,7 +1226,25 @@ class TemplateInstance implements Disconnectable {
...
+    // In dev mode, we issue a warning if `importNode` is used, since it's
+    // deprecated and will be removed in the next major version.
+    const {creationScope} = options ?? {};
+    const fragment =
+      creationScope?.importNode?.(content, true) ??
+      d.importNode(content, creationScope ?? true);
+    if (DEV_MODE && creationScope?.importNode !== undefined) {
+      issueWarning(
+        'import-node',
+        'RenderOptions `creationScope.importNode` used to create template. ' +
```
justinfagnani, line 1242: I wonder if we should start filing issues for APIs to remove in the next major, jus to track.

PR #5249, opened 2026-03-02
[labs/virtualizer] CSS-based direction detection and axis API

### LINKED ISSUE #4505: [labs/virtualizer] Document `grid` layout

graynorton:
Overhaul how Virtualizer handles writing mode and direction. Instead of requiring an explicit `direction: 'horizontal'` layout config, the virtualizer now reads `writing-mode` and `direction` from CSS automatically. A new `axis` property provides the easy path for controlling virtualization direction (e.g., `axis="inline"` for horizontal carousels).

### Changes

- **CSS-based direction detection**: Internal coordinates converted from physical (`left`/`top`/`width`/`height`) to logical (`block`/`inline`). All CSS writing modes supported (`horizontal-tb`, `vertical-lr`, `vertical-rl`), including window scrolling.
- **`axis` property**: New `'block' | 'inline'` property for inline-axis scrolling.
  - Works by swapping the virtualizer's own `writing-mode` and restoring the original on children.
  - **Dual writing-mode tracking**: Host and scroller writing-modes tracked separately for correct scroll coordinates when they differ.
- **`pin` property**: New first-class `pin` property on Virtualizer, LitVirtualizer, and the virtualize directive, replacing the layout config approach.
  - This change, along with replacing the layout's `direction` property with the virtualizer's `axis` property, means that all universally applicable API features are exposed on virtualizer itself, leaving layout options for truly layout-specific API features
- **Deprecation**: `direction` and `pin` layout config options still work but emit console warnings.
- **Zero-size warning**: Removed the `min-block-size: 150px` scroller default; a console warning now fires if a scroller-mode virtualizer has zero size.
- **Warning infrastructure**: New `src/warnings.ts` with global and per-instance deduplication.
- **README**: Rewritten direction/writing-mode docs, new `axis` and `pin` docs, updated API reference.
  - Bonus: added docs for `grid` layout (closes #4505)

### DIFF
--- a/packages/labs/virtualizer/src/layouts/flow.ts
+++ b/packages/labs/virtualizer/src/layouts/flow.ts
@@ -5,15 +5,11 @@
  */
 
 import {SizeCache} from './shared/SizeCache.js';
-import {BaseLayout, dim1} from './shared/BaseLayout.js';
+import {BaseLayout} from './shared/BaseLayout.js';
 import {
   Positions,
-  Size,
-  Margins,
-  margin,
-  ScrollDirection,
-  offsetAxis,
-  ChildMeasurements,
+  LogicalSize,
+  ChildLayoutInfo,
   BaseLayoutConfig,
   LayoutHostSink,
 } from './shared/Layout.js';
@@ -44,18 +40,6 @@ export const flow: FlowLayoutSpecifierFactory = (config?: BaseLayoutConfig) =>
     config
   );
 
-function leadingMargin(direction: ScrollDirection): margin {
-  return direction === 'horizontal' ? 'marginLeft' : 'marginTop';
-}
-
-function trailingMargin(direction: ScrollDirection): margin {
-  return direction === 'horizontal' ? 'marginRight' : 'marginBottom';
-}
-
-function offset(direction: ScrollDirection): offsetAxis {
-  return direction === 'horizontal' ? 'xOffset' : 'yOffset';
-}
-
 function collapseMargins(a: number, b: number): number {
   const m = [a, b].sort();
   return m[1] <= 0 ? Math.min(...m) : m[0] >= 0 ? Math.max(...m) : m[0] + m[1];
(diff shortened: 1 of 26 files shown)


packages/labs/virtualizer/src/layouts/shared/GridBaseLayout.ts
```diff
@@ -84,12 +83,9 @@ export abstract class GridBaseLayout<
         );
       }
       if (gapValue === Infinity && gap === '_gap2') {
+        // TODO: Revise this to reflect the fact that 'direction' is no longer a thing
```
graynorton: We should take care of this TODO before removing DRAFT status from the PR.

packages/labs/virtualizer/src/layouts/shared/Layout.ts
```diff
@@ -4,39 +4,93 @@
...
  */
 
-export type dimension = 'height' | 'width';
-export type Size = {
-  [key in dimension]: number;
+export type logicalSizeDimension = 'blockSize' | 'inlineSize';
+export type LogicalSize = {
+  blockSize: number;
+  inlineSize: number;
+  // [key in logicalSizeDimension]: number;
```
graynorton: Let's get rid of this and any other commented-out lines in changed blocks within the PR (no need to clean up commented-out lines in unmodified files / sections).

packages/labs/virtualizer/src/layouts/shared/GridBaseLayout.ts
graynorton: Can we explore retiring the too-cute "rolumns" and "cows" terminology here and throughout the relevant source files, now that all of our layouts are direction-agnostic? I think we can just say we have "columns" along the inline axis and "rows" along the block axis.

packages/labs/virtualizer/src/layouts/flow.ts
```diff
@@ -412,6 +399,8 @@ export class FlowLayout extends BaseLayout<BaseLayoutConfig> {
     this._physicalMin = this._anchorPos - anchorLeadingMargin;
     this._physicalMax = this._anchorPos + anchorSize + anchorTrailingMargin;
 
+    let crossSize = this._getCrossSize(this._anchorIdx);
```
graynorton: Should we use `inlineSize` instead of `crossSize` everywhere?

packages/labs/virtualizer/src/layouts/masonry.ts
```diff
@@ -38,26 +58,52 @@ export const masonry: MasonryLayoutSpecifierFactory = (
...
   private _positions = new Map<number, Positions>();
   private _rangeMap = new Map<number, RangeMapEntry>();
-  private _getAspectRatio?: GetAspectRatio;
+  private _getAspectRatioFromItem?: GetAspectRatioFromItem;
+  private _getAspectRatioFromElement?: GetAspectRatioFromElement;
 
-  protected _getDefaultConfig(): MasonryLayoutConfig {
-    return Object.assign({}, super._getDefaultConfig(), {
-      getAspectRatio: () => 1,
+  protected get _defaultConfig(): MasonryLayoutConfig {
```
graynorton: Maybe this can just use the inherited property, since we seem not to have any defaults anymore?

packages/labs/virtualizer/src/layouts/masonry.ts
```diff
@@ -4,17 +4,37 @@
...
+  element: Element,
+  requestReflow: () => void
+) => number;
 
 export interface MasonryLayoutConfig
   extends Omit<GridBaseLayoutConfig, 'flex' | 'itemSize'> {
   flex: boolean;
   itemSize: PixelSize;
-  getAspectRatio: GetAspectRatio;
+  /**
```
graynorton, line 17: In the process of preparing this PR, we came to the apparently erroneous conclusion that the replacement of `getAspectRatio` with `getAspectRatioFromItem` and `getAspectRatioFromElement` had already landed in `main`.

Since that seems to have been incorrect, we should see what it would look like to factor this change out into its own commit; it's conceptually distinct from the larger change in this PR. That would imply updating the changeset file as well.

Please confirm this is the right way forward.

graynorton (commented):
Leaving comments for Claude.

PR #5252, opened 2026-03-04
[labs/router] Prefix Routing Exclusion

ChristopherDavenport:
Website like Github Pages, or if you are hosting multiple single page applications on the same origin have a difficulty currently as the routing behavior of anchors intercepts those to the single page application. So if you have multiple apps and link between them, the router current intercepts the link since they are from the same origin.

I attempted to add support to exclude from the anchor, and then ran down the other issues that brought up, and some of those I dont really love how they ended up.

I had implemented something like this, but hadn't tried going upstream before, so I apologize if I should've started with the issue first rather than a proof of concept, I didn't see an issue and figured a demonstration of the functionality would illuminate the problem I was attempting to solve.


ChristopherDavenport:
Open to changes here, but in many contexts this is a necessary item to make sure you aren't gated in a sub-site.

PR #5260, opened 2026-03-10
Improve scoped custom elements support

sorvell:
DO NOT MERGE UNTIL https://github.com/lit/lit/pull/5246

* LitElement uses its renderRoot as `creationScope` by default. This means its `customElementsRegistry` is used by lit when creating elements.
* [scoped-registry-mixin]: updated to support latest platform spec
* [ssr-client]: hydration now supports closed shadowRoots.


packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -5,16 +5,25 @@
...
-import {adoptStyles} from '@lit/reactive-element/css-tag.js';
 
 // Proposed interface changes
 declare global {
+  interface CustomElementRegistry {
+    initialize(root: Element | DocumentFragment): void;
+  }
   interface ShadowRootInit {
+    customElementRegistry?: CustomElementRegistry;
+    // old property, retained for backwards compatibility
```
justinfagnani: Can you mark this deprecated with `@deprecated`?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -5,16 +5,25 @@
...
     customElements?: CustomElementRegistry;
   }
   interface ShadowRoot {
+    customElementRegistry?: CustomElementRegistry;
+    // old property, retained for backwards compatibility
```
justinfagnani: deprecate

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -5,16 +5,25 @@
  */
 
 import type {LitElement} from 'lit';
-import {adoptStyles} from '@lit/reactive-element/css-tag.js';
 
 // Proposed interface changes
```
justinfagnani: remove?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
      * Obtains the scoped elements definitions map
      */
     static elementDefinitions?: ElementDefinitionsMap;
+    protected static _elementDefinitions?: ElementDefinitionsMap;
     static registry?: CustomElementRegistry;
 
-    override createRenderRoot() {
+    protected static finalize() {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (superclass as any).finalize.call(this);
```
justinfagnani, line 54: Does `super.finalize()` not work on the static side?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
...
+      if (
+        this.hasOwnProperty('elementDefinitions') &&
```
justinfagnani, line 56: Prefer `Object.hasOwn()` these days. I like to tear it off at the top of the module too.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
...
+        !this.hasOwnProperty('_elementDefinitions')
+      ) {
+        this._elementDefinitions = {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ...(this._elementDefinitions ?? {}),
```
justinfagnani: You can spread `undefined`


```suggestion
          ...(this._elementDefinitions),
```

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
      * Obtains the scoped elements definitions map
      */
     static elementDefinitions?: ElementDefinitionsMap;
+    protected static _elementDefinitions?: ElementDefinitionsMap;
```
justinfagnani, line 47: Can you add a comment for this so we know how it's different from `elementDefinitions`?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
      * Obtains the scoped elements definitions map
```
justinfagnani: Can you update this comment? I don't know how "Obtains" works here.
```suggestion
     * An object describing the scoped custom element registrations to make for this class. They keys are the tag names to use for the definitions.
```

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
```diff
@@ -32,33 +41,63 @@ export function ScopedRegistryHost<SuperClass extends LitElementConstructor>(
...
+        };
+        this.registry = undefined;
+      }
+    }
+
+    override attachShadow(init: ShadowRootInit) {
       const constructor = this.constructor as typeof ScopedRegistryMixin &
         typeof LitElement;
+      // Detect supported API and supply expected property.
+      // For backwards compatibility, use older property when:
```
justinfagnani, line 71: Do we really need to worry about backwards compatibility with the mixin? It's in labs and we can release a new major version and say it requires at least lit / lit-html 3.4.0.

packages/lit-html/src/lit-html.ts
```diff
@@ -1220,7 +1226,29 @@ class TemplateInstance implements Disconnectable {
       el: {content},
       parts: parts,
     } = this._$template;
-    const fragment = (options?.creationScope ?? d).importNode(content, true);
+    // Moving forward, the `customElementRegistry` should be passed via
```
justinfagnani, line 1229: If you chain this PR into `creation-scope` instead of `main`, then we won't see this commits.

justinfagnani (commented):
looks good!

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Done. It's just so we don't mess with the user provided object.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: There's probably an incantation but it was a rabbit hole of errors.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: We need to change TS to use es2022 for this. Does that add too much stress?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Given the complexity of the setups here, it seems worth it?

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Done.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Removed.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Done.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
sorvell: Done.

packages/lit-html/src/lit-html.ts
sorvell: Done.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
justinfagnani: oh, nevermind then. es2022 beings class fields and we should do a major for that. At that point we can change the creationScope API.

packages/labs/scoped-registry-mixin/src/scoped-registry-mixin.ts
justinfagnani: I worry about the complexity of even trying to support new/old combinations here. It might be easier to just say that devs _have_ to update the polyfill, mixin, and lit to try to new stuff.

PR #5262, opened 2026-03-11
[labs/analyzer] Warn instead of error on CSS module import resolution

### LINKED ISSUE #5261: [labs/analyzer] Analyzer reports errors for CSS module imports

Analyzer (@lit-labs/analyzer)

### Description

The Lit Analyzer (via the CLI) reports errors for valid CSS module imports using `import ... with {type: 'css'}`:

```
src/components/button-dropdown-base/button-dropdown-base.ts:9:20 - error TS548000: Could not resolve specifier ./button-dropdown-base-styles.css to filesystem path.

9 import styles from './button-dropdown-base-styles.css' with {type: 'css'};
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

These imports are valid and don't cause errors in this project's TypeScript build and work correctly at runtime in Chrome.

### Reproduction

Any CSS import with import attributes should reproduce the error.

### Workaround

None

### Is this a regression?

### Affected versions

3.3.x

n/a

justinfagnani:
Fixes #5261 

TypeScript's resolveModuleName() doesn't work for non-TS/TS imports, so when that fails we now check to see if TypeScript should have been able to resolve the name by checking the extension, and only report an error if the file was .js, .ts, .d.ts, etc. If the file is something like .css, we emit a wraning.

Added test packages for both relative and bare specifier imports including package exports and monorepo workspace symlinks.

A previous version of this PR used enhanced-resolve to actually resolve the file, but that didn't work in the browser. Hopefully a future version of TypeScript will be able to resolve .css files similar to how it can with .json files with the resolveJsonModule option.

### DIFF
--- a/.changeset/chubby-moose-remain.md
+++ b/.changeset/chubby-moose-remain.md
@@ -1,0 +1,5 @@
+---
+'@lit-labs/analyzer': patch
+---
+
+Warn instead of error on unresolved non-TypeScript or non-JavaScript imports
--- a/package-lock.json
+++ b/package-lock.json
@@ -27142,7 +27142,6 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@parse5/tools": "^0.7.0",
-        "enhanced-resolve": "^5.20.0",
         "lit-html": "^3.1.2",
         "package-json-type": "^1.0.3",
         "parse5": "^7.3.0",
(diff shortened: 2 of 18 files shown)


rictic (commented):
The PR description doesn't seem to match the implementation / tests.

Looks like we still give a diagnostic for css files, we just give a slightly better one?

PR #5265, opened 2026-03-15
[labs/ssr] Add configuration on LitElementRenderer for calling connectedCallback on LitElement subclasses

### LINKED ISSUE #5175: More control over `connectedCallback` when server-side Rendering
### Should this be an RFC?

SSR (@lit-labs/ssr)

### Description

Re https://github.com/shoelace-style/webawesome/issues/1858

I stared building an app using Web Awesome components and `@lit-labs/context`. To get context running, events and `connectedCallback` need to be enabled. I enable them both with

```js
globalThis.window = getWindow({
    includeJSBuiltIns: true,
})
globalThis.litSsrCallConnectedCallback = true
```

This is problematic when 3rd party elements come into play because I cannot expect authors to not call client-only APIs when `connectedCallback` is not used on the server by default.

More shimming is probably a rabbit hole so I'm thinking that maybe it could be possible to selectively enable `connectedCallback` for my elements where I use context.

A decorator maybe?

```ts
import { ssrConnectedCallback } from '@lit-labs/ssr`

@customElement('my-element')
@ssrConnectedCallback
export class MyElement {
}
```

This way, when `litSsrCallConnectedCallback` is set, `connectedCallback` would only be called on annotated element.

My reflex was to decommission the `litSsrCallConnectedCallback` flag. If you'd intend to keep the current behavior of enabling all `connectedCallback`, it would probably need to be improved and not removed. Thus, to enable the callback ony to annotated elements, one could instead set something like

```
globalThis.litSsrCallConnectedCallback = {
  annotatedOnly: true
}
```

kyubisation:
This PR introduces the static property `renderOptions?: ((element: LitElement) => LitElementRendererRenderOptions | undefined)[]` on `LitElementRenderer`. This allows an easier or more fine granular configuration for which elements `connectedCallback` should be called or SSR should be skipped.

This PR also deprecates `globalThis.litSsrCallConnectedCallback` as `LitElementRenderer.renderOptions` is a full replacement.

Closes #5175 (more utility is planned, but that might take longer to get consensus)

### DIFF
--- a/packages/labs/ssr/src/lib/lit-element-renderer.ts
+++ b/packages/labs/ssr/src/lib/lit-element-renderer.ts
@@ -36,15 +36,59 @@ LitElement.prototype['createRenderRoot'] = function () {
 };
 
 /**
+ * The render options for a specific element in LitElementRenderer.
+ */
+export interface LitElementRendererRenderOptions {
+  /**
+   * Whether to call connectedCallback during SSR.
+   * @default false
+   */
+  connectedCallback?: boolean;
+  /**
+   * Whether to disable SSR for the element.
+   * @default false
+   */
+  disableSsr?: boolean;
+}
+
+/**
  * ElementRenderer implementation for LitElements
  */
 export class LitElementRenderer extends ElementRenderer {
   override element: LitElement;
 
+  private _disabled = false;
+
   static override matchesClass(ctor: typeof HTMLElement) {
     // This property needs to remain unminified.
     return (ctor as unknown as typeof LitElement)['_$litElement$'];
   }
+
+  /**
+   * Configure options for specific elements.
+   * Callbacks are called in order for each element being rendered and can be
+   * used to configure options such as whether to call connectedCallback for
+   * a given element or to disable SSR.
+   *
+   * @example
+   *
+   * ```ts
+   * import {LitElementRenderer} from '@lit-labs/ssr';
+   *
+   * // Disable SSR for `my-element`.
+   * LitElementRenderer.renderOptions.push(
+   *   (element) => element.localName === 'my-element' ? {disableSsr: true} : undefined
+   * );
+   *
+   * // Call connectedCallback for `my-element` by returning an options object with `connectedCallback` set to true.
+   * LitElementRenderer.renderOptions.push(
+   *   (element) => element.localName === 'my-element' ? {connectedCallback: true} : undefined
+   * );
+   * ```
+   */
+  static readonly renderOptions: ((
+    element: LitElement
+  ) => LitElementRendererRenderOptions | undefined)[] = [];
 
   constructor(tagName: string) {
     super(tagName);
@@ -82,9 +126,31 @@ export class LitElementRenderer extends ElementRenderer {
   }
 
   override connectedCallback() {
-    // Optionally call connectedCallback via setting: `litSsrCallConnectedCallback`
-    // Enable this flag to process events dispatched handled via connectedCallback.
     if (globalThis.litSsrCallConnectedCallback) {
+      console.warn(
+        'litSsrCallConnectedCallback is deprecated. ' +
+          'Please use LitElementRenderer.renderOptions instead.'
+      );
+    }
+
+    let renderOptions: LitElementRendererRenderOptions | undefined;
+    for (const optionsCallback of LitElementRenderer.renderOptions) {
+      const options = optionsCallback(this.element);
+      if (options) {
+        renderOptions = options;
+        break;
+      }
+    }
+
+    if (renderOptions?.disableSsr) {
+      this._disabled = true;
+      return;
+    }
+
+    if (
+      globalThis.litSsrCallConnectedCallback ||
+      renderOptions?.connectedCallback
+    ) {
       // Prevent enabling asynchronous updating by overriding enableUpdating
       // with a no-op.
       this.element['enableUpdating'] = function () {};
@@ -95,9 +161,10 @@ export class LitElementRenderer extends ElementRenderer {
         const className = this.element.constructor.name;
         console.warn(
           `Calling ${className}.connectedCallback() resulted in a thrown ` +
-            'error. Consider removing `litSsrCallConnectedCallback` to ' +
-            'prevent calling connectedCallback or add isServer checks to ' +
-            'your code to prevent calling browser API during SSR.'
+            'error. Consider configuring `LitElementRenderer.renderOptions` ' +
+            'to prevent calling connectedCallback for unsupported elements ' +
+            'or add isServer checks to your code to prevent calling browser ' +
+            'API during SSR.'
         );
         throw e;
       }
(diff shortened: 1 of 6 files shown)


packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -35,17 +35,52 @@ LitElement.prototype['createRenderRoot'] = function () {
...
 
+  /**
+   * Configure options for a specific element.
+   * This callback is called for each element being rendered and can be used
+   * to configure options such as whether to call connectedCallback for a given
+   * element or to disable SSR.
+   *
+   * @example
+   *
+   * ```ts
```
thescientist13, line 75: Apologies for missing the obvious here as I'm not super familiar with LitRenderer, but is this meant to be part of the custom element class itself, or "globally" defined somewhere like `globalThis.litSsrCallConnectedCallback` was intended to be?

For example:
```js
import { LitElement, html } from "lit";

export default class ProductsList extends LitElement {
  connectedCallback() {
    this.products = [{
      id: 1,
      name: "Product 1",
    }, {
      id: 2,
      name: "Product 2",
    }];
  }

  render() {
    const { products } = this;

    return html`
      <h1>Products List</h1>
      <ul>
        ${products.map((product) => {
          const { id, name } = product;

          return html`<li>${id}) ${name}</li>`;
        })}
      </ul>
    `;
  }
}

customElements.define('products-list', ProductsList);
```

I was initially interpreting the opt-in would be something like this at the class level, per the PR description
> This PR introduces the static property callConnectedCallback?: boolean | ((element: LitElement) => boolean) on LitElementRenderer.
```js
import { LitElement, html } from "lit";

export default class ProductsList extends LitElement {
  static callConnectedCallback = true;

  // ...
}

customElements.define('products-list', ProductsList);
```

But if I am understanding this current proposal correctly here, would I be correct in saying that where I am instrumenting `globalThis.litSsrCallConnectedCallback` currently (within a worker thread that is running the SSR operations / APIs), I would replace that the `renderOptions` configuration, and thus _independent_ of the class definition?
```js
import {LitElementRenderer} from '@lit-labs/ssr';

LitElementRenderer.renderOptions = (element) => element.localName === 'product-list' ? {connectedCallback: true} : true;
```

Or is it indeed opt-in at the custom element definition and I am just reading this all wrong? 😅 

Thanks!

kyubisation: Hello @thescientist13 
Thanks for reaching out. I just noticed I forgot to update the PR description. I have fixed that now.

Your last interpretation is correct; This is a global configuration separate from the custom element class body.
While I would like to extend this possibility in the future to the custom element class, during the discussion in the community call, it was much more difficult to reach consensus on that, so I wanted to provide the basic functionality without including class level configuration.

So this PR targets the following usage (taken from your examples):

```ts
import {LitElementRenderer} from '@lit-labs/ssr';

LitElementRenderer.renderOptions = (element) => element.localName === 'product-list' ? {connectedCallback: true} : true;
```

thescientist13: I see, thanks for confirming @kyubisation .

That makes sense, but I guess one thing I was curious about in this approach vs something on the element class itself was that say you are a meta-framework (11ty, Astro, etc) creating a plugin for Lit (e.g. so as a developer you are not in control of the Lit SSR integration directly), does that mean the end user would need to pass a list of "known" tag names for this particular approach to "work"?

Take the WebAwesome example from the issue, in this case would the expectation be that you have to support something like this as the plugin / Lit SSR wrapper author?

```ts
// my-framework.config.ts
import { litSsrPlugin } from '@my-framework/lit-ssr';

export default {
  plugins: [
    litSsrPlugin({
      // passing strings, or Regex, or whatever feels like being supported
      ssrTagNames: [
        /wc-/,
        'products-list',
        // etc
      ]
    })
  ]
}
```

Then that plugin / framwork author would have to do something like this?
```ts
import {LitElementRenderer} from '@lit-labs/ssr';
import { config } from '@my-framework/lit-ssr';

LitElementRenderer.renderOptions = (element) => {
  for(const tag of config.ssrTagNames) {
    if(tag instanceof Regex) {
      if(tag.test(element.localName) {
        return {connectedCallback: true}
      }
    } else if(typeof tag === 'string') {
      if(tag === element.localName) {
        return {connectedCallback: true}
      }
    }
  }

  return true;
}
```

kyubisation: Yes, at the moment that would be the case.
However, as mentioned above, I plan to implement more functionality, which would allow defining configuration on class level. But during the community call the API shape was contentious, and most likely will require a few rounds of discussion and implementation.
Therefore this PR is intended to alleviate the immediate problem and pave the way for future improvement.

thescientist13: Ok, thanks @kyubisation and sounds good; thanks for confirming that snippet would be a reasonable way to support this capability while awaiting potential enhancements for defining the opt-in at the component level. 🫡  

I've got an implementation of the `globalThis` approach working at present, so looking forward to seeing this land so I can formally update on my side to account for this particular iteration, before pushing it out from my end.  👍

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -35,17 +35,52 @@ LitElement.prototype['createRenderRoot'] = function () {
...
+   * import {LitElementRenderer} from '@lit-labs/ssr';
+   *
+   * // Disable SSR for `my-element` by returning false.
+   * LitElementRenderer.renderOptions = (element) => element.localName !== 'my-element';
```
justinfagnani: My big worry about this kind of API is multiple actors overwriting each other. Note: talked about this live.

kyubisation: Addressed via Set of callbacks.

packages/labs/ssr/src/lib/lit-element-renderer.ts
```diff
@@ -60,26 +65,30 @@ export class LitElementRenderer extends ElementRenderer {
...
-   * LitElementRenderer.renderOptions = (element) => element.localName === 'my-element' ? {connectedCallback: true} : true;
+   * LitElementRenderer.renderOptions.add(
+   *   (element) => element.localName === 'my-element' ? {connectedCallback: true} : undefined
+   * );
    * ```
    */
-  static renderOptions?: (
-    element: LitElement
-  ) => boolean | LitElementRendererRenderOptions;
+  static readonly renderOptions = new Set<
```
justinfagnani: Looking at how this is used, I think an Array might be a better container than a Set. Iteration of an array should be faster than over the items in a set, and this collection is iterated over once or twice for every element. Hopefully it remains a small list.

justinfagnani (changes requested):

packages/labs/ssr/src/lib/lit-element-renderer.ts
kyubisation: Refactored accordingly 👍

kyubisation:
Squashed and rebased in order to avoid the CLA failure

justinfagnani (approved):
thanks!

justinfagnani:
We should add lit.dev docs for this

Issue #5291, opened 2026-04-10
[labs/virtualizer] Reconsider viewport semantics — current logic conflates visibility with available layout space


graynorton:
## Problem

A core premise of virtualization is that *the virtualized view should be laid out and rendered just as an unvirtualized equivalent would be* — virtualization is purely an optimization that limits which items are rendered along the scroll axis, not a transformation of the layout itself.

The current `Virtualizer._updateView()` computes a single `viewportSize` from the geometric intersection of the host element and all of its clipping ancestors, and uses this same value for **both** layout-sizing decisions and range calculation, on **both** axes. This conflates two concerns that should be independent.

## Two distinct concerns

1. **Available layout space** — what the layout uses for sizing decisions (column counts, item dimensions, etc.). This should always reflect the host's configured layout dimensions on **both axes**, regardless of clipping ancestors. The layout's decisions should match what an unvirtualized equivalent would produce.

2. **Currently visible range** — what determines which items fall in the rendered range. This **should** be affected by clipping ancestors (e.g., a virtualizer fully off-screen should render nothing; one partially clipped along its scroll axis should render only the visible portion), but **only along the virtualization axis**. Clipping along the non-virtualization axis should never affect which items are rendered.

## Examples

**Cross-axis layout (currently broken):** A virtualizer with explicit `inline-size: 800px` using a grid layout, placed inside a parent of width `600px` with `overflow: auto`. The user has scrolled the parent so that only the leftmost 600px of the virtualizer is visible.

- *Unvirtualized equivalent*: the grid lays out across the full 800px (e.g., 4 columns at 200px each); the user scrolls the parent horizontally to reveal the rightmost column.
- *Current behavior*: `viewportSize.inlineSize` is 600px, so the grid lays out 3 columns. The configured 800px is effectively ignored.

**Scroll-axis visibility for a `scroller: true` virtualizer (status uncertain):** A `scroller: true` virtualizer placed inside a scrollable parent that scrolls it in and out of view.

Tracing through the current code: scroll listeners are attached to the parent (via `getClippingAncestors`), so the virtualizer does re-run its update cycle as the parent scrolls. The inset/viewport math then produces a reduced `viewportSize.blockSize` and a `viewportScroll.block` adjusted to account for the clipped portion, which has the *approximate* effect of rendering only the items that fall within the visible window. When the virtualizer is fully off-screen the math degenerates to a zero-size viewport, which likely results in no items rendered.

[...]

PR #5292, opened 2026-04-10
[labs/virtualizer] Add managed viewport mode and refactor scroll handling into ScrollSource strategies

### LINKED ISSUE #5295: [labs/virtualizer] Add managed viewport mode for custom/synchronized scrollers

### Description

The virtualizer currently supports two scroll modes: ancestor scrolling (default) and self-scrolling. Both rely on DOM observation — a `ResizeObserver` for viewport size and scroll events for position. This is limiting for use cases where the virtualizer should follow an externally-managed scroll position rather than observe its own DOM:

- Custom scroller implementations that want to drive the virtualizer without bolting onto a real scrollable DOM element
- Synchronized scrollers (e.g. pinned headers, side-by-side lists) where multiple virtualizers need to stay in lockstep
- Embeddings inside hosts that do their own scroll/viewport management

### Proposal

Add a new `'managed'` scroller mode in which DOM observation is skipped entirely. The virtualizer is driven by an externally-set `viewport` property instead.

Alongside the new mode, refactor scroll position and viewport size acquisition out of `Virtualizer` into a `ScrollSource` strategy interface with three implementations:

- `ManagedScrollSource` — new, driven by the external `viewport` property
- `SelfScrollSource` — existing self-scroller behavior
- `AncestorScrollSource` — existing ancestor-scroller behavior

The two DOM-based modes share a `BaseDomScrollSource`. `Virtualizer` becomes a slimmer orchestrator that delegates scroll/viewport concerns to the active source.

### Additional changes bundled with the refactor

- String aliases `'self'` / `'ancestor'` / `'managed'` for the `scroller` config (existing boolean values still supported).
- Public subpath export for `Virtualizer.js` so the underlying class can be imported and instantiated directly (currently only `LitVirtualizer` and the `virtualize` directive are exported).

### Notes

Smooth-scroll `scrollIntoView({behavior: 'smooth'})` is not supported in managed mode in the initial implementation; pin-based instant scroll-into-view works. A future enhancement could add an optional synchronous handshake for managed-mode consumers that want gap-free scroll-error correction.

Addressed by #5292.

graynorton:
Fixes #5295

## Summary

Adds a new `'managed'` scroller mode to the virtualizer in which DOM observation for scroll position and viewport size is skipped entirely; the virtualizer is instead driven by an externally-set `viewport` property. Useful for custom scroller implementations and synchronized scrollers.

Also includes a substantial internal refactor that extracts scroll-position and viewport-size acquisition out of `Virtualizer` into a `ScrollSource` strategy interface with three implementations (`ManagedScrollSource`, `SelfScrollSource`, `AncestorScrollSource`). The two DOM-based modes share a `BaseDomScrollSource` class. `Virtualizer` is now a slimmer orchestrator that delegates all scroll/viewport concerns to the active source.

Adds string aliases `'self'` / `'ancestor'` / `'managed'` for the `scroller` config (boolean values still supported), and a public subpath export for `Virtualizer.js` so the underlying class can be imported and instantiated directly.

This lands after #5249 (CSS-based direction detection) and #5279 (bug fixes). It is a sibling of #5280 (positioning method) and does not depend on it.

A separate issue #5291 was filed during this work to capture broader concerns about how the virtualizer's viewport math conflates "available layout space" with "currently visible range." That issue is independent of this PR; existing viewport semantics are preserved here.

### Scroll-into-view contract

A second wave of work in this PR completes the public scroll-into-view API for managed mode (and tightens it across the DOM modes). The previous managed-mode implementation set `layout.pin = options` and returned void, with no way for the consumer to learn the destination, animate toward it, receive mid-flight refinements, or be notified when the operation ended.

The new contract:

- `Virtualizer.element(N).scrollIntoView({behavior?, signal?, ...})` synchronously returns the destination coordinates (in viewport coords) across all scroll modes. New `signal: AbortSignal` option is also accepted across all modes.

- Managed mode branches by behavior:
  - **Instant** — synchronous teleport via the existing pin/`_scrollError` chain. The source suppresses the initial `ScrollErrorEvent` dispatch (the consumer already has the destination via the return value) but performs its internal viewport self-correction so subsequent reflows stay consistent. No intent state.
  - **Smooth** — registers a `ManagedSmoothIntent` that tracks destination shifts (via `destinationchanged`), at-rest arrival inference, and AbortSignal-driven cancellation. No pin is held; the consumer drives the animation and the layout's rendered range follows their progressive viewport pushes.

- DOM modes now route smooth scrolls through a `DomSmoothIntent` that bridges the existing `ScrollerController.managedScrollTo` arrival/retarget/end callbacks into a unified intent lifecycle, including `AbortSignal` cancellation.

- Three new event classes:
  - `ScrollErrorEvent` — formalizes the previously-untyped `'scrollerror'` CustomEvent (managed mode only). The source self-corrects its own internal viewport tracking at dispatch time, so the consumer is not required to push viewport back.
  - `DestinationChangedEvent` — fires during an active managed-smooth intent when the layout refines the destination beyond ε.
  - `ScrollIntoViewEndedEvent` — fires for smooth intents in any mode when the intent ends, with `reason: 'arrived' | 'cancelled' | 'replaced'`.

- Native DOM `scrollIntoView` fallback for in-range targets is removed in managed mode (the model is "consumer owns scroll state, so the virtualizer never touches the DOM scroller"). Other modes keep it.

- At-rest arrival detection only counts *consumer-pushed* viewports as samples, not all `updateView` invocations. Without this gate, internal reflows produce phantom "consecutive equal samples" that fire `arrived` falsely. Verified in manual testing.

### Flow layout estimation fix

Bonus bug fix surfaced during verification: `FlowLayout._estimatePosition` was multiplying `(delta - 1)` by `c.averageChildSize` and `c.averageMarginSize` directly. Both are `0` until measurements arrive, so the formula collapses to `refItem.pos` for any index past `_last` — meaning two indices well past the rendered range would return the same position (e.g. items 500 and 999 both returning the position of item 25). Fix uses `_getAverageSize()` (which falls back to `_itemSize.blockSize`) so positions remain monotonic before measurements land.

## Changes

- **`Virtualizer.ts`**: New `Viewport` type and re-export, `viewport` config option, reactive `viewport` getter/setter, three-way `scroller` config typing. ~600 lines deleted as the inline DOM-observation paths move into the new `ScrollSource` implementations. Now also: `_scrollElementIntoView` returns destination coordinates; managed-mode path skips native fallback; new event types declared in `HTMLElementEventMap`; `VirtualizerChildElementProxy.scrollIntoView` signature updated for `signal` + return value.
- **`scroll-sources/` (new directory)**:
  - `ScrollSource.ts` — strategy interface and `ScrollSourceHost` callback interface
  - `_dom-utils.ts` — shared shadow-DOM-aware ancestor-traversal helpers
  - `ManagedScrollSource.ts`, `AncestorScrollSource.ts`, `SelfScrollSource.ts`, `BaseDomScrollSource.ts`
  - **New:** `ScrollIntoViewIntent.ts` — base class + `ManagedSmoothIntent` + `DomSmoothIntent` (smooth-scroll intent abstraction)
- **`ScrollerController.ts`**: New `cancelManagedScrollTo()` method for AbortSignal-driven termination.
- **`events.ts`**: Three new event classes (`ScrollErrorEvent`, `DestinationChangedEvent`, `ScrollIntoViewEndedEvent`), `PhysicalCoordinates` type, `ScrollIntoViewEndedReason` union.
- **`layouts/flow.ts`**: `_estimatePosition` falls back to `_getAverageSize()` when the metrics cache is empty.
- **`virtualize.ts` / `LitVirtualizer.ts`**: Pass-through for `viewport`; widened `scroller` typing. `LitVirtualizer.scroller` uses a custom attribute converter so both `<lit-virtualizer scroller>` (legacy bare-attribute boolean form) and `<lit-virtualizer scroller="managed">` work.
- **`README.md`**: Expanded "Managed viewport mode" section covering the new scroll-into-view contract, return value, AbortSignal, and the three events. Updated `scrollIntoView` example with return value and signal. New API entries for the three events. New appendix at the end documenting consumer patterns for managed mode (Pattern A vs Pattern B for smooth scroll, animation timing strategies, keyboard navigation patterns), capturing the empirical findings from manual verification.
- **`package.json`**: New `./Virtualizer.js` subpath export; `scroll-sources/` added to `files` and the wireit `output` so the new internal modules ship in the published tarball.
- **Tests**: 24 new tests total across `managed-viewport.test.ts` (managed mode end-to-end + the new scroll-into-view contract) and the new `scroll-into-view-cross-mode.test.ts` (cross-mode contract: return value, AbortSignal, scrollintoviewended).

## Commit structure

Nine commits matching a staged plan, each independently committable and verifiable:

1. `5962f731` — Add `ManagedScrollSource` and `viewport` property (managed mode via inline branching)
2. `303937d1` — Migrate ancestor mode into `AncestorScrollSource`
3. `44c1c16d` — Migrate self-scroller mode and complete the refactor (extract `BaseDomScrollSource`, remove all dead inline code)
4. `00ba1c7c` — Add string aliases for `scroller`
5. `1a770de0` — Expose `viewport` through `<lit-virtualizer>` and the directive
6. `6dbd3cc1` — Changeset, public `Virtualizer.js` export, README docs

(Plus `d4bea179` syncing ignore files for the new `/scroll-sources/` build output.)

Then the second wave:

7. `bf51604e` — Fix `FlowLayout._estimatePosition` degeneracy with empty metrics cache
8. `50b45dbd` — Add scroll-into-view intent contract for managed mode
9. `7966d357` — Document managed viewport mode and consumer patterns

## Test plan

- [x] All 227 tests pass (203 existing + 24 new)
- [x] Manual verification via a playground demo (host inside `overflow: hidden`, translated by CSS transitions) that exercises:
  - Both `behavior` modes (instant + smooth)
  - Both consumer animation patterns for smooth (push-destination-once + push-current-per-RAF)
  - Three time-allocation policies (fixed, total budget, velocity-proportional)
  - Five key-handling policies (none, ignore-in-flight, coalesce, acceleration, acceleration+coalesce)
  - All three event types (scrollerror, destinationchanged, scrollintoviewended)
  - AbortSignal cancellation (mid-flight abort, replacement by new intent, natural arrival)

## Notes

The README appendix captures consumer patterns and trade-offs that emerged from the manual verification work — particularly around smooth-scroll animation strategies and keyboard-navigation pitfalls — that real consumers building on managed mode will likely want to know. It's structurally additive (no separate doc page yet); future restructuring is unblocked.


sorvell:
[Claude review](https://claude.ai/share/48a0a4b4-410f-4407-9368-3a7b4f620502)

PR #5313, opened 2026-04-21
fix: set correct path to sub path imports in generators

tibnor:
Befor the fix, when generating wrappers on windows the path became invalid for the import
```js
import {TypeSub} from '@lit-internal/test-element-a/sub\type-sub.js';
```
This fix, changes this to
```js
import {TypeSub} from '@lit-internal/test-element-a/sub/type-sub.js';
```
It also adds test for this.

### DIFF
--- a/packages/labs/gen-wrapper-angular/package.json
+++ b/packages/labs/gen-wrapper-angular/package.json
@@ -20,7 +20,7 @@
     "build": "wireit",
     "test": "wireit",
     "test:gen": "wireit",
-    "test:update-goldens": "UPDATE_TEST_GOLDENS=true npm run test:gen"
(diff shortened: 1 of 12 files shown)


packages/labs/gen-wrapper-angular/package.json
```diff
@@ -20,7 +20,7 @@
     "build": "wireit",
     "test": "wireit",
     "test:gen": "wireit",
-    "test:update-goldens": "UPDATE_TEST_GOLDENS=true npm run test:gen"
+    "test:update-goldens": "cross-env UPDATE_TEST_GOLDENS=true npm run test:gen"
```
louis-bompart: From what I check, cross-env is rarely used in the repo. Instead, environment variable are often set using wireit.

I think it might be best to adhere to the most frequent practice here.

Issue #5193, opened 2025-12-24
[labs/motion] Spring tests are flaky on Firefox


justinfagnani:
### Which package(s) are affected?

Motion (@lit-labs/motion)

### Description

One of the new Spring tests is flaking on Firefox:

```
../labs/motion/development/test/spring_test.js:

 ❌ Spring > SpringController > spring starts when connected, stops when disconnected (failed on Firefox)
      AssertionError: expected false to be true
      + expected - actual
      
      -false
      +true
      
      at ../labs/motion/src/test/spring_test.ts:72:13
```

These tests depend a bit on timing and comparison of floats, and they were hard to get working on Safari. We may need a different approach.

PR #5316, opened 2026-04-21
[tests] Hoist `.skipInCI` Mocha helper into shared test config

graynorton:
## Summary

Factors the `describe.skipInCI` mechanism that lived in `@lit-labs/virtualizer`'s local `web-test-runner.config.js` into a reusable helper at `packages/tests/src/mocha-skip-in-ci.ts`, and adopts it from two places:

- **Virtualizer**: replaces a Chrome-only `chromeLauncher.createPage` + `evaluateOnNewDocument` mechanism with a cross-browser `testRunnerHtml` hook. Existing `describe.skipInCI(...)` callsites continue to work unchanged.
- **`@lit-labs/motion`**: marks the two known-flaky Spring `"spring starts when connected, stop(s) when disconnected"` tests as `test.skipInCI(...)` so they run locally (where developers can still exercise the behavior) but skip in CI. Same underlying timing/float-comparison flakiness as tracked in #5193.

The helper installs via web-test-runner's `testRunnerHtml`, which works across all browsers supported by `@web/test-runner-playwright` (Chromium, Firefox, Webkit) and `@web/test-runner-chrome`. It adds a `.skipInCI(title, fn)` variant to Mocha's UI globals (`suite`, `test`, `describe`, `it`) that routes to `fn.skip` when `process.env.CI` is truthy and to `fn` otherwise.

## Alternative to #5315

[#5315](https://github.com/lit/lit/pull/5315) just skips both flaky tests outright. This PR unblocks the same CI failure but preserves the ability to run the tests locally, and leaves behind infrastructure other packages can adopt. Merge whichever approach is preferred and close the other.

## Test plan

- [x] Verified `packages/tests` builds and emits `mocha-skip-in-ci.{js,d.ts}` alongside the existing WTR config.
- [x] Verified shared tests pass locally with `CI=` (spring tests run, may flake) and with `CI=true` (2 extra skips, no failures).
- [x] Verified virtualizer tests pass with `CI=` (114 pass, 0 skipped) and `CI=true` (82 pass, 32 skipped — same as the prior Chrome-only mechanism produced).
- [x] Verified motion tests pass with `CI=true` (17 pass, 2 spring tests skipped).
- [ ] `Changeset Status` CI job is green.


sorvell (approved):
def seems useful... lgtm

justinfagnani:
@graynorton what do you think of this PR instead of or in addition to? https://github.com/lit/lit/pull/5317

Seems to fix the flakiness and keep the motion tests. Would it help in virtualizer tests?

PR #5317, opened 2026-04-22
test(motion): use sinon fake timers for better timing control

louis-bompart:
Relates to #5316

The changes is mostly motivated by how the tests were done in the underlying library (see [here](https://github.com/skevy/wobble/blob/develop/src/__tests__/Springs-test.ts))

I am still scratching my head on why this is flaky in the beginning however.

I ran the test in the CI 50 times without the changes in this PR (https://github.com/louis-bompart/lit/pull/1), and 50 times with (https://github.com/louis-bompart/lit/pull/2).

14 failures out of 50 in the first one, and no failure in the second (still possible that it's actually not fixed if the dice rolled in my favor 50 times in a row, but well 😂 )


justinfagnani:
Thanks @louis-bompart!

I don't love spies and mocks (and new dependencies) if they can be avoided, but if this truly fixes the flake, then I'll guess it's worth it.

Checking in w/ @graynorton and @sorvell on this one.

louis-bompart:
> Thanks @louis-bompart!
> 
> I don't love spies and mocks (and new dependencies) if they can be avoided, but if this truly fixes the flake, then I'll guess it's worth it.
> 
> Checking in w/ @graynorton and @sorvell on this one.

Yeah I tried to do without but from experience time-based API (including animation frame) are hard to control deterministically.

An alternative is to use a combination of Promise & timeout, but even then, YMMV, [(this article](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) is a good read on the topic, although it is probably ancillary to the situation at play here) 
For the dependencies FWIW, `sinon` is already present in another package, so I used it here to avoid having multiple mocking libraries.

I'm fairly confident this PR will fix the flakes, and my hunch is around the timing, but I did not determine with absolute certitude as to why.

Issue #5323, opened 2026-04-27
Amend @types/node constraint


SignpostMarv:
### Which package(s) are affected?

SSR (@lit-labs/ssr)

### Description

lit-labs/ssr runs fine under node 25, but can't be installed alongside the appropriate `@types/node` package due to the version constraint excluding the package targetting node 25.

PR #5324, opened 2026-04-27
[labs/ssr] amend version constraint to allow use of @types/node for node 26 (#5323)

SignpostMarv:


packages/labs/ssr/package.json
```diff
@@ -237,7 +237,7 @@
     "parse5": "^7.1.1"
   },
   "peerDependencies": {
-    "@types/node": ">=20.0.0 <25.0.0"
+    "@types/node": ">=20.0.0 <26.0.0"
```
justinfagnani: I suspect we actually want to remove the upper bound, but this is a good incremental improvement till we figure that out.

justinfagnani (approved):
thanks!

SignpostMarv:
Updated this for node 26

kyubisation (approved):
LGTM

PR #5325, opened 2026-04-28
fix(react): forward element props in node export when DOM is present

mhoritani:
## Summary

This change fixes a behavior gap in `@lit/react` when tooling resolves the `node` export in a DOM-like runtime (for example Vitest + jsdom).

Today, the node build treats all renders as SSR-oriented and skips the DOM prop-application path. In that situation, element props like `id` (and custom element prototype props) can be dropped even though a DOM is present.

This PR keeps SSR behavior intact, but applies element props in node mode when:

- a DOM runtime is present (`window` exists), and
- Lit SSR patching is not active (`litPatchedCreateElement` / `litSsrReactEnabled`).

## Why this is scoped this way

The goal is not to change SSR semantics. The goal is to avoid false-SSR behavior in client-like test environments that happen to resolve the node export.

So the behavior split becomes:

- **True SSR / Lit SSR active:** unchanged (`_$litProps$` path still used)
- **Node export + DOM runtime + no Lit SSR patching:** apply browser-style element prop forwarding

## Changes

- `packages/react/src/create-component.ts`
  - Adds a runtime guard (`shouldApplyDomElementProps`) for node export when DOM is present and Lit SSR is not active.
  - Preserves existing SSR-oriented behavior when Lit SSR is active.
- `packages/react/src/test/create-component_test.tsx`
  - Adds a regression test that imports the **node build** directly and asserts forwarded `id` and object property values in DOM tests.

## Validation

Ran in this branch:

- `npx playwright install chromium firefox webkit`
- `npm run test:dev --workspace=@lit/react -- --force`
- `npm run test:prod --workspace=@lit/react -- --force`
- `npm run test:node --workspace=@lit/react -- --force`

Browser matrix passed (Chromium/Firefox/WebKit).

I also validated against two downstream reproductions that originally surfaced the problem:

- https://github.com/stenciljs/output-targets/issues/791
- https://github.com/stenciljs/output-targets/pull/793

Both pass once this patched `@lit/react` is installed.

## Notes

If you’d prefer a different guard for determining “DOM-like but not Lit SSR” in node mode, I’m happy to adjust it.

### DIFF
--- a/.changeset/fair-mangos-brush.md
+++ b/.changeset/fair-mangos-brush.md
@@ -1,0 +1,5 @@
+---
+'@lit/react': patch
+---
+
+Fix node-export prop forwarding in DOM-like runtimes (e.g. Vitest + jsdom) while preserving Lit SSR behavior.
--- a/packages/react/src/create-component.ts
+++ b/packages/react/src/create-component.ts
@@ -247,6 +247,12 @@ export const createComponent = <
   const ReactComponent = React.forwardRef<I, Props>((props, ref) => {
     const prevElemPropsRef = React.useRef(new Map());
     const elementRef = React.useRef<I | null>(null);
+    const hasDom = typeof window !== 'undefined';
+    const isLitSsrRendering =
+      React.createElement.name === 'litPatchedCreateElement' ||
+      globalThis.litSsrReactEnabled;
+    const shouldApplyDomElementProps =
+      !NODE_MODE || (hasDom && !isLitSsrRendering);
 
     // Props to be passed to React.createElement
     const reactProps: Record<string, unknown> = {};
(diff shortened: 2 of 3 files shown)


mhoritani:
Hey @kevinpschaaf, is there anything I can help with to get this PR to a point where it can get reviewed?

kevinpschaaf:
Sorry don't have time now to dig in, maybe @justinfagnani or @sorvell could take a look?

mhoritani:
Hey @kevinpschaaf, @justinfagnani, @sorvell sorry to ping again, but any chance someone could have a look here? 👀 Appreciate it!

Armand-Lluka:
We would also like to have this fix in. Is there anything blocking this @justinfagnani ?

klh:
this looks good to me

PR #5334, opened 2026-05-13
[localize-tools] Fix generic call expressions inside msg() substitutions

jimsimon:
## Summary

`parseStringAsTemplateLiteral` in `@lit/localize-tools` parsed template literal bodies with `ts.ScriptKind.JS`. JavaScript has no generic-call syntax, so an expression like `${foo<T>(arg)}` inside a `msg()` template substitution was parsed as `(foo < T) > (arg)` (a `BinaryExpression` chain) instead of a `CallExpression` with a type argument list.

In contexts that re-emit the parsed AST, this prints as `foo < T > (arg)`, which evaluates to `false` at runtime and silently drops any side effects of the call. Source/English branches are unaffected because they reuse the original TypeScript-parsed source AST; translated branches that flow back through `parseStringAsTemplateLiteral` are vulnerable.

The fix is a one-line, surgical change: parse with `ts.ScriptKind.TS` so the TypeScript parser correctly disambiguates `<T>` as a type argument list. There is no behavior change for non-generic substitutions; TypeScript's parser is a strict superset of JavaScript's for this input.

### Changes
- `packages/localize-tools/src/typescript.ts`: switch `parseStringAsTemplateLiteral` to `ts.ScriptKind.TS` (with a comment explaining why).
- `packages/localize-tools/src/tests/typescript.unit.test.ts`: new unit test file with 4 tests, including a regression test that asserts a generic call substitution parses as a `CallExpression` (not a `BinaryExpression`).
- `.changeset/`: patch changeset for `@lit/localize-tools`.

## Test plan
- [x] `npm test` in `packages/localize-tools` passes (`test:unit` 80/80, `test:unit:ssr` 4/4, `test:check-tsc` clean).
- [x] New regression test fails on `main` (with `ScriptKind.JS`) and passes with the fix.
- [x] `npm run format` and ESLint clean on changed files.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-localize-tools-generic-call-substitution.md
+++ b/.changeset/fix-localize-tools-generic-call-substitution.md
@@ -1,0 +1,5 @@
+---
+'@lit/localize-tools': patch
+---
+
+Fix `parseStringAsTemplateLiteral` misparsing TypeScript generic call expressions inside `${...}` template substitutions. Previously the helper parsed template literal bodies with `ts.ScriptKind.JS`, so an expression like `${foo<T>(arg)}` was parsed as a chain of comparison operators (`(foo < T) > (arg)`) instead of a generic call. Parsing as TypeScript now correctly recognizes the type argument list and produces a `CallExpression`.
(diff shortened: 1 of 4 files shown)


justinfagnani (approved):
thanks!

PR #5340, opened 2026-05-20
[context] Expose Options types for ContextConsumer and ContextProvider

### LINKED ISSUE #4837: [Lit/Context]: Expose "Options" types
### Should this be an RFC?

Context (@lit/context)

### Description

Lit context consumer and provider controllers have `Options` types:

https://github.com/lit/lit/blob/1eb179f69d663440fd2ebd3589b6f2808d87494f/packages/context/src/lib/controllers/context-consumer.ts#L17-L21

https://github.com/lit/lit/blob/1eb179f69d663440fd2ebd3589b6f2808d87494f/packages/context/src/lib/controllers/context-provider.ts#L40-L43

These types are not exposed outside the package:

https://github.com/lit/lit/blob/main/packages/context/src/index.ts

Consider exposing them for usage outside the package.

I can't use `ConstructorParameters<>` to extract the options type because of the constructor signature overload (TypeScript's Parameters/ConstructorParameters don't play nicely with overloads)

Workaround:
- Manually copy paste the implementation of those types. sigh

Use case: I need to provide a wrapper for ContextProvider and ContextConsumer controllers to better support my lazy-loading implementation (in my case the ReactiveElementHost element and HTMLElement are separate objects, where as these controllers assumes it is the same object)

adv0r:
> **Dear maintainer** — this PR has a permanent home with methodology + [opt-out](https://github.com/adv0r/tokens-for-good/blob/main/MAINTAINER_REMOVAL.md) at [tokens-for-good](https://github.com/adv0r/tokens-for-good). A one-line "no thanks" → auto-close + blacklist. Sorry for the notification this edit caused.
>
> ---

## Summary

Re-export the `Options` interfaces from `@lit/context`'s public entry
point so that user code wrapping `ContextConsumer` / `ContextProvider`
can reference them without having to copy-paste their definitions.

Adds the following named re-exports from `packages/context/src/index.ts`:

- `ContextConsumerOptions` — alias for the `Options` interface in
  `./lib/controllers/context-consumer.js`
- `ContextProviderOptions` — alias for the `Options` interface in
  `./lib/controllers/context-provider.js`

Aliases are needed because both source modules export a type named
`Options`; the new public names make their package origin obvious and
let both be imported from the same entry point.

Also adds a changeset for `@lit/context` (patch).

Fixes #4837

## Why

The two `Options` interfaces are part of the public constructor
signatures of `ContextConsumer` / `ContextProvider`, but they were
not reachable from `@lit/context`. `ConstructorParameters<typeof X>`
cannot extract them because of the deprecated constructor overload on
each controller. The only workaround today is to copy-paste the type
definitions into user code (see the issue for a concrete wrapper /
lazy-loading use case).

## Backward compatibility

Type-only, additive change. No runtime behavior changes; existing
imports are untouched.

## AI disclosure

Per AGENTS.md / CONTRIBUTING — this is a small AI-generated, unattended
contribution made by a Cursor Anthropic Claude Opus 4.7 agent under
@adv0r as a short personal token-burn initiative before my Cursor
subscription billing cycle ends. Happy to iterate on review.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/expose-context-options-types.md
+++ b/.changeset/expose-context-options-types.md
@@ -1,0 +1,16 @@
+---
+'@lit/context': patch
+---
+
+Expose `Options` types for the `ContextConsumer` and `ContextProvider`
+controllers from `@lit/context` as `ContextConsumerOptions` and
+`ContextProviderOptions` respectively.
+
+These types were already publicly reachable via the
+`new ContextConsumer(host, options)` and `new ContextProvider(host, options)`
+constructor signatures but could not be referenced from user code (e.g.
+when wrapping the controllers), because `ConstructorParameters<…>` does
+not see through the deprecated constructor overload. Exposing them
+under distinct names avoids the collision between the two same-named
+`Options` interfaces in their source modules while keeping the existing
+public API unchanged.
--- a/packages/context/src/index.ts
+++ b/packages/context/src/index.ts
@@ -17,7 +17,9 @@ export {
 } from './lib/create-context.js';
 
 export {ContextConsumer} from './lib/controllers/context-consumer.js';
+export type {Options as ContextConsumerOptions} from './lib/controllers/context-consumer.js';
 export {ContextProvider} from './lib/controllers/context-provider.js';
+export type {Options as ContextProviderOptions} from './lib/controllers/context-provider.js';
 export {ContextRoot} from './lib/context-root.js';
 
 export {provide} from './lib/decorators/provide.js';


adv0r:
Noted the CLA check — Google CLA is being signed under @adv0r and the check should turn green once the bot picks it up. Will leave a follow-up here if anything else is needed from my side.

(Tangentially: per AGENTS.md / CONTRIBUTING this PR is AI-generated under a personal token-burn initiative, see PR body for details. Happy to revise on review.)

PR #5344, opened 2026-05-23
fix(labs/router): correct enter() callback docs to match implementation

### LINKED ISSUE #5328: [labs/router] README documentation for enter() callback doesn't match implementation

### Description

The `@lit-labs/router` README's documentation of the `enter()` callback in route configurations does not accurately describe how `enter()` actually behaves in `routes.ts`.

Specifically:

1. **The README describes `enter()` as being able to "optionally reject that route as a match,"** which suggests the router falls through to the next matching route on rejection. The implementation does not do this — `routes.ts` strict-equality checks the return value against `false` and bails out of `goto()` entirely without trying any other routes:

```ts
   // packages/labs/router/src/routes.ts
   if (typeof route.enter === 'function') {
       const success = await route.enter(params);
       if (success === false) {
           return;
       }
   }
```

2. **The dynamic-routes example comment in the README** ("Reject this route so the dynamic one is matched") reinforces the same incorrect mental model. The example only works because it manually calls `goto()` again before returning `false` — not because returning `false` causes any fall-through.

3. **The return-value contract is undocumented.** Only strict `false` cancels; `undefined`, `null`, or no return value all proceed. The README never states this.

4. **Cancellation is silent.** No event, no thrown error — the `goto()` Promise still resolves. Callers have no signal that the navigation was cancelled. Undocumented.

5. **Error propagation through `goto()` is undocumented.** A thrown or rejected `enter()` surfaces out of `goto()` with no internal handling.

6. **The `enter` type signature isn't shown in the README.** The published `routes.d.ts` declares it as `(params) => Promise<boolean> | boolean`, but neither the type nor the param shape (`{[key: string]: string | undefined}`) appears in the docs.

I'd like to fix this with a docs-only PR that rewrites the `enter() callbacks` section to match the source. Happy to follow up with a separate PR for related stale typedefs (`PathRouteConfig` / `URLPatternRouteConfig` in the README don't include `enter` and use the wrong param type).

xxiaoxiong:
Fixes #5328

Why
The README for `@lit-labs/router` describes `enter()` as being able to "optionally reject that route as a match," which suggests the router falls through to the next matching route on rejection. In reality, returning `false` from `enter()` cancels the entire navigation (`goto()` returns without trying other routes).

Changes
Updated the description to reflect the actual behavior: "optionally cancel the navigation by returning `false`".

Updated component: `packages/labs/router/README.md`

Testing
Documentation only change. No code behavior was modified.

Surface area
- Labs / Router
- Documentation fix

### DIFF
--- a/packages/labs/router/README.md
+++ b/packages/labs/router/README.md
@@ -154,7 +154,7 @@ html`<main>${this.routes.outlet()}</main>`;
 
 #### enter() callbacks
 
-A route can define an `enter()` callback that lets it do work before rendering and optionally reject that route as a match.
+A route can define an `enter()` callback that lets it do work before rendering and optionally cancel the navigation by returning false.
 
 `enter()` can be used to load and wait for necessary component definitions:
 


xxiaoxiong:
Closing due to Google CLA signing requirement

PR #5350, opened 2026-06-08
Add ExportDecoratedItems for closure compiler compat

kevinpschaaf:
Fixes issue with closure compiler eliding readonly fields decorated with `@provide` / `@consume` decorators.

### DIFF
--- a/packages/context/src/lib/decorators/consume.ts
+++ b/packages/context/src/lib/decorators/consume.ts
@@ -39,6 +39,7 @@ import {Context} from '../create-context.js';
  * }
  * ```
  * @category Decorator
+ * @ExportDecoratedItems
  */
 export function consume<ValueType>({
   context,
--- a/.changeset/old-geese-try.md
+++ b/.changeset/old-geese-try.md
@@ -1,0 +1,5 @@
+---
+'@lit/context': patch
+---
+
+Add ExportDecoratedItems for closure compiler compatibility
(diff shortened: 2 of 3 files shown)


rictic:
I guess this is about `@provide` more than `@consume`, since if the consuming field isn't used anywhere it seems fine to elide it?

packages/context/src/lib/decorators/consume.ts
```diff
@@ -39,6 +39,7 @@ import {Context} from '../create-context.js';
  * }
  * ```
  * @category Decorator
+ * @ExportDecoratedItems
```
rictic, line 42: Do we need this on `@consume`?

rictic (changes requested):

packages/context/src/lib/decorators/consume.ts
kevinpschaaf: `@provide` is definitely vulnerable, but the OP bug report was actually on `@consume`.

Consider:

```ts
  @consume(someContext)
  private readonly someContext: string | undefined = undefined;

  someMethod() {
    if (this.someContext) {
      // Compiler sees nothing could ever set the field, so elides
      // the field and treats this code block as dead
    }
  }
}
```

rictic: Oh, interesting! Right!

rictic (approved):

PR #5356, opened 2026-06-23
[localize] Convert and expand localize tests

aomarks:
This adds a bunch of missing unit tests for localize, FNV-1A hash implementation, and a benchmark for the FNV-1A implementation. It also converts from uvu to the node test runner.

Also included a fix for a Vite Vue plugin test that was failing CI.


packages/labs/gen-wrapper-vue/src/lib/package-json-template.ts
```diff
@@ -43,7 +43,7 @@ export const packageJsonTemplate = (
         vite: '^5.3.1',
         'vue-tsc': '^2.0.22',
       },
-      files: [...moduleNames.map((f) => `${f}.*`)],
+      files: [...moduleNames.flatMap((f) => [`${f}.*`, `${f}2.*`])],
```
rictic: Why the 2 suffix here?

rictic (approved):

packages/labs/gen-wrapper-vue/src/lib/package-json-template.ts
aomarks: Vite sometimes splits up modules into files like Foo.vue.js and Foo2.vue.js, so the previous pattern was missing the "2" ones. Improved the fix further to just do a Foo* search instead of Foo.*.

PR #5367, opened 2026-07-25
[reactive-element] Fix useDefault with null and undefined initial values

### LINKED ISSUE #5366: `useDefault: true` doesn't work when a property is initialized to `null`

### Description

The docs for `useDefault` say a property must be initialized, but don't say the initial value can't be null or undefined and it can't: `useDefault` uses `??` to both record and restore the default value, so it can't tell "no default recorded" from "the default is `null`/`undefined`".

This causes bugs in a few cases:

- `null` initial value with standard decorators throws when the element is constructed.
- `foo: string | undefined = undefined`: removing the attribute sets the property to `null` rather than `undefined`.
- `accessor foo: string | undefined = undefined`, or a getter/setter pair: the default is never recorded at initialization (the init paths skip `undefined` values), so the first value the user sets is recorded as the "default".
  Removing the attribute then restores that value instead of `undefined`, i.e. it appears to do nothing.

### Reproduction

The null-default case:

```ts
class MyEl extends LitElement {
  @property({reflect: true, useDefault: true})
  accessor foo: string | null = null;
}

new MyEl();
// TypeError: Private element is not present on this object
//   at ReactiveElement._$changeProperty (reactive-element.ts:1339)
//   at init (decorators/property.ts:157)
```

### Workaround

Don't use null defaults?

### Is this a regression?

### Affected versions

3.3

n/a

justinfagnani:
Fixes https://github.com/lit/lit/issues/5366

Defaults were recorded and restored with `??` chains, which can't distinguish an unrecorded default from one that's `null` or `undefined`.

This change records the initial value explicitly and use `has()` when restoring it.

### DIFF
--- a/packages/reactive-element/src/reactive-element.ts
+++ b/packages/reactive-element/src/reactive-element.ts
@@ -295,7 +295,8 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
    *
    * When set, properties must be initialized, either with a field initializer, or an
    * assignment in the constructor. Not initializing the property may lead to
-   * improper handling of subsequent property assignments.
+   * improper handling of subsequent property assignments. The initial value may be
+   * `null` or `undefined`; it is restored as such when the attribute is removed.
    *
    * While this behavior is opt-in, most properties that reflect to attributes should
    * use `useDefault: true` so that their initial values do not reflect.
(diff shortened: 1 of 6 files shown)


bencbradshaw (approved):

packages/reactive-element/src/reactive-element.ts
```diff
@@ -1298,18 +1309,26 @@ export abstract class ReactiveElement
...
-          options.reflect &&
-          newValue === this.__defaultValues?.get(name) &&
-          !this.hasAttribute(ctor.__attributeNameForProperty(name, options)!));
+        // When there is no change, check the corner cases that can occur with
+        // `useDefault`.
+        (options.useDefault === true &&
+          // 1. The default value has not been recorded yet, which happens when
+          // a property is initialized to `undefined`. The default must be
+          // recorded so that it can be restored when an attribute is removed.
+          (defaultValues?.has(name) !== true ||
```
sorvell, line 1321: I'm not sure change properties are handled correctly for default values. Need to add tests for that. When `useDefault` is `true` and the property is not user set, it should not appear in changed properties.

PR #5368, opened 2026-07-29
Fix import statement for React module

stevematney:
React is used as a concrete object in this file, so it must be imported regularly (not as solely a type).


stevematney:
Actually, I just realized this is unnecessary. I misunderstood the flow of the code. (`React` defined as a type at the top level, and then `React` defined separately destructured from a function parameter.)

PR #5375, opened 2026-08-30
Make asyncAppend generic

### LINKED ISSUE #4819: Make asyncAppend generic
### Should this be an RFC?

### Description

Right now, `asyncAppend` types the parameter passed to the optional mapper function as `unknown` (`async-append.d.ts`):

```typescript
export declare const asyncAppend: (value: AsyncIterable<unknown>, _mapper?: ((v: unknown, index?: number) => unknown) | undefined) => import("../directive.js").DirectiveResult<typeof AsyncAppendDirective>;
// −−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−^^^^^^^
```

This is awkward, particularly with `strict: true` in TypeScript, if you want to do nearly anything with the value in the mapper, such as operate on it or use it as the value of a typed property.

Instead, if we make `asyncAppend` generic like `map` is, it's more convenient to use. Here's the change:

```typescript
export declare const asyncAppend: <T>(value: AsyncIterable<T>, _mapper?: ((v: T, index?: number) => unknown) | undefined) => import("../directive.js").DirectiveResult<typeof AsyncAppendDirective>;
// −−−−−−−−−−−−−−−−−−−−−−−−−−−−−−−^^^−−−−−−−−−−−−−−−−−−−−−−^−−−−−−−−−−−−−−−−−−^
```

I could be mistaken, but this should be very broadly backward-compatible with existing code, possibly entirely so.

I'm happy to provide a PR if it would be welcome.

### Documentation Example

My first motivating example is in [the documentation for `asyncAppend`](https://lit.dev/docs/templates/directives/#asyncappend), which produces an error if you have `strict: true` in `tsconfig.json`. Here's the documentation's example:

```typescript
async function *tossCoins(count: number) {
  for (let i=0; i<count; i++) {
    yield Math.random() > 0.5 ? 'Heads' : 'Tails';
    await new Promise((r) => setTimeout(r, 1000));
  }
}

@customElement('my-element')

  @state()
  private tosses = tossCoins(10);

  render() {
    return html`
      <ul>${asyncAppend(this.tosses, (v: string) => html`<li>${v}</li>`)}</ul>`;
  }
}
```

That produces the following error on the mapper callback in the `asyncAppend` call:

> Argument of type '(v: string) => TemplateResult<1>' is not assignable to parameter of type '(v: unknown, index?: number | undefined) => unknown'.
>&nbsp;&nbsp;Types of parameters 'v' and 'v' are incompatible.
>&nbsp;&nbsp;&nbsp;&nbsp;Type 'unknown' is not assignable to type 'string'.ts(2345)

[Here's that error in a TypeScript playground](https://www.typescriptlang.org/play/?experimentalDecorators=true#code/JYWwDg9gTgLgBAbzgCxiANgGjgGWDAUXQFMRiA7eAXzgDMoIQ4AidfZgbgChRJZE4AYwCuAZxiMipCjGziAhjGJwa9RizYwA9ABNig6IuiiAdACtRnHuGjwk80QE9yggIJgwFHSroMmrfF1gKH0YYAA3YlEtB2dBAFp5Dy9zS24uWJc6YRcwiHI4ACoJUVEAYQhgclEACgMcmAAuOHJhEAAjYigASkQuODpoOBqSeGAAXgAGDjhgAB56yhngAGoV3oR+gbhHYGJ0bwBZRWQTKHlyHUYa3oA+OEmTAFY4AH44AHIACWJ5HVEPnBmh8ACryYDoAHcbZweQAd3B8HIxDhcAACn5gKJiDUaj04ON7tiYCDQMQIMIYHjsABGSb07rdaEqLhULhcAACInEkhIZEoNQ+IEc8X20koH26XEE6Acojgh0cUn58GIAA8lJd5XhCHyZH0thyFEoblswFAIoplCVsfLxnAbRUqrU6Uz2QMQpcujc+jCQjBhFACqgMAADLYwubCdC3AAkCEybmSlxqMGQWJMNqi2Bq4Wa4gt5AA5r1CSg0OhQ3M2HGEOEqHMtDXQ90G1po7dQ8y2VQgA) (you can ignore the "Cannot find global type 'AsyncIterableIterator'." error in that playground). Alternatively, you can see it by cloning [this GitHub project](https://github.com/tjcrowder/lit-asyncappend-issue) and opening `src/my-element.ts`.

In that specific case, even with the existing definition you could just remove the type annotation on `v` in the mapper function, letting `v` be `unknown`, since all the mapper does is dump it out as text.

### My Use Case

My second motivating example is what brought this problem to my attention: I wanted to pass the mapper value to a component with a typed property:

```html
<div>
  ${asyncAppend(
    this.values,
    (v) => html`<example-element .someProperty=${v}></example-element>`
  )}
</div>
```

That fails because `someProperty` on `example-element` is of type `string`, and I can't directly assign `unknown` to it. I'd have to use `${v as string}` instead, which both is inconvenient and sacrifices type safety.

Please let me know if there's any further information etc. I can provide.

Thanks for Lit!

It's possible to work around issues with the parameter's type being `unknown` with type assertions and such, but type assertions tend to become maintenance hazards.<sup>[citation needed 😉]</sup>

vjymisal0:
Fixes #4819

## What

`asyncAppend`'s mapper callback was always typed as `(v: unknown, index?: number) => unknown`, even though the value passed to it always matches the async iterable's element type. This made it awkward to use under `strict: true`, as described in the issue (both the doc example and the "assign to a typed property" use case).

## Why this happened

`AsyncAppendDirective` inherits `render<T>(value: AsyncIterable<T>, _mapper?: Mapper<T>)` from `AsyncReplaceDirective` — `render` is already generic. But `asyncAppend` is produced via `directive(AsyncAppendDirective)`, and `directive()`'s exported function type is `Parameters<InstanceType<C>['render']>`. Extracting `Parameters<>` from a generic method collapses its type parameter, which is why the `.d.ts` shows `unknown` instead of the generic.

## Fix

Give `asyncAppend` an explicit generic call signature (`interface AsyncAppend { <T>(...): ... }`) and assign the `directive(AsyncAppendDirective)` value to it via `as`, the same approach already used for `guard` in `guard.ts` (whose exported `Guard` interface is generic for the same reason). A plain type annotation isn't enough here — because `T` appears in a contravariant (parameter) position in the mapper callback, TypeScript's structural check rejects the direct assignment, so an explicit cast is used instead. Runtime behavior is unchanged; this is a type-only fix.

## Tests

Added a compile-time-only case to `test/directives/types_are_inferrable_test.ts` (the file already used for exactly this kind of directive-type-inference check on `guard`, `classMap`, `keyed`, `live`, and `until`) asserting the mapper's parameter is inferred as the iterable's element type rather than `unknown`. Confirmed this test fails against the pre-fix code (`Type 'unknown' does not satisfy the expected type 'string'`) and passes after the fix.

`tsc --noEmit -p .` in `packages/lit-html` passes (aside from pre-existing, unrelated `@lit-labs/testing` module-resolution errors present on `main` too — that package isn't built in my environment). `eslint` and `prettier --check` both pass on the changed files.

I wasn't able to run the browser-based (`web-test-runner`) suite in this environment, but the fix is type-only and covered by the type-checker test above, which is how the existing directives with this same generic-inference concern (`guard`, `keyed`, `live`, `until`) are tested.

Added a changeset (patch) for `lit-html`, `lit-element`, and `lit`, per CONTRIBUTING.md's guidance to include core-library dependents.


vjymisal0:
I checked the old Google CLA failure link. The current blocking check is actually DCO (`ACTION_REQUIRED`), not an active Google CLA check; both PR commits have verified signatures but neither commit message contains a `Signed-off-by:` trailer. I have already signed the Google CLA, so the remaining fix is to rewrite the branch commits with the required DCO trailers and update the branch. I have not force-pushed that history without explicit approval.

vjymisal0:
DCO should be fixed now. I rewrote the branch to a single signed-off commit on current main.

